PR #4977, opened 2020-03-04
MM-23015: Enable or disable group mentions

catalintomai:
#### Summary
Adds support for:
1.  Enabling or disableing group mentions in System Console
2. adding group mentions to the list of suggestions in the channel textbox

Note: Includes webapp changes for https://mattermost.atlassian.net/browse/MM-23016 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23015

#### Related Pull Requests
redux: https://github.com/mattermost/mattermost-redux/pull/1066
server: https://github.com/mattermost/mattermost-server/pull/14010


components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -112,8 +117,15 @@ export default class GroupDetails extends React.PureComponent {
         this.setState({loadingTeamsAndChannels: false});
     }
 
+    onToggle = async (allowReference) => {
+        this.setState({allowReference});
+        await this.props.actions.patchGroup(this.props.groupID, {allow_reference: allowReference});
+    }
```
mkraft: Does `onToggle` need to `await` patchGroup? I wasn't seeing the response being used.

components/admin_console/group_settings/group_details/group_profile_and_settings.jsx
```diff
@@ -0,0 +1,81 @@
...
+    <LineSwitch
+        disabled={isDefault}
+        toggled={allowReference}
+        last={allowReference}
+        onToggle={() => {
+            if (isDefault) {
+                return;
+            }
+            onToggle(!allowReference);
+        }}
```
mkraft: [The React docs](https://reactjs.org/docs/handling-events.html) recommend against inline functions because...

> a different callback is created each time [it] renders

... whether this affects performance in this particular case, I don't know.

components/textbox/index.js
```diff
@@ -4,22 +4,46 @@
...
     const getProfilesInChannel = makeGetProfilesInChannel();
     const getProfilesNotInChannel = makeGetProfilesNotInChannel();
 
-    return (state, ownProps) => ({
+    const team = getCurrentTeam(state);
+    const channel = getCurrentChannel(state);
+
+    let autocompleteGroups;
+
+    if(team && team.group_constrained && channel && channel.group_constrained) {
```
mkraft: I think you can do this because we have [optional chaining enabled in our babel config](https://github.com/mattermost/mattermost-webapp/blob/master/babel.config.js#L33).
```suggestion
    if(team?.group_constrained && channel?.group_constrained) {
```

components/textbox/index.js
```diff
@@ -4,22 +4,46 @@
...
+        let groupsFromChannel = getGroupsAssociatedToChannelForReference(state, channel.id);
+        let groupsFromTeam = getGroupsAssociatedToTeamForReference(state, team.id);
+        autocompleteGroups = groupsFromChannel.concat(groupsFromTeam.filter((item) => groupsFromChannel.indexOf(item) < 0));
+    } else if(team && team.group_constrained) {
```
mkraft:
```suggestion
    } else if(team?.group_constrained) {
```

components/textbox/index.js
```diff
@@ -4,22 +4,46 @@
...
+        autocompleteGroups = getGroupsAssociatedToTeamForReference(state, team.id);
+    } else if(channel && channel.group_constrained) {
```
mkraft:
```suggestion
    } else if(channel?.group_constrained) {
```

components/needs_team/needs_team.jsx
```diff
@@ -205,6 +207,11 @@ export default class NeedsTeam extends React.Component {
         this.props.actions.loadStatusesForChannelAndSidebar();
         this.props.actions.loadProfilesForDirect();
 
+        this.props.actions.getAllGroupsAssociatedToChannelsInTeam(team.id);
+        if (team.group_constrained) {
+            this.props.actions.getAllGroupsAssociatedToTeam(team.id);
+        }
```
catalintomai: need to add getGroups() call for the case where neither the team nor the channels in the team are synced - could be expensive to do it for every team switch.

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -34,34 +34,39 @@ export default class GroupDetails extends React.PureComponent {
...
         groupTeams: [],
         groupChannels: [],
-        group: {display_name: ''},
+        group: {display_name: '', allow_reference: false},
         memberCount: 0,
     };
 
     constructor(props) {
         super(props);
+        const group = props.group;
```
fm2munsh: I dont see the value in this creating this variable could just do 
`Boolean(props.group.allow_reference),` since there is only a single props attribute being referenced

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -34,34 +34,39 @@ export default class GroupDetails extends React.PureComponent {
...
         this.state = {
             loadingTeamsAndChannels: true,
             addTeamOpen: false,
             addChannelOpen: false,
+            allowReference: Boolean(group.allow_reference),
         };
     }
 
     componentDidMount() {
         const {groupID, actions} = this.props;
```
fm2munsh: could change this to `const {groupID, actions, group} = this.props;` 
And then modify line 69 to be
```
this.setState({group, allowReference: Boolean(group.allow_reference)});
```

components/admin_console/group_settings/group_details/group_details.jsx
fm2munsh: 👍 could drop the async / await on this method - unless theres something im missing

components/suggestion/at_mention_provider/at_mention_provider.jsx
```diff
@@ -162,6 +202,9 @@ export default class AtMentionProvider extends Provider {
             if (user.username) {
                 return '@' + user.username;
             }
+            if (user.display_name) {
+                return '@' + user.display_name;
+            }
```
fm2munsh: This would be a `group` right? maybe the code should be updated to be resource agnostic into something like `item` or `userOrGroup`

catalintomai: yes, it is on my list of things to refactor here.

components/suggestion/at_mention_provider/at_mention_suggestion.jsx
```diff
@@ -82,6 +84,22 @@ export default class AtMentionSuggestion extends Suggestion {
                     )}
                 </FormattedMessage>
             );
+        } else if (user.type === Constants.MENTION_GROUPS) {
+            username = user.display_name;
+            description = `- (${user.display_name})`;
+            icon = (
+                <FormattedMessage
```
fm2munsh: same here - renaming `user` to something more appropriate would make it less confusing i think

catalintomai: agree

components/textbox/textbox.jsx
```diff
@@ -42,6 +42,7 @@ export default class Textbox extends React.PureComponent {
         preview: PropTypes.bool,
```
fm2munsh: Somewhere in this file we would want to check for the group mentions permission in order to not show autocomplete if the user does not have permission to groups

catalintomai: wouldn't that be covered by the permission gating of the API call to the store though? In that case the list of groups would be empty and there would be nothing to show.

fm2munsh: ah true - would and the entire section be hidden if there were no groups right?

catalintomai: I will double check, but my understanding is the section header will not be present is there are no items of that type (group mentions here) to suggest.

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: agree.

components/admin_console/group_settings/group_details/group_profile_and_settings.jsx
catalintomai: This looks to be the pattern when using the LineSwitch component (see team_modes, channel_modes). I say if we were to change this to make it more performant we should do it across the board (maybe a community ticket?). Will defer the change till then.

components/textbox/index.js
catalintomai: moved this logic to redux, where for some reason I cannot do `?.`, so it'll have to stay this way fo now.

lieut-data:
@catalintomai, for context, is this a feature branch you're merging to master, or the "initial implementation"? My reviewing strategy is similar, but less exhaustive for the former, since I'm assuming you've already gotten a peer review on code style, etc.

catalintomai:
> @catalintomai, for context, is this a feature branch you're merging to master, or the "initial implementation"? My reviewing strategy is similar, but less exhaustive for the former, since I'm assuming you've already gotten a peer review on code style, etc.

This is the initial cut, to be merged into master after all code reviews (we don't stage them necessarily) and QA test passes, feel free to be as exhaustive as you need to be. Thanks.

catalintomai:
> Can you resolve the errors that are causing the build to fail

I believe they are there because the redux changes are not merged in yet. Do you see other reasons?

catalintomai:
> > @catalintomai, for context, is this a feature branch you're merging to master, or the "initial implementation"? My reviewing strategy is similar, but less exhaustive for the former, since I'm assuming you've already gotten a peer review on code style, etc.
> 
> @lieut-data This is the initial cut, to be merged into master after all code reviews (we don't stage them necessarily) and QA test passes, feel free to be as exhaustive as you need to be. Thanks.

@lieut-data, we are trying to have this for 5.22 so if you can take a look and detect anything blocking, it would be very helpful to find out soon. We can tackle smaller issues next week. Thanks.

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -51,17 +52,20 @@ export default class GroupDetails extends React.PureComponent {
...
-        const {groupID, actions} = this.props;
+        const {groupID, actions, group} = this.props;
         actions.getGroup(groupID);
+
         Promise.all([
             actions.getGroupSyncables(groupID, Groups.SYNCABLE_TYPE_TEAM),
             actions.getGroupSyncables(groupID, Groups.SYNCABLE_TYPE_CHANNEL),
         ]).then(() => {
             this.setState({loadingTeamsAndChannels: false});
+            this.setState({group, allowReference: Boolean(this.props.group.allow_reference)});
```
lieut-data: "Duplicate" call to `setState` here -- I suspect this is a non-issue, but just highlighting in case it makes more sense to merge with the line above.

components/admin_console/group_settings/group_details/group_profile_and_settings.test.jsx
```diff
@@ -0,0 +1,31 @@
...
+describe('components/admin_console/group_settings/group_details/GroupProfileAndSettings', () => {
+    test('should match snapshot, with toggle off', () => {
+        const wrapper = shallow(
+            <GroupProfileAndSettings
+                displayname='GroupProfileAndSettings'
+                name='GroupProfileAndSettings'
+                allowReference={false}
+                onToggle={jest.fn()}
+            />);
+        expect(wrapper).toMatchSnapshot();
```
lieut-data: nit: Would love to see explicit assertions vs. generic snapshot matches: I've historically found the snapshots too easy to change without noticing.

components/suggestion/at_mention_provider/at_mention_provider.jsx
```diff
@@ -68,6 +70,27 @@ export default class AtMentionProvider extends Provider {
...
+        const groupSuggestions = [];
+        if (!group) {
+            return groupSuggestions;
+        }
+
+        if (group.name) {
+            const groupnameSuggestions = getSuggestionsSplitByMultiple(group.name.toLowerCase(), Constants.AUTOCOMPLETE_SPLIT_CHARACTERS);
+            groupSuggestions.push(...groupnameSuggestions);
+        }
+        [group.display_name].forEach((property) => {
```
lieut-data: I'm stumbling over this construct -- why doesn't this look like the `group.name` block above?

catalintomai:
> A few comments below, but nothing strictly blocking. 👍

Thanks!

components/needs_team/needs_team.tsx
```diff
@@ -224,6 +227,13 @@ export default class NeedsTeam extends React.Component<Props, State> {
         this.props.actions.loadStatusesForChannelAndSidebar();
         this.props.actions.loadProfilesForDirect();
 
+        this.props.actions.getAllGroupsAssociatedToChannelsInTeam(team.id, true);
+        if (team.group_constrained) {
+            this.props.actions.getAllGroupsAssociatedToTeam(team.id, true);
+        } else {
+            this.props.actions.getGroups(true);
```
mkraft: As discussed, I think this call to `getGroups` should be invoked once rather than each time the user changes teams.

catalintomai: agree

srkgupta:
Hi @catalintomai 
Can you please resolve the merge conflicts and also check that the tests are passing.

catalintomai:
> Hi @catalintomai
> Can you please resolve the merge conflicts and also check that the tests are passing.

I've resolved the conflicts. The tests are not passing because the redux changes are missing. Once they are checked-in then they should pass.

fm2munsh:
@catalintomai couldnt you update the `package.json` in this PR to point to the redux commit hash that it needs?

srkgupta:
Hello @catalintomai 
The PR is approved. Please fix the failing tests and merge the PR.

PR #4978, opened 2020-03-05
[MM-23018] Add use_channel_mentions to system and team permission schemes

fm2munsh:
#### Summary
- Adds the new channel mentions permission to the edit permission scheme pages for both system and team schemes
- Not sure about the wording or placement of this permission @michaelgamble let me know what the actual description and name should be and where in the list it should show up 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-23018

#### Screenshots
Screen Shot 2020-03-04 at 7 10 13 PM
Screen Shot 2020-03-04 at 7 10 06 PM


fm2munsh:
Will add `Dev Review` and `QA Review` labels once UX has approved

srkgupta:
@fm2munsh Please ignore the issue #2 mentioned above. I was seeing incorrect results because the user was actually a Channel Admin. I am resuming my testing and will test the different combinations and will post any other issues as separate comment.

srkgupta:
@fm2munsh Some other issues:

2) In the System Schemes, if Channel Mentions is disabled for All Members and Channel Admins and allowed for Team Admin, when a Team Admin tries to use a Channel Mention, it does not displays in the autocomplete dropdown. However if the Team Admin uses the channel mentions and sends a message, other user gets notified as expected. So the issue is only with the autocomplete and highlight.
Screenshot 2020-03-12 at 9 11 13 PM
Screenshot 2020-03-12 at 9 12 47 PM

3) Can you please add some Cypress E2E tests for this feature. Some of the tests which you can add are:
* Change the values in System Scheme page, save and reload and check if values are saved. 
* Change the values in Team Scheme page, save and reload and check if values are saved. 
* Removing Permission for All Members in System Scheme and check if permission is removed and a regular member is not able to use channel mentions. 
* Removing Permission for Channel Admin & Team Admin in Team Scheme and check if permission is removed and a channel admin or team admin is not able to use channel mentions.

srkgupta:
4. Channel Mentions System Scheme is not applied for Guest users if Team Override Scheme exists

I guess this happened because there was an existing Team Override scheme present in the system and then when this PR was merged and the permission for Channel Mentions for Guests was removed, the changes were not applied to the existing Team Override Scheme.  When Channel Mentions is removed for Guest Users at System Scheme level, a guest user is still allowed to use the channel mentions as shown in this screen. 
Screenshot 2020-03-12 at 10 19 42 PM

Screenshot 2020-03-12 at 10 30 55 PM

fm2munsh:
#4 is related to the issue we discussed earlier - and not related to this PR its a behaviour that will happen with any of the guest permissions

fm2munsh:
For issue # 1 that is by design - the channel admin / team admin roles revert back to their previous state if a permission is unchecked at the member level

So if
1. `Channel Mentions` is removed at the member level
2. `Channel Mentions` is removed at the team admin level 
3. `Channel Mentions` is enabled at the member level
4. `Channel Mentions` will be enabled at the team admin level now 
5. `Channel Mentions` is removed at the member level again
6. `Channel Mentions` will be removed at the team admin level (reverting back to its state in step 2)

fm2munsh:
Also is issue number 2 a duplicate of issue number 6 in this PR: https://github.com/mattermost/mattermost-server/pull/13813?

srkgupta:
@fm2munsh regarding issue #1, I guess there is something wrong with the default permissions in the system scheme. By default, channel Admins and Team Admins should have the new permission enabled. However as per your explanation, it looks like it doesn't and hence when we remove the permission for All Members, it gets removed for Channel Admin and Team Admin. Can you please let me know if this needs to be raised as a separate ticket or can be handled in this one?

srkgupta:
Regarding issue #2, it is completely different from the issue #6 in the other PR. Please note that in this case, it's a system scheme and has nothing to do with the channel moderations. Basically when the channel mentions permission is allowed to Team Admins in System schemes (and no other permissions are changed), then the team admin should still be able to use the channel mentions. However it is observed that for Team Admin, the channel mentions is not shown at the UI side. However it is properly handled in the server side.

fm2munsh:
Yea issue 1 needs to be raised as a separate ticket @srkgupta 
issue 2 should be resolved in the latest commit

srkgupta:
@fm2munsh here is the update:

Issue 1 has been raised separately here:
https://mattermost.atlassian.net/browse/MM-23430

Issue 2 is fixed and is working fine now. 

Waiting for Cypress tests to be added as mentioned in Issue 3. Please let me know once this is added so that the PR can be approved. 

Issue 4 has been filed separately here:
https://mattermost.atlassian.net/browse/MM-23197

hahmadia:
Is it possible to update the title + Jira ticket to mention that this implements use_channel_mentions for both System Scheme and Team Override Scheme. Thanks! 😄

e2e/cypress/integration/enterprise/system_console/system_scheme_spec.js
```diff
@@ -0,0 +1,231 @@
...
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+import users from '../../../fixtures/users.json';
+
+import {deleteExistingTeamOverrideSchemes} from './team_scheme_spec.js';
```
srkgupta: When we import a specific function from another test, Cypress runs the entire imported test like this:
Screenshot 2020-03-24 at 4 32 41 PM

So it is suggested to have the function local in multiple tests even if its redundant.

e2e/cypress/integration/enterprise/system_console/team_scheme_spec.js
```diff
@@ -0,0 +1,176 @@
...
+
+        // # Save config
+        cy.get('#saveSetting').click();
+
+        // * Ensure that the team scheme was created and assigned to the team
+        cy.findByTestId('permissions-scheme-summary').within(() => {
+            cy.get('.permissions-scheme-summary--header').should('include.text', 'Test Team Scheme');
+            cy.get('.permissions-scheme-summary--teams').should('include.text', 'eligendi');
+        });
+    });
```
srkgupta: Since this is a generic testcase, you can leave this test here in this file `team_scheme_spec.js` as it makes more sense and can move the other test to a different file `channel_mentions_team_scheme_spec.js`

fm2munsh: my assumption was that eventually we would add more permission tests to this same file - splitting out half the tests into a different file would result in a bunch of duplication - thoughts @srkgupta?

e2e/cypress/integration/enterprise/system_console/system_scheme_spec.js
```diff
@@ -0,0 +1,231 @@
...
+
+    // # Click reset to defaults and confirm
+    cy.findByTestId('resetPermissionsToDefault').click();
+    cy.get('#confirmModalButton').click();
+
+    // # Save
+    saveConfig();
+};
+
+describe('System Scheme Permissions Test', () => {
```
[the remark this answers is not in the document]
fm2munsh: My assumption here is that more tests for system scheme permissions will be eventually added to this file too which is why I named it `system_scheme_spec` - is that a wrong assumption do we want a different file for each permission tested? maybe a folder `system_scheme/use_channel_mentions_spec.js` makes more sense? 

Would still result in a tonne of duplicated code due to the fact that we cant export functions

e2e/cypress/integration/enterprise/system_console/system_scheme_spec.js
fm2munsh: done

e2e/cypress/integration/enterprise/system_console/team_scheme_spec.js
fm2munsh: I decided to keep this test case in `use_channel_mentions_scheme_spec.js` since it is required as setup for the `use_channel_mentions` part of the test - maybe later it can be extracted into a separate file

PR #4985, opened 2020-03-05
MM-22843 - Updating RHS size on wider screens

asaadmahmood:
#### Summary
MM-22843 - Updating RHS size on wider screens

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22843

### DIFF
--- a/sass/responsive/_desktop.scss
+++ b/sass/responsive/_desktop.scss
@@ -3,27 +3,20 @@
 @media screen and (min-width: 1680px) {
     .sidebar--right {
         width: 500px;
-        @include translateX(100%);
-
-        &.move--left {
-            @include translateX(0);
-        }
     }
 
-    .inner-wrap {
-        &.move--left {
-            .search__form {
-                width: 260px;
-            }
+    .move--left {
+        .search__form {
+            width: 260px;
+        }
 
-            .app__content {
-                > #post-list {
-                    margin-right: 500px;
-                }
+        #post-list {
+            margin-right: 500px;
+        }
 
-                > .post-create__container {
-                    width: calc(100% - 500px);
-                }
+        .app__content {
+            > .post-create__container {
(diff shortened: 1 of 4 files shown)


sass/responsive/_desktop.scss
```diff
@@ -1,5 +1,34 @@
...
+    .inner-wrap {
+        &.move--left {
+            .search__form {
+                width: 260px;
+            }
+
+            .app__content {
+                > #post-list {
+                    margin-right: 500px !important;
+                }
```
nevyangelova: I would stay away from ```!important``` and look into the selector specificity in order to make this work correctly.

sass/responsive/_desktop.scss
```diff
@@ -1,5 +1,34 @@
...
+
+                > #post-create {
+                    width: calc(100% - 500px) !important;
```
nevyangelova: Same here

sass/responsive/_desktop.scss
```diff
@@ -1,5 +1,34 @@
...
+        @include translateX(100%);
+
+        &.move--left {
+            @include translateX(0);
+        }
+    }
+
+    .inner-wrap {
+        &.move--left {
+            .search__form {
```
nevyangelova: Consider avoiding nesting and using classes instead.

asaadmahmood: @nevyangelova I'm sorry, I don't understand. Isn't nesting already using classes?
Can you provide an example of what you mean?

asaadmahmood: I tried simplifying it, but the reason why we have `move--left` is because that class is added dynamically, and we want to apply certain styles only if that class is there.

And the nesting for `app__content` is there because there are 2 `post-create` containers within the app, and I only want to target the one directly within `app__content`.

PR #4988, opened 2020-03-05
MM-22754 Fix for missing closing bracket

sudheerDev:
#### Summary
Adding the missing bracket in translation files

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22754

PR #4402, opened 2019-12-02
[GH-8665/MM-9832]: Refactor utility functions that use getState and upd…

nevyangelova:
#### Summary
This PR introduces a refactor of utility functions, which would call getState. I've been passing more things into the functions, in order to not use the whole state in them, and in some cases I chose to call them in mapStateToProps. Some functions were redundant so I ended up deleting them. I am opening this as WIP as there are a lot of changes and I am hoping to start some discussions and see if I am on the right track. I haven't updated any tests yet so the builds will fail. 

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/8665

PR #4992, opened 2020-03-05
Define handleFormattedTextClick as an arrow function

nevyangelova:
https://mattermost.atlassian.net/browse/MM-23045

Fixes https://github.com/mattermost/mattermost-webapp/pull/4988#issuecomment-595347421 

this was introduced after merging https://github.com/mattermost/mattermost-webapp/pull/4402

### DIFF
--- a/components/post_view/message_attachments/message_attachment/message_attachment.jsx
+++ b/components/post_view/message_attachments/message_attachment/message_attachment.jsx
@@ -242,9 +242,7 @@ export default class MessageAttachment extends React.PureComponent {
         );
     };
 
-    handleFormattedTextClick(e) {
-        handleFormattedTextClick(e, this.props.currentRelativeTeamUrl);
-    }
+    handleFormattedTextClick = (e) => handleFormattedTextClick(e, this.props.currentRelativeTeamUrl);
 
     render() {
         const {attachment, options} = this.props;
--- a/components/post_view/message_attachments/message_attachment/message_attachment.jsx
+++ b/components/post_view/message_attachments/message_attachment/message_attachment.jsx
@@ -7,7 +7,7 @@ import truncate from 'lodash/truncate';
 
 import {isUrlSafe} from 'utils/url';
 import {Constants} from 'utils/constants';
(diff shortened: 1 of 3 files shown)


components/post_view/message_attachments/message_attachment/message_attachment.jsx
```diff
@@ -242,9 +242,7 @@ export default class MessageAttachment extends React.PureComponent {
         );
     };
 
-    handleFormattedTextClick(e) {
-        handleFormattedTextClick(e, this.props.currentRelativeTeamUrl);
-    }
+    handleFormattedTextClick = (e) => handleFormattedTextClick(e, this.props.currentRelativeTeamUrl);
```
Willyfrog: can we change it like the others into Util.handleFormattedTextClick for the imported one so it is easy to differentiate between them?
I know it was like that earlier, but leaving it a bit more readable would be great.

jgilliam17:
Thank you @nevyangelova 
Alt+click on user posts works as expected, no error.
Following errors are related to alt+click on the system message at the top of the channel and the system message with (Only visible to you) text. 

Note: Edge cases, as user may alt+click on sys message by accident. I checked and these errors are also present on the current release server, so I'm unsure if it should be fixed here, filed separately or needs further discussion. @sudheerDev, @Willyfrog what are your thoughts? Thanks.

Screen Shot 2020-03-07 at 8 50 25 AM
Screen Shot 2020-03-07 at 8 49 41 AM

sudheerDev:
>  checked and these errors are also present on the current release server, so I'm unsure if it should be fixed here, filed separately

@jgilliam17 I think we should file them separately. The changes that caused this regression are only present in master. 

Just to be clear are you saying the above error is present in 5.21v?

jgilliam17:
Yes, I can see this alt+click on system message error on 5.21 release server.

sudheerDev:
@jgilliam17 Ya, I suggest we log it as a separate ticket and i just checked and it isn't a regression either.

PR #4916, opened 2020-02-16
[MM-22622] Add mentionHighlightDisabled props on create post and create comment

fm2munsh:
#### Summary
- Adds a check in the markdown component to set `options.mentionHighlight` to `false` if  `props.mentionHighlightDisabled` is `true`
- Sets `mentionHighlightDisabled` to `true` when creating a comment or post 
(this is also checked on the server but it is also needed on the webapp or else the `@ channel` text will temporarily flash as highlighted before the message gets validated on the server) 
- Tests

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22622

#### Related Pull Requests
- Has Server changes: https://github.com/mattermost/mattermost-server/pull/13902

#### Screenshots
Screen Shot 2020-02-15 at 3 12 47 PM

PR #4819, opened 2020-02-03
MM-22017 Fix Markdown component not respecting mentionHighlight flag

jwilander:
#### Summary
When `mentionHighlight: false` was passed as an option to our text formatting function it would output HTML that did not highlight any mentions for the current user but the AtMention component created by that HTML being passed to our `messageHtmlToComponent` function would not take that into account and would highlight the mentions for the current user anyways.

@hmhealey the patch works but please let me know if there's a better way I should go about this

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22017

PR #4994, opened 2020-03-06
MM-22970: Fix highlighting of at mentions of self

sbishel:
#### Summary
This PR fixes and issue with highlighting at mention myself followed by a Period(.) or underscore (_). This issue has already been fixed in master by this PR -
https://github.com/mattermost/mattermost-webapp/pull/4916

However, that PR, was for another feature and is too large to cherry-pick back to v5.21. This PR only contains the changes necessary to fix the highlighting issue.

This issue was introduced when fixing the highlighting of the header - https://github.com/mattermost/mattermost-webapp/pull/4819
I made sure this fix was retained, see screenshot. 

Essentially, when the `options.mentionHighlight` is not set, it gets passed as false. Therefore, when it is empty, it needs set to true by default. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22970

#### Screenshots
Screen Shot 2020-03-05 at 8 16 45 PM

### DIFF
--- a/components/post_markdown/__snapshots__/post_markdown.test.jsx.snap
+++ b/components/post_markdown/__snapshots__/post_markdown.test.jsx.snap
@@ -12,6 +12,11 @@ exports[`components/PostMarkdown plugin hooks can build upon other hook message 
   imageProps={Object {}}
   isRHS={false}
   message="hello world!"
+  options={
+    Object {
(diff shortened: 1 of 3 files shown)


amyblais:
@sbishel This needs to be cherry picked.

PR #4998, opened 2020-03-06
Merging of CSS styles into new sidebar code

devinbinnie:
#### Summary
Adding in of styles provided by @bradjcoughlin. Some tweaking done and a few other fixes to the logic.

NOTE: Currently this isn't perfect visually speaking, but I'm hoping to have this merged so we can spin up an instance for Eric and Andrew to review.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22464


components/sidebar/sidebar_channel/sidebar_direct_channel/sidebar_direct_channel.tsx
```diff
@@ -92,10 +92,19 @@ export default class SidebarDirectChannel extends React.PureComponent<Props, Sta
             return icon;
         }
 
+        let className = '';
```
hmhealey: Shouldn't this be part of the StatusIconNew itself? Are we planning on using it somewhere where it wouldn't be using these class names?

devinbinnie: It's because our `status-x` classes belong to the sidebar, and I'm not sure what the reusability of the `StatusIconNew` component will be. I know we're moving a lot of our icons over to the new Mattermost designed ones, so I figured I would keep it as vanilla as possible and let other components dictate the behaviour, at least for now.

PR #5005, opened 2020-03-06
[MM-22478] Unread Filtering and Create New Channel Menu

devinbinnie:
#### Summary
This adds the unread filtering and create new channel menu to the sidebar.
It also hooks up the unread filter and the collapsing of categories to redux.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22478


components/sidebar/sidebar_category/index.ts
```diff
@@ -4,24 +4,28 @@
...
+import {setCategoryCollapsed} from 'actions/views/channel_sidebar';
+import {isCategoryCollapsed} from 'selectors/views/channel_sidebar';
+import {GlobalState} from 'types/store';
+
 import SidebarCategory from './sidebar_category';
 
-function mockSetCollapsedState(categoryId: string, isCollapsed: boolean) {
-    return (dispatch: any, getState: any) => {};
+type OwnProps = {
+    category: any;
```
hmhealey: This can use the `Category` type from `mattermost-redux/types/channel_categories`

components/sidebar/sidebar_category_list/index.ts
```diff
@@ -35,7 +37,7 @@ function getCategoryFromChannel(channelCategories: ChannelCategory[]) {
 const categoriesFunc = memoizeResult(getCategoryFromChannel);
 
 // TODO: temp typing until we fix redux
```
hmhealey: This comment can be removed

components/sidebar/channel_filter/channel_filter.tsx
```diff
@@ -2,25 +2,80 @@
...
+        this.props.actions.setUnreadFilterEnabled(!unreadFilterEnabled);
+    }
+
     render() {
+        const {unreadFilterEnabled} = this.props;
+
+        let filterTitle = (
+            <FormattedMessage
+                id='sidebar_left.channel_filter.viewing'
+                defaultMessage='VIEWING:'
```
hmhealey: Would it make sense to capitalize these with a `text-transform` instead of having the translation strings be in all caps? 0/5

components/sidebar/channel_filter/channel_filter.tsx
```diff
@@ -2,25 +2,80 @@
 // See LICENSE.txt for license information.
 
 import React from 'react';
+import {FormattedMessage} from 'react-intl';
+import classNames from 'classnames';
```
hmhealey: Are you adding this library somewhere? I don't see it in the package.json

components/sidebar/channel_filter/channel_filter.tsx
devinbinnie: I'd say no, because then different languages can't choose whether capitalization matters to them or not. This is more customizable per language, and is also much easier to understand, since a `text-transform` is hidden away in the CSS and isn't immediately apparent.

components/sidebar/channel_filter/channel_filter.tsx
devinbinnie: `classnames` is used in a lot of our components. It's just a simple way of adding classes via booleans. It's been around a while AFAIK.

components/sidebar/channel_filter/channel_filter.tsx
hmhealey: I don't think that'd necessarily be something we care to change per-language though since using all caps is an aesthetic decision. 0/5 still though

components/sidebar/channel_filter/channel_filter.tsx
hmhealey: Huh, strange. It looks like it's a dependency-of-a-dependency that someone started using at some point, so that's why it's not in the package.json

devinbinnie: Oh, we should add that into `package.json` then.

devinbinnie:
Whoops, should have waited for Brad. @bradjcoughlin if there's anything about this PR that seems off let me know.

components/sidebar/channel_filter/channel_filter.tsx
bradjcoughlin: I prefer the text-transform for strings usually FWIW, but probably not a blocker at this point.

bradjcoughlin:
No worries, I had one comment about the uppercase vs. title case question. I prefer using title case and letting CSS specify uppercase as @hmhealey suggested but I don’t think it’s a blocker at this stage. This just allows for more flexibility in case visual guidance changes from the UI design side.

PR #5016, opened 2020-03-09
Mobile/Desktop Sidebar View + Restored Team Sidebar and Header

devinbinnie:
#### Summary
Mobile/desktop sidebar view fixes so that it works correctly.
Also added back the team sidebar and sidebar header since we aren't changing those in Phase 1.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22835


sass/responsive/_mobile.scss
```diff
@@ -548,10 +548,13 @@
     }
 
     .sidebar--left .team__header .Avatar,
+    #SidebarContainer .team__header .Avatar,
     .sidebar--menu .team__header .Avatar {
         display: none;
     }
 
+    Unread
```
hmhealey: Looks like this was added accidentally

PR #5019, opened 2020-03-09
[MM-23103] Fix redirection after editing the channel you were just viewing

hahmadia:
**Note To Self**
- [x] Update the redux hash to master after redux commit is merged to master

#### Summary

This addresses the problem where when a user edits a channel's setting and saves (System Console > Channel) for the last channel they were viewing, they will be redirected to that channel.

The cause of this problem is due to the fact that `currentChannelId` in the state does not get reset when they visit the System Console therefore when they edit the channel they last viewed, the following gets triggered which sends them back to the channel they just viewed ( https://github.com/mattermost/mattermost-webapp/blob/master/actions/websocket_actions.jsx#L451 ) 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23103

#### Related Pull Requests

- Has redux changes https://github.com/mattermost/mattermost-redux/pull/1070

#### Screenshots

When Admin visits the system console, the following action is emitted and the state gets reset.

image
image


hmhealey:
Regarding the changes suggested in the other PR, you could also define an action creator somewhere that just calls `selectChannel('')` if you'd prefer to keep the nicer name.

hahmadia:
@hmhealey 
I have updated the PR and addressed your comments. I think it's okay to do it without a nicer name cause now we are resetting both team and channel ID so I think doing it the way you requested is fine.

hmhealey:
Looks good. Like I mentioned on the redux PR, I might be taking a look at modifying the SELECT_CHANNEL and SELECT_TEAM actions at some point soon anyway so that we can implement a channel history for the desktop app, so those functions may change or get renamed at some point anyway.

PR #5025, opened 2020-03-09
[MM-22360] - Handle closing of editPostModal when user has unsaved changes

nevyangelova:
#### Summary

This fixes the closing of the editPostModal when the user has unsaved changes. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22360

### DIFF
--- a/components/edit_post_modal/edit_post_modal.jsx
+++ b/components/edit_post_modal/edit_post_modal.jsx
@@ -251,12 +251,12 @@ class EditPostModal extends React.PureComponent {
         this.props.actions.hideEditPostModal();
     }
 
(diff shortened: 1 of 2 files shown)


components/edit_post_modal/edit_post_modal.jsx
```diff
@@ -251,6 +251,14 @@ class EditPostModal extends React.PureComponent {
         this.props.actions.hideEditPostModal();
     }
 
+    handleCheckForChangesHide = (doRefocus = true) => {
```
sudheerDev: This is nitpicking but Why even bother adding the `doRefocus` here? it is just called once and you can just call `this.handleHide()`

PR #4931, opened 2020-02-19
[MM-22498] Refactor channel header popover to match requirements

nevyangelova:
#### Summary
Refactored the popover for channel header info to match the desired behaviour linked in the Jira ticket

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22498

PR #5029, opened 2020-03-10
MM-22986 - Updating channel header interaction

asaadmahmood:
#### Summary
MM-22986 - Updating channel header interaction

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22986

#### Screenshots
edit


i18n/en.json
```diff
@@ -2203,6 +2203,7 @@
   "edit_channel_header_modal.title": "Edit Header for {channel}",
   "edit_channel_header_modal.title_dm": "Edit Header",
   "edit_channel_header.editHeader": "Edit the Channel Header...",
+  "channel_header.editLink": "(Edit)",
```
jespino: This is not in the place that is expected (it must be alphabetically ordered by id in the file).

components/channel_header/channel_header.js
```diff
@@ -478,6 +478,12 @@ class ChannelHeader extends React.PureComponent {
                                     id='channel_header.addChannelHeader'
                                     defaultMessage='Add a channel description'
                                 />
+                                <span className='button__edit ml-1'>
+                                    <FormattedMessage
+                                        id='channel_header.editLink'
+                                        defaultMessage='(Edit)'
+                                    />
+                                </span>
```
cometkim: This will probably create a nested span tag because [we use it as default](https://github.com/mattermost/mattermost-webapp/blob/7b65d4e641b17023d30e828e09e1eead143109ba/components/intl_provider/intl_provider.jsx#L71).

so this would be a bit better:

```suggestion
                                <FormattedMessage
                                    id='channel_header.editLink'
                                    defaultMessage='(Edit)'
                                />
                                      {(message) => <span className='button__edit ml-1'>{message}</span>}
                                </FormattedMessage>
```

asaadmahmood: Cool, thanks.

esethna:
@asaadmahmood please see this PR to make sure we aren't introducing conflicts? https://github.com/mattermost/mattermost-webapp/pull/4931

asaadmahmood:
@esethna There may be a conflict, but a minor one, and it could be easily resolved. My changes are quite small.

lindy65:
Thanks @asaadmahmood - I've tested on the spinwick and see the `(Edit)` option appear when I hover over `Add a channel description' when one has never been added before.

However, once a description has been added, there is no hover effect on the channel description and you cannot re-edit it from here. Is it intended then to have to edit the existing channel description from the channel drop-down menu for subsequent channel header edits?

asaadmahmood:
@lindy65 Yes. This is not changing any functionality. It's just adding the edit link. The Channel header if it is not set is already editable if it is not set. Yet if a channel header is set, it is not possible to edit it, thus not changing anything from that.

asaadmahmood:
Ideally we want the edit link in the subsequent headers as well, but that is a larger ticket.

PR #5030, opened 2020-03-10
MM-22818 - Removing status icon from channel

asaadmahmood:
#### Summary
MM-22818 - Removing status icon from channel

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22818

#### Screenshots
localhost_8065_

### DIFF
--- a/components/post_view/post/index.js
+++ b/components/post_view/post/index.js
@@ -65,6 +65,7 @@ function makeMapStateToProps() {
         return {
             post,
             createAriaLabel: createAriaLabelForPost(state, post),
+            currentChannel: channel,
             currentUserId: getCurrentUserId(state),
             isFirstReply: isFirstReply(post, previousPost),
             consecutivePostByUser,
--- a/components/post_view/post/index.js
+++ b/components/post_view/post/index.js
@@ -15,7 +15,7 @@ import {markPostAsUnread} from 'actions/post_actions';
 import {selectPost, selectPostCard} from 'actions/views/rhs';
 
 import {isArchivedChannel} from 'utils/channel_utils';
-import {Preferences} from 'utils/constants';
+import {Constants, Preferences} from 'utils/constants';
 import {makeCreateAriaLabelForPost, makeGetReplyCount} from 'utils/post_utils.jsx';
 
 import Post from './post.jsx';
@@ -67,6 +67,7 @@ function makeMapStateToProps() {
             createAriaLabel: createAriaLabelForPost(state, post),
             currentChannel: channel,
             currentUserId: getCurrentUserId(state),
+            isDMChannel: Boolean(channel && channel.type === Constants.DM_CHANNEL),
             isFirstReply: isFirstReply(post, previousPost),
             consecutivePostByUser,
             previousPostIsComment,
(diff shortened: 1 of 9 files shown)


lindy65:
Hi @asaadmahmood @adamjclarkson 

I've tested on the spinwick and the status icon is removed in the center channel.

However, the status icon still remains on reply posts in the right-hand-side which looks weird to me.

I also noticed your comment on the ticket Adam:

> It also pushes the firstname initial letter off to the left when user has a default avatar.

Even although the status icon is removed, this doesn't fix the first-name initial letter being off to the left when a user has a default avatar. In my opinion, the default avatar looks very strange now.

image

asaadmahmood:
@lindy65 I can update the ticket, but the letter in the avatar is not positioned to the left because of the status icon, that is a separate thing.

I have removed it from the RHS thread view now as well.

asaadmahmood:
I've assigned @esethna to the PR. @esethna the letters on the left are a bit separate from the emojis, and maybe coming from elsewhere. We can tackle that in another PR.

lindy65:
Thanks @asaadmahmood - I've re-tested and see the status is removed from the user avatar in the right-hand-side now too.

I realise that the letter on a default user avatar is over to the left whether or not the status icon is present. The reason I drew attention to this in my comment above is because Adam had mentioned on the Jira ticket that it was because of the status icon on the avatar that the letter was over to the left - which isn't the case.

esethna:
Discussed with UX team, propose we only do this only for DMs as a test (already have the online status in the channel header), and allow it to soak for feedback on daily for a release cycle

components/profile_picture.tsx
```diff
@@ -51,6 +52,11 @@ export default class ProfilePicture extends React.PureComponent<Props> {
             this.props.src;
 
         const profileIconClass = `profile-icon ${this.props.isEmoji ? 'emoji' : ''}`;
+        let statusIcon = <StatusIcon status={this.props.status}/>;
+
+        if (this.props.statusHidden) {
+            statusIcon = <></>;
+        }
```
saturninoabril: Instead of `<></>`, we normally do the other way, something like:
```suggestion
        let statusIcon;
        if (!this.props.statusHidden) {
            statusIcon = <StatusIcon status={this.props.status}/>;
        }
```

asaadmahmood:
@sudheerDev @saturninoabril Eric requested we remove the status icon from the DMs only, so please rereview the PR.

components/post_view/post/index.js
```diff
@@ -81,6 +81,7 @@ function makeMapStateToProps() {
         return {
             post,
             createAriaLabel: createAriaLabelForPost(state, post),
+            currentChannel: channel,
```
saturninoabril: Instead of passing the entire channel object which may cause unnecessary re-render, you may pass what's needed only, could be:
```suggestion
            isDMChannel: Boolean(channel && channel.type === Constants.DM_CHANNEL),
```

esethna:
@asaadmahmood, thinking about this more, doing it only in DMs creates an inconsistency that people may find frustrating. I think something like this should probably require a setting as SA suggests - that would apply to all channels. Propose we close this ticket for now

asaadmahmood:
@esethna Cool.

PR #5031, opened 2020-03-10
MM-22989 - Adding reply count to root posts in RHS (Search, Flags, Pinned)

asaadmahmood:
#### Summary
MM-22989 - Adding reply count to root posts in RHS
Comment count should be added in the search sidebar for root posts.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22989

#### Screenshots
Screenshot 2020-03-10 at 5 15 54 PM

### DIFF
--- a/components/search_results_item/index.js
+++ b/components/search_results_item/index.js
@@ -3,13 +3,15 @@
 
 import {connect} from 'react-redux';
 import {bindActionCreators} from 'redux';
+import {createSelector} from 'reselect';
+
 import {getChannel} from 'mattermost-redux/selectors/entities/channels';
 import {getConfig} from 'mattermost-redux/selectors/entities/general';
 import {getUser} from 'mattermost-redux/selectors/entities/users';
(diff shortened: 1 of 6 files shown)


components/search_results_item/index.js
```diff
@@ -23,7 +25,23 @@ import {getDirectTeammate, getDisplayNameByUser} from 'utils/utils.jsx';
...
+            if (!postIds) {
+                return 0;
+            }
+
+            // Count the number of non-ephemeral posts in the thread
+            return postIds.map((id) => allPosts[id]).filter((post) => post && !isPostEphemeral(post)).length;
+        }
+    );
+}
+
```
saturninoabril: This is the same as in [Post component](https://github.com/mattermost/mattermost-webapp/blob/6c0a915972b8d671db5676b6ec04554911c334a5/components/post_view/post/index.js#L40-L53).  How about move the function and corresponding tests to `utils/post_utils.jsx` and import from there?

asaadmahmood:
@saturninoabril Done, have a look.

PR #5034, opened 2020-03-10
Update dependencies.

grundleborg:
#### Summary
Update dependencies.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21581


package.json
```diff
@@ -105,14 +105,14 @@
...
-    "@types/react": "16.9.19",
+    "@types/react": "^16.9.23",
     "@types/react-bootstrap": "0.32.20",
     "@types/react-color": "3.0.1",
     "@types/react-dom": "16.9.5",
     "@types/react-redux": "7.1.4",
     "@types/react-router-dom": "5.1.2",
     "@types/react-select": "3.0.10",
-    "@types/react-transition-group": "4.2.3",
+    "@types/react-transition-group": "^4.2.4",
```
reflog: I think the consensus is - use exact versions and not "^". Right?

grundleborg: Yes good point - I'm separately going to propose changing that consensus, but we shouldn't do that in this PR.

PR #5041, opened 2020-03-11
MM-23042: Fix select teams when there are no teams

agnivade:
#### Summary

When there are no teams, then handleScroll does not
get called which does not allow it to set isEndOfData = true.

So we also check if isFetching is set to true or not,
which won't show the loader screen until it first gets set
to true when handleScroll gets called.

We also remove the unncessary CSS settings which made the
component to be wider.

This regression was introduced in 07b847a4f3b3710b86842a3ae8c97d0ce658ffba.

Thanks to @jespino and @ethervoid for helping me work on this. :)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23042

/cc @AGMETEOR


agnivade:
@ogi-m - I think that's a bit outside the scope of the PR. The PR was just about fixing the infinite spinner. Granted I made 2 small css changes, but anything more needs a separate ticket.

agnivade:
@hmhealey - It would be good to get this in 5.22 if possible. Any chance you could take a look ? Thanks.

PR #5042, opened 2020-03-11
[MM-23095] - Fix console error when marking system post as unread

ghost:
#### Summary
Fixes a JS error related to marking a system post as unread. There are two types of posts and system posts have an array of posts ID's, which needs to be taken into account in the mattermost-redux action. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23095

#### Related Pull Requests
- Has redux changes (https://github.com/mattermost/mattermost-redux/pull/1076)


Willyfrog:
if system posts are not supposed to be marked as unread (as it says on the ticket) why are we sending it to redux to manage? why not prevent it from the interface directly?

Willyfrog (approved):

nevyangelova:
Closing because mattermost/mattermost-redux#1076 has updated and won't need the change

hmhealey (changes requested):
Looks good, but this has to be updated with the redux changes

hmhealey (approved):

jgilliam17:
Thank you @nevyangelova 
I am seeing one error (see image) for `only visible to you` system post 
Screen Shot 2020-03-19 at 11 44 52 AM

hmhealey:
That error would've likely happened regardless. We probably don't check for ephemeral posts before we try to mark the post as unread. We could split that off into a separate ticket if we wanted.

jgilliam17 (approved):
Tested, looks good to merge.
- No error on alt+click on system message.
- Per previous comment, ephemeral system message error is not addressed in this PR.

amyblais:
@nevyangelova Can you help make sure to cherry pick this?

nevyangelova:
@amyblais already cherry-picked this from release-5.22

PR #5046, opened 2020-03-12
[MM-21517][MM-23451] Members in the Manage Members Team Modal roles not updating when changing pages and update call to fetch correct users

hahmadia:
#### Summary

This PR addresses two problems. 

First problem is MM-23451 which is when someone updates a users role on the manage team members modal (hamburger menu -> manage members) for users on the second page and further, other users who view that modal and go to see that users role will not see the updated role (this is because we are not refetching the users on those pages which this PR addressed to fix).

The second problem is that the manage team members modal omits deleted users (users who are inactive) and sorts by Username rather than UserID, however when we fetch users data from the backend, we fetch deleted users and sort users by their UserID. The backend has been updated to accept an options object to sepcify sorting by Username and excluding deleted users. This change needs to be reflected here and the redux.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21517
https://mattermost.atlassian.net/browse/MM-23451

#### Dependent PR's
**Server:** https://github.com/mattermost/mattermost-server/pull/14043
**Redux:** https://github.com/mattermost/mattermost-redux/pull/1091


components/member_list_team/member_list_team.tsx
```diff
@@ -116,8 +116,11 @@ export default class MemberListTeam extends React.Component<Props, State> {
         this.setState({loading: false});
     }
 
-    nextPage = (page: number) => {
-        this.props.actions.loadProfilesAndTeamMembers(page + 1, USERS_PER_PAGE);
+    nextPage = async (page: number) => {
+        this.setState({loading: true});
+        this.props.actions.loadProfilesAndTeamMembers(page, USERS_PER_PAGE),
```
fm2munsh: should this be `await Promise.all(...` ?

hahmadia: > should this be `await Promise.all(...` ?

@fm2munsh 
I was doing that initially but then I realized that it wasn't waiting on the `loadProfilesAndTeamMembers` prior to this change so I thought it might be better if I don't wait on it too cause it will just delay. I only need to wait on the getTeamMembers and the other request can run in the background. 

Thoughts? I also wasn't 100% but I mean to follow the pattern I use in the componentDidMount , I could do a `Promise.all`

fm2munsh: works for me although you still need to modify this line 
```suggestion
        this.props.actions.loadProfilesAndTeamMembers(page, USERS_PER_PAGE);
```

srkgupta:
As per discussion with @hahmadia this PR does not still fixes the original issue found while testing on the RC.test instance. It only partially fixes the issue where the users on the second page will never get updated. Have asked @hahmadia to fix this issue completely so that it can be tested thoroughly and ensured that the role for a user is updated every time in the Manage Members dialog.

srkgupta:
> As per discussion with @hahmadia this PR does not still fixes the original issue found while testing on the RC.test instance. It only partially fixes the issue where the users on the second page will never get updated. Have asked @hahmadia to fix this issue completely so that it can be tested thoroughly and ensured that the role for a user is updated every time in the Manage Members dialog.

hahmadia:
@sbishel @fm2munsh 
Hey guys,
This PR now addresses two problems (easier for Rohitesh to test as it's dealing with the same area of UI) so I've added some more code which requires a re-review. Thank you!

components/member_list_team/index.ts
```diff
@@ -26,7 +26,7 @@ type Props = {
 }
 
 type Actions = {
-    getTeamMembers: (teamId: string) => Promise<{data: {}}>;
+    getTeamMembers: (teamId: string, page?: number, perPage?: number, options?: {}) => Promise<{data: {}}>;
```
fm2munsh: based on my comment in redux  - you would then import the `GetTeamMembersOpts` type above here and change this to 
```suggestion
    getTeamMembers: (teamId: string, page?: number, perPage?: number, options?: GetTeamMembersOpts) => Promise<{data: {}}>;
```

components/member_list_team/member_list_team.tsx
```diff
@@ -62,7 +62,7 @@ export default class MemberListTeam extends React.Component<Props, State> {
     async componentDidMount() {
         await Promise.all([
             this.props.actions.loadProfilesAndTeamMembers(0, Constants.PROFILE_CHUNK_SIZE, this.props.currentTeamId),
-            this.props.actions.getTeamMembers(this.props.currentTeamId),
+            this.props.actions.getTeamMembers(this.props.currentTeamId, 0, Constants.DEFAULT_MAX_USERS_PER_TEAM, {sort: 'Username', exclude_deleted_users: true}),
```
sbishel: 'Username' make constant.

hahmadia: Done! Thanks

components/member_list_team/index.ts
hahmadia: Done! Thanks

PR #5052, opened 2020-03-12
Fix messaging specs

josephbaylon:
#### Summary
Fix failing specs on messaging, at mentions, elastic search, and interactive dialog. Added new plugin for clipboard.

#### Ticket Link
NA

### DIFF
--- a/e2e/cypress/integration/messaging/focus_move_spec.js
+++ b/e2e/cypress/integration/messaging/focus_move_spec.js
@@ -149,12 +149,6 @@ describe('Messaging', () => {
         });
     });
     it('M17455 - Focus does not move for non-character keys', () => {
-        //# Make sure main input is visible and focused
-        cy.get('#post_textbox').should('be.visible').and('be.focused');
-
-        //#Click anywhere in the body to move the focus out of the main input box
-        cy.get('body').click();
-
         //# Make sure main input is not focused
         cy.get('#post_textbox').should('not.be.focused');
 
--- a/e2e/cypress/integration/post_header/post_header_spec.js
+++ b/e2e/cypress/integration/post_header/post_header_spec.js
@@ -8,6 +8,8 @@
 // - Use element ID when selecting an element. Create one if none.
 // ***************************************************************
 
+import * as TIMEOUTS from '../../fixtures/timeouts';
+
 describe('Post Header', () => {
     before(() => {
         // # Go to Main Channel View with "user-1"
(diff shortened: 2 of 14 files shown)


e2e/cypress/integration/post_header/post_header_spec.js
```diff
@@ -33,8 +35,11 @@ describe('Post Header', () => {
             // # Click timestamp of a post
             cy.clickPostTime(postId);
 
-            // * Check that it redirected to permalink URL
-            cy.url().should('include', `/ad-1/pl/${postId}`);
+            // * Check if url include the permalink
+            cy.url().should('include', `/ad-1/channels/town-square/${postId}`);
+
+            // * Check if url redirects back to parent path eventually
+            cy.wait(TIMEOUTS.SMALL).url().should('include', '/ad-1/channels/town-square').and('not.include', `/${postId}`);
```
saturninoabril: Succeeding steps failed.  Looks like no longer applicable?
Screen Shot 2020-03-13 at 10 00 30 AM

e2e/cypress/integration/messaging/focus_move_spec.js
```diff
@@ -149,12 +149,6 @@ describe('Messaging', () => {
         });
     });
     it('M17455 - Focus does not move for non-character keys', () => {
-        //# Make sure main input is visible and focused
-        cy.get('#post_textbox').should('be.visible').and('be.focused');
```
saturninoabril: Why remove?  I thought the focus should be on the center textbox on visit to a channel.

e2e/cypress/plugins/getClipboard.js
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+const clipboardy = require('clipboardy');
+
+module.exports = async () => {
+    return clipboardy.readSync();
+};
```
saturninoabril: 💯

e2e/cypress/integration/post_header/post_header_spec.js
josephbaylon: This failure is known and listed in the audit I posted in the chatroom. Will address in a separate PR

e2e/cypress/integration/messaging/focus_move_spec.js
josephbaylon: For some reason, loading the page manually on browser puts focus on center textbox correctly. However, in cypress, the focus never happens. I don't think it's an issue with the app since it works manually. The test case seems to just verify focus hasn't changed (or in this case, lack of focus). I can put it back and let it fail if you think otherwise.

saturninoabril: That makes sense.  I've checked the [test case](https://docs.google.com/spreadsheets/d/1-3usQeVbXS798rNt2IkcsKM-KxcckLnsehztSkeSiNw/edit#gid=1636776662) and we're still align on it.
> 1. Left click on a post to move the focus out of the main input box
> 2. Hit the following keys and watch to see if the focus is returned to the main input box: Num Lock, down arrow, Shift, F7, Pg Up(from numpad with num lock off), enter and a system key such as win or fn

PR #5055, opened 2020-03-13
[MM-22212] Add channel scheme websocket event

fm2munsh:
#### Summary
Adds a websocket event to dispatch `getMyChannelMember` when a channel_scheme is either created or deleted to update the list of roles tied to that channel.

Related Server PR: https://github.com/mattermost/mattermost-server/pull/14057

Test steps: 
1.  Login to a channel as regular member
2. Login as sysadmin in different tab and make the channel read only
3. Ensure that the channel changes to read only without reload for the regular member


actions/websocket_actions.jsx
```diff
@@ -309,6 +310,10 @@ export function handleEvent(msg) {
         handleRoleRemovedEvent(msg);
         break;
 
+    case SocketEvents.CHANNEL_SCHEME_UPDATED:
+        handleChannelSchemeUpdatedEvent(msg);
+        break
```
fm2munsh:
```suggestion
        break;
```

PR #5056, opened 2020-03-14
[MM-22482] Reset focus on CreatePost Textbox to address Firefox bug

bradjcoughlin:
#### Summary
In Firefox, attempts to focus a disabled input field or textbox results in the React app losing focus, therefore, ignoring further keyboard shortcuts keypress events to be captured. In the case of cycling through channels using `alt + up/down` (or `option + up/down`), this was occurring when a read-only channel was reached, effectively "freezing" the UI on said channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22482


components/create_post/create_post.jsx
```diff
@@ -668,8 +668,10 @@ class CreatePost extends React.PureComponent {
     }
 
     focusTextbox = (keepFocus = false) => {
-        if (this.refs.textbox && (keepFocus || !UserAgent.isMobile())) {
-            this.refs.textbox.getWrappedInstance().focus();
+        if (!this.readOnlyChannel) { // On Firefox, setting focus on a disabled Textbox causes keyboard shortcuts to not be ignored
```
hmhealey: Should this check `this.readOnlyChannel()` instead?

components/create_post/create_post.jsx
```diff
@@ -1123,16 +1125,21 @@ class CreatePost extends React.PureComponent {
         this.setState({renderScrollbar: height > maxHeight});
     }
 
+    readOnlyChannel = () => {
```
hmhealey: Total nitpicking, but maybe call this `isReadOnlyChannel`? I personally prefer getters to be called `isX` when it returns a boolean

components/create_post/create_post.jsx
```diff
@@ -1195,7 +1202,7 @@ class CreatePost extends React.PureComponent {
         }
 
         let preview = null;
-        if (!readOnlyChannel && (draft.fileInfos.length > 0 || draft.uploadsInProgress.length > 0)) {
+        if (!this.readOnlyChannel && (draft.fileInfos.length > 0 || draft.uploadsInProgress.length > 0)) {
```
hmhealey: Similarly, should this also use `this.readOnlyChannel`?

components/create_post/create_post.jsx
```diff
@@ -1376,7 +1383,7 @@ class CreatePost extends React.PureComponent {
                                 characterLimit={this.props.maxPostSize}
                                 showPreview={this.state.showPreview}
                                 updatePreview={this.updatePreview}
-                                message={readOnlyChannel ? '' : this.state.message}
+                                message={this.readOnlyChannel ? '' : this.state.message}
```
hmhealey: And this

hmhealey:
Looks good on my end. That's interesting that attempting to focus a disabled textbox messes up focus on Firefox in particular. I definitely would not have expected that

components/create_post/create_post.jsx
```diff
@@ -1195,7 +1202,7 @@ class CreatePost extends React.PureComponent {
         }
 
         let preview = null;
-        if (!readOnlyChannel && (draft.fileInfos.length > 0 || draft.uploadsInProgress.length > 0)) {
+        if (!this.isReadOnlyChannel() && (draft.fileInfos.length > 0 || draft.uploadsInProgress.length > 0)) {
```
sudheerDev: Nit: we could assign the value at the top of the render function

sudheerDev:
@bradjcoughlin It is nice to have a unit test for this asserting that `this.refs.textbox.getWrappedInstance()` is not called in case of readonly. channels

components/create_post/create_post.jsx
bradjcoughlin: I was wondering about that and if that is our preferred practice.

bradjcoughlin:
> @bradjcoughlin It is nice to have a unit test for this asserting that `this.refs.textbox.getWrappedInstance()` is not called in case of readonly. channels

Agree. I can file a follow-up ticket to do this.

jgilliam17:
Thank you @bradjcoughlin 
I am still seeing this in Firefox for user-1. Alt+up/down arrow scrolling stops at Town Square.

hmhealey:
Oh yeah, I'm seeing it as well. I did test this before, but it was with the sysadmin account that can post either way

bradjcoughlin:
@jgilliam17 @hmhealey I may be back to the drawing board on this one 😖 . This fix was working at the time of the original commit, but now it's not working locally for me either.

components/create_post/create_post.jsx
```diff
@@ -668,6 +668,9 @@ class CreatePost extends React.PureComponent {
     }
 
     focusTextbox = (keepFocus = false) => {
+        if (this.refs.textbox) {
```
sudheerDev: i don't understand the reasoning behind this. Can we not focus in the first place for read only channels in all brwosers?

bradjcoughlin: We can't for Firefox, it simply doesn't behave correctly when we do. I'm not 100% clear on the reasons, because this isn't documented anywhere and we have competing `focus()`'s throughout the webapp. Over about a week's time of experimenting on/off I narrowed it down to this - and I can't explain exactly why Firefox is acting weird, but suffice it to say Chrome/Safari don't have this problem. I'm wondering if this issue just goes away in future Firefox releases.

bradjcoughlin: I'm essentially unsetting the `focus()` here before re-focusing because I believe FF gets confused when the component changes to read-only while focus is still applied. Like I said, it's been really difficult to track down so this is a bit of black magic.

bradjcoughlin: I also renamed the ticket to match this new fix.

Willyfrog:
code looks good, i'm only concerned about the video I sent you, as I'm not sure if it's the expected behaviour

components/create_post/create_post.jsx
sudheerDev: Can you add a comment to this line explaining which bug it solves.

sudheerDev:
> > @bradjcoughlin It is nice to have a unit test for this asserting that `this.refs.textbox.getWrappedInstance()` is not called in case of readonly. channels
> 
> Agree. I can file a follow-up ticket to do this.

This unit test was supposed to be added along with this PR. Based on previous dev decisions

Every bug fix should have a unit test was one of the. 
Features should have e2e and unit tests

Key change at least should have unit tests if not entire change.

bradjcoughlin:
@sudheerDev Fair, and @deanwhillier and I just looked at this issue together and now that we're 100% sure how to reproduce, I'll definitely add the unit test. For a browser-specific bug, that test will make it clear what we're trying to fix with this code.

hmhealey:
@bradjcoughlin Is there anything that still has to change here? These changes were approved, but it sounds like you still have changes to be made based on this
> @sudheerDev Fair, and @deanwhillier and I just looked at this issue together and now that we're 100% sure how to reproduce, I'll definitely add the unit test. For a browser-specific bug, that test will make it clear what we're trying to fix with this code.

sudheerDev:
@bradjcoughlin I feel it is fine the way it is. It is not worth the mana cost wrt to the impact of the bug IMO

bradjcoughlin:
@sudheerDev sounds good.
@hmhealey should be good to go now.
- Added the condition that `blur()` only fire if `textbox` is read-only.
- Added the bug issue reference number to the comment

components/create_post/create_post.jsx
```diff
@@ -668,6 +668,11 @@ class CreatePost extends React.PureComponent {
     }
 
     focusTextbox = (keepFocus = false) => {
+        const postTextboxDisabled = this.props.readOnlyChannel || this.props.canPost;
```
sudheerDev: should this be `!this.props.canPost`?

PR #5061, opened 2020-03-16
MM-23197: Add Guest permissions to Team Override Settings.

sbishel:
#### Summary
Adds guest permissions to be set in Team Override Schemes.  Currently, the guest permissions cannot be overridden at the team level.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23197
https://mattermost.atlassian.net/browse/MM-23650

#### Screenshots
Screen Shot 2020-03-16 at 1 23 57 PM


components/admin_console/permission_schemes_settings/permission_team_scheme_settings/permission_team_scheme_settings.jsx
```diff
@@ -36,11 +37,21 @@ const EXCLUDED_PERMISSIONS = [
     Permissions.LIST_PRIVATE_TEAMS,
 ];
 
+const GUEST_INCLUDED_PERMISSIONS = [
+    Permissions.CREATE_PRIVATE_CHANNEL,
+    Permissions.EDIT_POST,
+    Permissions.DELETE_POST,
+    Permissions.ADD_REACTION,
+    Permissions.REMOVE_REACTION,
+];
```
mkraft: Adding a newly exposed guest permission to `GuestPermissionsTree` adds the permission to the UI for both of the`PermissionTeamSchemeSettings` and `PermissionSystemSchemeSettings` components, but both are *also* required to update their `GUEST_INCLUDED_PERMISSIONS` arrays otherwise the new checkbox does nothing.

I think both of the `Permission*SchemeSettings` components should import this array from `GuestPermissionsTree` and then `GuestPermissionsTree` component is the source of truth and their behaviour is more likely to be kept in lockstep.

fm2munsh: `EXCLUDED_PERMISSIONS` is also duplicated across two files `permission_system_scheme...` and `permission_team_scheme..` 

I think both `EXCLUDED_PERMISSIONS` and `GUEST_INCLUDED_PERMISSIONS` can be added to the `permissions_tree` file as exported constants

sbishel: Agreed

PR #5062, opened 2020-03-16
[MM-22670] - Hide tooltips in header when the dropdowns are open

nevyangelova:
#### Summary
Hides the tooltips that are displayed when the user clicks on a channel popover item and the dropdown is displayed

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-22670

Fix versions
v5.22 (April 2020)

### DIFF
--- a/components/channel_header/components/header_icon_wrapper.js
+++ b/components/channel_header/components/header_icon_wrapper.js
@@ -129,4 +129,5 @@ HeaderIconWrapper.propTypes = {
     onClick: PropTypes.func.isRequired,
     tooltipKey: PropTypes.string,
     tooltipText: PropTypes.node,
+    isRhsOpen: PropTypes.bool,
 };
--- a/components/channel_header/components/header_icon_wrapper.js
+++ b/components/channel_header/components/header_icon_wrapper.js
@@ -20,6 +20,7 @@ export default function HeaderIconWrapper({
     onClick,
     tooltipKey,
     tooltipText,
+    isRhsOpen,
 }) {
     const toolTips = {
         flaggedPosts: {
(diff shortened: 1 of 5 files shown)


components/channel_header/components/header_icon_wrapper.js
```diff
@@ -92,7 +93,7 @@ export default function HeaderIconWrapper({
                     trigger={['hover']}
                     delayShow={Constants.OVERLAY_TIME_DELAY}
                     placement='bottom'
-                    overlay={tooltip}
+                    overlay={isRhsOpen ? <></> : tooltip}
```
devinbinnie: How come this can't just return `null` when the RHS is open?

nevyangelova: It threw an error that it's not the right type of element for the overlay. 🤷‍♀

devinbinnie: That's odd, the type of `overlay` in `overlay_trigger.tsx` is `any`.
What's the error that you get?

nevyangelova:
```react-hot-loader.development.js?c2cb:2408 Uncaught TypeError: Cannot read property 'props' of null
    at eval (overlay_trigger.tsx?1b61:29)
    at updateContextConsumer (react-dom.development.js?7f13:19771)
    at beginWork$1 (react-dom.development.js?7f13:20148)
    at HTMLUnknownElement.callCallback (react-dom.development.js?7f13:363)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?7f13:412)
``` 

And I get it for all the components I updated

devinbinnie: Ohhh, it's because it's trying to read the props from the component you pass in...gotcha.
I wonder if there's like a null component you can use that provides those...otherwise I'm okay with this.

PR #5065, opened 2020-03-17
[MM 23264] Show Confirmation Modal for group mentions 

hahmadia:
#### Summary

Show the confirmation modal for group mentions 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23264

#### Related Pull Requests

https://github.com/mattermost/mattermost-server/pull/14068
https://github.com/mattermost/mattermost-redux/pull/1086

PR #5068, opened 2020-03-17
[MM-20775] Group mentions feature

fmunshi:
#### Summary
https://github.com/mattermost/mattermost-webapp/pull/4977 - Merged, In Dev Review
https://github.com/mattermost/mattermost-webapp/pull/5066 - Merged, Dev Review Complete ✅ 
https://github.com/mattermost/mattermost-webapp/pull/5065 - Merged, Dev Review Complete ✅ 

#### Ticket Link
Epic link https://mattermost.atlassian.net/browse/MM-20775

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1087
https://github.com/mattermost/mattermost-server/pull/14067


components/admin_console/permission_schemes_settings/permission_schemes_settings.jsx
```diff
@@ -328,6 +328,8 @@ t('admin.permissions.permission.revoke_user_access_token.description');
 t('admin.permissions.permission.revoke_user_access_token.name');
 t('admin.permissions.permission.upload_file.description');
 t('admin.permissions.permission.upload_file.name');
+t('admin.permissions.permission.use_group_mentions.description');
+t('admin.permissions.permission.use_group_mentions.name')
```
fm2munsh:
```suggestion
t('admin.permissions.permission.use_group_mentions.name');
```

srkgupta:
Hi @fm2munsh and @hahmadia 

I am seeing the following issues. Can you please check and fix them:

1) The Group Mention becomes undefined when we add a new Team or Channel in the Groups configuration page. 
Screenshot 2020-03-18 at 11 46 09 AM

2) The Group Mentions handle should be shown as lowercase. However it's displayed as is in the webapp. 
Screenshot 2020-03-18 at 11 57 46 AM

3) The Group Mentions feature does not works on a different team which have not been added to the LDAP Groups. 
a) Login as a member and create a new team. 
b) Ensure developers Group mentions is enabled in System console. 
c) For the newly created team, type `@developers` and check. Nothing happens. 

The server displays the group mentions only in the teams which have been added in the Groups Configuration page. However this behaviour is incorrect and a LDAP user can never be invited with this assumption. The reason being if the teams are added to a LDAP group, then automatically all the LDAP users are added to the team. Hence there will never be a case where LDAP users can be invited. 

Expected: Group Mentions should be allowed to use in other teams as well and if LDAP users are not present in that team, then an option to invite LDAP users should be provided. 

I am still doing my testing and will add more issues as separate comments.

srkgupta:
Some more issues:
4. If we use the group mentions on a channel, the LDAP users belonging to that group do not get notified. 
a) On browser 1, visit a channel on which Group mentions is available. 
b) On browser 2, login as a LDAP user belonging to the developers group. 
c) Ensure Groups mention is enabled for developers group. 
d) On browser 1, use the group mention `@developers` and check if user in step 2 gets notified. 
e) User on browser 2 does not gets notified. 
Expected: User on browser2 should get notified and a desktop/browser notification should be sent. The channel should be marked as unread with a notification count. 

5. Guest users should not be able to use the group mentions feature
a) Login as a guest user and access the channel on which group mentions is enabled. 
b) Try to use the group mentions. A guest user is allowed to use the group mentions.
c) As per expectations, the guest user should not be allowed to use the group mentions.

catalintomai:
> Please see the list of issues mentioned earlier.

Hi 

> Please see the list of issues mentioned earlier.

Hi @srkgupta, 
I looked at the bugs above and here are my findings:
#1 - existing bug, exposed now since we are displaying the group name in the UI - on a an existing dev deployment you can look at the JS console when linking a team from "Group Configuration" panel :"failed prop type: The prop `name` is marked as required in `GroupProfile`, but its value is `undefined". This should be fixed in the feature branch now.
#2 - should be fixed in the feature branch now
#3, #4, #5 - could not repro these ones - I started with a clean env (make nuke, make test-data, make config-reset, make config-ldap). We should have a Zoom session to see what I am doing differently. The guest issue has been fixed late, so it could be that you were using an earlier drop.
#6 - Since we are not doing invite groups yet - are you referring to this message: "..did not get notified by this mention because they are not in the channel. Would you like **to add them to the channel**?.."? 
Fwik we did not change the logic there but @fm2munsh can confirm. Or are you referring to not being able to group mention at all, unless the user has the permission to manage groups? If yes, please add a comment to the Design/tech spec to consider as part of the design. or open a ticket).
#7 -  When you say "Group Mentions invite" are you referring to the notifications we are sending to the members of the groups we are mentioning? If yes, please add a comment to the Design/Tech spec (or open a ticket if you think we should have a trace of the issue) 
#8 - I tried the same action with a user name : "``` @username ```, and the user mention was not filtered out. Should we expect a different behavior for groups vs. users?  
Thanks.

hahmadia:
@srkgupta 
I have addressed problem #8. You should not longer be getting a notification modal when mentioning groups inside code blocks. Thanks for raising this.

srkgupta:
This PR is no longer valid and does not needs QA and needs to be closed without any merge. Individual PRs are being tested and will be merged accordingly.

PR #5069, opened 2020-03-17
Improve E2E test duration of interactive menu by using waitUntil instead of explicit wait when posting an incoming webhook

saturninoabril:
#### Summary
Improve E2E test duration of interactive menu by using waitUntil instead of explicit wait when posting an incoming webhook

#### Ticket Link
none

### DIFF
--- a/e2e/cypress/integration/interactive_menu/basic_options_spec.js
+++ b/e2e/cypress/integration/interactive_menu/basic_options_spec.js
@@ -30,6 +30,9 @@ describe('Interactive Menu', () => {
     before(() => {
         cy.requireWebhookServer();
 
+        // # Login as sysadmin
+        cy.apiLogin('sysadmin');
+
         // Set required ServiceSettings
         const newSettings = {
             ServiceSettings: {
@@ -38,10 +41,9 @@ describe('Interactive Menu', () => {
                 EnablePostIconOverride: true,
             },
         };
-        cy.apiUpdateConfig(newSettings);
+        cy.apiUpdateConfigBasic(newSettings);
 
-        // # Login as sysadmin and ensure that teammate name display setting is set to default 'username'
-        cy.apiLogin('sysadmin');
+        // # Update teammate name display setting is set to default 'username'
         cy.apiSaveTeammateNameDisplayPreference('username');
         cy.apiSaveMessageDisplayPreference('clean');
 
@@ -59,9 +61,6 @@ describe('Interactive Menu', () => {
             cy.apiCreateWebhook(newIncomingHook).then((hook) => {
                 incomingWebhook = hook;
             });
-
-            // # Login again, this is a temp fix as cy.createNewUser, logs out the current user
-            cy.apiLogin('sysadmin').visit(`/ad-1/channels/${channel.name}`);
         });
     });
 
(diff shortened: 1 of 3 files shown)


e2e/cypress/integration/interactive_menu/basic_options_spec.js
```diff
@@ -546,7 +540,16 @@ describe('Interactive Menu', () => {
         });
     });
 
-    it('IM21038 - Selected options with long usernames are not cut off in the RHS', () => {
+    it.only('IM21038 - Selected options with long usernames are not cut off in the RHS', () => {
```
josephbaylon: Did you mean to keep .only here?

e2e/cypress/support/task_commands.js
```diff
@@ -28,7 +28,12 @@ Cypress.Commands.add('postMessageAs', ({sender, message, channelId, rootId, crea
 * @param {Object} data - payload on incoming webhook
 */
 Cypress.Commands.add('postIncomingWebhook', ({url, data}) => {
-    cy.task('postIncomingWebhook', {url, data}).its('status').should('be.equal', 200).wait(TIMEOUTS.MEDIUM);
+    cy.task('postIncomingWebhook', {url, data}).its('status').should('be.equal', 200);
+
+    cy.waitUntil(() => cy.getLastPost().then((el) => {
```
josephbaylon: What if you have 2 or more tests that already posted and the intended incoming message hasn't come in yet? Wouldn't this get whatever the latest even if the current test hasn't posted yet?

saturninoabril: No, the retry starts with getting the last post then evaluating whether it has the expected value ``() => cy.getLastPost().then()``. So if the incoming message hasn't posted yet, it will retry to get the last post until it's true or timed out.

e2e/cypress/integration/interactive_menu/basic_options_spec.js
saturninoabril: Missed it, now reverted, thanks

e2e/cypress/support/task_commands.js
josephbaylon: If response of postIncomingWebhook is synchronous then my previous question is invalid, but if it is async, then I think it's valid.

josephbaylon: Ah I see .includes(data.attachments[0].pretext)). Cool

saturninoabril:
@srkgupta Thanks for checking. Fixed by revisiting a channel after creating a new user, PTAL.

The issue in general is that we have API commands that does not do one thing only.  Like some convenient commands do login/logout a user (specifically a sysadmin) implicitly, and as a result, it breaks/clears the session in the middle when we're using it in our spec files.  My proposal is for API command to do one thing only and to follow what's needed (minimum) by the corresponding REST API, and then be more explicit in the spec file. I'll submit a separate PR to address it.

srkgupta:
I agree. 
>The issue in general is that we have API commands that does not do one thing only. Like some convenient commands do login/logout a user (specifically a sysadmin) implicitly, and as a result, it breaks/clears the session in the middle when we're using it in our spec files. My proposal is for API command to do one thing only and to follow what's needed (minimum) by the corresponding REST API, and then be more explicit in the spec file. I'll submit a separate PR to address it.

PR #5070, opened 2020-03-17
[MM-20903] Category collapsing and unread filter tests

devinbinnie:
#### Summary
Adds E2E tests that test the functionality of collapsing categories, and the unread filter of the new experimental sidebar.

NOTE: Make sure you are running the latest `master` server code, which contains the code necessary to enable the experimental sidebar.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20903


e2e/cypress/integration/channel_sidebar/category_collapsing_unread_filter_spec.js
```diff
@@ -0,0 +1,287 @@
...
+
+        // # Click on PUBLIC CHANNELS
+        cy.get('.SidebarChannelGroupHeader:contains(PUBLIC CHANNELS)').should('be.visible').click();
+
+        // * Verify that Off-Topic is no longer visible but Channel Test still is
+        cy.get('.SidebarChannel:contains(Off-Topic)').should('not.be.visible');
+        cy.get('.SidebarChannel:contains(Channel Test)').should('be.visible');
+    });
+
+    // Test should be working but isn't because the state doesn't seem to be persisting in Cypress
```
devinbinnie: This test is failing and I'm trying to determine why.

e2e/cypress/integration/channel_sidebar/category_collapsing_unread_filter_spec.js
```diff
@@ -0,0 +1,291 @@
...
+    before(() => {
+        cy.apiLogin('user-1');
+
+        cy.visit('/');
+    });
+
+    it('should display collapsed state when collapsed', () => {
+        // # Start with a new team
+        const teamName = `team-${getRandomInt(999999)}`;
+        cy.createNewTeam(teamName, teamName);
```
saturninoabril: Should each test/it block needs to be tested against a new team? If possible, create a new team once for all the tests in this spec file. Same for other spec files.

e2e/cypress/integration/channel_sidebar/category_collapsing_unread_filter_spec.js
```diff
@@ -0,0 +1,291 @@
...
+
+        // # Disable the unread filter
+        cy.get('.SidebarFilters_filterButton').should('be.visible').click();
+
+        // * Verify that the unread filter is back in all channels state
+        cy.get('.SidebarFilters:contains(VIEWING:)').should('be.visible');
+        cy.get('.SidebarFilters:contains(All channels)').should('be.visible');
+    });
+
+    it('should only show unreads when the unread filter is enabled', () => {
```
saturninoabril: This test failed on me. PTAL
Screen Shot 2020-03-18 at 8 38 24 PM

e2e/cypress/integration/channel_sidebar/channel_sidebar_spec.js
```diff
@@ -0,0 +1,130 @@
...
+            const channel = response.body;
+            cy.postMessageAs({sender: sysadmin, message: 'Test', channelId: channel.id});
+        });
+
+        // * Verify that Off Topic is unread and Town Square is read
+        cy.get('.SidebarChannel.unread:contains(Off-Topic)').should('be.visible');
+        cy.get('.SidebarChannel:not(.unread):contains(Town Square)').should('be.visible');
+    });
+
+    it('should remove channel from sidebar after leaving it', () => {
```
saturninoabril: This and the succeeding test failed on me. PTAL.
Screen Shot 2020-03-18 at 8 46 41 PM

e2e/cypress/integration/channel_sidebar/hotkeys_spec.js
```diff
@@ -0,0 +1,113 @@
...
+
+        // # Press alt + down
+        cy.get('body').type('{alt}', {release: false}).type('{downarrow}').type('{alt}', {release: true});
+
+        // * Verify that the channel changed to the Town Square
+        cy.url().should('include', `/${teamName}/channels/town-square`);
+        cy.get('#channelHeaderTitle').should('contain', 'Town Square');
+    });
+
+    it('should switch to unread channels when pressing the alt + shift + arrow hotkeys', () => {
```
saturninoabril: This test failed on me. PTAL.
Screen Shot 2020-03-18 at 8 48 53 PM

e2e/cypress/integration/channel_sidebar/channel_sidebar_spec.js
devinbinnie: These are in Harrison's PR, he will fix them there.

e2e/cypress/integration/channel_sidebar/hotkeys_spec.js
devinbinnie: These are in Harrison's PR, he will fix them there.

e2e/cypress/integration/channel_sidebar/category_collapsing_unread_filter_spec.js
hmhealey: This is copied from my code. We wanted to start with a new team each time to avoid having one test depend on the state left over by the previous test. Maybe we should have some best practices listed around this?

e2e/cypress/integration/channel_sidebar/category_collapsing_unread_filter_spec.js
```diff
@@ -0,0 +1,294 @@
...
+        cy.url().should('include', `/${teamName}/channels/town-square`);
+
+        // Create a new channel and post a message into it
+        cy.getCurrentTeamId().then((teamId) => {
+            cy.apiCreateChannel(teamId, 'channel-test', 'Channel Test').then((res) => {
+                cy.postMessageAs({sender: sysadmin, message: 'Test', channelId: res.body.id});
+
+                // * Verify that all channels are visible
+                cy.get('.SidebarChannel:contains(Off-Topic)').should('be.visible');
+                cy.get('.SidebarChannel:contains(Channel Test)').should('be.visible').should('has.class', 'unread');
```
saturninoabril: Test failed here. 
Screen Shot 2020-03-20 at 7 31 26 AM

"Channel Test" is not turning into unread since sysadmin is not channel member and is not able to successfully post a message.  Before `cy.postMessageAs` above, I suggests to explicitly add sysadmin:
```javascript
                const channel = res.body;
                cy.apiGetUserByEmail(sysadmin.email).then((emailResponse) => {
                    const user = emailResponse.body;
                    cy.apiAddUserToTeam(channel.team_id, user.id);
                    cy.apiAddUserToChannel(channel.id, user.id);
                });

                cy.postMessageAs({sender: sysadmin, message: 'Test', channelId: channel.id});
```

e2e/cypress/integration/channel_sidebar/category_collapsing_unread_filter_spec.js
```diff
@@ -0,0 +1,294 @@
...
+        cy.url().should('include', `/${teamName}/channels/town-square`);
+
+        // # Create a new channel and post a message into it
+        cy.getCurrentTeamId().then((teamId) => {
+            cy.apiCreateChannel(teamId, 'channel-test', 'Channel Test').then((res) => {
+                cy.postMessageAs({sender: sysadmin, message: 'Test', channelId: res.body.id});
+
+                // * Verify that all channels are visible
+                cy.get('.SidebarChannel:contains(Off-Topic)').should('be.visible');
+                cy.get('.SidebarChannel:contains(Channel Test)').should('be.visible').should('has.class', 'unread');
```
saturninoabril: Same here.
Screen Shot 2020-03-20 at 7 31 32 AM

e2e/cypress/integration/channel_sidebar/category_collapsing_unread_filter_spec.js
saturninoabril: I totally agree on it. One thing that I'm worried is with the number of new teams for "user-1" which is a permanent/default test user. One downside is it might introduce flakiness to other tests like needing to do "scrollIntoView" since the expected team is outside the window.
Instead of user-1, one option is to create a new (throw-away) test user where it can be added into many teams as required; like fire and forget after the test.
But I'm fine as is. We'll look into this more during daily Cypress test.

e2e/cypress/integration/channel_sidebar/category_collapsing_unread_filter_spec.js
devinbinnie: Thanks for the tip, but this is actually causing the tests to fail consistently for me now. Do I need to wait for each API call maybe?

saturninoabril: Can you share the fail logs you're seeing?  May not need explicit wait since `cy.get` is doing it under the hood already.  How about try moving `cy.postMessageAs` and the rest of `cy.get` inside the last `.then` ==> 
```
cy.apiGetUserByEmail(sysadmin.email).then((emailResponse) => { // all here //}); 
```

devinbinnie: Still not working, it looks like the channel isn't being marked as unread using the above logic.
`CypressError: Timed out retrying: expected '<li.SidebarChannel>' to have class 'unread'`

hmhealey: The systems admin shouldn't have to join the channel to post in it. System admins can do basically anything. This is probably failing because of something else.

saturninoabril: Yeah, one issue I'm seeing is on how Cypress handles websocket . Maybe instead of waiting for message via websocket, reload the page after posting a message.  Otherwise, I'm good and will take it from here as part of Cypress test maintenance.

devinbinnie: Okay, added the `reload()` after `postMessageAs()`. Let me know if that works for you.

saturninoabril: Thanks, works for me.

PR #5072, opened 2020-03-18
MM-23094: allow full screen modals to self focus

lieut-data:
#### Summary
https://mattermost.atlassian.net/browse/MM-17534 introduced a focus trap for full-screen modals, preventing the marketplace from focussing the search box within itself on component mount. I have a proposed fix here, but I don't understand the full scope of the original focus trap, so I'd appreciate context to know if this is the right solution.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23094


components/widgets/modals/full_screen_modal.tsx
```diff
@@ -29,7 +29,7 @@ class FullScreenModal extends React.Component<Props> {
     public componentDidMount() {
         document.addEventListener('keydown', this.handleKeypress);
         document.addEventListener('focus', this.enforceFocus, true);
-        this.resetFocus();
+        this.enforceFocus();
```
lieut-data: `resetFocus` was unconditional, but `enforceFocus` allows the focus to stay within the modal, still resetting it otherwise. Thoughts?

e2e/cypress/integration/plugins/marketplace/marketplace_spec.js
```diff
@@ -197,7 +198,7 @@ describe('Plugin Marketplace', () => {
             cy.get('#error_bar').should('not.be.visible');
 
             // * search should be visible
-            cy.findByPlaceholderText('Search Plugins').should('be.visible').click();
+            cy.findByPlaceholderText('Search Plugins').should('be.visible');
```
lieut-data: This `click` was preventing the focus test from being meaningful, since it was already focussed manually. There are other failing e2e tests in my local sandbox, but they appear to be untouched by this change.

marianunez: I think this was there before we introduced the change to be focused.

PR #5074, opened 2020-03-18
[MM-22212] Update manage_members from server after switching channel privacy

fm2munsh:
#### Summary
- Updates the value of `manage_members` from the server after the user switches the channel privacy 
- Also fixes a bug where `isPrivacyChanging` would remain true even after submitting the form

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/13813

#### Screenshots
captured

### DIFF
--- a/components/admin_console/team_channel_settings/channel/details/channel_details.tsx
+++ b/components/admin_console/team_channel_settings/channel/details/channel_details.tsx
@@ -125,7 +125,7 @@ export default class ChannelDetails extends React.Component<ChannelDetailsProps,
                 then(() => actions.getChannel(channelID)).
                 then(() => this.setState({groups: this.props.groups}))
             );
-            actionsToAwait.push(actions.getChannelModerations(channelID).then(() => this.updateChannelPermissions()));
+            actionsToAwait.push(actions.getChannelModerations(channelID).then(() => this.restrictChannelMentions()));
         }
 
         if (channel.team_id) {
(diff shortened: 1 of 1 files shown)


fm2munsh:
This relates to issue 3 from https://github.com/mattermost/mattermost-server/pull/13813

fm2munsh:
@hahmadia I readded `getChannelModerations` in the case where `isPrivacyChanging = true`.
Since there is a really small edge case i encountered:
1 - Enable `manage_members` 
2 - Toggle channel privacy
3 - Hit Save
4 - if the privacy you are now switching to has `manage_members` disabled in its inherited scheme the request would fail since you would be trying to add a permission that is disabled in the parent scheme 

My solution is to simply update the `manage_members` permission from the server before submitting the patch

components/admin_console/team_channel_settings/channel/details/channel_details.tsx
```diff
@@ -164,6 +141,30 @@ export default class ChannelDetails extends React.Component<ChannelDetailsProps,
 
         await Promise.all(actionsToAwait);
     }
+
+    private updateChannelPermissions() {
```
hahmadia: NIT: Maybe we can rename this to `updateUseChannelMentions` or something since currently it's only updating those and not the other permisisons. Thoughts? More specific but don't mind either way.

fm2munsh: I have renamed it to `restrictChannelMentions` since that is what its doing at its core - restricting channel mentions based on the create post value

PR #5075, opened 2020-03-18
MM-23415: Reload channel moderations upon change to channel type (pri…

mkraft:
…vate/public).

#### Summary

Update the channel moderations from the server if the channel type (public vs private) has been changed and saved.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23415

### DIFF
--- a/components/admin_console/team_channel_settings/channel/details/channel_details.tsx
+++ b/components/admin_console/team_channel_settings/channel/details/channel_details.tsx
@@ -124,7 +124,7 @@ export default class ChannelDetails extends React.Component<ChannelDetailsProps,
                 then(() => actions.getChannel(channelID)).
                 then(() => this.setState({groups: this.props.groups}))
             );
-            actionsToAwait.push(actions.getChannelModerations(channelID).then(this.getChannelModerationsCallback));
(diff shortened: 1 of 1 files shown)


hahmadia:
I think this is a duplicate of https://github.com/mattermost/mattermost-webapp/pull/5074

components/admin_console/team_channel_settings/channel/details/channel_details.tsx
```diff
@@ -375,6 +351,9 @@ export default class ChannelDetails extends React.Component<ChannelDetailsProps,
                 serverError = <FormError error={resultWithError.error.message}/>;
             } else {
                 await actions.getGroups(channelID);
+                if (isPrivacyChanging) {
+                    await actions.getChannelModerations(channelID).then(this.getChannelModerationsCallback);
+                }
```
fm2munsh: this will be called before the patch and therefore if the user has modified channel moderated permissions while also toggling from private to public the channel moderated permissions changes will tossed and replaced with what exists on the server

mkraft: That won't do at all! I'll fix it, unless Hossein is right and you're talking care of this in another PR?

PR #5077, opened 2020-03-18
[MM-22937] Fix Edge case by using includes instead of contains

fm2munsh:
#### Summary
- Fixes an issue where dragging and dropping a file on pre-chromium edge would result in console errors and the overlay not appearing
- Edge does not support `.contains` so replacing it with `.includes` fixes the drag and drop issue 
- This issue is only reproducible on pre-chromium edge.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22937


Willyfrog:
@fm2munsh can you fix the tests before we can merge?

fm2munsh:
@Willyfrog Its the build step failing with some mattermost-redux issue (also seems to be happening on master) not sure whats causing it... but its definitely not related to this PR

fm2munsh:
Looks like the issue is highlighted here https://mattermost.atlassian.net/browse/MM-23538

PR #5080, opened 2020-03-19
E2E: add partial default config which aimed to not require reverting config changes

saturninoabril:
#### Summary
- add partial default system config which aimed to not require reverting config changes. This fixes the issue of `mention_email_notification_spec` and others. Note this has corresponding PR on private repo that updates config of Cypress test server
- fix tests on `customization_spec`

#### Ticket Link
none


saturninoabril:
@srkgupta Good catch, will update it and other test that requires LDAP. Thanks.

srkgupta:
> @srkgupta Good catch, will update it and other test that requires LDAP. Thanks.

@saturninoabril LDAP was just an example. It will also affect other tests for which the default settings were disabled in the platform-private PR. For example:
ExperimentalTimezone, ExperimentalEnableAutomaticReplies

saturninoabril:
For `ExperimentalTimezone`, that is handled already by the test itself like in accessibility and local date time. For `ExperimentalEnableAutomaticReplies`, I don't see which test requires it.

e2e/cypress/integration/emoji/react_to_last_message_shortcut_spec.js
```diff
@@ -42,14 +42,6 @@ describe('Keyboard shortcut for adding reactions to last message in channel or t
         });
     });
```
josephbaylon: Getting this failure consistently
Screen Shot 2020-03-19 at 10 33 04 AM

saturninoabril: Oh I see.  It's not related to this PR but a bug introduced by https://github.com/mattermost/mattermost-webapp/commit/c6c1b5e205ac1994e3c41247fdb687e6b3ac4774.  I suggests to deal with it separately.

PR #5090, opened 2020-03-19
MM-22757 Fix for date toast placement in web mobile view

sudheerDev:
#### Summary
Push the toast by 50px when toast is present

Issue: `FloatingTimestamp` component need to know if any toasts are present in the UI. toasts state is only available in ToastWrapper as of now.

Solutions:
1. Either use DOM manipulation like we do in few places of codebase and add a class when toast is present with `add.classList`
2. Add a redux state to keep track of toast and use a selector 
3. Use something like ReactClone and get the required prop from `ToastWrapper`.
4. We could have `FloatingTimestamp` in ToastWrapper 

i went with 3 reasons being
1. Really hacky and msot places we use in codebase tend to fail and cause regressions 
2. Feel like too much of work for such a small use case
4. need to add couple more props to ToastWrapper so it can be passed to  `FloatingTimestamp`

#### Ticket Link
Mar-20-2020 03-38-41
https://mattermost.atlassian.net/browse/MM-22757

### DIFF
--- a/actions/views/channel.test.js
+++ b/actions/views/channel.test.js
@@ -519,4 +519,15 @@ describe('channel view actions', () => {
             expect(markChannelAsRead).not.toHaveBeenCalled();
         });
     });
+
+    describe('updateToastStatus', () => {
+        test('should disptach updateToastStatus action with the true as argument', async () => {
+            await store.dispatch(Actions.updateToastStatus(true));
+
+            expect(store.getActions()).toEqual([{
+                data: true,
+                type: 'UPDATE_TOAST_STATUS'
+            }]);
+        });
+    });
 });
--- a/components/post_view/floating_timestamp/floating_timestamp.jsx
+++ b/components/post_view/floating_timestamp/floating_timestamp.jsx
@@ -14,22 +14,24 @@ export default class FloatingTimestamp extends React.PureComponent {
             PropTypes.instanceOf(Date),
             PropTypes.number,
         ]).isRequired,
+        toastPresent: PropTypes.bool,
         isRhsPost: PropTypes.bool,
-        stylesOverride: PropTypes.object,
     }
 
     render() {
-        if (!this.props.isMobile) {
+        const {isMobile, createAt, isScrolling, isRhsPost, toastPresent} = this.props;
+
+        if (!isMobile) {
             return null;
         }
 
-        if (this.props.createAt === 0) {
+        if (createAt === 0) {
             return null;
         }
 
         const dateString = (
             <RecentDate
-                value={this.props.createAt}
+                value={createAt}
                 weekday='short'
                 day='2-digit'
                 month='short'
@@ -38,18 +40,18 @@ export default class FloatingTimestamp extends React.PureComponent {
         );
 
         let className = 'post-list__timestamp';
-        if (this.props.isScrolling) {
+        if (isScrolling) {
             className += ' scrolling';
         }
 
-        if (this.props.isRhsPost) {
+        if (isRhsPost) {
             className += ' rhs';
         }
 
         return (
             <div
                 className={className}
-                style={this.props.stylesOverride}
+                style={toastPresent ? {top: '50px'} : null}
                 data-testid='floatingTimestamp'
             >
                 <div>
--- a/components/post_view/floating_timestamp/floating_timestamp.jsx
+++ b/components/post_view/floating_timestamp/floating_timestamp.jsx
@@ -48,10 +48,13 @@ export default class FloatingTimestamp extends React.PureComponent {
             className += ' rhs';
         }
 
+        if (toastPresent) {
+            className += ' toastAdjustment';
+        }
+
         return (
             <div
                 className={className}
-                style={toastPresent ? {top: '50px'} : null}
                 data-testid='floatingTimestamp'
             >
                 <div>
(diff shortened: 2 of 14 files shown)


sudheerDev:
@nevyangelova @Willyfrog Let me know what you guys think. I am going to add tests after you guys have a look at it.

components/post_view/floating_timestamp/floating_timestamp.jsx
```diff
@@ -38,18 +40,18 @@ export default class FloatingTimestamp extends React.PureComponent {
...
-        if (this.props.isRhsPost) {
+        if (isRhsPost) {
             className += ' rhs';
         }
 
         return (
             <div
                 className={className}
-                style={this.props.stylesOverride}
+                style={toastPresent ? {top: '50px'} : null}
```
Willyfrog: Why don’t we  use classname for this? We are already adding classes depending on what is around the floating timestamp, so another if wouldn’t hurt and it would probably be cleaner.

nevyangelova: 0/5 maybe we can drive this by className.

components/toast_wrapper/toast_wrapper.jsx
```diff
@@ -229,45 +230,50 @@ class ToastWrapper extends React.PureComponent {
...
-        if (this.state.showUnreadToast && this.state.unreadCount > 0) {
+        if (showUnreadToast && unreadCount > 0) {
             unreadToastProps = {
                 ...unreadToastProps,
                 onDismiss: this.hideUnreadToast,
                 onClick: this.scrollToLatestMessages,
                 onClickMessage: Utils.localizeMessage('postlist.toast.scrollToBottom', 'Jump to recents'),
                 show: true,
-                showActions: !this.props.atLatestPost || (this.props.atLatestPost && !this.props.atBottom),
+                showActions: !atLatestPost || (atLatestPost && !atBottom),
```
nevyangelova: 1/5 can be simplified to !(atLatestPost && atBottom)

components/post_view/floating_timestamp/floating_timestamp.jsx
sudheerDev: i am not a fan of using classes if they are not reusable. It is hard to clean up with the amount of css we already have. I am gonna change to class though.

components/toast_wrapper/toast_wrapper.jsx
sudheerDev: Aah, Nice. i cannot wrap my head around conditions usually. Will change

components/post_view/floating_timestamp/floating_timestamp.jsx
sudheerDev: Classes already exist and there weren't part of the PR. As i said i am not a fan personally to use classes in out codebase anymore for things which are not reusable, or at least applied unconditionally.  Things like these tend to make maintenance of css hard. There were some arguments in the past about performance of css vs inline styles in react but with the level of nesting we normally create the browsers will have a much harder time figuring out the specificity IMO.

In this case i am gonna change to class as it is popular behaviour and i am like 1/5 on this.

sudheerDev:
> So I think I’d have gone for #2 as it would allow for further usage if needed, not only for this view but for any other we might have. Having a way to know if the toast is present and probably some info with it, might be an easy way to share all the calculations already done for displaying on the toast.
Ya, i think i am gonna change it to have a state in redux.

> With that said, this is an interesting approach, I didn’t know about react.clone, so that’s a new thing, but if it will always depends on the toast to be rendered, shouldn’t we pass the needed info to have the toast render it and remove cloning? Or are we planning to reuse the toast on other situations?

The problem removing the cloning and passing in the props is that we need `isMobile` prop which means i need to add a state and a listener again in this component or pass that to `ToastWrapper` which only needs because of the `FloatingTimestamp`. Seemed like it has its own downsides.

sudheerDev:
@Willyfrog @nevyangelova this is up for review.

actions/views/channel.test.js
```diff
@@ -519,4 +519,15 @@ describe('channel view actions', () => {
...
         });
     });
+
+    describe('updateToastStatus', () => {
+        test('should disptach updateToastStatus action with the true as argument', async () => {
+            await store.dispatch(Actions.updateToastStatus(true));
+
+            expect(store.getActions()).toEqual([{
+                data: true,
+                type: 'UPDATE_TOAST_STATUS'
```
Willyfrog: slightly concerned about the naming here: what should someone think of "update_toast_status=true"? maybe we should rename into something more descriptive?

sudheerDev: i am 0/5 on this but this is not something that people stumble up on when reading code IMO.

components/post_view/floating_timestamp/floating_timestamp.jsx
```diff
@@ -38,18 +40,21 @@ export default class FloatingTimestamp extends React.PureComponent {
...
             className += ' rhs';
         }
 
+        if (toastPresent) {
+            className += ' toastAdjustment';
+        }
+
         return (
             <div
                 className={className}
```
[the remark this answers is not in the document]
sudheerDev: Done. Changed it i guess it is a new convention but If else is better human readable code than functions and conditional operators.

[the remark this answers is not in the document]
sudheerDev: Thanks, i didn't know we were using classNames util though. Will try and make it a habit.

jgilliam17:
Please disregard all the label changes and a review re-request. 
I misunderstood that history toast fix will be included in this PR and changed a bunch labels 😞

PR #5094, opened 2020-03-20
[MM-23428] - Check for single new lines in channel header popover

nevyangelova:
#### Summary
After merging the functionality for the new channel header popover to expand on hover and also show only when the lines are longer than the space available, there was a bug introduced, when there are bullet points in the description but the first line was as long as the space available, you couldn't expand the popover. This fix checks for single new lines as well and it seems to cover all edge cases. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23428

Fix versions
v5.22 (April 2020)

### DIFF
--- a/components/channel_header/channel_header.js
+++ b/components/channel_header/channel_header.js
@@ -484,7 +484,7 @@ class ChannelHeader extends React.PureComponent {
                         >{popoverContent}</Overlay>
 
                         <Markdown
-                            message={headerText.replace(/\n/, '\n\n')}
+                            message={headerText.replace(/\n/g, '\n\n')}
                             options={this.getHeaderMarkdownOptions(channelNamesMap)}
                         />
                     </span>
--- a/components/channel_header/channel_header.js
+++ b/components/channel_header/channel_header.js
@@ -462,7 +462,7 @@ class ChannelHeader extends React.PureComponent {
                         ref={this.headerPopoverTextMeasurerRef}
                     >
                         <Markdown
-                            message={headerText.replace(/\n{2,}/g, ' ').replace(/\n/, ' ')}
+                            message={headerText.replace(/\n+/g, ' ')}
                             options={this.getHeaderMarkdownOptions(channelNamesMap)}
                         /></div>
                     <span
--- a/components/channel_header/channel_header.js
+++ b/components/channel_header/channel_header.js
@@ -442,7 +442,7 @@ class ChannelHeader extends React.PureComponent {
                         onClick={this.handleFormattedTextClick}
                     >
                         <Markdown
-                            message={headerText}
+                            message={headerText.replace(/\n/, '\n\n')}
                             options={this.getPopoverMarkdownOptions(channelNamesMap)}
                         />
                     </span>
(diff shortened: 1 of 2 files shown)


components/channel_header/channel_header.js
```diff
@@ -462,7 +462,7 @@ class ChannelHeader extends React.PureComponent {
                         ref={this.headerPopoverTextMeasurerRef}
                     >
                         <Markdown
-                            message={headerText.replace(/\n{2,}/g, ' ')}
+                            message={headerText.replace(/\n{2,}/g, ' ').replace(/\n/, ' ')}
```
Willyfrog: why not changing the first expression instead of replacing twice? also, your regex will only happen on the first line (was that intended?)
```suggestion
                            message={headerText.replace(/\n/g, ' ')}
```
also, should we trim spaces at the begining? like `    \n`, not the most common case, but people tend to hit the spacebar sometimes.

nevyangelova: First it looks for double newlines and then single newlines, the trick is that it only finds the single newlines after the bigger ones are already replaced. But it might not cover all edge cases. I am really open to suggestions on this.

sudheerDev:
Swapping reviewers as HH might have better opinion on this.

components/channel_header/channel_header.js
Willyfrog: sorry, but I think i'm missing something, what difference would make on the result? although to be fair it would look better if we do it like:
```js
    message={headerText.replace(/\n+/g, ' ')}
```
to prevent having multiple spaces (not that it matters much) but it would be done on a single regex instead of 2

nevyangelova: I will see how it behaves, makes sense what you are saying :)

jgilliam17:
Thank you @nevyangelova the issue as reported in the linked ticket is fixed and the popover shows as expected when first line is fully visible and without ellipses. 


I have one concern here and it relates to what user sees on the preview of channel description modal compared to what is finally displayed in the popover.
 In 5.21 (and before) what was shown on the modal as preview mirrored the popover layout, but now we are going away from that and text often merges, disregarding new lines, when it displays on the popover
See second row. @esethna any thoughts?
|Edit modal Preview for both versions| Before (v5.21)  | Now (v5.22)|
|------------| ------------- | ------------- |
|Screen Shot 2020-03-23 at 11 19 33 AM|Screen Shot 2020-03-23 at 11 11 21 AM| Screen Shot 2020-03-23 at 11 12 03 AM|
|Screen Shot 2020-03-23 at 11 21 35 AM|Screen Shot 2020-03-23 at 11 31 05 AM|Screen Shot 2020-03-23 at 11 30 52 AM|

nevyangelova:
Opps this was meant to be fixed @jgilliam17  sorry got confused with the comments and re-introduced the behaviour, pushing the fix now

esethna:
@jgilliam17 I'm not seeing the behaviour you described above ^ on this server

jgilliam17:
> @jgilliam17 I'm not seeing the behaviour you described above ^ on this server

@esethna for the second row of images-  If you look at the Off Topic channel, compare the channel description popover on that channel versus edit channel description modal in preview mode

esethna:
> If you look at the Off Topic channel, compare the channel description popover on that channel versus edit channel description modal in preview mode

The observed on the server isn't the same as in your post. Can you file a separate ticket since it seems out of scope for this one?  

image

jgilliam17:
> > If you look at the Off Topic channel, compare the channel description popover on that channel versus edit channel description modal in preview mode
> 
> The observed on the server isn't the same as in your post. Can you file a separate ticket since it seems out of scope for this one?
> 
> image

You're totally right. I mixed up my screenshots. 
I'll file a separate ticket.

PR #5099, opened 2020-03-20
[MM-22988] return focus to input

Willyfrog:
#### Ticket Link

[MM-22988](https://mattermost.atlassian.net/browse/MM-22988)


Willyfrog:
thanks for the reminder @ali-farooq0 I'll add the e2e tests for that

ali-farooq0:
> thanks for the reminder @ali-farooq0 I'll add the e2e tests for that

@Willyfrog also the e2e test doesn't have to be part of this PR as i know that can take some time. I would suggest creating a jira and getting to it later, and just merging this PR for now. :)
When I suggested the test i was thinking more on the jest side than e2e.

components/quick_input.jsx
```diff
@@ -95,6 +95,7 @@ export default class QuickInput extends React.PureComponent {
         if (this.props.onClear) {
             this.props.onClear();
         }
+        this.focus();
```
sudheerDev: Looks fine to me but could you link on how we caused this regression?

Willyfrog: i don't think it ever had its focus back, at somepoint it didn't have the X button to clear, so it probably wasn't needed. we later added the x, but never thought on focus.

PR #5102, opened 2020-03-21
[MM-23448] - Fix drag and drop in desktop app

nevyangelova:
#### Summary
After some digging it turned out that the events were working correctly in Desktop app, but the button wasn't draggable. Actually only the side indicator was draggable and working even before that change. Now that we are using the same component for webapp and desktop it works as expected. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23448

Fix versions
v4.5 Desktop App


esethna:
Nice find, thanks Nevy!

amyblais:
@nevyangelova Has this been cherry-picked?

PR #5105, opened 2020-03-23
[MM-18953] Migrate 'components/change_url_modal' to TypeScript

haydenhw:
#### Summary
Migrates 'components/change_url_modal' module and associated tests to TypeScript

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/12473


components/change_url_modal/change_url_modal.tsx
```diff
@@ -136,9 +144,9 @@ export default class ChangeURLModal extends React.PureComponent {
         return error;
     }
 
-    onSubmit = (e) => {
+    onSubmit = (e: React.MouseEvent<HTMLButtonElement>) => {
         e.preventDefault();
-        const url = this.refs.urlinput.value;
+        const url = this.urlInput.current ? this.urlInput.current.value : '';
```
haydenhw: When I left this simply as `const url = this.urlInput.current.value;` Typescript gave me this error:
`TS2531: Object is possibly 'null'`

Not sure if there's a cleaner fix for this than the ternary expression I used?

cpoile:
@devinbinnie I don't think I'm a good fit for this PR -- I've only done a little TS, so I'm still on the learning curve. Can you switch someone else in please?

devinbinnie:
@cpoile It's important that all staff members get a handle on TypeScript, since all projects are being converted over to use it. Please review, and if you have questions/need any advice, please let me know and I'll be happy to assist :)

components/change_url_modal/change_url_modal.test.tsx
```diff
@@ -65,25 +65,28 @@ describe('components/ChangeURLModal', () => {
...
             <ChangeURLModal {...baseProps}/>
         );
-        const refURLInput = wrapper.find('input[type="text"]').instance();
-        refURLInput.value = 'urlexample';
+        const value = 'urlexample';
+        const refURLInput = wrapper.find('input[type="text"]').getDOMNode() as HTMLInputElement;
+        refURLInput.value = value;
 
-        wrapper.instance().onSubmit({preventDefault: jest.fn()});
+        const instance = wrapper.instance() as ChangeURLModal;
```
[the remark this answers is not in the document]
cpoile: Fine by me, especially since it's a test. You know what you're casting after all.

components/change_url_modal/change_url_modal.tsx
```diff
@@ -12,73 +11,82 @@ import Constants from 'utils/constants';
...
+    currentTeamURL: string;
 
-        /**
-        * Server error from failed channel creation
-        */
-        serverError: PropTypes.node,
+    /**
+     * Server error from failed channel creation
+     */
+    serverError: JSX.Element | string | null;
```
cpoile: Will this work if it's a JSX.Node? That should allow null assignment (not sure about string though).

cpoile:
@devinbinnie Makes sense. 👍

components/change_url_modal/change_url_modal.tsx
```diff
@@ -12,73 +11,82 @@ import Constants from 'utils/constants';
...
         title: 'Change URL',
         submitButtonText: 'Save',
         currentURL: '',
         serverError: null,
     }
 
-    constructor(props) {
+    constructor(props: Props) {
         super(props);
+        this.urlInput = React.createRef();
```
cpoile: 👍

components/change_url_modal/change_url_modal.tsx
cpoile: That seems fine to me.

components/change_url_modal/change_url_modal.tsx
```diff
@@ -12,73 +11,82 @@ import Constants from 'utils/constants';
...
+    title: JSX.Element | string;
 
-        /**
-        * Set to change the submit button text
-        */
-        submitButtonText: PropTypes.node,
+    /**
+     * Set to change the submit button text
+     */
+    submitButtonText: JSX.Element | string;
```
hahmadia: The equivalent for `PropTypes.node` in typescript is actually `React.ReactNode`.
As this not required, could we add a conditional `?` to it?

components/change_url_modal/change_url_modal.tsx
```diff
@@ -12,73 +11,82 @@ import Constants from 'utils/constants';
...
+    show: boolean;
 
-        /**
-        * Set to change the title of the modal
-        */
-        title: PropTypes.node,
+    /**
+     * Set to change the title of the modal
+     */
+    title: JSX.Element | string;
```
hahmadia: The equivalent for `PropTypes.node` in typescript is actually `React.ReactNode`.
As this not required, could we add a conditional `?` to it? Like this
```suggestion
    title?: React.ReactNode;
```

components/change_url_modal/change_url_modal.tsx
```diff
@@ -12,73 +11,82 @@ import Constants from 'utils/constants';
...
+    submitButtonText: JSX.Element | string;
 
-        /**
-        * Set to change the current URL
-        */
-        currentURL: PropTypes.string,
+    /**
+     * Set to change the current URL
+     */
+    currentURL: string;
```
hahmadia: As this not required, could we add a conditional `?` to it? 
```suggestion
    currentURL?: string;
```

components/change_url_modal/change_url_modal.tsx
hahmadia: The equivalent for `PropTypes.node` in typescript is actually `React.ReactNode`. Could we change this to the following:
```suggestion
    serverError?: React.ReactNode;
```

components/change_url_modal/change_url_modal.tsx
hahmadia: Yep same for me! I don't see a problem with this.

hahmadia: Actually I just wanted to make an update. Currently our ESlinter doesn't allow this (not setup correctly for Typescript) however in such scenarios, Typescript best practices is to use the optional conditioning (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining). So in this case it would look something like 
`this.urlInput?.current?.value`.

hahmadia: UPDATE! We have updated ESlint. You can now do
`const url = this.urlInput?.current?.value` and you should not see the `Object is possibly null` issue. Let me know how it goes!

haydenhw: Optional conditioning... very cool! It worked out well.

Thanks for the docs link.

components/change_url_modal/change_url_modal.tsx
haydenhw: Fixed !

components/change_url_modal/change_url_modal.tsx
```diff
@@ -12,73 +11,82 @@ import Constants from 'utils/constants';
 import {getShortenedURL, cleanUpUrlable} from 'utils/url';
 import {t} from 'utils/i18n';
 
-export default class ChangeURLModal extends React.PureComponent {
-    static propTypes = {
+export type Props = {
```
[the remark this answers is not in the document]
haydenhw: Fixed!

components/change_url_modal/change_url_modal.tsx
haydenhw: Fixed!

components/change_url_modal/change_url_modal.tsx
haydenhw: Fixed!

components/change_url_modal/change_url_modal.tsx
haydenhw: Fixed!

components/change_url_modal/change_url_modal.tsx
```diff
@@ -146,7 +146,7 @@ export default class ChangeURLModal extends React.PureComponent<Props, State> {
 
     onSubmit = (e: React.MouseEvent<HTMLButtonElement>) => {
         e.preventDefault();
-        const url = this.urlInput.current ? this.urlInput.current.value : '';
+        const url = this.urlInput?.current?.value || '';
```
haydenhw: Still converting a potential ```null``` value to a string here since the ```cleanUpUrlable``` function only takes strings

PR #5107, opened 2020-03-23
MM-23367 Add test metadata to selectively run Cypress tests

saturninoabril:
#### Summary
__Changes included in this PR__
- add test metadata to group or skip tests like `// @prod @messaging @headless` --> To follow, will submit PR to docs
- moved spec files into its folder according to groups, mostly following tabs in the spreadsheet
- clean up spec headers
- assigned all 167 spec files as `@prod` and 50 as `@smoke` --> To follow, will sort out which specs should qualify as prod and smoke tests (to be decided by QA team).

__Test Metadata details__:
I initially designed grouping/filtering to be flexible up to `describe/context` and `it/test` block by using Mocha suite but there's some issue when running Cypress tests.  So, I ended up dong it on a higher level, which is per file basis.  Here are some of the details:
- Assign test metadata by adding comment per file
  - Add stage like `// Stage: @prod @smoke` which means a spec file is for production and smoke testing
  - Add group like `// Group: @accessibility`
  - Exclude test environment `Skip: @headless` which means a test can run in all environment except on headless mode
  - "Stage:", "Group:" and "Skip:" are written for guide only. The important keyword is "@" followed by a "word"
- Filtering spec files before passing into Cypress makes the reporting accurate since it doesn't need to mark a describe/test as skip or pending on runtime. It also prevents Cypress from unexpected error since it doesn't need to run hooks (e.g. before, after...) if test will eventually skipped.
- Easier to invert matches. If there are 7/10 assigned as `@prod`, inverting it by `node run_tests --stage='@prod' --invert` will run test to other 3/10 spec files. Those unassigned spec files can be from a newly submitted PR.

__How to run__:
1. Run with `node run_tests.js` - tests all spec files
2. Add `--stage` flag, like `node run_tests.js --stage='@prod'` - tests all files marked as `@prod`.  It can be of multiple values, comma-separated which will tests files if either of values were found.
Screen Shot 2020-03-23 at 10 06 51 PM
3. Add `--group` flag like `node run_tests.js --group='@accessibilty'`. Same logic as stage flag.
4. Add `--invert` to invert matches, like `node run_tests.js --group='@accessibilty' --invert` - test all files excluding `@accessibilty` group
5. Can combine all flags, like `node run_tests.js --stage='@prod @smoke ' --group='@messaging @accessibilty' --invert` - test all files excluding those that matches either stage values AND either group values.
6. Test can be skipped by environment: platform, browser or mode (headless/headed). If a test should run only on Chrome (headed) on any platform, then assigning with `@headless @firefox @electron` will evaluate whether the test can run on a given environment.
Screen Shot 2020-03-23 at 10 07 13 PM

#### Ticket Link
JIRA ticket: https://mattermost.atlassian.net/browse/MM-23367


srkgupta:
Hi @saturninoabril 
Great Work. Couple of suggestions:
1) Can we add a README which details all the options available and some examples on how to run run_tests.js (as you have mentioned in the PR's description). It can also include information on how to run headed vs headless mode etc. 

2) Is there a way to get list of all groups, stages so that it could be referenced while writing the test. We just do not want to have duplicate objects conveying the same group/stage. For example: Someone might just use `elasticsearch` instead of `elasticsearch_autocomplete`. We can have this list in the README file itself (or) any other separate files. 

3) There are some merge conflicts. Can you please resolve the same.

saturninoabril:
@srkgupta Thanks for the review.
1. My original plan is to add it to https://developers.mattermost.com/contribute/webapp/end-to-end-tests/, will submit PR today.  Let me know if that's fine with you, instead of adding README here.
2. Good point.
  - In getting the list, I added additional npm script - ex. `META=Group npm run uniq-meta`, "META" can be Stage, Group or even Skip (though not meaningful enough for Skip).
  - I can add the list in the documentation but such would be on general term only, like those list should follow the naming in tabs of release testing spreadsheet.  Others that don't have corresponding name is subject for QA team's approval.  Note that the test metadata should be controlled/added/edited only by QA team.

Now updated. PTAL.

saturninoabril:
@srkgupta Added per request - https://github.com/mattermost/mattermost-webapp/pull/5107/commits/a1156bfba9f47353c02ebf701e741b98a203680e#diff-39eb77461c7ea5114f9b1c1a57805effR6-R40

srkgupta:
Awesome. Thanks Saturn. Looks great

e2e/run_tests.js
```diff
@@ -77,21 +120,128 @@ function writeJsonToFile(jsonObject, filename, dir) {
...
+        console.log(chalk.magenta.bold(`${invert ? 'All Except --> ' : ''}${testStage}${stage && group ? '| ' : ''}${testGroup}`));
+        console.log(chalk.magenta(`(Testing ${i + 1} of ${finalTestFiles.length})  - `, testFile));
 
-    for (const dir of testDirs) {
         const {totalFailed} = await cypress.run({
-            browser: `${BROWSER}`,
-            headless: 'true',
-            spec: `./cypress/integration/${dir}/**/*`,
+            browser: BROWSER,
+            headless: HEADLESS,
```
josephbaylon: I'm guessing the possible values for HEADLESS is undefined, headed, headles, true, etc. If it's headed or any other string, would headless parameter recognize the value or would it set it to true if it's a non-empty string?

e2e/run_tests.js
```diff
@@ -77,21 +120,128 @@ function writeJsonToFile(jsonObject, filename, dir) {
...
+    }
+
+    // Return files if both in stage and group
+    return intersection(stageFiles, groupFiles);
+}
+
+function getSkippedFiles(initialTestFiles, platform, browser, headless) {
+    const platformFiles = grepFiles(grepCommand(`@${platform}`));
+    const browserFiles = grepFiles(grepCommand(`@${browser}`));
+    const headlessFiles = grepFiles(grepCommand(`@${headless ? 'headless' : 'headed'}`));
```
josephbaylon: If headless='headed' and some spec files have `Skip: @headed`, wouldn't that skip those files even if the intent was to run headed on those files? Or is `@headed` not allowed in Skip: only `@headless`?

saturninoabril: I'd say, the intention is not correct. It's allowed to skip both of those.

e2e/run_tests.js
saturninoabril: No, values of HEADLESS is either true or false only - https://github.com/mattermost/mattermost-webapp/pull/5107/files#diff-39eb77461c7ea5114f9b1c1a57805effR219.

PR #4825, opened 2020-02-04
MM-11505 Update permalink view

sudheerDev:
#### Summary
  * Use existing channel view for permalinks
  * Add a new URL route so channelView can be used for permalinks
  * Add new history toast
  * Highlight of permalink post should fadeout after 6 secs. 
  * new permalink url changes to channel url after 6 secs. 
  
One of the bigger changes in the code is to remove the view state of focussedPostId as having that and depending on that causes issues when we change url to channel view. When url changes to channel view it will clear the redux view state and hence the selectors for posts will also change from permalink posts to channel posts. Instead we will be depending on react state instead for permalink post id and use that for selectors and clear that only on channel change again.

 Initial idea was to drop the use of `/pl/postId` url from UI and use only the new url `/channel/postId` but this will cause us to to expose channelNames at times raising few security and privacy concerns. 
We will use the old url in the UI but redirect internally to new url to show the channelView instead. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-11505
https://mattermost.atlassian.net/browse/MM-19783
https://mattermost.atlassian.net/browse/MM-19782
https://mattermost.atlassian.net/browse/MM-20614

#### Screenshots
Screenshot 2020-02-04 at 9 58 08 PM

PR #5109, opened 2020-03-23
MM-23458 Fix missing toast for unreads

sudheerDev:
#### Summary
This should be a race cause with atBottom somehow changing this to be how it used to be in 5.20v. I changed the condition for not mounting the toasts till `atBottom` has an assigned value is because to have a simpler logic on mount. 
The change was made at https://github.com/mattermost/mattermost-webapp/pull/4825/files#diff-dacf7e63a72c7be54c626d95244525efL560 

* Not sure of the tests to add here, suggestions are welcome.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23458

#### Related Tickets
Ticket to fix the onScroll event in virt list https://mattermost.atlassian.net/browse/MM-23494

### DIFF
--- a/components/toast_wrapper/toast_wrapper.jsx
+++ b/components/toast_wrapper/toast_wrapper.jsx
@@ -40,10 +40,8 @@ class ToastWrapper extends React.PureComponent {
 
     constructor(props) {
         super(props);
-        const showMessageHistoryToast = props.focusedPostId !== '' && (props.initScrollOffsetFromBottom > 1000 || !props.atLatestPost);
         this.state = {
             unreadCountInChannel: props.unreadCountInChannel,
-            showMessageHistoryToast,
         };
     }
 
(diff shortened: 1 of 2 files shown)


components/toast_wrapper/toast_wrapper.jsx
```diff
@@ -235,7 +238,7 @@ class ToastWrapper extends React.PureComponent {
         };
 
         const archiveToastProps = {
-            show: this.state.showMessageHistoryToast,
+            show: this.state.showMessageHistoryToast || false,
```
Willyfrog: minor nitpick
```suggestion
            show: Boolean(this.state.showMessageHistoryToast),
```

components/toast_wrapper/toast_wrapper.jsx
```diff
@@ -73,6 +71,10 @@ class ToastWrapper extends React.PureComponent {
             showUnreadToast = unreadCount > 0 && !props.atBottom;
         }
 
+        if (typeof showMessageHistoryToast === 'undefined' && props.focusedPostId !== '') {
```
Willyfrog: if focusedPostId is empty, show messageHistoryToast can be `undefined`, is this expected?

sudheerDev: Yes, it can be that means it was not set Previously. Ideally i would have preferred to have this in mount by not mounting this component till it is needed. But i cannot do that because of this particular issue and the shortcomings of virt list.

PR #5111, opened 2020-03-23
[MM-22877] - Fix download image button tooltip

nevyangelova:
#### Summary
Fixes the download image tooltip that was off-centre because it was relating to the svg and not the whole button.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22877

Fix versions
v5.22 (April 2020)


asaadmahmood:
@nevyangelova If I remove the padding, this is what i see.
Screenshot 2020-03-24 at 11 55 16 AM
With the padding, I see this.
Screenshot 2020-03-24 at 11 55 24 AM

The issue looks like the button is not taking its original height.
So we can just change a of `display: inline-block`, or `block` to the span, and it would be fixed.
The padding solution may not work in cases where the button is larger, because it is still not fixing the main issue.

You can try increasing the height/width on the button and see that the padding may not work.
Screenshot 2020-03-24 at 11 58 15 AM

PR #5116, opened 2020-03-23
MM-23203 Fix for missing new messages line on socket reconnect

sudheerDev:
#### Summary
  * Add an additional check to `atBottom` with `null`

Theoretically this could happen with this series of events
1. User switches channel to a previously visited out of sync channel(System from sleep)
2. atBottom value is not set yet but the posts from since API loads.

The condition `this.props.postListIds[0] !== prevProps.postListIds[0];` is true and `!this.props.atBottom` is also true making a call `updateNewMessagesAtInChannel` and updating the line.

It would have been perfect if the logic was considered only for new messages but atm it is not possible and has a high dev cost to it too. Possibly with upcoming changes to redux state idata .e marking channels out of sync can help.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23203


hmhealey:
Makes sense to me to be stricter on the checks for it. This does have some conflicts that need to be resolved though

PR #5120, opened 2020-03-24
Adding bleve to the admin console

jespino:
#### Summary
Adding bleve to the admin console

#### Ticket Link
[MM-23144](https://mattermost.atlassian.net/browse/MM-23144)


wiersgallak:
@mgdelacroix Per our discussion, this can actually be moved to the Experimental section at the bottom of the LHS.

actions/admin_actions.jsx
```diff
@@ -254,6 +254,20 @@ export async function elasticsearchPurgeIndexes(success, error) {
     }
 }
 
+export async function blevePurgeIndexes(success, error) {
+    const purgeBleveIndexes = bindClientFunc({
```
ethervoid, line 258: Are we calling here the client function instead of in redux because of the deprecation?

mgdelacroix: It was initially put in the redux PR and moved here [out of a discussion there](https://github.com/mattermost/mattermost-redux/pull/1094#discussion_r397226988). The only piece needed in redux is the client function.

PR #5129, opened 2020-03-24
MM-23559: change quick input "X" tooltip text

lieut-data:
#### Summary
Default to `Clear`, but allow future customization via a new prop. All existing instances of this component use the default.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-23559

#### Screenshots
image
image
image

### DIFF
--- a/components/__snapshots__/quick_input.test.jsx.snap
+++ b/components/__snapshots__/quick_input.test.jsx.snap
@@ -1,0 +1,199 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`components/QuickInput should render clear button with customized tooltip component 1`] = `
+<div
+  className="input-clear visible"
+  onClick={[Function]}
+>
+  <OverlayTrigger
+    defaultOverlayShown={false}
+    delayShow={400}
+    overlay={
+      <Tooltip
+        bsClass="tooltip"
+        id="InputClearTooltip"
+        placement="right"
+      >
+        <span>
+          Custom
+        </span>
+      </Tooltip>
+    }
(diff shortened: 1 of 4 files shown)


lieut-data:
@andrewbrown00, heads up that I've added you as reviewer to confirm the correct implementation of our discussion at https://community-daily.mattermost.com/core/pl/dwr46bh69fbc5es51fxypkctay.

components/__snapshots__/quick_input.test.jsx.snap
```diff
@@ -0,0 +1,199 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
lieut-data: I normally prefer to avoid snapshots, but in this case it seemed onerous to bind the tests to a particular prop deep in the rendered hierarchy. So I verified the expected text changes, and then recorded them as the full snapshot.

components/__snapshots__/quick_input.test.jsx.snap
```diff
@@ -0,0 +1,199 @@
...
+    }
+    placement="bottom"
+    trigger={
+      Array [
+        "hover",
+        "focus",
+      ]
+    }
+  >
+    <OverlayTrigger
```
lieut-data: This double nesting surprised me, but it's not an artifact of my changes. It seems to just be how we propagate the props through to `BaseOverlayTrigger`.

andrewbrown00:
@lieut-data will you spark up a test server so I can review the PR. I added the `Setup Cloud Test Server` label but the server creation **failed**.

PR #5132, opened 2020-03-24
[MM-23481] - Turn down opacity for muted channels

nevyangelova:
#### Summary
Turns down the opacity if a channel is muted

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23481#### Related Pull Requests


jgilliam17:
Thank you @nevyangelova 
- Verified muted channels have reduced opacity (greyed out) on the LHS

There is an issue where muted channel have their full opacity restored when mention badge is present see gif

muted channel with mention

nevyangelova:
@jgilliam17 good catch! I didn't check how mentions work. I think it looks good now

PR #5138, opened 2020-03-25
MM-22595: local date time fixes

lieut-data:
#### Summary
Chrome 80 introduced support for honouring the locale's `hourCycle`, but the current Unicode specifications configure `en-US` to render the hour after midnight as `24:00`-`24:59`.

Override this behaviour by forcing `h23` semantics (00-23) for every locale. Since react-intl doesn't [currently](https://github.com/formatjs/react-intl/issues/1577#issuecomment-603514480) support passing this field down, just use the `Intl.DateTimeFormat` directly.

The unit tests have been refactored in various commits, and then changed to assert the required behaviour before this fix was applied. I've tested the changes in the latest versions of Firefox, Chrome, Safari, and Edge on MacOS, but as with anything time related, I expect some surprises on edge cases I haven't considered.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-22595


components/local_date_time/local_date_time.tsx
```diff
@@ -48,14 +48,25 @@ class LocalDateTime extends React.PureComponent<Props> {
...
+        }
+        if (useMilitaryTime) {
+            options.hourCycle = 'h23';
+        } else {
+            options.hourCycle = 'h12';
+        }
 
-        if (formattedTime === String(date)) {
+        let formattedTime;
+        try {
```
lieut-data: I believe the previous check was effectively a mechanism to detect that `formatTime` was unsuccessful. `Intl.DateTimeFormat` happily throws an exception in this case, so I'm handling it that way.

components/local_date_time/local_date_time.tsx
```diff
@@ -48,14 +48,25 @@ class LocalDateTime extends React.PureComponent<Props> {
...
+            formattedTime = (new Intl.DateTimeFormat(this.props.intl.locale, options)).format(date);
+        } catch {
+            // Fallback to Moment.js as a default rendering strategy. We could probably just
```
lieut-data: I debated touching this code as described, but opted to keep the current semantics to reduce the number of changes here.

e2e/cypress/integration/messaging/local_date_time_spec.js
```diff
@@ -26,14 +26,17 @@ describe('Messaging', () => {
...
-            // # Post messages from the past and now
+            // # Post messages from the past
             [
-                Date.UTC(2020, 0, 5, 4, 30), // Jan 5, 2020 4:30am
-                Date.UTC(2020, 0, 5, 12, 30), // Jan 5, 2020 4:30am
-                Date.UTC(2020, 0, 5, 20, 30), // Jan 5, 2020 4:30am
+                Date.UTC(2020, 0, 5, 4, 30), // Jan 5, 2020 04:30
+                Date.UTC(2020, 0, 5, 12, 30), // Jan 5, 2020 12:30
+                Date.UTC(2020, 0, 5, 20, 30), // Jan 5, 2020 20:30
+                Date.UTC(2020, 0, 6, 0, 30), // Jan 6, 2020 00:30
```
lieut-data: This post is the key test addition.

e2e/cypress/integration/messaging/local_date_time_spec.js
```diff
@@ -49,46 +52,82 @@ describe('Messaging', () => {
...
-                name: 'for react-intl unsupported timezone',
+                name: 'in react-intl unsupported timezone',
                 channelIntro: 'Inicio de Channel Test',
                 locale: 'es',
                 manualTimezone: 'NZ-CHAT',
                 localTimes: [
-                    {postIndex: 0, match12hour: re12Hour, match24hour: re24Hour},
-                    {postIndex: 1, match12hour: re12Hour, match24hour: re24Hour},
-                    {postIndex: 2, match12hour: re12Hour, match24hour: re24Hour},
+                    {postIndex: 0, standard: '06:15 PM', military: '18:15'},
```
lieut-data: We'll see if this passes CI (@saturninoabril, can you help verify?), but it /seemed/ unnecessary to have a regular expression here given all the inputs are deterministic.

e2e/cypress/integration/messaging/local_date_time_spec.js
```diff
@@ -49,46 +52,82 @@ describe('Messaging', () => {
...
+                    {postIndex: 2, standard: '10:15 AM', military: '10:15'},
+                    {postIndex: 3, standard: '02:15 PM', military: '14:15'},
                 ]
             },
-        ].forEach((testCase) => {
-            it(testCase.name, () => {
-                verifyLocalizeTime(testCase);
+        ];
+
+        testCases.forEach((testCase) => {
```
lieut-data: Restructured this so that each post is asserted separately, making it much easier to iterate on the changes.

e2e/cypress/integration/messaging/local_date_time_spec.js
saturninoabril: As I remembered, it's due to DST period on that timezone that makes it inconsistent but I'm good as is.  Let's see how it goes in the next cycle.

e2e/cypress/integration/messaging/local_date_time_spec.js
```diff
@@ -49,46 +52,82 @@ describe('Messaging', () => {
...
+                    testCase.localTimes.forEach((localTime, index) => {
+                        it('post ' + index + ' should match', () => {
+                            // # Change user preference to 12-hour format
+                            setTo24HourTimeFormat(false);
+
+                            // # Set user locale and timezone
+                            setLocaleAndTimezone(testCase.locale, testCase.manualTimezone);
+
+                            // * Verify that the channel is loaded correctly based on locale
+                            cy.findByText(testCase.channelIntro).should('be.visible');
```
saturninoabril: This is failing on me due to channel intro being out of view partially.  I'd suggest to increase viewport height by `cy.viewport(1300, 800)` before each test. (We're currently using 600 height for testing but maybe we'll change this in the future as I'm seeing unnecessary "scrollIntoView" action).

components/local_date_time/local_date_time.tsx
```diff
@@ -48,14 +48,25 @@ class LocalDateTime extends React.PureComponent<Props> {
...
         const options = {
-            ...timezoneProps,
-            hour12: !useMilitaryTime,
+            hour: 'numeric',
+            minute: 'numeric',
         };
-        let formattedTime = this.props.intl.formatTime(date, options);
+        if (enableTimezone && timeZone) {
+            options.timeZone = timeZone;
+        }
```
fm2munsh: looks like this is causing a type check fail - quick fix would be to define `const options: any = {`

e2e/cypress/integration/messaging/local_date_time_spec.js
hmhealey: If it starts to fail again, we can file a ticket and then push it off for a few weeks until it stops reproducing 😛

components/local_date_time/local_date_time.tsx
lieut-data: Thanks, @fm2munsh! I ended up defining an inline interface here. I had hoped to use `Intl.DateTimeFormatOptions`, but found [it does not yet support hourCycle](https://github.com/microsoft/TypeScript/issues/34399).

e2e/cypress/integration/messaging/local_date_time_spec.js
lieut-data: Fixed!

lieut-data:
It looks like the tests are failing on CI because the version of Node is too old and doesn't support `hourCycle`. I may have to do some feature detection here to handle backwards compatibility.

lieut-data:
Apologies for the thrashing on this! @saturninoabril & @hmhealey, I've pushed two changes:
* feature test support for `hourCycle`, for backwards compatibility with Internet Explorer and old NodeJS versions. When `hourCycle` isn't supported, we just use `hour12` as before.
* `set TZ=Etc/UTC` for all tests. This eliminates the need to force a global timezone with moment, and fixes the unit tests given the new code paths.

I've also identified the reason for why Firefox fails the E2E tests. In Chrome, the following:
new Intl.DateTimeFormat("es", {timeZone: "NZ-CHAT", hour: 'numeric', minute: 'numeric'}).format()
11:40:59.857 VM263:1 Uncaught RangeError: Invalid time zone specified: NZ-CHAT
    at new DateTimeFormat (<anonymous>)
    at <anonymous>:1:1

amyblais:
@lieut-data Which version was this cherry-picked to?

lieut-data:
Apologies, @amyblais -- I merged this to v5.22 by mistake. Shall I revert, and keep this for v5.23 instead?

amyblais:
@lieut-data It's already cherry-picked so it's fine to keep it (as long as it's not risky)

PR #5151, opened 2020-03-25
[MM-23583] Fix all channel mentions getting highlighted even when they shouldn't

hahmadia:
#### Summary

All channel mentions getting highlighted even when they shouldn't has been fixed in this PR. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23583

#### Screenshots

image


fm2munsh:
Looks like there are some broken snapshot tests @hahmadia

PR #5152, opened 2020-03-25
Adding support for the new field for team_name in the post mentions

jespino:
#### Summary
Adding support for the new field for team_name in the post mentions. This is
added in the server PR (referenced at the bottom). At the same time, if the
field is not found in the post props it keeps the old behavior.

#### Ticket Link
[MM-22476](https://mattermost.atlassian.net/browse/MM-22476)

#### Related Pull Requests
- Has server changes (mattermost/mattermost-server#14150)


hmhealey:
@jespino I just saw that this went in. How does this work when the mentions come from the post? When the `ChannelNamesMap` is passed in by the `Markdown` component ([link](https://github.com/mattermost/mattermost-webapp/blob/master/components/markdown/index.js#L19)), that object doesn't have a `team_name` field. The `getChannelsNameMapInCurrentTeam` selector in `mattermost-redux` needs to be updated to support the `team_name` field or else we could cause problems.

amyblais:
@jespino Has this comment been resolved https://github.com/mattermost/mattermost-webapp/pull/5152#issuecomment-606879230?

jespino:
@hmhealey this works because we are adding that information through the server side PR. That information is only added for the new posts, but if the post doesn't provide this new information we simply not included the fix. That means, the fix is going to work for any new link created, but is going to keep working as was working before for any old post.

PR #5159, opened 2020-03-25
[MM-23517] - Fix channel header popover

ghost:
#### Summary
This fixes the line breaks that were removed in the markdown formatting. Checked the messages everywhere that use the 'marked' function and nothing looks broken. The style change is needed because on the new side bar the popover needs a different left offset. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23517

### DIFF
--- a/components/channel_header/channel_header.js
+++ b/components/channel_header/channel_header.js
@@ -85,6 +85,7 @@ class ChannelHeader extends React.PureComponent {
         }).isRequired,
         teammateNameDisplaySetting: PropTypes.string.isRequired,
         currentRelativeTeamUrl: PropTypes.string.isRequired,
+        newSideBarPreference: PropTypes.bool,
     };
 
     constructor(props) {
@@ -290,6 +291,7 @@ class ChannelHeader extends React.PureComponent {
             rhsState,
             hasGuests,
             teammateNameDisplaySetting,
+            newSideBarPreference,
         } = this.props;
         const {formatMessage} = this.props.intl;
         const ariaLabelChannelHeader = Utils.localizeMessage('accessibility.sections.channelHeader', 'channel header region');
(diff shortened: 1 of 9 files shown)


hmhealey (approved):

esethna (changes requested):
Thanks Nevy! Nice find on the sidebar left offset, although I'm still seeing a slight shift:

c186ca1d3762a23b84267555d73c6345

esethna (approved):

devinbinnie (commented):
There's actually a way you can do this without using JavaScript.
You can define a CSS selector that checks to see if `#SidebarContainer` is a sibling, then select `.channel__wrap` and you can go down the CSS tree from there.

Should look something like:
`#SidebarContainer + .channel__wrap .channel-header__popover`

nevyangelova:
@devinbinnie I see the point of doing it without Javascript but I feel in this case is more clear to drive it from the state, because the reason for adding the class is clearer and its grouped together with the other variable.

jgilliam17:
@nevyangelova Looks good
Only one issue
there is a shift to the left in the popover on DM channels, due to status below name. See gif.
Popover shift

nevyangelova:
@hmhealey @devinbinnie I am not super happy with the hacky CSS of the popover and would like to open a follow up ticket to investigate re-doing it and fixing the positioning in a more elegant and consistent way, similar to the dot menu exploration.

devinbinnie:
> @devinbinnie I see the point of doing it without Javascript but I feel in this case is more clear to drive it from the state, because the reason for adding the class is clearer and its grouped together with the other variable.

@nevyangelova I look at it as a separation of concerns issue, where the styling and component logic should be kept separate. I would only use a custom class driven by component logic when CSS breaks down.

I'm only about 2/5 on this, not sure what other people think but I won't hold up the PR.

devinbinnie (approved):

nevyangelova:
I agree with that and I am not happy with the final solution with adding classes, lets explore redesigning the component and styles in the ticket I mentioned above.

jgilliam17 (approved):
Thank you @nevyangelova 

Tested, looks good to merge.
- Verified line breaks, new lines, fixed first line of channel description and popover not shifting on hover.
- Popover text formatting matches edit channel header modal.

PR #5170, opened 2020-03-25
MM-23458 Add missing atBottom condition for showMessageHistoryToast

sudheerDev:
#### Summary
Based on the latest changes from https://github.com/mattermost/mattermost-webapp/pull/5109 we also need a check for `atBottom` for archive toast which was missed in the previous PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23458

### DIFF
--- a/components/toast_wrapper/toast_wrapper.jsx
+++ b/components/toast_wrapper/toast_wrapper.jsx
@@ -78,7 +78,7 @@ class ToastWrapper extends React.PureComponent {
             showUnreadToast = unreadCount > 0 && !props.atBottom;
         }
 
-        if (typeof showMessageHistoryToast === 'undefined' && props.focusedPostId !== '') {
+        if (typeof showMessageHistoryToast === 'undefined' && props.focusedPostId !== '' && props.atBottom !== null) {
             showMessageHistoryToast = props.initScrollOffsetFromBottom > 1000 || !props.atLatestPost;
         }
 
--- a/components/toast_wrapper/toast_wrapper.test.jsx
+++ b/components/toast_wrapper/toast_wrapper.test.jsx
@@ -164,9 +164,12 @@ describe('components/ToastWrapper', () => {
                 ...baseProps,
                 focusedPostId: 'asdasd',
                 atLatestPost: false,
+                atBottom: null
             };
             const wrapper = shallowWithIntl(<ToastWrapper {...props}/>);
+            expect(wrapper.state('showMessageHistoryToast')).toBe(undefined);
 
+            wrapper.setProps({atBottom: false});
             expect(wrapper.state('showMessageHistoryToast')).toBe(true);
         });
 


sudheerDev:
@jgilliam17 Here is the PR i was talking about the issue you pointed out about history toast

jgilliam17:
Thank you @sudheerDev 
I am seeing few issues (testing on Mac)
1. When viewing permalink to first few posts on any channel, history toast is not present on Chrome and Safari (yes on Firefox)
2. `Jump to` on history toast doesn't go to bottom of the channel right away, shows a screen near the bottom then refreshes/reloads to actually show the bottom (white screen on the main view). 
I am not seeing this issue with the new message toast.

sudheerDev:
@jgilliam17 

> 1. When viewing permalink to first few posts on any channel, history toast is not present on Chrome and Safari. 
That is how it is supposed to work based on the spec 
```
Archive toast: When user jumps to an old post/permalink in a read channel
Toast is present only if jumping to a permalink when your not viewing the bottom of the channel + 1000 pixels
```
Link for spec https://docs.google.com/document/d/1lANE8O3jVn4yjHNBbe4tMtXVcl3BYcRLWS0IViQ2Vds/edit#bookmark=id.lyco82r7doa7

I cannot replicate the issue with firefox. Can you try a post above that probably that post is just above 1000px in firefox? 


> 2. Jump to on history toast doesn't go to bottom of the channel right away, shows a screen near the bottom then refreshes/reloads to actually show the bottom (white screen on the main view).
I am not seeing this issue with the new message toast.

This is how it is supposed to work too. 
If the new post and the adjoining posts as well are on client, all we have to do it scroll to it. If we dont have the adjoining posts then we need to fetch from server so we show a loader and then show the posts once API returns it.

in the case of `new messages toast` it did not happen because the new post and the adjoining posts are all in the view so it just scrolled to it.

We can simulate scrolling or loading to show the new set of posts in all three toasts(unread toast, new messages toast, history toast) based on the above logic.

Like for example. 
1. Visit a channel
2. Mark few posts in channel as unread by scrolling a bit(such that unread message toast appears).
3. scroll a bit more and load more posts.
4. Click on a post timestamp to load the permalink view
5. click on the unread toast which says `X new messages since`

Observed: scrolled to the bottom without without refresh.

But instead 
1. Visit a channel
2. scroll a bit more load more posts
3. mark a post from the new set of posts as unread
4. Refresh
5. click on the unread toast which says `X new messages since`

Observed: You see a loader and then followed by new set of posts

On client we do whatever it takes to take you to the post but that means at times we have to call server because we don't have all the posts in case of like looking at permalink of older posts and trying to go to latest post hence the loader.

jgilliam17:
Thank you Sudheer,

> 1. Visit a channel
> 2. Mark few posts in channel as unread by scrolling a bit(such that unread message toast appears).
> 3. scroll a bit more and load more posts.
> 4. Click on a post timestamp to load the permalink view
> 5. click on the unread toast which says `X new messages since`
> 
> Observed: scrolled to the bottom without without refresh.

I see a page load here and it happens after the bottom is already in view. I am not sure this is as intended. 
When I see the bottom I am under the impression my `jump to` bottom is done, but 3 seconds later main view reloads. 

> 1. Visit a channel
> 2. scroll a bit more load more posts
> 3. mark a post from the new set of posts as unread
> 4. Refresh
> 5. click on the unread toast which says `X new messages since`
> 
> Observed: You see a loader and then followed by new set of posts

This works as described, I didn't see an issue with reload of the main view or any lag when jumping to the bottom with new messages.

sudheerDev:
> I see a page load here and it happens after the bottom is already in view. I am not sure this is as intended.
> When I see the bottom I am under the impression my `jump to` bottom is done, but 3 seconds later main view reloads.

Sounds like https://github.com/mattermost/mattermost-webapp/pull/5188/. Can you wait for highlight to  fadeout i.e wait for 6 secs and then try? If waiting fixes that then this issue will be fixed with the other PR. If not we can create a separate ticket tomorrow after these get it.

jgilliam17:
> > I see a page load here and it happens after the bottom is already in view. I am not sure this is as intended.
> > When I see the bottom I am under the impression my `jump to` bottom is done, but 3 seconds later main view reloads.
> 
> Sounds like #5188. Can you wait for highlight to fadeout i.e wait for 6 secs and then try? If waiting fixes that then this issue will be fixed with the other PR. If not we can create a separate ticket tomorrow after these get it.

Thank you for pointing that out Sudheer!
I was clicking on the `jump to` before highlight fade/url change.

PR #5173, opened 2020-03-25
[MM-23589] Fixed the handleRemovedUserEvent to wait for the user to leave the channel before redirecting

devinbinnie:
#### Summary
The websocket action for removing a user from a channel was dispatching the `LEAVE_CHANNEL` action too late, so that `redirectUserToDefaultTeam` would redirect the user back to the same channel they just left.

This PR fixes that behaviour and adds tests for the new and old sidebar (since the old sidebar handled this by calling `redirectUserToDefaultTeam` after the state had updated)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23589

### DIFF
--- a/e2e/cypress/integration/websocket/handle_removed_user_spec.js
+++ b/e2e/cypress/integration/websocket/handle_removed_user_spec.js
@@ -42,9 +42,10 @@ describe('Handle removed user - old sidebar', () => {
             const userId = res.body.id;
             return cy.removeUserFromChannel(channelId, userId);
         }).then(() => {
-            // * Verify that the channel changed back to Town Square
+            // * Verify that the channel changed back to Town Square and that Off-Topic has been removed
             cy.url().should('include', `/${teamName}/channels/town-square`);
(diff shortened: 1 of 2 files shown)


e2e/cypress/integration/websocket/handle_removed_user_spec.js
```diff
@@ -0,0 +1,94 @@
...
+        cy.getCurrentChannelId().then((res) => {
+            channelId = res;
+            return cy.apiGetMe();
+        }).then((res) => {
+            const userId = res.body.id;
+            return cy.removeUserFromChannel(channelId, userId);
+        }).then(() => {
+            // * Verify that the channel changed back to Town Square
+            cy.url().should('include', `/${teamName}/channels/town-square`);
+            cy.get('#channelHeaderTitle').should('be.visible').should('contain', 'Town Square');
```
saturninoabril: Can we add another assertion here to check that Off-Topic is removed from the LHS? 
```
cy.get('.SidebarChannel:contains(Off-Topic)').should('not.exist');
```

Same below.

jwilander:
Nice test 👍

PR #5175, opened 2020-03-25
MM-23577 Update More Unread Indicators after animations

hmhealey:
@devinbinnie @bradjcoughlin Let me know what you think about the 100ms debounce time for this. We could probably drop it since all of the `onTransitionEnd` events from each child should come in faster than that, but I default to 100ms for this sort of thing.

I also haven't figured out a way to add tests for this since I haven't been able to reproduce it, but if anyone has any suggestions, I'd appreciate them.

@srkgupta I haven't been able to reproduce the exact issue as you're seeing it, but I'm not surprised there's some edge cases around the showing and hiding of those indicators. I've added some extra logic to update them after an animation happens, so hopefully this should reduce the chances that we run into an edge case like that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23577

### DIFF
--- a/components/sidebar/sidebar_category_list/__snapshots__/sidebar_category_list.test.tsx.snap
+++ b/components/sidebar/sidebar_category_list/__snapshots__/sidebar_category_list.test.tsx.snap
@@ -5,6 +5,7 @@ exports[`components/sidebar/sidebar_category_list should match snapshot 1`] = `
   className="SidebarNavContainer a11y__region"
   data-a11y-sort-order="7"
   id="sidebar-left"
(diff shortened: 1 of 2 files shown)


hmhealey:
@srkgupta I've updated the repro steps in the ticket with some steps to guarantee that the More Unreads indicator appears at the bottom of the sidebar. I assume that's the same issue even though you saw the indicator at the top that wouldn't go away.

srkgupta:
Thanks @hmhealey. Waiting for Dev Reviews to complete to begin QA.

PR #5199, opened 2020-03-26
[MM-20550] Convert email notification setting to typescript

hiendinhngoc:
#### Summary

Migrate 'components/user_settings/notifications/email_notification_setting' module and associated tests to TypeScript 

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/13691
  JIRA: https://mattermost.atlassian.net/browse/MM-20550


hiendinhngoc:
I temporarily set `enableEmailBatching`  to number type because `getEmailInterval` requires number type at the second argument. Please suggest the correct way to deal with `enableEmailBatching`

hanzei:
NVM @agnivade, GH played tricks on me

devinbinnie:
@hiendinhngoc Looks like the type on `getEmailInterval` is wrong, the parameter `enableEmailBatching` should be a boolean. I can make a fix for that, work under the assumption that it's a boolean. Sorry about that.

devinbinnie:
https://github.com/mattermost/mattermost-redux/pull/1102

devinbinnie:
@hiendinhngoc The fix has been merged, and I've updated the branch for you.
There are some tests failing, which I've indicated below. Please make sure to `git pull` on your branch before making any changes.

```
Summary of all failing tests
 FAIL  components/user_settings/notifications/email_notification_setting/email_notification_setting.test.tsx
  ● components/user_settings/notifications/EmailNotificationSetting › should match snapshot

    expect(received).toMatchSnapshot()

    Snapshot name: `components/user_settings/notifications/EmailNotificationSetting should match snapshot 1`

    - Snapshot
    + Received

    @@ -6,11 +6,11 @@
        }
        activeSection="email"
        currentUserId="current_user_id"
        emailInterval={0}
        enableEmail={false}
    -   enableEmailBatching={0}
    +   enableEmailBatching={false}
        onCancel={[MockFunction]}
        onChange={[MockFunction]}
        onSubmit={[MockFunction]}
        saving={false}
        sendEmailNotifications={true}

      32 |         const wrapper = mountWithIntl(<EmailNotificationSetting {...requiredProps}/>);
      33 | 
    > 34 |         expect(wrapper).toMatchSnapshot();
         |                         ^
      35 |         expect(wrapper.find('#emailNotificationImmediately').exists()).toBe(true);
      36 |         expect(wrapper.find('#emailNotificationNever').exists()).toBe(true);
      37 |         expect(wrapper.find('#emailNotificationMinutes').exists()).toBe(false);

      at Object.test (components/user_settings/notifications/email_notification_setting/email_notification_setting.test.tsx:34:25)

  ● components/user_settings/notifications/EmailNotificationSetting › should match snapshot, enabled email batching

    expect(received).toMatchSnapshot()

    Snapshot name: `components/user_settings/notifications/EmailNotificationSetting should match snapshot, enabled email batching 1`

    - Snapshot
    + Received

    @@ -6,11 +6,11 @@
        }
        activeSection="email"
        currentUserId="current_user_id"
        emailInterval={0}
        enableEmail={false}
    -   enableEmailBatching={1}
    +   enableEmailBatching={true}
        onCancel={[MockFunction]}
        onChange={[MockFunction]}
        onSubmit={[MockFunction]}
        saving={false}
        sendEmailNotifications={true}

      46 |         const wrapper = mountWithIntl(<EmailNotificationSetting {...props}/>);
      47 | 
    > 48 |         expect(wrapper).toMatchSnapshot();
         |                         ^
      49 |         expect(wrapper.find('#emailNotificationMinutes').exists()).toBe(true);
      50 |         expect(wrapper.find('#emailNotificationHour').exists()).toBe(true);
      51 |     });

      at Object.test (components/user_settings/notifications/email_notification_setting/email_notification_setting.test.tsx:48:25)


Snapshot Summary
 › 2 snapshots failed from 1 test suite. Inspect your code changes or run `npm run test-ci -- -u` to update them.
```

hiendinhngoc:
@devinbinnie I updated, pls check

components/user_settings/notifications/email_notification_setting/email_notification_setting.test.tsx
```diff
@@ -168,6 +168,7 @@ describe('components/user_settings/notifications/EmailNotificationSetting', () =
 
     test('should derived state from props', () => {
         const nextProps = {
+            enableEmail: true,
```
larkox, line 171: Here you are changing the behavior of the test (which may explain why you had to change INTERVAL_NEVER to INTERVAL_INMEDIATE). When we left it as undefined, it was used as false. Later in line 187 was set to true to test a different thing.
TL;DR: here enableEmail should be false.

components/user_settings/notifications/email_notification_setting/email_notification_setting.tsx
```diff
@@ -14,27 +13,37 @@ import SettingItemMin from 'components/setting_item_min';
...
+    onCancel: () => void;
+    onChange: (enableEmail: boolean) => void;
+    serverError?: string;
+    saving?: boolean;
+    focused?: boolean;
+    sendEmailNotifications: boolean;
+    enableEmailBatching: boolean;
+    actions: {
+        savePreferences: (currentUserId: string, emailIntervalPreference: {}) =>
+        Promise<{}>;
```
larkox, line 32: Maybe is too verbose here, since we do not use the value, but if we wanted to be strict, the promise result is also not {}, but {data: bool} (if I am not mistaken).

components/user_settings/notifications/email_notification_setting/email_notification_setting.test.tsx
```diff
@@ -184,8 +185,8 @@ describe('components/user_settings/notifications/EmailNotificationSetting', () =
         expect(wrapper.state('sendEmailNotifications')).toBe(nextProps.sendEmailNotifications);
         expect(wrapper.state('newInterval')).toBe(Preferences.INTERVAL_NEVER);
 
-        nextProps.enableEmail = true;
         nextProps.emailInterval = Preferences.INTERVAL_FIFTEEN_MINUTES;
+        nextProps.enableEmail = true;
         wrapper.setProps(nextProps);
```
larkox, line 190: Nit: Here you have change the order of the lines. Not a real problem, but if you can put it in the original order, it will make the changes log clearer.

components/user_settings/notifications/email_notification_setting/index.ts
```diff
@@ -11,10 +11,17 @@ import {savePreferences} from 'mattermost-redux/actions/preferences';
...
 import {getConfig} from 'mattermost-redux/selectors/entities/general';
 import {get as getPreference} from 'mattermost-redux/selectors/entities/preferences';
+import {GlobalState} from 'mattermost-redux/types/store';
+import {ActionFunc} from 'mattermost-redux/types/actions';
 
 import EmailNotificationSetting from './email_notification_setting';
 
-function mapStateToProps(state) {
+type Actions = {
+    savePreferences: (currentUserId: string, emailIntervalPreference: {}) =>
```
larkox, line 20: You changed the type of emailIntervalPreference in the other file, but not in this one. Is it a mistake, or there is any reason to leave it like this?

hiendinhngoc: That's my mistake, gonna fix it now.

hiendinhngoc: That's my mistake, gonna fix it now.

DHaussermann:
HI @hiendinhngoc, on the spin-up server above, I'm seeing that the account setting for email notification does not seem to update. It starts out set to `Immediately` and cannot be updated. 
EmailSettings

Not sure if this could be related to your change or if there could be something else going on. I'm currently unable to repro this on master.

stylianosrigas:
I am removing the Setup Cloud Test Server label because we need to deprecate the cluster the installation is running. Please feel free to add it again.

DHaussermann:
Hi @hiendinhngoc, can you confirm if you're able to reproduce the issue shown above?

hiendinhngoc:
> Hi @hiendinhngoc, can you confirm if you're able to reproduce the issue shown above?

@DHaussermann 
Hi, I've run the code in my local. I didn't get this issue. I can change to `never` normally.
Peek 2020-07-28 23-39

hiendinhngoc:
@DHaussermann hi, is there anything you need me to do to complete  the PR

DHaussermann:
Hi @hiendinhngoc thanks for the ping. I meant to retest this after pulling in master. 

I'm still seeing the same issue as before where setting does not appear to save. This can be reproduced on the server above ☝️ 

Do you believe it could be related to this PR?

hiendinhngoc:
> Hi @hiendinhngoc thanks for the ping. I meant to retest this after pulling in master.
> 
> I'm still seeing the same issue as before where setting does not appear to save. This can be reproduced on the server above point_up
> 
> Do you believe it could be related to this PR?

@DHaussermann hi, honestly I'm not sure because when I ran the code on my local previously, it worked normally. I will try to update the newest code from master to this branch and check again.

hiendinhngoc:
@DHaussermann hi, I've run the code on my local(master branch which not including my change in this PR) and I got the same issue as mentioned(cannot update from `Immediately` do `Never`). So, I guess the problem comes from the server-side. However, I don't have the experience to debug server. Do you have any suggestions?

DHaussermann:
@hiendinhngoc If it reproduces on master, then I will create a separate issue to address it. But that wasn't what I saw when I looked. Sorry let me check again.

DHaussermann:
Hi @hiendinhngoc,

From what I can see this does not repro for me on master. I can't think of any config that would prevent me from saving 'never'. I can't reproduce this on my local when set to master.  

I also spun up this sever:  https://dkh-master.test.mattermost.cloud  (same admin credentials).  The email and SMTP config looks the same and this is only 2 server commits ahead of the PR server. Given that the issue does not repro here, is there a way you can confirm for sure it's not the webapp code making the difference.

DHaussermann:
@devinbinnie I'm still seeing the issue on the updated server after the last merge. Any thoughts on if we can confirm the issue is related to this PR?

hiendinhngoc:
> @devinbinnie I'm still seeing the issue on the updated server after the last merge. Any thoughts on if we can confirm the issue is related to this PR?

@DHaussermann @devinbinnie FYI, as I noticed by checking [the link](https://mattermost-webapp-pr-5199.test.mattermost.cloud), we can change from `immediately` to `never` if at the first time we keep it `immediately` and save, then at the second time change to `never` and save. I'm currently investigating.

DHaussermann:
@hiendinhngoc  thanks for narrowing this down. I also see the same behavior. The option can be toggled if save it as `immediately` first. With this additional info, I have gone back and confirmed that I do not see the behavior on Master. Hopefully this helps narrow it down.

components/user_settings/notifications/email_notification_setting/email_notification_setting.tsx
```diff
@@ -94,12 +103,12 @@ export default class EmailNotificationSetting extends React.PureComponent {
         return null;
     }
 
-    handleChange = (e) => {
-        const enableEmail = e.currentTarget.getAttribute('data-enable-email');
+    handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        const enableEmail = JSON.parse(e.currentTarget.getAttribute('data-enable-email')!);
```
devinbinnie, line 107: @hiendinhngoc is it possible that adding this `JSON.parse` statement could be causing the error seen by @DHaussermann? I checked the PR SpinWick and can also reproduce it there.

stylianosrigas:

components/user_settings/notifications/email_notification_setting/email_notification_setting.tsx
hiendinhngoc: @devinbinnie I've checked `enableEmail` variable and it returns the right value base on user selection. So, I don't think this is the cause.

devinbinnie: @hiendinhngoc See this line: https://github.com/mattermost/mattermost-webapp/blob/5da1daa84bdbf58eec4c7872091d6ead01a7d457/components/user_settings/notifications/user_settings_notifications.jsx#L942

The parent component is expecting that `enableEmail` is a string, so even when it is set to `true`, it would evaluate to `false` due to the explicit type checking. When `handleChange` is called, it makes a call to a parent method `onChange` that sets the state of the parent component to the pure boolean value, which we don't want. We'll want to change this back to use a `string` for now until we migrate the parent component at least.

hiendinhngoc: @devinbinnie thank you so much for your suggestion, I'm going to update it.

hiendinhngoc:
@DHaussermann hi, I updated the code to fix the issue, please check it when you have time, thanks.

PR #4499, opened 2019-12-11
MM-20939 - File picker opens twice when trying to attach a file through the paperclip button in mobile view

ali-farooq0:
#### Summary
Attaching only 1 click handler when in mobile view or not in mobile view.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-20939

PR #5211, opened 2020-03-27
MM-23402 Fix for Mobile Web: Attach button only works with double tap on iOS

sudheerDev:
#### Summary
This was a regression was caused when we added conditional checks for mobile for preventing opening of file picket twice when clicked on attach button https://github.com/mattermost/mattermost-webapp/pull/4499.

The `isMobileApp` util func used here is not all that appropriate as it skips considering mobile browsers, i think it should have been `isMobile` util func(if we need a similar check In future we probably should check if touch listeners are available instead for reliability). 
Fixing this did not help as i think we need `onClick` to get around some of the security concerns in browsers with user initiated actions and trying to open file pickers programmatically. 

I resorted to passing `simulateClick` event to both `onClick` and `onTouchEnd` events and stop propagation as well as defaults to prevent them from opening the file picker twice. If touch events exist then it is fired first and the checks prevent click method from executing solving the root issue too.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23402


components/file_upload/file_upload.jsx
```diff
@@ -565,15 +565,16 @@ class FileUpload extends PureComponent {
         this.setState({menuOpen: false});
     }
 
-    simulateInputClick = () => {
+    simulateInputClick = (e) => {
+        e.preventDefault();
+        e.stopPropagation();
```
ali-farooq0: i may be wrong but i remember looking into this at one point and attempting something similar but it still opened up the file picker twice for me. i.e. was seeing https://mattermost.atlassian.net/browse/MM-20939 behaviour. 

@sudheerDev can you please confirm if you're not able to reproduce the issue mentioned in MM-20939 with this change?

ali-farooq0: using chrome/ff

sudheerDev: yes, i am positive it does not cause a regression with the ticket you mentioned.

sudheerDev: Ahh, it does not cause that issue but it has not fixed the issue of picker not opening without double click. It worked lcaoclly for me. Gotto check this again

ali-farooq0: 👍

sudheerDev: @ali-farooq0 The PR works. I don't know what i was testing earlier. It fixes the ticket i linked and it does not cause a regression with the ticket you linked.

sudheerDev:
waiting to see if this is needed for 5.22v

PR #5217, opened 2020-03-27
Experimental Sidebar Drag 'n Drop - Proof of Concept with react-beautiful-dnd

devinbinnie:
#### Summary

This is a proof-of-concept for the drag 'n drop implementation for the Experimental Channel Sidebar using `react-beautiful-dnd`.
Please except things not to work, but this PR is up so that we can quickly iterate over design issues and keep everyone in the loop w.r.t. the progress of the implementation.

#### Ticket Link
No ticket yet.


esethna:
@devinbinnie looks like build is failing

esethna:
@devinbinnie FYI this doesn't work on desktop ^ Probably the same issue Nevy fixed with the drag and drop of teams https://mattermost.atlassian.net/browse/MM-23448

devinbinnie:
@esethna Desktop dragging issue should be fixed

devinbinnie:
@esethna @andrewbrown00 I've updated the PoC to allow for persistent dragging (ie. when you drag something, it should stay there now).

Some notes:
- The drag will only persist in the current browser session, once you refresh the changes will be lost
- The current mock does not account for external changes to the sidebar (ie. creating/leaving/favouriting channels does not work correctly and will crash the app). This is only temporary to allow you to play with the dragging animations and feel, since we have not filled out the functionality in `mattermost-redux` yet.

Feel free to continue playing with this PoC, and let me know if there are changes you would like me to make. For the time being, i'll take off Work in Progress as I believe I have covered most of the functional requirements.

esethna:
@devinbinnie let's schedule a review call to walk through the functional notes and comment on UI/UX improvements to make sure we're on the right track with react dnd

esethna:
Nice work!

actions/views/channel_sidebar.ts
```diff
@@ -28,3 +31,70 @@ export function setUnreadFilterEnabled(enabled: boolean) {
         enabled,
     };
 }
+
+export function setCategoryOrder(teamId: string, categoryId: string, channelId: string, newIndex: number) {
+    const getChannelsForCategory = makeGetChannelsForCategory();
```
hmhealey, line 36: This isn't strictly necessary to do here because the memoization when it's called less frequently, but it's totally fine to leave this way.

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -98,12 +103,26 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
...
+
+    getStyle = (style: DraggingStyle | NotDraggingStyle | undefined, snapshot: DraggableStateSnapshot) => {
+        if (this.props.isDMCategory && !snapshot.isDragging) {
+            return {
+                ...style,
+                transform: 'none',
+            };
+        }
+
+        return {...style};
```
hmhealey, line 121: Could we just return `style` here without creating a new object?

hmhealey:
One other thing that might be weird is having separate actions for adding a channel to a new category and removing it from the old one. It's possible we could get into a weird state where a channel appears in multiple categories at once. We can always change that when the other infrastructure is available though.

esethna:
Sweet, CSS is coming along nicely! No need to make any changes now, but things I want to take over with Andrew: 

1. Playing with it I think we should rethink this effect. Perhaps we can use the same as the DM category drop target for consistency.

image

2. Also wondering if we should consider using the sidebar text color instead of the online color for the drop zone. 

3. channels and categories are too dimmed I think right now, not enough contrast

4. I think we should rethink the expanding of collapsed categories when a read channel is dropped in them. It's pretty jarring right now, and testing slack they have channels disappear if dropped in a collapsed category and it doesn't feel off. Especially if we had an undo toast that appears (Phase 3)

devinbinnie:
@esethna @andrewbrown00 Should be ready to review the look and feel including CSS at this point.

esethna:
@devinbinnie 

1. remove highlight/gradient effect on the category header
- Add the DM drop target effect on the category headers itself when dropping a channel directly on a sidebar header
- Use sidebar text color for the drop target outline at 60% opacity
- Team sidebar should be under the dragables
- Hover effect overlay should remain on the draggable after you pick it up so it doesn't change colors when picked up
- GM icon does not reduce in opacity when a channel is being dragged
- Do not expand category on dropping channels
- Allow channels to be dropped into collapsed categories and maintain order as per: https://docs.google.com/document/d/17Uw_HMptsXADgganXffjB1k8_WuAn5nyOU5JrqZj_AA/edit#bookmark=id.cfz6w49d3x67
- Check that alpha sorting is disabled in all categories after first drag (there seems to seom alpha sorting observed on teh spinmint)
- When hovering over a category header name should show the pointer cursor
- QA: Test on a low end machine, chrome and firefox
- Fix alignment and usability on mobile web view
- Increase opacity on "x" to remove DM to 
- Add an exception in the keyboard accessibility controller when a channel is picked up so drag/drop doesn't conflict with keyboard shortcuts
- Tabbing through channels in collapsed categories should only go to visible channels
- Tabbing on a category header should only have a single focus effect (currently there are two tabs required)

devinbinnie:
> * [ ]  Allow channels to be dropped into collapsed categories and maintain order as per: https://docs.google.com/document/d/17Uw_HMptsXADgganXffjB1k8_WuAn5nyOU5JrqZj_AA/edit#bookmark=id.cfz6w49d3x67
> * [ ]  Increase opacity on "x" to remove DM to
> * [ ]  Add an exception in the keyboard accessibility controller when a channel is picked up so drag/drop doesn't conflict with keyboard shortcuts
> * [ ]  Tabbing through channels in collapsed categories should only go to visible channels
> * [ ]  Tabbing on a category header should only have a single focus effect (currently there are two tabs required)

These should be separate tickets, since they're all Phase 1 issues.
@esethna Can you make bugs for these? We can probably just have the team pick them up as needed.

esethna:
@devinbinnie 

1. Category outline should only appear when directly on top of the category header

esethna:
@devinbinnie the mobile web view alignment issues are only introcuded on this PR

esethna:
@devinbinnie followup tickets:

https://mattermost.atlassian.net/browse/MM-24204
https://mattermost.atlassian.net/browse/MM-24205
https://mattermost.atlassian.net/browse/MM-24164
https://mattermost.atlassian.net/browse/MM-24206

But we should fix mobile web view in this PR probably, since the issue is only introduced here?

devinbinnie:
@esethna I can reproduce the issues on mobile view on `community-daily` as well.

esethna:
@devinbinnie ticket created, however I don't see the issues as severely as in this PR on my daily instance: https://mattermost.atlassian.net/browse/MM-24323

My daily (definitely not perfect, but better than what I see in this PR): 
image

actions/views/channel_sidebar.ts
```diff
@@ -28,3 +34,91 @@ export function setUnreadFilterEnabled(enabled: boolean) {
...
+        }
+
+        // TODO: Will need to call an actual redux action so that this also goes to the server
+        // But for now, this will update the redux state correctly.
+        dispatch({type: ChannelCategoryTypes.RECEIVED_CATEGORY, data: {...category, channel_ids: channelIds}});
+    };
+}
+
+export function setDraggingState(data: DraggingState) {
+    return (dispatch: DispatchFunc) => {
```
hmhealey, line 112: This can just return the action instead of explicitly dispatching it

actions/views/channel_sidebar.ts
```diff
@@ -28,3 +34,91 @@ export function setUnreadFilterEnabled(enabled: boolean) {
...
+        dispatch({
+            type: ActionTypes.SIDEBAR_DRAGGING_SET_STATE,
+            data,
+        });
+    };
+}
+
+export function stopDragging() {
+    return (dispatch: DispatchFunc) => {
```
hmhealey, line 121: Same here

components/sidebar/sidebar_category_list/sidebar_category_list.tsx
```diff
@@ -216,6 +230,10 @@ export default class SidebarCategoryList extends React.PureComponent<Props, Stat
     }
 
     updateUnreadIndicators = () => {
+        if (this.props.draggingState.state) {
```
hmhealey, line 233: It might make sense to hide the unread indicators when dragging in addition to not recalculating it.

components/sidebar/sidebar_category_list/sidebar_category_list.tsx
```diff
@@ -330,6 +349,62 @@ export default class SidebarCategoryList extends React.PureComponent<Props, Stat
         this.updateUnreadIndicators();
     }, 100);
 
+    onBeforeCapture = (before: BeforeCapture) => {
+        // // Ensure no channels are animating
+        this.channelRefs.forEach((ref) => ref.classList.remove('animating'));
+
+        // TODO: Refactor into better code
+        // Turn off scrolling temporarily so that dimensions can be captured
+        const droppable = [...document.querySelectorAll<HTMLDivElement>('[data-rbd-droppable-id*="droppable-categories"]')];
```
hmhealey, line 358: We'll definitely want automated tests for this since it seems like we're reaching into the library's internals here

components/sidebar/sidebar_category_list/sidebar_category_list.tsx
```diff
@@ -330,6 +349,62 @@ export default class SidebarCategoryList extends React.PureComponent<Props, Stat
         this.updateUnreadIndicators();
     }, 100);
 
+    onBeforeCapture = (before: BeforeCapture) => {
+        // // Ensure no channels are animating
+        this.channelRefs.forEach((ref) => ref.classList.remove('animating'));
+
+        // TODO: Refactor into better code
```
hmhealey, line 356: Do you have concrete plans for refactoring this, or can we remove this comment? We shouldn't be leaving TODOs around without an accompanying ticket.

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -75,11 +87,11 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
     }
 
     isCollapsed = (props: Props) => {
-        return props.isCategoryCollapsed && !this.isUnread() && !props.isCurrentChannel;
+        return /*props.isDropDisabled || */props.isCategoryDragged || (props.isCategoryCollapsed && !this.isUnread() && !props.isCurrentChannel);
```
hmhealey, line 90: Remember to remove the commented out code once we've got the correct behaviour here.

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -98,12 +110,34 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
...
-    setRef = (ref: HTMLLIElement) => {
-        this.props.setChannelRef(this.props.channel.id, ref);
+    setRef = (refMethod: (element: HTMLLIElement) => any) => {
+        return (ref: HTMLLIElement) => {
+            this.props.setChannelRef(this.props.channel.id, ref);
+            refMethod(ref);
+        };
+    }
+
+    getStyle = (style: DraggingStyle | NotDraggingStyle | undefined, snapshot: DraggableStateSnapshot) => {
```
hmhealey, line 120: What do these different styles do? Some comments would be helpful here.

components/sidebar/sidebar_category_list/sidebar_category_list.tsx
```diff
@@ -330,6 +349,62 @@ export default class SidebarCategoryList extends React.PureComponent<Props, Stat
...
+        const droppable = [...document.querySelectorAll<HTMLDivElement>('[data-rbd-droppable-id*="droppable-categories"]')];
+        droppable[0].style.height = `${droppable[0].scrollHeight}px`;
+
+        const draggingState: DraggingState = {
+            state: 'capture',
+            id: before.draggableId,
+        };
+
+        if (this.props.categories.some((category) => category.id === before.draggableId)) {
+            draggingState.type = 'category';
```
hmhealey, line 367: We should probably have constants to track these values between components

components/sidebar/sidebar_category_list/sidebar_category_list.tsx
devinbinnie: Agreed, made a ticket: https://mattermost.atlassian.net/browse/MM-24439

components/sidebar/sidebar_category_list/sidebar_category_list.tsx
devinbinnie: I don't, it's just not code I'm happy with. I'd like to look into a better way to control the scrolling, any suggestions?

esethna:
@devinbinnie 

1. I think we discussed that the category should only show the outline when dragging directly on top of it:
https://i.gyazo.com/c9ad666a52f6a7171962dd8e868c5419.gif

2. DM category outline when collapsed is too tall:
image

-------

@andrewbrown00 can you help do a final review so we can move this PR forward? A few thoughts:

1. The decreased opacity state when a category can't be dragged into is slightly too faint imo
image

2. When the categories get decreased/increased opacity as above, should we have some sort of fade in/out effect ^ 0/5

devinbinnie:
> @devinbinnie
> 
> 1. I think we discussed that the category header should only show the outline when dragging directly on top of it:
>    https://i.gyazo.com/c9ad666a52f6a7171962dd8e868c5419.gif

This is fixed now.

> 2. DM category outline when collapsed is too tall:
>    image

This is happening because the outline wraps around the entire category, not just the header. We can look into fixing this but I'd like to push back until after feature complete, since it's just a small visual issue but will require some thought to fix.

> 3. Dragging a category or channel does not show the closed hand cursor during drag

I can't reproduce this. Tried on Mac and Windows and the closed hand shows up for me on both.

> 
> @andrewbrown00 can you help do a final review so we can move this PR forward? A few thoughts:
> 
> 1. The decreased opacity state when a category can't be dragged into is slightly too faint imo
>    image
> 2. When the categories get decreased/increased opacity as above, should we have some sort of fade in/out effect ^ 0/5 transition feels jarring right now
> 
> https://i.gyazo.com/63cdbb5613deae0a7e0304d03dbe7094.gif

devinbinnie:
> This is amazing @devinbinnie 🎉
> 
> **Few items to resolve:**
> 
> * [x]  The opacity on the dimmed items while dragging should be `40%`, it looks like it's `8%`
>   image

Fixed!

> * [x]  Can we also dim the selected channel (and DM/GMs) state too, only the channel name is dimmed
>   image

Fixed!

> * [x]  Category header should maintain the hover state style while being dragged, this will avoid a color flash on the header text when going from the hover to the drag state. Same way we maintain the last state for the channels.
>   image

Fixed!

> * [ ]  Can we increase the timing when Direct or Group message shows up in the DM category after dropping? It seems to have a lag, instead of being as smooth as dropping channels. **Note** sometimes the lag happens when dragging channels too.
>   

You'll have to be more specific. Some of the time taken there is just the time it takes for the redux state to update. This should hopefully speed up a bit once we removed the bad mocked code.
The only thing I have control of here is the transition time for the dropped channel to make it into place. I can adjust that if you'd like, what should I adjust it to?

> * [ ]  Can we remove the flash effect after dropping a channel and the hover effect flashes in-and-out?
>   

What you're seeing is the style of the button change from when it has been dropped back to its default state. Since I assume they disable all hover events while dragging, you're seeing the flash because the hover gets turned back on because your cursor is over top of the channel. I don't see an easy way around this since I can't tap into that state it's in after dropping, and I don't think this is something that will be a huge issue since it depends on you leaving your cursor on top of the dragged item after it's dropped.

Can we look into this one outside of this PR?

> * [ ]  Can we add an animation when bringing dimmed items back to full opacity, I agree with Eric it feels jarring when they flash back in. It's most noticeable for DMs.
>

deanwhillier:
I noticed this with the sticky category headers, not sure if it's already on the list but thought I'd add here in case.

image

devinbinnie:
> I noticed this with the sticky category headers, not sure if it's already on the list but thought I'd add here in case.
> 
> image

When are you seeing this? I haven't seen this before

deanwhillier:
> When are you seeing this? I haven't seen this before

I'm seeing on the test server right now ... after a fresh load having done nothing.

devinbinnie:
> > When are you seeing this? I haven't seen this before
> 
> I'm seeing on the test server right now ... after a fresh load having done nothing.

Oh! I know what I did :p, will look into fixing.

devinbinnie:
> > > When are you seeing this? I haven't seen this before
> > 
> > 
> > I'm seeing on the test server right now ... after a fresh load having done nothing.
> 
> Oh! I know what I did :p, will look into fixing.

@deanwhillier Okay this is fixed. However, this will still happen if you have part of the sidebar focused using the `a11y_controller` and you start scrolling using the mouse. Personally I think this is a case that won't really be encountered (since if you're using the keyboard to navigate you won't be scrolling in the same sense) so I think this is fine. We can also look into addressing it later.

devinbinnie:
> Awesome work @devinbinnie, I'm loving the drag-n-drop 🎉
> 
> **One item to resolve**
> 
> * [ ]  Channels are shifting too early and sometimes oddly when a channel gets near or over the **Category headers**
>   
> 

Well this one's weird, is it alright if we create a follow-up ticket for this one as well? It doesn't seem to be easily solvable. For some reason 'Favorites' seemed to be allowing for dragging into it via placeholder way before the other categories. Not entirely sure why...

components/sidebar/sidebar_category_list/sidebar_category_list.tsx
hmhealey: No, but we shouldn't be leaving comments around for stuff that we don't have anything actionable for improving or else we'll just be cluttering the code.

devinbinnie: I'll remove the TODO then and we can revisit later on.

PR #5220, opened 2020-03-27
[MM-23638] Prevent Safari from handling team number keyboard shortcuts

deanwhillier:
**Summary**

Safari favorite shortcuts share the keyboard shortcut's MM uses for switching between teams (i.e. CMD+OPTION+#). When using the active team's shortcut, the keyboard event handler was not calling `event.preventDefault()` and so Safari was navigating away from MM to the corresponding favourite shortcut. This PR ensures that the event handler always calls `event.preventDefault()` while still 'ignoring' the shortcut for the active team.

**Ticket Link**

https://mattermost.atlassian.net/browse/MM-23638


amyblais:
@deanwhillier @esethna Is this a new regression in v5.22?

esethna:
@amyblais no it's a bug with the new team switcher shortcuts

PR #5221, opened 2020-03-29
show UserProfile popover instead of performing search in discontinuing post

rakhi2104:
#### Summary
In `discontinuing` post type, show UserProfile of Post's parent user instead of performing search of Post's parent user posts in a channel

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-server/issues/14137

Jira ticket: https://mattermost.atlassian.net/browse/MM-23480

#### Screenshot
mm-23480


components/user_profile/user_profile.jsx
```diff
@@ -22,6 +22,7 @@ export default class UserProfile extends PureComponent {
         overwriteName: PropTypes.node,
         user: PropTypes.object,
         userId: PropTypes.string,
+        showGuestBadge: PropTypes.bool,
```
Willyfrog: What does this have to do with the current issue?

rakhi2104: @Willyfrog the User profile component by default has the `Guest` label/tag. So to avoid it in the post comment, this change was brought in. 
(if this tag should be carried forward, we have to play with css and we can remove this piece of code.)

Screenshot from 2020-03-30 20-20-12

Willyfrog: but it should only have the tag if the account is a guest one, right? Taking a look at the jira ticket, I don't see any notes asking for removal.

rakhi2104: Thanks @Willyfrog. I'll update it soon.

components/post_view/commented_on/commented_on.jsx
```diff
@@ -59,14 +54,18 @@ export default class CommentedOn extends PureComponent {
     render() {
         const username = this.makeUsername();
         const message = this.makeCommentedOnMessage();
+        const parentPostUser = this.props.parentPostUser;
+        const parentPostUserId = parentPostUser && parentPostUser.id && parentPostUser.id;
```
nevyangelova, line 58: Third branch is redundant :D

components/user_profile/user_profile.jsx
rakhi2104: Addressed in latest commit.

components/post_view/commented_on/commented_on.jsx
rakhi2104: Addressed in latest commit.

components/post_view/commented_on/commented_on.jsx
```diff
@@ -59,14 +54,18 @@ export default class CommentedOn extends PureComponent {
...
-                className='theme'
-                onClick={this.handleOnClick}
-            >
-                {username}
-            </a>
+        const parentUserProfile = (
+            <UserProfile
+                user={parentPostUser}
+                userId={parentPostUserId}
+                displayName={username}
```
Willyfrog, line 64: wouldn't this username be related to the current post instead of the parent's?
also, looks like this one is being provided by [mapping the state](https://github.com/mattermost/mattermost-webapp/blob/master/components/user_profile/index.js#L15), maybe you wanted to set `displayUserName` to true so it uses the name from the `user` object?

components/post_view/commented_on/commented_on.jsx
```diff
@@ -59,14 +54,18 @@ export default class CommentedOn extends PureComponent {
...
+                hasMention={true}
+                disablePopover={false}
+                showGuestBadge={false}
```
Willyfrog, line 67: what happens if the parent post is created by a guest account? shouldn't we show it?

rakesh-sureify: This was leftover from cleaning done earlier. I'll update the snaps also

rakhi2104: This `showGuestBadge` flag was leftover from past cleaning. I'll update in the snaps too.

components/post_view/commented_on/commented_on.jsx
jasonblais: @rakhi2104 Do you have thoughts on Guillermo's questions above? We'd be glad to help!

Willyfrog: @rakhi2104 let me know if you want to discuss this either here or in the community server :)

Willyfrog: from community's conversation: 
> The CommentedOn's makeUsername generates the Original post's username

> The User object will be fetched asynchronously, and will not have username at the point when rendered, so it should be passed explicitly.

Willyfrog (approved):
after talking about this on community server, lgtm

thanks for taking the time to create the PR and answering my questions :D

lindy65 (approved):
Thanks @rakhi2104 - looks good to me - Clicking on the username opens the profile popover. 👍

PR #5223, opened 2020-03-30
[MM-18992] Migrate 'components/status_icon.jsx' to TypeScript

haydenhw:
#### Summary
Migrate 'components/status_icon.jsx' to TypeScript

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/12437

### DIFF
--- a/components/status_icon.tsx
+++ b/components/status_icon.tsx
@@ -1,7 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import PropTypes from 'prop-types';
 import React from 'react';
 
 import StatusAwayAvatarIcon from 'components/widgets/icons/status_away_avatar_icon';
@@ -13,14 +12,14 @@ import StatusOfflineIcon from 'components/widgets/icons/status_offline_icon';
 import StatusOnlineAvatarIcon from 'components/widgets/icons/status_online_avatar_icon';
 import StatusOnlineIcon from 'components/widgets/icons/status_online_icon';
 
-export default class StatusIcon extends React.PureComponent {
-    static propTypes = {
-        button: PropTypes.bool,
-        status: PropTypes.string,
-        className: PropTypes.string,
-        type: PropTypes.string,
-    };
+type Props = {
+    button?: boolean;
+    status?: string;
+    className?: string;
+    type?: string;
+}
 
+export default class StatusIcon extends React.PureComponent<Props> {
     static defaultProps = {
         className: '',
         button: false,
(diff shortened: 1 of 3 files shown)


components/status_icon.tsx
```diff
@@ -60,6 +60,6 @@ export default class StatusIcon extends React.PureComponent {
             IconComponent = StatusOfflineIcon;
         }
 
-        return <IconComponent className={className}/>;
+        return <IconComponent {...iconComponentProps}/>;
```
haydenhw: I'm passing the props with the spread operator like this to fix the following TS error on ```<IconComponent className={className}/>``` 

> TS2322: Type '{ className: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<{}, any, any>> & Readonly<{}> & Readonly<{ children?: ReactNode; }>'.   Property 'className' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<{}, any, any>> & Readonly<{}> & Readonly<{ children?: ReactNode; }>'.

I found the solution from [this SO post](https://stackoverflow.com/a/57312722
)

components/status_icon.tsx
```diff
@@ -36,10 +35,11 @@ export default class StatusIcon extends React.PureComponent {
         let className = 'status ' + this.props.className;
 
         if (button) {
-            className = this.props.className;
+            className = this.props.className || '';
```
enahum: why the fallback? `className` seems to have a default value

haydenhw: If I leave out  the fallback here I get this TS error on ```className```:

TS2322: Type 'string | undefined' is not assignable to type 'string'.   Type 'undefined' is not assignable to type 'string'.

The way I have it written here it seems like the TS compiler isn't taking the defaultProps into account and is expecting ```this.props.className``` to possibly be undefined since it's marked as optional in the ```Props``` type.

Another alternative is to remove the ```?```  from the ```className```  property in the ```Props``` type  so that it's just ```className: string;```.  Then the fallback can be removed from and the error goes away.

Actually, If I'm understanding [this section of the TS docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#support-for-defaultprops-in-jsx) and this [SO post correctly](https://stackoverflow.com/a/37282264), in TS 3.0+ it's recommended to omit the ```?``` in the ```Props``` type definition for any prop that has a default value supplied in ```defaultProps```.

Any thoughts on this?

saturninoabril: @enahum @devinbinnie Thoughts on this?  Feel free to merge if all are good.

enahum: I’m good

PR #5224, opened 2020-03-30
[MM-23065] - Return to a DM/channel when you reload or go to root url

ghost:
#### Summary
 This is the same fix as before for returning to a DM or channel after reloading the page or going to root URL. Tested both webapp and desktop, however I am able to close DM's and leave channels successfully this time. Let's check on the test server as well. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23065

Fix v5.23.0


actions/global_actions.jsx
```diff
@@ -292,8 +292,14 @@ async function getTeamRedirectChannelIfIsAccesible(user, team) {
         teamChannels = getChannelsNameMapInTeam(state, team.id);
     }
 
-    let channelName = LocalStorageStore.getPreviousChannelName(user.id, team.id);
+    const channelName = LocalStorageStore.getPreviousChannelName(user.id, team.id);
     channel = teamChannels[channelName];
+
+    if (typeof channel === 'undefined') {
```
devinbinnie: What happens if you're removed from a channel you're viewing? Usually we're redirected to Town Square when that happens, but from this logic it looks like it will redirect you to the last DM/GM you were in.

nevyangelova:
Tested and can actually reproduce the bug in the test server, back to investigating.

actions/global_actions.jsx
nevyangelova: This works correctly for me after my latest change.

actions/views/channel.js
```diff
@@ -149,6 +149,15 @@ export function leaveChannel(channelId) {
     };
 }
 
+export function leaveDirectChannel() {
+    return async (dispatch, getState) => {
+        const state = getState();
+        const currentUserId = getCurrentUserId(state);
+        const currentTeam = getCurrentTeam(state);
+        LocalStorageStore.removePreviousChannelName(currentUserId, currentTeam.id, state);
```
devinbinnie, line 157: I'm confused how this works, how does this remove the user from a DM channel? It looks like they would just get re-added on refresh.

actions/global_actions.jsx
devinbinnie: So upon leaving a direct channel, this function will return you to the default channel because the previous channel has been removed from the system, that makes sense.

actions/views/channel.js
Willyfrog: maybe i missread this, but if I understood it correctly is like this: if I leave a DM i get the last channel I accessed, whether is the one I'm in or not. is this right? I think we should only remove it if I was really in that channel, we would also need to take a look at the penultimate one (in case then i go and remove another one and I'm back to the dm which i shouldn't). Just to try and clarify I thought of two weird case scenarios:
- **case 1**: go to dm A, go to dm B, close dm A, reload. Haven't tested but it looks like I'd be back at A
- **case 2**: go to dm A, go to dm B, close dm C, reload. I think we are redirected to dm A.

Willyfrog (commented):
I have doubts on how are we managing this cases, which might show as bugs in weird case escenarios (if I understood correctly) can you clarify if I'm missing something?

actions/views/channel.js
nevyangelova: @Willyfrog the first scenario worked fine but in the second one I can't close dm C if I have clicked on dm A and B before so I have to check the penultimate like you said. Thanks

nevyangelova:
@Willyfrog @devinbinnie I couldn't find any bugs and couldn't reproduce the cases but I added an additional check which would make it more robust.

Willyfrog:
if you have tested and couldn'r repro, I guess that's good enough!

Willyfrog (approved):
LGTM

devinbinnie (approved):
What @Willyfrog said :P

jgilliam17 (approved):
Thank you @nevyangelova 
Tested, looks good to merge.
- Verified user is returned to DM/channel they were viewing after reload and go to root URL.
- Verified on desktop, user remains on the DM/channel they were viewing on reload and returns to the same DM/channel on restart of the app.
- Verified user can close DMs they are actively viewing, redirected to Town Square, and close DMs they not viewing without any issues.

PR #5232, opened 2020-03-30
[MM-23686] - Fix user settings modal bug on master

nevyangelova:
#### Summary
There is a bug on master that prevents the user from clicking on an option in the user settings modal. This seems to fix the issue. IMO important to merge soon because it happens on daily. 

#### Ticket Link 
https://mattermost.atlassian.net/browse/MM-23686


amyblais:
Is this bug also happening on v5.22?

jgilliam17:
> Is this bug also happening on v5.22?

Release server 5.22, does not have this issue I verified this morning.

PR #5236, opened 2020-03-31
Don't compress already-compressed files, after uploaded, as it's very…

esbeeb:
…, very CPU-expensive!

Please read starting from here:
https://forum.mattermost.org/t/processing-during-image-upload/8655/7?u=esbeeb
Basically:
1) imagemin is a very inefficient way to compress things, compared to just zip (72x faster) or tar (30x faster) on the command line.
2) There is little to no disk space saved (like 1%), from compressing already-compressed image files.

#### Related Pull Requests

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -194,7 +194,8 @@ var config = {
                 ],
             },
             {
-                test: /\.(png|eot|tiff|svg|woff2|woff|ttf|gif|mp3|jpg)$/,
+                // svg and ttf are certainly not compressed already
+                test: /\.(svg|ttf)$/,
                 use: [
                     {
                         loader: 'file-loader',
@@ -205,6 +206,20 @@ var config = {
                     {
                         loader: 'image-webpack-loader',
                         options: {},
+                    },
+                ],
+            },
+            {
+                // all these file formats are certainly already compressed
+                // There is virtually no disk space savings in compressing them again.
+                // tiff is a strange exception.  It's sometimes compressed.
+                test: /\.(png|eot|tiff|woff2|woff|gif|mp3|jpg)$/,
+                use: [
+                    {
+                        loader: 'file-loader',
+                        options: {
+                            name: 'files/[hash].[ext]',
+                        },
                     },
                 ],
             },


esbeeb:
Ok, I'll sign the CLA...

esbeeb:
OK, it's signed...

esbeeb:
Your linting tool wants lines around my comments.
"Expected line before comment  lines-around-comment"
  Fine, I don't object to that!  That's easy.

esbeeb:
Please be patient with me, as this is the first PR I've gone through this process with...

esbeeb:
Is there anything more you need from me?  Does this PR need to be discarded, and a new one made, where I put a line before each of my 2 comments?

esbeeb:
BTW: the free gift which is currently offered to new contributors: I'm way over in Malaysia, so it's probably not worth sending it to me.  Please don't bother...

webpack.config.js
```diff
@@ -194,7 +194,8 @@ var config = {
                 ],
             },
             {
-                test: /\.(png|eot|tiff|svg|woff2|woff|ttf|gif|mp3|jpg)$/,
+                // svg and ttf are certainly not compressed already
```
hanzei:
```suggestion

                // svg and ttf are certainly not compressed already
```

webpack.config.js
```diff
@@ -208,6 +209,20 @@ var config = {
                     },
                 ],
             },
+            {
+                // all these file formats are certainly already compressed
```
hanzei:
```suggestion

                // all these file formats are certainly already compressed
```

hanzei:
@esbeeb There is no need to open a new PR. :wink: The CLA is green now.

hanzei:
@jupenur Including you as a reviewer because it's a security related change. See also https://community-release.mattermost.com/core/pl/h3sqfoaeriy4tmo5yohuocq7yo

jupenur:
@esbeeb have you tested to verify this change actually resolves the original issue? I would find it very strange for something in the webpack config to be the root cause.

A more likely candidate would be the explicit image processing we do in [file.go](https://github.com/mattermost/mattermost-server/blob/master/app/file.go#L959-L978) on the server. That would also still line up with the testing you did using `.xcf` files, because `HandleImages` is only invoked on files with an `image/*` MIME type -- whereas the MIME type for `.xcf` is (on many systems) `application/x-xcf`.

hmhealey:
Hi @esbeeb. As I mentioned in the linked forum post, this wouldn't have any effect on file uploads since webpack is not used for those. The code linked by @jupenur is what is actually used for processing images uploaded by users. If you can find any improvements we could make there, those would certainly be welcome, but I'm afraid this PR is invalid.

esbeeb:
Thanks for tracking down where the code is, which is making things difficult on my RPi4.  You're right @jupenur , I did no testing, as I installed a pre-compiled build for my RPi4 (not re-compiling to see how the change to mattermost-webapp/webpack.config.js would make an effect).  

The build process is very, very tough to do on an RPi4, as the official build process will fail, since there is no MySQL docker image for ARM 32-bit (and the build process assumes there will be such a docker image; this assumption holds true on AMD64).

Thanks for explaning, @hmhealey .  Anyone who really wants to make an impact on the Mattermost codebase, to improve performance on a Raspberry Pi (or suchlike SBCs), now has a much better chance of knowing where to focus their efforts.  So feel something helpful did come out of this.   Thanks for looking into this...

We can close this PR, it wasn't the right move to make.

PR #5241, opened 2020-03-31
MM-23252 Show parent post link when deleting first reply

ethervoid:
#### Summary

Right now if we delete the first reply message the post body
doesn't show the parent post link

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23252

#### Screenshots

Peek 2020-03-31 21-06


ethervoid:
@hmhealey I want to check if this change makes sense this change. I'm wondering if this could unleash side effects

ethervoid:
@hmhealey gentle ping :)

hmhealey:
What's the change in behaviour here? I would've expected the behaviour of `isEphemeral` to be the same as checking that constant

ethervoid:
@hmhealey [isPostEphemeral](https://github.com/mattermost/mattermost-webapp/blob/master/utils/utils.jsx#L1465) checks either for ephemeral type posts and for deleted posts. So in this case, I' just check for ephemeral posts but not for deleted ones.

ethervoid:
@hmhealey gentle ping

marianunez:
I do not have enough domain knowledge here. Swapping out my review on this one.

ethervoid:
@sudheerDev gentle ping for this CR :)

PR #5242, opened 2020-03-31
[MM-23715] Modify all calls to imageURLForUser to pass a user id and default last_picture_update to 0

fm2munsh:
#### Summary
- The utility method `imageURLForUser` has a single method `userIdOrObject` but it looks like it always expects a user object as opposed to a user_id.
- ~Since there were a lot of places that called it with just an id I decided to modify the existing method to handle strings _or_ user object as it looks like thats how it was defined.~
- Modified all places that call the utility function to pass in a user or an object with an id property
- Also renamed the parameters for `imageURLForUser` and `imageURLForTeam` since they only expect `user` and `team` objects and id strings

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-23751

#### Test steps
- Go to `integrations/bots`, create a bot and give it an avatar.
- Observe that the avatar is loaded when hitting the integrations/bots page after reloading

#### Screenshots
Screen Shot 2020-03-31 at 4 27 20 PM

### DIFF
--- a/utils/utils.jsx
+++ b/utils/utils.jsx
@@ -1348,12 +1348,7 @@ export function displayEntireNameForUser(user) {
     return displayName;
 }
 
-export function imageURLForUser(userIdOrObject) {
-    let user = userIdOrObject;
-    if (typeof userIdOrObject === 'string') {
-        const state = store.getState();
-        user = getUser(state, userIdOrObject) || {id: userIdOrObject};
-    }
+export function imageURLForUser(user) {
     return Client4.getUsersRoute() + '/' + user.id + '/image?_=' + (user.last_picture_update || 0);
 }
 
--- a/utils/utils.jsx
+++ b/utils/utils.jsx
@@ -1348,8 +1348,8 @@ export function displayEntireNameForUser(user) {
     return displayName;
 }
 
-export function imageURLForUser(user) {
-    return Client4.getUsersRoute() + '/' + user.id + '/image?_=' + (user.last_picture_update || 0);
+export function imageURLForUser(userId, lastPictureUpdate = 0) {
+    return Client4.getUsersRoute() + '/' + userId + '/image?_=' + lastPictureUpdate;
 }
 
 export function defaultImageURLForUser(userId) {
--- a/utils/utils.jsx
+++ b/utils/utils.jsx
@@ -1349,7 +1349,12 @@ export function displayEntireNameForUser(user) {
 }
 
 export function imageURLForUser(userIdOrObject) {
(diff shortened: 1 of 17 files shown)


utils/utils.jsx
```diff
@@ -1349,16 +1349,21 @@ export function displayEntireNameForUser(user) {
 }
 
 export function imageURLForUser(userIdOrObject) {
-    return Client4.getUsersRoute() + '/' + userIdOrObject.id + '/image?_=' + (userIdOrObject.last_picture_update || 0);
+    let user = userIdOrObject;
+    if (typeof userIdOrObject === 'string') {
```
lieut-data, line 1353: If `userIdOrObject` is a string, and thus the user id, can we just use it directly in the URL?

fm2munsh: I tried to get the user id from the state so that we could get `last_picture_update` - not sure what purpose that serves but I figured we wanted that query parameter to be included whenever possible

lieut-data: Never mind, just realized we need the `last_picture_update` here too. Hmm.

How painful would it be to not rely on the `store.getState()`? We've been making steady progress towards eliminating global use of the state.

fm2munsh:
@lieut-data Looks like It was last modified https://github.com/mattermost/mattermost-webapp/commit/5ff65b261116646fe6d3dd067c431e3e28de1796#diff-3fadf2896914d755b6f1baa150454ef1L1405

That change seems to have basically reverted what I have here... in an attempt to get rid of direct store calls by utility functions. 

Maybe I should refactor the places that call this function to send the `user` object?

fm2munsh:
@lieut-data There are several places that call that call this utility function with a `user_id` and I think that modifying each of those components to get the user from the store before hand is too large of a change to make this late into the release.

It makes most sense to me go ahead with the change as proposed in this PR and then open a new ticket to refactor calls to `imageURLForUser` to not rely on state.

Thoughts on this @nevyangelova @hmhealey ?

nevyangelova:
@fm2munsh yes we will like to avoid calling global state in the utility functions from now on as it impacts performance, I would suggest you pass the user in the connected components where necessary. I am worried that if we roll this out it might take time to get it fixed and thus reintroducing the bad practice.

fm2munsh:
I have scrapped my initial changes and changed the PR to instead ensure all calls made to `imageURLForUser` pass a user id, or an object containing an id property in cases where there was no user available.

nevyangelova:
LGTM!

fm2munsh:
Nice - i like that approach better makes it more clear whats actually being passed. I have implemented your requested changes in the latest commit @lieut-data

PR #5245, opened 2020-04-01
MM-23395: Add websockets support for update group

catalintomai:
#### Summary
Ensuring that when we delete or update a MM group, the websocket event is triggered and we are updating the referenceable groups (et al) in Redux for group-mentions

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23395 
Note: this addresses items(events) 1-9 of the ticket - see ticket comments. 

#### Related Pull Requests
redux: https://github.com/mattermost/mattermost-redux/pull/1113
server: https://github.com/mattermost/mattermost-server/pull/14202

Note: removed the mobile changes: mattermost/mattermost-mobile#4116 from this PR since they require extra consideration and are not blocking for the current set for webapp.


actions/websocket_actions.jsx
```diff
@@ -1127,3 +1148,39 @@ function handleOpenDialogEvent(msg) {
...
+    store.dispatch({
+        type: GroupTypes.RECEIVED_GROUP,
+        data,
+    });
+}
+
+function handleGroupAssociatedToTeamEvent(msg) {
+    store.dispatch({
+        type: GroupTypes.RECEIVED_GROUP_ASSOCIATED_TO_TEAM,
+        data: {teamID: JSON.parse(JSON.stringify(msg.broadcast.team_id)), groups: [{id: msg.data.group_id}]}
```
hahmadia, line 1163: I guess I'm a little confused here but why do we do JSON.parse(JSON.stringify(...)) ? 
Isn't it valid to just do a `teamID: msg.broadcast.team_id` ?

catalintomai: Yes, I removed the unnecessary transformation, was leftover from a previous iteration.

PR #5252, opened 2020-04-01
[MM-21405] Show warning first when Sync group members is turned on without members and then show error if save without groups added

hahmadia:
#### Summary

Display a warning if no groups are added and user tries to sync members and after they press save turn it into an error (matches team configuration page)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21405

#### Screenshots

MM-21405


srkgupta:
Hi @thefactremains 
I am not sure if you noticed this, but as per the JIRA ticket, the behavior should be in consistence with Team Profile page. However on the Team Profile page, the moment we toggle the Sync Group Members radio button, we get a message at the bottom, something like this, before clicking on Save button. 

 
Should the behavior be changed in Team Profile page as well and message be displayed only after clicking on Save button (like in Channel page)

thefactremains:
@srkgupta you are correct. The purpose of this was to match the functionality of the team config page alert process. It was clearly an error on my part as the JIRA ticket states conflicting info. I have confirmed that the team config page shows the error prior to save and change the message as groups are added. 

@hahmadia my apologies, but this PR may have been built on misdirection from me. Can you please ensure that the functionality matches the team config page? I will add a comment to the JIRA ticket as well.

hahmadia:
@thefactremains @srkgupta 
Updated! Please see main post for details.

hahmadia:
> @hahmadia just found one inconsistency still, i.e. on Channel Configuration page when you hit save, it displays "Saving Config..." and then displays the error in red. However on Team Configuration page, when you hit save, it does not displays "Saving Config...". It just displays the error. The problem with displaying "Saving Config..." is that the error message shifts to the right a few pixels and then comes back which looks odd. Can we make this consistent too?

Done. I tagged Michael just to get his approval on this too.

thefactremains:
@srkgupta yes. same verbiage (though replacing "team" with "channel") is good by me.

hahmadia:
@srkgupta 
I have updated the text. Please ensure that channel config page works as usual and that the text on team config page has also not changed.

michaelgamble:
> @michaelgamble @thefactremains can you also confirm if the message is fine on the Channel configuration page. It says ` You must add at least one group to manage this team by sync group members.` whereas it should be ` You must add at least one group to manage this channel by sync group members.`. I can raise this as a separate issue on JIRA since its existing on the prior versions as well. Please confirm.

Yes you're correct in what the warning should be changed too.

michaelgamble:
Ha cool, i just noticed my review wasnt required anymore, seems like you guys have this PR under control

PR #5257, opened 2020-04-02
Fixed few failures related to toast_spec & System Scheme

srkgupta:
#### Summary
1. Deleted create_post_permission_spec as this was modifying system scheme within the test and removing create_posts permission. We have detailed tests in form of channel_moderation_spec which validates create_posts permission & other channel moderation settings. Hence removing this problematic test which is no longer required. 

2. Updated few steps in toast_spec so that all the test passes. 
3. Increased the timeout in cypress.json file from 20 seconds to 60 seconds as some of the pages takes a long time to load. 

#### Ticket Link
N/A. Fixed based on analysis of daily runs. 

#### Screenshots
Screenshot 2020-04-02 at 8 03 35 PM


e2e/cypress/integration/toast/toast_spec.js
```diff
@@ -39,10 +39,8 @@ describe('toasts', () => {
 
     beforeEach(() => {
         // # Click on town-square then off-topic channels in LHS
-        cy.get('#publicChannelList').scrollIntoView().should('be.visible').within(() => {
-            cy.findByText('Town Square').scrollIntoView().should('be.visible').click();
-            cy.findByText('Off-Topic').scrollIntoView().should('be.visible').click().wait(TIMEOUTS.TINY);
-        });
+        cy.get('#sidebarItem_town-square').should('be.visible').click();
+        cy.get('#sidebarItem_off-topic').should('be.visible').click().wait(TIMEOUTS.TINY);
```
saturninoabril, line 43: If not causing the main problem, I suggests not to go back on this type of getting target elements.

e2e/cypress.json
```diff
@@ -1,8 +1,8 @@
 {
     "baseUrl": "http://localhost:8065",
     "viewportWidth": 1300,
-    "defaultCommandTimeout": 20000,
-    "taskTimeout": 20000,
+    "defaultCommandTimeout": 60000,
+    "taskTimeout": 60000,
```
saturninoabril, line 5: I'm good here as temporary fix only but we need to further investigate whether the issue is application's performance degradation (in terms of first loading and needs to address by Dev) or is related to Cypress framework itself (SDETs to fix).

srkgupta: I had checked that the slowness is in the app. For example the posts does not loads within 20 seconds on center channel and it times out and fails the test

e2e/cypress/integration/toast/toast_spec.js
srkgupta: The problem is the channel keeps juggling between unreads section and public Channel section. Hence made this generic

PR #5262, opened 2020-04-02
[MM-22171] Move `loadedPlugins` into redux-store

Qovaros:
#### Summary
This PR removes `loadedPlugins` global variable, and switches to using the `plugins` redux store in `mattermost-webapp`.

#### Ticket Link
Fixes: https://github.com/mattermost/mattermost-server/issues/13928
JIRA: https://mattermost.atlassian.net/browse/MM-22171


reducers/plugins/index.js
```diff
@@ -321,6 +344,10 @@ function adminConsoleCustomComponents(state = {}, action) {
 
 export default combineReducers({
 
+    // object where every key is a plugin id and values are webapp plugin manifests
+    // tracks which plugins have been added as script tags to the page
+    loadedPlugins,
```
ali-farooq0, line 349: Ahh I hadn't realized that there is already a `plugins` store in here when I wrote the proposal for this ticket.

Rather than adding a new store/reducer, I wonder if we can utilize `plugins` on line 352. Would simplify things quite a bit.

So then the new proposal becomes:

1. Remove `dispatch({type: ActionTypes.RECEIVED_WEBAPP_PLUGINS, data: plugins});` from https://github.com/mattermost/mattermost-webapp/blob/63bcb51368cf8161663ad89c8107edd4653a9a7c/plugins/index.js#L66

2. Remove `loadedPlugins` store/reducer.

3. `loadedPlugins[manifest.id] = manifest;` gets replaced with `dispatch({type: ActionTypes.RECEIVED_WEBAPP_PLUGIN, data: manifest});`

4. `delete loadedPlugins[manifest.id];` gets replaced with `store.dispatch({type: ActionTypes.REMOVED_WEBAPP_PLUGIN, data: manifest});`

The overall idea is instead of creating a new store `loadedPlugins`, just use `plugins` from [line 352](https://github.com/mattermost/mattermost-webapp/pull/5262/files#diff-b62c6fcf6082fdbefb2139d0ea16d5a1R352) and rework the logic on when to add plugins to that store. Right now it's being added when we fetch all plugins from the server by invoking [`getPlugins`](https://github.com/mattermost/mattermost-webapp/blob/63bcb51368cf8161663ad89c8107edd4653a9a7c/plugins/index.js#L57) and it doesn't look like we do much with it afterwards, so maybe it's better if we use that as our `loadedPlugins` state rather than just storing it and keeping our loaded state somewhere else.

I apologies for not picking up on this earlier and it was an oversight on my part. However the new proposal above isn't much different from what you have in this change, just dispatching different actions as opposed to the new ones for `loadedPlugin`.
The benefit of the new change is that we'll have all the plugin state in one spot i.e. `store.getState().plugins.plugins` and that'll be the source of truth.

What do you think @Qovaros?

utils/constants.jsx
```diff
@@ -138,6 +138,8 @@ export const ActionTypes = keyMirror({
 
     BROWSER_CHANGE_FOCUS: null,
 
+    LOADED_PLUGIN: null,
+    UNLOADED_PLUGIN: null,
```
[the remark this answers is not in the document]
ali-farooq0: this comment can be disregarded in light of https://github.com/mattermost/mattermost-webapp/pull/5262#discussion_r413503953 since this will be removed.

reducers/plugins/index.js
ali-farooq0:
```
    RECEIVED_WEBAPP_PLUGINS: null,
    RECEIVED_WEBAPP_PLUGIN: null,
    REMOVED_WEBAPP_PLUGIN: null,
```

can be renamed to 
```
    RECEIVED_WEBAPP_PLUGINS: null, // this action will be removed entirely.
    ADD_WEBAPP_PLUGIN: null,
    REMOVE_WEBAPP_PLUGIN: null,
```

Qovaros: Thanks for review @ali-farooq0 ! I had a similar thought as you did about those two being combined. I just thought that: 
- plugins is supposed to track plugins that the app received during `getPlugins()`
- loaded plugins was supposed to track of all the plugins that are loaded currently.

That said I agree that those two should be combined, but I feel like we shouldn't remove `RECEIVED_WEBAPP_PLUGINS`. I think there should be some indication, that we received new plugins. For example:

- There should be a way to notice, that we received new versions and for some reason they weren't updated. 
- Or that a plugin wasn't removed for some reason, even though we didn't get it in `getPlugins()`.

So my final thoughts are that we should do points 2,3 and 4 from https://github.com/mattermost/mattermost-webapp/pull/5262#discussion_r413503953 but we should keep `RECEIVED_WEBAPP_PLUGINS` to keep track of received plugins. What do you think?

ali-farooq0: Great observations @Qovaros!


> There should be a way to notice, that we received new versions and for some reason they weren't updated.

Fair point, this would be a concern if we had a retry mechanism for when a plugins fail to install. Even then I would argue to fetch the latest and greatest from the server and then do the comparison, as it's possible that the cache we have locally might be out of date. 

Emitting the `RECEIVED_WEBAPP_PLUGINS` action and storing the plugins the app just received as is, is kinda like storing throw-away data. Because right when we fetch it, we attempt to install it immediately. Also looking at the current code, doesn't look like we are using the stored plugins anywhere, unless maybe I'm mistaken 🤔. So in the light of all that I would still recommend removing `RECEIVED_WEBAPP_PLUGINS` and not storing the plugins that were JUST received. I wouldn't mind keeping the state if we were actually using it. Otherwise what will happen is someone will look at this and get confused as to why we store the received plugins and then we have loaded plugins, and then every time they need to use it they will have to investigate which manifest to use, the stored or the loaded. This also simplifies all that. i.e. we will have only one source of truth. i.e. the loaded plugins.

What do you think?

Qovaros: I agree. I guess right now this data is obsolete and may cause confusion. I will update the PR and remove `RECEIVED_WEBAPP_PLUGINS` completely.

ali-farooq0: Thanks @Qovaros! And apologies again for not thinking through the original proposal and causing churn on this. However I did enjoy our discussion on this :)

Qovaros: I applied the suggested changes and also change `getPlugins` a little bit, as it doesn't need dispatch anymore.

plugins/index.js
```diff
@@ -82,15 +75,15 @@ const describePlugin = (manifest) => (
 export function loadPlugin(manifest) {
     return new Promise((resolve, reject) => {
         // Don't load it again if previously loaded
-        const oldManifest = loadedPlugins[manifest.id];
+        const oldManifest = store.getState().plugins.plugins[manifest.id];
```
ali-farooq0, line 78: Hey @Qovaros, you'll hate me for this but wondering if you can rename the `plugins` store to `manifests` as then we won't have this stutter, then it'll become
`store.getState().plugins.manifests[manifest.id];` which looks much nicer :)

Qovaros: Yeah, sure, why not. It is a good idea.

ali-farooq0:
Hey @DHaussermann this is ready for QA. Will fill in the details in the jira.

DHaussermann:
@Qovaros or @ali-farooq0  I tested the majority of the plugins in the market place. 
- No issues on install
- Ensured that enable / disable works without needing a refresh 
- No user side console errors on enable / disable. Checked Browser and Desktop client
- No need for refresh on uninstall 

The only issue I found is that with Zoom - when disabled the icon disappears but the `/zoom` remains in the slash command auto-complete. This did not occur with any other plugin I tested. Any thoughts on if it could be related to this PR?

ali-farooq0:
Thanks for the thorough QA @DHaussermann, can you confirm if same behaviour is observed in `master`? If so then it's probably isolated to the Zoom plugin and not related to this PR. I suspect it is isolated but better to confirm.

ali-farooq0:
Thanks @DHaussermann!

Thanks @Qovaros for your contribution! 🎉

merging this now.

PR #5263, opened 2020-04-03
[MM-23727] Fix validation for channel url when renaming channel

fmunshi:
#### Summary
- Adds the expected validation for channel urls when renaming a channel through the `Rename Channel` modal
- Also renames `name` in the component code to `url` since its confusing with `DisplayName` also being referred to as `name` in some places 
- Also modifies the code to allow multiple errors to be displayed if url is failing multiple validations

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23727

#### Screenshots
Screen Shot 2020-04-02 at 8 15 41 PM


components/rename_channel_modal/rename_channel_modal.jsx
```diff
@@ -161,20 +188,39 @@ export class RenameChannelModal extends React.PureComponent {
...
-                state.nameError = formatMessage(holders.lowercase);
-                state.invalid = true;
+        const urlErrors = [];
+        const url = channel.name;
+        if (url) {
+            if (url.length < 2) {
+                urlErrors.push(this.formattedError(holders.minLength));
             }
+
+            if (url.charAt(0) === '-' || url.charAt(0) === '_') {
```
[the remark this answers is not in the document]
fm2munsh: I didnt use a regex since I just copied the same url validation from the `change_url_modal`.
Do you think it makes sense to extract the validation into a utility function and refactor it for both of them as part of this PR?

fm2munsh:
@Willyfrog the reason I renamed it to `url` is because the same attribute is called `url` when creating a channel inside the `change_url_modal` component

Willyfrog:
> @Willyfrog the reason I renamed it to `url` is because the same attribute is called `url` when creating a channel inside the `change_url_modal` component

while I can wrap my head around the concept in code, what I'm concerned is about the error messages being shown, which might confuse the user. E.g.: url must start with a letter or a number, while for the wide audience the url would start by `h`

components/rename_channel_modal/rename_channel_modal.jsx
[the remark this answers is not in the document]
Willyfrog: if you are using the same validation, I'd definitely go for it, that way if at some point one changes we don't want to allow it to happen somewhere else.

fm2munsh:
> while I can wrap my head around the concept in code, what I'm concerned is about the error messages being shown, which might confuse the user. E.g.: url must start with a letter or a number, while for the wide audience the url would start by `h`

I understand the confusion but the UI has always referred to it as `URL` and all the existing translations for error messages relating to this attribute refer to `URL` - so i think changing this modal to align with the existing terminology makes the most sense as opposed to making a one off change?

utils/url.tsx
```diff
@@ -89,3 +93,41 @@ export function getScheme(url: string): string | null {
...
+}
+
+export function validateChannelUrl(url: string): React.ReactElement[] {
+    const errors: React.ReactElement[] = [];
+
+    const cleanedURL = cleanUpUrlable(url);
+    const urlMatched = url.match(/[a-z0-9]([-_\w]*)[a-z0-9]/);
+    if (cleanedURL !== url || !urlMatched || urlMatched[0] !== url || url.indexOf('__') > -1) {
+        if (url.length < 2) {
+            errors.push(formattedError(t('change_url.longer'), 'URL must be two or more characters.'));
```
Willyfrog, line 114: error strings for user still call it URL, maybe `channel's url name` or something along those lines will be less confusing? As I said I'm not concerned about code, but user experience. URL is the full thing, not just the name for it.

fm2munsh: I understand your point - `Slug` is technically the correct word for it that but I don't think this is the right place to do it.
If anything we would need to create new ticket (to be reviewed by UX potentially?)that modifies all existing references of `channel url` to something else but changing it in this one place while leaving it `URL` in all other places doesn't make sense to me. 

thoughts?

Willyfrog: sure, creating a new ticket to be reviewed by UX would be good engough :)

fmunshi:
@amyblais I dont think I can cherry-pick this for 5.23 since this component was migrated to Typescript here https://github.com/mattermost/mattermost-webapp/pull/5105 on master before I opened this PR.

Therefore these changes cannot be cleanly cherry-picked onto 5.24 without requiring a decent amount of work and a second round of QA. I think it makes most sense to just mark this (and https://github.com/mattermost/mattermost-webapp/pull/5342) for 5.24.

 since the typescript PR is also marked for 5.24.

amyblais:
@fmunshi Ok. The Jira ticket is assigned to v5.23 though. I'll change it

PR #5264, opened 2020-04-03
Updating autocomplete popovers

asaadmahmood:
#### Summary
Updating autocomplete popovers

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22956


esethna:
@asaadmahmood @matthewbirtch this is awesome!! A few things:

1. Text size look different:
image

2. Alignment is off:
image

3. Alignment and text size:

image

4. Do we want to consider slightly more compact row heights? 

5. Can we make the scroll bar only appear when scrolling or hovering over the area?
image

6. Do we want to consider a slightly different BG color for the autocomplete do it doesn't blend with the canter channel area so much?

image

7. "x" icon feels misaligned

image

asaadmahmood:
1. Intentionally different.
2. Intentional, because we are center aligning it for the large pictures that come if you have DMs in the list.
3. Intentional.
4. I'll defer that to @matthewbirtch but I like them.
5. We may need a custom scrollbar, which I would refrain from adding in this PR.
6. I'll increase the box shadow, but yes, that would still be a valid question for darker themes.
7. Intentional, per the design spec.

matthewbirtch:
1. Yes, the secondary content is intentionally smaller - although it doesn't appear to be applied everywhere (for example:  "Notifies everyone in this channel" is not smaller). 
2. What @asaadmahmood said
3. See #1 - should be resolved when this is applied everywhere
4. Since we enlarged the profile image in the autocomplete, the row height has also increased. I'm good with it as-is. It works well for me.
5. My preference certainly is for a custom scrollbar (like the designs proposed), but @asaadmahmood and I spoke and he wants to create a separate ticket for thist
6. Shadow will help with this a lot - it's not implemented yet - I don't think we should consider a different background color
7. X icon is intentionally pinned to the top right of the container - rather than aligned with the title. The title needs more breathing room from the edge, but the X icon feels like it's floating in white space if I align with the title. @asaadmahmood maybe we can bump the icon 4px up and to the right to make this look more intentional rather than a mistake.

esethna:
1. Remove glow effect on dark themes for pop-overs

matthewbirtch:
@asaadmahmood I'm still seeing a glow in dark theme. 
image

lindy65:
Thanks @asaadmahmood @esethna 

1. In the right-hand-side the suggestion modals for at-mentions and channels are not left-aligned with the "Add a comment" message input box.  There are no borders at the bottom or on the right of the modal - it just seems to disappear off the bottom and the right-side of the screen.

image

image

2. In the center channel, the positioning of the modal also looks a bit "off" to me. Is it supposed to be left-aligned with the beginning of the Write to_ message input box? Also, the bottom of the modal overlaps a little with the message input box which looks odd to me.

image

   - A comparison of the modal for the search suggestions - the modal is left-aligned with the search input box and slightly below the input field (it doesn't overlap with the input box):

image


3. In the channel switcher modal, does this line serve any purpose? Would it look better if it were removed?

image

asaadmahmood:
@lindy65 
1. Should be fixed.
2. Is as expected.
3. Is as expected.

esethna:
Thanks Lindy, @nevyangelova @bradjcoughlin ready for review

PR #5269, opened 2020-04-04
[GH-14152] Retain emoji search gif picker

ebaker:
#### Summary

- After inputting text filter in `EmojiPicker`, switching to `GifPicker` will populate `GifPicker` search bar with the filter from `EmojiPicker` and show gif results for the search text. Any changes to `GifPicker` search text will be reflected in `EmojiPicker` if the user switches back to `EmojiPicker` from `GifPicker`

- Moved `filter` from `EmojiPicker` state to `EmojiPickerTabs` state. Passed to `EmojiPicker` a `handleFilterChange` prop to update `filter` state in `EmojiPickerTabs` from `EmojiPicker`. 

- Propagated `EmojiPickerTabs` `filter` to `SearchBar` through `GifPicker` components as `defaultSearchText` prop. Likewise, passed `handleFilterChange` through `GifPicker` components as `handleSearchTextChange` prop to update `filter` state in `EmojiPickerTabs`.

- NOTE: I stored `filter` in `EmojiPickerTabs` state in a similar fashion to `EmojiPicker` which was suggested in the issue. I'd be open to placing `filter` in the redux store with some guidance on the process of aligning changes in `mattermost-redux` & `mattermost-webapp`. Since this is my first contribution to mattermost I tried to keep this fix straight forward.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/14152

#### Screenshots

mattermost-emoji-search-retain-gif-picker_2020-04-04 15-53-40 2020-04-04 15_57_33


sudheerDev:
> I'd be open to placing filter in the redux store with some guidance on the process of aligning changes in mattermost-redux & mattermost-webapp

@ebaker I think it makes sense to leave it in react state at the moment instead of moving it to redux.  we don't need the state across a whole of different components and it feels like an overkill for now. It would have been a good case If we had either more components or we need the state shared between components far in hierarchy IMO.

ebaker:
> > I'd be open to placing filter in the redux store with some guidance on the process of aligning changes in mattermost-redux & mattermost-webapp
> 
> @ebaker I think it makes sense to leave it in react state at the moment instead of moving it to redux. we don't need the state across a whole of different components and it feels like an overkill for now. It would have been a good case If we had either more components or we need the state shared between components far in hierarchy IMO.

thanks for the review and response @sudheerDev 🙂 your rationale on leaving it in react state at the moment makes sense to me and appreciate your insight on cases in which redux would make more sense. thanks 👍

PR #5270, opened 2020-04-05
[MM-23841] Allow confirm modal to update when state.checked is modified

fm2munsh:
#### Summary
- The component was not updating when checked was enabled so the checkbox would never toggle from on to off, this PR fixes that 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23841


fm2munsh:
@lieut-data i updated the existing unit tests to also check if the checkbox becomes checked after its state is set `checked: true` - this test will fail if the component does not update after the state changes so i think it should be sufficient.

PR #5275, opened 2020-04-06
Fixing javascript errors in the invitation process

jespino:
#### Summary
Fixing javascript errors in the invitation process. Added some tests that needed to fix other things to make it working properly.

The added tests fail before the change and works with the change.

#### Ticket Link
[MM-23856](https://mattermost.atlassian.net/browse/MM-23856)

### DIFF
--- a/components/invitation_modal/invitation_modal.jsx
+++ b/components/invitation_modal/invitation_modal.jsx
@@ -93,21 +93,21 @@ export default class InvitationModal extends React.Component {
             this.setState({step: STEPS_INITIAL, hasChanges: false, lastInviteChannels: [], lastInviteMesssage: '', prevStep: this.state.step});
         }
         if (this.modal && this.modal.current) {
-            this.modal.current.resetFocus();
+            this.modal.current.enforceFocus();
         }
     }
 
     goToMembers = () => {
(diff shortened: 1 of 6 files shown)


components/invitation_modal/invitation_modal.jsx
```diff
@@ -93,21 +93,21 @@ export default class InvitationModal extends React.Component {
             this.setState({step: STEPS_INITIAL, hasChanges: false, lastInviteChannels: [], lastInviteMesssage: '', prevStep: this.state.step});
         }
         if (this.modal && this.modal.current) {
-            this.modal.current.resetFocus();
+            this.modal.current.enforceFocus();
```
lieut-data, line 96: Gah, sorry about this @jespino! I didn't track that `resetFocus` had been used elsewhere when I removed it.

jespino: No problem, is not a very common usage of methods inside components while using React, but sometimes is used.

devinbinnie: Does this not break TypeScript? `enforceFocus` is a private method on `FullScreenModal`

PR #5282, opened 2020-04-07
Bucket 3: Posts

asaadmahmood:
#### Summary
Updating File Attachment
Updating Single Image title

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22842
https://mattermost.atlassian.net/browse/MM-22848

#### Screenshots
Screenshot 2020-04-07 at 1 31 13 PM
Screenshot 2020-04-07 at 1 31 20 PM


esethna:
Looks great, minor things for UX to consider:

1. Should corner rounding should be increased slightly to match single image previews? 
image

2. How can we handle other themes better? Can we be clever about the color choice for the icons so it's compatible with various center pane BG colors (without needing a white background)?  

image

Example: 
image

3. Should we consider decreasing the vertical height and icon size even more? Example comparison:

image

esethna:
Adding @andrewbrown00 given he did the mobile design and would have best knowledge about how best to adapt for webapp ^

asaadmahmood:
1. They are both 4px right now.
2. Yeah, that's one thing I was also thinking about.

matthewbirtch:
@esethna @asaadmahmood @andrewbrown00 to address themes, I think all we need to brighten the colors a bit. I did a quick test with the colors below. This should work with any theme

image
image

Excel: 1CA660
PDF: E81023
Text: 999999
PPT: D95431
Word/Others: 1976F2

As for the size of the icons, I think going down to 40px in height would work. I'll defer to @andrewbrown00 though since it sounds like he designed the original icons

matthewbirtch:
Also, can we update the download icon for this PR (and remove the circle and update to the new hover style you used with the other PRs you've been working on)? Or should that be a separate effort?

andrewbrown00:
Thanks for all the feedback here and I'm 5/5 good

> Also, can we update the download icon for this PR (and remove the circle and update to the new hover style you used with the other PRs you've been working on)? Or should that be a separate effort? See the screenshot below (this also shows a smaller attachment icon for reference)
> 
> image

I really like this design with the updated **download** icon button and reducing the overall attachment icon size to 40px.

Exploring color options to increase contrast and using one color to cover all themes. Matt's proposal is solid, we're quickly exploring tweaking some colors to hit accessibility contrast across all themes.

esethna:
+1 on the above. If we want to go a set further for cleanliness we can probably only have the download button appear when hovering over the attachment. ie doing a part implementation of this ticket: https://mattermost.atlassian.net/browse/MM-22841

cc// @asaadmahmood on all the above

esethna:
Also if we are reducing the icon height should we also consider reducing the height of the container?

asaadmahmood:
@esethna Will have a look at this tomorrow.

asaadmahmood:
@andrewbrown00 Is doing some modifications to the icons, so I'll update this PR with that.

esethna:
Looks great!

esethna:
Thanks guys. @andrewbrown00 @asaadmahmood, regarding the zoom in cursor, maybe it makes sense for images that are previewed in the post list like this:

image

But for the file attachments it feels off in my opinion, given some are previewable and other aren't, and if it does 

________

Also for the draft state, do we want to tweak the alignment of the "x" button to move it to the same position as the download button? Or are we okay with it floating in the corner? 0/5

asaadmahmood:
@esethna We do offer a way to open up the file in a modal, which has additional options (even if the preview for that is not available), so the zoom does identify some functionality, so I'll be inclined to keep it.

And regarding the x button, I would like to keep it location at the top right, as that what we usually do the x icon.

I also don't want to make it as prominent as the download button by increasing the size or placing it in the same place, as it is quite a different action.

esethna:
@asaadmahmood @andrewbrown00 I have never seen that type of a cursor used before for file type attachments. Check out gmail as an example. Even Slack only shows the zoom in cursor on images (or some things they can preview), if there's no preview they show the normal hand cursor. 

I'm 4/5 that someone will report this as a bug within a week of it getting on community-daily.

andrewbrown00:
@esethna maybe we can take a quick pole on Monday's call and see which cursor feels right.

I did check out a few other products and it seems like most use the `pointer` cursor for attachments, but I'm 0/5 if it's based on research or defaulting to using the `pointer` cursor for all **actionable** elements /which I'm 4/5 is wrong/. In this context, clicking the image or the attachment /based on type/ does open a preview /larger or zoomed view/ for almost all attachments the `zoom-in` cursor feels like the right affordance.

I do agree that if we can't /or don't currently/ show an actual preview we should use the `pointer` cursor; and also consider making **download** the primary action for non-preview-able attachments.

Slack does use the `zoom-in` cursor for images and preview-able attachments. They don't for text/code files and .zips and instead use the `pointer` cursor, but the primary action is either to download or open (for text/code files) in the RHS.

asaadmahmood:
@esethna @andrewbrown00 I don't mind the pointer, as that's just very common behaviour, while zoom based on my experience has been quite seldom.

@andrewbrown00 I don't want a different behaviour for different type of attachments, as that would just look inconsistent. You decided we should have the download button hidden because it was not the primary action. So I'm supposing that some file types will have the download button visible (if that's their primary action), while on others it would be hidden, which would be quite weird, and may even seem like a bug. As there are various file extensions that do have a preview that are not images, like code files, pdf etc.
I would prefer if we are consistent.

asaadmahmood:
I would suggest keeping the download button hidden, irrespective of primary/secondly action, and use either the zoom in pointer, or the cursor pointer (I lean more towards the cursor) for all file attachments.

The file attachments in the draft area can have a normal cursor the way it is right now.

esethna:
@asaadmahmood @andrewbrown00 since the goal of this PR is UI reskin, I'd propose if we want to make changes to UX we should create separate tickets and pass them through review.

So I'd propose in this PR to keep it moving we: 

1. Make no to changes to primary actions (ie clicking the attachment should open the preview window as it's done always)
2. Keep the pointer cursor on hover, and 
3. Have the download button hidden unless hovering on the attachment

If there's strong feels against that we can discuss on Monday?

asaadmahmood:
I'm fine with that, or keeping the zoom in cursor the way it is right now.
Visually or functionally, I don't want the thumbnails to behave differently depending on whether the preview modal shows an actual preview or not.

bradjcoughlin:
Going to review the PR but just wanted to throw out one piece of feedback: I prefer the larger thumbnail for images. Sometimes the extra few pixels make a big difference. Can we keep the thumbnail "bleed" to the edge?
image-attach-thumb-current
image-attach-thumb-pr-5282
image-attachment-thumbnail@2x

matthewbirtch:
@andrewbrown00, @asaadmahmood, I agree with @esethna on the cursor. I understand the idea behind it, but I don't think there is a mental model for that kind of cursor change for attachments. It doesn't feel quite right to me. I think we should stick with the standard click pointer cursor.

Also, I get what @bradjcoughlin is saying. I miss having a larger preview of the image attachments.  I know we discussed this idea, but in order to keep things aligned, we would have to do something like the below if we want to go with the larger full-bleed attachment image:

image

asaadmahmood:
@matthewbirtch I was already in favour of the cursor, so that sounds good to me.
And I also miss the larger thumbnail, but I'm guessing we both don't really like the border for the icons.

andrewbrown00:
We chatted about the cursors on Monday's call and decided to be consistent and use the `pointer` cursor for all images and attachments. 

We have historical feedback that added more context to why/how we handle image previews and why the **download-only** action causes friction for users since it makes previewing batch uploads confusing since we'd omit the **download-only** attachments from the preview.

asaadmahmood:
@bradjcoughlin We'll be taking in the changes right now, we're going to see how this feels. I also personally prefer the larger thumbnails. We may end up increasing the size and figure out a workaround if it feels off to others too.

esethna:
@nevyangelova @lindy65 ready for dev and QA review

nevyangelova:
LGMT minus the lint error :)

asaadmahmood:
@saturninoabril We can show it on that size. I'll fix that.

PR #5290, opened 2020-04-08
Fixing discard changes confirm modal accesibility for invitation process

jespino:
#### Summary
Fixing discard changes confirm modal accesibility for invitation process

#### Ticket Link
[MM-20934](https://mattermost.atlassian.net/browse/MM-20934)

### DIFF
--- a/components/confirm_modal.tsx
+++ b/components/confirm_modal.tsx
@@ -194,7 +194,9 @@ export default class ConfirmModal extends React.Component<Props, State> {
                 onExited={this.props.onExited}
                 id='confirmModal'
                 role='dialog'
+                aria-modal={true}
                 aria-labelledby='confirmModalLabel'
+                aria-describedby='confirmModalBody'
             >
                 <Modal.Header closeButton={false}>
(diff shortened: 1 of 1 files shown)


srkgupta:
Sorry @Willyfrog Re-requested your review by mistake

srkgupta:
@jespino just to confirm that this PR only handles one part of the JIRA ticket, i.e. Fixing discard changes confirm modal accessibility for invitation process. There are still many more things for which we need support for accessibility. Right?

jespino:
@srkgupta Yes, actually I think the most sensible approach is to split the ticket in different tickets, specially because for me is hard to use and debug with JAWS (I don't have a windows computer at home) and I'm using VoiceOver in mac for most of the tasks.

srkgupta:
@jespino honestly it will be too much to add these issues as separate ticket. Generally each section is added as a single ticket, something like this: https://mattermost.atlassian.net/browse/MM-22256. We can have multiple PRs for single JIRA ticket.

PR #5291, opened 2020-04-08
[MM-24016] - Fix channel header when there is an announcement bar in …

nevyangelova:
#### Summary
One last thing to take into consideration with the header popover was the Announcement bar on top that would change the vertical positioning of the popover.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24016


jgilliam17:
Thank you @nevyangelova 
seeing one issue
After banner is dismissed, the popover position doesn't get adjusted until refresh
Screen Shot 2020-04-08 at 12 50 38 PM

nevyangelova:
@devinbinnie @hmhealey  please take a look again. In order to get the correct top offset we were dependent on the announcement bar count which wasn't connected to redux so there was no way for the popover to know the count.

components/announcement_bar/announcement_bar.jsx
```diff
@@ -119,3 +124,20 @@ export default class AnnouncementBar extends React.PureComponent {
...
+function mapDispatchToProps(dispatch) {
+    return {
+        actions: bindActionCreators({
+            incrementAnnouncementBarCount,
+            decrementAnnouncementBarCount,
+        }, dispatch),
+    };
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(AnnouncementBar);
```
hmhealey, line 143: The connector for this component should go in its own file. You'll have to move this into its own folder for that though.

components/announcement_bar/__snapshots__/text_dismissable_bar.test.jsx.snap
```diff
@@ -1,119 +1,11 @@
...
-      }
-    />
-  }
-  onDismissal={[MockFunction]}
-  showCloseButton={true}
-  textColor=""
-  type="critical"
-/>
-`;
+exports[`components/TextDismissableBar should match snapshot 1`] = `""`;
```
hmhealey, line 3: It looks like these snapshots were broken. Do the tests for this need to be updated or is there something wrong here? [This line](https://github.com/mattermost/mattermost-webapp/pull/5291/files#diff-5d31096ca67f4f64a81f8a0587adf9c7R54) looks like it might be causing this.

reducers/views/modals.js
```diff
@@ -26,6 +26,18 @@ function modalState(state = {}, action) {
             },
         };
 
+    case ActionTypes.TRACK_ANNOUNCEMENT_BAR:
```
hmhealey, line 29: I think these should probably be in their own reducers since I don't think of the announcement bar as a modal.

components/announcement_bar/__snapshots__/text_dismissable_bar.test.jsx.snap
nevyangelova: Oops

components/announcement_bar/announcement_bar.jsx
```diff
@@ -119,3 +124,20 @@ export default class AnnouncementBar extends React.PureComponent {
         );
     }
 }
+
+function mapStateToProps(state) {
```
devinbinnie, line 128: Usually we create separate connector files for this stuff. I assume this wasn't done because there's already an `index.js` in this directory. I think it makes sense to separate these components out if we're going to be connecting to redux.

components/announcement_bar/text_dismissable_bar.jsx
```diff
@@ -51,7 +51,7 @@ export default class TextDismissableBar extends React.PureComponent {
     }
 
     render() {
-        if (this.state.dismissed) {
+        if (!this.state.dismissed) {
```
devinbinnie, line 54: Why is the logic flipped here?

nevyangelova: mistake leftover from testing 👀

utils/utils.jsx
```diff
@@ -1761,3 +1761,7 @@ export function getSortedUsers(reactions, currentUserId, profiles, teammateNameD
 
     return {currentUserReacted, users};
 }
+
+export function getAnnouncementBarCount(state) {
```
hmhealey, line 1765: This can go in something like `selectors/views/announcement_bar` or even just in `default_announcement_bar/index` to keep it with the only code that uses it. I'm still trying to avoid adding more code to `utils.jsx` since it has so much random stuff in it.

nevyangelova:
> Is the default part of "default_announcement_bar" needed? 0/5 on that one

I am 2/5 that we add default since the main folder for all bars is called 'announcement_bar' and for me it would look weird 'announcement_bar/announcement_bar' and since this is the default one I think this is more verbose.

hmhealey:
Yeah, I started to write that `components/announcement_bar/announcement_bar/announcement_bar.jsx` didn't seem that bad, but then I deleted that after I wrote it 😛

PR #5295, opened 2020-04-08
Fix messaging, plugins, team settings

josephbaylon:
#### Summary
- fix some messaging specs
- fix marketplace spec
- fix invite members spec

Don't mind the merge commits. Only 17 files are affected

#### Ticket Link
NA


srkgupta:
Hi @josephbaylon 
I see that you have modified in most of the tests and moved `before` method to `beforeEach`. Any reasons why this is done in this PR?

PR #5297, opened 2020-04-08
[MM-23977] Ensure users not in team when searching for users

fmunshi:
#### Summary
- This modal is used for inviting users to a team however it was never sending a `not_in_team_id` parameter, and ended up just falling back to `UserStore.Search` on the server. However some server changes made it so that that behaviour would default to `SearchUsersInTeam`.
- This PR ensures now that the invite to team modal sends the `not_in_team_id` and ends up calling the correct store function always 
 
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23977


amyblais:
@fm2munsh Saturn is off, maybe another dev can review?

PR #5299, opened 2020-04-09
Fix emoji, guest accounts

josephbaylon:
#### Summary
- fix react to last message shortcut spec
- fix guest invitation ui spec

Don't mind the merge commits. Only 2 files are affected

#### Ticket Link
NA


josephbaylon:
Hi @srkgupta 
I felt it necessary to change it to beforeEach because the original setup introduces more failures (and sometimes doesn't run the rest as in the cypress server results) when one of the tests fail for some reason. I had to choose between stability/coverage and length of test run.

I tried to revert back to before/after hook but it's harder to trace the issue when multiple tests fail.

PR #5301, opened 2020-04-09
[MM-24035] Fix getDeactivatedChannel selector for DM channel

fm2munsh:
#### Summary
- There was an extra argument so teammate was always undefined resulting in a deactivated DM channel never showing up as deactivated

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24035

#### Screenshots

Screen Shot 2020-04-08 at 10 49 49 PM

### DIFF
--- a/components/channel_view/index.js
+++ b/components/channel_view/index.js
@@ -23,7 +23,7 @@ const getDeactivatedChannel = createSelector(
     (state, channelId) => {
         return getDirectTeammate(state, channelId);
     },
-    (users, channelId, teammate) => {
+    (users, teammate) => {
         return Boolean(teammate && teammate.delete_at);
     }
 );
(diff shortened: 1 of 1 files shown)


components/channel_view/index.js
```diff
@@ -23,7 +23,7 @@ const getDeactivatedChannel = createSelector(
     (state, channelId) => {
         return getDirectTeammate(state, channelId);
     },
-    (users, channelId, teammate) => {
+    (users, teammate) => {
```
nevyangelova, line 26: 👍  Seems like we can also remove the users param since its not being used too.

amyblais:
@fm2munsh Mario is off the rest of the week, maybe another dev can review?

PR #5302, opened 2020-04-09
[MM-24027] Add profile pop-over when clicking a user in the View/Manage members modals

shred86:
#### Summary

Added the ability to view user profile pop-over when clicking the profile picture or username from the view/manage members modal. Mimics the behavior when clicking a profile picture or username from a post.

Tested on Chrome 81.0.4044.92, Safari 13.1 (15609.1.20.111.8) and the Mattermost web app 4.4.0 (4.4.0.5682).

#### Ticket Link

JIRA ticket: [MM-24027](https://mattermost.atlassian.net/browse/MM-24027)
Fixes https://github.com/mattermost/mattermost-server/issues/14266
#### Screenshots

Screen Shot 2020-04-10 at 10 46 31 PM


shred86:
One issue I'll need to fix, the ` - ` is showing up when a user doesn't have a nickname.

shred86:
> @shred86 sweet! One issue I found is that clicking "Send Message" in the profile pop-over should clode the view/manage members modal

Hm, I'm trying to duplicate this but not sure how. If I click "Send Message" from the pop-over profile, it opens a direct message window thus "closing" everything.

esethna:
> @esethna Hm, I'm trying to duplicate this but not sure how. If I click "Send Message" from the pop-over profile, it opens a direct message window thus "closing" everything.

Here's an example of the repro @shred86 :
6a527cd5b9ed6f88942bd6e411be33a8

Thanks for pointing out those other cases. Given you are taking action on another modal in the case of "Account Settings" and "Add to channel", we should avoid having modals over modals, hence closing the orginal view members modal when those actions are taken

shred86:
Ah, yeah that's not good! That's strange I'm not seeing the same behavior in Safari or Chrome. I am trying to figure out how to close the View/Manage Members modal though. I posted a message in the [Developers: Web App](https://pre-release.mattermost.com/core/channels/webapp) channel hoping to get some pointers because I'm at a bit of a road block with my knowledge of React. Will keep working to figure it out.

shred86:
Some additional issues identified and fixed: the `hasMention`prop wasn't being passed for `UserProfile` or `ProfilePicture` thus making the `@username` not a clickable link to populate the message bar. Yet another case the View/Manage Members modal will need to be closed...

shred86:
This is the best solution I can come up with right now for being able to close the View/Manage Members modal. I'm passing the `handleExit` method as a prop from `ChannelMembersModal` down the component tree to the `ProfilePopover` component. From here, it's a simple check to see if the prop that I passed down exists (which will only exist if the profile popover is accessed from the View/Manage Members modal), then call `handleExit` to close the View/Manage Members modal. This check is obviously required since the profile popover can be accessed from main channel message windows when clicking a username.

There is one known issue I haven't quite figured out but when you load up the web app for the first time (i.e. refresh the browser), the first time you select "Send Message" from the popover profile when accessed from the View/Manage Members modal, it throws this error:
```
react-dom.development.js:506 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
```
It only occurs the first time when clicking "Send Message" and only from the popover profile when accessed from the View/Manage Members modal.

components/profile_popover/profile_popover.jsx
```diff
@@ -188,6 +190,19 @@ class ProfilePopover extends React.PureComponent {
...
+        this.handleChannelMembersModalExit();
+    }
+
+    handleAddToChannel = (e) => {
+        e.preventDefault();
+
+        this.handleChannelMembersModalExit();
+    }
+
+    handleChannelMembersModalExit = () => {
```
shred86, line 202: This method should be renamed to something a little more generic since it’s handling the closing of both the channel and team member as modals. Perhaps, `handleCloseMembersModal`.

shred86:
After doing some more investigating, I think I'm understanding this a bit better and where the potential issue is. How I'm currently closing the View/Manage Team and Members modal works (passing `handleExit` prop through many different child components), but I think a potentially better way is using `closeModal` in actions/views/modals. However, there is a slight "issue" with how the View/Manage Team and Members modal is being created. When it's opened via the menu or clicking the channel name, that occurs through the `ChannelHeaderDropdown` component which uses `Menu.ItemToggleModalRedux`. Using this will result in the `openModal` action being called which saves the modal ID and state to redux. This allows us to access and close the modals using `closeModal`.

When the View/Manage Team and Members modal is accessed through the `PopoverListMembers` component (small person icon on top right of the screen), all of the states for modals opened this way are stored in this component. The modals are being created directly (e.g. importing the `ChannelMembersModal` component directly), thus the `openModal` action is never called and the model ID and state is not stored using redux, meaning `closeModal` can't be utilized.

So I see two options:
1. Keep the current implementation and just pass the `handleExit` prop down the component tree as I'm doing currently.
2. Restructure `PopoverListMembers` so it's calling the `openModal` action to store the modal ID and state in redux. This way we can close the modals using `closeModal`.

The second option seems "cleaner", but one issue still remains which is I would have to call `closeModal` on both the View/Manage Team **and** Members modals, which would result in `closeModal` being called on a modal that doesn't exist. The obvious work around is check where it's being called from, but my understanding is that would still require passing a prop down the component tree unless there's another option like checking if that modal state exists first.

shred86:
Major changes made with the last commit:
- When `ChannelMembersModal` is accessed from `PopoverListMembers`, it now opens it using `openModal` which allows us to now close the modal using `closeModal`. With these changes, I've removed the passing of props down several children components to close the View/Manage Members modal.
- Added a new method to `ProfilePopover` called `handleCloseModals` which checks if the modal exists in redux. If it does, it then checks the `open` property and if it's `true`, `closeModal` is called.
- Minor refactoring of some code and clean up between `ChannelMembersModal` and `TeamMembersModal`.

There is a lot of commented out code in popover_list_members.jsx for ease of review. Some of it is based on the new `handleCloseModals` method but most of it is code that I cannot figure out its purpose and was hoping to get some feedback. I've spent a couple hours going over every scenario I could think of and I can't figure out why or when the Team Members and Channel Invite modals would need to be opened or closed from the `PopoverListMembers` component.

I'm holding off on updating the tests for now since depending on whether or not I'm able to remove the code mentioned above in popover_list_members.jsx will determine if some changes to the tests need to be made.

bradjcoughlin:
@shred86 I would suggest removing the comments because as a PR reviewer I can see the removed lines. The comments just end up looking like new lines in the diff 😃

shred86:
@bradjcoughlin Makes sense - I probably shouldn't have even made the commit yet. I basically wanted to know if those commented out functions are even used and didn't want to delete them quite yet since I wasn't sure, but it's easy enough to copy and paste them back in! :)

jgilliam17:
@shred86 
I noticed during PR testing, but thought it was out of scope here and an issue on master.
Confirmed it is not broken on master.test, with help from @ogi-m. 
Can you please check if this PR is breaking the guest label, appears doubled on the manage member modal?
Thank you.

Screen Shot 2020-04-17 at 12 17 28 PM

shred86:
@jgilliam17 Great catch! I found the issue. Since we're using the `UserProfile` component now to display the username which includes the bot and guest badges. I've removed the redundant bot and guest badge in `UserListRow`.

components/popover_list_members/popover_list_members.jsx
```diff
@@ -80,26 +75,15 @@ export default class PopoverListMembers extends React.Component {
...
-    showChannelInviteModal = () => {
-        this.setState({showChannelInviteModal: true});
-    };
+        this.closePopover();
 
-    hideChannelInviteModal = () => {
-        this.setState({showChannelInviteModal: false});
-    };
+        const modalData = {
+            modalId: ModalIdentifiers.CHANNEL_MEMBERS,
```
nevyangelova, line 81: How are we handling the other removed modals?

shred86: Is this in reference to the methods I deleted (`hideTeamMembersModal`, `showChannelInviteModal`, etc.)? I was very confused as to why these modals were referenced here. I could not find any situation where the Team Members Modal and Channel Invite Modal would ever be active/open when the profile pop-over is used. I tested every situation I could think of and couldn't find a reason to not delete them. I definitely could be missing something though. I asked twice in the Developers channel to be sure but haven't gotten an answer yet.

nevyangelova: Makes sense, I am also new and not that familiar with the codebase, I would wait for an answer on that, otherwise LGTM 👍

hmhealey: I did some digging and I found out why that unnecessary code had existed. At one point, both the header, the mobile header, and the dropdown all had their own versions of those modals. In https://github.com/mattermost/mattermost-webapp/pull/1666, the ChannelHeader and Navbar (the mobile header) components were refactored to remove a bunch of duplication there, but this component was forgotten about so it still used the old way of opening those modals.

Thanks for helping us clean up a bunch of old, dead code!

shred86: Perfect! Thanks for looking into that.

PR #5303, opened 2020-04-09
Quick wins bucket 4

asaadmahmood:
#### Summary
MM-22851 - Emoji reactions improvements
MM-22807 - Mention highlight improvements
- I didn't update the highlight color for the mention highlight, as that would be a theme change. So would prefer to look at any theme changes separately.
- Also didn't add the animation for the emoji reaction count bump, as that may cause a lot of issues, especially if a count update animation is happening, and another user adds a reaction while the animation is happening, we would either have to change the number in the middle of the animation, or wait for the animation to complete, and then trigger another animation to update to the new count, etc. Making it quite complicated.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22851
https://mattermost.atlassian.net/browse/MM-22807

#### Screenshots
Screenshot 2020-04-09 at 2 17 50 PM
Screenshot 2020-04-09 at 2 18 01 PM


matthewbirtch:
@asaadmahmood for the animation, if the number changes while the animation is happening, I think we would finish the animation and then update the count (without any animation) after the animation finishes. Also, I envisioned the count animation only happening for your own reaction - just to visually make it clear that your own reaction counted. I wouldn't expect the animation to happen for all other reactions from other users.

asaadmahmood:
@matthewbirtch I think we may need to get that done in another ticket, or maybe have another develop work on it. As this may be that straightforward or quick.

matthewbirtch:
Okay. Could you create a separate ticket for that effort?

asaadmahmood:
@matthewbirtch Here it is, you can add more details on the exact behaviour on the ticket:
https://mattermost.atlassian.net/browse/MM-24052

(i.e. the animation should only occur for the current user, and how other cases are to be handled).

esethna:
Awesome! Looks great! 

@matthewbirtch propose we create a separate quick wins ticket with the animation details and cases, so we can close this ticket and run it through QA

matthewbirtch:
@esethna already created: https://mattermost.atlassian.net/browse/MM-24052

esethna:
@asaadmahmood the reaction is missing a click effect when *removing* a reaction

matthewbirtch:
@asaadmahmood, as @esethna noted, can you take a look at the hover state? It seems to be using a white background regardless of the selected theme. It should be using the channel background color.

image

image

asaadmahmood:
Done @matthewbirtch @esethna.

esethna:
@lindy65 @nevyangelova ready for dev and QA review

lindy65:
Thanks @asaadmahmood 

Just a tiny thing...

- The padding on the left of the `at-here`, `at-all` and `at-channel` seems to be less than the padding to the left of a mentioned user:

image

asaadmahmood:
@lindy65 Fixed.

PR #5304, opened 2020-04-09
Added E2E Cypress test to verify Accessibility Support in different Buttons

srkgupta:
#### Summary
Added E2E Cypress test to verify Accessibility Support in different Buttons

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-22624

#### Screenshots
Screenshot 2020-04-09 at 6 44 12 PM


e2e/cypress/integration/accessibility/accessibility_buttons_spec.js
```diff
@@ -0,0 +1,80 @@
...
+        cy.get('#toggleFavorite').should('have.attr', 'aria-label', 'add to favorites').and('have.class', 'a11y--active a11y--focused').click();
+
+        // * Verify accessibility support if Channel is added to Favorites
+        cy.get('#toggleFavorite').should('have.attr', 'aria-label', 'remove from favorites');
+
+        // # Ensure the focus is on the Toggle Favorites button
+        cy.get('#toggleFavorite').focus().tab({shift: true}).tab();
+
+        // # Press tab until the focus is on the Pinned posts button
+        cy.focused().tab().tab().tab().tab();
```
prapti, line 66: This generally works fine, but in my case where I have a plugin that has an icon in the header between the Pin and Members icons, it's failing. Not sure if we want to tackle this scenario right now to check if the focus is on the pin button?

srkgupta: @prapti It will be difficult to tackle this, i.e. there might be some other plugins locally installed which the test might not know. It will be too much to set up a plugin in this test as its not dependent & out of scope of this test & generally on the Cypress test instance we would not have the plugin enabled while running this test. Hence I think it would be okay to skip this for now. 
Screenshot 2020-04-14 at 6 10 54 PM

PR #5319, opened 2020-04-13
MM-24108 Hide next/prev team shortcuts on Linux

hmhealey:
Due to conflicts between the default hotkeys on Ubuntu, we're disabling this shortcut on Linux temporarily while we figure out either a replacement hotkey or a more general solution to the problem.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24108


hmhealey:
Removing the Do Not Merge label since we can cherry-pick it after release-5.23 if it gets merged before then.

DHaussermann:
@hmhealey This needs 1 more small change.
The shortcut for switching to a team by number is still shown. This also does not work and should be be removed. (On my machine this shortcut moves the application to a specific display)

hmhealey:
@DHaussermann That's on Ubuntu? The ticket says that those shortcuts still work fine and shouldn't be disabled, so we're getting some mixed behaviour here.

DHaussermann:
@hmhealey On a second look, it seems like we can leave this in. 
The behavior I was describing where the app changes to the that display is ONLY when using the number keys from the number pad (with numlock enabled). I thought this was happening for all number keys but was mistaken. 

This actually has nothing to do with our app. Operating Systems treat the number pad keys differently when you're holding down other modifiers keys so they don't actually produce a number.
 
The instruction are valid.
LGTM!

PR #5320, opened 2020-04-13
MM-23994 Fix slash command /leave causing webapp stuck trying to load left channel

sudheerDev:
#### Summary
We recently made a change https://github.com/mattermost/mattermost-webapp/pull/5173 to use dispatch leave channel before the redirect of the url which is causing the channel to be rejoined again because of the behaviour of remounting the center channel with `/teamUrl`

The last change i could trace for dispatching leave channel is https://github.com/mattermost/mattermost-webapp/commit/14d6c93688b0e467526c3b57f868e61c6e78d177 which does not seem to be necessay as we fetch things like channel again on leave events.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23994


sudheerDev:
@amyblais i propose this to be considered for 5.23v instead of 5.22v. Change is not huge but there are some moving parts and few races conditions which are evident with recent changes. 

Reverting the code which caused this will leave us with https://mattermost.atlassian.net/browse/MM-23589 which seems like a bigger issue to me

jgilliam17:
Thank you @sudheerDev 
I am seeing an issue when leaving a public channel, server opens in new tab, non secure and in a weird state e.g. no longer able to be added to some public or private channels by another user or using /join)
(sorry, used QuickTime to record my video forgetting github doesn't take .mov, zip instead)
[Screen Recording 2020-04-16 at 10.15.36 AM.mov.zip](https://github.com/mattermost/mattermost-webapp/files/4487725/Screen.Recording.2020-04-16.at.10.15.36.AM.mov.zip)

sudheerDev:
@jgilliam17 weird the changes seem to work fine locally and I don't see how the issues mentioned above can be caused by my changes.

PR #5325, opened 2020-04-14
[MM-20598] Migrate components/delete_channel_modal to TypeScript

shred86:
#### Summary

Migrates components/delete_channel_modal and associated test to TypeScript. First attempt at trying to convert a component to TypeScript and I'm having some issues with the tests specifically. Everything works as expected and the tests passed locally, but I'm receiving some TypeScript errors in VSCode which I've left comments on in the code. I'll make comments on each section as well with the things I've tried. Once I've resolved the TypeScript errors in VSCode, I'll make a new commit and remove the comments.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/13476

#### Related Pull Requests

N/A

#### Screenshots

N/A


components/delete_channel_modal/delete_channel_modal.test.tsx
```diff
@@ -56,7 +64,7 @@ describe('components/delete_channel_modal', () => {
         );
 
         wrapper.setState({show: true});
-        wrapper.instance().onHide();
+        wrapper.instance().onHide(); // Property 'onHide' does not exist on type 'Component<{}, {}, any>'
```
shred86, line 67: As noted in the comments, I'm getting a TypeScript error in VSCode: `Property 'onHide' does not exist on type 'Component<{}, {}, any>'`. I noticed a couple other components made `onHide` a public method in the actual component with a note that it was required for tests, but even after doing that I was still getting the same error.

components/delete_channel_modal/delete_channel_modal.test.tsx
```diff
@@ -69,7 +77,7 @@ describe('components/delete_channel_modal', () => {
         );
 
         wrapper.setState({show: true});
-        wrapper.instance().handleDelete();
+        wrapper.instance().handleDelete(); // Property 'handleDelete' does not exist on type 'Component<{}, {}, any>'
```
shred86, line 80: As noted in the comments, I'm getting a TypeScript error in VSCode: `Property 'handleDelete' does not exist on type 'Component<{}, {}, any>'`.

components/delete_channel_modal/delete_channel_modal.test.tsx
```diff
@@ -84,7 +92,7 @@ describe('components/delete_channel_modal', () => {
             <DeleteChannelModal {...props}/>
         );
 
-        wrapper.find(Modal).props().onExited();
+        wrapper.find(Modal).props().onExited(document.createElement('div')); // Cannot invoke an object which is possibly 'undefined'
```
shred86, line 95: The first error I was receiving here was that `onExited` requires an HTML argument. I noticed some other tests passing in `document.createElement('div')` which I'm assuming is just to provide some sort of HTML element for the test, so I ended up doing the same. That resolved that error.

However, as the comment mentions, I'm still receiving this error: `Cannot invoke an object which is possibly 'undefined'`.

components/delete_channel_modal/delete_channel_modal.test.tsx
```diff
@@ -37,7 +45,7 @@ describe('components/delete_channel_modal', () => {
         channel,
         currentTeamDetails,
         actions: {
-            deleteChannel: emptyFunction,
+            deleteChannel
         },
         onHide: emptyFunction,
```
shred86, line 50: I've noticed some tests using `jest.fn()` instead of `emptyFunction`. Is there any reason not to just use `jest.fn()` which, to my understanding, basically returns an empty mocked function?

jasonblais:
Tagging @saturninoabril to help with the above questions on tests

components/delete_channel_modal/delete_channel_modal.test.tsx
devinbinnie: `jest.fn()` allows our testing library to detect whether a function has been called or not. It's not always necessary, but it's the preferred way of mocking out functions.

components/delete_channel_modal/delete_channel_modal.test.tsx
devinbinnie: You need to specific a type for the generic `shallow()` method, since it can't autodetect.

Something like:
```
const wrapper = shallow<DeleteChannelModal>(
    <DeleteChannelModal {...baseProps}/>
);
```

components/delete_channel_modal/delete_channel_modal.test.tsx
devinbinnie: See comment above.

components/delete_channel_modal/delete_channel_modal.test.tsx
devinbinnie: You shouldn't need to pass anything into `onExited`.
I think this issue is that `wrapper.find()` is returning a potentially `undefined` object since it might not find the object it's looking for. So you probably need to do some optional chaining or assertion on the result of `wrapper.find()`.

components/delete_channel_modal/delete_channel_modal.test.tsx
shred86: Ah, got it. Thanks!

components/delete_channel_modal/delete_channel_modal.test.tsx
shred86: That makes sense. I've replaced the empty function with `jest.fn()`.

components/delete_channel_modal/delete_channel_modal.test.tsx
shred86: I'm using optional chaining:

`wrapper.find(Modal).props().onExited?.();`

I'm seeing two errors:

1. Eslint: "Expected an assignment or function call and instead saw an expression." and one from typescript: "Expected 1 arguments, but got 0."  I don't think that one should matter for a test and can be ignored, but I'm not entirely sure.

2. Typescript error, that is what I mentioned before with `onExited` requiring an argument, specifically an HTMLElement.

components/delete_channel_modal/delete_channel_modal.test.tsx
```diff
@@ -87,7 +87,7 @@ describe('components/delete_channel_modal', () => {
             <DeleteChannelModal {...props}/>
         );
 
-        wrapper.find(Modal).props().onExited?.(document.createElement('div'));
+        wrapper.find(Modal).props().onExited!(document.createElement('div'));
```
shred86, line 90: Fixed the previous TypeScript errors by using a non-null assertion operator which makes sense for this case, but I still have to provide an argument for `onExited`. Otherwise, it will throw a TypeScript error:

"Expected 1 arguments, but got 0.", specifically an HTMLElement. It appears an argument is required:

"(method) TransitionCallbacks.onExited?(node: HTMLElement): any"

Just want to make sure I'm not missing anything here. This PR should hopefully be good to go.

devinbinnie: Can you pass in `null`? Our implementation just doesn't happen to need any arguments. I'm not sure what the `onExited` method is looking for otherwise. It's probably coming from the `react-bootstrap` type definitions.

shred86: @devinbinnie I did attempt that but I get a TypeScript error:

"Argument of type 'null' is not assignable to parameter of type 'HTMLElement'."

shred86: Any other thoughts on this one? I’m out of ideas on how to make TypeScript happy without passing an HTMLElement in. :)

devinbinnie: Yeah, I guess we'll have to live with that for now.
As long as the comment is there and the tests are passing we should be okay. Thanks for looking into this :)

PR #5330, opened 2020-04-14
[MM-23159] Highlight Group Names

hahmadia:
#### Summary

Highlights group name mentions 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23159

#### Related Pull Requests

Waiting on:
https://github.com/mattermost/mattermost-redux/pull/1127

#### Screenshots

image


package.json
```diff
@@ -37,7 +37,7 @@
     "localforage-observable": "2.0.0",
     "mark.js": "8.11.1",
     "marked": "github:mattermost/marked#5a06d1af25ca48927b9efe0d9c42c01c84e7839b",
-    "mattermost-redux": "github:mattermost/mattermost-redux#e3e67f1bc0ebe852d586ee19323fa3e9394a7dd1",
+    "mattermost-redux": "github:mattermost/mattermost-redux#1329ad2223aadc96cf47be6cd2527a553aba3c9e",
```
fmunshi, line 40: i think you need to change the package-lock too

hahmadia: Good point!

package-lock.json
```diff
@@ -14258,39 +14258,58 @@
       "integrity": "sha512-6qE4B9deFBIa9YSpOc9O0Sgc43zTeVYbgDT5veRKSlB2+ZuHNoVVxA1L/ckMUayV9Ay9y7Z/SZCLcGteW9i7bg=="
     },
```
catalintomai, line 14259: Do you need all these changes or just the redux hash update?

hahmadia: I update the package.json redux hash and then do a `npm install` and submit whatever npm changed in the package-lock.json file. Is there a better way to do it? I'm just submitting whatever NPM decided to update in the package lock file.

catalintomai: Maybe it does, I usually err on the side of doing only the changes explicitly related to the feature and try to avoid lateral changes that I do not understand, that can create side-effects.

fmunshi: yea I would avoid checking in these package changes - easy way to fix is 
`git checkout upstream/master package-lock.json` 
Then manually update the mattermost-redux hash inside of `package-lock.json` and commit only those changes

hahmadia: Noted. Thanks!

catalintomai: you might need to update the hash to the one of the checkin for this PR: https://github.com/mattermost/mattermost-redux/pull/1129/files

hahmadia: Done! Thanks

hahmadia:
@srkgupta 
Hey Rohitesh,

I realised for me to fix MM-24177 , I need some of the work I've done in https://github.com/mattermost/mattermost-webapp/pull/5331 (some helper functions). I think it would be better for me to fix this after that PR is merged rather than re-doing a lot of the work I've already done in that PR. So I will wait for the above PR to get merged first and then move forward with fixing this bug.

PR #5331, opened 2020-04-14
[MM 23264] Show Confirmation Modal for group mentions

hahmadia:
**Change redux hash before merge to master!**

#### Summary

Show the confirmation modal for group mentions 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23264

#### Related Pull Requests

https://github.com/mattermost/mattermost-server/pull/14068
https://github.com/mattermost/mattermost-redux/pull/1086


components/create_comment/create_comment.jsx
```diff
@@ -211,6 +218,13 @@ class CreateComment extends React.PureComponent {
          * Should preview be showed
          */
         shouldShowPreview: PropTypes.bool.isRequired,
+
+        /*
+            Group member mention
+        */
+        selectChannelMemberCountsByGroup: PropTypes.func.isRequired,
+        allowReferencedGroups: PropTypes.array,
```
fmunshi, line 226: the wording of this is kinda weird it makes me think it should be a boolean maybe something like `groupsWithAllowReference` would make more sense as a name? 0/5 on renaming it though

components/create_comment/create_comment.jsx
```diff
@@ -465,17 +490,45 @@ class CreateComment extends React.PureComponent {
             this.setState({draft: updatedDraft});
         }
 
-        if (enableConfirmNotificationsToChannel && useChannelMentions &&
+        const notificationsToChannel = enableConfirmNotificationsToChannel && useChannelMentions;
+        let memberNotifyCount = 0;
+        let channelTimezoneCount = 0;
+        let mentions = [];
+        const notContainsAtChannel = !containsAtChannel(draft.message);
+        if (enableConfirmNotificationsToChannel && notContainsAtChannel) {
```
fmunshi, line 498: i think a `useGroupMentions` check here would make sense?

components/create_comment/create_comment.jsx
```diff
@@ -465,17 +490,45 @@ class CreateComment extends React.PureComponent {
...
+                        }
+                        return `@${group.name}`;
+                    });
+            }
+        }
+
+        if (notificationsToChannel &&
             channelMembersCount > Constants.NOTIFY_ALL_MEMBERS &&
-            containsAtChannel(this.state.draft.message)) {
+            notContainsAtChannel) {
```
fmunshi, line 516: why is this `notContainsAtChannel`?

components/create_post/index.js
```diff
@@ -112,6 +116,8 @@ function makeMapStateToProps() {
             canPost,
             useChannelMentions,
             shouldShowPreview: showPreviewOnCreatePost(state),
+            allowReferencedGroups: getAllAssociatedGroupsForReference(state),
```
fmunshi, line 119: thoughts on converting this into a map of the format 
`{groupAName: groupA, groupBName: groupB}` to make some of the computations faster when calling groups mentioned in text?

utils/post_utils.jsx
```diff
@@ -122,6 +124,17 @@ export function containsAtChannel(text, options = {}) {
...
 
+export const groupsMentionedInText = (text, groups) => {
+    // Don't warn for slash commands
+    if (!text || text.startsWith('/')) {
+        return [];
+    }
+
+    const mentionableText = formatWithRenderer(text, new MentionableRenderer());
+    const mentions = allAtMentions(mentionableText);
+    return mentions.length > 0 && groups.filter((group) => mentions.includes(`@${group.name}`));
```
fmunshi, line 135: specifically talking about optimizing this function here right now it would be quite inefficient with a large list of mentions and a large list of groups. 

could replace it with something like `mentions.filter((mention) => groups[mention]));`

components/create_comment/create_comment.jsx
hahmadia: Good idea! Done.

components/create_comment/create_comment.jsx
hahmadia: Mind explaining how I would go about doing this? Little confused

components/create_comment/create_comment.jsx
hahmadia: True that! Done! It had to be `!notContainsAtChannel`

components/create_post/index.js
hahmadia: Prodigy thoughts Farhan! Thanks for this! I implemented using map instead.

utils/post_utils.jsx
hahmadia: Implemented! Thanks

components/create_comment/create_comment.jsx
hahmadia: After your explanation, I better understood and implemented. Please let me know what you think!

components/create_post/create_post.jsx
```diff
@@ -275,8 +276,14 @@ class CreatePost extends React.PureComponent {
             /**
              * Function to set or unset emoji picker for last message
              */
-            emitShortcutReactToLastPostFrom: PropTypes.func
+            emitShortcutReactToLastPostFrom: PropTypes.func,
+
+            selectChannelMemberCountsByGroup: PropTypes.func,
         }).isRequired,
+
+        groupsWithAllowReference: PropTypes.object,
```
fmunshi, line 284: should this either be `required` _or_ have a default value?

components/create_post/create_post.jsx
```diff
@@ -275,8 +276,14 @@ class CreatePost extends React.PureComponent {
...
+        channelMemberCountsByGroup: PropTypes.object,
```
fmunshi, line 285: same here might make sense to just make both of these default `{}`

hahmadia: For `getChannelMemberCountsByGroup`, the selector already returns an empty `{}`.
For `groupsWithAllowReference`, I do a 
```
new Map(getAssociatedGroupsForReference(state, channel.team_id, channel.id).map((group) => [`@${group.name}`, group]))
```
Which if the array inside not present, it would get `new Map([])` which simply returns a empty Map without any entries.

hahmadia:
@srkgupta I have addressed those bugs you have mentioned. Ready for re-test. Thanks

PR #5332, opened 2020-04-14
Fix username, fullname, nickname

josephbaylon:
#### Summary
Fix tests failing on username, fullname, nickname

#### Ticket Link
NA


josephbaylon:
@srkgupta I'm not getting those failures locally. We can probably address those when they fail on cypress server?
Screen Shot 2020-04-14 at 10 43 02 AM

PR #5333, opened 2020-04-14
MM-23933 Fix for post text box overriding content on paste of github codeblock

sudheerDev:
#### Summary
 * Message area should consider existing text and append any codeblock based on the cursor position
 * Consider the same logic for RHS

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23933


components/create_comment/create_comment.jsx
```diff
@@ -334,9 +344,17 @@ class CreateComment extends React.PureComponent {
...
+
         if (isGitHubCodeBlock(table.className)) {
-            message = '```\n' + getPlainText(e.clipboardData) + '\n```';
+            const codeBlock = '```\n' + getPlainText(e.clipboardData) + '\n```';
+            const {firstPiece, lastPiece} = splitMessageBasedOnCaretPosition(this.state.caretPosition, message);
+
+            // check whether the first piece of the message is empty when cursor is placed at beginning of message and avoid adding an empty string at the beginning of the message
+            message = firstPiece === '' ? `${codeBlock}\n${lastPiece}` : `${firstPiece}\n${codeBlock}\n${lastPiece}`;
+
+            const newCaretPosition = firstPiece === '' ? `${codeBlock}`.length : `${firstPiece} ${codeBlock}`.length;
```
Willyfrog, line 356: You can simplify these lines calculating if you need the extra LF first. Also added a check for LF after the codeblock to prevent adding an innecesary one (not that most people would notice).
```suggestion
            const {firstPiece, lastPiece} = splitMessageBasedOnCaretPosition(this.state.caretPosition, message);
            const requireStartLF = firstPiece === '' ? '' : '\n';
            const requireEndLF = lastPiece === '' ? '' : '\n';
            const codeBlock = requireStartLF + '```\n' + getPlainText(e.clipboardData) + '\n```' + requireEndLF;

            // check whether the first piece of the message is empty when cursor is placed at beginning of message and avoid adding an empty string at the beginning of the message
            message = `${firstPiece}${codeBlock}${lastPiece}`;

            const newCaretPosition = this.state.caretPosition + codeBlock.length;
```

components/create_post/create_post.jsx
```diff
@@ -763,12 +763,19 @@ class CreatePost extends React.PureComponent {
 
         e.preventDefault();
 
-        let message = '';
+        let message = this.state.message;
         if (isGitHubCodeBlock(table.className)) {
-            message = '```\n' + getPlainText(e.clipboardData) + '\n```';
-        } else {
-            message = formatMarkdownTableMessage(table, this.state.message.trim());
+            const codeBlock = '```\n' + getPlainText(e.clipboardData) + '\n```';
```
Willyfrog, line 768: not that it would be a huge gain in processing, but I would have put both in the same function to make sure it is treated the same way in both places

components/create_comment/create_comment.jsx
sudheerDev: Makes sense. TBH I pretty much copied this from a similar method.

components/create_post/create_post.jsx
sudheerDev: Done

Willyfrog:
I was going to test if any code from github has a table (it might have, but i couldn't see it locally) so i logged into the server and tried to add a post with some code block from github and it crashed:
Screenshot 2020-04-22 at 23 26 18

sudheerDev:
> I was going to test if any code from github has a table (it might have, but i couldn't see it locally) so i logged into the server and tried to add a post with some code block from github and it crashed:
> Screenshot 2020-04-22 at 23 26 18

@Willyfrog That's weird. Does it happen on master? I don't see how I could have caused an issue like that.

Willyfrog:
on daily it doesn't crash, but it deletes previous content. 
Steps on firefox dev edition, macos:
1) open cloud test server
2) copy partially the github linked file from the ticket (all but a couple of lines from the end)
3) write a line, shift enter, write another line
4) position caret at the end of the first line
5) cmd+v

sudheerDev:
@Willyfrog I think it was an issue during the update of spinmints or something. I cannot replicate this locally or on the new spinmint I just spun. Care to give it another try.

Willyfrog:
> @Willyfrog I think it was an issue during the update of spinmints or something. I cannot replicate this locally or on the new spinmint I just spun. Care to give it another try.

whatever it was, it is fixed now ¯\_(ツ)_/¯

components/create_comment/create_comment.jsx
sudheerDev: That was a good suggestion 👍. Thank you

sudheerDev:
@amyblais should this be for 5.23v? I don't see a milestone on the ticket

amyblais:
@sudheerDev It's fine for v5.23 if it's not risky.

PR #5336, opened 2020-04-14
[MM-22828]  Improve the Main Menu popover position/placement

Qovaros:
#### Summary
This PR does two things:

1. Aligns `main menu` and `set status` drop down with user icon.
2. Removes tooltip showing on hover over `main menu` drop down and `set status` icon

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/14228
JIRA: https://mattermost.atlassian.net/browse/MM-22828

#### Screenshots

mattermost


components/main_menu/main_menu.jsx
```diff
@@ -148,6 +148,7 @@ class MainMenu extends React.PureComponent {
                 mobile={this.props.mobile}
                 id={this.props.id}
                 ariaLabel={formatMessage({id: 'navbar_dropdown.menuAriaLabel', defaultMessage: 'main menu'})}
+                customStyles={{left: '-44px', marginLeft: 0}}
```
Qovaros, line 151: It seems to me, like providing styles object is a preferred solution, but wouldn't it be better to give className prop to `Menu` object to keep styles in `.scss` files?

i18n/en.json
```diff
@@ -3231,7 +3231,6 @@
   "sidebar.createPublicPrivateChannel": "Create a public or private channel",
   "sidebar.directchannel.you": "{displayname} (you)",
   "sidebar.leave": "Leave the channel",
-  "sidebar.mainMenu": "Main Menu",
```
Qovaros, line 3234: I'm no sure it should be removed. If so, should appearances in other translations also be removed?

components/status_dropdown/status_dropdown.jsx
```diff
@@ -135,6 +135,7 @@ export default class StatusDropdown extends React.Component {
                 <Menu
                     ariaLabel={localizeMessage('status_dropdown.menuAriaLabel', 'Set a status')}
                     id='statusDropdownMenu'
+                    customStyles={{marginLeft: '6px'}}
```
nevyangelova, line 138: Can we keep the styles in the css files?

i18n/en.json
nevyangelova: Yes we should remove from other translation files if verified that its not used.

Qovaros: I removed them.

components/status_dropdown/status_dropdown.jsx
Qovaros: Done.

components/main_menu/main_menu.jsx
Qovaros: I modified `sass/layout/_headers.scss` instead.

Qovaros:
Also in https://github.com/mattermost/mattermost-webapp/commit/ba7358aa78058553c910fc9cfc4a17172fbe117d `z-index` for `channel-header` got changed to 15 and because of that main menu dropdown displays under channel header. Should it be like that?
Screenshot from 2020-05-05 23-17-00

sass/layout/_headers.scss
```diff
@@ -520,6 +520,12 @@
             .MenuWrapper {
                 flex: 1;
 
+                &.main-menu {
```
bradjcoughlin, line 523: @nevyangelova I wonder if we shouldn't move these types of styles into a `.scss` file next to the component - i.e. `main_menu.scss` like we're doing with the `widgets` components. This would become `.MainMenu` to match the component name.

bradjcoughlin:
> Also in [ba7358a](https://github.com/mattermost/mattermost-webapp/commit/ba7358aa78058553c910fc9cfc4a17172fbe117d) `z-index` for `channel-header` got changed to 15 and because of that main menu dropdown displays under channel header. Should it be like that?
@asaadmahmood thoughts?

asaadmahmood:
@bradjcoughlin Nops, we can increase the z-index for the dropdowns.

Qovaros:
@asaadmahmood @bradjcoughlin would you like me to create a PR for that?

bradjcoughlin:
@Qovaros you can go ahead and commit the `z-index` change you feel appropriate against this PR.

sass/layout/_headers.scss
bradjcoughlin: @Qovaros don't worry about this comment for now - it's just something we're thinking about but haven't made a decision yet 😄

Qovaros:
> @bradjcoughlin Nops, we can increase the z-index for the dropdowns.

I don't think we can fix it by changing dropdown z-index. The problem is [channel-header](https://github.com/mattermost/mattermost-webapp/blob/da27823327902d37543986d6d03decf0d4eb2d3c/sass/layout/_headers.scss#L5) has `z-index: 15`, while [sidebar-left](https://github.com/mattermost/mattermost-webapp/blob/da27823327902d37543986d6d03decf0d4eb2d3c/sass/layout/_legacy-sidebar-left.scss#L11) has `z-index: 12`. I added a commit changing it to 16

Qovaros:
@hanzei so should this PR be merged now?

hanzei:
@Qovaros Indeed, thanks for the reminder. 

@esethna Should this go into `v5.25`?

PR #5337, opened 2020-04-14
[MM-22802] [MM-22801] Dont show createPost until channel roles are loaded

fmunshi:
#### Summary
- Currently reloading the page while on a channel results in the create post text box showing the channel as read only for a split second this is because while the roles are not loaded into the state the `haveIChannelPermission` returns false for the `create_post` permission.

- This PR makes it so that the `createPost` component is not rendered (i.e. the loading indicator is shown) if the channel's roles are not in the state.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22801
https://mattermost.atlassian.net/browse/MM-22802

#### Test steps
- Reload page while on a channel
- Ensure that the text box does not flash as read only
- Ensure that the text box is in focus after reloading
- Repeat test with some channel moderation setting enabled


hahmadia:
Code looks good! Before I approve, are you able to provide a before and after GIF? Thanks!

fmunshi:
Before
Kapture 2020-04-16 at 15 15 13

After 
Kapture 2020-04-16 at 15 18 55

vdvukhzhilov:
Hi, everyone.

Sorry for posting here, but I have a question
Why we break **outside** of `if` statement here https://github.com/mattermost/mattermost-webapp/pull/5337/files#diff-6010dfd3202d738799f6e3e6006ff7e893bfe0be6a516de82537fbdbf96c340aR48 ? Is there a chance that we won't find a needed role in first iteration and break earlier than needed?

lieut-data:
Indeed, at first glance, that code doesn't make sense to me @vdvukhzhilov. @hmhealey, would you be able to help triage this observation?

hmhealey:
Yeah, I agree that that's likely a mistake. Would you mind creating an issue for that, @vdvukhzhilov? Alternatively, we'd appreciate a PR to fix that as well 😄

vdvukhzhilov:
https://github.com/mattermost/mattermost-webapp/pull/11097 here you go

hmhealey:
Awesome, thanks

PR #5339, opened 2020-04-15
Update multiselect string in en.json

justinegeffen:
#### Summary
Updated string for ""multiselect.numPeopleRemaining": "Use ↑↓ to browse, ↵ to select. You can add {num, number} more {num, plural, one {person} other {people}}. "," in line with suggestions in the Jira ticket but slightly edited. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23406

#### Screenshots


esethna:
I think the JIRA ticket proposed "You can add up to 20 new members at a time"?

Using the word "invitation" is confusing to me since it's not an "invitation", the user is being added directly

justinegeffen:
> I think the JIRA ticket proposed "You can add up to 20 new members at a time"?
> 
> Using the word "invitation" is confusing to me since it's not an "invitation", the user is being added directly

Yep, that is what the Jira ticket proposed and I started with that but when applying it to the singular it didn't seem to read well so I went with "invitation" as an alternative. 

For the option suggested in the Jira ticket, using the string below and the suggestion in the Jira ticket, it would read:
"multiselect.numPeopleRemaining": "Use ↑↓ to browse, ↵ to select. You can add up to {num, number} more {num, plural, one {person} other {people}}. "," 

**"You can add up to one more person"** and **"you can add up to 20 more people".** 

One option could be to break that up and have one string for singular "you can add one more person at a time" and one for plural "you can add up to 10 more people at a time". 

That isn't how the rest of the singular/plural strings are written and I don't want to introduce unnecessary complexity. Happy to go with the original suggestion if it works best. :)

justinegeffen:
I gave this some more thought and a less complicated option would be to have a static message of "You can add up to 20 people at a time", cutting out the need for singular and plural. So, disregard my previous comment.

esethna:
> "You can add up to 20 people at a time"

Works for me :)

esethna:
@justinegeffen can you help make the updates discussed above and then move this to dev review? Thanks!

i18n/en.json
```diff
@@ -2917,7 +2917,7 @@
   "multiselect.loading": "Loading...",
   "multiselect.numGroupsRemaining": "Use ↑↓ to browse, ↵ to select. You can add {num, number} more {num, plural, one {group} other {groups}}. ",
   "multiselect.numMembers": "{memberOptions, number} of {totalCount, number} members",
-  "multiselect.numPeopleRemaining": "Use ↑↓ to browse, ↵ to select. You can add {num, number} more {num, plural, one {person} other {people}}. ",
+  "multiselect.numPeopleRemaining": "Use ↑↓ to browse, ↵ to select. You can add {num, number} more {num, plural, one {person} other {people}} in this invitation. ",
```
justinegeffen, line 2920:
```suggestion
  "multiselect.numPeopleRemaining": "Use ↑↓ to browse, ↵ to select. You can add up to {num, number} more {num, plural, one {person} {people}} at a time. ",
```

justinegeffen:
Thanks, @esethna. Update is done; added @jespino as reviewer, let me know if anyone else should be added. Thanks!

justinegeffen:
Thanks everyone for the reviews, looks like this is good to go so I changed it to reviews complete. :)

PR #5341, opened 2020-04-16
[MM-22806] Make search hints interactive

nevyangelova:
#### Summary
Improving the UI of our search bar and making the search hint interactive. Used the design specs from figma, but since there wasn't a sidebar right design I improvised. UX and PM can advise on behaviour and styles. Also fix version is not specified, is this for v.5.23.0? 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22806

#### Screenshots
Screenshot 2020-04-16 at 13 50 02


esethna:
Adding @matthewbirtch for UX review since it's his design

bradjcoughlin:
Though I just noticed its failing `i18n-check`

jgilliam17:
Thank you @nevyangelova 
Adding few observations for UX input:
1. Exclude search term option posts as "dash space" `- ` if user doesn't backspace to clear the space and enters the term, text is not excluded.
 Is it possible to place the blinking cursor right after dash automatically?
2. Similarly for `""`,  blinking cursor is placed after the double quotes, so to include the search term user has to manually move it inside the quotes 
Can we place the blinking cursor inside the double quotes automatically when selecting this option from the dropdown?
3. Clearing the text box using `x` leaves the search box with cursor, but otherwise looking unfocused with missing box border highlight and no visible hint menu

nevyangelova:
@matthewbirtch thanks for the suggestions. I only have a question regarding the space between tag and text. The spacing is 16px at a 2560 x 1600 resolution like it is in the figma design. Unfortunately the figma file didn't show the popover for larger resolutions like the one you are showing in the screenshot. The popover expands with the viewport and its based on the search bar width, however if we keep the space 16px I think it looks weird as the side padding becomes too large and all the text is crammed in the middle. My suggestion is to keep the spaces expanding with the popover width. Also based on the position of the tags, baseline left, the space on each row will be different than 16px since the tags have different length. If they are all spaces between 16 px they wouldn't be aligned left, so which tag do we take as a target?

matthewbirtch:
Thanks for the response @nevyangelova. Here are my thoughts:

1. I think it's more important that the descriptive text feels connected to the tag it relates to. We would just use the longest tag as the reference point for the 16px distance I mentioned (which would be the `BEFORE:` tag). So it would look like this...
image


2. This extra space was intentionally added in the design because the first and last list items need extra room from the edges. I'm totally fine with it looking this way on the hover state - it's more important that it looks right in the default state. This is how the autocomplete are designed as well so we need to be consistent with those (see screenshot below)
image

3. Not sure what I was seeing before....it seems to be fine now.

components/search_hint/search_hint.tsx
```diff
@@ -1,32 +1,84 @@
...
+                        key={option.searchTerm}
+                        onClick={() => props.updateSearchTerms(option.searchTerm)}
+                        onMouseOver={() => setIndex(optionIndex)}
+                    >
+                        <div className='search-hint__suggestion-list__flex-wrap'>
+                            <span className='search-hint__suggestion-list__label'>{option.searchTerm}</span>
+                        </div>
+                        <div className='search-hint__suggestion-list__value'>
+                            <FormattedMessage
+                                id={t(option.message.id)}
```
hmhealey, line 76: `t` needs to receive a string literal for it to be picked up by the i18n tools, so this won't stop them from messing up `en.json`. You'll need to use `t` wherever `props.options` is defined.

nevyangelova:
PR Updated and ready for second review @matthewbirtch @bradjcoughlin @hmhealey @esethna @jgilliam17

components/search_hint/search_hint.tsx
```diff
@@ -1,32 +1,84 @@
...
+    options: SearchTerm[];
 }
 
-export default class SearchHint extends PureComponent<Props> {
-    public render() {
-        return (
-            <React.Fragment>
-                {this.props.withTitle &&
-                <h4>
+export const SearchHint = (props: Props) => {
```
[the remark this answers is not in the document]
hmhealey: Thanks for changing that. I could've sworn you could do `export default const Foo = ...` all on one line, but I guess not. At the very least this component can be imported like it would in other places now.

components/search_hint/search_hint.tsx
hmhealey: @nevyangelova This still needs to be looked at. This is why your search strings aren't being added to `en.json`, and I think it's why there's an entry being created for "undefined"`. You can wrap the IDs defined in the constants with `t` to fix this.

jgilliam17:
Thanks @nevyangelova 
Few more issues

1. Unable to keep focus on Search field, it dismisses if you click on `Members` first followed by click on Search
2. I've experienced No hint popover when clicking on Pinned or Flagged icon first with RHS open, but can't reproduce reliably
3. Unable to dismiss hint popover, Search box styled as focused, but missing the cursor. 
Steps:
- Click on Mentions `@`
- Clear mention text by clicking on the `x` in Search field
- Click outside of the hint popover, you can even place the focus in main channel text box at his point and post text -> hint popover remains visible, Search box styled as focused, missing cursor
Screen Shot 2020-04-27 at 9 26 47 AM

4. Mobile view - hint list that opens does not follow the same behavior. This may be related/mentioned in the previous comment (point 7). Also, popover appeared once or twice layered over the existing hint list, but this I am not able to repro consistently 
Screen Shot 2020-04-27 at 11 54 16 AM

nevyangelova:
PR updated. There might still be some issues with focusing/refocusing as the code for that is a bit all over the place and there might be uncaught cases. Help with testing is much appreciated 🙏

hmhealey:
> FYI in order to not conflict with the mention search since they use the same input and focus functions, the only way I can avoid that is by not showing the popover at all when mention search is present in the search bar or RHS. In order to see it again you need to navigate away from mentions. Maybe devs can advice.

Do you mean that when the `from:` autocomplete is visible, the search hint is disabled? That's what I'd expect, since I think only having 1 popover visible at once is ideal.

components/search_bar/search_bar.jsx
```diff
@@ -155,29 +186,90 @@ export default class SearchBar extends React.Component {
...
+    focus = () => {
+        // This is to allow redux to process the search term change
+        setTimeout(() => {
+            this.search.focus();
+            this.setState({focused: true});
+        }, 0);
+
+        if (this.search.value === '""') {
+            this.search.selectionStart = this.search.length / 2;
+            this.search.selectionEnd = this.search.length / 2;
```
hmhealey, line 213: In this case, shouldn't `this.search.length / 2` always equal 1? In that case, it would be clearer to just put 1 here. It might also be nice to add a comment here saying that we're putting the selection in between the quotes.

utils/constants.jsx
```diff
@@ -1429,6 +1429,15 @@ export const Constants = {
 
 export const AcceptedProfileImageTypes = ['image/jpeg', 'image/png', 'image/bmp'];
 
+export const searchHintOptions = [{searchTerm: 'From:', message: {id: 'search_list_option.from', defaultMessage: 'Messages from a user'}},
+    {searchTerm: 'In:', message: {id: 'search_list_option.in', defaultMessage: 'Messages in a channel'}},
```
hmhealey, line 1433: These still need to be wrapped with `t` to have the i18n tools pick them up like
```suggestion
    {searchTerm: 'In:', message: {id: t('search_list_option.in'), defaultMessage: 'Messages in a channel'}},
```

esethna:
A few things I noticed: @nevyangelova @matthewbirtch 

1. After using two search modifiers the pop-over no longer appears after hitting space. Example:

image

2. Can we lower case the modifiers in the search box? ie "Before" > "before", "In" > "in"

3. If you you play with it for a bit I can sometimes get it into a state where the pop-over doesn't appear anymore, even if I fully clear the search input box. See image, I don't have clear repro steps though 

image

4. Can we have the pop-over disappear immediately when the user clicks outside of it? Compared to @-autocomplete in the post text box, this pop-over takes a lot longer to close after clicking out of it. If we have a transition on it I'd propose we cut it, it's making the app feel laggy. 

https://i.gyazo.com/23599e67660b4ea665f1d2a7e188edba.gif

nevyangelova:
Thank you all for the great feedback! I have taken into account and implemented all steps, apart from the point regarding showing the popup 4 times. In my search experience as a user, seems like twice is optimal, if I get the popup displayed more than that it seems annoying because most of the time I wouldn't need all the search criteria for every single search. My suggestion is that we keep it at two (or max 3) times per search. I am curious about your reasoning however and will implement as many as needed. So far while testing I didn't get any issues with focusing, let me know if there is anything else I can do! @matthewbirtch @esethna @jgilliam17 @hmhealey

esethna:
+1 so awesome! Adding PM approval, @matthewbirtch please help rereview and approve if you're happy with it. We can probably start getting some dev review eyes on it.

esethna:
@bradjcoughlin probably worth doing a re-review given the changes since you last reviewed. @devinbinnie feel free to review now

bradjcoughlin:
@nevyangelova I can take a look at the `i18n` issue tomorrow.

jgilliam17:
Looks great @nevyangelova 
few issues
1. When "" is used as a second hint, cursor is outside the quotes 
Screen Shot 2020-05-07 at 3 21 53 PM
2. Mobile web view search only allows one hint to be used at a time e.g. search for `from: sysadmin` when adding second hint it replaces first

One hint at a time

3. Mobile web view search is stuck on loading, without the hint list (same when selecting recent mentions and flagged posts) I am able to add search terms, prompt calendar modal to show up and from: user list etc. but not able to get out of Searching spinner

esethna:
@hmhealey should we unassigned you from this PR a reassign to another dev?

nevyangelova:
@jgilliam17 fixed the cursor issue. For the mobile view we decided to keep the functionality to only work on one click (as observed) for now.

jgilliam17:
Thank you @nevyangelova
Looks great! 1 issue (+ one for later investigation)
1. When the hints are part of the RHS it doesn't follow the same behavior as the popover, hints are capitalized and focus is missing in the text box when you select any hint. 
Screen Shot 2020-05-12 at 5 03 28 PM

- Popover is missing all together in some instances. Per discussion with Eric, I am going to create a follow up ticket for this. I've seen that a few times, but have not been able to narrow down the steps to repro. For now refresh gets me out of that weird state when it happens, and it is occurring rarely so that's good.

nevyangelova:
Thanks @jgilliam17 yeah exactly that weird issue you are seeing is related to the regression that happened on master with the search in RHS. Alejandro is working on a fix so I think we can merge this and the fix will be applied in his PR. I will keep a close track of this to make sure its working flawlessly. FYI @esethna

PR #5344, opened 2020-04-16
[MM-23605] Reload team memberships when team scheme is updated

fmunshi:
#### Summary
- Assigning (or un-assigning) a scheme to a team requires reloading the team membership in order to get the new role name to load permissions from.
- `UPDATED_TEAM` which was previously being used only reloads the team in the state and not the memberships, so I created a new websocket event for team scheme being updated on a team

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-23605

#### Related Pull Requests
- Has server changes: https://github.com/mattermost/mattermost-server/pull/14305


actions/websocket_actions.jsx
```diff
@@ -645,6 +649,10 @@ function handleUpdateTeamEvent(msg) {
     dispatch({type: TeamTypes.UPDATED_TEAM, data: JSON.parse(msg.data.team)});
 }
 
+function handleUpdateTeamSchemeEvent(msg) {
```
fmunshi, line 652:
```suggestion
function handleUpdateTeamSchemeEvent() {
```

fmunshi:
@srkgupta this needs to be tested with the corresponding server changes here https://github.com/mattermost/mattermost-server/pull/14305

PR #5347, opened 2020-04-17
Adds support for TypeScript syntax highlighting

Nirei:
#### Summary
Activates the, already latent, TypeScript language syntax highlighting capabilities provided by `highlighting.js` dependency.

#### Ticket Link
There is no issue or ticket addressing this. I happen to use both mattermost and typescript at work and I think it'd just make sense to provide highlighting to the language in which mattermost's own webapp is being written. :)

#### Related Pull Requests
N/A

#### Screenshots
Screenshot of highlighted TypeScript


utils/constants.jsx
```diff
@@ -1319,6 +1319,7 @@ export const Constants = {
         swift: {name: 'Swift', extensions: ['swift']},
         tex: {name: 'TeX', extensions: ['tex'], aliases: ['latex']},
         text: {name: 'Text', extensions: ['txt', 'log']},
+        typescript: {name: 'TypeScript', extensions: ['ts'], aliases: ['ts', 'tsx']},
```
hmhealey, line 1322: If we're including `tsx` as an alias, does this also support highlighting TypeScript with React embedded in it? If so, we should also include `tsx` as an extension so that `tsx` files are highlighted when viewed as file attachments.

Nirei: That sounds completely reasonable. :grin:

bradjcoughlin:
Thanks @Nirei , I was looking for Typescript highlighting the other day. 🙌

PR #5351, opened 2020-04-18
[GH-13718] Migrate rhs_card component and its tests to typescript

abdusabri:
#### Summary
Migrates `rhs_card` component and its related tests to TypeScript

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/13718


components/post_profile_picture/post_profile_picture.jsx
```diff
@@ -14,15 +14,16 @@ import * as Utils from 'utils/utils';
 export default class PostProfilePicture extends React.PureComponent {
     static propTypes = {
         compactDisplay: PropTypes.bool.isRequired,
-        enablePostIconOverride: PropTypes.bool.isRequired,
-        hasImageProxy: PropTypes.bool.isRequired,
+        enablePostIconOverride: PropTypes.bool,
+        hasImageProxy: PropTypes.bool,
```
abdusabri, line 18: The component was already used without providing `enablePostIconOverride` and `hasImageProxy` props. Like:
```javascript
const avatar = (
            <PostProfilePicture
                compactDisplay={false}
                post={selected}
                userId={selected.user_id}
            />
        );
```
So, had to make them optional because TypeScript was complaining about this 🙂

components/post_profile_picture/post_profile_picture.jsx
```diff
@@ -14,15 +14,16 @@ import * as Utils from 'utils/utils';
...
         isBusy: PropTypes.bool,
         isRHS: PropTypes.bool,
         post: PropTypes.object.isRequired,
         status: PropTypes.string,
         user: PropTypes.object,
         isBot: PropTypes.bool,
         postIconOverrideURL: PropTypes.string,
+        userId: PropTypes.string,
```
abdusabri, line 26: Was provided to the component when used in `rhs_card`, I've added it as optional to please the TypeScript compiler 🙂

types/store/plugins.ts
```diff
@@ -38,7 +39,7 @@ export type PostPluginComponent = {
     id: string;
     pluginId: string;
     type: string;
-    component: React.Component;
+    component: React.ComponentType<{post: Post}>;
```
abdusabri, line 42: When a react component is used like in the following code snippet: 

```javascript
if (pluginPostCardTypes.hasOwnProperty(postType)) {
            const PluginComponent = pluginPostCardTypes[postType].component;
            content = <PluginComponent post={selected}/>;
        }
```

TypeScript is looking for a type not a constructor function, so it was giving an error that the `PluginComponent` variable `does not have any construct or call signatures`. So, had to change it.

Here are couple of links for reference:

https://github.com/microsoft/TypeScript/issues/28631

https://stackoverflow.com/questions/31815633/what-does-the-error-jsx-element-type-does-not-have-any-construct-or-call

abdusabri:
Closing this PR to allow another contributor to work on it. More details are provided in a comment on the issue ticket.

PR #5353, opened 2020-04-19
MM-22894/MM-23046/MM-24274/MM-24375 Quick wins bucket list 5

asaadmahmood:
#### Summary
Quick wins bucket list 5
Escape to close the expanded RHS is not in, as there are a lot of corner cases around that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22894
https://mattermost.atlassian.net/browse/MM-23046
https://mattermost.atlassian.net/browse/MM-24375

#### Screenshots
localhost_8065_mattermost_channels_town-square
Screenshot 2020-04-19 at 2 58 17 PM


esethna:
Nice!

1. We can probably remove this text now:
image

2. Channel name in the RHS header is missing a click effect

3. Do we want to include the changes to the "Add comment" button in this PR?

4. 0/5 on the changes to the RHS expanded state in terms of the UI? If you think we can do better, feel free to change what was written in the ticket

5. Can we allow "ESC" to collapse the expanded RHS?

asaadmahmood:
1. Done
2. Done
3. We should do that globally in a separate PR.
4. I am concerned about the header being white, that doesn't look good to me, but lets see what others think, and then the full opacity channel header at the top not vertically aligning with the full opacity RHS below it. I'll see what others think before making any changes.
5. That may have a bunch of corner cases, so I would avoid doing that in this PR.

asaadmahmood:
@matthewbirtch I did update that, both the close button and the expand button aren't there, and we don't have a close button because the RHS is closed in the mobile web view with the back button at the top beside the search bar.

The PR just did not update automatically, which is why you can't see it right now.

asaadmahmood:
@MathewtheCoder 
1. Cool.
2. Done.
3. Wasn't a fan of that, so removed it.
4. Done. Good idea.
5. As mentioned above.

matthewbirtch:
Thanks @asaadmahmood. I think you may have tagged the wrong github user in your comment above :)

Things are looking good. However, I noticed now that we've added a max-width on the expanded state of the RHS, I think it messed up the animation :(

asaadmahmood:
@matthewbirtch Will fix that.

esethna:
>Do we want it to close the sidebar? We already have another way to close the sidebar on mobile. For most cases, a user will already be in the current channel, unless he is jumping from the search bar, in which case, he can jump to the channel from the search results, or we can also add Jump to be very explicit. I don't want it to function in different ways on mobile and desktop, with it closing it on mobile, but not on the desktop app. Especially since its not clear that clicking it will close the sidebar and you will lose the thread you were viewing. You would have to either search for it again from search, or manually search for it in the channel.

@asaadmahmood I think either we keep the button and it should close the sidebar and take the user to that channel or we just make it not clickable in the mobile view. 0/5

asaadmahmood:
@esethna I feel like we can make it not clickable. I don't want people to accidentally click it and lose the RHS, as it does not close it right now.
In the case of the user opening the reply thread from a post in the center channel, he would always remain in that channel, as one can't easily navigate to a different channel with the RHS open, so I think its fine.

esethna:
@matthewbirtch @asaadmahmood 0/5, but if we are adding a max width to the RHS exoand state, which is a great idea, do we need the channel sidebar to have an overlay? I'm imagining that if a user switches channel we can just switch the channel and shrink the RHS back to normal size?

asaadmahmood:
@esethna Removed the click on mobile.
And for the RHS overlay. It's a good way to collapse it right now. If we remove that, then the user would have to manually collapse it, or redirect himself to a different channel to do so.

Plus, if we are removing the overlay, then I would imagine we should also squish the center channel based on the space available, so the user can interact with it, but that may pose challenges on most normal browser screen as the width would be quite small.

esethna:
@asaadmahmood I'm referring to removing the overlay on the channel sidebar, not the center pane

asaadmahmood:
@esethna Hmmm, yeah I think we can try that.

asaadmahmood:
Or maybe it may not look as good. It would look like this.

asaadmahmood:
With the overlay, its clear that clicking the sidebar would collapse it, but without it, its not as clear.

asaadmahmood:
Also, having the sidebar collapse on channel switch might be something we can consider for a different ticket, as that would be adding some extra functionality on a global level, so the channel sidebar can close it.

utils/utils.jsx
```diff
@@ -733,9 +732,9 @@ export function applyTheme(theme) {
         changeCss('.app__body .post .post-collapse__show-more-line', `background-color:${changeOpacity(theme.centerChannelColor, 0.1)}`);
 
         if (theme.centerChannelBg) {
-            const ownPostBg = blendColors(theme.centerChannelBg, theme.centerChannelColor, 0.05);
-            const hoveredPostBg = blendColors(theme.centerChannelBg, theme.centerChannelColor, 0.08);
-            const hoveredPostBgLight = blendColors(theme.centerChannelBg, theme.centerChannelColor, 0.05);
+            const ownPostBg = theme.centerChannelBg;
```
hmhealey, line 735: Instead of just changing the colour here, could you remove the logic that refers to `ownPostBg`? This part of the code is really complicated, so removing that case would simplify it a fair bit.

andrewbrown00:
@asaadmahmood things look great here 🎉 

**A few items to resolve/consider**:

- [ ] We should use the permalink highlight pattern after a user clicks on the `channel` name in the **Thread** header. Right now, clicking the `channel` name in the **Thread** header opens the channel (unless it's already open); but we don't provide the user with an affordance to where the **thread** is in the channel.


- [ ] 0/5 and open to feedback, but something feels off with the timing when clicking to expand the RHS. I think it's because the overlay comes in faster than the RHS slides out. I like the animation on the RHS but wonder if we can tie them together?


- [ ] 0/5 but something feels off with leaving a sliver of the center channel showing when the RHS is expanded; it's create a stepped/staggered visual. **Maybe it won't appear this way once the overlay is removed from the LHS.**
image

- [ ] @esethna should we add tooltips to the `expand` and `close` icons. We need to define the rules for when we do/do not want to introduce tooltips for icon buttons.
image  image

- [ ] The overlay sits on top of the **Thread** view when reducing the view down to mobile.

esethna:
@andrewbrown00 

That's a really great idea to permalink to the root post in the center pane - looks like that's what slack does also. We should do this for sure.

Re tooltips, already have a ticket completed for that by community: https://mattermost.atlassian.net/browse/MM-22876

_______

@asaadmahmood I don't mind this but I'm 0/5, I've leave it up to you guys and @andrewbrown00 on what's the best UI for the expanded state. Wonder if we can find some examples in other products

> localhost_8065_mattermost_channels_this-is-a-really-long-channel-name-like-really-really-long

esethna:
Also @asaadmahmood FYI on this: https://mattermost.atlassian.net/browse/MM-24274

asaadmahmood:
@andrewbrown00 
1. Are you saying we just have the permalink styling, or do you want the channel name linked to the post?
2. I'll see if I can tie it together, though there was an issue with that as one's fading, the other is sliding.
3. It still does feel that way, as shown in the screenshot I posted.
4. Those are done (not in this PR).
5. On mobile/tablet view, the RHS is positioned on top of the header, so there is no reason to have it appear below, as a person can rarely click anything in the header.

matthewbirtch:
@asaadmahmood as per our chat, I noticed a couple more things:
1. For the collapsing animation of the expanded RHS, there is a flash that happens over the RHS. It looks like the overlay is coming on top of the RHS during the animation?
2. Can we change the 'x' icon on the search field to the filled x icon?
image

andrewbrown00:
> @andrewbrown00
> 
> 1. Are you saying we just have the permalink styling, or do you want the channel name linked to the post?
> 2. I'll see if I can tie it together, though there was an issue with that as one's fading, the other is sliding.
> 3. It still does feel that way, as shown in the screenshot I posted.
> 4. Those are done (not in this PR).
> 5. Should be fixed.

1. Yup, link the channel name to the parent post in the center channel; **looks like you already did it ✅ **
2. Feels much smoother and more connected
3. Visually I like it without the overlay on the sidebar. I also agree with Eric's previous feedback that if the LHS doesn't have the overlay then users can use the LHS to switch channels which would close the expanded RHS. I'm 4/5 it feels like an unnecessary barrier to block the LHS in this context. 

I stubbed out version using an elevation instead of the overlay @asaadmahmood @esethna @matthewbirtch thoughts?



4. Thanks for confirming
5. Yup, looks good

asaadmahmood:
@andrewbrown00 @matthewbirtch Not really a fan of the elevation, also, do we no longer collapse the sidebar when we click the center channel (if we do, it would no longer be obvious).

matthewbirtch:
To me the elevation solution it's way more elegant than dimming out only a small part of the background. Dimming out only a small area actually makes it too stark and brings more attention to the dimmed area than we want. 

With the elevation solve, clicking anywhere on the center channel would still collapse the RHS. Clicking on a channel in the sidebar would switch to that channel and collapse the RHS. To me, this is a pretty solid solve. What is that is not working for you?

andrewbrown00:
Thanks @asaadmahmood the elevation on the RHS feels really good to me; I say we go with it and let it soak on daily. 

**One change**, let's also collapse the expanded RHS when interacting with the LHS and switching channels and direct or group messages using any of the current switching method, i.e. clicking to switch, keyboard shortcuts, channel switcher, switching by browsing or creating new channels or starting a new direct or group message.

As for the `clear` icon, I prefer to see it filled, IMO it sets it apart from the other icons and is more obvious when it pops in, especially since it's ephemeral, even though the outline version matches the rest of the icons. I'll leave this one to a vote 😉

image

asaadmahmood:
Done @andrewbrown00

andrewbrown00:
Looks awesome @asaadmahmood 🎉

esethna:
@andrewbrown00 can you help update the JIRA ticket for the expanded RHS updates so QA can reference it for testing?

esethna:
@nevyangelova @bradjcoughlin please go ahead with dev review

sass/layout/_sidebar-right.scss
```diff
@@ -2,35 +2,43 @@
...
     .sidebar--right__bg {
-        @include single-transition(background-color, .5s, ease);
-        background-color: alpha-color($black, 0);
+        background-color: transparent;
         height: 110%;
         position: absolute;
         top: -70px;
         visibility: hidden;
-        width: 300%;
+        width: 1000%;
```
nevyangelova, line 34: is there a reason for these values? Maybe a comment to explain?

asaadmahmood:
Done @nevyangelova

components/channel_header/channel_header.js
```diff
@@ -124,6 +125,10 @@ class ChannelHeader extends React.PureComponent {
         if (header !== prevHeader) {
             this.props.actions.getCustomEmojisInText(header);
         }
+
+        if (this.props.channel.id !== prevProps.channel.id) {
+            this.props.actions.setRhsExpanded(false);
```
hmhealey, line 130: This should be done in the SidebarRight component since that's the one that cares about whether or not the sidebar is expanded or not. It's likely to get lost if it's left here since others wouldn't expect to find sidebar logic here.

components/rhs_root_post/rhs_root_post.jsx
```diff
@@ -229,25 +229,13 @@ class RhsRootPost extends React.PureComponent {
     };
 
     render() {
-        const {post, isReadOnly, teamId, channelIsArchived, channelType, channelDisplayName} = this.props;
+        const {post, isReadOnly, teamId, channelIsArchived} = this.props;
```
hmhealey, line 232: Since we're no longer using `channelType` or `channelDisplayName`, can you remove both of those from the props and make sure they aren't passed in from `mapStateToProps` as well?

components/rhs_header_post/rhs_header_post.jsx
```diff
@@ -145,13 +163,21 @@ export default class RhsHeaderPost extends React.Component {
...
                         id='rhs_header.details'
-                        defaultMessage='Message Details'
+                        defaultMessage='Thread'
                     />
+                    {channelName &&
+                        <button
+                            onClick={this.handleJumpClick}
+                            className='style--none sidebar--right__title__channel'
+                        >
+                            {channelName}
```
hmhealey, line 173: Is it intentional that this label will change from "Direct Message" to the other user's name in a DM channel? If so, it's not consistent with the label that appears on search results in this same channel
image
image
If you did want to change the channel name displayed in the search result header, that comes from the `SidebarRight` component.

sass/utils/_mixins.scss
```diff
@@ -40,3 +40,14 @@
         max-height: $line-height * $lines * 1px;
     }
 }
+
+@mixin linear-gradient($top, $bottom){
```
hmhealey, line 44: 0/5 on whether we keep this, but we don't really need this mixin since every browser we support also supports linear-gradient according to [this](https://caniuse.com/#feat=mdn-css_types_image_gradient_linear-gradient)

asaadmahmood: Cool, I can just use it as it is and remove it.

components/rhs_root_post/rhs_root_post.jsx
asaadmahmood: Will do.

components/channel_header/channel_header.js
asaadmahmood: Cool. I'll have a look at it.

components/rhs_header_post/rhs_header_post.jsx
asaadmahmood: I think we can make it consistent and have the user's name appear.
So for this ticket, as its touching only the sidebar, I think we can have the user's name in the header, similar to the channel names, and create a new ticket to change it for other search related items.

Screenshot 2020-04-27 at 6 41 40 PM

asaadmahmood: @esethna ^

asaadmahmood: I'll update the Pinned posts thing with the DM/GM name instead of the generic Direct Messages for now to be consistent.

components/sidebar_right/index.js
```diff
@@ -22,12 +25,12 @@ import SidebarRight from './sidebar_right.jsx';
 
 function mapStateToProps(state) {
     const rhsState = getRhsState(state);
-
+    const channel = getCurrentChannel(state) || {};
```
hmhealey, line 28: We can't be possibly returning object literals like `{}` from this function since that can cause `SidebarRight` to re-render constantly if the current channel is empty. You're already null-checking `channel` in `SidebarRight`, so this can just be replaced with
```suggestion
    const channel = getCurrentChannel(state);
```
safely.

asaadmahmood:
Done @hmhealey

lindy65:
Thanks @asaadmahmood - pshew, this is a hectic PR to review - so many comments! ;)

I tested via the Jira tickets linked at the top of this PR, our DM where you gave me the high level changes and also from design and PM comments above. Hope I did not miss anything!

**1. In RHS thread view**, the "<" (back arrow) is misaligned with "Thread". This is visible after a search is performed and "reply" is hit on one of the search results.

image

**2. In RHS thread view**, where the channel name is clickable:
    - This doesn't seem very "obvious" to me as a user (there is a grey hover state and mouse pointer changes to a hand but just looking at the channel name doesn't say to me "hey, you can click here to open the channel in the center pane").  Is it possible to show the channel name in a "link colour" (e.g. blue) so it looks like a link you can click on (both on hover and without hover over)? Similar to the "Help" link below "Add a comment" in this screenshot...

image

**3. In expanded RHS view:**
    - I do not see a drop shadow on the LHS bar?
    - The top left corner is not rounded (but not sure whether this was implemented as it has a question mark next to it on the [Jira ticket](https://mattermost.atlassian.net/browse/MM-23046)
    - ESC is not working (to close the expanded RHS)

image

asaadmahmood:
@lindy65 
1. Fixed.
2. I agree, however, with the pattern we are developing, with the icons in the header, or the post menu icons, it is consistent. Gray, gray and background on hover, and then button bg on click and hover. So we'll probably keep it that way, and maybe add an icon next to the name in the future.
3. 
- There isn't supposed to be, we removed it.
- We went with not rounding it off.
- Yeah, that was not done in this PR.

PR #5357, opened 2020-04-20
[MM-24356] Dispatch getChannelMemberCountsByGroup if timezones are enabled when user is updated

fmunshi:
#### Summary
- Builds off of https://github.com/mattermost/mattermost-webapp/pull/5331 look at latest commit for my changes
- Dispatch `getChannelMemberCountsByGroup` when a user updates is called in order to ensure timezone counts are accurate since a user may change their timezone resulting in the confirmation modal displaying out of date information if the page hasnt been reloaded in a while.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24356

### DIFF
--- a/actions/websocket_actions.jsx
+++ b/actions/websocket_actions.jsx
@@ -836,15 +836,22 @@ export async function handleUserUpdatedEvent(msg) {
     const currentUser = getCurrentUser(state);
     const user = msg.data.user;
 
-    if (isGuest(user)) {
+    const config = getConfig(state);
+    const isTimezoneEnabled = config.ExperimentalTimezone === 'true';
+    const userIsGuest = isGuest(user);
+    if (userIsGuest || isTimezoneEnabled) {
         let members = getMembersInCurrentChannel(state);
         const currentChannelId = getCurrentChannelId(state);
-        if (members && members[user.id]) {
-            dispatch(getChannelStats(currentChannelId));
-        } else {
+        if (!members || !members[user.id]) {
             await dispatch(getChannelMember(currentChannelId, user.id));
             members = getMembersInCurrentChannel(getState());
-            if (members && members[user.id]) {
+        }
+
+        if (members && members[user.id]) {
+            if (isTimezoneEnabled) {
+                dispatch(getChannelMemberCountsByGroup(currentChannelId, true));
+            }
+            if (isGuest(user)) {
                 dispatch(getChannelStats(currentChannelId));
             }
         }
--- a/actions/websocket_actions.jsx
+++ b/actions/websocket_actions.jsx
@@ -842,12 +842,14 @@ export async function handleUserUpdatedEvent(msg) {
     if (userIsGuest || isTimezoneEnabled) {
         let members = getMembersInCurrentChannel(state);
         const currentChannelId = getCurrentChannelId(state);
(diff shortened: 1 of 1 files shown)


actions/websocket_actions.jsx
```diff
@@ -825,15 +826,22 @@ export async function handleUserUpdatedEvent(msg) {
...
+    const config = getConfig(state);
+    const isTimezoneEnabled = config.ExperimentalTimezone === 'true';
+    const userIsGuest = isGuest(user);
+    if (userIsGuest || isTimezoneEnabled) {
         let members = getMembersInCurrentChannel(state);
         const currentChannelId = getCurrentChannelId(state);
-        if (members && members[user.id]) {
-            dispatch(getChannelStats(currentChannelId));
-        } else {
+        if (!members || !members[user.id]) {
```
hahmadia, line 835: Could we possibly do something like 
```
const memberExist = members && members[user.id]
if (!memberExist) { ... }
if (memberExist) { ... } 
```
Just thought it may be clearer. What do you think?

fmunshi: done

actions/websocket_actions.jsx
```diff
@@ -825,15 +826,24 @@ export async function handleUserUpdatedEvent(msg) {
     const currentUser = getCurrentUser(state);
     const user = msg.data.user;
 
-    if (isGuest(user)) {
+    const config = getConfig(state);
+    const isTimezoneEnabled = config.ExperimentalTimezone === 'true';
+    const userIsGuest = isGuest(user);
+    if (userIsGuest || isTimezoneEnabled) {
```
mkraft, line 832: I don't know the background: why is this branching statement only for guests?

components/create_post/create_post.jsx
```diff
@@ -551,21 +560,50 @@ class CreatePost extends React.PureComponent {
...
+                        const mappedValue = channelMemberCountsByGroup[group.id];
+                        if (mappedValue && mappedValue.channel_member_count > Constants.NOTIFY_ALL_MEMBERS && mappedValue.channel_member_count > memberNotifyCount) {
+                            memberNotifyCount = mappedValue.channel_member_count;
+                            channelTimezoneCount = mappedValue.channel_member_timezones_count;
+                        }
+                        return `@${group.name}`;
+                    });
+            }
+        }
+
```
mkraft, line 589: There seems to be a hefty amount of logic in both create_post.jsx and create_comment.jsx, which makes sense but my only callout would be to consider if there's any opportunity for reuse?

actions/websocket_actions.jsx
fmunshi: Im not sure what the background was either but it just so happened I needed the same logic for `getChannelMemberCounts...` that was being done for guest users

components/create_post/create_post.jsx
fmunshi: Agreed @mkraft  - but this PR never actually had changes to that file that was from Hosseins previous PR that this was built off of. Maybe we can make a ticket to cleanup and reduce duplication between the two files

mkraft: Yeah for sure. I was thinking this was new code. NVM then.

PR #5358, opened 2020-04-20
MM-24379: Adds create post permission to UI.

mkraft:
#### Summary

Adds the `create_post` permission to the UI.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24379

#### Screenshots

Screen Shot 2020-04-20 at 3 18 39 PM
Screen Shot 2020-04-20 at 3 24 19 PM


fmunshi:
I think this probably needs UX review re: what the permission description should be?

mkraft:
> I think this probably needs UX review re: what the permission description should be?

I had added @thefactremains for that, but either way.

srkgupta:
@thefactremains currently when Create Posts permission is removed from System Scheme, user is not able to create a DM/GM as well. Can you please confirm if this is expected and is it fine? 

@mkraft can you please update the permission description as mentioned by @thefactremains, i.e. 

> Allow guests to create posts.
Allow users to create posts.

srkgupta:
Also @mkraft, can you please add an E2E Cypress test which verifies this new permission: 
- New permission displayed on System Schemes and save and reload should work fine. 
- Member should not be able to create post if the permission is disabled.

mkraft:
> Allow guests to create posts.
> Allow users to create posts.

Currently the code doesn't support role-specific descriptions for permissions shared across multiple roles. Can you please rephrase to be neutral to the role? You can add a ticket to support it if you'd like.

thefactremains:
> Allow users to create posts.

works for me

mkraft:
@amyblais Are you planning to include this in a dot release or should I move it to a later version?

amyblais:
@mkraft How urgent is this and how does it affect customers? Does this resolve the customer's issue?

mkraft:
> @mkraft How urgent is this and how does it affect customers? Does this resolve the customer's issue?

There are currently no known issues that require this change. 

However, if there were to be issues related to instances with roles that do not have the `create_post` permission, this this change would provide customers with a means to diagnose and solve the problem. 

This change would only be put in place in an abundance of caution.

amyblais:
On further discussion with a PM, if we have no other reports and this doesn't fix the customer issue, we can hold off on a dot release until at least one of those occur. If there is another issue we need to fix for a v5.22 dot release, we can probably include this.

e2e/cypress/integration/enterprise/system_console/system_scheme_permission_spec.js
```diff
@@ -0,0 +1,277 @@
...
+    before(() => {
+        // * Check if server has license
+        cy.requireLicense();
+
+        // Reset permissions in system scheme to defaults.
+        resetPermissionsToDefault();
+
+        deleteExistingTeamOverrideSchemes();
+    });
+
```
srkgupta, line 157: Can you please add an after method to resetPermissionsToDefault to ensure that even if this test fails anywhere in between, the system scheme permissions are reset to defaults and other tests are not affected. 

```
after(() => {

resetPermissionsToDefault();

});

File: e2e/cypress/integration/enterprise/system_console/team_scheme_permission_spec.js [hunk 2]
```diff
@@ -117,62 +134,115 @@ const resetPermissionsToDefault = () => {
...
 
         // Reset permissions in system scheme to defaults.
         resetPermissionsToDefault();
 
+        // Delete any existing team override schemes
         deleteExistingTeamOverrideSchemes();
+
+        // # Visit the permissions page
+        cy.visit('/admin_console/user_management/permissions');
     });
```
srkgupta, line 157: Similarly in this test as well, can you please add an after method to resetPermissionsToDefault()

PR #5361, opened 2020-04-21
[MM-29881] Add ability to zoom in/out on PDFs

shred86:
#### Summary
Adds the ability to zoom in and out on PDFs. This required a change in how `ViewImageModal` is initially rendered. Instead of displaying based on the size of the PDF (or other supported media), it now opens to `height: 90vh` and `width: 90:vw`. This was to prevent the modal from continuously expanding and shrinking when the zoom in/out functionality was used. Additional notes:
- Implemented most of the "core" zoom controls and methods to include a new `scale` state  in `ViewImageModal` so it can be applied to other types of media if desired.
- The zoom in/out is occurring by adjusting the scale factor in `getViewport` from PDF js. The reason I took this approach was to provide a higher quality PDF when you zoom in. There's some PDFs that if you try to zoom using the initial render size/scale factor, it's very blurry and hard to read. The other implementation option is to just render the PDF to a higher quality initially, then use pure CSS for the zoom in/out functionality. This would make it easier to implement with other media types (I think).
- Refactored some of the code to display the buttons for consistency and to make it a bit cleaner.
- Icons were made by me. I'm not a graphic designer by any means, so I'm open to change it to anything.

One minor "issue" where the horizontal scroll bar doesn't show up unless you scroll all the way to the bottom. I've spent the afternoon trying to figure this one out but will continue to investigate. I'm assuming it's just a styling issue.
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29881

#### Related Pull Requests
N/A

#### Screenshots
Video Clip: https://imgur.com/3JieFTx


components/view_image/view_image.jsx
```diff
@@ -196,13 +201,45 @@ export default class ViewImageModal extends React.PureComponent {
...
         this.setState({showCloseBtn: false});
+
+        if (fileInfo && fileInfo.extension && fileInfo.extension === FileTypes.PDF) {
+            this.setState({showZoomBtn: false});
+        }
     }
 
+    handleZoomIn = () => {
+        let newScale = this.state.scale;
+        newScale = (newScale * DEFAULT_SCALE_DELTA).toFixed(2);
```
shred86, line 225: I initially pulled this from a repository that implemented some zoom functionality for PDF JS. The more I look at how they’re handling the scale factor, the less I understand why they made something so simple, so complicated. I’m going to refactor this to simply:

```
const newScale = Math.min(this.state.scale + DEFAULT_SCALE DELTA), MAX_SCALE);
```

esethna:
This is sweet! Adding @andrewbrown00 to do a UX review

shred86:
Noticing some errors in the console when you try to click the zoom in or out button very quickly. Occurs more frequently with more complex PDFs:
```
pdf.js:9770 Uncaught (in promise) Error: Cannot use the same canvas during multiple render() operations. Use different canvas or ensure previous operations were cancelled or completed.
```

components/view_image/view_image.jsx
```diff
@@ -312,9 +346,30 @@ export default class ViewImageModal extends React.PureComponent {
...
+            );
+        }
+
+        let zoomButtons;
+        if (this.state.showZoomBtn) {
+            zoomButtons = (
+                <div>
+                    <div
+                        className='modal-zoom-in'
+                        onClick={this.handleZoomIn}
```
shred86, line 365: Perhaps a possible solution is using `debounce` from lodash to set a min time between clicks to call the function as well as a max wait time to call. Something like:

`onClick={debounce(this.handleZoomIn, 300, {maxWait: 300})}`

However, the errors I mentioned in [my comment]() still exists for larger file sized PDFs, so I'm not sure if this is the best solution.

shred86:
Thanks for the feedback!

**Items to resolve**
1. Not sure what's going on with the initial render where it's only showing the first page. Will need to troubleshoot this.
2. This should be easy to resolve.
3. Public links was just turned off in that Mattermost instance by default. I just turned it on.

**Recommend improvements**
1. I see the `fonts` directory but the only two icons I'm seeing in mattermosticons.ttf is a pin and clock looking icon. Where can I find the +, - and magnifying glass icons you have in your prototype?
2. Reset zoom will be easy. Fit to zoom, I'll have to mess around with this.
3. I'll see what I can do about the disable/dimming at max zoom out. I think users should be able to zoom out past the default scale. The reason being is PDFs render to a scale of 1 (this is the current behavior). If the PDF resolution is larger than the modal, then you will want to zoom out in some cases. Now, some logic could be put in place to maybe check to see if the PDF resolution is larger than the modal view to set a different default scale value, but that may have to be a separate PR.
4. Same as above. Styling stuff should be do-able.
5. I should be able to do that.

This will likely take me some time but I'll get as far as I can. I'm hoping someone will have some tips on how to resolve [this issue](https://github.com/mattermost/mattermost-webapp/pull/5361#discussion_r412666824).

esethna:
Hi @shred86 let us know if you need some help addressing the items identified

esethna:
> I see the fonts directory but the only two icons I'm seeing in mattermosticons.ttf is a pin and clock looking icon. Where can I find the +, - and magnifying glass icons you have in your prototype?

@andrewbrown00 can you help answer this?

andrewbrown00:
Hi @shred86, are you able to see the [fonts in this directory](file:///Volumes/GoogleDrive/Shared%20drives/Marketing/Design/Product%20Design/Mattermost%20Design%20Guide/Icons/demo.html)? You can copy and paste the glyph or `check` the **show codes** option to get the hex values. It should look like this:

Glyphs | Codes
----- | -----
image | image



> > I see the fonts directory but the only two icons I'm seeing in mattermosticons.ttf is a pin and clock looking icon. Where can I find the +, - and magnifying glass icons you have in your prototype?

shred86:
@andrewbrown00 I'm not sure which directory you're referring to. I'm looking at the [fonts/mattermosticons](https://github.com/mattermost/mattermost-webapp/tree/master/fonts/mattermosticons) on this repository. Can you provide me a link to the pictures of the mattermost icons you posted?

Also, do I need to create a new jsx file for these icons individually in the components/widgets/icons directory or can I just use this mattermosticon font type directly?

shred86:
Just made a commit with where I'm at. I've made most of the changes requested from the UX review but here's what's left:

- Need to update icons using mattermosticons (currently just using regular text as a place holder).
- Determine a solution for the [issue](https://github.com/mattermost/mattermost-webapp/pull/5361#issuecomment-617515676) I mentioned above. Using `debounce` helps, but with very large PDFs, it still occurs.
- Update Download to read more pages button.
- Add a "Fit to width" option.

I should be able to take care of the first one if I can figure out how to access the mattermost icons. The second option I'll need some feedback/assistance. I'd recommend the last two we save for another PR unless someone else wants to work on it.

andrewbrown00:
@asaadmahmood will you help @shred86 to make sure we are using the `latest version` of the `mattermosticon` font.

> @andrewbrown00 I'm not sure which directory you're referring to. I'm looking at the [fonts/mattermosticons](https://github.com/mattermost/mattermost-webapp/tree/master/fonts/mattermosticons) on this repository. Can you provide me a link to the directory you have pictured of the mattermost icons?
> 
> Also, do I need to create a new jsx file for these icons individually in the components/widgets/icons directory or can I just use this mattermosticon font type directly? Unfortunately styling is a weak area for me so I may need some assistance on this one.

asaadmahmood:
@shred86 The icons are already there, you can find all the icons this directory.
Screenshot 2020-05-05 at 10 31 07 PM

There you would be able to see the class names of all the icons.
You can use each icon like this:
`<i className='icon icon-arrow-right'/>`

shred86:
Appreciate the help! Some remaining issues:
- There doesn't appear to be a `icon-minus`, so I'm just using `icon-minus-box` as a placeholder (see image below).
- I'm not sure how to implement the "fit to zoom" option. I propose we just have a "reset zoom" option in the middle for now and let someone else make another PR to add this feature if it's desired. I do think we should use a different icon though for reseting the zoom level to default (currently just have the `icon-magnify` as a placeholder).
- Still not sure what is to be done about the issue I [mentioned above](https://github.com/mattermost/mattermost-webapp/pull/5361#issuecomment-617515676).

Screen Shot 2020-05-07 at 8 54 41 PM

esethna:
> * Still not sure what is to be done about the issue I [mentioned above](https://github.com/mattermost/mattermost-webapp/pull/5361#issuecomment-617515676).

@bradjcoughlin can you help with this question above ^

esethna:
cc// @andrewbrown00 on these two questions:

> Appreciate the help! Some remaining issues:
> 
> * There doesn't appear to be a `icon-minus`, so I'm just using `icon-minus-box` as a placeholder (see image below).
> * I'm not sure how to implement the "fit to zoom" option. I propose we just have a "reset zoom" option in the middle for now and let someone else make another PR to add this feature if it's still desired.

andrewbrown00:
> cc// @andrewbrown00 on these two questions:
> 
> > Appreciate the help! Some remaining issues:
> > 
> > * There doesn't appear to be a `icon-minus`, so I'm just using `icon-minus-box` as a placeholder (see image below).
> > * I'm not sure how to implement the "fit to zoom" option. I propose we just have a "reset zoom" option in the middle for now and let someone else make another PR to add this feature if it's still desired.

Hi @shred86 

- [ ] Do you see the code `0xf0374` in the font icons?   
   image
- [ ] I'm good with pulling the `fit to zoom` feature for now. We can just show the `Reset to zoom` icon and keep it disabled until as user zooms in.

shred86:
Thanks @andrewbrown00. I had to rebase my branch but now I'm seeing it. As for the `Reset zoom` icon, which would be preferred to use? Since you can reset the zoom from a zoomed in or zoomed out state, the magnifying glass with a minus sign doesn't always make sense. Perhaps we can add another icon that better represents the `Reset zoom` functionality?

andrewbrown00:
Welcome @shred86 glad that worked for the icon 👍

Can we swap between the `magnify minus` and magnify plus` icons for the `Reset zoom` functionality based on what you mentioned in previous comment that..
> PDFs render to a scale of 1 (this is the current behavior). If the PDF resolution is larger than the modal, then you will want to zoom out in some cases.

We can keep the UI aligned with prototype and make a small functional change and remove the `Fit to width` logic (and tooltip) and make the control always `Reset zoom` but it will either zoom in or out based on the zoom state the user defined.

shred86:
@andrewbrown00 Yep, just made the change. The only question is which icon do we use for when the scale is 1? The `icon-magnify` is slightly different than the `icon-magnify-minus` and `icon-magnify-plus`. I could just use one of the minus or plus icons and have it grayed out, or we would just need to add another icon that matches the minus and plus icon.

andrewbrown00:
Thanks @shred86 👍 lets use the `icon-magnify-minus` as the default icon (disabled) since I'm 4/5 that users will likely zoom in first rather than zooming out.

shred86:
> Thanks @shred86 👍 lets use the `icon-magnify-minus` as the default icon (disabled) since I'm 4/5 that users will likely zoom in first rather than zooming out.

I was thinking the same thing. Changes made.

shred86:
Resolved the following:
- Fixed the loading screen to show a full black background when loading. 
- Added tooltips to zoom controls.
- Updated the `Download to read more pages` button to reflect the current theme.

I've tried fixing the horizontal scroll bar multiple times but I'm having no luck. If there's any CSS experts, I'll definitely need help with this.

sass/components/_modal.scss
```diff
@@ -338,13 +338,14 @@
 
         .modal-image__content {
             background: #F3F3F3;
-            max-height: 91.5vh;
-            max-width: 96vw;
+            height: 90vh;
+            width: 90vw;
             overflow-x: hidden;
             overflow-y: visible;
```
shred86, line 344: I believe the horizontal scrolling bar issue has something to do with these overflows. I've tried different variations of `overflow-x` and `overflow-y` to include adding/modifying child styling classes, but no luck.

andrewbrown00:
Thanks @shred86 for knocking out the changes 🎉 

**Item to resolve**
- [ ] Add 4px of space between the download icon and label
   
   Current | Per the specs
   ----- | -----
   image | image

> I've tried fixing the horizontal scroll bar multiple times but I'm having no luck. If there's any CSS experts, I'll definitely need help with this.

I see if @deanwhillier can provide some guidance on ^^^

andrewbrown00:
@shred86 will you update the `branch`, seems there was a **conflict** with the `test server`

esethna:
cc// @andrewbrown00 build is passing if we want to retest?

asaadmahmood:
@shred86 I can see that there is an icon-minus in the webapp repo.

Screenshot 2020-06-12 at 11 39 45 AM

shred86:
Ah sorry, forgot to update my previous post. It's using the `icon-minus` now.

I am not sure why the test server creation fails.

components/view_image/view_image.jsx
```diff
@@ -203,6 +205,33 @@ export default class ViewImageModal extends React.PureComponent {
         this.setState({showCloseBtn: false});
     }
 
+    handleZoomIn = () => {
+        let newScale = this.state.scale;
+        newScale = Math.min(newScale + ZoomSettings.SCALE_DELTA, ZoomSettings.MAX_SCALE);
+        this.setState(() => ({
```
sudheerDev, line 211: Can this be simplify this to be ` this.setState({scale: newScale);` since we are not using `this.setState((prevState) => ({` prevState?

components/view_image/view_image.jsx
```diff
@@ -203,6 +205,33 @@ export default class ViewImageModal extends React.PureComponent {
...
+            scale: newScale,
+        }));
+    };
+
+    handleZoomOut = () => {
+        let newScale = this.state.scale;
+        newScale = Math.max(newScale - ZoomSettings.SCALE_DELTA, ZoomSettings.MIN_SCALE);
+        this.setState(() => ({
```
sudheerDev, line 219: same comment as above and for other instances as well.

esethna:
Ping @andrewbrown00 for UX review

esethna:
Hi @shred86, let us know if you have any questions on the above dev review?

components/view_image/view_image.jsx
shred86: Changed.

shred86:
Just made the changes. There's still the issue with larger PDFs where zooming in/out quickly will result in these errors in the console:

> Uncaught (in promise) Error: Cannot use the same canvas during multiple render() operations. Use different canvas or ensure previous operations were cancelled or completed.

Not sure how much this matters but it's something I mentioned early on and was hoping someone could look at it. I implemented debounce to prevent a user from being able to call the zoom in/out functions too fast which helps but the error still exists on large/complex PDFs.

esethna:
@sudheerDev any thoughts here? ^

sudheerDev:
@shred86 Hey, i had a look at the issue and it is not caused by the PR but it is evident because we are trying to render multiple canvases with zoom now.

https://github.com/mattermost/mattermost-webapp/blob/08a4c5726d5c886abe4b1921481b85b91f61b0ec/components/pdf_preview.jsx#L83-L86

Here we are using the same canvas and re-rendering it with every zoom change and so far from what i understood we need to cancel the operations or wait for them to complete before again initiating render. I don't see any methods on the PDF-JS we are using to do that for suggesting here. We seem to have a really old version of the library. 

Options to fix the root problem. use react version of the library(Hopefully they have it handled internally) or render new canvas on each render like https://github.com/FranckFreiburger/vue-pdf/pull/77

I suggest we leave out of the scope of the PR and make a new ticket and PR for the issue above. 

Let me know if you want to look at the options and want to work on it.

shred86:
Thanks @sudheerDev. It looks like the issue was fixed in the pdf.js library with [this pull request](https://github.com/mozilla/pdf.js/pull/9853) (if I'm understanding that PR correctly). Would we just be able to update the version of pdf.js we're using to the latest?

esethna:
cc// @sudheerDev ^

sudheerDev:
As discussed in the community instance

> There is no blocker for updating the library. But the pr link here is for fixing the issue of re-rendering when after a cancel operation. The issue we are seeing in the zoom pdf PR is that we are not canceling to even hit this issue

I suggest we continue with the PR and log a separate ticket for the issue.

components/pdf_preview.jsx
```diff
@@ -167,16 +168,24 @@ export default class PDFPreview extends React.PureComponent {
...
+                    key='previewpdfmorepages'
                 >
-                    <FormattedMessage
-                        id='pdf_preview.max_pages'
-                        defaultMessage='Download to read more pages'
-                    />
-                </a>,
+                    <button
+                        className='btn btn-primary'
+                        onClick={(e) => {
```
sudheerDev, line 177: Can you make onClick call back a function on the instance?
```
onClick={this.downloadFile}
```

components/view_image/view_image.jsx
```diff
@@ -240,11 +263,13 @@ export default class ViewImageModal extends React.PureComponent {
                     />
                 );
             } else if (fileInfo && fileInfo.extension && fileInfo.extension === FileTypes.PDF) {
+                this.setState({showZoomControls: true});
```
sudheerDev, line 266: setting setState in render is not a good idea. Can we change the setSate to be on `getDerivedStateFromProps` or any other lifecycle you see fit?

components/pdf_preview.jsx
shred86: Done.

components/view_image/view_image.jsx
shred86: Having some trouble with this one. More I think about it, I'm not sure if it's even worth making `showZoomControls` a part of the state? My understanding is when `setState` is called, `render()` is called which seems unnecessary. All I'm trying to do is set a variable to true if it's a PDF so the popup bar will display the zoom controls.

shred86: I think I figured it out. `getDerivedStatesFromProps` seems like the best place to put it as far as I can tell.

shred86:
Moving `this.setState({showZoomControls: true});` out of `render()` seems to have been what was primarily causing the [issue](https://github.com/mattermost/mattermost-webapp/pull/5361#issuecomment-653683636) I mentioned above. I'm guessing because it was probably rendering twice?

I tried removing debounce but when I do, the error starts occurring again if you rapidly zoom in or out on a large PDF, so I kept debounce in but reduced the delay between clicks.

shred86:
Found some issues on mobile devices. The popover bar was extending outside of the viewable screen area on an iPad Pro and iPhone 11. There were also issues based on how I was aligning the zoom controls. Reworked the popover bar styling to fix some of the issues, but one of the issues I can't seem to figure out is with the iPhone specifically as you can see pictured below. There is a gray "padding" on the right edge of the modal view. This is only occurring on the iPhone in portrait mode. Additionally, you can see the zoom icons sort of get stacked which is less than ideal.

Simulator Screen Shot - iPhone 11 - 2020-07-11 at 15 10 44

andrewbrown00:
@shred86 are we able to make the modal dynamic? It feels odd to have a large modal with small images, like:
image

Versus how the modal would resize based on on the files size, like:
image

shred86:
That is what I desired but couldn't figure out how to get it to work with the PDF zooming. I can look at it again but I remember spending a considerable amount of time trying to get that behavior and couldn't figure it out without keeping the size fixed.

andrewbrown00:
> That is what I desired but couldn't figure out how to get it to work with the PDF zooming. I can look at it again but I remember spending a considerable amount of time trying to get that behavior and couldn't figure it out without keeping the size fixed.

Thanks for the context @shred86, Let me know if we can help in any way, maybe pull another dev in to help work through it?

shred86:
Just committed a change that should address this.

andrewbrown00:
> Just committed a change that should address this. Basically, PDFs will display at a fixed height and width (91.5vh and 96vw). All others will display based on the size of the media up to a max of 91.5vh and 96vw. I also added a min height and width of 240 and 440px respectively to prevent the public link and download link from wrapping to a second line.

Thanks @shred86, are you able to center the images vertically and horizontally in the frame?

Screen Shot 2020-07-16 at 8 50 03 PM

shred86:
@andrewbrown00 Should be resolved now.

andrewbrown00:
> @andrewbrown00 Should be resolved now.

Thanks @shred86, the images are showing centered now 👍 

image

-----

**Items to look in to:**

- [ ] Will you check the other file types and confirm if the are being clipped and centered because of the recent change. If you look at the screenshots below, they both look to be clipped but I can't scroll up any higher.

**PDF shows a partial page**

Screen Shot 2020-07-19 at 7 24 42 PM

**JSX file doesn't start at 1**

Screen Shot 2020-07-19 at 7 23 44 PM

shred86:
Ah I can't believe I missed that. Yeah, I'm probably going to need some CSS assistance on this one. Styling is a weak area for me unfortunately.

andrewbrown00:
> Ah I can't believe I missed that. Yeah, I'm probably going to need some CSS assistance on this one. Styling is a weak area for me unfortunately.

All good @shred86 🙂 . I'm pulling in @asaadmahmood to see if he can provide you some `CSS` assistance.

jasonblais:
@asaadmahmood are you open to helping with CSS - see notes from @andrewbrown00 above for further context: https://github.com/mattermost/mattermost-webapp/pull/5361#issuecomment-660726360

shred86:
Spent some time looking at this and I think an easy solution, which I've made in this latest commit, is removing `min-height` on the `.modal-image__content` class. This eliminates the need to vertically center anything since the height will just adjust to the size of the content. I left `min-width` in there to prevent the popover bar text from wrapping to a second line.

jasonblais:
@shred86 Heads up that I spoke to Asaad and he'll help take a look at the CSS changes tomorrow (Tuesday), as he's out today. Big thanks for your contribution 🚀

asaadmahmood:
@shred86 I've fixed the scrolling issue you were having trouble with on this branch.

shred86:
Thank you @asaadmahmood!

jasonblais:
@asaadmahmood @sudheerDev @andrewbrown00 this contribution is ready for your re-review

jasonblais:
Actually, @asaadmahmood @shred86 it looks like the build tests fail, and as a result the test server creation fails. Would you be open to help take a look?

sudheerDev (approved):
LGTM

shred86:
I'm unable to duplicate the zoom controls showing up on non-PDF files on my local MM instance. It seems like this is only occurring in the build that SpinWick is creating. I'm going to try to make a change but the only way I can test it is in the build that SpinWick creates which is not ideal.

As for the "Download to read more pages" button not going back to the default state, I think I see what you're saying but I'm not sure where the 'Add Comment' button is for reference.

andrewbrown00:
Thanks @shred86 for the update, glad you were able to resolve the issue.

@asaadmahmood did you see this comment?
> Unfortunately it looks like I overrode your changes to fix the horizontal scroll bar during my rebase. Is there any chance you could push those changes again? Sorry about that. I tried looking for your changes in the commit history but can't seem to find it.

asaadmahmood:
@andrewbrown00 Looks like I missed it, will have a look.

asaadmahmood:
@shred86 the horizontal scrollbar does not reproduce for me.

Screenshot 2020-09-03 at 4 07 31 PM

shred86:
@asaadmahmood You have to zoom in on the PDF and you’ll see the bottom horizontal scroll bar doesn’t appear until you scroll to the very bottom of the page. It was a small styling issue that you fixed previously (I think adding a z-index property or something). It’s not a show stopper, just something that was requested to be cleaned up.

asaadmahmood:
I'm at the bottom, I still don't see it.
Screenshot 2020-09-05 at 3 39 06 PM

asaadmahmood:
@shred86 ^

shred86:
@asaadmahmood You have to zoom in on the PDF. There is no horizontal scroll bar if your zoom level fits the entire width of the page. See the screenshot below. Note the lack of a horizontal scroll bar - that is the issue. 

Screen Shot 2020-09-07 at 8 56 31 AM

asaadmahmood:
@shred86 I've pushed a fix.

shred86:
So it looks like this change causes a few issues:

1) The blank black loading screen width is now very narrow.
2) The width is no longer fixed causing the links on the bottom right to extend to a second line.
3) There's a small artifact on the right side of the horizontal scroll bar.

Screen Shot 2020-09-08 at 5 59 37 PM

asaadmahmood:
@shred86 Pushed a fix.
1. That should be fixed.
2. That should be fixed as well.
3. I don't see that.

shred86:
Thanks @asaadmahmood, looks great! The artifact only seems to show up in Safari when you zoom in and the horizontal scroll bar appears. I'm not seeing it in Chrome.

asaadmahmood (approved):

hanzei:
@esethna Is this PR ready for QA review?

esethna:
@hanzei @jgilliam17 yes! :)

jgilliam17:
@hanzei Can you please help me with the error updating? Thanks

hanzei:
@jgilliam17 Doen

jgilliam17:
Thank you @shred86 
- PDF zooming  in, out and reset works great; smooth horizontal and vertical scrolling; verified tooltip on on `+`, `-`, and 🔍 

- I am seeing one small issue with the scrollbar when using a dark theme on Chrome, Edge, Safari and Desktop app on macOS, Windows and Ubuntu. Note: Works as expected on Firefox (macOS, Windows & Ubuntu).
See both videos for comparison. When dark theme is selected scroll indicator is not visible.
scrollbar light theme
scrollbar dark theme

shred86:
Thanks @jgilliam17!

 @asaadmahmood any chance you could help with this one as well? The only addition I made to _modal.scss is the addition of line 339:

```
        // provides a background for the scrollbar
        .modal-image__background {
            background: #F3F3F3;
        }
```

Not sure how this would affect it but again, I'm not very good with CSS and unfortunately very limited on time to be able to troubleshoot.

jasonblais:
@asaadmahmood kind reminder to help with the latest comment :) 

https://github.com/mattermost/mattermost-webapp/pull/5361#issuecomment-703332322

asaadmahmood:
@shred86 Pushed and fixed.

asaadmahmood (approved):

jgilliam17:
Thank you @shred86 
Tested, looks good to merge.
-Verified zoom in, out and reset for PDF files. Verified get public link and download buttons. Zoom options are not available when previewing files that do not support zooming. Scroll bars visible in both, light and dark themes.

jgilliam17 (approved):

jgilliam17:
@amyblais I'll create a Jira Ticket and link here shortly.

hanzei:
Thanks you very much for this amazing contribution @shred86 :rocket: :tada:

PR #5362, opened 2020-04-21
MM-23816: Group Mentions: Add ability to rename group names

catalintomai:
#### Summary
Covers reqs #9, #10, #11 in https://mattermost.atlassian.net/wiki/spaces/EN/pages/389710117/EE+Group+Mentions+MVP

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23816

#### Related Pull Requests
redux:  https://github.com/mattermost/mattermost-redux/pull/1135
server:  https://github.com/mattermost/mattermost-server/pull/14338

#### Screenshots
Screen Shot 2020-04-21 at 4 15 39 PM


components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -115,14 +132,59 @@ export default class GroupDetails extends React.PureComponent {
         this.setState({loadingTeamsAndChannels: false});
     }
 
-    onToggle = async (allowReference) => {
-        this.setState({allowReference});
-        this.props.actions.patchGroup(this.props.groupID, {allow_reference: allowReference});
+    onMentionToggle = async (allowReference) => {
```
hahmadia, line 135: Is it possible to provide screenshot of what this function effects? It be helpful to visualize.

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -115,14 +132,59 @@ export default class GroupDetails extends React.PureComponent {
...
+        const isOriginalAllowReference = group.allow_reference;
+
+        this.setState({
+            saveNeeded: true,
+            allowReference,
+            hasAllowReferenceChanged: allowReference !== isOriginalAllowReference}
+        );
+    }
+
+    onMentionChange = async (e) => {
```
hahmadia, line 146: Is it possible to provide a screenshot on what part of the group config page this function effects? It help for better visualization.

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -115,14 +132,59 @@ export default class GroupDetails extends React.PureComponent {
         this.setState({loadingTeamsAndChannels: false});
     }
 
-    onToggle = async (allowReference) => {
-        this.setState({allowReference});
-        this.props.actions.patchGroup(this.props.groupID, {allow_reference: allowReference});
+    onMentionToggle = async (allowReference) => {
+        const {group} = this.props;
+        const isOriginalAllowReference = group.allow_reference;
```
hahmadia, line 137: I would prefer if this was just named `originalAllowReference`.

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -115,14 +132,59 @@ export default class GroupDetails extends React.PureComponent {
...
+        this.setState({
+            saveNeeded: true,
+            allowReference,
+            hasAllowReferenceChanged: allowReference !== isOriginalAllowReference}
+        );
+    }
+
+    onMentionChange = async (e) => {
+        const {group} = this.props;
+        const isOriginalGroupMentionName = group.name;
```
hahmadia, line 148: I would prefer if this was just named `originalGroupMentionName`.

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -115,14 +132,59 @@ export default class GroupDetails extends React.PureComponent {
...
+    }
+
+    handleSubmit = async () => {
+        this.setState({saving: true});
+        const {allowReference, groupMentionName, hasAllowReferenceChanged, hasGroupMentionNameChanged} = this.state;
+
+        let serverError = null;
+        let saveNeeded = false;
+
+        if (!groupMentionName || groupMentionName.length === 0) {
```
hahmadia, line 167: Why do we need a `!groupMentionName` ? If `groupMentionName` is a string, wouldn't the `groupMentionName.length === 0` be enough if your checking for a empty string? You initialize it as a string object so I'm lead to believe it will never be anything but a string. Also, I think it may be a little more clearer if you did `groupMentionName === ""` instead of using `.length` just to make it more definitive that it's a string (but this just for more clarity and isn't something that needs to be changed).

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -115,14 +132,59 @@ export default class GroupDetails extends React.PureComponent {
...
+            saveNeeded = true;
+            serverError = <NeedGroupNameError/>;
+            this.setState({allowReference, serverError, saving: false, saveNeeded});
+        } else if (hasAllowReferenceChanged || hasGroupMentionNameChanged) {
+            saveNeeded = false;
+            const result = await this.props.actions.patchGroup(this.props.groupID, {allow_reference: allowReference, name: groupMentionName});
+            if (result.error) {
+                saveNeeded = true;
+                if (result.error.server_error_id === 'store.sql_group.unique_constraint') {
```
hahmadia, line 176: Ah I'm not 100% if we want to be introducing backend errors into frontend in such manner. I'm okay with it but I will let @mkraft make a comment. Not sure if there is a better way to handle it though so let's wait to see what others have to say about it.

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -235,6 +298,13 @@ export default class GroupDetails extends React.PureComponent {
                     </div>
                 </div>
 
+                <SaveChangesPanel
+                    saving={saving}
+                    cancelLink='/admin_console/user_management/teams'
```
hahmadia, line 303: Should cancelling redirect us to teams? Or groups?

components/admin_console/group_settings/group_details/group_details_errors.jsx
```diff
@@ -0,0 +1,37 @@
...
+};
+
+export const DuplicateGroupNameError = () => (
+    <FormError
+        iconClassName={'fa-exclamation-triangle'}
+        textClassName={'has-error'}
+        error={(
+            <FormattedMessage
+                id='admin.group_settings.group_detail.duplicateMentionNameError'
+                defaultMessage='You must specify the group name for mentioning.'
```
hahmadia, line 34: Should the error message here be different than the one above?

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: I saw this pattern used in some other places in the client, therefore I used it. The thing is, the server does not have the context that the groupname is used as a mention, it only knows that there is a duplicate name. The context of that error exists at the client.

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: probably groups, thanks for pointing out.

components/admin_console/group_settings/group_details/group_details_errors.jsx
catalintomai: yes, definitely.

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: I'll add a screenshot.

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: I'll add a screenshot.

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: sure

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: sure.

components/admin_console/group_settings/group_details/group_details.jsx
hahmadia: I guess if it's being used in other places then it's fine to be used here as well. I just thought since we had discussions in developer meetings regarding like backend errors bubbling up to front end and wanting to avoid this is why I thought we shouldn't be doing it. @mkraft any thoughts?

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: Actually, according to https://stackoverflow.com/questions/5515310/is-there-a-standard-function-to-check-for-null-undefined-or-blank-variables-in just `!groupMentionName` should be enough.

hahmadia: I know this is common way to check for falsy values but I personally dislike it 😆 It makes it feel like I'm working with a boolean when I'm not. It just makes it more readable when you do `variable === ''` so then I know immediately this should be a string and not something else. But I'm 0/5 and if you prefer to keep it `!groupMentionName` then I'm fine with it too.

components/admin_console/group_settings/group_details/group_details.jsx
catalintomai: It could be that that the discussion was about them not bubbling through unchecked - in this case we are converting it. to a client error. I wish there was a more streamlined way of doing it, however this is the pattern I found.

hahmadia: True say. Okay in that case it's fine then.

components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -152,9 +217,10 @@ export default class GroupDetails extends React.PureComponent {
 
                         <GroupProfileAndSettings
                             displayname={group.display_name}
-                            name={group.name}
+                            mentionname={groupMentionName}
```
hahmadia, line 220: I think we want this to be `mentionName` (camel case)

components/admin_console/group_settings/group_details/group_profile_and_settings.jsx
```diff
@@ -71,17 +73,21 @@ export const GroupProfileAndSettings = ({displayname, name, allowReference, onTo
         </div>
         {allowReference &&
             <GroupProfile
-                name={'@' + name}
+                name={mentionname}
```
hahmadia, line 76: `mentionName` ? (camel case)

components/admin_console/group_settings/group_details/group_profile_and_settings.jsx
```diff
@@ -71,17 +73,21 @@ export const GroupProfileAndSettings = ({displayname, name, allowReference, onTo
...
+                showAtMention={true}
+                onChange={onChange}
             />
         }
     </AdminPanel>);
 
 GroupProfileAndSettings.propTypes = {
     displayname: PropTypes.string.isRequired,
-    name: PropTypes.string.isRequired,
+    mentionname: PropTypes.string.isRequired,
```
hahmadia, line 89: `mentionName` ?

srkgupta:
Some more issues:

[MM-24680](https://mattermost.atlassian.net/browse/MM-24680): Group Mentions with upper case letters should automatically be converted to lowercase instead of displaying an error

[MM-24681](https://mattermost.atlassian.net/browse/MM-24681): Existing Group Mention check should be matched only with the group mentions which are enabled

catalintomai:
> Some more issues:
> 
> [MM-24680](https://mattermost.atlassian.net/browse/MM-24680): Group Mentions with upper case letters should automatically be converted to lowercase instead of displaying an error
> 
> [MM-24681](https://mattermost.atlassian.net/browse/MM-24681): Existing Group Mention check should be matched only with the group mentions which are enabled

@srkgupta , I’ve updated https://github.com/mattermost/mattermost-webapp/pull/5362 with the fixes (except https://mattermost.atlassian.net/browse/MM-24677).

PR #5367, opened 2020-04-21
MM-24126 Supporing plugin full page routes.

crspeller:
#### Summary
Supports full page plugin routes. Two available routes. First is under `/plug/` where plugins can hook full pages independent of team. Second is under `/:team/` where plugins can hook routes that depend on a specific team.

`/plug/` was chosen instead of allowing plugins to define at true root to avoid additional team name conflicts. _why didn't we do /team/:team? ... 6.0?_


marianunez:
Should we also include the `pluginId` in the route to avoid conflicts between plugins as well? I can see multiple plugins registering their `/admin` or `/backstage` that can cause conflicts.

crspeller:
@marianunez Good point. I was unsure about that. Ideally yes, but these are user visible routes and the plugin might want to make them nicer then their plugin ID. Not all plugins have nice IDs like `jira` or `github` some do reverse dot notation like `com.mattermost.incident-response`. I feel like UX is not going to like that as a prefix on every route. Thoughts?

marianunez:
> @marianunez Good point. I was unsure about that. Ideally yes, but these are user visible routes and the plugin might want to make them nicer then their plugin ID. Not all plugins have nice IDs like `jira` or `github` some do reverse dot notation like `com.mattermost.incident-response`. I feel like UX is not going to like that as a prefix on every route. Thoughts?

Plugin id is what is currently used in the System Console for the plugins configuration pages (i.e. `http://localhost:8065/admin_console/plugins/plugin_com.mattermost.custom-attributes`). 

I suggest we do the same and maybe just replace `plug` completely with just `plugin-id`.

marianunez:
> @marianunez Good point. I was unsure about that. Ideally yes, but these are user visible routes and the plugin might want to make them nicer then their plugin ID. Not all plugins have nice IDs like `jira` or `github` some do reverse dot notation like `com.mattermost.incident-response`. I feel like UX is not going to like that as a prefix on every route. Thoughts?

Plugin id is what is currently used in the System Console for the plugins configuration pages (i.e. `http://localhost:8065/admin_console/plugins/plugin_com.mattermost.custom-attributes`). 

I suggest we do the same and maybe just replace `plug` completely with just `plugin-id`.

crspeller:
@marianunez 
Right, but that's the system console and the plugin doesn't get a say in that. We don't really have another choice here. I feel like when we want to do the incident response plugin's route we are going to want `/:team/incident_backstage` or `/incident_backstage` and not `/:team/com.mattermost.incident-respose/backstage` or `/com.mattermost.incident-respose/backstage`

marianunez:
We could enforce the initial part to include the plugin id and the plugin author has the say for the rest of the route. I think we should avoid collisions in some way with plugin ids or any other solution that would not leave any possibility of conflicts.

We need some input on UX/PM here - not sure who that would be for this Toolkit question. @lieut-data Thoughts?

plugins/registry.js
```diff
@@ -584,4 +584,58 @@ export default class PluginRegistry {
...
+
+    // Register a component to be displayed at a custom route under /plug/
+    // Accepts the following:
+    // - route - The route to be displayed at.
+    // - component - A react component to display.
+    // Returns:
+    // - id: a unique identifier
+    registerCustomRoute(route, component) {
+        const id = generateId();
+        let fixedroute = route.trim();
```
hmhealey, line 623: Minor nit, but the casing on this should be `fixedRoute`

hmhealey:
We're finally getting App Centre! 🎉 

I don't think including the plugin ID in the UI is too much of a problem. URLs are usually mostly meaningless for users with browsers even going so far as to grey out the unimportant parts of the URL (aka everything other than the domain name). Also, the desktop app won't even show that URL at all. It might be worth adding the package name automatically to make it easy to prevent naming clashes.

hmhealey:
Also, regarding `/plug/`, we have reserved `plugins` as an invalid team name [recently](https://github.com/mattermost/mattermost-server/pull/13606), although I could still see conflicts happening with teams created prior to 5.22.

lieut-data:
Adding @aaronrothschild as PM since this will impact plugins beyond just incident response.

components/needs_team/index.ts
```diff
@@ -31,10 +30,11 @@ type OwnProps = {
     };
 }
 
-function mapStateToProps(state: GlobalState, ownProps: OwnProps) {
+function mapStateToProps(state: any, ownProps: OwnProps) {
```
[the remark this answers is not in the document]
crspeller: Awesome! I figured it existed but for some reason didn't find it.

plugins/registry.js
crspeller: I don't know why I keep doing this. I keep not camelCasing stuff when I am typing.

crspeller:
@hmhealey 

Haha yup. Finally app center. 

The prefix can't be `/plugins/<pluginid>` because we use that for the plugin API.

If we really don't care about how the URL looks then I guess we can just add the plugin ID as a prefix. That would be preferable. @aaronrothschild That's the question we would like an answer to.

aaronrothschild:
> @hmhealey
> 
> Haha yup. Finally app center.
> 
> The prefix can't be `/plugins/<pluginid>` because we use that for the plugin API.
> 
> If we really don't care about how the URL looks then I guess we can just add the plugin ID as a prefix. That would be preferable. @aaronrothschild That's the question we would like an answer to.

@crspeller I Agree with Harrison's points and don't think this is a big deal from a UX perspective, particularly as I wouldn't be expecting the user to ever share these URLs either. 

Let's go the easy path and insert the plugiID to avoid collisions. 

Just an FYI for @joebordes and @jespino as this seems related or a replacement for https://github.com/mattermost/mattermost-webapp/pull/4607 ?

joebordes:
Finally !!! Couldn't be happier  :-D
Great!

crspeller:
Ok. Pushed an update to add the plugin id prefixes. 
Routes are now:
- `/:team/pluginid/custompart`
- `/plug/pluginid/custompart`

Ready for re-review @marianunez @hmhealey

plugins/registry.js
```diff
@@ -584,4 +584,60 @@ export default class PluginRegistry {
 
         return {id, showRHSPlugin: showRHSPlugin(id), hideRHSPlugin: hideRHSPlugin(id), toggleRHSPlugin: toggleRHSPlugin(id)};
     }
+
+    // Register a Needs Team component by providing a route past /:team/ to be displayed at.
```
marianunez, line 588:
```suggestion
    // Register a Needs Team component by providing a route past /:team/:pluginId/ to be displayed at.
```

plugins/registry.js
```diff
@@ -584,4 +584,60 @@ export default class PluginRegistry {
...
+                pluginId: this.id,
+                component,
+                route: fixedRoute,
+            },
+        });
+
+        return id;
+    }
+
+    // Register a component to be displayed at a custom route under /plug/
```
marianunez, line 616:
```suggestion
    // Register a component to be displayed at a custom route under /plug/:pluginId
```

plugins/registry.js
```diff
@@ -584,4 +584,60 @@ export default class PluginRegistry {
...
         return {id, showRHSPlugin: showRHSPlugin(id), hideRHSPlugin: hideRHSPlugin(id), toggleRHSPlugin: toggleRHSPlugin(id)};
     }
+
+    // Register a Needs Team component by providing a route past /:team/ to be displayed at.
+    // Accepts the following:
+    // - route - The route to be displayed at.
+    // - component - A react component to display.
+    // Returns:
+    // - id: a unique identifier
+    registerNeedsTeamComponent(route, component) {
```
marianunez, line 594: For consistency with the other one: 
```suggestion
    registerNeedsTeamRoute(route, component) {
```

components/root/root.jsx
```diff
@@ -389,6 +391,18 @@ export default class Root extends React.Component {
                         path={['/_redirect/integrations*', '/_redirect/pl/:postid']}
                         component={PermalinkRedirector}
                     />
+                    {this.props.plugins?.map((plugin) => (
+                        <Route
+                            key={plugin.id}
+                            path={'/plug/' + plugin.route}
```
marianunez, line 397: Do we still need `/plug`?

crspeller: Yes. We don't want plugin IDs to conflict with team names.

lieut-data:
@crspeller, also should we register `plug` as a reserved team name?

crspeller:
@lieut-data Probably. I think there are actually a lot of the client routes missing from that list. https://mattermost.atlassian.net/browse/MM-24516

prapti:
Re-reviewing a tiny UI discrepancy.

PR #5368, opened 2020-04-21
MM-24421 Upgrade Cypress version

saturninoabril:
#### Summary
Per investigation, crashes on daily E2E run is most likely related to Cypress itself than the application code or infra.  Looking at Cypress issues, there were several reports [related to crashes, out of memory error, hang up, etc](https://github.com/cypress-io/cypress/issues?q=is%3Aissue+is%3Aopen+memory).  With regards to the new version of v4.4.1 as compared to v4.0.2, there are several fixes [related to crashes](https://github.com/cypress-io/cypress/releases) too.
This PR hopefully fixes the issue since it's not reproducible on local but on the test server only.

Upgrade Cypress version to 4.4.1

#### Ticket Link
JIRA ticket: https://mattermost.atlassian.net/browse/MM-24421


saturninoabril:
Hmmm, does not happen to me. Have you tried deleting node_modules and then reinstalling packages again?

PR #5372, opened 2020-04-21
[MM-24177] Don't highlight group mention if doesn't have required permission

hahmadia:
**Blocked by:**
https://github.com/mattermost/mattermost-server/pull/14348

#### Summary

Fix group mentions shouldn't be highlighted when mentioned by someone who doesn't have permission and should highlight for everyone when triggered by someone who has the permission

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24177


components/post_markdown/post_markdown.jsx
```diff
@@ -63,17 +63,25 @@ export default class PostMarkdown extends React.PureComponent {
...
         if (post && post.props) {
-            options.mentionHighlight = !post.props.mentionHighlightDisabled;
+            mentionKeys = allMentionKeys;
+
+            if (post.props.disable_group_highlight) {
+                mentionKeys = mentionKeysWithoutGroups;
+            }
+
+            if (post.props.disable_channel_highlight) {
+                mentionKeys = mentionKeys.filter((value) => !["@all", "@channel", "@here"].includes(value))
```
fmunshi, line 83: this should probably be 
```suggestion
                mentionKeys = mentionKeys.filter((value) => !['@all', '@channel', '@here'].includes(value))
```

components/post_markdown/post_markdown.jsx
```diff
@@ -63,17 +63,25 @@ export default class PostMarkdown extends React.PureComponent {
...
+        let mentionKeys = [];
         if (post && post.props) {
-            options.mentionHighlight = !post.props.mentionHighlightDisabled;
+            mentionKeys = allMentionKeys;
+
+            if (post.props.disable_group_highlight) {
+                mentionKeys = mentionKeysWithoutGroups;
+            }
+
+            if (post.props.disable_channel_highlight) {
```
fmunshi, line 82: if you're using a different prop now you should replace all instances of the previous prop with this - or remove the old logic added here https://github.com/mattermost/mattermost-webapp/pull/4916/

hahmadia:
I'm going to pause working on this as it requires some code from https://github.com/mattermost/mattermost-webapp/pull/5331 and I don't want to re-duplicate work that is done so I will wait for the PR to get merged first so I can re-use some of those functions.

components/post_markdown/post_markdown.jsx
hmhealey: I'd say we should stick with the old prop names. As bad as it feels to mix up camel case and snake case here, being consistently bad with it is preferable.

components/post_markdown/post_markdown.jsx
```diff
@@ -63,17 +65,25 @@ export default class PostMarkdown extends React.PureComponent {
...
         if (post && post.props) {
-            options.mentionHighlight = !post.props.mentionHighlightDisabled;
+            mentionKeys = allMentionKeys;
+
+            if (post.props.disable_group_highlight) {
+                mentionKeys = mentionKeysWithoutGroups;
+            }
+
+            if (post.props.mentionHighlightDisabled) {
+                mentionKeys = mentionKeys.filter((value) => !['@all', '@channel', '@here'].includes(value.key));
```
hmhealey, line 85: This is going to cause `Markdown` to re-render whenever this does because `filter` creates a new array each time render is called. To solve that, we could memoize the final list of `mentionKeys` either by creating a selector in `componentspost_markdown/index.js` or we could use add a method to this component that returns this result, but it's wrapped with `memoizeResult` from `mattermost-redux/utils/helpers`.

hahmadia: Done! Thanks :)

components/post_markdown/post_markdown.jsx
```diff
@@ -50,6 +54,10 @@ export default class PostMarkdown extends React.PureComponent {
         options: {},
     };
 
+    memoizedGetMentionKeysWithoutChannelMentions(mentionKeys) {
+        return memoizeResult(() => mentionKeys.filter((value) => !['@all', '@channel', '@here'].includes(value.key)));
+    }
```
fmunshi, line 59: I think this should be in the container component and then passed in as a prop shouldnt it? i.e. `/components/post_markdown/index.js`

hahmadia: I'm not sure if I quite understand what you mean.

fmunshi: Look at how the memoization is done in `components/more_direct_channels/index.ts`

hahmadia: Done! I made it the same way it's done there.

components/post_markdown/index.js
```diff
@@ -3,19 +3,29 @@
...
+    const allMentionKeys = getAllUserMentionKeys(state);
+    const mentionKeysWithoutChannelMentions = filterOutChannelMentions(allMentionKeys);
     return {
         channel,
         pluginHooks: state.plugins.components.MessageWillFormat,
         hasPluginTooltips: Boolean(state.plugins.components.LinkTooltip),
         isUserCanManageMembers: channel && canManageMembers(state, channel),
+        allMentionKeys,
+        mentionKeysWithoutGroups: getCurrentUserMentionKeys(state),
+        mentionKeysWithoutChannelMentions,
```
fmunshi, line 25: Looks good to me but does it make sense maybe to check the `post.props` here and only send one set of mention keys to the component? thoughts?

hahmadia: I think the way it currently is fine. I don't see moving the logic to `index.js` will give it any sort of improvement so I think keeping it the way it currently is should be okay.

components/post_markdown/post_markdown.jsx
```diff
@@ -63,17 +66,25 @@ export default class PostMarkdown extends React.PureComponent {
...
         if (post && post.props) {
-            options.mentionHighlight = !post.props.mentionHighlightDisabled;
+            mentionKeys = allMentionKeys;
+
+            if (post.props.disable_group_highlight) {
+                mentionKeys = mentionKeysWithoutGroups;
+            }
+
+            if (post.props.mentionHighlightDisabled) {
+                mentionKeys = mentionKeysWithoutChannelMentions;
```
hmhealey, line 86: I think this is a slight behaviour change from the previous version since the version with channel-wide mentions disabled may still include group mentions, wouldn't it?

components/post_markdown/index.js
hmhealey: It would definitely be cleaner to do it that way since we'd have all of the logic for it in one place instead of having to pass multiple props through here and then do stuff with them later on. You could combine them all into one selector like

```javascript
function makeGetMentionKeysForPost() {
    return createSelector(
        getAllUserMentionKeys,
        getCurrentUserMentionKeys,
        (state, post) => post,
        (allMentionKeys, mentionKeysWithoutGroups, post) => {
            let mentionKeys = allMentionKeys;
            if (post?.props?.disable_group_highlight) {
                mentionKeys = mentionKeysWithoutGroups;
            }

            if (post?.props?.mentionHighlightDisabled) {
                mentionKeys = mentionKeys.filter((value) => !['@all', '@channel', '@here'].includes(value.key));
            }

            return mentionKeys;
        }
    );
}

    const getMentionKeysForPost = makeGetMentionKeysForPost();

    return (state, ownProps) => {
        return {
            mentionKeys: getMentionKeysForPost(state, ownProps.post);
        };
    };
}
```
Note that this is using a selector factory. Since the selector takes an argument, each post needs its own version of the selector to not overwrite each others' memoized results.

This version should also fix the behaviour change I mentioned before.

hahmadia: @hmhealey 
Ok I have refactored it and added test for that.

PR #5376, opened 2020-04-22
Experimental Sidebar Features - Quick Menus look and feel

devinbinnie:
#### Summary
This PR adds new quick menus to the new sidebar, allowing for quick channel and category operations from the sidebar.
Currently only the look and feel is implemented, along with a couple menu items that already existed in other parts of the system.
There will be a follow up PR to hook up the rest of the functionality.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23970
https://mattermost.atlassian.net/browse/MM-23980


hmhealey:
Known issues:
1. Category menu should only show on hover
2. Menu for muted channels is faded out
3. Menu button and scroll handle goes over dropdown menus

jgilliam17:
Thank you @devinbinnie 
Adding few items
1. LHS Scroll bar shows on top of menu
Screen Shot 2020-04-22 at 3 17 51 PM

2. `Move to` does not line up with options
Screen Shot 2020-04-22 at 3 21 44 PM

3. DM category has ..., instead of `+`, unable to add new DM/GM
Screen Shot 2020-04-22 at 5 29 32 PM

4. Maybe we should we dismiss all tooltips on click? 
Screen Shot 2020-04-22 at 5 23 23 PM

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -170,13 +187,13 @@ export default class SubMenuItem extends React.PureComponent<Props, State> {
                 >
                     <span
                         id={'channelHeaderDropdownIconLeft_' + id}
-                        className={'fa fa-angle-left SubMenu__icon-left' + (hasSubmenu && !isMobile ? '' : '-empty' + (isMobile ? ' mobile' : ''))}
+                        className={'fa fa-angle-left SubMenu__icon-left' + (hasSubmenu && !isMobile && (direction === 'left') ? '' : '-empty' + (isMobile ? ' mobile' : ''))}
```
nevyangelova, line 190: 3/5 prefer the use of 'classnames'

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -170,13 +187,13 @@ export default class SubMenuItem extends React.PureComponent<Props, State> {
...
                         aria-label={Utils.localizeMessage('post_info.submenu.icon', 'submenu icon').toLowerCase()}
                     />
                     {textProp}
                     <span
                         id={'channelHeaderDropdownIconRight_' + id}
-                        className={'fa fa-angle-right SubMenu__icon-right' + (hasSubmenu && isMobile ? '' : '-empty')}
+                        className={'fa fa-angle-right SubMenu__icon-right' + (hasSubmenu && (isMobile || direction === 'right') ? '' : '-empty')}
```
nevyangelova, line 196: 3/5 prefer the use of 'classnames'

sass/layout/_sidebar-left.scss
```diff
@@ -257,6 +257,82 @@
...
+
+            i {
+                font-size: 16px;
+                color: var(--center-channel-color-72);
+                line-height: 16px;
+            }
+        }
+
+        .Menu {
+            position: fixed;
```
nevyangelova, line 285: can we consider doing more BEM?

devinbinnie:
> Thanks @devinbinnie, looking good. Noting some more items here:
> 
> 
> image
> 
> 1. Mark as read icon doesn't match spec
> 
> image
> 

Fixed!

> 1. Copy link option isn't working on desktop

This wasn't working for me and then fixed itself...let me know if you can still reproduce...it could be a bug with the desktop app itself.

> 2. Leave channel option is broken (stuck in selected state)
> 
> image

So this was keeping me in the selected channel when leaving initially, but then stopped. Is that the issue you were having?

> 
> 1. Town Square should not have a leave channel option.

Fixed!

> 2. Can you update tooltips to "Channel options" and "Category options"

Done!

> 3. @andrewbrown00 we should look at the opacity and weight of  the dots, they feel heavy 0/5
> 4. Are the menus's hooked up to permissions? ie if I don't have create channel or manage members permissions, are we hiding those options?

They weren't, but they are now. Good catch.

> 
> @andrewbrown00 please help review. I'm adding PM approval, will let UX comment and iterate on the above or anything else I missed.

devinbinnie:
> Thank you @devinbinnie
> Adding few items
> 
> 1. LHS Scroll bar shows on top of menu
> 
> Screen Shot 2020-04-22 at 3 17 51 PM

This one is known, I'll probably need a follow-up ticket for this one to deal with the CSS headaches.

> 
> 1. `Move to` does not line up with options
> 
> Screen Shot 2020-04-22 at 3 21 44 PM

Same as above for this one.

> 
> 1. DM category has ..., instead of `+`, unable to add new DM/GM
> 
> Screen Shot 2020-04-22 at 5 29 32 PM

You can click on the 3-dot menu and click Create Channel (currently it's not working but it will be)

> 
> 1. Maybe we should we dismiss all tooltips on click?
> 
> Screen Shot 2020-04-22 at 5 23 23 PM

Hmm, this one might also need a follow-up ticket, since I don't have a lot of control over the tooltips AFAIK. I will investigate though.

components/widgets/menu/menu_items/submenu_item.tsx
devinbinnie: Agreed, but that's outside the scope of this PR. We're going to refactor all this menu code at some point since it's not quite robust enough.

components/widgets/menu/menu_items/submenu_item.tsx
devinbinnie: Same as above.

sass/layout/_sidebar-left.scss
devinbinnie: Sure, again for the `Menu` stuff we'll address that outside of this PR.

devinbinnie:
> Looking good @devinbinnie
> 
> **A few items to resolve**
> 
> * [ ]  Scrollbar sits on top of the menu
> 
> image

This one needs a separate ticket, it's a more complicated fix.

> 
> * [x]  Muted channels also mute the menu
> 
> image

Fixed!

> 
> * [x]  Can we position the secondary menu to always open top aligned to **Move to** list item.
> 
> image

Fixed, but I think we need to support the menu opening up with the submenus as well. Imagine you open the menu for the bottom channels and you have 10+ categories in the "Move to..." list, it will clip off the screen. We can do this later, but I think it's important from a UX POV.

> 
> * I agree with Eric's feedback around the `...` icon states, something I already updated in the Figma files.
> 
> Hover States	Selected States
> Decrease the icon on hover to match the default state of the `+` icons i.e. "Direct Messages +" and remove the fill state on the icon hover. Should be `64%` or `rgba (255,255,255,0.64)`	Increase the icon to `100% opacity` or `rgba (255,255,255,1)` when the menu is open and remove the fill state on the icon hover
> -----	-----
> image	image

I'm a bit confused here, I assume this meant to remove the background hover state and reduce the icon opacity to 64%, but bump it back to 100% on hover, so I did that. Is this correct? Let me know if I missed something.

> * [x]  Menu doesn't match the specs for spacing and icon colors:
> 
> * Should be 16px of left and right padding
> * Icons should have an opacity of `48%`
> 
> Current	Expected
> image	image

Added the opacity fix, but the 16px padding was there before. Should it be wider than that?

> * [x]  Direct messages should say **Mute/Unmute Conversation**, they say **Channel**
> * [x]  Direct messages should say **Close Conversation**, they say **Leave**
> 
> image

Fixed!

> 
> * [x]  0/5 if this part of the functional phase, but we shouldn't show the category a channel is currently in in the **Move to** secondary menu, i.e. the channel is already in the **Public Channels** category in the example below:
> 
> image

Yeah this wasn't worth the effort to mock out, so I have a TODO to add this later on.

> 
> **Also, we should show the full category name**
> 

This is only happening because the default categories are using the wrong `display_name`. This should change when we hook them up to the backend.

> * [x]  Direct Messages category should not have the `...` icon and keep the `+`. It's also missing the `sorting toggle`, [spec doc](https://docs.google.com/document/d/17Uw_HMptsXADgganXffjB1k8_WuAn5nyOU5JrqZj_AA/edit#).
> 
> Current	Expected
> image	image

Fixed. Based on spec it looks like the background hover state was removed as well, so I've done that. Let me know if that needs to change.

components/sidebar/sidebar_menu.tsx
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    componentWillUnmount() {
+        const scrollbars = document.querySelectorAll('#SidebarContainer .SidebarNavContainer .scrollbar--view');
+        if (scrollbars && scrollbars[0]) {
+            scrollbars[0].removeEventListener('scroll', this.setMenuPosition);
+        }
+    }
+
+    refCallback = (ref: Menu) => {
```
hmhealey, line 61: Similar to my other comment, could more of this be done in `handleMenuToggle` so we position the menu right before its shown instead of hijacking the ref for it?

sass/layout/_sidebar-left.scss
```diff
@@ -257,6 +257,81 @@
         outline-style: none;
     }
 
+    .SidebarMenu {
```
hmhealey, line 260: Maybe more of this should be done in a separate file? Similar to the widget code, you could have these styles next to the component that they're styling so that we can keep them apart from the older code that's in need of cleanup

components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    renderDropdownItems = () => {
+        const {intl, isMuted, category} = this.props;
+
+        let deleteCategory;
+        let renameCategory;
+        if (category.type === CategoryTypes.CUSTOM) {
+            deleteCategory = (
+                <Menu.Group>
+                    <Menu.ItemAction
```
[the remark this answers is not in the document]
devinbinnie: It's called `isDangerous` :P

components/sidebar/sidebar_channel/sidebar_channel_menu/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        let managePrivateChannelMembers = false;
+
+        if (currentTeam) {
+            managePublicChannelMembers = haveITeamPermission(state, {team: currentTeam.id, permission: Permissions.MANAGE_PUBLIC_CHANNEL_MEMBERS});
+            managePrivateChannelMembers = haveITeamPermission(state, {team: currentTeam.id, permission: Permissions.MANAGE_PRIVATE_CHANNEL_MEMBERS});
+        }
+
+        return {
+            currentUserId: getCurrentUserId(state),
+            categories: getCategoriesForTeam(state, currentTeam.id),
```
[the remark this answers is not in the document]
devinbinnie: I'm actually not sure, this is a pattern I copied from the older sidebar. Is it possible for `getCurrentTeam()` to return a null? According to TypeScript it isn't

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    moveToNewCategory = () => {
+        // TODO
+    }
+
+    copyLink = () => {
+        const channelUrl = this.props.channelLink;
+
+        const clipboard = navigator.clipboard;
```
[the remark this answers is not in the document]
devinbinnie: This is copied from Nevy's `dot_menu` code. @nevyangelova Is there a reason you went with this code over `copyToClipboard`?

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -0,0 +1,329 @@
...
+    favoriteChannel = () => {
+        this.props.actions.favoriteChannel(this.props.channel.id);
+    }
+
+    unfavoriteChannel = () => {
+        this.props.actions.unfavoriteChannel(this.props.channel.id);
+    }
+
+    unmuteChannel = () => {
+        this.props.actions.updateChannelNotifyProps(this.props.currentUserId, this.props.channel.id, {
```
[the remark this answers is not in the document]
devinbinnie: I suppose. Where should they live?

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -0,0 +1,329 @@
...
+            case 'G':
+                return category.type !== CategoryTypes.PRIVATE && category.type !== CategoryTypes.PUBLIC;
+            default:
+                return true;
+            }
+        }).map((category) => {
+            return {
+                id: `moveToCategory-${channel.id}-${category.id}`,
+                icon: category.type === CategoryTypes.FAVORITES ? (<i className='icon-star-outline'/>) : (<i className='icon-folder-outline'/>),
+                direction: 'right' as any,
```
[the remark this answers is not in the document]
devinbinnie: TypeScript has some weirdness, where if you restrict the type to a specific set of strings, for some reason it won't recognize when you use one of them. I think there's an issue in our config, or we're not typing it correctly.

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -0,0 +1,329 @@
...
+            dialogType: ChannelInviteModal,
+            dialogProps: {channel},
+        });
+    }
+
+    renderDropdownItems = () => {
+        const {intl, isUnread, isFavorite, isMuted, channel, categories} = this.props;
+
+        let markAsRead;
+        if (isUnread) {
```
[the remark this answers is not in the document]
devinbinnie: There isn't any `show` property on `Menu.ItemAction`, it's always rendered regardless of props.

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -0,0 +1,329 @@
...
+                <Menu.Group>
+                    {copyLink}
+                    {addMembers}
+                </Menu.Group>
+                {leaveChannel}
+            </React.Fragment>
+        );
+    }
+
+    refCallback = (ref: SidebarMenu) => {
```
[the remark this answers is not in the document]
devinbinnie: We have an `eslint` rule against updating state in `componentDidUpdate`, so I assume we're not really supposed to be doing that.

This is also the same method used in Nevy's `dot_menu` code.

components/sidebar/sidebar_menu.tsx
devinbinnie: See above comment.

sass/layout/_sidebar-left.scss
devinbinnie: This file is purely for the new sidebar code, so it won't affect any of the older stuff.

I'm putting it here because the styles used here are from the spec for the sidebar, and we don't have a concrete plan of how we're fixing the menus yet. I assume some of this code will make it into the `Menu` component, but I don't think it makes sense to refactor any of this now without a plan.

esethna:
cc// @cpanato ^

cpanato:
Why me @esethna ?

devinbinnie:
> Why me @esethna ?

@cpanato I think it was just to kick the SpinWick, but it seems to be working fine now. Sorry to bug you.

cpanato:
Not a problem at all, just need to add some context and not just ^

devinbinnie:
> Thanks @devinbinnie, all items have been resolved except:
> 
> * [x]  Clicking to open a menu should immediately hide all tooltips, i.e. hide `Category options` and `Channel options` tooltips after clicking the `...` icons or `Group Message names` tooltip

Fixed!

> * [x]  The padding is still off on both sides, plus it isn't taking in to account for the 'Move to' chevron icon on the right
>   image
> 

I think we're gonna have to jump on a call and do some live editing for this one, cause it looks right to me.

> **Direct Messages category:**
> 
> * [x]  The `sort` icons should only show up on hover of the **Direct Messages** category header

Fixed!

> * [x]  Clicking the `sort` icons should toggle the sort between `Recency` and `Alphabetical` sorting

This won't be functional until after this branch is merged and we start hooking things up. It doesn't really make sense to mock this currently, so we'll have it ready later.

>   **I updated the Figma file to make it more clear in the design**
>   
>   * [Recency](https://www.figma.com/file/zVNYaShAXDK8i1RdpuEkr8/Channel-Sidebar-All-Phases?node-id=1295%3A11971)
>   * [Alphabetical](https://www.figma.com/file/zVNYaShAXDK8i1RdpuEkr8/Channel-Sidebar-All-Phases?node-id=1738%3A22644)
>   * Here's [functional notes](https://docs.google.com/document/d/17Uw_HMptsXADgganXffjB1k8_WuAn5nyOU5JrqZj_AA/edit#bookmark=id.2ib7z27ipy51) that were in the spec doc for reference

devinbinnie:
@andrewbrown00 Ready for your re-review

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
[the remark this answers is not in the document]
nevyangelova: @devinbinnie oh, this change was made in my first week I believe and I didn't know we had this Util :/

deanwhillier:
@devinbinnie, I started looking at this on my iPad Pro. Here are a couple things I found:

- With no hover on touchscreen tablets, obviously none of the `...` menu’s are visible until you tap on a category header or channel. We may need to default show all menus on touch screens. @andrewbrown00, thoughts?
- Unfortunately there are some z-index and overflow issues with the `...` menus. :(

image

image

components/sidebar/sidebar_category/sidebar_category.tsx
```diff
@@ -114,52 +137,89 @@ export default class SidebarCategory extends React.PureComponent<Props> {
...
+                />
+            );
         }
 
         let displayName = category.display_name;
         if (category.type !== CategoryTypes.CUSTOM) {
             displayName = localizeMessage(`sidebar.types.${category.type}`, category.display_name);
         }
 
+        console.log(this.categoryMenuRef?.current?.menuRef?.current?.state.isMenuOpen);
```
hmhealey, line 213: Looks like you left in some debug code by mistake.

components/sidebar/sidebar_category/sidebar_category.tsx
```diff
@@ -29,13 +32,23 @@ type Props = {
...
 };
 
-export default class SidebarCategory extends React.PureComponent<Props> {
+type State = {
+    isMenuOpen: boolean;
+}
+
+export default class SidebarCategory extends React.PureComponent<Props, State> {
     categoryTitleRef: React.RefObject<HTMLButtonElement>;
+    categoryMenuRef: React.RefObject<SidebarCategoryMenuType>;
```
hmhealey, line 41: Regardless of this being debug code, shouldn't this use the type wrapped with `mapStateToProps`? Unless `connect(SidebarCategoryMenu)` forwards calls to the `SidebarCategoryMenu`, I wouldn't think you could access fields of the wrapped component anyway.

components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+            type: CategoryTypes.CUSTOM,
+            display_name: 'custom_category_1',
+        },
+        canCreatePublicChannel: true,
+        canCreatePrivateChannel: true,
+        isMuted: false,
+        actions: {},
+    };
+
+    test('should match snapshot', () => {
```
hmhealey, line 26: Is there something specific we can look for in these tests like the existence of some component or how the props are passed into children? Just confirming that they match some snapshot doesn't do much since, if they break in the future, someone will likely just run the tests with `--update` without reading what actually changed.

components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.tsx
```diff
@@ -0,0 +1,119 @@
...
+import SidebarMenu from 'components/sidebar/sidebar_menu';
+import Menu from 'components/widgets/menu/menu';
+
+type Props = {
+    category: ChannelCategory;
+    canCreatePublicChannel: boolean;
+    canCreatePrivateChannel: boolean;
+    onToggle: (open: boolean) => void;
+
+    //currentUserId: string;
```
hmhealey, line 19: You forgot some commented out code

components/sidebar/sidebar_channel/sidebar_channel_menu/index.ts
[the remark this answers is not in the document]
hmhealey: Not normally, but it could theoretically happen since `getCurrentTeam` just returns `teams[currentTeamId]` which could possibly be undefined, but TypeScript seems to automatically type that as `Team` instead of `Team | undefined` like I'd expect. Teams is using the definition
```typescript
export type RelationOneToOne<E extends {id: string}, T> = {
    [x in $ID<E>]: T;
};
```
so I guess that definition means that every possible ID should map to a `Team`

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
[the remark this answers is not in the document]
hmhealey: Either redux or in `actions/channel_actions` in the web app

components/sidebar/sidebar_category/sidebar_category.tsx
devinbinnie: The problem here is actually `injectIntl`, since it doesn't return a wrapped class, it returns an object. Nonetheless, this is just used to type the object and since the class not wrapped by redux contains all the fields I need anyways (since we define the props and state in there), it works.

components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.test.tsx
devinbinnie: Most of the behaviour wrt. what's rendered into the snapshot is covered by the `baseProps`, so I think this test is still valuable. I don't think we need to also check for component existence as well, that just does what `toMatchSnapshot` already does.

We need to just be dilligent with PRs and make sure people aren't updating snapshots without checking if they changed something major in the component.

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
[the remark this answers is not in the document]
devinbinnie: I'll change over to `copyToClipboard`

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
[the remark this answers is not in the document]
devinbinnie: > @devinbinnie, I started looking at this on my iPad Pro. Here are a couple things I found:
> 
> * With no hover on touchscreen tablets, obviously none of the `...` menu’s are visible until you tap on a category header or channel. We may need to default show all menus on touch screens. @andrewbrown00, thoughts?
> * Unfortunately there are some z-index and overflow issues with the `...` menus. :(
> 
> image
> 
> image

Fun, seems to be an iOS only issue currently cause I can't reproduce on Safari on Mac. Time to boot up the simulator!

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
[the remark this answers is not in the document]
hmhealey: Strange. I wonder what the difference between MenuItem and MenuItemAction is then? Why have both of them

devinbinnie:
> @devinbinnie, I started looking at this on my iPad Pro. Here are a couple things I found:
> 
> * With no hover on touchscreen tablets, obviously none of the `...` menu’s are visible until you tap on a category header or channel. We may need to default show all menus on touch screens. @andrewbrown00, thoughts?
> * Unfortunately there are some z-index and overflow issues with the `...` menus. :(
> 
> image
> 
> image

Weird, I don't see this on any desktop browsers...time to open the simulator!

sass/layout/_sidebar-left.scss
hmhealey: We could move this into the same folder as the rest of the `SidebarMenu` code at least even if some of it will be moved out to be with the proper menu components later on. It's more about making the styles easy to find and work with rather than refactoring the menus at this point.

components/widgets/menu/menu_wrapper.tsx
```diff
@@ -77,12 +78,23 @@ export default class MenuWrapper extends React.PureComponent<Props, State> {
...
+         * version (ie the one that uses a modal) needs propagation to close the modal after selecting something
+         * We need to refactor this so that the modal is explicitly closed on toggle, but for now I am aiming to preserve the existing logic
+         * so as to not break other things
+        **/
+        if (this.props.stopPropagationOnToggle) {
+            e.preventDefault();
+            e.stopPropagation();
         }
+        const newState = !this.state.open;
+        this.setState({open: newState}, () => {
```
hmhealey, line 93: Why does this need to wait until after the state has been updated to call `onToggle`?

components/sidebar/sidebar_category/sidebar_category.tsx
hmhealey: In that case, I'm surprised it actually works though when the type here is set to the inner type, but the ref is attached to the wrapper.

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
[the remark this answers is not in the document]
hmhealey: Looking again, I think the `Menu.ItemAction` might actually have a `show` prop that it gets from the `menuItem` wrapping it. I'd need to better understand why it works that way, but it looks like it should at least have that prop

components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.test.tsx
devinbinnie: It's not likely anyone touching this component won't have some understanding of what it does. It's not a global component being re-used by other parts of the app.

If a snapshot is going to break on this component, it'll only be from editing the component itself. I think whoever would see this snapshot will have enough info to know if they've broken something, it's pretty clear if the snapshot no longer renders the menu or one of the menu items.

If people are rubber-stamping without looking at any of the snapshot changes, that's a failure of the PR review system.

I can add some comments here or something to make sure other developers know what to look for if that would ease your mind, but I think we're being overzealous here on something that's just a verification step. Other tests will also be covering a lot of this behaviour, so I think we have enough safe-guards here.

sass/layout/_sidebar-left.scss
devinbinnie: Can we do this as a later step perhaps? I'd like to first understand what needs to be refactored out, and since we're taking the steps to fix the menu sooner rather than later, I don't want to have to do my work twice.

components/widgets/menu/menu_wrapper.tsx
devinbinnie: I don't think there's a requirement to, is this a bad pattern or something?

components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.test.tsx
hmhealey: We're not the only ones that look at and modify this code, especially since we're open source, but even if we were, we're not going to remember how many of the details of this work in a few months. Clear tests like "when a user does X, Y happens" help to document how this is expected to work and ensures that the important parts of a component continue to work. Snapshot tests aren't very helpful since they fail as soon as anything changes, and when that happens, they don't do a good job saying what did change without combing through a hundred or more lines of snapshot.

components/widgets/menu/menu_wrapper.tsx
hmhealey: It's not a bad pattern to use it when it's needed, but we shouldn't use it unless it's actually necessary since it complicates code flow and may cause difficult to understand bugs like wrapping the callback in a `setTimeout(..., 0)` would.

components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.test.tsx
devinbinnie: I can remove the snapshot test if you'd like. There's not much else really to test other than setting different `props` to the component to see which menu items show up. I think the snapshot is useful for that.

components/widgets/menu/menu_wrapper.tsx
devinbinnie: Sure, I can remove it then

devinbinnie: Looked into it, it actually is required so that the menu renders as open and then the `SidebarMenu` can move the menu to the correct position.

sass/layout/_sidebar-left.scss
hmhealey: It's not a matter of major refactoring, it's just to move the newly added styles to some `.../sidebar_channel_menu/styles.scss` and import that in the component instead of adding it as part of the `sass` folder. Doing that doesn't change the names or anything, but it helps keep the CSS nearer to the code that should be affected by it.

devinbinnie: Is that a pattern we use often? I've never seen any CSS files outside the `sass` directory.

andrewbrown00:
@deanwhillier and @devinbinnie 

> * With no hover on touchscreen tablets, obviously none of the `...` menu’s are visible until you tap on a category header or channel. We may need to default show all menus on touch screens. @andrewbrown00, thoughts?

For now, I would say we don't show the `...` unless a user taps to select a channel or the category headers. Having the `...` always visible will add a lot of noise in the UI.

devinbinnie:
@deanwhillier @andrewbrown00 I've added a temporary look-and-feel for menus on the mobile view. Currently it's going to show the menu buttons always. We also have some issues wrt. the sub-menu modal, apparently it's not working correctly at all. I have a version of it working right now but it's not remotely ideal.

We should chat about this soon since this is a miss and needs a plan to be handled.

andrewbrown00:
@devinbinnie let's sync up early next week. Will you throw something on our calendars.

components/sidebar/sidebar_menu.tsx
```diff
@@ -0,0 +1,155 @@
...
+        this.menuButtonRef = React.createRef();
+    }
+
+    // TODO: Temporary code to keep the menu in place while scrolling
+    componentDidMount() {
+        const scrollbars = document.querySelectorAll('#SidebarContainer .SidebarNavContainer .scrollbar--view');
+        if (scrollbars && scrollbars[0]) {
+            scrollbars[0].addEventListener('scroll', this.setMenuPosition);
+        }
+    }
```
deanwhillier, line 51: 2/5 we should instead disable scrolling (and even interaction with the sidebar) while a menu is open. I know we're going to be having a conversation about menus in general, so just adding as a comment for later reference.

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -41,17 +42,20 @@ type Props = {
     ariaLabel?: string;
     root?: boolean;
     show?: boolean;
+    direction?: 'left' | 'right';
```
deanwhillier, line 45: Should we make consistent with the `openUp` prop and use `openLeft` instead? Both are defining direction.

devinbinnie: We actually need to rework all this anyways, will discuss outside of this PR.

deanwhillier: Yup, no worries.

components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.test.tsx
devinbinnie: I've added some additional checks to the snapshot tests.

sass/layout/_sidebar-left.scss
devinbinnie: We will move this out once we look into refactoring the menu.

PR #5379, opened 2020-04-22
MM-24378 MM-24430 Inteactive dialog improvments

crspeller:
#### Summary
- Interactive dialogs will focus the first field allowing for immediate text entry if the first field is a text field. 
- Interactive dialogs will submit on enter if enter pressed in a text field. 
- Focus will be on the submit button if there are no fields in the interactive dialog, allowing immediate submission. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24378
https://mattermost.atlassian.net/browse/MM-24430


components/interactive_dialog/interactive_dialog.jsx
```diff
@@ -200,31 +200,33 @@ export default class InteractiveDialog extends React.Component {
...
+                                    id='interactiveDialogModalIntroductionText'
+                                    value={introductionText}
+                                    emojiMap={this.props.emojiMap}
+                                />
+                            )}
+                            {elements &&
+                            elements.map((e, index) => {
                                 return (
                                     <DialogElement
+                                        autoFocus={index === 0}
```
cpoile, line 229: Nice.

cpoile:
@crspeller Not sure who from QA should be assigned?

saturninoabril:
@crspeller Sorry for the delay, just need to prioritize other issue.

I did testing on the ff:
- dialog without elements (cancel and submit only) - focus not on either button
- dialog with boolean element only and cancel/submit buttons - focus not on those items
- dialog with several elements - no focus inside the modal

Key press (enter or other) after opening a modal does nothing.

Also, we have E2E for interactive dialog at https://github.com/mattermost/mattermost-webapp/tree/master/e2e/cypress/integration/interactive_dialog. Would be good to add related test or I can add separately if you wish.

crspeller:
@saturninoabril Not entirely sure on how to parse your report there. 

The bulleted points sound like what I would expect. 

> Key press (enter or other) after opening a modal does nothing.

In what context? If there are no elements or if there is a text element, enter should submit the modal.

> Would be good to add related test or I can add separately if you wish.

I don't have experience with our procedures around e2e testing. Does QA typically add the tests to the PR or separately?

saturninoabril:
1. Dialog without elements (cancel and submit only) - enter didn't submit the modal
Screen Shot 2020-04-29 at 12 31 32 AM

2. Dialog with boolean element only and cancel/submit buttons - I was expecting for the checkbox to have focus, and on keypress to spacebar, I thought the checkbox will toggle but it didn't happen.
Screen Shot 2020-04-29 at 12 29 40 AM

3. Dialog with several elements - on press to letters/numbers, nothing happened.  I thought that should type into the first textbox.
Screen Shot 2020-04-29 at 12 43 46 AM

For writing E2E, we are requesting for the submitter to add it but we can help out either on the PR itself or separately.

crspeller:
@saturninoabril

1. I do not see this. 
2. I added code to make this happen (not part of the original ticket but it's easy)
3. When I test this it will focus the first textbox. 

I noticed when trying to write cypress tests that cypress does not seem to handle the focus properly. It also doesn't seem to handle pressing enter properly. I see 1, and 3 but only in cypress. I will leave that problem to you.

PR #5385, opened 2020-04-23
MM-23665: Read channel right before leaving it

agarciamontoro:
#### Summary

When switching channels, we need to mark the one we are leaving as read as soon as possible, to avoid race conditions when new posts are arriving.

The problem this PR solves is the following:

Users A and B are in channel X. User B sends a message and as soon as they sent it, user A clicks on channel Y to switch to it. The webapp receives the post in channel X, but it's not added into the posts list before user A leaves.
User A then sees channel X marked as unread, but a few milliseconds later, the unread mark is gone.  This happens because we mark channels X and Y as read at the same time, when channel Y has already been loaded and the new post has already been counted into the total messages of channel X. So we incorrectly mark the new post as read.

The solution is simple: mark channel X as read right before leaving it, not after having switched to Y.

Publishing as a draft PR because I am not sure of the implications of my changes, both performance-wise and about possible side-effects.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23665

#### Related Pull Requests
None

#### Screenshots
None


components/post_view/post_list/post_list.jsx
```diff
@@ -147,6 +147,7 @@ export default class PostList extends React.PureComponent {
     }
 
     componentWillUnmount() {
+        this.markChannelAsReadAndViewed(this.props.channelId);
```
hmhealey, line 150: I'm trying to think if there's anything else that could cause this component to unmount that wouldn't represent the user having interacted with the app, but I can't think of anything that wouldn't already happen with the current logic.

hmhealey:
This code looks good, although we might want to bump it to 5.24 since this is a pretty critical piece of functionality, and that would give it more time to soak on community-daily

agarciamontoro:
> we might want to bump it to 5.24

Fine for me. This issue has been around for a long long time (it reproduces in 5.16, at least) and is not critical, but any error in this PR may be.

I'll mark this as ready for review.

components/post_view/post_list/post_list.jsx
```diff
@@ -205,15 +206,11 @@ export default class PostList extends React.PureComponent {
...
-    markChannelAsReadAndViewed = () => {
-        const currentChannelId = this.props.channelId;
-        const prevChannelId = this.props.prevChannelId;
-
-        // Mark previous and next channel as read
+    markChannelAsReadAndViewed = (channelId) => {
         // Posts are marked as read from here to not cause a race when loading posts
         // marking channel as read and viewed after calling for posts in channel
-        this.props.actions.markChannelAsViewed(currentChannelId, prevChannelId);
-        this.props.actions.markChannelAsRead(currentChannelId, prevChannelId);
```
sudheerDev, line 216: i dont  `prevChannelId` used anymore. Can you remove the reference from this file and also from the connected component and the parent components state etc? Search for `prevChannelId` and you should see references for it.

agarciamontoro: Thank you for pointing it out, @sudheerDev, there was quite a lot of dead code around this!

agarciamontoro:
Re-requesting reviews because I've deleted a bunch of dead code and I want more eyes on this to double-check it was actually dead code. Thank you!

jgilliam17:
Thank you @agarciamontoro 
I am seeing few errors on the PR server on Chrome, not sure if they are related to this fix
Screen Shot 2020-05-11 at 10 31 51 AM

There is also an issue with manually marking posts as unread.
I am expecting a channel with the posts I just marked as unread to remain Unread when I navigate away to a different channel. But on the test server the Unread status does not persist, in this case, channel is marked as unread as soon as I click on a different channel.

agarciamontoro:
Thank you for testing this, @jgilliam17!

I'll review the errors you shared in the screenshot. I'd say that they are not related to these changes, but I'm handwaving here, I need to check.

Regarding the marking messages as unread, I see how this could be breaking that. Let me investigate this. Thank you for bringing this up!

agarciamontoro:
The last commit should fix the issue with the manually marking the channel as unread.

Regarding the console errors: I was not able to reproduce them. Please tell me if you still see them and, in case they do, when they show up.

Thank you, @jgilliam17!

agarciamontoro:
Fixed failing tests (forgot to add the new prop there!) and the failing builds (master was unstable). Ready to go, @jgilliam17 :rocket:

jgilliam17:
Thank you @agarciamontoro 
The console error I am seeing shows up on Chrome only(not seeing this on Firefox), every time I switch channels, something about channel member count and timezones.
Screen Shot 2020-05-14 at 10 58 22 AM

I am still seeing the issue with marking posts as unread on the test server - the setting does not persist when I switch channels.

agarciamontoro:
@jgilliam17 Looks like the test server did not update correctly. It's on 26ae698c5fb87a20348760b62e32c977ce8d04cc Should we restart it?
image

agarciamontoro:
Closing this for the moment. Reasons described in the ticket: https://mattermost.atlassian.net/browse/MM-23665?focusedCommentId=88206

agarciamontoro:
Reopening after @devinbinnie's work on this, that confirmed that this PR does not fix the issue completely but it at least mitigates it. See [his comment on the ticket](https://mattermost.atlassian.net/browse/MM-23665?focusedCommentId=88439).

PR #5387, opened 2020-04-23
MM-24381 - Improve channel header button options

cpoile:
#### Summary
- Plugins can now register their channel header button in `registerChannelHeaderButtonAction` with a priority flag. If set to true, it will put the button in a new array in the store, and those buttons will be rendered to the right of the search box. 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24381

#### Screenshots
- The incident response plugin (the exclamation point) without the priority set:
image
- With the priority set:
image

### DIFF
--- a/components/channel_header/channel_header.js
+++ b/components/channel_header/channel_header.js
@@ -87,6 +87,7 @@ class ChannelHeader extends React.PureComponent {
         currentRelativeTeamUrl: PropTypes.string.isRequired,
         newSideBarPreference: PropTypes.bool,
         announcementBarCount: PropTypes.number,
+        priorityPluginButtons: PropTypes.array,
     };
 
     constructor(props) {
@@ -280,6 +281,20 @@ class ChannelHeader extends React.PureComponent {
 
     handleFormattedTextClick = (e) => Utils.handleFormattedTextClick(e, this.props.currentRelativeTeamUrl);
 
+    createPriorityPluginButton = (plug) => {
+        return (
+            <HeaderIconWrapper
+                key={'channelHeaderButton' + plug.id}
+                buttonClass='channel-header__icon style--none'
+                iconComponent={plug.icon}
+                onClick={() => plug.action(this.props.channel, this.props.channelMember)}
+                buttonId={plug.id}
+                tooltipKey={'plugin'}
+                tooltipText={plug.tooltipText ? plug.tooltipText : plug.dropdownText}
+            />
+        );
+    }
+
     render() {
         const {
             teamId,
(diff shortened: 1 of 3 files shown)


asaadmahmood:
@cpoile Actually, we want to move them to the far left, because then the search bar, mentions, pinned, or the channel member icons won't change position based on how many plugins one has. They will remain in their place, and the plugins will be appended to the left.

cpoile:
@asaadmahmood But... That's not how the mockups are https://www.figma.com/file/krMfTWyPSmh5J8EpV20ceS/Incident-Response-(Current)?node-id=1361%3A6222 , and that's what made us create the ticket: https://mattermost.atlassian.net/browse/MM-24381

asaadmahmood:
@cpoile Yup, sorry about that, but I previously thought that the incident icon was going to be a global icon.
The icons on the right of the channel are global actions (i.e. mentions, flags), while the ones on the left of the searchbar are channel specific.

But either way, we decided to move them on the left to keep the button positions consistent.
Let me know if that would be hard to do.

cpoile:
@asaadmahmood I was unaware of that decision... Can you show me an example of what you mean? 

Just for clarity -- the incident channel icon is global (it's not channel specific), so shouldn't it belong on the right of the search bar?
Second: if it's on the left of the search bar, it will be moved when the search bar expands (and you can't click it again to close it, because it just moved). 

I would recommend trying it out first -- the UX is much better this way.

asaadmahmood:
@cpoile If we go with my initial decision, which is to place global icons on the left, then we would have to dislocate channel plugins and global plugins.
Channel plugins would go on the left, global on the right.
Global plugins on the left would most probably push the icons (even the normal ones, pinned, channel members) on the left, and thus change their position default position.

And the issue you have with the icon jumping around when the search bar expands, that would still be faced by channel level pluguins.

So I am still leaning towards keeping all the plugins together, and to the left.

asaadmahmood:
We are prioritising UX on having a dedicated space for the plugins in a RHS bar, so hopefully your concern about stuff jumping around would be solved there.

cpoile:
@asaadmahmood Sorry, I'm getting a little confused -- you just said global icons (mentions, flags, etc.) are on the right of the search bar, correct? The incident side bar is a global icon. Shouldn't it be on the right of the search bar?

cpoile:
Chatted with Asaad on zoom -- a bit of context that I was missing: the design team doesn't want to allow plugins to register icons to the right of the search bar because it will push the search bar an unknown amount to the left, and then the right margin of the search bar will not align with the left margin of the expanded RHS window. 
I suppose that makes this PR moot. :). I will close.

PR #5388, opened 2020-04-23
MM-24380 Allowing plugins to hide the RHS header when displaying a custom RHS.

crspeller:
#### Summary
Allows plugins to suppress the default RHS header by passing `null` as the second parameter to the registration function instead of a title. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22917

### DIFF
--- a/plugins/rhs_plugin/rhs_plugin.jsx
+++ b/plugins/rhs_plugin/rhs_plugin.jsx
@@ -18,14 +18,21 @@ export default class RhsPlugin extends React.PureComponent {
     }
 
     render() {
+        let header = null;
+        if (this.props.title !== null) {
+            header = (
+                <SearchResultsHeader>
+                    {this.props.title}
+                </SearchResultsHeader>
+            );
+        }
+
         return (
             <div
                 id='rhsContainer'
                 className='sidebar-right__body'
             >
-                <SearchResultsHeader>
-                    {this.props.title}
-                </SearchResultsHeader>
+                {header}
                 <Pluggable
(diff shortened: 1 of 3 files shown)


plugins/rhs_plugin/rhs_plugin.jsx
```diff
@@ -18,14 +18,21 @@ export default class RhsPlugin extends React.PureComponent {
     }
 
     render() {
+        let header = null;
+        if (this.props.title !== null) {
+            header = (
+                <SearchResultsHeader>
+                    {this.props.title}
+                </SearchResultsHeader>
+            );
```
marianunez, line 27: I think we shouldn't couple the title with completely hiding the header. Removing the header removes the option to close and expand the RHS. Not all plugins use the toggle feature, they can just use open the RHS and then they won't have a way to close it. 

If my plugin doesn't want a title but wants to keep the close button?

crspeller: `registerRightHandSidebarComponent(component, '')` and
`registerRightHandSidebarComponent(component)`
Will both work.

crspeller:
@marianunez Given the requirements outlined in the figma for incident response. I don't think we are going to get away with anything other than a total override.

lieut-data:
We discussed this proposal today at the workflows architecture sync-up, and decided to pause until we can sync with the broader UX plan.

crspeller:
Closing as UX has changed it's mind on this.

PR #5390, opened 2020-04-23
[MM-21938, MM-22819, MM-22855] Toast dates, Date Separator; Post time tooltip

calebroseland:
#### Summary

- Unread Toast: friendlier date formats and time removed (desktop view; mobile view omits the date)
- Date Separator: friendlier date formats (center & rhs in desktop & mobile views)
- New post_time (permalink) tooltip with same simplified format as other date/timestamps.
- New `<Timestamp/>` component for unified handling of date and time formatting
- E2E: `date_separator_spec` updated for additions/changes.

| Time Period | Localized Date Format |
| --- | --- |
| Today | [Today] |
| Yesterday | [Yesterday] |
| Last 6 days | {weekday} |
| Current year | {month, day} |
| Last year or older | {month, day, year} |

#### I18n: changes to:

- `postlist.toast.newMessages`
- `postlist.toast.newMessagesSince`

#### Ticket Links

- https://mattermost.atlassian.net/browse/MM-21938
- https://mattermost.atlassian.net/browse/MM-22819
- https://mattermost.atlassian.net/browse/MM-22855

#### Screenshots

(Format breakpoints in red)
date formats

scrolling

Screen Shot 2020-04-23 at 16 05 00
Screen Shot 2020-04-23 at 16 03 26
Screen Shot 2020-04-23 at 16 10 27

**Post time tooltip**, uses same format as specified in [MM-22855](https://mattermost.atlassian.net/browse/MM-22855).
Screen Shot 2020-07-17 at 14 18 46

#### Relative Time (currently only used for Today and Yesterday)
If `<Timestamp>`'s relative time options are used, the difference is automatically calculated from the given value and is with deeply configurable with its breakpoints (`ranges`) and outputs.

This breakpoint configuration: 
```ts
const ranges = [
    {within: ['second', -15], display: ['second', 0]}, // second, 0 is equal to 'now'
    {within: ['second', -54], display: ['second', 10]}, // round to nearest 10 seconds
    {within: ['hour', -1], display: ['minute']},
    {within: ['hour', -24], display: ['hour']},
];
```
Produces the following flow:

now
20 sec. ago
30 sec. ago
40 sec. ago
50 sec. ago
1 min. ago
2 min. ago
...
59 min. ago
1 hr. ago
2 hr. ago
... 
23 hr. ago
(then reverts to fixed time)

You can also simply specify a unit and it will run the diff calc for you. 

```tsx
<Timestamp
    value={new Date(/* three days ago */)}
    useTime={false}
    unit='day'
/>
//-> '3 days ago'

<Timestamp
    value={new Date(/* yesterday, xx:xx */)}
    useTime={true}
    unit='day'
/>
//-> 'yesterday at 3:21 PM'

```


calebroseland:
@esethna @andrewbrown00 cc @deanwhillier 
Adding ordinal days would be nice, but we'd have to bring our own localization of `Intl.PluralRules` and probably have to DIY the implementation, as Intl.DateTimeFormat doesn't natively support ordinals (yet?). (https://github.com/formatjs/react-intl/issues/1043)

We could possibly take a step back and write up a more capable date-time component that could support all these different date-time formats and use cases.

Let me know what you guys think.

esethna:
Thanks @calebroseland, I think what you have is fine in that case, can you help create a follow-up ticket for the above and set for app grooming to discuss?

components/toast_wrapper/toast_wrapper.jsx
```diff
@@ -216,20 +216,14 @@ class ToastWrapper extends React.PureComponent {
             return (
                 <FormattedMessage
                     id='postlist.toast.newMessagesSince'
-                    defaultMessage={'{count, number} new {count, plural, one {message} other {messages}} since {date} at {time}'}
+                    defaultMessage='{count, number} new {count, plural, one {message} other {messages}} {isToday, select, true {} other {since}} {date}'
```
hmhealey, line 219: Have we used `select` anywhere else in translation strings? I'm wondering if we need to add it to some documentation somewhere so that translators know how to translate it. I [started a thread in the I18n channel](https://community-daily.mattermost.com/core/pl/915d8y916tgyjjcn1kwttax1oe) to ask if we have any docs on those.

components/recent_date.tsx
```diff
@@ -14,37 +15,70 @@ type Props = {
...
+                    />
+                );
+            }
         }
 
-        const options = {
-            timeZone,
-            weekday: 'short',
-            month: 'short',
+        const options: Intl.DateTimeFormatOptions = dateTimeFormat ?? is.withinPastSevenDays ? {
```
hmhealey, line 69: Could you maybe wrap the first part of this in brackets? Mixing `??` and a ternary statement is a bit difficult to read.

```suggestion
        const options: Intl.DateTimeFormatOptions = (dateTimeFormat ?? is.withinPastSevenDays) ? {
```

calebroseland: Thanks for catching that @hmhealey... I clarified it and added a test for the that condition.

components/recent_date.tsx
```diff
@@ -14,37 +15,72 @@ type Props = {
...
-        const options = {
-            timeZone,
-            weekday: 'short',
-            month: 'short',
-            day: '2-digit',
-            year: 'numeric'
-        };
-        const formattedDate = this.props.intl.formatDate(value, options);
+        const options: Intl.DateTimeFormatOptions = dateTimeFormat ?? (
+            is.withinPastSixDays ? {
```
sudheerDev, line 70: Nit: This seems a little long. I am 0/5 though. You call.

sudheerDev:
@calebroseland I think this also affects the floating timestamp used in the mobile view of the app. Could you just make the changes to the description if that is true so QA can verify changes

components/toast_wrapper/toast_wrapper.jsx
calebroseland: @hmhealey I think this is the first `select` usage. I agree we should add docs... what if we added a Message Syntax section to https://docs.mattermost.com/developer/localization.html with a 1-paragraph overview with links to https://formatjs.io/docs/icu-syntax and https://formatjs.io/docs/react-intl ?

components/recent_date.tsx
```diff
@@ -62,14 +98,35 @@ class RecentDate extends React.PureComponent<Props> {
...
 }
 
-export function isSameDay(a: Date, b: Date) {
-    return a.getDate() === b.getDate() && a.getMonth() === b.getMonth() && a.getFullYear() === b.getFullYear();
+function fromNow(a: Date) {
+    const now = new Date();
+    const today = isSameDay(a, now);
+
+    return {
+        is: {
```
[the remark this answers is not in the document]
calebroseland: I was going back and forth on that, 0/5 either direction, I'll go ahead and simplify the destructuring.

components/toast_wrapper/toast_wrapper.jsx
hmhealey: Yeah, that would be great. The source for those docs live [here](https://github.com/mattermost/docs/blob/master/source/developer/localization.rst) if you could make a PR

components/recent_date.tsx
hmhealey: Can you break this up to use an if statement? I actually thought we didn't allow nested ternary statements

components/toast_wrapper/toast_wrapper.jsx
calebroseland: [Done](https://github.com/mattermost/docs/pull/3590) ✓

calebroseland:
@saturninoabril Thanks for pointing this out—I'll add these.

hmhealey:
We could also use regular Enzyme unit tests for this instead of Cypress ones if you want since it's mostly around how things display rather than testing interaction with the components.

calebroseland:
@hmhealey that was my original thought, and there are these tests in `recent_date.test.tsx` to cover the changes, but there does already appear to be Cypress tests in `date_separator_spec.js`, so I'll just add/adjust these to match the new date ranges.

esethna:
Hey @calebroseland, just curious on next steps here?

calebroseland:
> Hey @calebroseland, just curious on next steps here?

Hey @esethna, I'm just about done wrapping up the tests for a more robust component to handle time formatting more consistently. Also adding https://mattermost.atlassian.net/browse/MM-22855 as it's a near cousin to the existing two stories.

e2e/cypress/integration/messaging/local_date_time_spec.js
```diff
@@ -77,10 +77,10 @@ describe('Messaging', () => {
...
                 manualTimezone: 'NZ-CHAT',
                 localTimes: [
-                    {postIndex: 0, standard: '06:15 PM', military: '18:15'},
-                    {postIndex: 1, standard: '02:15 AM', military: '02:15'},
-                    {postIndex: 2, standard: '10:15 AM', military: '10:15'},
-                    {postIndex: 3, standard: '02:15 PM', military: '14:15'},
+                    {postIndex: 0, standard: '6:15 p. m.', military: '18:15'},
+                    {postIndex: 1, standard: '2:15 a. m.', military: '2:15'},
+                    {postIndex: 2, standard: '10:15 a. m.', military: '10:15'},
+                    {postIndex: 3, standard: '2:15 p. m.', military: '14:15'},
```
saturninoabril, line 83: I've ran the test and the changes here failed. Please take a look at https://mm-cypress-report.s3.amazonaws.com/663-webapp-pr-5390-1361efd/mochawesome.html

calebroseland: Hmmm, okay.  These E2E tests run fine in Chrome, FF, and Edge, but not Electron.

I think there are discrepancies between CI and runtime environments regarding timezones supported by Intl.DateTimeFormat.

All current versions of Chromium/FF/Safari/Edge have full `Intl.DateTimeFormat` support for all timezones that `moment-timezone` supports. In these environments (`cypress:run`, `cypress:run:firefox`), `NZ-CHAZ` is supported and proceeds to get translated to `es`. 

In other environments (Electron, Jest, etc.) `NZ-CHAZ` and other timezones are not supported and subsequently fail, which causes the fallback `moment-timezone` formatter to kick in, which does not have translations—resulting in `en`.


What do you think, should I loosen the check here to accept both `en` and `es`? (Hopefully at least until we can unify CI/runtime `Intl` support?)

calebroseland: I went ahead and switched to `en` as the test case is validating unsupported timezone and `es` is checked in the case above.

saturninoabril: Our CI runs in chrome 83 (headless) currently. Not yet running on Firefox 'cause there's still an issue with it, nor with Edge. I'll try to run again then see what's going on.

cpanato:
test server removed due some underlying issues in the cluster that made this one unstable.

components/post_view/date_separator/index.js
```diff
@@ -1,30 +1,4 @@
...
-    }
-
-    return {
-        enableTimezone: areTimezonesEnabledAndSupported(state),
-        timeZone: getUserCurrentTimezone(userTimezone),
-    };
-}
-
-export default connect(mapStateToProps)(DateSeparator);
+export {default} from './date_separator';
```
sudheerDev, line 4: Why do we need this file? Can't we directly use `./date_separator` file?

sudheerDev: I thought i had an outstanding review on this one. Really minor nit anyway

calebroseland: Thanks!

> Why do we need this file? Can't we directly use `./date_separator` file?

Yeah, 0/5 I think it's generally better to always have an index, but I could go either way.

hmhealey: We've written that as
```javascript
import DateSeparator from './date_separator';
export default DateSeparator;
```
elsewhere, but I think that's because we didn't yet support export from if I remember correctly. I'm 0/5 on whether or not we move the file or reexport from an index file though. As long as we're using importing it as `components/post_view/date_separator` everywhere, it won't matter to any calling code

PR #5393, opened 2020-04-24
[MM-24150] [MM-24357] Update special mentions regex to better match server notification behaviour

fmunshi:
#### Summary
- The current special mentions regex has the potential to conflict with LDAP group names such as `@all-developers` or `@all.developers`, technically there is even the possibility the conflict with user names such just as `@here.smith` although less likely still possible.
- The behaviour on the server has always been to not treat `@all-developers` as an `@all` mentions so this PR simply updates the regex to better match what the server does and also moves the regex into a the constants file.
- Also adds more tests to the `text_formatting` file to cover more cases

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24510
- https://mattermost.atlassian.net/browse/MM-24357

#### Screenshots
Screen Shot 2020-04-23 at 11 03 15 PM


utils/constants.jsx
```diff
@@ -787,7 +787,9 @@ export const Constants = {
     },
 
     SPECIAL_MENTIONS: ['all', 'channel', 'here'],
+    SPECIAL_MENTIONS_REGEX: /\B@(channel|all|here)(?!(\.|-|_)*[^\W_])/gi,
```
Willyfrog, line 790: the `(?!(\.|-|_)*[^\W_])` part seems a bit complex, is it trying to say that only \w caracters are allowed? although testing it on an online regex does have a match with `@all---` which I'm not sure that's intended.

fmunshi: That is intentional, the idea is that `@all` can be followed up by any number of `-` `.` `_`  as long as they are not followed by a word character (excluding underscores).

So the following should be valid according to this regex
`@here....._` 
`@all-----_` 
`@all_______` 
But the following should not be valid
`@all------s`
`@all_______s`
`@all.........s`

Is there a simpler way to achieve that last bit? this was the cleanest solution that I could come up with

Willyfrog: if that's intentional, that's good enough, then :D

PR #5394, opened 2020-04-24
MM-24487 - Updating icons in invite modal dropdown

asaadmahmood:
#### Summary
MM-24487 - Updating icons in invite modal dropdown

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24487

#### Related Pull Requests
Screenshot 2020-04-24 at 2 28 50 PM

### DIFF
--- a/components/suggestion/search_channel_with_permissions_provider.jsx
+++ b/components/suggestion/search_channel_with_permissions_provider.jsx
@@ -14,9 +14,6 @@ import {Permissions} from 'mattermost-redux/constants';
 import {sortChannelsByTypeAndDisplayName} from 'mattermost-redux/utils/channel_utils';
 import {logError} from 'mattermost-redux/actions/errors';
 
-import GlobeIcon from 'components/widgets/icons/globe_icon';
-import LockIcon from 'components/widgets/icons/lock_icon';
-import ArchiveIcon from 'components/widgets/icons/archive_icon';
 import store from 'stores/redux_store.jsx';
 import {Constants} from 'utils/constants';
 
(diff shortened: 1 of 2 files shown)


components/suggestion/search_channel_with_permissions_provider.jsx
```diff
@@ -44,15 +41,15 @@ class SearchChannelWithPermissionsSuggestion extends Suggestion {
         let icon = null;
         if (channelIsArchived) {
             icon = (
-                <ArchiveIcon className='icon icon__archive'/>
+                <i className='icon icon--no-spacing icon-archive-outline'/>
```
jespino, line 44: Are we going back to font-awesome with this icons? Shouldn't we use the icon widgets for this, because if not in the future we have to modify each one of this, instead of modifying only the component one.

jgilliam17:
Thank you @asaadmahmood 
- Spacing is present but it appears a bit wider than currently on community 5.22.
- There's an issue where channel icon is missing or glitching with max or close to max character count
Screen Shot 2020-04-24 at 5 10 51 PM

components/suggestion/search_channel_with_permissions_provider.jsx
asaadmahmood: @jespino No, this is the mattermost icon font. From now on, we'll try to use the mattermost icon font in the webapp instead of the svgs.

jespino: Great, still I think have value to wrap it in a component. But that is another subject. Thanks for the explanation.

PR #5396, opened 2020-04-25
Use inline images for gif picker

bbodenmiller:
#### Summary
Move to use inline images for gif picker. Inserting the raw link is ugly and not user friendly.

Needs further testing from someone else.

### DIFF
--- a/components/create_comment/create_comment.jsx
+++ b/components/create_comment/create_comment.jsx
@@ -441,7 +441,7 @@ class CreateComment extends React.PureComponent {
 
         let newMessage = '';
         if (draft.message === '') {
-            newMessage = `![](${gif})`;;
+            newMessage = `![](${gif}) `;
         } else if ((/\s+$/).test(draft.message)) {
             // Check whether there is already a blank at the end of the current message
             newMessage = `${draft.message}![](${gif}) `;
--- a/components/create_comment/create_comment.jsx
+++ b/components/create_comment/create_comment.jsx
@@ -441,12 +441,12 @@ class CreateComment extends React.PureComponent {
 
         let newMessage = '';
         if (draft.message === '') {
-            newMessage = gif;
+            newMessage = `![](${gif})`;;
         } else if ((/\s+$/).test(draft.message)) {
             // Check whether there is already a blank at the end of the current message
-            newMessage = `${draft.message}${gif} `;
+            newMessage = `${draft.message}![](${gif}) `;
         } else {
-            newMessage = `${draft.message} ${gif} `;
+            newMessage = `${draft.message} ![](${gif}) `;
         }
 
         const modifiedDraft = {


hanzei:
Thanks for the PR @bbodenmiller :+1: Could you please fix the two linter issue:
```
/root/mattermost-webapp/components/create_comment/create_comment.jsx
 
  444:39  error  Missing whitespace after semicolon  semi-spacing
 
  444:40  error  Unnecessary semicolon               no-extra-semi
 ```

bbodenmiller:
Maybe fixed now? @hanzei can you respin the sever?

bbodenmiller:
@hanzei server spin up didn't work?

esethna:
@bbodenmiller thanks for the PR! Is there a ticket open for this?

My concern with this approach is that the GIFs cannot be collapsed if they are added as in-line markdown images. 

cc// @andrewbrown00

bbodenmiller:
> Is there a ticket open for this?

No. I created a PR to try and address an issue I saw.

> My concern with this approach is that the GIFs cannot be collapsed if they are added as in-line markdown images.

That sounds like an entirely separate issue that should be addressed independently. It's an issue for plugins too e.g. https://github.com/moussetc/mattermost-plugin-giphy/issues/12. In my experience most people are already changing these in the inline image links and this just simplifies the process.

esethna:
Thanks @bbodenmiller 

> That sounds like an entirely separate issue that should be addressed independently. 

I'd be concerned though that this is removing a feature of collapsing GIFs that currently exists. I commented on the thread you mentioned to try and understand the problem: https://github.com/moussetc/mattermost-plugin-giphy/issues/12#issuecomment-620179601

esethna:
@bbodenmiller are you good with closing this as per: https://github.com/moussetc/mattermost-plugin-giphy/issues/12#issuecomment-621387037

bbodenmiller:
@esethna is there any usage analytics data on how often the collapse feature is actually used?

esethna:
@bbodenmiller unfortunately not but given the user feedback in the thread linked I’d be hesitant to make this change until we can offer a collapse for inline images.

andrewbrown00:
Sorry I'm late to replying @bbodenmiller and I agree with @esethna

> I'd be concerned though that this is removing a feature of collapsing GIFs that currently exist

and I agree with you we should better understand the data around how often the collapse feature is used and not having the data supports my decision to not remove the feature at this time.

> is there any usage analytics data on how often the collapse feature is actually used?

bbodenmiller:
@esethna @andrewbrown00 in light of the above how can we move along the ability to collapse inline images? As the gif picker doesn't work at all in my instance because link previews don't work on our instance.

esethna:
Hi @bbodenmiller, I've already queued up the ticket for our design team to investigate. 

Can you explain more about why link previews don't work for your instance?

teresa-novoa:
@esethna where are we with this PR? Thanks!

esethna:
@teresa-novoa as discussed above, I'd propose closing this PR until we can sort out the expected UX for collapsing in-line images (queued with the UX team): https://mattermost.atlassian.net/browse/MM-12290

@bbodenmiller are you good with closing this until we complete the design ticket? ^

bbodenmiller:
> Can you explain more about why link previews don't work for your instance?

@esethna because the Mattermost server does not have internet access and the link previews require internet access. In line images do not require internet access as they are loaded client side.

PR #5398, opened 2020-04-26
MM-24549 Fix for skipping user info calls for hidden GM's

sudheerDev:
#### Summary
This PR fixes the issue with making unnecessary calls or members info in GM channels when hidden in LHS. This only skip calls if experimental sidebar settings are turned on

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24549


actions/user_actions.test.js
```diff
@@ -5,10 +5,14 @@ import thunk from 'redux-thunk';
 import configureStore from 'redux-mock-store';
 
 import {Preferences} from 'mattermost-redux/constants';
+import channelCategories from 'mattermost-redux/selectors/entities/channel_categories';
+import {CategoryTypes} from 'mattermost-redux/constants/channel_categories';
 
 import * as UserActions from 'actions/user_actions';
 
 const mockStore = configureStore([thunk]);
+const mockChannelsObj1 = [{id: 'gmChannel', type: 'G1'}];
```
hmhealey, line 14: If these are supposed to be gm channels, shouldn't those types be using `General.GM_CHANNEL`? `G1` and `G2` aren't valid channel types.

sudheerDev: It was a mock so it wouldn't have mattered all that much. That wasn't great in the way I had to hack in the first place. I spent quite a lot of time trying to figure out how to pass out of scope variable to mock and there is no way to do that.

hmhealey: I'm confused. We're mocking the constants? This test file is a mess, but I didn't know it was that bad.

actions/user_actions.test.js
```diff
@@ -21,6 +26,19 @@ jest.mock('mattermost-redux/actions/users', () => {
     };
 });
 
+jest.mock('mattermost-redux/selectors/entities/channel_categories', () => {
```
hmhealey, line 29: In particular, why mock these selectors?

sudheerDev: If we are calling the selectors then isn't that a good enough test? Selectors are supposed to filter and they have tests to cover that.

deanwhillier:
Just noting that I've seen this, but am removing myself from review as Devin and Harrison have this.

actions/user_actions.test.js
hmhealey: The main thing I'm concerned about is that the tests confirm that the code call the selectors, but it doesn't confirm that calling the selectors actually prevents the extra requests. In the end, they both cover the same thing assuming that the selectors work as expected, so while I personally think that testing for the bug would be better, doing it this way is probably fine as well, so 0/5 on whether or not we change it.

PR #5401, opened 2020-04-27
MM-24490 - Updating UI for link previews name

asaadmahmood:
#### Summary
MM-24490 - Updating UI for link previews name

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24490

#### Screenshots
Screenshot 2020-04-27 at 6 47 22 PM

### DIFF
--- a/sass/components/_post.scss
+++ b/sass/components/_post.scss
@@ -1576,7 +1576,7 @@
         height: 1.6rem;
 
         &.pull-left {
-            margin: 4px 5px 0 0;
+            margin: 2px 0px 0 -4px;
         }
 
         &:hover {


components/post_view/post_body_additional_content/post_body_additional_content.jsx
```diff
@@ -152,13 +152,13 @@ export default class PostBodyAdditionalContent extends React.PureComponent {
...
-            const prependToggle = (/^\s*https?:\/\/.*$/).test(this.props.post.message);
 
             return (
                 <div>
-                    {(toggleable && prependToggle) && this.renderToggle(true)}
-                    {this.props.children}
-                    {(toggleable && !prependToggle) && this.renderToggle(false)}
+                    <div className='file-preview__link'>
+                        {this.props.children}
+                        {toggleable && this.renderToggle(true)}
```
larkox, line 160: Render toggle does not longer receive any value, so we can call it just as `renderToggle()`

components/single_image_view/single_image_view.jsx
```diff
@@ -124,14 +124,14 @@ export default class SingleImageView extends React.PureComponent {
...
-        let imageNameClass = 'image-name';
+        let imageNameClass = 'file-preview__link';
         if (compactDisplay) {
             imageNameClass += ' compact-display';
         }
 
         const fileHeader = (
             <div
-                data-testid='image-name'
+                data-testid='file-preview__link'
```
larkox, line 134: Is this testid being used? After a quick search in the code I found no other references apart of the snapshots.

If it is not being used, I would remove it completely.
If it is being used, the related tests should updated (or at least give a heads up to the test team)

asaadmahmood: I don't really know, but I didn't wanted to mess it up, but makes sense, I'll remove it.

jespino: The references to this where added here 94d285bc648eac5bf71f29ffffb42969e4639d06 and removed here beb80fb7add37d9e7fff355cd5b37bb1f018ed5c so I guess we can remove it.

asaadmahmood:
Done @jespino @larkox

esethna:
@cpanato this test server looks like it's having issues starting up? Can you help take a look? Thanks!

esethna:
@asaadmahmood not sure we want to update the link styling? See cases where this breaks: 
image

asaadmahmood:
Yup @esethna. Looks like there are some other cases that would need a larger refactor, so I think we can skip this for now.
 
Screenshot 2020-04-29 at 2 08 15 PM

esethna:
@asaadmahmood sorry, the bug is still valid (in terms of positioning of the arrow in various cases), I mean we should just avoid changing the styling of the text from a link to being greyed out, since link previews may not appear directly below the link itself

asaadmahmood:
@esethna We can't bring to arrow to the right of the link, based on how the post is structured in these cases.

What apart from bringing the arrow to the right do we want to do?
Should I just keep it on the left and update the color (gray).

esethna:
@asaadmahmood okay then keeping it on the left is fine, but we should fix the alignmnet: 
image

asaadmahmood:
Cool, but we would be inconsistent then with the images, but I'll just fix the alignment in this PR, and we can look at the other stuff later.

asaadmahmood:
@andrewbrown00 We have to enable link previews in the system console first, but I'll do that when I make my change to the PR.

andrewbrown00:
> @andrewbrown00 We have to enable link previews in the system console first, but I'll do that when I make my change to the PR.

Hmm, I did that but the `caret` never showed up. Also, should we change the `font-size` to be consistent with the font size for images?

asaadmahmood:
@andrewbrown00 that would be a bit hard to do, as the link for the preview is not easily distinguishable right now.
It can be the first one, or the middle one in a message.

Screenshot 2020-05-04 at 7 27 53 PM

asaadmahmood:
I've enable the previews and posted a link in the PR. @andrewbrown00

andrewbrown00:
Thanks @asaadmahmood looks good.

image

What's the `dev effort` to [move the icon to the right](https://github.com/mattermost/mattermost-webapp/pull/5401#issuecomment-621394505) to be consistent with images?

asaadmahmood:
@andrewbrown00 Not sure, maybe @hmhealey knows as he worked on the markdown stuff.

esethna:
Can we do anything about this case as part of this ticket for fixing the alignmnet issues? 
image

asaadmahmood:
@esethna Not much, its quite hard, I can move the icon somewhere at the top, but then there may be cases where the preview is being generated by an image that's not at the start but somewhere at the middle, so I would suggest we look at that when we can look at this issue in detail.

lindy65:
Hey @asaadmahmood,

Is the icon supposed to be left-aligned with the start of the username or positioned half-way between the first and second letter of the username as per the screenshot below?

image

Compared to the screenshot ES posted to the ticket as an example of what should be fixed, it seems the icon has been moved only marginally left?  Is this correct?

Screenshot from ticket:
image

asaadmahmood:
@lindy65 Expected, as its a button, so it's not aligned.

asaadmahmood:
And based on the ticket, the main fix is the icon's vertical alignment with the text, it has moved up a bit.

PR #5403, opened 2020-04-27
Highlight hovered items when searching with "in:"

agarciamontoro:
#### Summary
https://github.com/mattermost/mattermost-webapp/pull/4402 introduced a minor regression on the search popover when using the `in:` keyword, as the `onMouseMove` event was not registered in the `SearchChannelSuggestion` component. This PR just adds it back.

I have tried to add a test for this using Cypress, but I have no luck hovering over the item to check that the CSS class is added. Looks like this is a well-known pain point in Cypress; it even has [a dedicated documentation page](https://docs.cypress.io/api/commands/hover.html#Workarounds), but none of the workarounds worked for me. I'd love to hear other ideas for testing this.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24296 (this PR only solves the first issue described in the ticket, I think we should discuss the CSS one separately)

#### Screenshots
Before:
image

After:
image

### DIFF
--- a/components/suggestion/search_channel_suggestion/search_channel_suggestion.jsx
+++ b/components/suggestion/search_channel_suggestion/search_channel_suggestion.jsx
@@ -91,6 +91,7 @@ export default class SearchChannelSuggestion extends Suggestion {
         return (
             <div
                 onClick={this.handleClick}
+                onMouseMove={this.handleMouseMove}
                 className={className}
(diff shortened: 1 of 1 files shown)


agarciamontoro:
Thank you for trying the mouseover thing, it's weird that it does not work!

PR #5405, opened 2020-04-27
MM-24054: Do not populate channels list after user clears the search

catalintomai:
#### Summary
Autocomplete list of channels remains populated after user clears the search on Add user to a channel modal

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24054

Conversation: https://community.mattermost.com/core/pl/fiznc65zq78mmmmbdz4cy9bt1o


lieut-data:
Makes sense to me, but swapping myself out for @hmhealey in case there is any extra context from the webapp team.

components/add_user_to_channel_modal/add_user_to_channel_modal.jsx
```diff
@@ -240,7 +240,6 @@ export default class AddUserToChannelModal extends React.Component {
                 className='form-control focused'
                 onChange={this.onInputChange}
                 value={this.state.text}
-                onKeyDown={this.handleKeyDown}
```
agarciamontoro, line 243: Why do we need to remove this?

lieut-data: I /think/ it's because the function is not defined at all, so we're just passing `undefined` through to the component anyway.

agarciamontoro: Oh, I just tried putting it back and saw no difference, so that explains it. I thought it was something related to this change. Thank you!

catalintomai: That's correct, @lieut-data , thanks = ). @agarciamontoro , in most cases I believe it to be a bad idea to make lateral, unrelated changes to the actual fix, but in this case it doesn't add or remove any functionality, it is more of a cleanup.

agarciamontoro: It's always a good time for cleanup :)

PR #5382, opened 2020-04-22
MM-24214: Add check for license for group mention API calls

catalintomai:
#### Summary
Avoids calling the getGroups* API methods if the user has no enterprise license

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24214

PR #5411, opened 2020-04-28
[MM-24521] Restrict group mentions permission to license with LDAP

fmunshi:
#### Summary
- Restricts the `USE_GROUP_MENTIONS` feature on the presence of the `LDAPGroups` license feature

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24521


components/admin_console/permission_schemes_settings/guest_permissions_tree/guest_permissions_tree.jsx
```diff
@@ -55,8 +56,11 @@ export default class GuestPermissionsTree extends React.Component {
                 ],
             },
             Permissions.USE_CHANNEL_MENTIONS,
-            Permissions.USE_GROUP_MENTIONS,
         ];
+
+        if (props.license.LDAPGroups === 'true') {
```
hahmadia, line 61:
```suggestion
        if (this.props.license && this.props.license.IsLicensed === 'true' && this.props.license.LDAPGroups === 'true') {
```
Maybe a little bit more precise if done this way? It's how I noticed it was done Catalin's PR and I thought it could be more accurate check (https://github.com/mattermost/mattermost-webapp/pull/5382/files#diff-1d27e4a784ea64f677a61481b5585a94R233)

components/admin_console/permission_schemes_settings/permissions_tree/permissions_tree.jsx
```diff
@@ -116,8 +116,9 @@ export default class PermissionsTree extends React.Component {
     }
 
     updateGroups = () => {
-        const {config, scope} = this.props;
+        const {config, scope, license} = this.props;
         const integrationsGroup = this.groups[this.groups.length - 1];
+        const postsGroup = this.groups[this.groups.length - 2];
```
hahmadia, line 121: I'm just confused where the order of the groups is defined? Where is it defined that this exact index is where the posts Group is? My worry is just for whatever reason if the order changes, wouldn't these become wrong?

components/admin_console/permission_schemes_settings/permissions_tree/permissions_tree.jsx
```diff
@@ -146,6 +147,9 @@ export default class PermissionsTree extends React.Component {
         if (scope === 'team_scope' && this.groups[0].id !== 'teams_team_scope') {
             this.groups[0].id = 'teams_team_scope';
         }
+        if (license.LDAPGroups === 'true' && postsGroup.permissions.indexOf(Permissions.USE_GROUP_MENTIONS) === -1) {
```
hahmadia, line 150: Update license check to match above suggested one? It seems that `postsGroup.permissions` is an array. Wouldn't it be better if we did `!postsGroup.permissions.inlcudes(Permissions.USE_GROUP_MENTIONS)` instead of using indexOf?

components/admin_console/permission_schemes_settings/permissions_tree/permissions_tree.jsx
fmunshi: yes they would - tests ensure that this order is accurate, it is the order in which the UI renders the permissions

hahmadia: Cool! Thanks for verifying.

components/admin_console/permission_schemes_settings/permissions_tree/permissions_tree.jsx
fmunshi: yea i used `indexOf` initially since the other checks in this function also used it. But I decided to change all of them to use `includes` since its much more readable

fmunshi:
bump @mkraft for dev review

PR #5414, opened 2020-04-29
[MM-24632] Show valid group mentions as blue

fmunshi:
#### Summary
- Shows valid group mentions as blue similar to user mentions but not clickable since popover card for groups is not in mvp spec
- *Note*: In order to actually see the blue you will have to inspect the element and remove the `mention--highlight ` class on the post message, since current behaviour shows all group mentions as highlighted. 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24632

#### Related Pull Requests
- https://github.com/mattermost/mattermost-redux/pull/1140

#### Screenshots
Screen Shot 2020-04-28 at 10 47 36 PM


components/at_mention/at_mention.jsx
```diff
@@ -71,8 +72,36 @@ export default class AtMention extends React.PureComponent {
         return '';
     }
 
+    getGroupFromMentionName(props) {
+        const groupsByName = props.groupsByName;
```
hahmadia, line 76: const {groupsByName} = this.props

components/at_mention/at_mention.jsx
```diff
@@ -71,8 +72,36 @@ export default class AtMention extends React.PureComponent {
...
+
+        return {};
+    }
+
     render() {
         const user = this.getUserFromMentionName(this.props);
+
+        if (!user) {
+            const group = this.getGroupFromMentionName(this.props);
+            if (group.allow_reference) {
```
catalintomai, line 100: I'm pretty sure you don't need this check, but prob. safer to keep it, if things change in the future.

fmunshi:
@srkgupta 
I believe I have fixed the behaviour of group mentions when users do not have permission as of the latest commit, it now looks like this: 

#### User with permission mentions a group:
- Group is highlighted yellow to members
- Group is highlighted blue to non members of the group

#### User without permission mentions a group:
- Group is shown as plain text to everyone

fmunshi:
Note: This will still apply when testing though

> In order to actually see the blue you will have to inspect the element and remove the mention--highlight class on the post message, since current behaviour shows all group mentions as highlighted.

PR #5417, opened 2020-04-29
MM-24590: fix roles check on leave

lieut-data:
#### Summary
As part of https://github.com/mattermost/mattermost-webapp/pull/5337, a check was added to avoid showing `createPost` until after channel roles are loaded. Unfortunately, when leaving a channel, the state is such that the channel exists but the channel membership does not, resulting in attempting to dereference undefined.

Fix the check not to assume channel memberships exist, and add an e2e test for leaving the channel via the header menu.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-24590

### DIFF
--- a/components/channel_view/index.js
+++ b/components/channel_view/index.js
@@ -40,12 +40,15 @@ function mapStateToProps(state) {
     let channelRolesLoading = true;
     if (channel && channel.id) {
         const roles = getRoles(state);
-        const channelRoles = getMyChannelRoles(state)[channel.id];
-        for (const roleName of channelRoles.values()) {
-            if (roles[roleName]) {
-                channelRolesLoading = false;
+        const myChannelRoles = getMyChannelRoles(state);
+        if (myChannelRoles[channel.id]) {
+            const channelRoles = myChannelRoles[channel.id].values();
+            for (const roleName of channelRoles) {
+                if (roles[roleName]) {
+                    channelRolesLoading = false;
+                }
+                break;
             }
-            break;
         }
     }
 
--- a/e2e/cypress/integration/messaging/channel_menu_spec.js
+++ b/e2e/cypress/integration/messaging/channel_menu_spec.js
@@ -79,4 +79,36 @@ describe('Channel header menu', () => {
             });
         });
     });
+
+    it('MM-24590 should leave channel successfully', () => {
+        let channel;
+
+        // # Go to "/"
+        cy.visit('/ad-1/channels/town-square');
+
+        cy.getCurrentTeamId().then((teamId) => {
+            // # Create new test channel
+            cy.apiCreateChannel(teamId, 'channel-test-leave', 'Channel Test Leave').then((res) => {
+                channel = res.body;
+
+                // # Select channel on the left hand side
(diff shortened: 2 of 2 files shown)


components/channel_view/index.js
```diff
@@ -40,12 +40,15 @@ function mapStateToProps(state) {
...
-            if (roles[roleName]) {
-                channelRolesLoading = false;
+        const myChannelRoles = getMyChannelRoles(state);
+        if (myChannelRoles[channel.id]) {
+            const channelRoles = myChannelRoles[channel.id].values();
+            for (const roleName of channelRoles) {
+                if (roles[roleName]) {
+                    channelRolesLoading = false;
+                }
+                break;
```
lieut-data, line 50: @fmunshi, I left this untouched, but I don't fully understand this. Are we intentionally just iterating the loop once and testing the value that happens to get returned?

fmunshi: This loop simply checks for the existence of a role matching the role name specified by the channel membership. If that role does not exist it means we are still loading the roles.

This was actually a bit of a bandaid fix for a bigger problem in the way our permission checks are done - they assume that the data is loaded into the state and so if the role is not yet loaded into the state it treats it as if the user does not have the permission.

I am planning on refactoring `haveIChannelPermission` and `haveITeamPermission` to return a different value if the permission is not found due to the roles not being found in the state vs permission not found due to the corresponding role not having the permission.

Since I wasnt sure if I would get around to doing before 5.23 I added this check here to make sure the post box doesnt show up when the roles are not loaded into state.

lieut-data: @fmunshi, I'm still stuck on the `break` statement. What if the role you're looking to check is the second one in the list? We'll never find it, since we unconditionally `break` after the first iteration.

fmunshi: `myChannelRoles` will only ever be a set of length > 1 if the user is a channel admin where it would be ['channel_user', 'channel_admin'] in all other cases it just looks like ["someuuid"] or ["channel_user"] or ["channel_guest"].

In the case of channel admin since all channel roles are loaded at the same time anyway if the `channel_user` role is loaded then we can assume the `channel_admin` role is also loaded.

Technically the break isnt really even needed you could just drop the if and the break altogether and have 
`channelRolesLoading = roles.hasOwnProperty(roleName);`

lieut-data: Thanks for clarifying -- given it's working as designed, I'm going to leave it to avoid my breaking something. Just wanted to double check intent :)

srkgupta:
@lieut-data I am not sure if this issue is related to this fix, but when I am trying to close a Group Message, it does not works as shown in this GIF. Can you please confirm if this is a different issue? 

Kapture 2020-04-30 at 18 18 35

lieut-data:
@srkgupta, I seem to be able to reproduce that in `master` as well. I've bisected the cause of that to https://mattermost.atlassian.net/browse/MM-23065. Would you be open to filing a separate ticket to track?

srkgupta:
Thanks for the confirmation @lieut-data. I found a similar issue already being reported on 5.23. 
https://mattermost.atlassian.net/browse/MM-24506

PR #3727, opened 2019-09-22
MM-18481 - Common component: Separator

reflog:
#### Summary
Refactored common separator styles into a component

#### Ticket Link
Fixed: https://mattermost.atlassian.net/browse/MM-18481

PR #5419, opened 2020-04-29
add new-separator class to NewMessageSeparator component

jfrerich:
#### Summary
This PR adds add the `new-separator` class to the `NewMessageSeparator` component. 

By wrapping the `NotificationSeparator` component with this class, the css will adjust according to the settings in `Account Settings` > `Display` > `Theme` > `Custom Theme` >  `Center Channel Styles` > `New Message Separator`

Tested the following cases:
* change in color exists after marking message unread (with post menu pulldown option)
* line and message can be seen changing colors (in the background of the account settings modal) as you change the color picker value
* As 2nd user, post message to channel user1 is not viewing, to create new unread messages. User1 switches to channel and sees the new line component with the correct color.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23885

#### Related Pull Requests
Original PR that introduced this issue. https://github.com/mattermost/mattermost-webapp/pull/3727

#### Screenshots
With the color set to blue (as an example for the color picker), the following was still loading the default color
image

After the fix, the color is adjusted to match the user's Account Settings
image

### DIFF
--- a/components/post_view/new_message_separator/__snapshots__/new_message_separator.test.jsx.snap
+++ b/components/post_view/new_message_separator/__snapshots__/new_message_separator.test.jsx.snap
@@ -1,13 +1,17 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`components/post_view/new_message_separator should render new_message_separator 1`] = `
-<NotificationSeparator
-  id="1234"
+<div
+  className="new-separator"
 >
-  <FormattedMessage
-    defaultMessage="New Messages"
-    id="posts_view.newMsg"
-    values={Object {}}
-  />
-</NotificationSeparator>
+  <NotificationSeparator
+    id="1234"
+  >
+    <FormattedMessage
+      defaultMessage="New Messages"
+      id="posts_view.newMsg"
+      values={Object {}}
+    />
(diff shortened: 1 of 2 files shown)


jfrerich:
@lieut-data, I am interested in how you bisected the commit that produced the issue...  :)

jfrerich:
Thanks for the feedback, @lieut-data.  I find that testing previous commits also requires me to find a compatible version of `enterprise`, `server`, `redux` repo commits.  How did you run `git bisect` on the webapp and keep the other repos in a state that compiles?  Did you do that manually for each git bisect on the webapp repo?

lieut-data:
@jfrerich, actually I find that the `server` is remarkably backwards compatible. That is, for at least this bisect, I was just running the `master` version of server and enterprise the whole time. For `redux`, since I'm just doing `npm install` each time, I get the version that's referenced by that commit's `package.json` automatically.

jfrerich:
@lieut-data thanks!  That is fantastic feedback.  I introduced `git bisect` during our off-site meeting, but didn't consider it for this ticket (because of the multiple repos).  I might have to blog this one because it seemed that it wasn't a known feature of `git` for some.

PR #5420, opened 2020-04-29
MM-24467: Add configuration setting for ServiceProviderIdentifier

sbishel:
#### Summary
The tag in the AuthnRequest is being created incorrectly. This PR uses a new configuration setting to set the . If the new configuration setting is empty, then the current logic is used. This will allow existing installations to continue to work.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24467

#### Related Pull Requests
https://github.com/mattermost/enterprise/pull/633
https://github.com/mattermost/mattermost-server/pull/14426

#### Screenshots
Screen Shot 2020-04-29 at 4 49 32 PM


e2e/cypress.env.json
```diff
@@ -0,0 +1,3 @@
+{
+    "setChromeWebSecurity": false
+}
```
fmunshi, line 3: How were the saml e2e tests working without this setting before?

sbishel: This setting is to be set before running these tests. This file should not be checked in, thanks for catching this.

sbishel: Removed...

sbishel:
Not going to do this implementation at this time.

PR #4791, opened 2020-01-30
[MM-21000] Fix using local storage to return to a DM/GM

Willyfrog:
#### Summary
Fix whenever using local storage to return to a previous DM/GM as we were only looking for channels.

#### Ticket Link

[MM-21000](https://mattermost.atlassian.net/browse/MM-21000)

PR #5422, opened 2020-04-30
MM-24506: Fix closing DM/GM from header dropdown menu

agarciamontoro:
#### Summary

PR #4791 fixed an issue with redirecting the user to direct or group messages when they were the last ones to be visited. But a collateral regression was introduced: closing DMs or GMs stopped working, as the user was returned to the closed channel immediately after leaving it.

That commit was reverted (#4995), and then applied again in PR #5224, which also introduced the new `leaveDirectChannel` function to correctly handle the new issue with closing DMs or GMs. But this function was only applied to the case where the user closed the channel through the `x` button on the sidebar channel item.

However, the bug remained in place when the user closed the channel using the "Close {Direct, Group} Message" option in the channel header dropdown menu.

This PR just fixes this second use case by calling the `leaveDirectChannel` function from the right place.

I have also added a couple of E2E tests to make sure that we don't lose this functionality again in the future.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24506

### DIFF
--- a/e2e/cypress/integration/channel/close_direct_group_spec.js
+++ b/e2e/cypress/integration/channel/close_direct_group_spec.js
@@ -7,7 +7,6 @@
 // - Use element ID when selecting an element. Create one if none.
 // ***************************************************************
 
-// Stage: @prod @smoke
 // Group: @channel @channel_settings
 
 // Make sure that the current channel is Town Square and that the
--- a/e2e/cypress/integration/channel/close_direct_group_spec.js
+++ b/e2e/cypress/integration/channel/close_direct_group_spec.js
@@ -1,0 +1,74 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Stage: @prod @smoke
+// Group: @channel @channel_settings
+
+// Make sure that the current channel is Town Square and that the
+// channel identified by the passed name is no longer in the channel
+// sidebar
+function verifyChannelWasProperlyClosed(channelName) {
+    // * Make sure that we have switched channels
+    cy.get('#channelHeaderTitle').should('contain', 'Town Square');
+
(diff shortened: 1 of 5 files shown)


e2e/cypress/integration/channel/close_direct_group_spec.js
```diff
@@ -0,0 +1,74 @@
...
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Stage: @prod @smoke
+// Group: @channel @channel_settings
```
agarciamontoro, line 11: @saturninoabril I just copied these lines from another file, but I'm not sure what they mean. Is there any place with docs on what to put here?

saturninoabril: Thanks for checking. These are what we called test metadata ([link](https://developers.mattermost.com/contribute/webapp/end-to-end-tests/#adding-test-metadata-on-spec-files)). You may remain the "Group" as is. For "Stage", please remove. Once merged, QA team will monitor this spec and will promote to `@prod` if we found the test to be stable.

agarciamontoro: Thank you! Done in https://github.com/mattermost/mattermost-webapp/pull/5422/commits/bac1b771a1f226ad2bc58a6c09068687cebd1b3c

lieut-data:
Nicely done, @agarciamontoro!

amyblais:
@saturninoabril Kind reminder to review for v5.23.

e2e/cypress/integration/channel/close_direct_group_spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+    it('Through x button on channel sidebar item', () => {
+        cy.createAndVisitNewGroupMessageWith(['sysadmin', 'samuel.tucker']).then((channel) => {
+            // # Click on the x button on the sidebar channel item
+            cy.get('#sidebarItem_' + channel.name + '>span.btn-close').click({force: true});
+
+            verifyChannelWasProperlyClosed(channel.name);
+        });
+    });
+});
```
saturninoabril, line 73: 💯

PR #5433, opened 2020-05-04
Adding option to invite users from invite errors list

jespino:
#### Summary
When you try to invite a regular user as a guest, this shows an error in the
invitation results, this error is fixable, so you can fix it right away there,
adding that user to the team a channels as a regular user.

#### Ticket Link
[MM-17502](https://mattermost.atlassian.net/browse/MM-17502)

#### Screenshots
image


components/invitation_modal/invitation_modal_confirm_step_row.jsx
```diff
@@ -61,6 +80,68 @@ export default class InvitationModalConfirmStepRow extends React.PureComponent {
...
+                        >
+                            <FormattedMessage
+                                id='invite.members.is-already-user-in-team.add-to-channels'
+                                defaultMessage='Add this member to the team and channels.'
+                            />
+                        </a>}
+                </div>
+            );
+        }
+
```
reflog, line 144: seems like alot of code repetition here. wouldn't it be cleaner if the id/defaultMessage were decided by 'if' and the rendering part would be the same for both cases?

stylianosrigas:
Removing the test server for maintenance purposes. If you need again please add the label. Sorry for any inconvenience.

PR #5436, opened 2020-05-04
MM-24781 Make GM profiles API calls to be concurrent

sudheerDev:
#### Summary
  * Add p-queue as a dependency and keep concurrency at 4 calls. 

On the web, we now make one call at a time for GM profiles as attached in the screenshot. This can be improved by making 4 calls at a time in parallel and adding a new request to the stack when one finishes. 
Making multiple calls will fetch all profiles sooner and instead of blocking the network stack with all the requests at once, we can cap them at 4. The reason to cap them at 4 is to have a little room for other network requests to go through in case they are needed while these are pending.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24781

#### Screenshots
Before
May-04-2020 19-03-10

After
May-05-2020 01-45-25


actions/user_actions.test.js
```diff
@@ -232,4 +241,62 @@ describe('Actions.User', () => {
...
+                },
+            },
+        };
+
+        const testStore = mockStore(state);
+        getState.mockImplementation(testStore.getState);
+
+        await UserActions.loadProfilesForGM();
+        expect(UserActions.queue.onEmpty).toHaveBeenCalled();
+        expect(UserActions.queue.add).toHaveBeenCalled();
```
sudheerDev, line 300: i wanted to assert that we are calling an action here but it was becoming hard to mock the class etc.

package.json
```diff
@@ -37,6 +37,7 @@
     "marked": "github:mattermost/marked#5a06d1af25ca48927b9efe0d9c42c01c84e7839b",
     "mattermost-redux": "github:mattermost/mattermost-redux#9a915443ab1d695540493406926c3dd269d380fc",
     "moment-timezone": "0.5.27",
+    "p-queue": "^6.4.0",
```
sudheerDev, line 40: The footprint of the library small and it has good API which I intend to use in upcoming features like prefetching posts. It has API's for clearing, pausing and adding tasks dynamically too which fits perfectly for some of the upcoming features.

actions/user_actions.jsx
```diff
@@ -256,11 +259,13 @@ export async function loadProfilesForGM() {
             });
         }
 
-        await dispatch(UserActions.getProfilesInChannel(channel.id, 0, Constants.MAX_USERS_IN_GM)); //eslint-disable-line no-await-in-loop
+        const getProfilesAction = UserActions.getProfilesInChannel(channel.id, 0, Constants.MAX_USERS_IN_GM); //eslint-disable-line no-await-in-loop
```
Willyfrog, line 262: i think you can remove the comment
```suggestion
        const getProfilesAction = UserActions.getProfilesInChannel(channel.id, 0, Constants.MAX_USERS_IN_GM);
```

PR #5439, opened 2020-05-05
Do not allow to navigate to deleted teams

jespino:
#### Summary
Do not allow to navigate to deleted teams

#### Ticket Link
[MM-24021](https://mattermost.atlassian.net/browse/MM-24021)


jespino:
@streamer45 actually, no, it is just fixing the symptom from the frontend perspective, I've created another ticket to handle properly the soft deleted teams. Now the API is quite inconsistent about this, so I think we should tackle this as a 6.0 change, and try to organize the API to make it sensible, for example, everything accepting the `includeDeleted` parameter, or have an special API for list archived and restore archived teams, and everything else filtering the soft deleted teams.

jespino:
@prapti yes, I think it can be addressed separately

PR #5440, opened 2020-05-05
E2E: fix tests for sidebar link in system console

saturninoabril:
#### Summary
Fix tests for sidebar link in system console.  Instead of navigating on each section in system console, simply click on the sidebar and verify the corresponding URL is correct.

#### Ticket Link
none, failing on daily Cypress tests.


prapti:
Most enterprise sidebar link navigation tests are failing for me right now. Queueing this one up for more detailed review.

saturninoabril:
We test on E20 currently.  So I think it's good as is.
But let's create a ticket to address it separately.

PR #5441, opened 2020-05-05
fix missing dash between tos/pp

lieut-data:
#### Summary
While the TOS/PP links are hard-coded on the about modal, the code still checked for an (empty) configured value (so as to hide on other screens) when rendering the dash between them. Fix this to make the dash equally hard-coded.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-24856

#### Screenshots
Before:
image

After:
image


lieut-data:
@lindalumitchell, not sure to whom to send this for QA Review -- please feel free to re-assign!

PR #5446, opened 2020-05-06
Show emoji aliases

Roy-Orbison:
#### Summary
Shows all aliases in the picker, assisting typing of emojis. Drops unnecessary hard-coded previewer width that truncates aliases.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/6276  
https://mattermost.atlassian.net/browse/MM-5883

#### Screenshots
image


components/emoji_picker/components/emoji_picker_preview.jsx
```diff
@@ -61,7 +61,7 @@ export default class EmojiPickerPreview extends React.Component {
                             id='emojiPickerAliasesPreview'
                             className='emoji-picker__preview-aliases'
                         >
-                            {':' + aliases[0] + ':'}
+                            {':' + aliases.join(': :') + ':'}
```
hmhealey, line 64: Feel free to leave this as-is, but it might be clearer what we're doing here if you use `map` to add the colons and then `join` to add the spaces between items.

Roy-Orbison:
Is "shit" really so profane that it requires censoring? This seems a bit prim. It's not like it's the *name* of the emoji.

To censor it will either be a bloat at the display-level, using a blacklist or something, or require a change the emoji data structure/generation to have a `visible_aliases` list.

components/emoji_picker/components/emoji_picker_preview.jsx
Roy-Orbison: `join()` is far more efficient. If it were a more complex transformation of each item, I'd agree with you.

Roy-Orbison:
I don't know how the `utils/emoji.json` file is generated, so I'm not going to edit it by hand. I can see how some servers would wish to censor emojis, but wouldn't it be better to have a System Console feature to block emojis or specific aliases entirely? E.g. 👌 means okay in many places but [is rude elsewhere](https://en.wikipedia.org/wiki/Obscene_gesture#Okay_gesture).

hmhealey:
We have thought about adding a feature to remove other default emojis in the past since there's other ones like the Doomguy emojis that we've inherited from GitHub (eg. :rage1:). If we had more aliases we wanted to filter out, something like a `visible_aliases` would be helpful, but I think we could just make this a one-off like `aliases.filter((alias) => alias != 'shit').join(': :')` since this is the only place it would appear.

hmhealey:
Actually, never mind. I just talked to @esethna, and we agreed that it was fine to leave in. In that case, this looks good to me.

esethna:
@Roy-Orbison, quick question, what happens if there's more aliases than can fit on the one line? 

Can we truncate them with "..."?

Roy-Orbison:
Good because filtering just one alias like that is pretty hacky.

You can add another rule for truncation if you like, but the example shown is the longest set at present, that's why I chose it.

Roy-Orbison:
Hmmm. It's actually `:raised_hand_with_fingers_splayed_medium_light_skin_tone:` but you'd have to redesign the picker preview to accommodate that.

hmhealey:
Thankfully, we don't display the different skin tones separately in the emoji picker, so you won't see that in the UI.

Roy-Orbison:
Has this stalled?

hmhealey:
I'll message @jgilliam17 about it. It might've been missed, but I know we're starting into testing for the next release, so they may just be a bit busy.

jgilliam17:
Thank you @Roy-Orbison and apologize about delay in getting to this PR.
Looks great!
One small issue.
I think @esethna already mentioned this previously, when there are more aliases than can fit in the provided space, can they be truncated with ... or similar?
They are partially visible on the bottom right now
Thank you again 
Screen Shot 2020-06-02 at 4 39 58 PM

Roy-Orbison:
image

hmhealey:
Yep, I'll merge this now. Thanks again for the PR, @Roy-Orbison! This'll be going in for our 5.28 release in August, although you'll be able to see it as soon as tomorrow on the nightly version of our Community server at https://community-daily.mattermost.com.

PR #5449, opened 2020-05-06
[MM-24154] - Prevent copying of certain elements in posts

nevyangelova:
#### Summary
As part of the quick wins UI spring we decided to exclude the copying of certain elements from posts: 
- User avatar image
- Reply count
- "Commented on" message
- "Edited" message 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24154

Fix v.Queued


bradjcoughlin:
What exactly is are the scenario here? For example, would I be attempting to highlight the following? Here's what I just did on the test server:
Screen Shot 2020-05-06 at 10 53 26 AM
Screen Shot 2020-05-06 at 10 53 10 AM

esethna:
It's working as expected for me based on a brief test

image

bradjcoughlin:
I'm still seeing the user profile image being selected:
Screen Shot 2020-05-06 at 7 51 50 PM

On Safari on OSX, It seems the `user-select: none` isn't applying. I'm guessing we need vendor prefixes such as:
```
-webkit-user-select: none;  /* Chrome all / Safari all */
  -moz-user-select: none;     /* Firefox all */
  -ms-user-select: none;      /* IE 10+ */
  user-select: none;          /* Likely future */      
```
I mentioned awhile back we could use `autoprefixer` but currently [we're using `compass` mixins](http://compass-style.org/reference/compass/css3/user_interface/#mixin-user-select) to accomplish this. (@deanwhillier). I haven't tried the above idea locally yet.

nevyangelova:
@bradjcoughlin thanks for the find I will implement the vendor prefixes. @andrewbrown00 its totally doable I think, with events, I will go ahead and try now :)

nevyangelova:
PR updated, note that the selector isn't working for Safari on image alt. It is the browser behaviour related to accessibility so it will still copy the text.

andrewbrown00:
Thanks @nevyangelova we are super close on this.

I do see a slight regression and it. seems to be copying and allowing me paste the `reply count`. Which is odd sine I don't see it when selecting the messages.

image

**Output** 
```
user-1 [1:51 PM]

1
This is a post I'd like to copy

[user-1 [1:58 PM]]
Here's a post I'm going to attempt to copy using highlight

[sysadmin [6:44 PM]]



1
Commented on user-1's message: This is a post I'd like to copy
This a reply that I edited
```

- [ ] Can we remove the **extra space** between the `username/timestamp` and the message and just have the space between messages? So it renders like:

```
user-1 [1:51 PM]
This is a post I'd like to copy

[user-1 [1:58 PM]]
Here's a post I'm going to attempt to copy using highlight

[sysadmin [6:44 PM]]
Commented on user-1's message: This is a post I'd like to copy
This a reply that I edited
```

nevyangelova:
@andrewbrown00 updated. Bear in mind, we can only influence selection behaviour so much. I am doing this with Regex, which is not an 'all-case' solution and it can get quite hacky. In the current example, by influencing the selection we ignore the 'user-select' CSS property so I have to manually exclude those elements with javascript. Hope this solution gives us the best of both worlds.

components/root/root.jsx
```diff
@@ -137,6 +137,31 @@ export default class Root extends React.Component {
...
         });
 
+        document.addEventListener('copy', (event) => {
+            const selectionArray = document.getSelection().toString().split('\n');
+
+            // Since the getSelection() doesn't take into account the CSS property 'user-select' we have to manually create a class and exclude those elements
+            const unselectableElements = document.getSelection().getRangeAt(0).commonAncestorContainer.querySelectorAll('.unselectable');
+            const unselectableElementsArray = Array.from(unselectableElements).map((element) => element.textContent);
+
+            const formattedSelection = selectionArray.filter((element) => !unselectableElementsArray.includes(element)).reduce((result, line, index) => {
```
bradjcoughlin, line 147: Non-blocking suggestion: Maybe add a comment here as to what this is filtering? I know it's looking for a post but in the future we might want to add additional use cases.

bradjcoughlin: Another thought: making copy/paste formatting pluggable would be really nice. That way, you could change the output of various post types upon copying. I may file a ticket for that 💡

hmhealey:
@andrewbrown00 Formatting the post in that way is out of scope for this ticket. We were going to add better formatting with https://mattermost.atlassian.net/browse/MM-15727, but that was something that needed more discussion.

This ticket was just to prevent selection of things like menus since that's straightforward enough to do without much discussion of how it should work or be implemented.

components/common/comment_icon.tsx
```diff
@@ -33,7 +33,7 @@ export default class CommentIcon extends React.PureComponent<Props> {
         if (this.props.commentCount > 0) {
             iconStyle += ' post-menu__item--show';
             commentCountSpan = (
-                <span className='post-menu__comment-count'>
+                <span className='post-menu__comment-count unselectable'>
```
hmhealey, line 36: Instead of marking things with a helper class, is it maybe better form to add `user-select: none` to the classes that are already on these components? 0/5, but maybe people with more experience with CSS would have a stronger opinion on that.

components/widgets/users/avatar.scss
```diff
@@ -31,3 +32,10 @@
         height: 128px;
     }
 }
+
+.unselectable {
+    -webkit-user-select: none; /* Chrome all / Safari all */
```
hmhealey, line 37: I think `user-select` should be enough here to cover all browsers that we support.

bradjcoughlin: I was seeing issues in the highlighting, so I suggested adding the vendor prefixes not knowing if that was the issue. I'm still seeing highlighting styles on Safari (I realize I already approved, but not sure if the highlighting is in scope for this issue).

First recording is Chrome, second is Safari:
highlight-post

components/common/comment_icon.tsx
nevyangelova: Yes the original solution was to use “user-select”, like it’s in the first commit. This helper class was added to help exclude those elements from the selection because document.getSelection() doesn’t take the css selector into account and was copying them regardless. I will revert to that solution since we won’t be modifying the selection anymore

nevyangelova:
Thanks everyone! I will revert these copy-paste behaviour changes and discuss the solution further with the team and this https://mattermost.atlassian.net/browse/MM-15727 @andrewbrown00 @bradjcoughlin @hmhealey @esethna

components/widgets/users/avatar.scss
hmhealey: Ugh, of course Safari has to be different...

PR #5452, opened 2020-05-06
MM-24033: Compute suggestion list width

agarciamontoro:
## Summary
The `SuggestionList` inside the `AddUserToChannelModel` component had a fixed width that rendered smaller than the modal input box width. This PR computes the width of the `SuggestionList` as the parent div width minus two times the padding in the parent div.

This is *very* hacky, and that is why I have narrowed the change to a specific class inside the component. But we have plans to rework this modal in the near future, so we should get rid of this code hopefully sooner than later.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24033

#### Screenshots
image
image

### DIFF
--- a/components/add_user_to_channel_modal/__snapshots__/add_user_to_channel_modal.test.jsx.snap
+++ b/components/add_user_to_channel_modal/__snapshots__/add_user_to_channel_modal.test.jsx.snap
@@ -71,7 +71,7 @@ exports[`components/AddUserToChannelModal should match snapshot 1`] = `
       <SuggestionBox
         className="form-control focused"
         completeOnTab={false}
-        containerClass=""
+        containerClass="add-user-to-channel-modal__suggestion-box"
         delayInputUpdate={true}
         isRHS={false}
         listComponent={[Function]}
--- a/components/add_user_to_channel_modal/__snapshots__/add_user_to_channel_modal.test.jsx.snap
+++ b/components/add_user_to_channel_modal/__snapshots__/add_user_to_channel_modal.test.jsx.snap
@@ -68,37 +68,41 @@ exports[`components/AddUserToChannelModal should match snapshot 1`] = `
           values={Object {}}
         />
       </div>
-      <SuggestionBox
-        className="form-control focused"
-        completeOnTab={false}
-        containerClass="add-user-to-channel-modal__suggestion-box"
-        delayInputUpdate={true}
-        isRHS={false}
-        listComponent={[Function]}
-        listStyle="bottom"
-        listenForMentionKeyClick={false}
-        maxLength="64"
-        onChange={[Function]}
-        onItemSelected={[Function]}
-        openOnFocus={false}
-        openWhenEmpty={true}
-        providers={
-          Array [
-            SearchChannelWithPermissionsProvider {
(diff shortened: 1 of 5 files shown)


asaadmahmood:
@agarciamontoro We face this issue in other modals as well, so instead of adding the class, we can just target all suggestion lists inside modals, and give them the css `width: calc(100% - 30px);`.
Screenshot 2020-05-07 at 2 08 10 PM

agarciamontoro:
@asaadmahmood after double and triple checking, looks like just removing the `width: 496px` does fix the issue in the `EditPost` modal. But if we add the `width: calc(100% - 30px)` to all the suggestion lists in addition, it actually brings back the issue with the `EditPost` modal, as it is now 30px narrower than before.

What I still don't understand is why removing the fixed width works with the `EditPost` modal and not with the `AddUserToChannel` modal. I have checked the DOM in both cases and it looks exactly the same, but of course I must be missing something.

I would like to understand the root cause of this, but it looks like it will become a time sink really soon. This is what I have tested:

| Change | AddUserToChannel fixed? | EditPost fixed? |
|-|-|-|
| Current PR (removing `width: 496px` from `.suggestion-list__content` and adding `width: calc(100% - 30px)` to `.add-user-to-channel__suggestion-box .suggestion_list` | Yes | Yes |
| Removing `width: 496px` from `.suggestion-list__content` | No | Yes |
| Replacing `width: 496px` in `.suggestion-list__content` with `width: calc(100% - 30px)` | Yes | No |
| Replacing `width: 496px` in `.suggestion-list` with `width: calc(100% - 30px)` | Yes | No |

Given the table above and the assumption that we will rework these modals in the near future, hopefully getting rid of all these moving parts, I propose to merge this as is now to fix the current ticker. Thoughts, @asaadmahmood?

asaadmahmood:
@agarciamontoro `width: 496px` is to have a reduced width in the center channel. And that's intentional, but I'll have a look at this PR, and submit a fix.
So keep this PR open, will submit a commit.

Screenshot 2020-05-07 at 4 08 58 PM

agarciamontoro:
Thank you, @asaadmahmood!

asaadmahmood:
@agarciamontoro Submitted my fix.

agarciamontoro:
Your solution is definitely way better than mine. Thank you, @asaadmahmood!

jgilliam17:
Thank you @agarciamontoro 
I'm seeing the channels with long name not truncated within the list. Is this out of scope of this PR?

agarciamontoro:
Thank you for testing this, @jgilliam17. I did not check that the names were truncated within the list, this PR was only to fix the width issue. Should we open a new ticket for that one?

jgilliam17:
> Thank you for testing this, @jgilliam17. I did not check that the names were truncated within the list, this PR was only to fix the width issue. Should we open a new ticket for that one?

I can add a new ticket for that 🙂

agarciamontoro:
@amyblais, should we add this to v5.23?

amyblais:
Added to v5.24; not a regression

agarciamontoro:
Thank you, @amyblais. Merging now!

PR #5453, opened 2020-05-06
[MM-23694] - Improve channel switcher feedback when there are no results

nevyangelova:
#### Summary
As part of the Improved autocomplete UI epic this tickets adds a better experience for the empty state of Channel Switcher autocomplete. Design is in the ticket. 

This component will be reused for all empty states listed here https://www.figma.com/file/MsznbIxwHeq91iNBrQ4Kv5/Components---Right-Sidebar?node-id=77%3A3056 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23694

Fix v.Queued


esethna:
@matthewbirtch @nevyangelova  0/5 first impressions:

1. Should we not show the loading indicator for subsequent key strokes (other than a backspace) after this image displays? Since we know further key strokes won't have any matches:

https://i.gyazo.com/9f82d58496d20fa6732ed2d6af47958e.gif

2. Image seems blurry on my 4k monitor

image

devinbinnie:
@nevyangelova Re-assigning to @ali-farooq0 so I can focus on the remaining sidebar work.

components/quick_switch_modal/quick_switch_modal.jsx
```diff
@@ -305,7 +312,15 @@ export default class QuickSwitchModal extends React.PureComponent {
                             renderDividers={renderDividers}
                             delayInputUpdate={true}
                             openWhenEmpty={true}
+                            onSuggestionsReceived={this.handleSuggestionsReceived}
                         />
+                        {!this.state.hasSuggestions &&
+                        <NoResultsIndicator
+                            variant='Search'
```
ali-farooq0, line 319: can we use the `Search` enum here? unless jsx won't allow it 😭.

If that's the case mind putting a comment here so whoever converts this to TS can use the enum.
`// NoResultsVariant.Search`

components/no_results_indicator/no_results_indicator.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    // this will be changed soon in the PRs following the other no results indicators
+    [NoResultsVariant.Mentions]: '',
+    [NoResultsVariant.FlaggedPosts]: '',
+    [NoResultsVariant.PinnedPosts]: '',
+};
+
+const titleMap: {[key in NoResultsVariant]: MessageDescriptor } = {
+    [NoResultsVariant.Search]: {
+        id: t('no_results.channels.title')
```
ali-farooq0, line 29: Feels odd to have a channel specific format string inside a generic component, no? I imagine this component being used in search scenarios as well, like text search in RHS, no?

If that's the case I wonder if we can pass this through the props? would help with the customizability, thoughts? 🤔

nevyangelova: I can change the string to make it specific to the quick switch modal, I thought I would use the same text for the Channel Search in the RHS, but later saw that the text was different :D So I can add a new variant for the RHS. 

I created the Enum in order to make it customisable and simple to use enough by passing a Variant to props only. Otherwise it looked quite weird to pass all the different formatted message values through props (ex. titleId, subtitleId, defaultMessageTitle, defaultMessagesubtitle etc.) and this is handled quite well by the messageDescriptor.

nevyangelova:
> Nice @nevyangelova! Just a concern regarding how generic we want to keep the `NoResultsIndicator` component. If you feel like it's tightly coupled with with the quick_switch component, we could move it under `components/quick_switch_modal` and then it might ok to have some channel/context in there.

As far as I know it will be used in quite a few places, not related to the quick switch modal and they will have different icons and text, as well as parent margins :) I think its for all autocomplete elements eventually

components/no_results_indicator/no_results_indicator.tsx
ali-farooq0: I see. Fair enough. Only issue is as soon as we have more customized types this list of variants will grow :P, which isn't a bad thing.

In which case then do you mind renaming the `Prop.value` to something that's close to FromattedMessage, otherwise who ever wants to use this component will first have to look at how `value` prop is being used and then use it. Could name it something like `formattedMessageValues`, what do you think?

jgilliam17:
Thank you @nevyangelova  
Looks good, I'd like to add only one small edge case
When displaying a not found search term using all of the available characters (64 character limit), there are no margins on the modal to stop the text from running out of view. Even with the scroll bar added, text is out of the view on the left side.

Screen Shot 2020-05-07 at 5 44 01 PM

esethna:
Looks good, thanks @nevyangelova. I'll leave it to @matthewbirtch to give UX approval

i18n/en.json
```diff
@@ -2963,6 +2963,8 @@
   "navbar.toggle2": "Toggle sidebar",
   "navbar.viewInfo": "View Info",
   "navbar.viewPinnedPosts": "View Pinned Posts",
+  "no_results.channels.subtitle": "Check the spelling or try another search.",
+  "no_results.channels.title": "No results for {channelName}",
```
ali-farooq0, line 2967: Recommend using a generic name for the both keys here:
```
  "no_results.search.subtitle": "Check the spelling or try another search.",
  "no_results.search.title": "No results for {searchText}",
```
Or maybe change the variant name to `ChannelSearch`

components/suggestion/suggestion_list.jsx
```diff
@@ -224,18 +225,17 @@ export default class SuggestionList extends React.PureComponent {
...
-                <div
-                    id='suggestionList'
-                    ref={this.contentRef}
-                    style={{...contentStyle}}
-                    className={contentClass}
-                    onMouseDown={this.props.preventClose}
-                >
-                    {items}
-                </div>
+        return this.props.supressLoadingSpinner ? '' : (<div className={mainClass}>
```
ali-farooq0, line 228: How does it look without suppressing the spinner?

components/suggestion/suggestion_box.jsx
```diff
@@ -132,6 +132,16 @@ export default class SuggestionBox extends React.Component {
...
+
+        /**
+         * Allows parent to access received suggestions
+         */
+        onSuggestionsReceived: PropTypes.func,
+
+        /**
+         * Supress loading spinner when necessary
+         */
+        supressLoadingSpinner: PropTypes.bool,
```
ali-farooq0, line 144: typo: `supPressLoadingSpinner`

components/suggestion/suggestion_list.jsx
```diff
@@ -30,6 +30,7 @@ export default class SuggestionList extends React.PureComponent {
         selection: PropTypes.string.isRequired,
         components: PropTypes.array.isRequired,
         wrapperHeight: PropTypes.number,
+        supressLoadingSpinner: PropTypes.bool,
```
ali-farooq0, line 33: typo: `supPressLoadingSpinner`

components/suggestion/suggestion_box.jsx
```diff
@@ -132,6 +132,16 @@ export default class SuggestionBox extends React.Component {
...
          */
         wrapperHeight: PropTypes.number,
+
+        /**
+         * Allows parent to access received suggestions
+         */
+        onSuggestionsReceived: PropTypes.func,
+
+        /**
+         * Supress loading spinner when necessary
```
ali-farooq0, line 142: `Suppress`

components/quick_switch_modal/quick_switch_modal.jsx
```diff
@@ -305,7 +316,16 @@ export default class QuickSwitchModal extends React.PureComponent {
...
+                            onSuggestionsReceived={this.handleSuggestionsReceived}
+                            supressLoadingSpinner={!this.state.hasSuggestions}
+                        />
+                        {!this.state.hasSuggestions &&
+                        <NoResultsIndicator
+                            variant={NoResultsVariant.Search}
+                            formattedMessageValues={{channelName: `"${this.state.pretext}"`}}
                         />
+                        }
+
```
ali-farooq0, line 328: nit: empty line

components/quick_switch_modal/quick_switch_modal.jsx
```diff
@@ -305,7 +316,16 @@ export default class QuickSwitchModal extends React.PureComponent {
                             renderDividers={renderDividers}
                             delayInputUpdate={true}
                             openWhenEmpty={true}
+                            onSuggestionsReceived={this.handleSuggestionsReceived}
+                            supressLoadingSpinner={!this.state.hasSuggestions}
+                        />
+                        {!this.state.hasSuggestions &&
+                        <NoResultsIndicator
+                            variant={NoResultsVariant.Search}
```
ali-farooq0, line 324: 👍

components/suggestion/suggestion_box.jsx
nevyangelova: 🙈

components/suggestion/suggestion_list.jsx
nevyangelova: It renders an empty line and/or a spinner even without any results because its mounting this component.

components/suggestion/suggestion_box.jsx
ali-farooq0: heh, i've been there before, mine is even committed 🤫

nevyangelova:
@ali-farooq0 implemented the suggestions. thx!

components/suggestion/suggestion_list.jsx
ali-farooq0: Gotcha, it doesn't work even if the spinner is hidden? wondering if we can just not include the spinner in the `items` (earlier in the function) rather than hiding this whole component. If it doesn't work then all good, just a thought.

nevyangelova: I tried that at first and it didn't hide the line 🙃

ali-farooq0: oh okay, not to thrash on the change but then shouldn't this be more like `hideSuggestionList` or something similar? and I imagine returning `null` here doesn't work either?

PR #5454, opened 2020-05-07
disable button until link shows

jfrerich:
#### Summary
This PR addresses a situation that occurs when users have a slow connection. When this happens, the modal that is used to display a link to a downloaded file is empty.  Also note this only occurs upon a new file upload and immediately trying to retrieve the public link, or after a refresh of the app or browser.  There are is more discussion in the comments section of the Linked Jira ticket.

The current solution disables the `Copy Link` button until the link appears.  There is also a suggestion to add the <LoadingSpinner> component until the text is visible.  The text is currently a <textarea> element and I am not sure how to superimpose the spinner on top of the <textarea>. 

I didn't want to add the spinner element and then replace it with the textarea element because the separation of components would visibly shift, causing a glitchy look.   Any advice?  

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23580

#### Related Pull Requests
N/A

#### Screenshots
Original issue shared by reporter
image

Link to video in Jira. This is the current solution which disables the button until the link shows.
blob:https://mattermost.atlassian.net/1cd4b135-fd0e-4934-8d74-76c05d882119

### DIFF
--- a/components/get_link_modal.tsx
+++ b/components/get_link_modal.tsx
@@ -64,6 +64,11 @@ export default class GetLinkModal extends React.PureComponent<Props, State> {
             );
         }
 
+        let disableButton = true;
+        if (this.props.link !== '') {
+            disableButton = false;
+        }
+
         let copyLink = null;
 
         if (document.queryCommandSupported('copy')) {
@@ -73,6 +78,7 @@ export default class GetLinkModal extends React.PureComponent<Props, State> {
                     data-copy-btn='true'
                     type='button'
                     className='btn btn-primary pull-left'
+                    disabled={disableButton}
                     onClick={this.copyLink}
                 >
                     <FormattedMessage
--- a/components/__snapshots__/get_link_modal.test.tsx.snap
+++ b/components/__snapshots__/get_link_modal.test.tsx.snap
@@ -72,6 +72,7 @@ exports[`components/GetLinkModal should have called onHide 1`] = `
     <button
       className="btn btn-primary pull-left"
       data-copy-btn="true"
+      disabled={false}
       id="linkModalCopyLink"
       onClick={[Function]}
       type="button"
@@ -168,6 +169,7 @@ exports[`components/GetLinkModal should have called onHide 2`] = `
     <button
       className="btn btn-primary pull-left"
       data-copy-btn="true"
+      disabled={false}
       id="linkModalCopyLink"
       onClick={[Function]}
       type="button"
@@ -259,6 +261,7 @@ exports[`components/GetLinkModal should match snapshot when all props is set 1`]
     <button
       className="btn btn-primary pull-left"
       data-copy-btn="true"
+      disabled={false}
       id="linkModalCopyLink"
       onClick={[Function]}
       type="button"
@@ -345,6 +348,7 @@ exports[`components/GetLinkModal should match snapshot when helpText is not set 
     <button
       className="btn btn-primary pull-left"
       data-copy-btn="true"
+      disabled={false}
       id="linkModalCopyLink"
       onClick={[Function]}
       type="button"


jfrerich:
@lieut-data, @andrewbrown00 - I went through quite a few iterations since creating this PR, but ready for review.  Notably, I discovered it was possible to copy the link by actually clicking on the text box, not only the `Copy Link` button.

Note that the current change will do the following until the link appears.
* disable copying the link from the button
* disable copying the link form the textarea 
* The `Copy Link` button will appear with no text, but include the Loading Spinner
  * I debated on including the `Copy Link` text here, but it kind of appears like the link is being copied once opening the modal.  In reality, we are just waiting for the link to show.

Once the link appears:
* enable all copy functions (button and textarea)
* transfer button text from the loading spinner to `Copy Link`

lieut-data:
Thanks, @jfrerich -- are you able to capture an animated screenshot illustrating the changes here to help @andrewbrown00 review the proposed tweak?

components/get_link_modal.tsx
```diff
@@ -65,6 +66,7 @@ export default class GetLinkModal extends React.PureComponent<Props, State> {
         }
 
         let copyLink = null;
+        const linkAvailable = this.props.link;
```
lieut-data, line 69: Can we make this boolean explicit?
```suggestion
         const linkAvailable = Boolean(this.props.link);
```

jfrerich: Added in the latest commit.

jfrerich:
> Thanks, @jfrerich -- are you able to capture an animated screenshot illustrating the changes here to help @andrewbrown00 review the proposed tweak?

@lieut-data.  Yes, I placed two new `.mp4` files in the Jira ticket.  (it's a shame we can't use .mp4 in github.  They are way smaller than .gif files and you can play/stop the video).   

The videos show 100kb/s and 10kb/s DL and UL connections.  At 10kb, you can see the result quite easily because the connection is so slow.  At 100kb/s, you can still see the corner case, although it is hard to catch now.  >100kb/s, the user will probably not ever see this case, and if they do, it's going to be difficult to reproduce.  You basically have to race to 1) refresh the page, 2) click the file, 3) click the `get public link`

jfrerich:
@andrewbrown00, I have two questions.  

1) Reserving the `Copy Link` button width 
width is defined in pixels. Do you suggest hard-coding this value, or somehow calculating the # of pixels on the fly, depending on the text in the button.  One thing to consider is the width of the button changing based on localization.

2) I discovered a <SpinnerButton> component.  Are you ok if I swap the current implementation for that component?

andrewbrown00:
Hey @jfrerich,

0/5 on how this is implemented, I'm going to defer to @deanwhillier to see if he can provide guidance

deanwhillier:
Thanks @andrewbrown00!

Stepping out of our webapp and currently available components for a minute, I would solve this by wrapping the txt of the button in a `<span>` tag and then based on a JS toggled state class on the parent element, animate the opacity of the `<span>` tag between 0 and 1 with a short CSS transition to allow it to fade in and out nicely.

Following that I would set the spinner icon to `position: absolute` (parent element needs `position: relative`) within the button and centre it using a combination of `margin-{top&left}: 50%` and `transform: translate3d(-50%, -50%, 0)` (or some other appropriate centring mechanism). Then apply the reverse transitioned `opacity` values based on the state class of the parent `button` element. The reverse transition will allow for a really nice, quick fade between the text and the spinner icon (150ms for instance). In the case of the spinner, I would also toggle between `visibility: visible` and `visibility: hidden` so it isn't sitting there spinning transparently when the button is not in a loading state. You could get fancier here and add a JS listener for `transitionend` on the spinner and then completely remove it from the DOM once it has transitioned to fully transparent, though you would have to re-add it back in if a scenario called for transitioning back and forth between loading and loaded states.

The benefit of this as an approach is that the button width is defined by the text and stays defined by the text when the text isn't visible without any complex calculations to explicitly set the width in pixels. Animating the `opacity` is hardware accelerated and doesn't require repainting in the browser so the transition should be quick and smooth without causing any performance degradation.

Stepping back into our webapp and available components, the downside of the way the `LoadingSpinner` is currently implemented is that it doesn't render the DOM elements for the text while in a loading state and conversely doesn't render the spinner when not in a loading state making it impossible to visually transition between the two or allowing the text to set and maintain the button width when the spinner is visible (as you are experiencing).

Happy to assist on the CSS front if you need (I know CSS can seem more like magic/mess 😛). Animating `opacity` and `visibility` are especially challenging if you're not familiar with working with transitions and visibility.

Let me know if you'd like to chat about this.

jfrerich:
@deanwhillier, Thanks for the extremely detailed suggestion. As this is a corner case, and not completely urgent, I'd like to iterate o this for a bit before reaching out.  This will be a learning experience for me and I'd like to give it some mana before just asking for help. I'll ping you at some point for sure!

jfrerich:
long_text_button

jfrerich:
@deanwhillier, thanks for the informative 1:1 earlier.  I have a new commit that I believe looks visually correct, although I had to introduce some extra `<span>` elements to get this working.  

Remember that we were seeing an issue when using the `margin-top: 50%` on the spinner. Somehow the height was changing considerably and causing the spinner to display below the button.

cpanato:
will terminate this test instance to move to another cluster

lieut-data:
Closing old PR.

PR #5462, opened 2020-05-08
MM-24715: fix getstate refactor

lieut-data:
### Summary
In testing the channel export plugin, we ran into issues with closed DMs not automatically re-opening when a new post was made. I traced the regression back to https://github.com/mattermost/mattermost-webapp/pull/4402, and fixed a few issues invoking `loadNewDMIfNeeded` and `getUserIdFromChannelName`.

I've broken out the commits for easier readability, but also commented below.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24715

### DIFF
--- a/actions/channel_actions.jsx
+++ b/actions/channel_actions.jsx
@@ -79,9 +79,9 @@ export function loadChannelsForCurrentUser() {
         for (const id of unreads) {
             const channel = getChannel(state, id);
             if (channel && channel.type === Constants.DM_CHANNEL) {
-                loadNewDMIfNeeded(channel.id);
+                dispatch(loadNewDMIfNeeded(channel.id));
             } else if (channel && channel.type === Constants.GM_CHANNEL) {
-                loadNewGMIfNeeded(channel.id);
+                dispatch(loadNewGMIfNeeded(channel.id));
             }
         }
 
--- a/actions/user_actions.jsx
+++ b/actions/user_actions.jsx
@@ -173,7 +173,7 @@ export function loadNewDMIfNeeded(channelId) {
     };
 }
 
-export async function loadNewGMIfNeeded(channelId) {
+export function loadNewGMIfNeeded(channelId) {
     return async (doDispatch, doGetState) => {
         const state = doGetState();
         const currentUserId = Selectors.getCurrentUserId(state);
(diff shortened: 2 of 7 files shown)


actions/channel_actions.jsx
```diff
@@ -79,9 +79,9 @@ export function loadChannelsForCurrentUser() {
         for (const id of unreads) {
             const channel = getChannel(state, id);
             if (channel && channel.type === Constants.DM_CHANNEL) {
-                loadNewDMIfNeeded(channel.id);
+                dispatch(loadNewDMIfNeeded(channel.id));
```
lieut-data, line 82: `loadNew(DM|GM)IfNeeded` is an action that must be dispatched -- calling it plain just returned an unused async function.

actions/user_actions.jsx
```diff
@@ -143,8 +143,8 @@ export function loadNewDMIfNeeded(channelId) {
         const state = doGetState();
         const currentUserId = Selectors.getCurrentUserId(state);
 
-        function checkPreference() {
-            const userId = Utils.getUserIdFromChannelName(currentUserId);
+        function checkPreference(channel) {
+            const userId = Utils.getUserIdFromChannelName(channel);
```
lieut-data, line 147: I stumbled over `getUserIdFromChannelName` a few times -- it actually accepts a `channel` and not a channel name, and initially tried to refactor this, but realized there's code duplication here with Redux and opted just to keep this PR simpler.

actions/user_actions.jsx
```diff
@@ -173,7 +173,7 @@ export function loadNewDMIfNeeded(channelId) {
     };
 }
 
-export async function loadNewGMIfNeeded(channelId) {
+export function loadNewGMIfNeeded(channelId) {
```
lieut-data, line 176: This `async` wasn't actually being used, but ended up tripping a test once we started dispatching this method. Fixed to align with `loadNewDMIfNeeded` which wasn't `async`.

actions/channel_actions.jsx
hmhealey: I wonder if this is something we could add an ESLint check for at some point. You'd have to find a function call that returns a redux action and then ensure that something is done with it. I guess that requires ESLint to understand types though, so that would probably be difficult.

PR #5463, opened 2020-05-08
MM-24918: Make sure flagged posts and recent mentions from old channel are excluded in RHS

agarciamontoro:
#### Summary
This PR fixes an issue with the RHS: when a user has the Recent Mentions RHS or the Flagged Posts RHS open and they are kicked out of a channel, the messages from that channel should disappear. This was introduced in https://github.com/mattermost/mattermost-webapp/pull/5320, that removed the dispatching of the `LEAVE_CHANNEL` event when a user was removed from a channel, as its position was causing race conditions when using the `/leave` command.

The thing is that we need to:
1. dispatch the `LEAVE_CHANNEL` event in order to fix https://mattermost.atlassian.net/browse/MM-24918.
2. in case the user is leaving a channel, make sure that we dispatch this event after we have pushed the new URL to the browser history in order to fix https://mattermost.atlassian.net/browse/MM-23994
3. in case the user is being kicked out of the channel, make sure that we dispatch this event before calling `redirectUserToDefaultTeam` in order to fix https://mattermost.atlassian.net/browse/MM-23589

The changes in this PR accomplish that by just dispatching the event when we need it. It's not the most elegant solution, but we want to have this in v5.23 and, given the time constraints, I thought this was the safest way to do it without breaking anything else.

I have also added e2e tests to make sure we don't lose this functionality again in the future.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24918

### DIFF
--- a/actions/websocket_actions.jsx
+++ b/actions/websocket_actions.jsx
@@ -769,6 +769,11 @@ export async function handleUserRemovedEvent(msg) {
         if (msg.data.channel_id === currentChannel.id) {
             if (msg.data.remover_id === msg.broadcast.user_id) {
                 browserHistory.push(getCurrentRelativeTeamUrl(state));
+
+                await dispatch({
+                    type: ChannelTypes.LEAVE_CHANNEL,
+                    data: {id: msg.data.channel_id, user_id: msg.broadcast.user_id},
+                });
             } else {
                 const user = getUser(state, msg.data.remover_id);
                 if (!user) {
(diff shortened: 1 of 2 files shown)


agarciamontoro:
Asked for three reviews so that it is easier for this to get reviewed sooner. The first two that review it win! :tada:

agarciamontoro:
Removing review request from @sudheerDev because it's already too late in their timezone. Hopefully @mkraft can take a look :)

PR #5465, opened 2020-05-08
MM-24086:Disable changing picture if LDAP sync'ed

sbishel:
#### Summary
When LDAP sync is turned on and a Picture attribute is selected, the user will not be able to change the profile picture. Instead they will be displayed a message that the picture is controlled by LDAP settings. This is the same way other LDAP Attributes work.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24086

#### Screenshots
Screen Shot 2020-05-08 at 3 51 08 PM


components/setting_picture.jsx
```diff
@@ -286,6 +261,55 @@ export default class SettingPicture extends Component {
...
+                            defaultMessage='Select'
+                        />
+                    </button>
+                    <button
+                        tabIndex={disableSaveButtonFocus ? '-1' : '0'}
+                        data-testid='saveSettingPicture'
+                        disabled={disableSaveButtonFocus}
+                        ref={this.confirmButton}
+                        className={confirmButtonClass}
+                        onClick={this.props.loadingPicture ? () => true : this.handleSave}
```
mkraft, line 301: The `() => true` function will create a new function every time causing unnecessary re-renders downstream vs creating a `this.returnTrue` method on the `SettingPicture` component.

components/setting_picture.jsx
```diff
@@ -286,6 +269,55 @@ export default class SettingPicture extends Component {
...
+                            defaultMessage='Select'
+                        />
+                    </button>
+                    <button
+                        tabIndex={disableSaveButtonFocus ? '-1' : '0'}
+                        data-testid='saveSettingPicture'
+                        disabled={disableSaveButtonFocus}
+                        ref={this.confirmButton}
+                        className={confirmButtonClass}
+                        onClick={this.props.loadingPicture ? this.returnTrue : this.handleSave}
```
hahmadia, line 305: I'm not a huge fan of how this is being done.  Could we do something like this possibly?
```suggestion
                        onClick={this.handleSave}
```
and then in the `handleSave()` function, we have something like this 
```Javascript
handleSave() {
     e.preventDefault()
     if (this.props.loadingPicture) {
            return;
     }
    // ... Rest of code
}
```

sbishel: I like it...done.

PR #5467, opened 2020-05-11
[MM-24052] Add animation to emoji reaction

bradjcoughlin:
#### Summary
Adds odometer-style animation when reacting to a post.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24052

#### Screenshots
reaction-animation-v2


bradjcoughlin:
@andrewbrown00 Thanks!

- Speed up animation: I think I left this slower than I intended for development reasons. I'll decrease the transformation duration first so you can see how that feels
- Don't animate if reaction is going to zero: If the reaction is being removed altogether, I can prevent the animation
- Add animation to new emoji: I can do this but it would increase scope/mana (possibly significantly)
- Tweak the emoji reaction container: Can you expand on that thought?

As a general aside, I'm trying to be cognizant of the performance implications of adding animation and handling of animation to every emoji reaction.

andrewbrown00:
Thanks @bradjcoughlin 

* Speed up animation: **Feels much better now**
* Don't animate if reaction is going to zero: If the reaction is being removed altogether, I can prevent the animation **It may be nice to have a subtle `fade-out` if it's not a performance issue**
* Add animation to new emoji: I can do this but it would increase scope/mana (possibly significantly) **We can create a separate ticket for this. What do you think the mana would be?**
* Tweak the emoji reaction container: Can you expand on that thought? **For sure, right now when you click to add a reaction the effect is instant on the border of the emoji reaction container vs having a subtle `ease` on the `border` of the container. **
* As a general aside, I'm trying to be cognizant of the performance implications of adding animation and handling of animation to every emoji reaction. **5/5 agree, we can look at the performance tomorrow on our call and see what type of performance hits we're looking at**

bradjcoughlin:
Working to update unit tests but wanted to push the new changes so you (@andrewbrown00 @esethna ) can try it out and look for bugs.

With this commit I've changed the behavior so that the redux/server event fire immediately after the animation. This should speed up the frame rates and reduce judder because the page isn't re-rendering or painting until after the animation is complete.

I've also added transition animation to the border and background colors at the same speed as the number animation.

bradjcoughlin:
@esethna it's at 200ms currently. I can easily bump it down, but I wonder if it it won't end up showing only a few animated frames on slower machines.

I definitely wouldn't got about 200ms, but there are tradeoffs decreasing the timing:

60 fps @ 1 sec - Decent GPU framerate
60 * .2 = 12 frames - 33% more than 150ms. May look smoother but feel slower
60 * .15 = 9 frames - May feel faster but look choppier

My vote is for slightly more frames but I leave it in your hands @andrewbrown00

bradjcoughlin:
@andrewbrown00 Great. I will address the first two points (though I wish I could edit the database directly). I may just have to create 8 more users : )

On the third point, animating the reaction component in and out of the DOM requires a transition wrapper. This is definitely something we can and should do, but it would be easier to tackle it as a separate layer. Mainly because of the code required but also to separate the concerns of animation within the component and the animation wrapping the component.

components/post_view/reaction/reaction.jsx
```diff
@@ -10,88 +10,152 @@ import OverlayTrigger from 'components/overlay_trigger';
...
+
+        this.reactionButtonRef = React.createRef();
+        this.reactionCountRef = React.createRef();
+        this.reactedNumeralRef = React.createRef();
     }
 
-    handleRemoveReaction = (e) => {
-        e.preventDefault();
-        this.props.actions.removeReaction(this.props.post.id, this.props.emojiName);
+    shouldComponentUpdate(nextProps) {
```
hmhealey, line 76: We shouldn't be using `setState` here. Could this be done in `getDerivedStateFromProps` instead?

components/post_view/reaction/reaction.jsx
```diff
@@ -10,88 +10,152 @@ import OverlayTrigger from 'components/overlay_trigger';
...
+                    reactedClass: 'Reaction reacted',
+                    displayNumber: reactedNumber,
+                };
+            }
+            return {
+                reactedClass: 'Reaction unreacted',
+                displayNumber: unreactedNumber,
+            };
+        });
+        if (this.state.userReacted) {
```
hmhealey, line 154: Perhaps it would be better to do this in response to the `setState` call in `handleClick` using `setState`'s callback parameter? I didn't expect to see these actions being fired in response to something other than the click event.

I suppose that could cause issues if the callback returns while we're still animating and updates the props though.

components/post_view/reaction/reaction.jsx
```diff
@@ -10,88 +10,152 @@ import OverlayTrigger from 'components/overlay_trigger';
...
+                displayNumber: reactionCount,
+                reactedNumber: reactionCount + 1,
+                unreactedNumber: reactionCount,
+            };
+        }
+
+        if (currentUserReacted && !canRemoveReaction) {
+            this.state = {
+                ...this.state,
+                reactedClass: 'Reaction read-only',
```
hmhealey, line 62: Instead of storing the whole CSS class in the state, maybe it would make sense to derive it in the `render` method? "Reaction" will always be there and "read-only" is inferred by the props, so the only thing that really needs to be in state is the transitions like "reacting"

components/post_view/reaction/reaction.jsx
```diff
@@ -103,10 +167,13 @@ export default class Reaction extends React.PureComponent {
         if (!this.props.emojiImageUrl) {
             return null;
         }
-
         const {currentUserReacted, users} = this.props.sortedUsers;
+        const {otherUsersCount, canAddReaction, canRemoveReaction} = this.props;
+        const {unreactedNumber, reactedNumber, displayNumber} = this.state;
+        const unreacted = (unreactedNumber > 0) ? unreactedNumber : '';
```
hmhealey, line 173: Why have these default to the empty string instead of 0?

Also, it might be easier to follow this if these ternaries were inline in the JSX below, although 0/5 about that.

components/post_view/reaction/reaction.jsx
```diff
@@ -207,28 +271,26 @@ export default class Reaction extends React.PureComponent {
                     />
                 );
             } else {
-                className += ' post-reaction--read-only';
+                //className += ' Reaction--read-only';
```
hmhealey, line 274: There's some left over code that can be cleaned up.

components/post_view/reaction/reaction.jsx
nevyangelova: Also can we use BEM for readability?

components/post_view/reaction/reaction.jsx
```diff
@@ -10,88 +10,152 @@ import OverlayTrigger from 'components/overlay_trigger';
...
-             * Function to add a reaction to a post
-             */
             addReaction: PropTypes.func.isRequired,
-
-            /*
-             * Function to get non-loaded profiles by id
-             */
             getMissingProfilesByIds: PropTypes.func.isRequired,
-
-            /*
```
nevyangelova, line 78: Why don't we keep the comments here?

components/post_view/reaction/reaction.jsx
```diff
@@ -10,88 +10,152 @@ import OverlayTrigger from 'components/overlay_trigger';
...
+                unreactedNumber: reactionCount,
+            };
+        }
+
+        if (currentUserReacted && !canRemoveReaction) {
+            this.state = {
+                ...this.state,
+                reactedClass: 'Reaction read-only',
+            };
+        } else if (!currentUserReacted && !canAddReaction) {
```
nevyangelova, line 64: this can be simplified `!(currentUserReacted || canAddReaction)`

components/post_view/reaction/reaction.jsx
```diff
@@ -10,88 +10,152 @@ import OverlayTrigger from 'components/overlay_trigger';
...
+                });
+                return true;
+            }
+        }
+        return true;
+    }
+
+    handleClick = () => {
+        // only proceed if user has permission to react
+        if (!this.props.canAddReaction || !this.props.canRemoveReaction) {
```
nevyangelova, line 122: this can be simplified `!(canAddReaction && canRemoveReaction)`

components/post_view/reaction/reaction.jsx
```diff
@@ -245,13 +307,23 @@ export default class Reaction extends React.PureComponent {
                 >
                     <span className='d-flex align-items-center'>
                         <span
-                            className='post-reaction__emoji emoticon'
+                            className='Reaction__emoji emoticon'
```
nevyangelova, line 310: can we use the previous class name format here too? I am 3/5 to not use capital letters :)

bradjcoughlin: I was following the widget component pattern, but I’m happy to standardize going forward. If we change these then we should probably change the widgets : )

components/post_view/reaction/reaction.jsx
bradjcoughlin: Yep, thanks.

components/post_view/reaction/reaction.jsx
bradjcoughlin: You’re right, thanks

components/post_view/reaction/reaction.jsx
bradjcoughlin: Thanks

components/post_view/reaction/reaction.jsx
bradjcoughlin: I could, let me look at that again. The render method was really messy previously with all of the conditions. But I agree the parent class name doesn’t need to change.

components/post_view/reaction/reaction.jsx
bradjcoughlin: The primary reason for calling here is to prevent the app from repainting during the animation. In Chrome Dev Tools calling the redux action during the animation caused re-painting in several places and slow the animation down to just a few frames.

components/post_view/reaction/reaction.jsx
bradjcoughlin: The empty string is so the animation doesn’t display a zero when removing a reaction. All aesthetics. I kept this logic up here just to keep the html elements readable below.

components/post_view/reaction/reaction.jsx
bradjcoughlin: I was going to switch this back to BEM format and forgot. Will do that now.

components/post_view/reaction/reaction.jsx
bradjcoughlin: I can, just removed for convenience when I was working. But I do wonder what our standard is regarding commenting on default props. Some have everything commented, others have nothing 🤷‍♂️

components/post_view/reaction/reaction.jsx
bradjcoughlin: I went back and forth on this a lot given the "you probably don't need `getDerivedStateFromProps`, but I think we might in this scenario. I had this logic under there originally.

bradjcoughlin: Long story short, I figure out how to do this in `getDerived...`. Because it fires frequently, you have to return the derived state when all of the props conditions are true. If you nest the conditions, it gets confusing as the props trickle in sequentially as they update.

components/post_view/reaction/reaction.jsx
bradjcoughlin: This should all match BEM conventions now, plus I'm only rendering the modifier classes now as the state updates.

components/post_view/reaction/reaction.jsx
bradjcoughlin: Leaving as capitalized for now because I'd like to have a discussion about this as a team.

andrewbrown00:
> On the third point, animating the reaction component in and out of the DOM requires a transition wrapper. This is definitely something we can and should do, but it would be easier to tackle it as a separate layer i.e. ticket. Mainly because of the code required but also to separate the concerns of animation within the component and the animation wrapping the component. They could still be released together for a harmonious result 🎷 😄

@bradjcoughlin thanks for the context here, let's create a separate ticket and we can prioritize to handle later on.

bradjcoughlin:
@nevyangelova @hmhealey I realized the handling of the read-only class wasn't quite correct. I tested this commit by switching Town Square from read-only to normal while viewing a reaction from a non-admin user. I wouldn't have to do this in `getDerived...` necessarily but it seemed clearer to leave that logic in the same lifecycle as the other state logic.

nevyangelova:
Nice find @bradjcoughlin  :)

components/post_view/reaction/reaction.jsx
```diff
@@ -83,15 +85,155 @@ export default class Reaction extends React.PureComponent {
...
+        }
+
+        this.reactionButtonRef = React.createRef();
+        this.reactionCountRef = React.createRef();
+        this.reactedNumeralRef = React.createRef();
+    }
+
+    static getDerivedStateFromProps(nextProps, prevState) {
+        // reaction count has changed, but not by current user
+        if (nextProps.reactionCount !== prevState.displayNumber && nextProps.sortedUsers.currentUserReacted === prevState.userReacted) {
```
hmhealey, line 133: What if multiple of these props change at the same time? This logic seems incorrect if, for example, the permissions and reaction count change at the same time.

components/post_view/reaction/reaction.jsx
hmhealey: What about using the `setState` callback to trigger it then? Does that still cause the repaint?

components/post_view/reaction/reaction.jsx
```diff
@@ -83,15 +85,155 @@ export default class Reaction extends React.PureComponent {
...
+                unreactedNumber: nextProps.reactionCount,
+            };
+        }
+
+        // if permission to add reactions has changed
+        if (nextProps.canAddReaction !== prevState.canAddReaction) {
+            // if can't add reaction
+            if (!nextProps.canAddReaction) {
+                return {
+                    reactedClass: 'Reaction--read-only',
```
hmhealey, line 156: Why store the read only class in the state? Couldn't we apply both the `Reaction--reacted` and `Reaction--read-only` classes to the component at the same time and have the `read-only` one take precedence? If we do that, we'll have less stuff to track in the component state.

components/post_view/reaction/reaction.jsx
bradjcoughlin: I can try it out locally, but my guess is anything that start to change redux state during the animation is going to slow down FPS because of the non-UI react lifecycle processing plus the UI repaint. Is there a reason you're concerned about waiting until the animation is done?

components/post_view/reaction/reaction.jsx
bradjcoughlin: I understand what you're saying, because I was also little confused around the design intent (trying to understand from the CSS). It seems there is only one "read-only" style that doesn't discriminate between whether the current user has reacted or not. I can make this change and see if it still passes the use-cases for QA.

bradjcoughlin: Updated per our discussion @hmhealey

components/post_view/reaction/reaction.jsx
bradjcoughlin: I've tested this thoroughly with several concurrent users so I don't think there's a problem, especially now that the canAdd/canRemove permissions aren't being handled here.

Essentially if the reaction of this user has NOT changed, but the reaction count has changed, we update the numbers (either up or down). Any other changes from props will be ignored.

components/post_view/reaction/reaction.jsx
bradjcoughlin: I inspected the performance and I'm convinced the animation should fire first, with the action triggered immediately after. I can see a ton of recomputes in the virtual DOM that do surge the CPU/GPU performance. The result is a drop in frame-rate. It is subtle, but the animation is so short it does make a difference (GIF at 30fps):
anim-comparison

jgilliam17:
Thank you @bradjcoughlin 
I see one small issue on read-only channel
If admin adds/removes an emoji, emojis are rearranged and ordered with the highest count on the left going to lowest on the right. See 😀 emoji jump to second to the last position from first.
I'm going to check on master and update here, if it is the same there.
Jun-02-2020 16-09-20

jgilliam17:
Update: Checked on master.test, not seeing the reordering there

cpanato:
terminating this test server because we need to clean the cluster that is running. If you still need a test server please add the label again in the future

esethna:
@andrewbrown00 is this UX approved? 

@bradjcoughlin @deanwhillier should we have someone take over this PR?

andrewbrown00:
@esethna there are some polish things I'd like to have addressed, namely around the make the transitions smoother and not so jumpy.

esethna:
Thanks @andrewbrown00, agreed, but I don't want to block this PR much longer, can you help file separate tickets as needed? 

I think we definitely need to figure out the animation for adding a new reaction and how we can make it optimistic vs waiting for the server before adding. Perhaps utilizing an animation is how we can do that.

esethna:
@calebroseland can you help address Jelena's feedback given it requires an MM license?

calebroseland:
> @calebroseland can you help address Jelena's feedback given it requires an MM license? 

@esethna sure, I'll take a look on Monday.

calebroseland:
@jgilliam17 could you help me out with the repro steps? I was able to reproduce _once_ on the cloud instance, but I can't repro again, and I could not repro locally at all.

I tried: 
Screen Shot 2020-07-10 at 14 26 09

And toggled emojis with an Admin user, but could not get it to happen locally. (Chrome, Safari, FF)

jgilliam17:
@calebroseland I am no longer able to reproduce on the server here. 
Adding/removing emoji works as expected, no jumping positions on read only channels:
- After enabling `Town Square is Read-Only` experimental feature 
- Removing all permissions, essentially making channel read only

PR #5468, opened 2020-05-11
MM-25386 Add support for PostgreSQL & PL/pgSQL syntax highlighting

ejose19:
#### Summary

Enables support for PostgreSQL & PL/pgSQL that is already supported on highlight.js library

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25386

#### Related Pull Requests

Implemented following #5347


teresa-novoa:
Thanks @ejose19 for your contribution! We will assign a Developer to review it. cc @hanzei

teresa-novoa:
@bradjcoughlin friendly reminder to review this PR, thanks!

jgilliam17:
Thank you @ejose19 
Highlighting looks great when using `pqsql`?
Quick follow up question: 
Should the highlighting also apply when using `postgresql` and `postgres`?
Let me know. Thanks.

ejose19:
Hello @jgilliam17, I've tested it and here are my observations:

Highlighting works great, but there are 3 minor bugs that apparently are related to mattermost implementation (those don't show when using the library alone)

1. highlight breaks when using $anyword$ instead of $$

Using $$myfunc$$
image
Using $$
image


2. The opening $$ is adding a newline that shouldn't be there

Notice line 2, it's a single line taking 2 spaces
image

3. The closing $$ is taking away spaces between other words in the same line
( should be $$ LANGUAGE plpgsql )

image


 I implemented it following https://github.com/highlightjs/highlight.js/blob/master/SUPPORTED_LANGUAGES.md which uses those 3 classes to apply the highlight. When testing in the browser just using the library alone, all 3 worked, but I'm not sure if it's because autodetection or indeed all 3 classes map to PostgreSQL syntax, perhaps we should only leave pgsql? Seems people will use it as it's shorter, the other option would be adding an alias for those 2.

Overall for people using postgresql it's great since highlight doesn't work on ```sql when using postgres keywords, so even with those small inconveniences the experience will be way better.

jgilliam17:
Thank you @ejose19 for a quick reply
>When testing in the browser just using the library alone, all 3 worked, but I'm not sure if it's because autodetection or indeed all 3 classes map to PostgreSQL syntax, perhaps we should only leave pgsql? Seems people will use it as it's shorter, the other option would be adding an alias for those

Let me get some feedback from the dev team about this and also about the bugs you mentioned.

jgilliam17:
Thank you for your patience @ejose19 
Can you please add aliases for `postgres` & `postgresql`, that way we would cover all the use cases.
As far as bugs are concerned, I'll create a follow up ticket to address those separately. Let me know if you have any questions or need more info.
Thank you again for your contribution 🙂

jgilliam17:
> /update-branch

@ejose19 can you please update the branch
I'm seeing a blank screen when opening the test server (I never tried updating before, no permissions I guess 🤔)

ejose19:
@jgilliam17 I'm also seeing that and was waiting for up to update to test

ejose19:
Ok, seems it's all good, all 3 works now.

PR #5470, opened 2020-05-11
[MM-24507] Affix comment input to bottom in right-hand side (RHS)

bradjcoughlin:
#### Summary
Currently, the comment input box will scroll out of view on longer threads. This change moves the input out of the scrollbars so it stays visible at all times.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24507


bradjcoughlin:
@esethna Yep, that's right. Back to the drawing board.

bradjcoughlin:
@esethna this is ready to test but the builds are failing. I'll check on it again in awhile.

esethna:
@bradjcoughlin ^

bradjcoughlin:
@esethna can you explain what's going on there? Is it getting cut off even when scrolled to the bottom?

esethna:
@bradjcoughlin there is a padding around the RHS input box that shouldn't be there (see the center pane in comparison).

bradjcoughlin:
@esethna got it. I set the background color to solid so I need to remove the padding or margin above the input box now. Stay tuned.

components/rhs_thread/rhs_thread.test.jsx
```diff
@@ -48,26 +48,32 @@ describe('components/RhsThread', () => {
 
     test('should match snapshot', () => {
         const wrapper = shallow(
-            <RhsThread {...baseProps}/>
+            <RhsThread {...baseProps}/>,
+            {disableLifecycleMethods: true},
```
bradjcoughlin, line 52: Disabling lifecycle method here because `componentDidUpdate` is calling the `ref` to `<Scrollbars/>`, throwing an `undefined` error.

components/rhs_thread/rhs_thread.test.jsx
```diff
@@ -48,26 +48,32 @@ describe('components/RhsThread', () => {
...
         );
-
-        wrapper.setProps({socketConnectionStatus: false});
+        const prevProps = {
+            ...baseProps,
+            socketConnectionStatus: false,
+        };
         wrapper.setProps({socketConnectionStatus: true});
+        wrapper.instance().componentDidUpdate(prevProps);
```
bradjcoughlin, line 67: To test the function within the lifecycle method, I'm manually calling `componentDidUpdate` and passing in a mock `prevProps` object.

asaadmahmood:
A few things.
1. When I click on the reply icon, it takes me to the top of the thread in the RHS, I expect the scroll/focus to be on the last message as it works right now in Mattermost. (I want to see where the conversation is right now).
2. There should be a padding given to the list of comments so that in the default state, the edge of the post does not align with the textbox, rather has some space.
This space however should not be visible if a user is scrolling and comments are being cropped by the comment box (similar to the issue Eric pointed out earlier). This would be done if the padding is given to list of comments instead of the comment button.
**Center Channel**
Screenshot 2020-05-15 at 4 42 24 AM
**RHS**
Screenshot 2020-05-15 at 4 42 29 AM

bradjcoughlin:
@asaadmahmood 
1. Got it. When the input was moved out of the scrollable area, the `focus` does not cause it to be scrolled into view. I can address this behavior in a different way.
2. This also makes sense.

jgilliam17:
Thank you @bradjcoughlin 
Looks good, just seeing one error when opening RHS in MobileWeb view.
Screen Shot 2020-05-19 at 9 51 42 AM

bradjcoughlin:
@jgilliam17 ah, that looks like a typescript issue.

jgilliam17:
@bradjcoughlin I am not able to open this server after update. Opens blank screen.

bradjcoughlin:
@jgilliam17 I opened the test server in Safari and discovered that scrolling in the RHS is working at all. Investigating now.

jgilliam17:
> @jgilliam17 I opened the test server in Safari and discovered that scrolling in the RHS isn't working at all. Investigating now.

Great catch, I missed that.
I checked Edge and desktop app on Win10 - no issue
Also, Chrome, Firefox and desktop app on macOS Catalina - no issue there either.

amyblais:
@bradjcoughlin Looks like this wasn't cherry-picked yet.

PR #5480, opened 2020-05-12
[MM-14912] Add channel members block to channel details page

fmunshi:
#### Summary
- Adds the channel members block to the channel details page builds off of https://github.com/mattermost/mattermost-webapp/pull/5469
- Adds a single new component (`TeamMembers`)
- Modifies the existing `channel_invite_modal` to allow for users to be added without committing and specifying users to include and exclude

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-14912

#### Related Pull Requests
- Has redux changes: https://github.com/mattermost/mattermost-redux/pull/1150

#### Screenshots
Screen Shot 2020-05-12 at 2 30 13 PM
Screen Shot 2020-05-12 at 2 30 07 PM

I have requested review from @devinbinnie since this touches a component (channel_invite_modal) that is also used on the client.


components/channel_invite_modal/index.js
```diff
@@ -6,16 +6,34 @@ import {bindActionCreators} from 'redux';
...
-    return {
-        profilesNotInCurrentChannel: getProfilesNotInCurrentChannel(state),
-        profilesNotInCurrentTeam: getProfilesNotInCurrentTeam(state),
+function makeMapStateToProps(initialState, initialProps) {
+    let doGetProfilesNotInChannel;
+    if (initialProps.channelId && initialProps.teamId) {
+        doGetProfilesNotInChannel = makeGetProfilesNotInChannel();
+    }
+
+    return function mapStateToProps(state, props) {
```
devinbinnie, line 21: Non blocking, but could this be an arrow function? That's usually the standard we go with for in-line functions.

components/channel_members_modal/channel_members_modal.tsx
```diff
@@ -31,7 +31,7 @@ type Props = {
         openModal: (modalData: {
             modalId: string;
             dialogProps: {};
-            dialogType: (props: {}) => React.ReactElement | null;
+            dialogType: (props: any) => React.ReactElement | null;
```
devinbinnie, line 34: Do we not have a type for this? We should try to avoid `any` wherever we can.

fmunshi: The issue is that the component being referenced is not typescript so I cant simply import the props for that component into here - is it worth creating the prop types in this file? just to add it as a type here?

devinbinnie: Ahh, that's unfortunate. No that's alright, we can import them in the future once that component is converted.

fmunshi:
Build is failing due to master having issues too... not related to the PR

components/admin_console/team_channel_settings/channel/details/channel_details.tsx
```diff
@@ -421,11 +440,77 @@ export default class ChannelDetails extends React.Component<ChannelDetailsProps,
             privacyChanging = false;
         }
 
-        this.setState({serverError, saving: false, saveNeeded, isPrivacyChanging: privacyChanging});
+        const usersToAddList = Object.values(usersToAdd);
+        const usersToRemoveList = Object.values(usersToRemove);
+        const userRolesToUpdate = Object.keys(rolesToUpdate);
+        const usersToUpdate = usersToAddList.length > 0 || usersToRemoveList.length > 0 || userRolesToUpdate.length > 0;
```
catalintomai, line 446: I'm not sure about this name(`userRolesToUpdate`), doesn't jive with the other two imo, but I don't have a better solution (maybe `usersWithRolesToUpdate`) 1/5 to look for a better name.

fmunshi:
Hi @srkgupta I have fixed all of the mentioned issues: 

> MM-25248 - Discard changes not shown when changes made in Members block and user clicks somewhere else
- This is fixed now and the discard changes modal will show if any changes are made to either of the members block pages

> MM-25249 - Search in Members block is not cleared when we navigate to a different channel
- This is fixed (Fix applied to both team and channel members block)

> MM-25250 - Searching in Members section of Channels page causes the whole block to shift causing unpleasant viewing experience
- This is fixed by setting the max height of the container to be the same as the _previous_ results. So if it was showing 10 users before it will keep the height for 10 users and then shift to a smaller height _after_ the loading is completed this is based on what @michaelgamble has requested.

> MM-25251 - Total count not displayed in Pagination section of Members block in Channel Configuration page when user searches for a member
- This is resolved, total count is showing properly on search (and pagination will now apply to search results as requested by @michaelgamble )

> MM-25252- Search on Members block in Channel Configuration page always returns the newly added members
- This works now, search will filter newly added users too

srkgupta:
Hi @fmunshi 
Currently I see that the Members block in Team Configurations and Channel Configuration page does not contains the Group column. Is this not part of scope of 5.24 release and will be handled later (or) was this missed? 

FYI @michaelgamble

fmunshi:
The groups column has been omitted from the designs now @srkgupta since we are not showing group members immediately after clicking the add group to channel button that column no longer made sense to keep

PR #3594, opened 2019-09-04
[MM-17353] Wait for transition on RHS before rendering expensive sub components

devinbinnie:
#### Summary
When the `search_results` component contained a large number of results, rendering would lag the transition on tablet view, which is not ideal from a UX standpoint. This PR stops the results from being rendered until the transition completes, improving visual performance.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-17353

PR #5483, opened 2020-05-13
MM-24965: Re-evaluate transitions on RHS state change

agarciamontoro:
#### Summary

Back in September, https://github.com/mattermost/mattermost-webapp/pull/3594 introduced a change to fix a laggy transition on the RHS when rendering on tablet view. The change looked for transitions when mounting the component and, in case one was found, it waited for the transition to finish before loading the results, placing a temporary loading spinner in the meanwhile.

A couple of weeks ago, https://github.com/mattermost/mattermost-webapp/pull/5353 made a big rework on the RHS. One of the changes added a transition for easing-in the expand and collapse of the RHS:

```diff
diff --git a/sass/layout/_sidebar-right.scss b/sass/layout/_sidebar-right.scss
index 78fd87e3f..7911781c5 100644
--- a/sass/layout/_sidebar-right.scss
+++ b/sass/layout/_sidebar-right.scss
@@ -13,10 +13,6 @@
         height: calc(100% - #{$announcement-bar-height});
     }
 
+    &.move--left {
+        @include single-transition(width, 0.25s, ease-in);
+    }
+
     &.sidebar--right--expanded {
         @include box-shadow(0px 8px 24px alpha-color($black, .24)); 
```

This transition is not added to the component until after it is mounted, as the `.move--left` class is only included when the RHS is already open. This interferes with the original fix for laggy transitions, which accounted only for transitions that were there when the component was first mounted, and makes the `isOpened` prop to be always false, causing an infinite loading spinner when opening the search/recent mentions/flagged posts results.

In order to maintain the new expand/collapse transition and not to make the original open/close transition laggy again, this PR makes sure that we re-evaluate the transitions, calling `determineTransition`, when the RHS is opened. I also added a new e2e test to make sure the loading spinner is eventually gone.

One thing that I still don't understand is why this reproduces in Chrome and not in Firefox, but I guess it may be related to how both browsers handle the transitions.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24965

#### Screenshots
rhs


nevyangelova:
@jgilliam17 this fixes https://mattermost.atlassian.net/browse/MM-24965 and I will have to open a new one for the loss of focus :)

PR #5488, opened 2020-05-13
Revert "[MM-22171] Move `loadedPlugins` into redux-store (#5262)"

lieut-data:
#### Summary
This reverts commit 00d6a96ad5580af1eba9f23eb584d31838cc7a58. Seeing some issues with the webapp bundles disappearing from community-daily that has a high probability of relating to this change. Reverting to verify.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25132


lieut-data:
Skipping QA review since this is a clean revert of the recently merged PR.

ali-farooq0:
Disappearing? or needs to be refreshed. Had the same issue, I just had to refresh  the page and that fixed it. The issue is that the PR changed the store names and probably messes with the local stored state, and requires a refresh. 

Can you confirm if refreshing the pages fixes the issue @lieut-data?

agarciamontoro:
Refreshing does solve the issue on my end.

lieut-data:
@ali-farooq0, refreshing solves the problem, until they disappear again randomly.

ali-farooq0:
@lieut-data are you sure it's not when you switched over to the `staff`/`contributors` team? they both require a refresh.

You've had the random disappear happen more than 2 times?

lieut-data:
(Discussed offline: still reproducing after refreshes, so proceeding with revert.)

PR #5502, opened 2020-05-15
[MM-25165] Fix search bar disappearing

hahmadia:
#### Summary

The search bar disappears when you search for a non-existent team or channel on system console (since the data returned is 0). This PR aims to fix this problem.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25165

#### Screenshots

image

image


hahmadia:
**NOTE** Github makes it seem there was way more code changes than there actually was. I simply removed `{this.searchBar()}` from the `header()` function and added it to render with a styled div tag. Nothing else was changed.

hahmadia:
@srkgupta Good catch. I didn't see that. I have fixed and you can re-test. Thanks 😄

amyblais:
@ali-farooq0 Was this a regression on master or an old issue?

PR #5505, opened 2020-05-15
MM-24380 - Add ability to add components to the left and right of the RHS

ali-farooq0:
#### Summary
Expands the `registerRightHandSidebarComponent` plugin API to accept `icons` field that are used to show plugin icons in the RHS.

**Changes**:
* Fixed showing tooltips for `expand` `collapse` icons in the RHS.
* `SearchResultsHeader` accepts an `icons` prop that are shown in the RHS

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-24380

#### Related Pull Requests
Incident-Response-Plugin: https://github.com/mattermost/mattermost-plugin-incident-response/pull/102

#### Incident response usage
May-15-2020 14-26-33

#### Incident response dynamic buttons
May-15-2020 22-58-16


ali-farooq0:
Great questions @cpoile!

1. Nope, since plugins let webapp manage their icons, I don't think there is any benefit in giving plugins the ability to hide certain system icons in the RHS.
 
2. Absolutely, spend most of my time figuring out how to make it dynamic. I had a working solution in the end, turns out as soon as I made the icons redux aware everything blew up :(. So had to rework how we update icons in the RHS. Mind giving this another look? 

> I'm assuming that we don't want to put buttons to the right of the expand and close icons (for UX reasons), right?

Correcto. Confirmed it with @asaadmahmood.

plugins/registry.js
```diff
@@ -579,10 +587,29 @@ export default class PluginRegistry {
                 pluginId: this.id,
                 component,
                 title,
+                icons: icons.map((icon) => ({...icon, icon: resolveReactElement(icon.icon)})),
```
marianunez, line 590: why pass them here too if they are processed in the action below `RECEIVED_PLUGIN_RHS_ICONS`?

components/search_results_header/index.jsx
```diff
@@ -8,16 +8,24 @@ import {
...
+        isExpanded: getIsRhsExpanded(state),
+    };
+}
+
 function mapDispatchToProps(dispatch) {
     return {
         actions: bindActionCreators({
             closeRightHandSide,
             toggleRhsExpanded,
+            dispatch: (action) => (action()), // Used to dispatch arbitrary action
```
marianunez, line 26: Could we pass in the icons action already bound to dispatch from the plugin like we do with the channel header icons? Not a fan of passing a generic dispatch to the component and we can keep the convention of how plugins have been passing actions up to now.

ali-farooq0: Since we can have an arbitrary number of `icons`, i'll have to have a mapping of icon->icon's action, which is generated somewhere, which is not ideal either. Also this is not passing dispatch down to the component, this is passing a function that accepts an action that _will_ be dispatched in the redux layer. So not really exposing dispatch to the component. I just happen to name it that, I can name it something different like `dispatchAction`?

plugins/registry.js
ali-farooq0: The action below is for subsequent updates, I only return the action below, not invoke it.

I could do something like:

```
        store.dispatch({
            type: ActionTypes.RECEIVED_PLUGIN_COMPONENT,
            name: 'RightHandSidebarComponent',
            data: {
                id,
                pluginId: this.id,
                component,
                title,
                icons: [], // set empty array
            },
        });

        const setRHSPluginIcons = (newIcons = []) => {
            return (dispatch) => {
                dispatch({
                    type: ActionTypes.RECEIVED_PLUGIN_RHS_ICONS,
                    data: {
                        id,
                        icons: newIcons.map((icon) => ({...icon, icon: resolveReactElement(icon.icon)})),
                    },
                });
            };
        };

	store.dispatch(setRHSPluginsIcons(icons));
```

components/search_results_header/search_results_header.jsx
```diff
@@ -51,16 +67,38 @@ export default class SearchResultsHeader extends React.Component {
...
+                                        key={id}
+                                        delayShow={Constants.OVERLAY_TIME_DELAY}
+                                        placement='top'
+                                        overlay={<Tooltip id={id}>{icon.tooltip}</Tooltip>}
+                                    >
+                                        <button
+                                            id={id}
+                                            type='button'
+                                            className='sidebar--right__expand btn-icon'
+                                            onClick={() => this.handleIconClick(icon.action)}
```
marianunez, line 85: If the plugin authors set this action to dispatch like we do here: 
https://github.com/mattermost/mattermost-plugin-demo/blob/3ed574f17f5c67ff67843a1cef391adcf40f154f/webapp/src/plugin.jsx#L68

then here it can just be 
```suggestion
                                            onClick={() => this.icon.action()}
```
like they do for the icon in the channel header:

https://github.com/mattermost/mattermost-webapp/blob/c46b47e7e0f81c49226b9e5795a9972b1e4b43a4/plugins/channel_header_plug/channel_header_plug.jsx#L127 without the need to do a generic `mapDispatchToProps` for the action. 

I think we should follow this approach to be consistent on how the webapp api is receiving actions in the current registry functions.

ali-farooq0: Ahh good call. Updated to pass in just a function rather than a dispatchable action.

ali-farooq0: Updated both repos.

components/search_results_header/search_results_header.jsx
```diff
@@ -51,16 +62,38 @@ export default class SearchResultsHeader extends React.Component {
...
+                                        key={id}
+                                        delayShow={Constants.OVERLAY_TIME_DELAY}
+                                        placement='top'
+                                        overlay={<Tooltip id={id}>{icon.tooltip}</Tooltip>}
+                                    >
+                                        <button
+                                            id={id}
+                                            type='button'
+                                            className='sidebar--right__expand btn-icon'
+                                            onClick={() => icon.action()}
```
marianunez, line 80: Should we check here if `icon.action` is not null first?

ali-farooq0: I did mention it in the docs that the field is required, however It is better to have a null check around it. Done 👍

ali-farooq0:
Thanks @prapti!

Merging changes.

PR #5518, opened 2020-05-18
MM-24818: Fix leaving archived channels

agarciamontoro:
#### Summary
This PR fixes a bug that prevented the user to leave already archived channels.

The bug was first introduced with https://github.com/mattermost/mattermost-webapp/pull/4402, which refactored how we used `getState` in several functions. In one of those functions, `leaveChannel`, we were reading the state at the beginning of the function, never updating it even when we synchronously dispatched an action that changed it. This caused the next calls to behave incorrectly, as they were being fed a state that was no longer the current one.

The solution was to call `getState` again after the state was changed. I have also added an e2e test that checks that the issue is gone.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24818

#### Screenshots
fix


agarciamontoro:
Friendly ping, @nevyangelova :)

jgilliam17:
@agarciamontoro 
Seeing one issue
When leaving an archived channel, user is returned to the 2nd to the last viewed channel instead of to the last viewed channel.
for example:
1. View `test2` channel
2. View `test4` channel
3. Use `More` to open `test1` archived 
4. Leave `test1` archived channel 
Expected: Back to viewing test4
Actual: Back to test2

agarciamontoro:
@jgilliam17 I just tested your steps in master and I could also reproduce it. Although it may be related to this issue, I think it is worth a new ticket. What do you think about merging this one and I'll open a new ticket for the issue you described?

agarciamontoro:
Any objections agains cherry-picking this into 5.24, @amyblais?

PR #5530, opened 2020-05-18
[MM-23937] Add e2e tests for group mentions

fmunshi:
### Summary
- Adds e2e tests for group mentions.

- The following two assertions run for two different tests, once to enable and disable the allow reference switch and once to enable and disable the group mentions permission on a regular user while keeping the group allow reference enabled

#### To test whether a group mention is enabled the tests assert the following: 
- the suggestion list shows the given group
- posting the message as a user not in the group shows it as a blue link
- viewing the message as a user in the group shows it as highlighted 

#### To test that a group mention is disabled the tests assert the following: 
- the suggestion list does _not_ show the given group
- posting the message containing the group does not apply any custom link styling on it
- viewing the message as a user in the group shows it as regular text since the mention did not trigger notifications 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-23937


components/admin_console/group_settings/group_details/group_profile.jsx
```diff
@@ -44,8 +44,7 @@ export default class GroupProfile extends React.PureComponent {
                         </div>
                         <input
                             type='text'
-                            id={customID}
-                            className='form-control group_at_mention_input'
+                            className='form-control group-at-mention-input'
```
fmunshi, line 47: this change is unrelated to the test but it just made sense to change the class to not be snake case since we always use hyphen case for class names

components/admin_console/group_settings/group_details/group_profile_and_settings.jsx
```diff
@@ -76,7 +77,7 @@ export const GroupProfileAndSettings = ({displayname, mentionname, allowReferenc
                 name={mentionname}
                 title={t('admin.group_settings.group_details.group_mention.name')}
                 titleDefault={'Group Mention:'}
-                customID={'GroupMention'}
+                customID={'group_mention'}
```
fmunshi, line 80: these changes are also seemingly unrelated but again it made sense to set these ids as snake case since the rest of the ids in this file also use the same format and these two components were the outliers

e2e/cypress/integration/enterprise/ldap/ldap_login_spec.js
```diff
@@ -24,7 +24,7 @@ function setLDAPTestSettings(config) {
 
 // assumes the CYPRESS_* variables are set
 // assumes that E20 license is uploaded
-// for setup with AWS: Follow the instructions mentioned in the mattermost/platform-private/config/ldap-test-setup.txt file
+// for setup with AWS: Follow the instructions mentioned in the mattermost/enterprise/config/ldap-test-setup.txt file
```
fmunshi, line 27: another unrelated change but `platform-private` is no longer a thing

components/admin_console/group_settings/group_details/group_profile.jsx
```diff
@@ -44,8 +44,7 @@ export default class GroupProfile extends React.PureComponent {
                         </div>
                         <input
                             type='text'
-                            id={customID}
```
fmunshi, line 47: also this id is being applied at the root of this component and was also being applied to the input -- ids should be unique so I removed it here since it wasnt required

e2e/cypress/integration/enterprise/ldap_group/group_mentions_spec.js
```diff
@@ -0,0 +1,263 @@
...
+        });
+    });
+
+    it('MM-23937 - Can enable and disable group mentions with a custom name for a group ', () => {
+        // # Login as sysadmin and navigate to board group page
+        navigateToGroup(groupID);
+
+        // # Click the allow reference button
+        cy.findByTestId('allow_reference_switch').then((el) => {
+            el.find('button').click();
```
saturninoabril, line 188: Click action to button works but it resets to original state (not active) when click action happened before the list of Users are loaded. My suggestion is to check Users first, something like:
```diff
+cy.findByText(boardUser.email).should('be.visible');
cy.findByTestId('allow_reference_switch').then((el) => {
    el.find('button').click();
});
```

Same case when disabling Group mentions.

e2e/cypress/support/api_commands.js
```diff
@@ -1279,3 +1279,42 @@ Cypress.Commands.add('apiActivateUser', (userId, active = true) => {
     const baseUrl = Cypress.config('baseUrl');
     cy.externalRequest({user: users.sysadmin, method: 'put', baseUrl, path: `users/${userId}/active`, data: {active}});
 });
+
+/**
+ * Get all groups via the API
+ *
+ * @param {Integer} page - The desired page of the paginated list
+ * @param {Integer} per_page - The number of groups per page
```
saturninoabril, line 1287:
```suggestion
 * @param {Integer} perPage - The number of groups per page
```

e2e/cypress/support/api_commands.js
```diff
@@ -1279,3 +1279,42 @@ Cypress.Commands.add('apiActivateUser', (userId, active = true) => {
...
+        expect(response.status).to.equal(200);
+        return cy.wrap(response);
+    });
+});
+
+/**
+ * Patch a group directly via API
+ *
+ * @param {String} name - The new name for the group
+ * @param {Boolean} allow_reference - Whether to allow reference (group mention) or not  - true/false
```
saturninoabril, line 1306: This does not match with the arguments `(groupID, patch)`

e2e/cypress/integration/enterprise/ldap_group/group_mentions_spec.js
```diff
@@ -0,0 +1,263 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Stage: @prod
```
saturninoabril, line 10: Please remove stage. We'll promote this spec to `@prod` when we found it to be stable for a few runs on CI.

e2e/cypress/integration/enterprise/ldap/ldap_login_spec.js
fmunshi: you are right - I was mistaken for some reason the link failed to load for me and I assumed it was the wrong repo

e2e/cypress/integration/enterprise/ldap_group/group_mentions_spec.js
```diff
@@ -0,0 +1,263 @@
...
+
+                    // # Set allow reference false to ensure correct data for test cases
+                    cy.apiPatchGroup(groupID, {allow_reference: false});
+                }
+            });
+        });
+
+        // # Login once as board user
+        boardUser = users['board-1'];
+        cy.apiLogin(boardUser.username, boardUser.password);
```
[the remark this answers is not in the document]
fmunshi: In order to ensure the user is created - since the user exists in ldap they need to login once before they are created

e2e/cypress/integration/enterprise/ldap_group/group_mentions_spec.js
```diff
@@ -0,0 +1,278 @@
...
+const navigateToGroup = (id) => {
+    // # Login as sysadmin and visit board group page, and wait until board user is visible
+    cy.apiLogin('sysadmin');
+    cy.visit(`/admin_console/user_management/groups/${id}`);
+
+    // # Scroll users list into view and then make sure it has loaded before scrolling back to the top
+    cy.get('#group_users').scrollIntoView();
+    cy.findByText(boardUser.email).should('be.visible');
+    cy.get('#group_profile').scrollIntoView();
+};
```
fmunshi, line 28: @saturninoabril I waited for the boardUser to be visible here instead of doing it in multiple places before clicking the switch

e2e/cypress/integration/enterprise/ldap_group/group_mentions_spec.js
```diff
@@ -0,0 +1,278 @@
...
+        assertGroupMentionDisabled(groupName);
+
+        // # Login as sysadmin and navigate to system scheme page
+        cy.apiLogin('sysadmin');
+        cy.visit('/admin_console/user_management/permissions/system_scheme');
+
+        // # Click reset to defaults confirm and save
+        cy.findByTestId('resetPermissionsToDefault').click();
+        cy.get('#confirmModalButton').click();
+        saveConfig();
```
srkgupta, line 276: Can you please add this in the after block (or) afterEach block instead of adding it at the end of the test. This is because the specified steps to reset the System permissions may get skipped if the test fails in between. However if its included in the after block, the steps will definitely be executed.

fmunshi: done

PR #5532, opened 2020-05-19
Adding request trial interface

jespino:
#### Summary
Adding request trial interface

#### Ticket Link
[MM-25341](https://mattermost.atlassian.net/browse/MM-25341)

#### Related Pull Requests
- Has server changes mattermost/mattermost-server#14603

#### Screenshots
image
image

image


i18n/en.json
```diff
@@ -940,6 +940,9 @@
   "admin.license.noFile": "No file uploaded",
   "admin.license.removing": "Removing License...",
   "admin.license.title": "Edition and License",
+  "admin.license.trial-request.different-data": "Or get a trial license with different data here: ",
```
jasonblais, line 943: Are we keeping the mattermost.com/trial link if requesting the trial license automatically fails? I'm just wondering if we'd remove it altogether.

Else, I'd tweak the text to:

```suggestion
  "admin.license.trial-request.different-data": "Or get a trial license manually at ",
```

since it may not be obvious to the admin if they should use different data.

jespino: Yes, on error the error text is shown below, the rest of the text is keep as is

jespino: Changed the text proposed.

PR #5540, opened 2020-05-20
MM-25363 Separate logic between running Cypress specs and saving/publishing test reports

saturninoabril:
#### Summary
Separate logic between running Cypress specs (`run_tests.js`) and saving/publishing test reports (`save_report.js`).  This is a step towards parallel testing in Cypress. Typical use would be running multiple test groups against its own test server, and then at the end, all reports will be processed all at once. Something like:
```bash
# Test run 1: against server 1
node run_tests.js --stage='@prod' --group='@messaging'

# Test run 2: against server 2
node run_tests.js --stage='@prod' --group='@enterprise'

# Test run 3: against server 3
node run_tests.js --stage='@prod' --group='@messaging,@enterprise' --invert

# Consolidate, save and publish all test data from all Test runs
node save_report.js
```

#### Ticket Link
Jira ticket - https://mattermost.atlassian.net/browse/MM-25363


saturninoabril:
@srkgupta Thanks for checking and yes, that makes sense.  I'll update the PR.

saturninoabril:
@srkgupta Updated, PTAL.

prapti:
Some checks aren't passing, couldn't merge @saturninoabril

saturninoabril:
Thanks Prapti. It was due to recent update of ESLint rule on comma-dangle. Now fixed.  Also I added small update on S3 folder - replacing dot with dash. Will merge after passing the build.

PR #5544, opened 2020-05-20
MM-22155 ESLint configuration updates part 1

hmhealey:
This is the first part of the configuration changes proposed [here](https://docs.google.com/document/d/1pcOyNxJrlnfs7FKS20aVzGnd08gi1krosx63HTtuls0/edit). It's focused on the simpler things to change like allowing `undefined` since it's better supported in TypeScript, getting rid of a few annoying warnings (file extensions and magic numbers), and re-enabling some stuff that was disabled during the TypeScript migration (camelcase naming, nested ternaries, and unused variables).

The biggest change in this one is that re-enabling the `no-unused-vars` check requires a bunch of redux action creators to have their return type changed, although this is consistent with what we had in the past.

Most of the rules changes in the `.eslintrc.json` will be moved to our ESLint plugin later once all the repos are consistent.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22155

#### Related Pull Requests
https://github.com/mattermost/mattermost-mobile/pull/4311
https://github.com/mattermost/mattermost-redux/pull/1158


utils/markdown/renderer.tsx
```diff
@@ -275,7 +275,7 @@ export default class Renderer extends marked.Renderer {
         return output;
     }
 
-    public listitem(text: string, bullet = '') {
+    public listitem(text: string, bullet = '') { // eslint-disable-line @typescript-eslint/no-unused-vars
```
devinbinnie, line 278: Are we removing this `eslint-disable-line` in a future PR? Or is this a special case?

utils/paste.test.tsx
```diff
@@ -68,7 +68,7 @@ describe('Paste.formatGithubCodePaste', () => {
         const message = "```\n// a javascript codeblock example\nif (1 > 0) {\n  return 'condition is true';\n}\n```";
         const codeBlock = "```\n// a javascript codeblock example\nif (1 > 0) {\n  return 'condition is true';\n}\n```";
 
-        const {formattedMessage, formattedCodeBlock} = formatGithubCodePaste(0, '', clipboardData);
+        const {formattedMessage, formattedCodeBlock} = formatGithubCodePaste(0, '', clipboardData); // eslint-disable-line @typescript-eslint/no-unused-vars
```
devinbinnie, line 71: Same with these.

utils/markdown/renderer.tsx
hmhealey: This is a special case. The type definitions provided for `marked` don't match what we use (either because we're based off an old version of marked or because of our fork, I can't remember), and some of the overridden Renderers need that extra parameter.

utils/paste.test.tsx
hmhealey: I filed a ticket to fix these. The tests weren't written correctly, so I added the ignore comments so that the argument handling doesn't need to be re-added when the tests are fixed.

hmhealey: Once https://github.com/mattermost/mattermost-webapp/pull/5559 is merged, these can be removed. I'll do that in this PR if that one makes it in first.

hmhealey:
Skipping dev review since these are only code style changes

PR #5545, opened 2020-05-20
Convert 'components/user_settings/general' module and associated tests to TypeScript

hiendinhngoc:
#### Summary
Migrate 'components/user_settings/general' module and associated tests to TypeScript

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/13692
  JIRA: https://mattermost.atlassian.net/browse/MM-20549


hiendinhngoc:
I get stuck with this test:
`wrapper.instance().submitUser(requiredProps.user, '');`
it shows the error: `Property 'submitUser' does not exist on type 'Component<{}, {}, any>'.`
it didn't fix even I tried with `(wrapper.instance() as UserSettingsGeneral).submitUser(requiredProps.user, '');`
Any suggestion with this?

devinbinnie:
> I get stuck with this test:
> `wrapper.instance().submitUser(requiredProps.user, '');`
> it shows the error:
> `Property 'submitUser' does not exist on type 'Component<{}, {}, any>'.`
> it didn't fix even I tried with
> `(wrapper.instance() as UserSettingsGeneral).submitUser(requiredProps.user, '');`
> Any suggestion with this?

@hiendinhngoc You'll want to type the call to `shallowWithIntl` or `mountWithIntl`, it should look like this:
`const wrapper = shallowWithIntl<ComponentType>(<Component {...props}/>)`

You also in this case might need to export the type seperately, since the only exported member is an wrapper context object, which does not have the type of the class you need.

hiendinhngoc:
> @hiendinhngoc You'll want to type the call to `shallowWithIntl` or `mountWithIntl`, it should look like this:
> `const wrapper = shallowWithIntl<ComponentType>(<Component {...props}/>)`
> 
> You also in this case might need to export the type seperately, since the only exported member is an wrapper context object, which does not have the type of the class you need.

@devinbinnie thanks, I guess the reason I got the issue is due to this:
`export default injectIntl(UserSettingsGeneralTab);`
in `user_settings_general.tsx`, but I'm not sure why it was used there. If I comment on this and use  
the normal export:
 `export default class UserSettingsGeneralTab extends React.Component<Props, State>` 
then the test will pass.

devinbinnie:
> > @hiendinhngoc You'll want to type the call to `shallowWithIntl` or `mountWithIntl`, it should look like this:
> > `const wrapper = shallowWithIntl<ComponentType>(<Component {...props}/>)`
> > You also in this case might need to export the type seperately, since the only exported member is an wrapper context object, which does not have the type of the class you need.
> 
> @devinbinnie thanks, I guess the reason I got the issue is due to this:
> `export default injectIntl(UserSettingsGeneralTab);`
> in `user_settings_general.tsx`, but I'm not sure why it was used there. If I comment on this and use
> the normal export:
> `export default class UserSettingsGeneralTab extends React.Component<Props, State>`
> then the test will pass.

We can't use the normal export here, as the component requires the `intl` prop to be passed in using `injectIntl`.

What you can do is just export the class seperately, not as the default, so that you can use it in your type:
`export default class UserSettingsGeneralTab extends React.Component<Props, State>`
and in your test:
```
import {UserSettingsGeneralTab as UserSettingGeneralTabType} from 'path/to/component'
...
const wrapper = shallowWithIntl<UserSettingsGeneralTabType>(<UserSettingsGeneralTab/>);
```

components/user_settings/general/user_settings_general.tsx
```diff
@@ -85,45 +83,93 @@ const holders = defineMessages({
...
-        samlLastNameAttributeSet: PropTypes.bool,
-        ldapNicknameAttributeSet: PropTypes.bool,
-        samlNicknameAttributeSet: PropTypes.bool,
-        ldapPositionAttributeSet: PropTypes.bool,
-        samlPositionAttributeSet: PropTypes.bool,
-        ldapPictureAttributeSet: PropTypes.bool,
-    }
+type Props = {
+    intl: IntlShape;
+    user: {
```
devinbinnie, line 88: We have a type in `mattermost-redux` called `UserProfile` that you can use here.

components/user_settings/general/user_settings_general.tsx
```diff
@@ -255,7 +301,7 @@ class UserSettingsGeneralTab extends React.PureComponent {
         this.submitUser(user, true);
     }
 
-    submitUser = (user, emailUpdated) => {
+    submitUser = (user: object, emailUpdated: boolean | string) => {
```
devinbinnie, line 304: It looks like this is only be used as a `boolean`, if there's a call to this method passing a `string`, it should be cast to a `boolean`.

components/user_settings/general/user_settings_general.tsx
```diff
@@ -356,39 +402,39 @@ class UserSettingsGeneralTab extends React.PureComponent {
...
         this.setState({confirmEmail: e.target.value});
     }
 
-    updateCurrentPassword = (e) => {
+    updateCurrentPassword = (e: React.ChangeEvent<HTMLInputElement>) => {
         this.setState({currentPassword: e.target.value});
     }
 
-    updatePicture = (e) => {
+    updatePicture = (e: {target: {files: {type: string; size: number}[]}}) => {
```
devinbinnie, line 437: this can use the `HTMLInputElement` type I believe.

components/user_settings/general/user_settings_general.tsx
```diff
@@ -650,7 +697,7 @@ class UserSettingsGeneralTab extends React.PureComponent {
                 />
             );
         } else {
-            let describe = '';
+            let describe: any = '';
```
devinbinnie, line 700: Should be a `string`, how come it's being cast as an `any`?

devinbinnie: Same with all the ones below.

components/user_settings/user_settings.tsx
```diff
@@ -11,11 +11,21 @@ import SecurityTab from './security';
 import SidebarTab from './sidebar';
 
 type Props = {
-    user: {};
+    user: {
```
devinbinnie, line 14: This can also use `UserProfile`

package-lock.json
```diff
@@ -10836,7 +10836,10 @@
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/full-icu/-/full-icu-1.3.0.tgz",
       "integrity": "sha512-LGLpSsbkHUT0T+EKrIJltYoejYzUqg1eW+n6wm/FTte1pDiYjeKTxO0uJvrE3jgv6V9eBzMAjF6A8jH16C0+eQ==",
-      "dev": true
+      "dev": true,
```
devinbinnie, line 10839: Can we revert these `package-lock.json` changes? They aren't necessary for this PR.

components/user_settings/general/user_settings_general.tsx
devinbinnie: @hiendinhngoc This is merged now, sorry about the delay :)

components/user_settings/general/user_settings_general.tsx
devinbinnie: It looks like we're allowing for either an `Element` or a `string` to be rendered.
You can define the type as `Element | string`.

components/user_settings/general/user_settings_general.tsx
hiendinhngoc: @devinbinnie sorry, but still got error `Property 'password' does not exist on type 'UserProfile'` even the password was added to the library. What should I do to get the updated from `mattermost-redux`?

devinbinnie: You'll have to update the commit has in `package.json` and `package-lock.json` to the most recent commit on `master` from `mattermost-redux` and then run `npm i` on the webapp to get the updated package.

jfrerich:
@hiendinhngoc, thanks for the PR.  I cloned locally and see the resulting errors from running `make check-types`.  Do you need any assistance?

devinbinnie:
> > @hiendinhngoc, thanks for the PR. I cloned locally and see the resulting errors from running `make check-types`. Do you need any assistance?
> 
> @jfrerich
> Thanks, currently I'm not sure how to resolve `incompatible type` user in `user_settings_general.test.tsx`. Do you have any suggestion?
> **Update**
> I fixed user type issue, but still got error with:
> `await (wrapper.instance() as UserSettingsGeneralTab).submitUser(requiredProps.user, '');`
> error: `Argument of type '""' is not assignable to parameter of type 'boolean'.`
> and
> `(wrapper.instance() as UserSettingsGeneralTab).updatePicture(event);`
> error: `Argument of type '{ target: { files: { type: string; size: number; }[]; }; }' is not assignable to parameter of type 'ChangeEvent<HTMLInputElement>'.`

@hiendinhngoc `submitUser` expects a `boolean` for the second argument, so I would just change those blank strings to `false`. That would be equivalent.

For `updatePicture` you can cast `event` as `any` for this case, since it's not necessary to mock out an entire `ChangeEvent`.

components/user_settings/general/index.ts
```diff
@@ -13,13 +13,42 @@ import {
...
+import {ActionFunc} from 'mattermost-redux/types/actions';
 
-function mapStateToProps(state) {
+import UserSettingsGeneralTab from './user_settings_general';
+
+type Actions = {
+    logError: ({message, type}: {message: any; type: string}, status: boolean) => void;
+    clearErrors: () => void;
+    getMe: () => void;
+    updateMe: (user: object) => Promise<{
```
devinbinnie, line 25: This type should be `UserProfile`

components/user_settings/general/user_settings_general.tsx
```diff
@@ -856,7 +893,7 @@ class UserSettingsGeneralTab extends React.PureComponent {
...
```
devinbinnie, line 896: This and all others like it can actually be of type `React.ReactNode`. It includes all of the possible types that can be rendered to the DOM. Sorry about the confusion.

components/user_settings/general/user_settings_general.tsx
```diff
@@ -1216,7 +1253,7 @@ class UserSettingsGeneralTab extends React.PureComponent {
                     defaultImageSrc={Utils.defaultImageURLForUser(user.id)}
                     serverError={serverError}
                     clientError={clientError}
-                    updateSection={(e) => {
+                    updateSection={(e: any) => {
```
devinbinnie, line 1256: Can this be some event type instead of `any`?

hanzei:
@jfrerich @devinbinnie Gentle reminder to review this PR. I would love if @hiendinhngoc could get some early feedback on the work.

devinbinnie:
@hiendinhngoc Can you resolve the conflicts on this PR? It looks like there are still changes to `package-lock.json` that don't need to be there. Let me know if you need any assistance.

hiendinhngoc:
> @hiendinhngoc Can you resolve the conflicts on this PR? It looks like there are still changes to `package-lock.json` that don't need to be there. Let me know if you need any assistance.

@devinbinnie  I got error when reverting `package-lock.json` file:
```
cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

devinbinnie:
@hiendinhngoc I'm still seeing changed to `package-lock.json` here: https://github.com/mattermost/mattermost-webapp/pull/5545/files#diff-32607347f8126e6534ebc7ebaec4853d

You might want to update your local branch to master and make sure to just take all changes from `master` for that file and then push that up. If it still doesn't build, let me know.

components/user_settings/general/user_settings_general.tsx
```diff
@@ -997,7 +1034,7 @@ class UserSettingsGeneralTab extends React.PureComponent {
             let extraInfo;
             let submit = null;
             if (this.props.user.auth_service === '') {
-                let usernameLabel = (
+                let usernameLabel: any = (
```
devinbinnie, line 1037: Can this also be `JSX.Element|string`?

components/user_settings/general/user_settings_general.tsx
```diff
@@ -1088,7 +1125,7 @@ class UserSettingsGeneralTab extends React.PureComponent {
                     </span>
                 );
             } else {
-                let positionLabel = (
+                let positionLabel: any = (
```
devinbinnie, line 1128: Can this also be `JSX.Element|string` ?

PR #5547, opened 2020-05-20
[MM-24917] Prefecthing of posts

sudheerDev:
## Summary
On webapp we fetch posts when user clicks on a channel. The aim of the PR is to prefetch posts without user intervention

**Conditions for prefetching posts**:
1. On load of webapp
2. On socket reconnect or system comes from sleep
3. On new message in a channel where user has not visited in the present session
4. On addition of user to a channel/GM
5. On Team switch

**Priority order**:
Channels with mentions are loaded first followed by unread channels.

**Prefetch process**:
In order not to impact the performance of clients and leave breathing space on network stack for user-initiated requests, we are capping the concurrency of requests made to 2 at a time. New prefetch requests are dispatched when previous requests are processed.

**Code Changes**:
* PR looks for changes in LHS and responds by making requests for channels that don't have a request dispatched. 
* Adds current channel posts requests to be dispatched as soon as it is set in redux state instead of dispatching it from actions down the hierarchy. Otherwise couple of prefetching requests are sent before the postlist makes a request for posts.
* Skipping making a duplicate request if prefetch requests are in progress when switching to a channel.
* Add a jitter(0-1sec) for delaying post requests in case of a new message in open/private channels. This is to prevent a case when all clients request messages when new post is made in a channel with thousands of users. 

## Tickets
https://mattermost.atlassian.net/browse/MM-24917
https://mattermost.atlassian.net/browse/MM-24920
https://mattermost.atlassian.net/browse/MM-24921
https://mattermost.atlassian.net/browse/MM-24924
https://mattermost.atlassian.net/browse/MM-24925


cpanato:
terminating this test server because we need to clean the cluster that is running. If you still need a test server please add the label again in the future

actions/user_actions.jsx
```diff
@@ -284,9 +284,9 @@ export function loadProfilesForGroupChannels(groupChannels) {
     };
 }
 
-export function loadProfilesForSidebar() {
-    loadProfilesForDM();
-    loadProfilesForGM();
+export async function loadProfilesForSidebar() {
```
nevyangelova, line 287: Is there a reason to not use Promise.all here? If I understand correctly this doesn't needs to run sequentially.

client/websocket_client.jsx
```diff
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 const MAX_WEBSOCKET_FAILS = 7;
-const MIN_WEBSOCKET_RETRY_TIME = 3000; // 3 sec
+const MIN_WEBSOCKET_RETRY_TIME = 30000; // 3 sec
```
nevyangelova, line 5: Nit: 30 sec? :P

components/data_prefetch/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import {prefetchChannelPosts} from 'actions/views/channel';
+
+import DataPrefetch from './data_prefetch';
+
+type Actions = {
+    prefetchChannelPosts: (channelId: string, delay: number | undefined) => Promise<{data: {}}>;
+};
+
+const prefetchQueue = memoizeResult((channels: Channel[], memberships: RelationOneToOne<Channel, ChannelMembership>) => {
```
nevyangelova, line 27: 👏

components/data_prefetch/index.ts
```diff
@@ -0,0 +1,75 @@
...
+            } else if (membership.notify_props && membership.notify_props.mark_unread !== 'mention' && channel.total_msg_count - membership.msg_count) {
+                return {
+                    ...acc,
+                    2: [...acc[2], channelId],
+                };
+            }
+        }
+        return acc;
+    }, {
+        1: [], // 1 being high priority requests
```
nevyangelova, line 46: Maybe this should be a priorityEnum type.

actions/user_actions.jsx
sudheerDev: Nope. Good call, I changed it to `Promise.all`

client/websocket_client.jsx
sudheerDev: Oops, It was a value left from testing socket disconnect scenario with PM. Removed it

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+    currentChannelId: string;
+    prefetchQueueObj: Record<string, string[]>;
+    prefetchRequestStatus: Record<string, any>;
+    unreadChannels: Channel[];
+    actions: {
+        prefetchChannelPosts: (channelId: string, delay: number | undefined) => Promise<any>;
+    };
+}
+
+export default class DataPrefetch extends React.PureComponent<Props, {}> {
```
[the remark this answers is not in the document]
sudheerDev: I don't feel this is a good example of usage for functional components or hooks. This is not a reusable component or needs any abstraction for the state. Both of them work great when we have multiple of these component instances in the app like dot menu etc where the benefit of a smaller footprint component is magnified because we can have thousands of them at the cost of small optimization for react rendering. 

Here, in this case, I don't think it fits the bill in either purpose

components/data_prefetch/index.ts
sudheerDev: Would it work? This is not an auto-incremented field. I am ignorant about this and I couldn't find any example using either.

nevyangelova: Enums values are by default numeric, I guess the benefit I had in mind is that you can express this in the value and not with comments. I am 1/5 on this.

components/data_prefetch/data_prefetch.tsx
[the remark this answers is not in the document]
sudheerDev: I cross-checked this again but it does not help much in this case. we could however do this for other components in the future mainly when we update redux to 7.1v to use `useSelector` or `useDispatch` etc otherwise here we will end up with one component for connecting to redux and then another hook component making this more complex than it needs to be.

hmhealey, line 23: Can you add some explanation of how this component works, perhaps as a comment so it's easy to find in the future?

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import PQueue from 'p-queue';
+import {Channel} from 'mattermost-redux/types/channels';
+
+import {Constants} from 'utils/constants';
+import {loadProfilesForSidebar} from 'actions/user_actions.jsx';
+
+const queue = new PQueue({concurrency: 2});
```
hmhealey, line 11: Shouldn't `queue` be part of the `DataPrefetch` component? Having it be global feels weird.

Also, why do we have two different queues here, one in this file and one coming from redux?

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+    };
+}
+
+export default class DataPrefetch extends React.PureComponent<Props, {}> {
+    private prefetchTimeout?: number;
+
+    async componentDidUpdate(prevProps: Props) {
+        const {currentChannelId, prefetchQueueObj} = this.props;
+        if (!prevProps.currentChannelId && currentChannelId) {
+            queue.add(async () => this.prefetchPosts(currentChannelId));
```
hmhealey, line 29: What if `currentChannelId` changes before this is called?

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+            await loadProfilesForSidebar();
+            this.prefetchData();
+        } else if (prevProps.prefetchQueueObj !== prefetchQueueObj) {
+            clearTimeout(this.prefetchTimeout);
+            queue.clear();
+            this.prefetchTimeout = window.setTimeout(() => {
```
hmhealey, line 35: Why use a `setTimeout` here? It seems like we could start prefetching immediately.

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    public componentWillUnmount(): void {
+        clearTimeout(this.prefetchTimeout);
+    }
+
+    public prefetchPosts = (channelId: string) => {
+        let delay;
+        const channel = this.props.unreadChannels.find((unreadChannel) => channelId === unreadChannel.id);
```
hmhealey, line 47: This looks like it's trying to fetch posts for the current channel when this is called from `componentDidUpdate`. Is that correct? Why would we need to do that when the PostList should already be doing that.

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+        if (channel && (channel.type === Constants.PRIVATE_CHANNEL || channel.type === Constants.OPEN_CHANNEL)) {
```
hmhealey, line 48: Why only add a delay for public/private channels?

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+            const isLatestPostInLastMin = (Date.now() - channel.last_post_at) <= 1000;
+            if (isLatestPostInLastMin) {
+                delay = Math.random() * 1000; // 1ms - 1000ms random wait to not choke server
+            }
+        }
+        return this.props.actions.prefetchChannelPosts(channelId, delay);
+    }
+
+    private shouldLoadPriorityQueue = (priority: number, numberOfPiorityRequests: Record<string, number>) => {
```
hmhealey, line 57: Can you add a comment to explain what's happening here?

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const queue = new PQueue({concurrency: 2});
+
+type Props = {
+    currentChannelId: string;
+    prefetchQueueObj: Record<string, string[]>;
+    prefetchRequestStatus: Record<string, any>;
+    unreadChannels: Channel[];
+    actions: {
+        prefetchChannelPosts: (channelId: string, delay: number | undefined) => Promise<any>;
```
hmhealey, line 19: I think this can be specified as
```suggestion
        prefetchChannelPosts: (channelId: string, delay?: number) => Promise<any>;
```
since delay is optional.

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import {Constants} from 'utils/constants';
+import {loadProfilesForSidebar} from 'actions/user_actions.jsx';
+
+const queue = new PQueue({concurrency: 2});
+
+type Props = {
+    currentChannelId: string;
+    prefetchQueueObj: Record<string, string[]>;
+    prefetchRequestStatus: Record<string, any>;
```
hmhealey, line 16: Can you add a type for the status of a request? Also, I know it doesn't change much, but is this keyed off of channel ID? I'm having a lot of trouble following this code.

components/data_prefetch/data_prefetch.tsx
hmhealey: Actually, I don't understand what this queue even does. I see that stuff is added to it in this file, but I don't see what's being added actually get used for anything.

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+        for (let priorityOrder = 1; priorityOrder <= 3; priorityOrder++) {
+            const priorityQueue = this.props.prefetchQueueObj[priorityOrder];
+            for (let channelIndex = 0; channelIndex < priorityQueue.length; channelIndex++) {
+                if (!this.props.prefetchRequestStatus[priorityQueue[channelIndex]] && this.shouldLoadPriorityQueue(priorityOrder, numberOfPiorityRequests)) {
+                    const channelId = priorityQueue[channelIndex];
+                    numberOfPiorityRequests[priorityOrder]++;
+                    queue.add(async () => this.prefetchPosts(channelId));
+                }
+            }
+            numberOfPiorityRequests[priorityOrder] = 0;
```
hmhealey, line 82: Why keep track of the number of requests outside of the loop if we're clearing it between loop iterations?

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private prefetchData = () => {
+        const numberOfPiorityRequests: Record<string, number> = {
+            3: 0,
+            2: 0,
+            1: 0, // high priority requests
+        };
+
+        for (let priorityOrder = 1; priorityOrder <= 3; priorityOrder++) {
```
hmhealey, line 73: Can you explain what this loop does? I'm having a really hard time understanding it.

It might make it easier to follow if the for loops used `for in` instead of looping through indices as well.

components/data_prefetch/index.ts
hmhealey: Yeah, you could define a TypeScript enum for this like
```typescript
enum Priority {
    High: 1,
    Medium: 2,
    Low: 3,
}
```

and then refer to them by name. It also helps since you can use `Priority` as a type elsewhere instead of our JS-based enums where you have to define a bunch of constants for them.

components/data_prefetch/index.ts
hmhealey, line 27: Can you add a comment explaining what this does?

Are we concerned if this returns a new object that's equal to the old one? Like renaming a channel or changing some settings will cause this to return a new, identical object.

components/legacy_sidebar/sidebar.test.jsx
```diff
@@ -216,6 +213,18 @@ describe('component/legacy_sidebar/sidebar_channel/SidebarChannel', () => {
...
+    test('should match snapshot, when isDataPrefechEnabled is false', () => {
+        const wrapper = shallowWithIntl(
+            <Sidebar
+                {...{
+                    ...defaultProps,
+                    isDataPrefechEnabled: false,
+                }}
+            />,
+        );
+        expect(wrapper).toMatchSnapshot();
```
hmhealey, line 225: What are you wanting to test here? Could we be more specific than just looking to see if we match a snapshot?

components/legacy_sidebar/sidebar.jsx
```diff
@@ -717,16 +735,17 @@ class LegacySidebar extends React.PureComponent {
...
         return (
             <div
-                className={classNames('sidebar--left', {'move--right': this.props.isOpen && Utils.isMobile()})}
+                className={classNames('sidebar--left', {'move--right': isOpen && Utils.isMobile()})}
                 id='sidebar-left'
                 key='sidebar-left'
                 role='navigation'
                 aria-labelledby='sidebar-left'
             >
+                {isDataPrefechEnabled && <DataPrefetch/>}
```
hmhealey, line 744: To avoid duplicating this between the old and new sidebars, could this go in the `ChannelController` that contains both of them? There's other invisible logic components there like the `FaviconTitleHandler` there as well

components/sidebar/sidebar.tsx
```diff
@@ -122,11 +124,19 @@ export default class Sidebar extends React.PureComponent<Props, State> {
     }
 
     render() {
+        const {
```
hmhealey, line 127: Can you also revert these changes? We have a lot of upcoming changes in the sidebar feature branch, so I'd like to avoid conflicts when possible.

components/post_view/post_list/index.js
```diff
@@ -43,15 +44,18 @@ function makeMapStateToProps() {
         let atOldestPost = false;
         let formattedPostIds;
         let latestAriaLabelFunc;
-        const lastViewedAt = state.views.channel.lastChannelViewTime[ownProps.channelId];
-        const channelManuallyUnread = isManuallyUnread(state, ownProps.channelId);
+        const {focusedPostId, unreadChunkTimeStamp, channelId} = ownProps;
```
hmhealey, line 47: 0/5, but it might be better to avoid these sort of changes when you're not directly touching most of this code since it makes the diff messier. I'm not saying it needs to be changed back, but when not writing new code, I think it's a bit better to stick with the existing coding style (assuming it isn't absolutely awful, of course 😛)

components/post_view/post_list/post_list.jsx
```diff
@@ -159,19 +164,23 @@ export default class PostList extends React.PureComponent {
...
-            await this.props.actions.loadPostsAround(channelId, this.props.focusedPostId);
-        } else if (this.props.isFirstLoad) {
-            await this.props.actions.loadUnreads(channelId);
-        } else if (this.props.latestPostTimeStamp) {
-            await this.props.actions.syncPostsInChannel(channelId, this.props.latestPostTimeStamp, false);
+        const {focusedPostId, isFirstLoad, latestPostTimeStamp, isPrefetchingInProcess, actions} = this.props;
+        if (focusedPostId) {
+            await actions.loadPostsAround(channelId, this.props.focusedPostId);
+        } else if (isFirstLoad) {
+            if (!isPrefetchingInProcess) {
```
hmhealey, line 171: Perhaps it would be better to still fire this request, regardless of prefetching, just in case the prefetching status gets stuck as in-progress somehow? It's a bit wasteful to make two requests for the same data if the user is switching channels while prefetching is happening, but I don't think it would cause us any real issues. 0/5 though.

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+        prefetchChannelPosts: (channelId: string, delay: number | undefined) => Promise<any>;
+    };
+}
+
+export default class DataPrefetch extends React.PureComponent<Props, {}> {
+    private prefetchTimeout?: number;
+
+    async componentDidUpdate(prevProps: Props) {
+        const {currentChannelId, prefetchQueueObj} = this.props;
+        if (!prevProps.currentChannelId && currentChannelId) {
```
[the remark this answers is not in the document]
sudheerDev: This happens only the first time when page loads. The current channel is set after the LHS is loaded.

components/data_prefetch/data_prefetch.tsx
sudheerDev: Not likely, because at this point this is the only item in the queue.

components/data_prefetch/data_prefetch.tsx
```diff
@@ -0,0 +1,89 @@
...
+            queue.add(async () => this.prefetchPosts(currentChannelId));
+            await loadProfilesForSidebar();
```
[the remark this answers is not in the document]
sudheerDev: Because it is hard to guess when the profiles are completely loaded so we can start prefetching with the way existing code works. Moving it here makes it easy to assert for dispatching prefetching actions.

components/data_prefetch/data_prefetch.tsx
sudheerDev: ya the queue is not cleared all that well without timeout. let me add a comment about it

components/data_prefetch/data_prefetch.tsx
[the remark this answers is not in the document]
sudheerDev: And we are loading the profiles on first load on LHS only after current channel is set.

components/data_prefetch/data_prefetch.tsx
sudheerDev: We don't have a redux queue, It is just the queue state in redux which is used to not make multiple requests as messages comes into the channel or prevent duplicate requests in cases other components require it.

components/data_prefetch/data_prefetch.tsx
sudheerDev: We wouldn't fetch for the current channel because if there is a prefetch request in progress then it will not be made again. The if loop below in component prevents itt

```
if(this.props.prefetchRequestStatus[priorityQueue[channelIndex]] && this.shouldLoadPriorityQueue(priorityOrder, numberOfPiorityRequests))
```

components/data_prefetch/data_prefetch.tsx
sudheerDev: Add a jitter(0-1sec) for delaying post requests in case of a new message in open/private channels. This is to prevent a case when all clients request messages when new post is made in a channel with thousands of users.

Some of the your questions might be answered if you just read the summary again. 

DM/GM's do not have enough users to cause an issue on the server

components/legacy_sidebar/sidebar.test.jsx
sudheerDev: test says `should match snapshot when isDataPrefechEnabled is disabled`

I am looking to not have `dataPrefetch` component mounted in the snapshot. it is a really simple change for prop and I felt it is good enough to have a snapshot.

components/legacy_sidebar/sidebar.jsx
sudheerDev: I don't feel right given the hierarchy of how the feature works. It works based on an LHS and having it somewhere as a channel controller feels out of place. Yes, it is duplicated but we will be removing the legacy sideebar component and so, I don't see it as a problem.

components/post_view/post_list/post_list.jsx
sudheerDev: I don't see a possibility of `stuck as in-progress somehow`. It will either be morphed into success or failure. 

It is not likely that this happens but if a request fails it still will be the same way it used to be without the failure retires epic.

components/data_prefetch/data_prefetch.tsx
sudheerDev: It is based of off channelId, it is the same as `channelPrefetchStatus` in redux state which has type defined.

components/data_prefetch/index.ts
sudheerDev: It is not for `channel` obj or memberships but in gerenal state changes. Selector felt like overkill and the memoization is just as a precaution.

components/data_prefetch/data_prefetch.tsx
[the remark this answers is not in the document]
hmhealey: This is going to make it so that we make multiple requests to load the profiles, although I can't really see any better way to do that since we don't set a flag or anything to say that profiles have been loaded for the whole sidebar.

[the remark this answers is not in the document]
sudheerDev: This is the only place we are loading the profiles for the first time now. I changed it so that global actions does not, Just in case if that wasn't clear

components/data_prefetch/data_prefetch.tsx
hmhealey: Is `queue.clear` async then? Perhaps we could do `queue.clear().then(() => {...})` instead?

components/data_prefetch/data_prefetch.tsx
hmhealey: Why not move the queue to be a field of the component though?

components/data_prefetch/data_prefetch.tsx
hmhealey: Actually, I think we're properly capturing `currentChannelId` here anyway so the queued function will still contain the `currentChannelId` from the time it was queued instead of any new value.

components/data_prefetch/data_prefetch.tsx
sudheerDev: Any particular reason why we want to move it to the component? I think it is a Usually way convention to initiate classes for external libraries, Why assign it to an instance variable in the component?

Things like this don't change in the lifetime of the app, so i don't see. a good reason to attach it to the instance unless it is a really big object that needs to be GC'ed by the browser.

components/data_prefetch/data_prefetch.tsx
hmhealey: Don't we have something to track whether or not a channel is in sync though? We shouldn't need to fire a prefetch for new messages when the channel is still in sync.

Also, could you add a comment to explain why we need a delay?

components/data_prefetch/data_prefetch.tsx
hmhealey: I don't mean that we'll have two prefetches being made for the current channel. I mean that we'll be making two requests for the posts: one prefetch and the other from the PostList.

sudheerDev: Nope we won't be doing that because postList won't be making a new request if prefetch is in progress

components/data_prefetch/data_prefetch.tsx
sudheerDev: >  We shouldn't need to fire a prefetch for new messages when the channel is still in sync.

yes, we don't.

This is the case when the channel did not have any unread so there was never a prefetch before. On new message say in bugs channel we don't want all users making the call for unreads at once hence the jitter.

components/legacy_sidebar/sidebar.test.jsx
hmhealey: You know what you're looking for here, but others reviewing this in the future might not. I'm trying to avoid us running into cases where something breaks because someone changes something and just updates snapshots without understanding what actually changes since there's so much stuff in the snapshot.

components/data_prefetch/data_prefetch.tsx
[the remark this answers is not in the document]
hmhealey: Ah, I missed that we weren't calling it elsewhere.

components/data_prefetch/data_prefetch.tsx
hmhealey: Avoiding global variables is generally considered to be a good practice since it prevents issues like multiple instances of DataPrefetch affecting each other's behaviour (if we ended up with multiple instances of it) and it helps prevents memory leaks by giving data an owner (like how Rust works without a garbage collector).

In this case, `queue` will be able to hold onto a references to an instance of `DataPrefetch` after it gets unmounted because the calls we're queuing directly reference ([like this one](https://github.com/mattermost/mattermost-webapp/pull/5547/files/e8634150fedef056f7ce3a8734baca9569ecd044#diff-799dd3cdcb584fe132d5d07f232327d0R79)) that hold a reference to the component.

components/data_prefetch/data_prefetch.tsx
hmhealey: Oh, okay. I missed that then.

components/data_prefetch/data_prefetch.tsx
sudheerDev: This is not a global variable. This is scoped inside a module. Like each file works as a module with our build systems and none of the values defined out of component are never global variables and are garbage collected when not needed. 

I don't think GC works that way in the case you described. They cause memory leaks only if they hang on to a detached node or held on with an eventlistner and that is not the case here. I remember seeing quite a few instances across open source projects where a new instance of object is created similar to this at the top of module as I said this is perfectly fine and indeed i say preferred if we don't need to change the way they are instantiated.

components/legacy_sidebar/sidebar.test.jsx
sudheerDev: FYI i pretty much followed how we usually always write for test description including everything in the file for snapshots, Update the text based on your comment though now.

components/data_prefetch/data_prefetch.tsx
sudheerDev: i coudn't replicate the issue anymore to cross check 🤷‍♂️. Changed it to await as i think it should fix it if it happens.

sudheerDev:
@hmhealey Up for review again

esethna:
@jgilliam17 please sync with Sudeer on testing this so we can get it soaking on daily

jgilliam17:
@sudheerDev found one error, I did have some network issues and lost connection for a bit, not sure if its related.
Screen Shot 2020-06-25 at 4 27 41 PM

sudheerDev:
@jgilliam17 Nice catch, Fixed it. 

Repo for the earlier issue:
1. Have a channel with more than 30 unreads
2. Refresh app with another channel in view.
3. Refresh the app
4. Disconnect from internet
5. Connect back to the internet

During the prefetch of the channel to sync we hit the earlier issue

esethna:
@hmhealey just needs your review

components/data_prefetch/data_prefetch.tsx
hmhealey: `queue` can hold onto references of detached nodes, at least for a time while it's working through the queued. It's the same as those warnings we'd get by calling `setState` from async code after a component unmounted, although we just silenced those by checking if the component is mounted which doesn't actually solve that problem now that I think of it. Either way, those references don't hang around for that long, so this is fine for now, although we might want to keep an eye on it if memory usage starts spiking in the future.

PR #5550, opened 2020-05-20
[MM-23690] Not showing updated / correct roles when searching for user in the team members modal after their role has been updated

hahmadia:
#### Summary

If you search for a particular member in the team members modal on chat facing side after a user has had their role updated, it will show an incorrect / out of date role. This PR aims to fix this problem.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23690

#### Related Pull Requests

- ~~Has server changes https://github.com/mattermost/mattermost-server/pull/14615~~
- ~~Has Redux changes https://github.com/mattermost/mattermost-redux/pull/1160~~

MM-23690

### DIFF
--- a/components/member_list_team/member_list_team.tsx
+++ b/components/member_list_team/member_list_team.tsx
@@ -34,7 +34,7 @@ type Props = {
         loadStatusesForProfilesList: (users: Array<UserProfile>) => Promise<{
             data: boolean;
         }>;
-        loadTeamMembersForProfilesList: (profiles: any, teamId: string) => Promise<{
+        loadTeamMembersForProfilesList: (profiles: any, teamId: string, reloadAllMembers: boolean) => Promise<{
             data: boolean;
         }>;
         setModalSearchTerm: (term: string) => Promise<{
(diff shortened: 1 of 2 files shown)


components/member_list_team/member_list_team.tsx
```diff
@@ -134,7 +134,14 @@ export default class MemberListTeam extends React.Component<Props, State> {
         this.loadComplete();
     }
 
-    search = (term: string) => {
+    search = async (term: string) => {
+        await this.props.actions.getTeamMembers(this.props.currentTeamId, 0, this.props.users.length,
```
hahmadia, line 138: When we search, we don't paginate so I set the limit to the number of users in the team.

fmunshi:
I think you can do fix this PR without having to do a server or redux PR.

@hahmadia The logic for searching in this component is done in the `componentDidUpdate` after setting a search term, within that it makes a call to `loadTeamMembersForProfilesList` however `loadTeamMembersForProfilesList` by default will not reload memberships that already exist in the state but I modified the function as part of my TeamMembers block PR to allow for memberships to be reloaded the function definition looks like
`export function loadTeamMembersForProfilesList(profiles, teamId, reloadAllMembers = false) {` now

So I think if you change the block at line 108 to just be
```
loadTeamMembersForProfilesList(data, this.props.currentTeamId, true).then(({data: membersLoaded}) => {
    if (membersLoaded) {
      this.loadComplete();
    }
});
``` 

Then this issue will be resolved

hahmadia:
@fmunshi 
Farhan, I can only describe you as a man with **PRODIGY** thoughts. You just saved me so much time and trouble I can't even explain. Thank you 🙏 !!!!!
I totally missed this and you mentioning it to me blew MY mind. Again thanks for this pointer!

PR #5554, opened 2020-05-21
[MM-22776] Implement auto scroll to permission that is disabled from channel moderation widget

hahmadia:
#### Summary

When you click on the system scheme link or team scheme link, you should be auto scrolled to that permission on the scheme site. This PR aims to implement this.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22776

#### Screenshots

MM-22776


components/admin_console/team_channel_settings/channel/details/channel_moderation.tsx
```diff
@@ -35,15 +35,15 @@ const formattedMessages: any = defineMessages({
         },
         disabledGuests: {
             id: t('admin.channel_settings.channel_moderation.createPosts.disabledGuest'),
-            defaultMessage: 'Create posts for guests are disabled in [{scheme_name}](../permissions/{scheme_link}).',
+            defaultMessage: 'Create posts for guests are disabled in [{scheme_name}](../permissions/{scheme_link}?rowIdFromQuery=create_post).',
```
fmunshi, line 38: Another concern but could this block `?rowIdFromQuery=create_post` be a value that you pass in instead of something thats hard coded in the translation? Since its not going to be changing between translations it would be better to pass it as a value instead of requiring each translation to have the query parameter set properly

hahmadia: Great idea! Done 😄 
dancing

PR #5555, opened 2020-05-21
MM-25014 -update group name if id pattern

sbishel:
#### Summary
When a group link gets created it is being created with a "NewId()" as the name value. That value is then used in the WebApp as the token or slug. This PR creates the slug name when a group is toggled on if the current name is 26 alpha-numeric characters. If the name contains any symbols [-_$] etc. It will not be changed.

It will be checked for uniqueness when saved.
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25014

#### Screenshots
Screen Shot 2020-05-21 at 2 47 36 PM


components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -137,10 +137,17 @@ export default class GroupDetails extends React.PureComponent {
         const {group} = this.props;
         const originalAllowReference = group.allow_reference;
         const saveNeeded = true;
+        let {groupMentionName} = this.state;
+
+        if (!originalAllowReference && allowReference &&
+            groupMentionName.length === 26 && new RegExp('^[a-z0-9]+$').test(groupMentionName)) {
```
fmunshi, line 143: can we define `26` as a constant something like `GROUP_ID_LENGTH` to avoid any confusion as to why we're comparing with an arbitrary number

sbishel: Yes, thank you. Meant to do that.

sbishel:
Closing - decision was made to null the Name field via update. This PR instead https://github.com/mattermost/mattermost-webapp/pull/5569

PR #5556, opened 2020-05-22
MM-25426 Continue to scroll RHS when scrolled to bottom of thread

bradjcoughlin:
#### Summary
Captures the scrolled state of the thread prior to adding the new post to the DOM. If the thread was scrolled to the bottom, the thread will continue to scroll to expose the next incoming message.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25426

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/5470

#### Screenshots
thread-scrolling-rhs

### DIFF
--- a/components/rhs_thread/rhs_thread.tsx
+++ b/components/rhs_thread/rhs_thread.tsx
@@ -120,8 +120,10 @@ export default class RhsThread extends React.Component<Props, State> {
 
         // if snapshot value, new posts have been added
         // if at bottom of thread, continue to scroll after adding new post
-        if (snapshot) {
-            this.scrollToBottom();
+        if (!Utils.areObjectsEqual(prevPostsArray, curPostsArray)) {
+            if (snapshot) {
+                this.scrollToBottom();
+            }
         }
 
         if (prevPostsArray.length >= curPostsArray.length) {
@@ -153,6 +155,10 @@ export default class RhsThread extends React.Component<Props, State> {
         }
 
         if (nextState.isScrolling !== this.state.isScrolling) {
+            return true;
+        }
+
+        if (nextState.isNearBottom !== this.state.isNearBottom) {
             return true;
         }
 
--- a/components/rhs_thread/rhs_thread.tsx
+++ b/components/rhs_thread/rhs_thread.tsx
@@ -102,7 +102,7 @@ export default class RhsThread extends React.Component<Props, State> {
         window.removeEventListener('resize', this.handleResize);
     }
 
-    public componentDidUpdate(prevProps: Props, prevState: State, snapshot?: boolean) {
+    public componentDidUpdate(prevProps: Props, prevState: State) {
         const prevPostsArray = prevProps.posts || [];
         const curPostsArray = this.props.posts || [];
 
@@ -118,10 +118,9 @@ export default class RhsThread extends React.Component<Props, State> {
             this.props.actions.getPostThread(this.props.selected.id);
         }
 
-        // if snapshot value, new posts have been added
-        // if at bottom of thread, continue to scroll after adding new post
+        // if near bottom of thread, continue to scroll after adding new post
         if (!Utils.areObjectsEqual(prevPostsArray, curPostsArray)) {
-            if (snapshot) {
+            if (prevState.isNearBottom) {
                 this.scrollToBottom();
             }
         }
@@ -173,14 +172,6 @@ export default class RhsThread extends React.Component<Props, State> {
         }
 
         return false;
-    }
-
-    public getSnapshotBeforeUpdate(prevProps: Props, prevState: State) {
-        // if adding posts to thread, capture state of rhs to determine if scrolled to botttom
-        if (!Utils.areObjectsEqual(prevState.postsArray, this.props.posts)) {
-            return prevState.isNearBottom;
-        }
-        return null;
     }
 
     private handleResize = (): void => {
--- a/components/rhs_thread/rhs_thread.tsx
+++ b/components/rhs_thread/rhs_thread.tsx
@@ -25,6 +25,8 @@ import {FakePost} from 'types/store/rhs';
 
 import './rhs_thread.scss';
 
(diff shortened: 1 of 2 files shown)


components/rhs_thread/rhs_thread.tsx
```diff
@@ -159,6 +171,14 @@ export default class RhsThread extends React.Component<Props, State> {
         return false;
     }
 
+    public getSnapshotBeforeUpdate(prevProps: Props, prevState: State) {
```
bradjcoughlin, line 174: For reference: https://5d4b5feba32acd0008d0df98--reactjs.netlify.app/docs/react-component.html#getsnapshotbeforeupdate

bradjcoughlin: Needed to capture scrolled position just prior to posts being added to the thread. React documentation shows a very similar example for this lifecycle method.

components/rhs_thread/rhs_thread.tsx
```diff
@@ -254,6 +274,14 @@ export default class RhsThread extends React.Component<Props, State> {
         this.setState({
             isScrolling: false,
         });
+
+        if (this.scrollbarsRef.current) {
```
bradjcoughlin, line 278: Capture scroll position upon stopping scrolling. This defaults to true as the thread is at the bottom when first rendered.

components/rhs_thread/rhs_thread.tsx
```diff
@@ -98,7 +100,7 @@ export default class RhsThread extends React.Component<Props, State> {
         window.removeEventListener('resize', this.handleResize);
     }
 
-    public componentDidUpdate(prevProps: Props) {
+    public componentDidUpdate(prevProps: Props, prevState: State, snapshot?: any) {
```
devinbinnie, line 103: `snapshot` is also of type `State` here.

components/rhs_thread/rhs_thread.tsx
sudheerDev: Could we add a buffer of few pixels from the bottom for this? Like 10px or something so posts are auto scrolled if they are left a bit above the posts?

components/rhs_thread/rhs_thread.tsx
bradjcoughlin: True. I was using `any` because I was getting back `any` 😆

components/rhs_thread/rhs_thread.tsx
sudheerDev: If we are doing the `isAtBottom` check here i.e `elem.scrollTop + elem.clientHeight === elem.scrollHeight` before adding the posts then we need to tap into this lifecycle otherwise i think we could just skip this and do

```
if (curPostsArray.length > prevPostsArray.length) {
            if (prevState.isAtBottom) {
                this.scrollToBottom();
            }
        }
``` 

Is there a reason why we need to check for in this lifecycle?

components/rhs_thread/rhs_thread.tsx
```diff
@@ -114,6 +116,14 @@ export default class RhsThread extends React.Component<Props, State> {
             this.props.actions.getPostThread(this.props.selected.id);
         }
 
+        // if snapshot value, new posts have been added
+        // if at bottom of thread, continue to scroll after adding new post
+        if (curPostsArray.length > prevPostsArray.length) {
```
sudheerDev, line 121: Would this work if a reaction is added?

bradjcoughlin: Possibly not, I should inspect that array.

bradjcoughlin: It seems this check is redundant, because `snapshot` should not return true unless the posts array has mutated per the condition in `getSnapshotBeforeUpdate`

components/rhs_thread/rhs_thread.tsx
bradjcoughlin: Based on limited testing with a few input devices and browsers, a "slight" scroll can easily be 150px. This isn't an exact science but I felt 200px seemed reasonable to assume a user "meant" to be at the bottom.

components/rhs_thread/rhs_thread.tsx
bradjcoughlin: You're right, that works. I think I was looking at this method first prior to deciding to track the scroll position in `handleScrollStop`.

components/rhs_thread/rhs_thread.tsx
bradjcoughlin: You're correct, it doesn't account for reaction and it doesn't appear that `rhs_thread.tsx` has readily available access to the reaction data currently (would require polling Redux). I'm asking @deanwhillier about breaking this issue into another ticket. Since it could potentially increase the scope I think it makes sense to submit another incremental fix.

bradjcoughlin: With the buffer added, the thread will continue to scroll but it's not the best UX, obviously. http://g.recordit.co/lDdklYbGSJ.gif

bradjcoughlin: @sudheerDev I created a new issue for this bug: https://mattermost.atlassian.net/browse/MM-25550

components/rhs_thread/rhs_thread.tsx
sudheerDev: 0/5 but 200px seems a lot to me, I think we do about 10px in center channel.

components/rhs_thread/rhs_thread.tsx
sudheerDev: One more thing, This won't be just for cases like this but it could even be for post edit. 

If someone edits a post and adds a lot of content then it will push the RHS down but does not keep the user to the bottom

bradjcoughlin:
The PR that is the parent feature of this ticket is being reverted. I plan to move it to a feature branch and resubmit this PR against it.

PR #5563, opened 2020-05-24
MM-25050: Enable cluster encryption

agnivade:
We add the admin console UI knobs to enable cluster encryption.

### Ticket link

https://mattermost.atlassian.net/browse/MM-25050


components/admin_console/admin_definition.jsx
```diff
@@ -1070,6 +1070,10 @@ const AdminDefinition = {
                 'admin.cluster.UseIpAddressDesc',
                 'admin.cluster.UseExperimentalGossip',
                 'admin.cluster.UseExperimentalGossipDesc',
+                'admin.cluster.EnableExperimentalGossipEncryption',
+                'admin.cluster.EnableExperimentalGossipEncryptionDesc',
+                'admin.cluster.EncryptionKey',
```
DSchalla, line 1075: I think this line and the one below is dead code?

agnivade: Yep.

agnivade:
Gentle ping on this @sudheerDev

components/admin_console/cluster_settings.jsx
```diff
@@ -211,6 +213,24 @@ export default class ClusterSettings extends AdminSettings {
...
+                    helpText={
+                        <FormattedHTMLMessage
+                            id='admin.cluster.EnableExperimentalGossipEncryptionDesc'
+                            defaultMessage='When true, all communication through the gossip protocol will be encrypted.'
+                        />
+                    }
+                    value={this.state.EnableExperimentalGossipEncryption}
+                    onChange={this.overrideHandleChange}
+                    setByEnv={this.isSetByEnv('ClusterSettings.EnableExperimentalGossipEncryption')}
+                />
```
saturninoabril, line 233: Please add Cypress test, specifically for this new setting. Minimum test could be check/uncheck and save the setting, and then see if it's saved into the server by getting the config (see [related test](https://github.com/mattermost/mattermost-webapp/blob/master/e2e/cypress/integration/system_console/ui_and_api/customization_spec.js)).  Let me know if you need help. Thanks.
(In the future, we might going to add test that query directly into the DB and inspect whether the secret key has been generated after enabling the setting.)

i18n/en.json
```diff
@@ -281,6 +281,8 @@
   "admin.cluster.ClusterNameDesc": "The cluster to join by name.  Only nodes with the same cluster name will join together.  This is to support Blue-Green deployments or staging pointing to the same database.",
   "admin.cluster.ClusterNameEx": "E.g.: \"Production\" or \"Staging\"",
   "admin.cluster.enableDescription": "When true, Mattermost will run in High Availability mode. Please see [documentation](!http://docs.mattermost.com/deployment/cluster.html) to learn more about configuring High Availability for Mattermost.",
+  "admin.cluster.EnableExperimentalGossipEncryption": "Enable Experimental Gossip encryption:",
+  "admin.cluster.EnableExperimentalGossipEncryptionDesc": "When true, all communication through the gossip protocol will be encrypted.",
```
saturninoabril, line 285: Would be nice to have PM review on this UI text.

components/admin_console/cluster_settings.jsx
agnivade: Oh my, back to JS again then :sweat_smile:

agnivade: Done

i18n/en.json
wiersgallak: @agnivade can you please modify to "Enable Experimental Gossip Encryption."  

Also wondering if there is a particular encryption method being used that would be valuable for the system admin to know?

agnivade: @wiersgallak - From what I understand, the ":" is there for all other settings though. See ClusterName, GossipPort, StreamingPort, EnableTitle etc. But happy to change it if you recommend.

The encryption method is AES-256 and is done in consultation with the security team. The choices are between AES-128, 192 and 256. The customer can choose to bring down the key size if they choose to. I will add all of these details in the release notes. For a simple config description, going to such details might not be worth it.

wiersgallak: @agnivade Good catch on the punctuation - my oversight. Just to confirm that there are no other settings to configure the key size? 
Will likely need your help on adding the details to the documentation for this if its something they do outside of the configuration settings in-product.

agnivade: Yes there are no other settings to configure key size. This is by design to make encryption automatic with the flip of a switch.

Yep, I have marked the server PR as "docs needed" and will work with Amy to provide the release docs when 5.26 gets released.

PR #5565, opened 2020-05-24
MM-18788 Remove deprecated lifecycle methods from file previews

CEOehis:
#### Summary
This Pull request removes deprecated lifecycle methods from file attachment previews. Methods like:
- componentWillMount
- componentWillReceiveProps
- componentWillUpdate

Using newer lifecycle methods:
* `getDerivedStateFromProps` to track prop changes
* `getSnapShotBeforeUpdate` to determine next prop changes

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/12337


components/code_preview.jsx
```diff
@@ -24,28 +24,46 @@ export default class CodePreview extends React.PureComponent {
...
+                prevFileUrl: props.fileUrl,
+            };
         }
+        return null;
     }
 
-    updateStateFromProps = (props) => {
-        const usedLanguage = SyntaxHighlighting.getLanguageFromFileExtension(props.fileInfo.extension);
+    componentDidUpdate() {
+        if (this.state.code === '') {
```
hmhealey, line 55: Could you change this so that it also checks if the previous value of `state.code` was empty as well? Right now, it's possible that we'll send multiple requests for the code if the component updates while we're waiting for the code to first load.

components/pdf_preview.jsx
```diff
@@ -42,17 +42,30 @@ export default class PDFPreview extends React.PureComponent {
...
+                loading: true,
+                success: false,
+                prevFileUrl: props.fileUrl,
+            };
         }
+        return null;
     }
 
     componentDidUpdate() {
+        if (this.state.pdf === null) {
```
hmhealey, line 64: Similar to my previous comment, could you add a check here against the previous value of `state.pdf`?

components/view_image/view_image.jsx
```diff
@@ -117,19 +117,34 @@ export default class ViewImageModal extends React.PureComponent {
...
+        if (this.props.show === true && prevProps.show === false) {
+            return 'modalShown';
+        } else if (this.props.show === false && prevProps.show === true) {
+            return 'modalHidden';
         }
+        return null;
+    }
+
+    componentDidUpdate(prevProps, prevState, snapshot) {
+        if (snapshot !== null) {
```
hmhealey, line 130: Instead of using `getSnapshotBeforeUpdate`, could you just compare `this.props.show` and `prevProps.show` here?

components/code_preview.jsx
```diff
@@ -24,28 +24,46 @@ export default class CodePreview extends React.PureComponent {
...
 
     componentDidMount() {
-        this.updateStateFromProps(this.props);
+        this.getCode();
     }
 
-    UNSAFE_componentWillReceiveProps(nextProps) { // eslint-disable-line camelcase
-        if (this.props.fileUrl !== nextProps.fileUrl) {
-            this.updateStateFromProps(nextProps);
+    static getDerivedStateFromProps(props, state) {
```
nevyangelova, line 30: Since we are removing old lifecycle method do you think this is a good candidate for using React Hooks?

components/pdf_preview.jsx
```diff
@@ -81,17 +94,8 @@ export default class PDFPreview extends React.PureComponent {
...
-            pdfPages: {},
-            pdfPagesLoaded: {},
-            numPages: 0,
-            loading: true,
-            success: false,
-        });
-
-        PDFJS.getDocument(props.fileUrl).then(this.onDocumentLoad, this.onDocumentLoadError);
+    getPdfDocument = () => {
+        PDFJS.getDocument(this.props.fileUrl).then(this.onDocumentLoad, this.onDocumentLoadError);
```
nevyangelova, line 98: Can we use then-catch if our promise library allows?

components/code_preview.jsx
hmhealey: I don't think we need to do that now. We can create a followup ticket to do that. For now, I just want to get rid of the deprecated lifecycle methods in favour of the new ones.

components/view_image/view_image.jsx
CEOehis: Alright...I totally agree. I'll make the changes.

components/code_preview.jsx
CEOehis: @hmhealey I think since we only first make a request for code when the component mounts, and the 
We have a guarantee from `getDerivedStateFromProps` that when the fileUrl changes then `this.state.code` would be`''`.

I tried checking if the `prevState.code` was`''` as well, but from the above reasoning, both `this.state.code` and `prevState.code` can't both be an empty string at the same time because we are only fetching the code when the fileUrl changes.

CEOehis: Currently checking if `this.state.code === ''` only sends one request, but if we check if `this.state.code === '' && prevState.code === ''` it doesn't make any request when the next or prev buttons are clicked, because both the current and prev state values for `code` aren't both empty strings at the same time.

components/pdf_preview.jsx
CEOehis: Same reasoning from my comment above applies to this component as well.

components/code_preview.jsx
hmhealey: You'd actually want to check `this.state.code === '' && prevState.code !== ''`. We'd just be double checking that something actually changed before firing the request. If we don't check that something changed, other state changes could cause extra requests to be fired.

CEOehis: Oh, I see what you are saying. For this component, we are only firing the request when the fileUrl changes, and we already guarantee this in `getDerivedStateFromProps` at line 31. That is where we check if the component receives a new `fileUrl` prop. With that, we set `this.state.code = ''` and then check for this in `componentDidUpdate` before actually firing any requests. So other state or prop changes won't result in multiple API calls.

The issue with also checking if `prevState.code !== ''` is if the user is viewing a large code file (above 500kb) we don't fetch the code. Therefore, when the user clicks "next", `prevState.code` will be `''` and no request will be fired for the new code item even though we need to.

hmhealey: What if we checked for something like `this.state.code !== prevState.code || this.props.fileUrl !== prevProps.fileUrl` then? Actions like server requests should only be dispatched when there's been some sort of event like a component being mounted or a state/props change.

CEOehis: Doing `this.state.code !== prevState.code || this.props.fileUrl !== prevProps.fileUrl` fires two API requests for the same document.

CEOehis: I'm 100% positive implementing it as `this.state.code === ''` only sends one API request. I've tested it to quite an extent mocking different scenarios regardless of other state or prop changes. The request is only fired when the component mounts or the fileurl changes.

hmhealey: Sorry, I meant to say that we could do`(this.state.code === '' && prevState.code !== '') || this.props.fileUrl !== prevProps.fileUrl` there, but now that I think of it, that's not all necessary. `this.props.fileUrl !== prevProps.fileUrl` should actually be correct here since `this.state.code` only becomes empty when the the file URL changes, so the extra check isn't needed.

Sorry to keep going on about looking for a change in props/state here before sending the request. I trust that it works as expected here, but my concern is that it might break going forward. The React docs explicitly say you should be looking for a change in props/state before firing network requests from `componentDidUpdate`, so that's what I want us to follow:
> This is also a good place to do network requests as long as you compare the current props to previous props (e.g. a network request may not be necessary if the props have not changed).

components/pdf_preview.jsx
hmhealey: And here would also be `this.props.fileUrl !== prevProps.fileUrl` as well then

components/code_preview.jsx
CEOehis: I understand, makes sense to me. I'll tweak that now.

hmhealey:
Yep, I can do that. This'll be ready to merge as soon as the branch is updated and the final tests are done.

PR #5566, opened 2020-05-25
MM-24180 Make emojis in posts selectable

bradjcoughlin:
#### Summary
Inserts transparent text into the emoticon span to allow copying & pasting of the emoji shortcode.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24180

#### Screenshots
copy-emoji-shortcode


sass/components/_emoticons.scss
```diff
@@ -23,6 +23,7 @@
     min-width: 21px;
     vertical-align: middle;
     width: 21px;
+    color: transparent;
```
bradjcoughlin, line 26: I'm simply making the text "invisible" using styling because both `visibility: hidden` and `display: none` result in a non-copyable element.

components/post_emoji/post_emoji.test.tsx
```diff
@@ -28,6 +34,7 @@ describe('PostEmoji', () => {
 
         const wrapper = shallow(<PostEmoji {...props}/>);
 
+        expect(wrapper.find('span')).toHaveLength(0);
```
bradjcoughlin, line 37: If the plain text is returned, the wrapping `<span>` shouldn't be present.

components/post_emoji/post_emoji.test.tsx
```diff
@@ -20,6 +20,12 @@ describe('PostEmoji', () => {
         });
     });
 
+    test('should render shortcode text within span when imageUrl is provided', () => {
```
bradjcoughlin, line 23: I'd like to add e2e testing, but it appears native browser commands aren't really supported by Cypress. Asking around, but it's looking like we'll have to add copy/paste testing to our manual suite of test.

bradjcoughlin: Perhaps @jgilliam17 can shed some more light. I'll tag @saturninoabril here as well - maybe he has some insights to share.

calebroseland: @bradjcoughlin re: e2e, maybe just checking the text contents (`cy.contains`?) of a post with emoji is good enough? (leaving the selection+copy/paste flows for another test to handle)

bradjcoughlin: @calebroseland yeah, I thought about that as well. I can add that to the existing e2e test, but of course it doesn't quite mimic the real-world interaction. But I agree it's better than nothing.

bradjcoughlin: I added a couple Cypress tests just to verify the element content as you suggested 👍

saturninoabril: +1 to @calebroseland's suggestion. Though it's a Cypress limitation, copy/paste starts with a text to select/highlight and so some test like below is good enough for minimum test requirement.
```javascript
const message = 'This post has a shortcode emoji :raising_hand_man: within it.';
cy.postMessage(message);

    cy.get(`#postMessageText_${postId}`).should('have.text', message);
});
```

e2e/cypress/integration/messaging/message_spec.js
```diff
@@ -137,4 +137,27 @@ describe('Message', () => {
...
     });
+
+    it('message with emoji contains hidden shortcode text', () => {
+        // # Post a message with a shortcode emoji
+        cy.postMessage('This post has a shortcode emoji :raising_hand_man: within it.');
+
+        cy.getLastPostId().then((postId) => {
+            const divPostId = `#post_${postId}`;
+
+            cy.get(divPostId).find('span.emoticon').should('have.text', ':raising_hand_man:');
```
saturninoabril, line 148: I'm good with this but please add one more assertion:
```javascript
cy.get(`#postMessageText_${postId}`).should('have.text', message);
```

bradjcoughlin: Added the assertion @saturninoabril

PR #5567, opened 2020-05-25
[MM-24200] - Create input that supports tags for search hint UI

nevyangelova:
#### Summary
This PR introduces a new component, input which supports rendering of tags. Design https://www.figma.com/file/wxJ9XcA2DPXdgzlO8S7N1G/MM-22806-Search-Popover---Quick-Wins?node-id=0%3A1&viewport=0%2C0%2C1

Due to extensive ref usage around search_bar, suggestion_box etc, it was quite more difficult than expected to make it happen. My intention was to use a functional component with React hooks, which proved impossible without a whole refactor of suggestion box. I would recommend a more ground-up refactor of the search area, so we don't overuse refs.

I remove the getTextbox function and instead expose the methods, I can't rely on getTextbox because its not a textbox anymore. I need to get the value from the tags instead.

There is a two-directional control and data flow when it comes to focusing, which is an anti-pattern. The refactor should fix that too. 

The UI was unclear for some behaviour, for example removing values, editing values etc. so I am expecting feedback. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24200

Fix v5.24


esethna:
Thanks @nevyangelova. @matthewbirtch can you help do a UX pass on this PR?

esethna:
FYI @amyblais we're pushing this to v5.26

components/suggestion/search_suggestion_tagged_input.jsx
```diff
@@ -0,0 +1,212 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import SearchSuggestionInput from './search_suggestion_input';
+const exampleTags = ['from:', 'in:', 'on:', 'before:', 'after:'];
+
+export default class SearchSuggestionTaggedInput extends SearchSuggestionInput {
```
hmhealey, line 10: Also, I'm really hesitant about having this extend `SearchSuggestionInput`. Inheritance isn't recommended with React, and whenever I've tried using it in the past, it's caused problems.

components/suggestion/search_suggestion_tagged_input.jsx
```diff
@@ -0,0 +1,212 @@
...
+        return exampleTags.some((tag) => lcStr.includes(tag));
+    }
+
+    determinePairs(value = this.props.value) {
+        const includesTags = this.includesTags(value);
+        let remainder;
+        let pairs = [];
+
+        if (includesTags) {
+            const split = value.split(/(?<!:) /);
```
hmhealey, line 63: Can we use negative lookbehinds? Last I checked, they weren't supported, and [according to this](https://caniuse.com/#feat=js-regexp-lookbehind), Safari and Firefox don't have support for them still.

hmhealey, line 63: Also, it doesn't look like this handles things like quoted search phrases. We might need to build a more sophisticated parser to handle this since we might be getting outside the bounds of what can be done with a simple split.

[This code](https://github.com/mattermost/mattermost-webapp/blob/master/utils/text_formatting.tsx#L647) may be helpful, although it kind of does the opposite thing we want here since it specifically removes the flags from the search query.

components/suggestion/search_suggestion_tagged_input.jsx
```diff
@@ -0,0 +1,212 @@
...
+        }).join(' ');
+
+        if (pairs.length > 0 && Boolean(pairs[pairs.length - 1].value)) {
+            pairsText += ' ';
+        }
+
+        return pairsText;
+    }
+
+    componentWillReceiveProps(nextProps) {
```
hmhealey, line 126: This method is deprecated. Can this be done with `getDerivedStateFromProps` instead? I think we'd just have to make `determinePairs` and its helpers static or pull them out of the component entirely for that.

Also, can you make sure this logic only triggers when `nextProps.value !== this.props.value` to avoid unnecessary state changes?

components/suggestion/search_suggestion_tagged_input.jsx
```diff
@@ -0,0 +1,212 @@
...
+                                    className='value'
+                                >
+                                    {value}
+                                </span>
+                            )}
+                        </li>
+                    ))}
+                </ul>
+                <input
+                    ref={(current) => {
```
hmhealey, line 198: Shouldn't this just be `ref={this.inputRef}`?

components/suggestion/suggestion_box.jsx
```diff
@@ -189,9 +196,11 @@ export default class SuggestionBox extends React.PureComponent {
             selection: '',
             allowDividers: true,
             presentationType: 'text',
+            tags: [],
```
hmhealey, line 199: Does this component actually need to keep track of its tags? This seems like it might be left over code.

components/suggestion/suggestion_box.jsx
```diff
@@ -189,9 +196,11 @@ export default class SuggestionBox extends React.PureComponent {
...
         };
 
         this.inputRef = React.createRef();
+        this.inputCompRef = React.createRef();
```
hmhealey, line 203: Same with using `inputCompRef` instead of `inputRef`

components/suggestion/suggestion_box.jsx
```diff
@@ -709,16 +695,29 @@ export default class SuggestionBox extends React.PureComponent {
...
-                    autoComplete='off'
-                    {...props}
-                    onInput={this.handleChange}
-                    onCompositionStart={this.handleCompositionStart}
-                    onCompositionUpdate={this.handleCompositionUpdate}
-                    onCompositionEnd={this.handleCompositionEnd}
-                    onKeyDown={this.handleKeyDown}
-                />
+                {renderTaggedInput ?
+                    <SearchSuggestionTaggedInput
```
hmhealey, line 699: Instead of explicitly rendering `SearchSuggestionTaggedInput` here, would it be possible to do this with `QuickInput`'s `inputComponent` prop? That's already used to switch between an `input`, `textarea`, and `LocalizedInput` in various places where `SuggestionBox` is used.

stylianosrigas:
I had to remove the Setup Cloud Test Server label because we need to deprecate the cluster the installation is running. Reapplied successfully.

components/suggestion/search_suggestion_tagged_input.jsx
```diff
@@ -0,0 +1,231 @@
...
+};
+
+const determinePairs = (value = this.props.value) => {
+    const pairIncludesTags = includesTags(value);
+    let remainder;
+    let pairs = [];
+
+    if (pairIncludesTags) {
+        const split = value.split(/\s(?=\w+:)/);
+        let lastEntry = split[split.length - 1];
```
hmhealey, line 29: We might've missed this in the previous version, but what if earlier entries aren't tags and values like you search for `apples in:town-square`?

nevyangelova: then the remainder will be the value and thus appear as the value of the input.

hmhealey: Looks like this actually works fine if text appears before a tag.

Screen Shot 2020-06-19 at 12 19 06 PM

hmhealey:
When I was trying to check out the case I commented on, I found a couple other bugs:

1. Clicking anywhere on the search box sets your cursor to the end of the text, even if you click earlier on it, even if you're already focused on the textbox and you're trying to move the cursor to a different part of the text.
2. Similarly to the ctrl+A one above, home/end and just moving the cursor with the arrow keys stop at a search tag.

esethna:
@nevyangelova should we push this to v5.28 or are we still aiming to get this in (feature complete is next friday)

nevyangelova:
@esethna it can go in 5.26 the only reason why this isn't addressed is because of the prior work on common components and now the sidebar bugs. But the changes are small and I had a video call with @matthewbirtch about them and we are on the same page.

nevyangelova:
@matthewbirtch @hmhealey with the current implementation this is the possible behaviour, regarding the position of the caret, since it's not really an input. I have some issues figuring out the blur on clearing if anyone can advice.

hmhealey:
Perhaps this should be revisited later then when we have more time to work on it? Being able to properly select text and navigate the search box with the keyboard is pretty critical to having it work correctly.

Perhaps we could solve the blurring problem by refocusing the textbox once the X button is clicked on? We have some logic for that in the post textboxes that we use to keep keyboard focus in them after tapping the send button on mobile. Hopefully it's more simple here. Alternatively, I don't know if there's any CSS that we can use to keep the X button clickable, but prevent it from gaining focus when clicked on.

nevyangelova:
> Perhaps this should be revisited later then when we have more time to work on it? Being able to properly select text and navigate the search box with the keyboard is pretty critical to having it work correctly.
> 
> Perhaps we could solve the blurring problem by refocusing the textbox once the X button is clicked on? We have some logic for that in the post textboxes that we use to keep keyboard focus in them after tapping the send button on mobile. Hopefully it's more simple here. Alternatively, I don't know if there's any CSS that we can use to keep the X button clickable, but prevent it from gaining focus when clicked on.

I think we can merge it and see how people feel about the behaviour. If its odd, we can postpone it for 5.28, but again what we are talking about achieving won't be possible with the current implementation and I highly suggest the refactor of search components @esethna @matthewbirtch @hmhealey @deanwhillier

matthewbirtch:
To be honest, I don't know if we should merge this in. The behavior isn't right as a number of other comments have identified.

1. I can't click on text that is tied to a tag
2. I can't click on text between two tags
3. I can't click on a tag to then remove it - I have to backspace the whole thing
4. If I try to click anywhere other than the end of the search query, it just puts the cursor at the end
5. If I'm focused on the input and then click a tag it feels buggy because the search hint and the focus state flickers

I agree with @hmhealey 
> Being able to properly select text and navigate the search box with the keyboard is pretty critical to having it work correctly.

If a refactor is the only way to resolve this, we should do that instead. Right now, it works better without the tags.

esethna:
Agree with Matt, the cursor behaviour is really funky right now and we can't push this feature with those usability issues

nevyangelova:
@esethna @matthewbirtch ok I am happy we all agree. I will create a project plan for the search refactor and introduce it after I am back from my sick leave :)

esethna:
@nevyangelova thanks, let's please discuss with Dean first as he is owning the approach to common components and we need to prioritize this among all the other components we have to refactor

cpanato:
test server removed due some underlying issues in the cluster that made this one unstable.

nevyangelova:
Closing in favour of the common components epic. This feature will be re-implemented and refactored for better UX experience when time allows. @esethna @deanwhillier @matthewbirtch

PR #5569, opened 2020-05-25
MM-25014: Set GroupMention Name if null when toggled

sbishel:
#### Summary
A group name will now be null if AllowGroupReference is false. When AllowGroupReference is toggled, this PR will create the GroupMentionName based off the Display Name.

It will be checked for uniqueness when saved.
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25014
https://mattermost.atlassian.net/browse/MM-25446

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/14651

#### Screenshots
Screen Shot 2020-05-21 at 2 47 36 PM


components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -137,10 +137,16 @@ export default class GroupDetails extends React.PureComponent {
         const {group} = this.props;
         const originalAllowReference = group.allow_reference;
         const saveNeeded = true;
+        let {groupMentionName} = this.state;
+
+        if (!originalAllowReference && allowReference && (typeof groupMentionName === 'undefined' || groupMentionName === '')) {
```
fmunshi, line 142: I think you could just do 
```suggestion
        if (!originalAllowReference && allowReference && !groupMentionName) {
```

package-lock.json
```diff
@@ -3190,8 +3190,7 @@
     "@types/highlight.js": {
       "version": "9.12.3",
       "resolved": "https://registry.npmjs.org/@types/highlight.js/-/highlight.js-9.12.3.tgz",
-      "integrity": "sha512-pGF/zvYOACZ/gLGWdQH8zSwteQS1epp68yRcVLJMgUck/MjEn/FBYmPub9pXT8C1e4a8YZfHo1CKyV8q1vKUnQ==",
-      "dev": true
```
fmunshi, line 3194: Need to checkout `package-lock`

components/admin_console/group_settings/group_details/group_details.test.jsx
```diff
@@ -142,4 +142,39 @@ describe('components/admin_console/group_settings/group_details/GroupDetails', (
         expect(actions.link).toBeCalledWith('xxxxxxxxxxxxxxxxxxxxxxxxxx', '22222222222222222222222222', 'team', {auto_add: true});
         expect(actions.link).toBeCalledTimes(2);
     });
+
+    test('update name for null slug', async () => {
+        const wrapper = shallow(
+            <GroupDetails
+                {...defaultProps}
+                group={{display_name: 'test group', allow_reference: false}}
```
mkraft, line 150: I'm not sure what the server-side validations are for the name field. Does `.toLowerCase().replace(/\s/g, '-')` ensure a valid name?

sbishel: No, the display_name could have what we consider "invalid" characters. But that would be verified server side on save, same as duplicate check. We don't necessarily guarantee this name is valid and unique. But is just a default name. If we expect illegal characters (like space) we can replace them. 

Handling this client side brings additional questions, do we replace the illegal characters or warn the user? Do we disable save until valid? Shouldn't we then validate whenever the user makes a change to be complete. I would suggest leaving that check on the server.

PR #5570, opened 2020-05-25
Rename Error type to ServerError

hmhealey:
In https://github.com/mattermost/mattermost-redux/pull/1162, I renamed the `Error` type we were using to `ServerError` since it conflicted with the built-in `Error` type, but I forgot to do that on the webapp side as well. This fixes that.

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1162

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5444,7 +5444,7 @@
         },
         "p-cancelable": {
           "version": "0.4.1",
-          "resolved": "http://registry.npmjs.org/p-cancelable/-/p-cancelable-0.4.1.tgz",
+          "resolved": "https://registry.npmjs.org/p-cancelable/-/p-cancelable-0.4.1.tgz",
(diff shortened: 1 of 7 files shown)


hmhealey:
Skipping QA review since this only causes build errors

package-lock.json
```diff
@@ -5444,7 +5444,7 @@
         },
         "p-cancelable": {
           "version": "0.4.1",
-          "resolved": "http://registry.npmjs.org/p-cancelable/-/p-cancelable-0.4.1.tgz",
+          "resolved": "https://registry.npmjs.org/p-cancelable/-/p-cancelable-0.4.1.tgz",
```
fmunshi, line 5447: I think a bunch of these `-lock` changes should not be checked in right? iirc I broke CI a while back by accidentally checking in similar changes

hmhealey: I think having them go back to https should be fine. I still don't know why npm switches those occasionally other than that they seem to change betweel `npm install` and `npm add`. Regardless, I'll revert them since they're not related to what I meant to change.

PR #5572, opened 2020-05-25
[MM-24705] Add telemetry for Experimental Sidebar Phase 2

devinbinnie:
#### Summary
Telemetry calls for various points in the new sidebar, as described by spec.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24705

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2.1
 
 orbs:
   aws-s3: circleci/aws-s3@1.0.11
+  owasp: entur/owasp@0.0.10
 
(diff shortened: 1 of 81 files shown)


esethna:
@devinbinnie were we able to add all events in the spec?

devinbinnie:
> @devinbinnie were we able to add all events in the spec?

Yes, except for this one:

```
Can we get data or do we already have data that tells us the number of DMs users have in their sidebar? Ie DMs that are visible in the sidebar because they haven’t been closed (or hidden via the hide after 7 day setting). 
```

PR #5592, opened 2020-05-26
[MM-25409] Switch to css variable for search options hover state

devinbinnie:
#### Summary
The new Search Options popout was using a static colour for the hover state.
This PR switches to using a dimmed version of the center channel colour, which works on all themes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25409


asaadmahmood:
@devinbinnie Please make them use the 08 variable instead of introducing 06.

PR #5598, opened 2020-05-26
MM-24751 - Upgrading the plugin from 0.2 to 0.3 changes the incident list to a blank RHS

ali-farooq0:
#### Summary
Going with a simple solution to close RHS when a plugin is upgraded/removed.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-24751

#### Screenshots (closes on RHS on upgrade)
2020-05-28 at 11 26 AM


plugins/rhs_plugin/index.js
```diff
@@ -12,13 +12,14 @@ function mapStateToProps(state) {
     const rhsPlugins = state.plugins.components.RightHandSidebarComponent;
     const pluginId = getPluginId(state);
 
-    const plugin = rhsPlugins.find((element) => element.id === pluginId);
-    const pluginName = plugin ? plugin.title : '';
-    const icons = plugin ? plugin.icons : [];
+    const pluginComponent = rhsPlugins.find((element) => element.pluginId === pluginId);
+    const pluggableId = pluginComponent ? pluginComponent.id : null;
```
cpoile, line 16: I might be misunderstanding this since they're both named the same, but does the change above (`(this.props.pluggableId !== undefined)`) conflict with this?

ali-farooq0: We're passing in a `null` value here when we can't locate the respective plugin component in the store. `Pluggable` component attempts to find from its list of components if `pluggableId` field is set, or displays all the components (in a row) if it's not set (`undefined`). In order to prevent showing other components in the RHS while the webapp loads the new manifest, we simply set the value of `null` to indicate RHS to not load any component (dictated by the logic defined in `Pluggable`).

Unfortunately there is a subtle distinction between how JS defines `null` vs `undefined.

Treating `null` as setting a unknown/uninitialized value.
Treating `undefined` as the field was never set.

https://codeburst.io/javascript-null-vs-undefined-20f955215a2

cpoile: What I meant was, prior to the change, pluggable was testing `plugableId`. Now it's equality testing on undefined, so null passes through. 
I didn't read down to see plugableId only being used for an equality check, so nothing will break. So nevermind.

ali-farooq0: ahh gotcha. Correct, null will be passed through now.

lieut-data: Is there an opportunity to make this behaviour explicit? I find the juggling of `undefined` and `null` resulting in different outcomes to be relatively unexpected.

ali-farooq0: Passing in an explicit `showPluggable` flag.

plugins/registry.js
```diff
@@ -608,9 +608,9 @@ export default class PluginRegistry {
 
         return {
             id,
-            showRHSPlugin: showRHSPlugin(id),
-            hideRHSPlugin: hideRHSPlugin(id),
-            toggleRHSPlugin: toggleRHSPlugin(id),
+            showRHSPlugin: showRHSPlugin(this.id),
```
lieut-data, line 611: Apologies for the thrashing, but I'm now better understanding the scope of the changes here -- does this mean a plugin can now register at most one RHS component? Would that be a breaking change for some plugins?

I wonder if we should just close the RHS when a plugin is unloaded, full stop, along with expunging all Redux state associated with a plugin. Then the empty RHS will never occur, albeit at the cost of no longer preserving any state between plugin instances. (Though, is effecting such a change complicated by the pending Redux changes to properly emit the actions when this occurs?) 

Using the force, I imagine that it's safer for plugins (and their registered components) to start from a clean slate than to assume they know how to handle old Redux state from a previous instance. 

Thoughts?

ali-farooq0: Ahh hadn't considered that we allowed registering for multiple plugin components 🤔 as I don't think that's a common practice. Although not impossible. K let me see if it's easy enough to dispatch closing the RHS action.

ali-farooq0: My initial assumption was that plugins couldn't register multiple component, which is why I hadn't gone down the path of closing RHS as I personally think that's a bit destructive (on upgrades).

However seeing that my assumption turned out to be incorrect, the only way to fix this /is/ to close RHS as we have no way of knowing what to replace RHS with post-upgrade.

Updated code to close RHS on plugin upgrade/removal.


Requesting re-review from @marianunez and @cpoile hopefully for the last time 😉. Apologies again for the review thrash.

plugins/rhs_plugin/index.js
```diff
@@ -10,15 +10,17 @@ import RHSPlugin from './rhs_plugin.jsx';
...
-    const icons = plugin ? plugin.icons : [];
+    const pluginComponent = rhsPlugins.find((element) => element.id === pluginComponentId);
+    const pluggableId = pluginComponent ? pluginComponent.id : null;
+    const pluginTitle = pluginComponent ? pluginComponent.title : '';
+    const icons = pluginComponent ? pluginComponent.icons : [];
 
     return {
-        pluggableId: pluginId,
-        title: pluginName,
+        showPluggable: Boolean(pluggableId),
```
ali-farooq0, line 21: deciding to keep this despite our pivot in direction as it's better to be explicit in showing/hiding the RHS pane for plugins.

plugins/rhs_plugin/index.js
```diff
@@ -10,14 +10,16 @@ import RHSPlugin from './rhs_plugin.jsx';
 
 function mapStateToProps(state) {
     const rhsPlugins = state.plugins.components.RightHandSidebarComponent;
-    const pluginId = getPluginId(state);
+    const pluginComponentId = getPluginId(state);
```
lieut-data, line 13: The `getPluginId` selector (and associated actions / reducers) remain confusing, since they imply a plugin id, but it's really this `pluggableId`. Worth fixing that for consistency?

Separately, I wonder if we can just call /this/ variable `pluggableId`, and then dispense with the redefinition of `plugableId` below and simply set `showPluggable` to `Boolean(pluginComponent)`?

plugins/index.js
```diff
@@ -203,3 +207,16 @@ export async function loadPluginsIfNecessary() {
         }
     });
 }
+
+// Closes RHS if its showing this plugin.
```
lieut-data, line 211: For consistency with other comments in this file, suggest:
```suggestion
 // hideRHSPlugin closes the RHS if currently showing this plugin.
```

plugins/index.js
```diff
@@ -90,6 +92,7 @@ export function loadPlugin(manifest) {
 
         if (oldManifest) {
             // upgrading, perform cleanup
+            hideRHSPlugin(oldManifest.id);
```
lieut-data, line 95: It's slightly annoying to have an action creator to dispatch an event that could conceivably just be handled by the `REMOVED_WEBAPP_PLUGIN` within the reducers. Would it be possible to store the real `pluginId` alongside the `pluginComponentId` in the store and just key off the existing event below?

ali-farooq0: Even if I were to store `pluginId` and `pluggableId` in the store (as reducers) I will still not be able to close RHS from within the reducer as  the [`rhsState`](https://github.com/mattermost/mattermost-webapp/blob/b806d43982e4d34beca05aa632f3f9cc3263cacd/reducers/views/rhs.js#L93) reducer does not have any knowledge about which `plugginId` or `pluggableId` is currently set. Which means I'll have to either:

1. Share state between reducers
   * Not recommended as it'll cause more confusion down the line

2. Include more data when dispatching `REMOVED_WEBAPP_PLUGIN` that can help `rhsState` make the proper deicion
3.

lieut-data: This approach is fine for now. I think the problem with `reducers/views/rhs.js` is that everything is so granular. If we merged some of the reducers together and stored an object for each, it would be easy to key off `REMOVED_WEBAPP_PLUGIN` in the right conditions with presumably no other changes to the selectors and such.

ali-farooq0: That's true. Combining them will work. Just need to put some thought into which ones to combine.

ali-farooq0:
@prapti heads up this is ready for QA.

prapti:
@ali-farooq0 I'm still seeing the original behavior where an empty RHS loads upon plugin update. The screenshot here is right after an update. I am running this PR branch but could I be missing anything? Doesn't seem like it but would I need to test with any specific plugin versions?
Screen Shot 2020-05-29 at 5 34 26 PM

ali-farooq0:
Thanks @prapti! Couldn't reproduce the issue we spoke about on the cloud env either, tried locally and seeing some oddities, however since I'm not able to replicate that on the cloud, makes me think it might be something to do with my local env. Please keep an eye out for the issue and let me know if you see it again.

Merging this change.

PR #5599, opened 2020-05-27
Update NOTICE.txt

amyblais:


amyblais:
@grundleborg Would you be open to merging this as I don't have permissions?

PR #5607, opened 2020-05-27
Add a simple test to allow team scheme delete issue to be reproduced easily

fmunshi:
#### Summary
- Adds a cypress test which creates and assigns a scheme to `eligendi` team and then deletes it 100 times.

This test case can be used to easily reproduce the issue highlighted in the ticket by logging in as a user in the `eligendi` team in a separate browser and observing the console for 400 errors, or watch the LHS as all channels disappear after a while

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24560

#### Related Pull Requests
- https://github.com/mattermost/mattermost-server/pull/14671


fmunshi:
This PR will be closed when https://github.com/mattermost/mattermost-server/pull/14671 is merged it is only open here to allow the issue to be reproduced easily

PR #5622, opened 2020-05-29
MM-22941 Add icons to status menu for quicker recognition

bradjcoughlin:
#### Summary
Adds Online, Away, Do Not Disturb, and Offline icons to the status menu. Indents help text with menu item text if icon is present.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22941

#### Related Pull Requests
N/A

#### Screenshots
status-menu-icons


bradjcoughlin:
@matthewbirtch I've implement the `theme.centerChannelColor` at `0.6` opacity (alpha). Let me know what you think:
Screen Shot 2020-05-29 at 7 27 57 PM
Screen Shot 2020-05-29 at 7 27 35 PM

sass/components/_status-icon.scss
```diff
@@ -193,3 +193,18 @@
         top: -1px;
     }
 }
+
+.status-dropdown-menu {
+    .online--icon > svg {
+        fill: rgba(var(--online-indicator-rgb), 1);
```
bradjcoughlin, line 199: When I tried to use `rgb()` here, I was getting a compile error stating that the `$green` argument was missing. Using `rgba()` for now until I figure out why. 🤔

sass/components/_status-icon.scss
```diff
@@ -193,3 +193,18 @@
...
+    }
+    .away--icon > svg {
+        fill: rgba(var(--away-indicator-rgb), 1);
+    }
+    .dnd--icon > svg {
+        fill: rgba(var(--dnd-indicator-rgb), 1);
+    }
+    .offline--icon {
+        fill: rgba(var(--center-channel-color-rgb), 0.6);
```
matthewbirtch, line 208: I believe we have variables for the different opacities of the center channel (in increments of 8%). So for this I think we should be able to use:

`fill: var(--center-channel-color-56);`

In fact, I'm not sure we need the `rgba()` wrapping the variables here at all.

bradjcoughlin: Sure thing, I'll adjust the alpha to `0.56`.

Apps team had a thread with @asaadmahmood about the variables and the `rgba()` function. The plan moving forward is to implement using the above pattern, but follow the guidelines set forth by design in regards to the `alpha` values.

Asaad provided the following for our reference.
Screen Shot 2020-06-01 at 9 26 58 AM

And in regards to the `rgba()` function, it's a CSS function that accepts an "rgb triplet" like '255, 255, 255'. I've added a `-rgb` value for each of the theme global variables. These will be used to apply opacity changes in the `.scss` rather than the `changeOpacity` function in the javascript.

components/widgets/menu/menu_items/menu_item.tsx
```diff
@@ -35,9 +35,11 @@ export default function menuItem(Component: React.ComponentType<any>) {
                 );
             }
 
+            const iconClassname = icon ? ' MenuItem--with-icon' : '';
+
             return (
                 <li
-                    className='MenuItem'
+                    className={`MenuItem${iconClassname}`}
```
hmhealey, line 42: Another way you could do this is using the `classnames` utility that we've started using elsewhere. It's a bit easier than fiddling around with appending strings and making sure spaces exist properly
```
className={classnames('MenuItem', {
    'MenuItem--with-icon': icon,
})
```

utils/utils.jsx
```diff
@@ -474,6 +474,12 @@ export function isHexColor(value) {
     return value && (/^#[0-9a-f]{3}([0-9a-f]{3})?$/i).test(value);
 }
 
+// given '#fffff', returns '255, 255, 255' (no trailing comma)
+export function toRgbValues(hexStr) {
+    const hslaStr = `${parseInt(hexStr.substr(1, 2), 16)}, ${parseInt(hexStr.substr(3, 2), 16)}, ${parseInt(hexStr.substr(5, 2), 16)}`;
```
hmhealey, line 479: Shouldn't this be `rgbStr` instead of `hslaStr`?

bradjcoughlin: Oops, yes.

utils/utils.jsx
```diff
@@ -474,6 +474,12 @@ export function isHexColor(value) {
     return value && (/^#[0-9a-f]{3}([0-9a-f]{3})?$/i).test(value);
 }
 
+// given '#fffff', returns '255, 255, 255' (no trailing comma)
+export function toRgbValues(hexStr) {
```
devinbinnie, line 478: Love this :)

components/status_dropdown/status_dropdown.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+import StatusDropdown from './status_dropdown';
+
+describe('components/StatusDropdown', () => {
+    const actions = {
+        openModal: jest.fn(),
+        setStatus: jest.fn(),
+    };
+
+    const baseProps = {
+        actions,
```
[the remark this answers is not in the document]
bradjcoughlin: The only other prop that affects the output of the snapshot is `profilePicture` which I added. I suppose we could add test for the other internal functions, but I didn't as this PR wasn't changing those. I did, however, test the status changing functions in the e2e test.

PR #5624, opened 2020-05-29
MM-25604 Add help text check to session length settings

saturninoabril:
#### Summary
- Add help text check to session length settings
- Add reference to TM4J test

#### Ticket Link
JIRA ticket - https://mattermost.atlassian.net/browse/MM-25604

PR #5630, opened 2020-06-01
MM-25604 E2E for extending session with web/email login

saturninoabril:
#### Summary
E2E for extending session with web/email login.  Session is controlled by the server and so the workaround from Cypress (`cy.clock`) is not possible.

To automate testing of sessions, the test directly manipulates session data, especially the `expiresat` field to move around -- make it soon to expire, expired, etc -- and then observe the side-effect to the user -- extend session, remain the same or automatically logout.

@wiggin77 Could you please help clarify the behavior on when to extend the session? Say the session length is set to 1 day. I observed that the session is being extended every time it reaches the threshold (~14 mins).
- First 14 minutes, the session is not extended (which is expected).
- 15 minutes onward, the session is being extended right away. Should it be extended only if it reaches "`1 day` - `~14 minutes`" which is a period before it expires?

#### Ticket Link
Jira ticket - https://mattermost.atlassian.net/browse/MM-25604

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/5624


wiggin77:
> @wiggin77 Could you please help clarify the behavior on when to extend the session? Say the session length is set to 1 day. I observed that the session is being extended every time it reaches the threshold (~14 mins)

The session is extended if there is user activity after the lessor of 1% of the session duration or 1 day has elapsed.  1% of a 1 day session length is a little over 14 minutes.  Therefore, if there is regular activity, the session will get extended every 14.4 minutes.  The session will be extended to current time plus the session length.  If the session length is 30 days then user activity will extend the session every 7.2 hours.

wiggin77:
> 15 minutes onward, the session is being extended right away. Should it be extended only if it reaches "1 day - ~14 minutes" which is a period before it expires?

@saturninoabril , are you saying that in 5.22, after the initial extension at 14.4 minutes, it then extends the session on every user activity instead of waiting another 14.4 minutes?  That's not what my tests are showing.

saturninoabril:
@wiggin77 No, it's exactly the way you explained. My only confusion is on why it needs to extend early on after 14.4 minutes and why not on the later stage? But I think I get it now. Thanks for taking time to explain.

e2e/cypress/integration/enterprise/extend_session/email_login_activity_spec.js
```diff
@@ -0,0 +1,228 @@
...
+        // # Login as test user and go to town-square channel
+        cy.apiLogin(testUser);
+        cy.visit(townSquarePage);
+
+        // # Get active user sessions as baseline reference
+        cy.dbGetActiveUserSessions({username: testUser}).then(({sessions: initialSessions}) => {
+            // Post a message to a channel
+            cy.postMessage(Date.now());
+
+            const expiredSession = parseDateTime(initialSessions[0].createat, 10) + 1;
```
wiggin77, line 53: I'm confused here how adding 1ms to existing ExpiresAt necessarily expires that session.  If the current ExpiresAt is past then it is already expired, and if not, adding 1ms is not likely to make it so.  I might be missing something.

saturninoabril: ExpiresAt here is still in the future which is on next day after a session is created on [login](https://github.com/mattermost/mattermost-webapp/pull/5630/files#diff-6298b162b7c6ee045aebc986b74f5489R45). I picked up the lowest value of 1ms to guarantee that it's expired during the time of [reloading a page](https://github.com/mattermost/mattermost-webapp/pull/5630/files#diff-6298b162b7c6ee045aebc986b74f5489R66).

e2e/cypress/integration/enterprise/extend_session/email_login_activity_spec.js
```diff
@@ -0,0 +1,228 @@
...
+
+                // # Invalidate cache and reload to take effect the expired session
+                cy.externalRequest({user: users.sysadmin, method: 'POST', path: 'caches/invalidate'});
+                cy.reload();
+
+                // # Try to visit town-square channel
+                cy.visit(townSquarePage);
+
+                // * Verify that it redirects to login page due to expired session
+                cy.url().should('include', `/login?redirect_to=${townSquarePage.replace(/\//g, '%2F')}`);
```
prapti, line 72: I haven't seen redirect being verified like this in the url, but this is very clever. I would normally verify that the user landed in the login page, but this makes testing easier 👍

saturninoabril:
@prapti Did you run `npm install`?  The error you're getting is looking for a missing package/module `lodash.mapkeys`.

e2e/cypress/integration/enterprise/extend_session/email_login_activity_spec.js
saturninoabril: Yeah, checking the page will be the full test for it.  But since we have several places where we verify redirection, I think this is meeting the minimum requirement of whether it redirects or not.

prapti:
I thought I did. I'll try again. Thanks Saturn!

prapti:
Yup, it indeed was only the missing dependency. Thanks again!
I'll review again once Doug's review is complete.

PR #5637, opened 2020-06-01
[MM-25685] Change out of channel msg to show X others when >3 users

fmunshi:
#### Summary
- When a large number of out of channel group members is mentioned it should not show the text as expanded by default, this fixes that.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25685

#### Screenshots
Screen Shot 2020-06-01 at 4 12 25 PM

### DIFF
--- a/components/post_view/post_add_channel_member/post_add_channel_member.jsx
+++ b/components/post_view/post_add_channel_member/post_add_channel_member.jsx
@@ -141,7 +141,7 @@ export default class PostAddChannelMember extends React.PureComponent {
                     />
                     {commaSeparator(1)}
                     <a
-                        id='post_body_other_users_link'
+                        className='PostBody_otherUsersLink'
                         onClick={this.expand}
                     >
                         <FormattedMessage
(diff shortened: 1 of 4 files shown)


components/post_view/post_add_channel_member/post_add_channel_member.jsx
```diff
@@ -94,26 +106,57 @@ export default class PostAddChannelMember extends React.PureComponent {
...
-                            return [...acc, commaSeparator(idx), el];
-                        }, [])
-                    }
+                    <AtMention
+                        key={firstUserName}
+                        mentionName={firstUserName}
+                    />
+                    {commaSeparator(1)}
+                    <a
+                        id='post_body_other_users_link'
```
devinbinnie, line 144: Is it possible for this component to be rendered in the DOM more than once?
If so, having 2 elements with this `id` is considered bad form.
If not it's alright but I would avoid using `id` here if you don't need it.

fmunshi: the main reason I added it was to discern between the other link within this component for testing... it is possible for this component to be on the page multiple times, would a class be preferable for this use case @devinbinnie?

devinbinnie: Yes definitely, or you can add a custom data tag like we use for Cypress.
Normally it's called `data-testid`.

PR #5644, opened 2020-06-03
MM-25696 - Updating z-order for channel header and sidebar

marianunez:
#### Summary
Minor CSS fix to z-order for channel header to avoid the related bug. 

#### Screenshots
**Before:**

Screen Shot 2020-06-03 at 12 27 16 PM

**After:**

Screen Shot 2020-06-03 at 12 26 06 PM

#### Ticket Link
[MM-25696](https://mattermost.atlassian.net/browse/MM-25696)

### DIFF
--- a/sass/layout/_headers.scss
+++ b/sass/layout/_headers.scss
@@ -2,7 +2,7 @@
 
 .channel-header {
     @include flex(0 0 63px);
-    z-index: 11;
+    z-index: 15;
     background: v(center-channel-bg);
     max-height: 63px;
     border-bottom: 1px solid;
--- a/sass/layout/_legacy-sidebar-left.scss
+++ b/sass/layout/_legacy-sidebar-left.scss
@@ -8,7 +8,7 @@
     left: 0;
     position: fixed;
     width: 220px;
-    z-index: 12;
+    z-index: 16;
 
     body.announcement-bar--fixed & {
         height: calc(100% - #{$announcement-bar-height});


bradjcoughlin:
I only increased the z-index for `#SidebarContainer` in my PR https://github.com/mattermost/mattermost-webapp/pull/5622/files#diff-49fef267656aa49c266e7f1536ebbf77R32

sass/layout/_headers.scss
```diff
@@ -2,7 +2,7 @@
 
 .channel-header {
     @include flex(0 0 63px);
-    z-index: 15;
+    z-index: 11;
```
asaadmahmood, line 5: Looks fine, but I think we can remove this change.

marianunez:
> I have a fix for this in #5622, but I think it's best that @asaadmahmood coordinates here. The changes he made previously were to address some changes to the header in RHS.

Thanks @bradjcoughlin! Wasn't aware you already had a fix, this ticket was passed to SET. Changes here were the values suggested by @asaadmahmood but let me know if we should just close this in favor of your PR.

sass/layout/_headers.scss
marianunez: Do we need the `.team-sidebar` one? 

@bradjcoughlin has the `#SidebarContainer` taken care of in his PR.

bradjcoughlin:
@marianunez you increased the team sidebar z-index, which I neglected to do (dragging team icons is hidden without it). I think we can merge this first and I will resolve any conflicts in my PR.

sass/layout/_sidebar-left.scss
```diff
@@ -29,7 +29,7 @@
     position: fixed;
     left: 0;
     height: 100%;
-    z-index: 12;
+    z-index: 16;
```
bradjcoughlin, line 32: We need to include the legacy sidebar as well https://github.com/mattermost/mattermost-webapp/pull/5622/files#diff-d921a34eb9bcfe6512043c12cc85fa21L11

marianunez:
> @marianunez you increased the team sidebar z-index, which I neglected to do (dragging team icons is hidden without it). I think we can merge this first and I will resolve any conflicts in my PR.

Sounds good! Also, this will be cherry-picked for v5.24.

sass/layout/_sidebar-left.scss
bradjcoughlin: I can use the same `z-index` as it displayed alternately.

sass/layout/_headers.scss
bradjcoughlin: @marianunez we do, so that team icons are above the sidebar when dragging.

sass/layout/_sidebar-left.scss
marianunez: Done!

PR #5647, opened 2020-06-03
[MM-25757] - Add latest UI requirements to search hint popover

nevyangelova:
#### Summary
Latest acceptance criteria from UX regarding the hint popover. Issues seem to be fixed for me On desktop (latest version) and webapp (chrome and firefox) 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25757


matthewbirtch:
Hey @nevyangelova does this resove this issue?

- If I click the `@` mentions icon it opens up the RHS with the title Recent Mentions. If I then clear the search field the content of the RHS remains the same (which is fine), but the title of the RHS changes to 'Search Results'. I think it's fine to leave the mentions content and title there until you actually enter a new search term. We shouldn't change the title to 'Search Results' until you actually press enter on a new search term

nevyangelova:
@matthewbirtch I will submit a different PR for that point since the Empty search results states didn't get into v5.24 and was pushed to v5.26 instead. You can check also in https://community-release.mattermost.com/ how the popover will look with the old search states.

jgilliam17:
Thank you @nevyangelova
One suggestion 
Now that hint appears for the 3rd modifier, we should probably expand 1. point from the linked issue to include it

"When I try to add second **or third** modifier in the search input, the hint shouldn't go away until I fully type from: (including the colon) or any other modifier"

nevyangelova:
@jgilliam17 thank you for the observation, will do :)

matthewbirtch:
Good call @jgilliam17. @nevyangelova I am still seeing that weird bug where sometimes I can't click the modifier in the search hint. NOTE: I clicked the from modifier here 3 times, but this is not the only case I'm experiencing this issue.
PR-5647-Search-Hint-Bug

nevyangelova:
@matthewbirtch this happens on desktop app latest version? I tried on both MacOS and Windows and just can't reproduce this. I suspect it might have to do with one module 'electron-context-menu' that handles click events. There is a PR that upgrades that package https://github.com/mattermost/desktop/pull/1313 and should handle this better.

jgilliam17:
I am not able to reproduce modifier needing more than one click. Not reproducing on the latest desktop v4.5-rc1 for both macOS and Windows.

matthewbirtch:
> this happens on desktop app latest version?

Apologies for the lack of detail, I experienced this in Chrome in the Webapp actually - not the desktop app. I'll try restarting my machine or something and see if this gets resolved, but I seem to be reproducing consistently.

jgilliam17:
Thank you @nevyangelova 
I'll wait for others to review before adding my approval, just in case I have to retest any changes.
Looks great so far!
1. Second and third modifiers can be added fully, including the colon, by typing & without hints going away too early. 
2. Not able to repro needing multi click to select modifier on Chrome, Firefox, Microsoft Edge and desktop app (mac&win)  Similar issue found on Safari has been fixed on Catalina, but on Mojave it still occurs intermittently, per discussion with Nevy - due to Safari having known problems with click handlers
3. Search hints can now come up and be added as a 3rd modifier
4. When RHS is closed, hint popover appears after search field is done expanding
5. When clearing search text input, field no longer tries to collapse and expand, x clears the field without trying to collapse it.

components/quick_input.jsx
```diff
@@ -141,7 +143,7 @@ export default class QuickInput extends React.PureComponent {
             {clearable && value && this.props.onClear &&
                 <div
                     className='input-clear visible'
-                    onClick={this.onClear}
+                    onMouseDown={this.onClear}
```
hmhealey, line 146: 0/5, but maybe `onMouseUp` would better match a click here?

components/search_hint/search_hint.tsx
```diff
@@ -46,7 +46,7 @@ const SearchHint = (props: Props) => {
                     <li
                         className={classNames('search-hint__suggestions-list__option', {highlighted: optionIndex === props.highlightedIndex})}
                         key={option.searchTerm}
-                        onClick={() => props.onOptionSelected(option.searchTerm)}
+                        onMouseDown={() => props.onOptionSelected(option.searchTerm)}
```
hmhealey, line 49: Same here

components/quick_input.jsx
nevyangelova: Not in this case because the blur action is tied to mouseDown, I tried both :P

components/search_hint/search_hint.tsx
nevyangelova: Safari comprehends this handler better as well it seems 🤷‍♀️

PR #5649, opened 2020-06-03
MM-24754 - Incident icon on the header closes RHS if clicked while viewing reply thread

ali-farooq0:
#### Summary
* Clearing out `pluggableId` state on `rhsReducer` when `SELECT_POST_CARD` or `SELECT_POST` dispatched.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-24754

#### Screenshots
Jun-03-2020 15-53-20

### DIFF
--- a/reducers/views/rhs.js
+++ b/reducers/views/rhs.js
@@ -115,6 +115,8 @@ function searchTerms(state = '', action) {
 function pluggableId(state = '', action) {
     switch (action.type) {
     case ActionTypes.UPDATE_RHS_STATE:
+    case ActionTypes.SELECT_POST:
+    case ActionTypes.SELECT_POST_CARD:
         if (action.state === RHSStates.PLUGIN) {
             return action.pluggableId;
         }
--- a/reducers/views/rhs.test.js
+++ b/reducers/views/rhs.test.js
@@ -107,6 +107,50 @@ describe('Reducers.RHS', () => {
         });
     });
 
+    test(`should wipe pluggableId on ${ActionTypes.SELECT_POST}`, () => {
+        const nextState = rhsReducer(
+            {
+                pluggableId: 'pluggableId',
+            },
+            {
+                type: ActionTypes.SELECT_POST,
+                postId: '123',
+                channelId: '321',
+                timestamp: 1234,
+            },
+        );
+
+        expect(nextState).toEqual({
+            ...initialState,
+            pluggableId: '',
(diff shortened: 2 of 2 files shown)


reducers/views/rhs.js
```diff
@@ -115,6 +115,8 @@ function searchTerms(state = '', action) {
 function pluggableId(state = '', action) {
     switch (action.type) {
     case ActionTypes.UPDATE_RHS_STATE:
+    case ActionTypes.SELECT_POST:
+    case ActionTypes.SELECT_POST_CARD:
```
marianunez, line 119: I'm surprised there is no value for thread in the `RHSState`. It should be ideal that the state would be updated to that when it is opened to reply. Not loving the mix of "unrelated" `SELECT_POST` and `SELECT_POST_CARD` actions here.

ali-farooq0: Agree, `SELECT_POST` and `SELECT_POST_CARD` should be moved under `RHSStates`. Unfortunately that requires a bigger refactor, not worth it for this bug. Have to see if it is how it is for legacy reasons or some other reason, if there is no good reason we should just create a HW ticket to clean it up.

@hmhealey might have more context on this.

reducers/views/rhs.js
```diff
@@ -115,6 +115,8 @@ function searchTerms(state = '', action) {
 function pluggableId(state = '', action) {
     switch (action.type) {
     case ActionTypes.UPDATE_RHS_STATE:
+    case ActionTypes.SELECT_POST:
```
hmhealey, line 118: Should these be a separate case in the switch statement? I don't think `SELECT_POST` or `SELECT_POST_CARD` pass a `state` field at all.

hmhealey:
Sorry for the delay. Not sure how this PR got dropped since I've done reviews a couple times since then.

reducers/views/rhs.js
ali-farooq0: It was mainly to reset the state back to `''` so `action.state` wasn't being used in those cases. But I think it's cleaner to give it its own case.

hmhealey: Yeah, I like that better. The previous way worked mostly because action happens to look different in those two cases, but I do agree that it's cleaner to separate them entirely.

ali-farooq0:
@prapti this is ready for QA.

PR #5658, opened 2020-06-04
Fix test: Invite Member via Email containing upper case letters

josephbaylon:
#### Summary
- only fixes the case ` Invite Member via Email containing upper case letters`

#### Ticket Link
NA


srkgupta:
One of the previous test was failing and an issue is raised to track this:
https://mattermost.atlassian.net/browse/MM-25816

PR #5662, opened 2020-06-05
[MM-25771] Dont show deleted profiles on team and channel members block

fmunshi:
#### Summary
- The team members block in the system console displays deleted users since the getProfiles selectors return deleted users by default.
- This applies the fix by creating a new filter at the redux level instead of filtering the users in the component that was how the team members modal does it on the app side does it but I felt that was not optimal. I think eventually we should add a flag to the backend to exclude deleted users too but for now this will suffice.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25771

#### Related PRs
- Has redux:

### DIFF
--- a/components/admin_console/team_channel_settings/channel/details/channel_members/index.ts
+++ b/components/admin_console/team_channel_settings/channel/details/channel_members/index.ts
@@ -66,15 +66,16 @@ function makeMapStateToProps() {
             member_count: 0,
             channel_id: channelId,
             pinnedpost_count: 0,
+            guest_count: 0,
         };
 
         const searchTerm = state.views.search.modalSearch;
         let users = [];
         if (searchTerm) {
-            users = searchProfilesInChannel(state, channelId, searchTerm);
+            users = searchProfilesInChannel(state, channelId, searchTerm, {skipInactive: false});
             usersToAdd = searchUsersToAdd(usersToAdd, searchTerm);
         } else {
-            users = doGetProfilesInChannel(state, channelId, false);
+            users = doGetProfilesInChannel(state, channelId, true);
         }
 
         return {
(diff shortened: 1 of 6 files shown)


hahmadia:
Maybe we want to add `Cherrypick Approved` if this is meant for 5.24?

components/admin_console/team_channel_settings/channel/details/channel_members/index.ts
```diff
@@ -66,15 +66,16 @@ function makeMapStateToProps() {
...
             channel_id: channelId,
             pinnedpost_count: 0,
+            guest_count: 0,
         };
 
         const searchTerm = state.views.search.modalSearch;
         let users = [];
         if (searchTerm) {
-            users = searchProfilesInChannel(state, channelId, searchTerm);
+            users = searchProfilesInChannel(state, channelId, searchTerm, false, true)
```
hahmadia, line 75:
```suggestion
            users = searchProfilesInChannel(state, channelId, searchTerm, false, true);
```

fmunshi:
> Code looks good but could we create another ticket (enhancement ticket?) that says we should remove this filter in favor of doing the filtering in the backend. I feel this could lead to performance hits and that this type of filtering is better if done in the backend. I understand it's meant as a quick fix for 5.24 but I feel it would make sense if we create another ticket since we should add a flag to exclude/include deleted members as you suggested.

Will always need to have client side filtering since you cannot guarantee the contents of the profiles state to not have deleted users due to interaction with other components.

Server side filtering is a plus but not necessarily required since client side filtering will have to be done either way

hmhealey:
Looks good, but https://github.com/mattermost/mattermost-redux/pull/1169 needs to be merged first so that the redux commit can be updated.

fmunshi:
> Hi @fmunshi
> I observed the following issue and I am not sure if its because of the changes in this PR (or) an existing issue unrelated to this PR. Can you please confirm it once.
> 
> https://mattermost.atlassian.net/browse/MM-25860
> 
> If it's not related to this PR, then we can go ahead and merge this PR and still provide a separate PR for the above ticket.

The issue is unrelated to this PR - however i think I should still fix it for v5.24 in a different PR

PR #5667, opened 2020-06-05
[MM-25828] Fix incorrect globe / lock icon for private and public channels

fmunshi:
#### Summary
- Fixes a small bug rendering the wrong icon introduced in https://github.com/mattermost/mattermost-webapp/pull/5264

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25828

Screen Shot 2020-06-05 at 11 37 53 AM

### DIFF
--- a/components/channel_selector_modal/channel_selector_modal.jsx
+++ b/components/channel_selector_modal/channel_selector_modal.jsx
@@ -157,10 +157,10 @@ export default class ChannelSelectorModal extends React.PureComponent {
                 <div
                     className='more-modal__details'
                 >
-                    {option.type === 'P' &&
-                        <i className='icon icon-globe'/>}
-                    {option.type === 'O' &&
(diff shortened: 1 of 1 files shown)


fmunshi:
Gonna merge this with just one dev review since its such a small change..

asaadmahmood:
@fmunshi I was thinking of suggesting some UX improvements here as the spacing between the icons and the text is consistent compared to other modals and it looks like the hover effect is also quite different.

fmunshi:
> @fmunshi I was thinking of suggesting some UX improvements here as the spacing between the icons and the text is consistent compared to other modals and it looks like the hover effect is also quite different.

@asaadmahmood Could you make a ticket for that and assign it to enterprise? since this was marked for 5.24 I wouldnt want to make changes that are not absolutely necessary to prevent potential regressions

PR #5671, opened 2020-06-05
[MM-24301] Added translation strings for new Direct Messages and Channels category headers

devinbinnie:
#### Summary
We changed a couple of the category headers with the new sidebar, so this PR copies over the translations that we have such that they are in place for the new sidebar.

We are missing "Favorites" in most languages, so that will need to be translated.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24301


jasonblais:
@devinbinnie Thanks! For future reference, you only need to make changes to en.json, the rest will be handled by our localization process.

hmhealey:
These changes look good, but they category types use `C`, `D`, and `F` as the internal values for the types, so that's mismatched. I submitted a PR to change the server to match this.

PR #5675, opened 2020-06-06
[MM-24518] Webapp: update 'More unreads' banner in sidebar to match new mobile UI

jakubnovak998:
#### Summary

Update the UI of the 'More Unreads' banner in the Channel Sidebar to match the mobile styling.

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/14476

#### Screenshots

Zrzut ekranu z 2020-06-06 10-42-07


jakubnovak998:
I wanted to change the translation of sidebar.unreads in Mattermost Translation Server to match Figma Design but I didn't have access to it.

hanzei:
@jakubnovak998 Thanks for the PR! I've queued it for review. Heads up that you need to update the snapshots for the build to pass: https://app.circleci.com/pipelines/github/mattermost/mattermost-webapp/18638/workflows/ff4a579c-927a-43ab-99b2-723cdec0aa7e/jobs/117564/parallel-runs/0/steps/0-103

jakubnovak998:
> 2. Can we ensure there is an equal 16px of margin all the way around. It looks slightly unevenly spaced
Which element ? Banner ?

matthewbirtch:
@jakubnovak998 yes, sorry for the lack of clarity. the uneven space is around the banner. Screenshot with pixel measurements below

image

sass/layout/_legacy-sidebar-left.scss
```diff
@@ -175,12 +175,15 @@
         justify-content: center;
         align-items: center;
         position: absolute;
-        margin-left: 50%;
+        margin-left: calc(50%);
```
hmhealey, line 178: What's the reason for having `calc(50%)` here instead of just `50%`? I'm not that in depth with CSS, so I'm not sure if there's something substantially different between the two.

hmhealey:
Also, could you please take a look at the CLA [linked above](https://github.com/mattermost/mattermost-webapp/pull/5675#issuecomment-644970190)? We're not able to accept submissions without having the CLA signed.

sass/layout/_legacy-sidebar-left.scss
jakubnovak998: Earlier Matt asked for a little cheat in horizontal margin so I used calc but now it's not necessary. I noticed it later. I read on https://docs.mattermost.com/developer/localization.html that translations are updated from mattermost Translation server so change of the string should be probably there. I don't have access to it. I let know about it in the first comment.

hmhealey:
Thanks for signing that and making those changes! This looks good now, but there's a couple snapshot tests failing now because of the case change. If you run `npm test -- -u`, that will update the snapshots.

jgilliam17:
Thank you @jakubnovak998 
Banner style looks good.
When `more unreads` text takes 2 lines, banner looks a bit crowded. @matthewbirtch is this acceptable?
Screen Shot 2020-07-15 at 5 01 50 PM
Screen Shot 2020-07-15 at 5 00 56 PM

matthewbirtch:
Can we do something like this when it wraps on multiple lines?
image

Here's the link to the Figma file: https://www.figma.com/file/hUsPd4kQn3cSCwcw7eaJDL/MM-24518-More-Unreads-Banner-Update?node-id=107%3A2284

sass/layout/_legacy-sidebar-left.scss
```diff
@@ -177,10 +177,13 @@
         position: absolute;
         margin-left: 50%;
         width: 188px;
+        height: 32px;
```
matthewbirtch, line 180: change to `height: auto` so this automatically grows for longer translations

matthewbirtch, line 180: add `padding: 8px;` to give the button enough padding when translation is longer

sass/layout/_legacy-sidebar-left.scss
```diff
@@ -191,7 +194,7 @@
             display: inline-flex;
             justify-content: center;
             align-items: center;
-            margin-left: 5px;
+            margin-right: 4px;
```
matthewbirtch, line 197: change to `margin-right: 8px;`

matthewbirtch:
@jakubnovak998 I commented on the lines that need to change to accommodate for the translation and the button growing in size. In addition to that, we will want to adjust the arrow svg as follows:

```
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
<path d="M8.696 13.88H7.184V4.88L3.062 9.002L2 7.94L7.94 2L13.88 7.94L12.818 9.002L8.696 4.88V13.88Z"></path>
</svg>
```

jgilliam17:
Thank you @jakubnovak998, banner looks good with longer text.
@matthewbirtch one small issue - arrow placement. Should the arrow keep the same margin on the left or adjust the placement based on the length of the text?
Screen Shot 2020-07-28 at 8 19 26 AM
Screen Shot 2020-07-28 at 8 19 05 AM

matthewbirtch:
@jakubnovak998 it looks like you've got the padding right, but the arrow icons need to be 16x16 instead of 24x24. It's fine if you want to rotate the arrow instead, but the height and width are important. Otherwise it makes this button too tall. The height of the button should be 32px tall in total (when there is only 1 line of text).

As for your other question, it's ok to position the arrow where it is with longer 2-line text

hmhealey:
Looks like we're all good to merge for 5.28 then. Thanks again for the PR, @jakubnovak998!

PR #5636, opened 2020-06-01
MM-24467 Add SAML ServiceProviderIdentifier

sbishel:
#### Summary
The tag in the AuthnRequest is being created incorrectly. This PR uses a new configuration setting to set the <. If the new configuration setting is empty, then the current logic is used. This will allow existing installations to continue to work.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24467

#### Related Pull Requests
https://github.com/mattermost/enterprise/pull/674
https://github.com/mattermost/mattermost-server/pull/14725

#### Screenshots
Screen Shot 2020-06-01 at 1 54 51 PM

PR #5683, opened 2020-06-08
MM-24467 Add SAML ServiceProviderIdentifier

sbishel:
#### Summary
The tag in the AuthnRequest is being created incorrectly. This PR uses a new configuration setting to set the <. If the new configuration setting is empty, then the current logic is used. This will allow existing installations to continue to work.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24467

#### Related Pull Requests
https://github.com/mattermost/enterprise/pull/674
https://github.com/mattermost/mattermost-server/pull/14725

#### Screenshots
Screen Shot 2020-06-19 at 3 01 06 PM


sbishel:
Opened on the repository rather than a fork.
https://github.com/mattermost/mattermost-webapp/pull/5636

sbishel:
Regarding #2. I think this field should be above `Verify Signature:`, since it is part of base setup. (In fact, I think `Service Provider Login URL:` should be above as well.) I can remove the text that mentions `Service Provider Login URL:` if that would make it flow better. 

Regarding #3. It uses `Identity Provider Issuer URL` by default for backwards compatibility reasons. But for a new installation we want them to be setting it to Service Provider Login Url.  May-be we should put a larger more thorough helper text explaining both.

sbishel:
Worked with Michael. Decision was to bring the field down below "Service Provider Login URL" as suggested by Rohitesh.

PR #5685, opened 2020-06-08
npminstall: skip if up-to-date

lieut-data:
#### Summary
Tweak the `make node_modules` directive to skip the expensive `npm install` step if `package.json` and `package-lock.json` haven't changed since the last such invocation. 

This is accomplished using an `.npminstall` file (already part of `.gitignore`) that is touched after a successful `make node_modules`. If the file is missing or older than either `package.json` or `package-lock.json`, `npm install` proceeds normally. Otherwise, the unnecessary `npm install` step is skipped.

Credit to @crspeller for the idea here while iterating on the incident response plugin.

Note that the experience here would be so much better if `npm install` would stop rewriting `package-lock.json` all the time, but I understand we're iterating on that separately.

#### Ticket Link
N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,10 +36,15 @@ test: node_modules ## Runs tests
 i18n-extract: ## Extract strings for translation from the source code
 	npm run mmjstool -- i18n extract-webapp
 
-node_modules: package.json package-lock.json
+.PHONY: node_modules
+node_modules: .npminstall
+
+.npminstall: package.json package-lock.json
 	@echo Getting dependencies using npm
 
 	npm install
+	touch $@
+
 
 package: build ## Packages app
 	@echo Packaging webapp


lieut-data:
Skipping QA review since this affects developer tooling only.

Makefile
```diff
@@ -36,10 +36,15 @@ test: node_modules ## Runs tests
...
-node_modules: package.json package-lock.json
+.PHONY: node_modules
+node_modules: .npminstall
+
+.npminstall: package.json package-lock.json
 	@echo Getting dependencies using npm
 
 	npm install
+	touch $@
+
```
lieut-data, line 47:
```suggestion
```

hmhealey:
Assuming I'm understanding Makefiles correctly, why do we need to declare the `node_modules` target as phony? I thought that only had to be done when the `node_modules` target doesn't actually produce a file named `node_modules`?

We used to use a `.npminstall` file, but I removed that a while back since it seems like it shouldn't be necessary.

hmhealey:
I found out what's wrong with what we currently have: https://stackoverflow.com/questions/44995853/a-makefile-recipe-runs-every-time-even-when-target-is-more-recent-than-dependenc

Basically, Makefiles test whether or not something needs to be rebuilt is based on timestamps of the dependencies and the target (package.json/package-lock.json and node_modules respectively here), BUT timestamps of a folder aren't updated if a child of a child changes, so if package.json changes after an `npm install`, package.json will always have a newer timestamp than node_modules.

So the fix is to update the timestamp of `node_modules` whenever `npm install` happens. This works locally:

```makefile
node_modules: package.json package-lock.json
    npm install
    touch node_modules
```
This works as intended now:
```
harrison-mm2:mattermost-webapp harrison$ touch package.json
harrison-mm2:mattermost-webapp harrison$ make node_modules
Getting dependencies using npm
npm install

> mattermost-webapp@0.0.1 postinstall /Users/harrison/go/src/github.com/mattermost/mattermost-webapp
# and so on
harrison-mm2:mattermost-webapp harrison$ make node_modules
make: `node_modules' is up to date.
```

PR #5688, opened 2020-06-08
translations PR 20200608

enahum:
Ideally this PR should be review by community members for each language.

If you as a core committer are requested a review, please confirm there are no changes that would cause build or unit test failures, and merge the PR.

Notes:
1. Further changes to translations should be made to the [Mattermost Translation Server(https://translate.mattermost.com/projects/mattermost). These will be included in next week's PR.
2. If you see an English translation instead of the desired language, then this PR was submitted before the translation was made on the [Mattermost Translation Server(https://translate.mattermost.com/projects/mattermost). New translations will be included in next week's PR.


jasonblais:
@metanerd Tests are failing for this PR

fmunshi:
Looks like the tests check the order of search results in `ES` and since the translations have changed for `ES` the order of the search results has changed... @enahum 
Could you just modify lines 52-60 of `admin_console_index.test.jsx` to
```
        expect(idx.search('ldap')).toEqual([
            'environment/session_lengths',
            'authentication/mfa',
            'authentication/ldap',
            'authentication/saml',
            'experimental/features',
            'authentication/email',
            'authentication/discover-ldap',
            'authentication/guest_access',
        ]);
``` 

That should fix the build...

enahum:
@fmunshi I can modify it yes, but are other languages going to be affected? why is this causing an issue ?

fmunshi:
@enahum looks like the spanish translations for session length changed to include the text LDAP
`"admin.service.webSessionDaysDesc.extendLength": "El número de días desde la última vez que un usuario tuvo actividad en Mattermost para que la sesión del usuario creada por autenticación de correo electrónico y AD/LDAP expire. Luego de cambiar esta configuración, la nueva duración de la sesión tendrá efecto luego de la próxima vez que el usuario ingrese sus credenciales.",`

So the test which checks search results for LDAP in spanish now returns session lengths as one of the first results - I opened a PR that ignores the order of languages other than english so we could merge that in _or_ we could make the changes here, but then this test could still have the potential to fail again in the future if the spanish translations for any of these changes again.

enahum:
@fmunshi I think the code should be fixed and it should be completely agnostic of the translation and this PR should be about translations only

jasonblais:
@enahum @fmunshi If it's only related to the order of the strings, this shouldn't impact the user as it's just a stylistic error affecting code? If so, could we merge this to 5.24 and fix the ordering on master?

enahum:
@jasonblais I think is only making the test to fail, so I do not see why it can’t be merged.. but if we can fix the tests that would be a good win.

PR #5689, opened 2020-06-09
Adding label for PR smoke tests

prapti:
Adding `@smoke` labels for tests to be considered in PR smoke test.
This PR covers smoke tests in the following E2E test files. Some test files already had the `@smoke` label and thus cover the smoke test criteria as such. Only some that needed according to the `Smoke Tests` doc have been added.
- User can create a new team -- `create_a_team_spec`
- User can create a channel -- `accessibility_sidebar_spec`
- Users can join an existing channel -- `add_users_to_channel_spec`
- User can post in a channel -- `messsge_spec`
- Basic search function -- `search_user_post`
- System administrator can access system console -- `ui_and_api > cluster_spec`
- User can install a plugin -- `marketplace_spec`
- User can log out -- `tutorial_navigation_and_links_spec`

### DIFF
--- a/e2e/cypress/integration/messaging/message_spec.js
+++ b/e2e/cypress/integration/messaging/message_spec.js
@@ -7,7 +7,7 @@
 // - Use element ID when selecting an element. Create one if none.
 // ***************************************************************
 
-// Stage: @prod
+// Stage: @prod @smoke
 // Group: @messaging
 
 import users from '../../fixtures/users.json';
--- a/e2e/cypress/integration/plugins/marketplace/marketplace_spec.js
+++ b/e2e/cypress/integration/plugins/marketplace/marketplace_spec.js
@@ -7,6 +7,7 @@
 // - Use element ID when selecting an element. Create one if none.
 // ***************************************************************
 
+// Stage: @smoke
 // Group: @plugin_marketplace @plugin
 
 import * as TIMEOUTS from '../../../fixtures/timeouts';
--- a/e2e/cypress/integration/search/search_user_post_spec.js
+++ b/e2e/cypress/integration/search/search_user_post_spec.js
@@ -7,7 +7,7 @@
 // - Use element ID when selecting an element. Create one if none.
 // ***************************************************************
 
-// Stage: @prod
+// Stage: @prod @smoke
 // Group: @search
 
 import users from '../../fixtures/users.json';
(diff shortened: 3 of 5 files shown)


saturninoabril:
@prapti Looks good to me. Just few requests:
- add more details in the description like how many test cases were initially selected and the duration it takes to run the test
- can't see the "Smoke Tests" doc mentioned in the description.  Can you link it here for reference? The only reference I have is the one you posted in our [GM](https://community-release.mattermost.com/core/pl/8y6g1t5qztduznnktii5ux6z5y)
- please add follow up documentation to https://developers.mattermost.com/contribute/webapp/end-to-end-tests/ on how it's selected, what are the criteria and how to run like should it be with args `--group="@smoke"` only or in combination with prod ``--stage="@prod" --group="@smoke"``
- can you confirm that all are consistently passing? The last time I checked was `marketplace_spec` still have failing tests that need to attend to.

prapti:
Thanks for a quick check @saturninoabril! I'll add make those changes and update the PR description. Meanwhile, I've changed it to a WIP PR for now as I'm running some tests from the smoke group to gather some data upfront. I'll update the PR soon. And thanks for reminding that the marketplace tests aren't 100%. Will need to remove them now until fixed.

e2e/cypress/integration/accessibility/accessibility_sidebar_spec.js
```diff
@@ -8,7 +8,7 @@
 // ***************************************************************
 
 // Stage: @prod @smoke
-// Group: @accessibility
+// Group: @accessibility @smoke
```
saturninoabril, line 11: I'm confused how do you plan to run this. 
If `@smoke` will be on "Group" metadata which I think makes sense, then the `@smoke` tags on "Stage" can be removed. With that, it could run as `node run_tests.js --stage='@prod' --group'@smoke'`
```suggestion
// Stage: @prod
// Group: @accessibility @smoke
```

saturninoabril: Oh, I just read your updated description with `node run_tests.js --stage='@smoke' --group='@smoke'`. However, passing `@smoke` on both args feels redundant to me.

prapti:
Removed '@smoke' tag from all other specs not intended for smoke tests. Now that the tag has been added as "Group", the tests can be run simply with `--group='@smoke'` flag for running the smoke tests, or with `--stage='@prod' --group='@smoke'` to run it within the production tests.

PR #5695, opened 2020-06-09
[MM-25921] Modify admin console index tests to ignore order check in spanish

fmunshi:
#### Summary
- These tests should not check the order of search results in languages other than english since they can change if the community updates translations as seen here https://github.com/mattermost/mattermost-webapp/pull/5688

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25921


fmunshi:
@bradjcoughlin I noticed you touched this file a bunch - does this change make sense to prevent issues like the one seen here https://github.com/mattermost/mattermost-webapp/pull/5688?

fmunshi:
Merging without QA Review since this only modifies test files

PR #5702, opened 2020-06-10
MM-25841 Retain search results even when quick switcher is out of focus

sudheerDev:
#### Summary
The previous behavior of clearing search results is because when the focus is lost we remove the suggestion list. This component is used in various places like auto-complete so default behavior of removing suggestions when focus is lost still holds true. Adding a new prop to make force suggestions even when the focus is lost.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25841


sudheerDev:
> looking good! but looks like something autoformated one file that is failing the linting

Ya I was having a tough time with my local setup yesterday. Prettier screwed up the formating somehow.

PR #5426, opened 2020-05-01
migrated do_verify_email component to typescript

pradeepmurugesan:
#### Summary
migrated do_verify_email component to typescript
#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/13715

#### Related Pull Requests

- Has server changes  - NO
- Has redux changes - NO
- Has mobile changes - NO

PR #5706, opened 2020-06-11
MM-25889 - Restored announcement type for email verification

marianunez:
#### Summary
Restored `AnnouncementType` for announcement banner when email has been verified that was mistakenly removed during [typescript migration](https://github.com/mattermost/mattermost-webapp/pull/5426). 

#### Ticket Link
[MM-25889](https://mattermost.atlassian.net/browse/MM-25889)


marianunez:
@amyblais This is a fix for a minor regression from v5.23. It might be a bit late but the change is very low impact, should we cherry pick this for 5.24?

amyblais:
That should fine as long as this won't lead to new regressions.

marianunez:
> We might need to cast the first argument to `logError` as an `any`, and we'll look into fixing `logError` at a later date.

Thanks @devinbinnie! I've updated the PR, hope the linter is happy now.

amyblais:
On further discussion with Jesse we can push to v5.25 as we're past code freeze,

PR #5707, opened 2020-06-11
[MM-25658] Replace Data Grid on Channel and Teams page in System console with Common Component

hahmadia:
#### Summary

Replace the data grids on Teams and Channels page with the new common component created by the wonderful @fmunshi . 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25658

#### Screenshots

image
image


components/admin_console/team_channel_settings/team/list/team_list.jsx
```diff
@@ -28,136 +32,190 @@ export default class TeamList extends React.PureComponent {
...
+            teams = response.data.teams;
+            total = response.data.total_count;
         }
-        this.setState({searchString: e.target.value});
-    }
 
-    handleTeamSearchKeyUp = async (e) => {
-        const {key} = e;
-        const {searchString} = this.state;
+        this.setState({page, loading: false, teams, total, term});
```
fmunshi, line 69: I think we should still continue using the redux state and pass teams / total / term as props

components/admin_console/team_channel_settings/channel/list/channel_list.test.tsx
```diff
@@ -4,29 +4,43 @@
...
+            last_post_at: 123,
+            total_msg_count: 123,
+            extra_update_at: 123,
+            creator_id: 'abc',
+            scheme_id: 'abc',
+            group_constrained: false,
+            team_display_name: 'teamDisplayName',
+            team_name: 'teamName',
+            team_update_at: 1,
         }];
```
fmunshi, line 33: Could you use the `utils/test_helper` `getChannelMock` methods instead?

components/admin_console/team_channel_settings/channel/list/channel_list.tsx
```diff
@@ -2,182 +2,227 @@
...
-            if (searchString.length > 1) {
-                const response = await this.props.actions.searchAllChannels(searchString, '', false, 0, PAGE_SIZE);
-                if ('data' in response) {
-                    this.setState({searchMode: true, channels: response.data.channels, searchTotalCount: response.data.total_count, pageResetKey: Date.now()});
-                }
+        if (term.length > 0) {
+            if (page > 0) {
+                this.searchChannels(page, term);
+            } else {
+                this.searchChannelsDebounced(page, term);
```
fmunshi, line 68: what happens if someone is already on page 3 and then they start searching wouldnt this result in a non debounced search? - does the page get reset to 0 when the search term changes?

fmunshi: Looks like it does

components/admin_console/team_channel_settings/channel/list/channel_list.tsx
```diff
@@ -2,182 +2,227 @@
...
+            />
         );
-    }
 
-    private renderRow = (item: ChannelWithTeamData) => {
         return (
-            <ChannelRow
-                key={item.id}
-                channel={item}
-                onRowClick={this.onChannelClick}
```
fmunshi, line 175: in order to prevent jittery behaviour when changing pages you'll need to do something similar to what I did in `user_grid` line 267
```
        const rowsContainerStyles = {
            minHeight: `${rows.length * ROW_HEIGHT}px`,
        };
```

And then pass `rowsContainerStyles` to `DataGrid` this prevents the datagrid from shrinking down while its loading between pages i.e. it retains its previous grid height when its loading the next page

fmunshi: Kapture 2020-06-11 at 20 49 22
Heres a gif showing what I mean

components/admin_console/data_grid/data_grid.scss
```diff
@@ -83,14 +89,19 @@
...
+            
             &:nth-child(odd) {
-                background: $bg--gray;
+                background-color: $bg--gray;
             }
+
+            &:hover {
+                background-color: lighten($primary-color, 45%);
+                border-left: 2px solid $primary-color;
+            }
```
fmunshi, line 100: Adding this border 2px causes the width of the element to increase so on hover it looks like the row is shifting slightly to the right as seen below
Kapture 2020-06-12 at 14 30 48

In order to fix that you should add a transparent border of the same size to `.DataGrid_row ` above
something like 
`border-left: 2px solid transparent;` should be good

components/admin_console/data_grid/data_grid.scss
```diff
@@ -83,14 +89,19 @@
 
         .DataGrid_row {
             min-height: 50px;
```
fmunshi, line 91:
```suggestion
        .DataGrid_row {
            min-height: 50px;
            border-left: 2px solid transparent;
```

fmunshi:
Just noticed an issue - previously clicking on any team / channel row would take you to the page for that team / channel... I think we're gonna want this to behave the same way, so you're going to have to setup a way to add an onclick handler to a full datagrid row 

I think refactoring the datagrid `Row` type to be
```
export type Row = {
    cells: {
        [key: string]: JSX.Element | string;
    }
    onClick?: () => void;
}
```
this would then require changing existing implementations to work with this new type too.

components/admin_console/team_channel_settings/team/list/team_list.jsx
```diff
@@ -28,136 +29,210 @@ export default class TeamList extends React.PureComponent {
...
+    }
 
-        if (key === Constants.KeyCodes.ENTER[0]) {
-            if (searchString.length > 1) {
-                const response = await this.props.actions.searchTeams(searchString, 0, PAGE_SIZE);
-                this.setState({searchMode: true, teams: response.data.teams, searchTotalCount: response.data.total_count, pageResetKey: Date.now()});
-            }
-        }
+    searchTeams = async (page = 0, term = '') => {
+        const response = await this.props.actions.searchTeams(term, page, PAGE_SIZE);
```
mkraft, line 70: Ditto here.

components/admin_console/team_channel_settings/team/list/team_list.scss
```diff
@@ -0,0 +1,20 @@
...
+                height: 80px;
+            }
+        
+            .TeamList_ManagementText {
+                opacity: 0.65;
+            }
+        
+            .TeamList_EditText {
+                padding-right: 20px;
+            }
```
mkraft, line 17: Ditto with some of these names.

components/admin_console/team_channel_settings/team/list/team_list.jsx
hahmadia: Handled!

components/admin_console/team_channel_settings/channel/list/channel_list.tsx
```diff
@@ -2,182 +2,238 @@
...
-                <div className='group-content'>
-                    <div className='group-description'>
-                        <FormattedMessage
-                            id='admin.channel_settings.channel_list.teamHeader'
-                            defaultMessage='Team'
-                        />
-                    </div>
-                    <div className='group-description adjusted'>
+    searchChannels = async (page = 0, term = '') => {
+        const response = await this.props.actions.searchAllChannels(term, '', false, page, PAGE_SIZE);
```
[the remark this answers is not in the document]
hahmadia: Handled! Thanks for the callout.

components/admin_console/team_channel_settings/channel/list/channel_list.scss
```diff
@@ -0,0 +1,19 @@
...
+        .DataGrid_row {
+            min-height: 40px;
+        }
+        
+        .DataGrid_cell {
+            .TeamList_editRow {
+                padding-right: 20px;
+            }
+
+            .TeamList_ChannelDisplayName {
```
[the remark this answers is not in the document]
deanwhillier: Here's the slight modification. See below for a hopefully clear example explaining the naming convention.
```suggestion
            .TeamList_channelDisplayName {
```

components/admin_console/team_channel_settings/team/list/team_list.scss
deanwhillier: And here:

```css
.TeamList_EditText {
```

should be

```css
.TeamList_editText {
```

components/admin_console/team_channel_settings/channel/list/channel_list.scss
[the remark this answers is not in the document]
hahmadia: Done!

components/admin_console/team_channel_settings/team/list/team_list.scss
hahmadia: Done + Edited all other classnames to follow this pattern.

deanwhillier:
### A "Quick" explanation for css naming:

Given this example HTML:
```html
<button class="Button Button___primary">
    <i class="Icon Icon___small Button_leftIcon"/>
    <span class="Button_label">Click Me!</span>
    <i class="Icon Icon___small Button_rightIcon"/>
</button>
```

A component has three naming considerations as follows:

1. The component itself representing the **Block** in BEM
2. The components child elements representing the **Element** in BEM
3. The components modifiers representing the **Modifier** in BEM

#### Blocks:

There are technically 3 **Blocks** above – the outer button and the two inner icons. The icons are **Blocks** as they are also components that are imported into the button component. So, the **Block** class names are:

* `.Button` for the Button component
* `.Icon` for the Icon components

**Block** names use **UpperCamelCase** (also known as **PascalCase**).

Generally speaking, **Block** classes are used to visually style the component and its child elements and not define any placement styles within its parent container.

#### Elements:

There are 3 **Elements** above – the two icon **Elements** and the span tag containing the button's text. While the icons are components and therefore are **Blocks** in their own right, they are also **Elements** from the buttons perspective, so they receive **Element** class names as well:

* `.Button_leftIcon`
* `.Button_label`
* `.Button_rightIcon`

**Element** names start with the **Block** name and add the **Element** name using **lowerCamelCase** (also know as just **camelCase**) separated by a _single_ underscore. **Element** class names are only really necessary if you need to modify the **Element's** styles in the context of the parent **Block**. 

**Elements** within a parent block that are components should be styled from the **Blocks** perspective using the **Element** class name (i.e. `.Button_leftIcon`) and not the **Element's** component name (i.e.`.Icon`).

Generally speaking, **Element** classes are used to position the element within the **Block** and not strictly to apply visual styling. Though visual styling can be applied if it doesn't make sense to support the needed styling within a child component's default styles or if the **Element** doesn't necessarily warrant the creation of a separate component.

#### Modifiers:

The overall button and the two icons above have modifiers applied:

* `.Button___primary` on the button to override the generic button with primary styling
* `.Icon___small` on the two icons to override the generic icon size to make them small icons
   * Note that this modifier name is from the perspective of the icon and not specifically the button because the icon already supports the size the button needs. If you needed an icon size within the button that the icon didn't already support and doesn't make sense to add to the icon component directly, you could use an element modifier class name like `.Button_rightIcon___superExtraSmall`. This approach should be used sparingly though!

**Modifier** names start with the **Block** name and add the **Modifier** name using **lowerCamelCase** (also know as just **camelCase**) separated by _three_  underscores to help visually distinguish them from **Element** names that use a _single_ underscore.

**Modifiers** can also be used for component state changes like `.Button___toggled` when a button is toggled to apply different visual styling when that state is triggered. That said, I also personally like to use HTML attributes to indicate state changes as some html elements already have state attributes. Eg. `[disabled]` or `[data-toggled="true"]` or ...

(Been meaning to write this as a general draft doc for CSS, so will copy from here)

fmunshi:
thanks for the explanation @deanwhillier

components/admin_console/team_channel_settings/channel/list/channel_list.tsx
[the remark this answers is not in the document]
mkraft: So this is handling null data, but it's not doing anything with the error. Do the designs have anything defined for how to display server errors?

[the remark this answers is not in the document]
hahmadia: @michaelgamble 
Do you have any designs for what should happen if we are searching for channels and we get a server error? Where to display the error or how to handle the error from the server.

Thanks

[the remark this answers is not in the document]
hahmadia: @mkraft 
I've updated to show error message as per our discussions with Micahel.
image

This is what is shown when the server errors when a user is searching.

clarmso:
@hahmadia I found your PR! I'd like to follow the mattermost repo more but I've a day job. I may be peeking around from time to time. "See" you around!

hahmadia:
> @hahmadia I found your PR! I'd like to follow the mattermost repo more but I've a day job. I may be peeking around from time to time. "See" you around!

@clarmso 
Welcome!! So cool to see you around! Consider joining www.community.mattermost.com . The entire Mattermost team is active (including myself) over their  and we are always happy to help out if you ever come across issues or problems.

srkgupta:
@deanwhillier can you please review the PR & confirm if the changes are good from code perspective. Will pickup the QA once it's done. Don't want to duplicate QA efforts if there are any new changes based on your review.

hahmadia:
MM-26568 - Resolved

michaelgamble:
@srkgupta talked to @hahmadia over zoom, determined all three need real quick fixes, we talked through the css changes should be pretty quick for him :)

srkgupta:
> @srkgupta talked to @hahmadia over zoom, determined all three need real quick fixes, we talked through the css changes should be pretty quick for him :)

Thanks @michaelgamble

hahmadia:
Addressed all problems @srkgupta

PR #5711, opened 2020-06-11
[MM-24708] Added What's New modal for Channel Sidebar Phase 2 changes

devinbinnie:
#### Summary
This PR adds a "What's New" modal that will popup on first use of the new Experimental Sidebar Features.
NOTE: This currently does not include the animated GIF that will display on the modal, we are waiting on UX to provide so currently it's just a blank box.

I've also refactored the category modals and this one into a new re-usable `generic_modal` component.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24708


components/channel_layout/channel_controller.jsx
```diff
@@ -88,6 +89,7 @@ export default class ChannelController extends React.Component {
                     <EditPostModal/>
                     <ResetStatusModal/>
                     <LeavePrivateChannelModal/>
+                    <SidebarWhatsNewModal/>
```
hmhealey, line 92: Maybe we could use the `ModalController` for this modal since it only appears once?

components/delete_category_modal/delete_category_modal.tsx
```diff
@@ -32,81 +32,40 @@ export default class DeleteCategoryModal extends React.PureComponent<Props, Stat
...
-                aria-labelledby='editCategoryModalLabel'
+            <GenericModal
+                onHide={this.props.onHide}
+                modalHeaderText={(
+                    <FormattedMessage
+                        id='delete_category_modal.deleteCategory'
+                        defaultMessage='Delete this category?'
+                    />
+                )}
+                handleCancel={() => {}}
```
hmhealey, line 49: This is going to cause the `GenericModal` to re-render whenever this does. Can we just omit this prop? If we need to null check it in the component, we can also use the new `this.props.handleCancel?.()` syntax

components/generic_modal.scss
```diff
@@ -0,0 +1,122 @@
+@charset 'UTF-8';
+
+.modal {
```
hmhealey, line 3: Is this extra layer needed or can we start with `.GenericModal`? I think `.modal` is part of react-bootstrap, so whenever we get rid of that, this would need to go away anyway.

components/generic_modal.scss
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        .modal-header {
+            background: transparent;
+            border: none;
+            min-height: 0;
+            padding: 0;
+
+            .close {
+                color: var(--center-channel-color-56);
```
hmhealey, line 45: Would you mind switching these transparent ones to the `-rgb` variants while you're in here?

components/generic_modal.tsx
```diff
@@ -0,0 +1,151 @@
...
+        if (this.props.handleConfirm) {
+            let confirmButtonText = (
+                <FormattedMessage
+                    id='generic_modal.confirm'
+                    defaultMessage='Confirm'
+                />
+            );
+            if (this.props.confirmButtonText) {
+                confirmButtonText = (
+                    <React.Fragment>
```
hmhealey, line 71: Are these fragments necessary or can we just do `confirmButtonText = this.props.confirmButtonText`?

components/sidebar/sidebar_whats_new_modal/index.ts
```diff
@@ -0,0 +1,40 @@
...
+import {get as getPreference, getNewSidebarPreference} from 'mattermost-redux/selectors/entities/preferences';
+import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
+
+import {GlobalState} from 'types/store';
+
+import SidebarWhatsNewModal from './sidebar_whats_new_modal';
+
+function mapStateToProps(state: GlobalState) {
+    const currentUserId = getCurrentUserId(state);
+    const hasSeenModal = getPreference(
```
hmhealey, line 18: There's a `getBool` selector that you can use here to skip the type conversion

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.scss
```diff
@@ -0,0 +1,31 @@
...
+    .SidebarWhatsNewModal__helpText {
+        margin-top: 8px;
+    }
+
+    .SidebarWhatsNewModal__img {
+        margin-top: 40px;
+        height: 360px;
+        border: 1px solid var(--center-channel-color-16);
+        border-radius: 4px;
+        min-height: 360px;  // TODO: remove
```
hmhealey, line 21: Is this something you meant to remove with this PR or later?

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+            <GenericModal
+                show={!this.props.hasSeenModal && this.state.isUsingNewSidebar}
+                onHide={this.onHide}
+                modalHeaderText={(
+                    <FormattedMessage
+                        id={'sidebar_whats_new_modal.header'}
+                        defaultMessage={'What\'s new'}
+                    />
+                )}
+                handleConfirm={() => {}}
```
hmhealey, line 56: This should also be omitted if possible.

hmhealey:
Yeah, we decided that would be acceptable for the time being since this is a one-off feature that'll be replaced eventually.

esethna:
Adding the do not merge label though, until we can get the updated GIF

components/generic_modal.tsx
nevyangelova: I also wanted to ask about the reasoning behind using the React.Fragment, the common use case is for many children rendering, but seems like this is only one button element?

components/generic_modal.tsx
```diff
@@ -0,0 +1,151 @@
...
+            let cancelButtonText = (
+                <FormattedMessage
+                    id='generic_modal.cancel'
+                    defaultMessage='Cancel'
+                />
+            );
+            if (this.props.cancelButtonText) {
+                cancelButtonText = (
+                    <React.Fragment>
+                        {this.props.cancelButtonText}
```
nevyangelova, line 102: Here as well

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.scss
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    .modal-footer {
+        margin-top: 20px;
+    }
+}
+
+.modal .GenericModal.modal-dialog {
+    margin-top: calc(50vh - 307px);
+}
```
nevyangelova, line 31: nit: new line

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.tsx
nevyangelova: Can we make this prop optional?

esethna:
@hmhealey is this PR going to get reopened against master?

components/generic_modal.tsx
devinbinnie: Apparently not, fixed.

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.scss
devinbinnie: I'm going to remove it once the GIF is actually in there, it's there to make sure the size of the modal is still the same.

components/generic_modal.tsx
devinbinnie: Since I allow `confirmButtonText` to be a `JSX.Element` or a `string`, it needs to be wrapped this way or TS will complain.

PR #5712, opened 2020-06-12
MM-25904 - Updating link tooltips z-index for plugins

asaadmahmood:
#### Summary
MM-25904 - Updating link tooltips z-index for plugins

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25904

#### Screenshots
Screenshot 2020-06-12 at 12 54 52 PM

### DIFF
--- a/components/link_tooltip/link_tooltip.tsx
+++ b/components/link_tooltip/link_tooltip.tsx
@@ -13,6 +13,9 @@ const tooltipContainerStyles: CSSProperties = {
     flexDirection: 'column',
     alignItems: 'center',
     zIndex: 1070,
+    position: 'absolute',
+    top: -1000,
+    left: -1000
 };
 
 type Props = {
--- a/components/link_tooltip/link_tooltip.tsx
+++ b/components/link_tooltip/link_tooltip.tsx
@@ -12,7 +12,7 @@ const tooltipContainerStyles: CSSProperties = {
     display: 'flex',
     flexDirection: 'column',
     alignItems: 'center',
-    zIndex: 10,
+    zIndex: 1070,
 };
 
 type Props = {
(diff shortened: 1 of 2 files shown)


hanzei:
I would love to see an E2e test for plugin tooltips.

asaadmahmood:
@hanzei Done.

DHaussermann:
@asaadmahmood I'm seeing an issue where previews in the pop-over are often inaccurate but only in the RHS. It seems like clearing the cache resolves this when I see it happen. Do you have any thoughts on if this is an issue with Webapp or one that should go in the plugin repo?
cc @hanzei 
InvalidPreviews

DHaussermann:
@asaadmahmood I missed an issue earlier. This seems to have introduced a bug where navigating to the admin side and returning to the users side will display the tooltip at the top center of the app and push the rest of the contents down. User must then refresh to resolve the issue.

I deployed this on a different cloud server and it does seem to be this PR that introduces the behavior. 
Steps:
- Install and configure GitHub
- Post some GitHub links that generate a preview
- Open RHS and render the preview there
- Open system console
- Return to user side by selecting your team      
**Observed**: 
GH-Preview

asaadmahmood:
@DHaussermann I see that system console issue sometimes but it automatically fixed itself for me, but that is not specific to this PR. I think we can open up another issue on that.

asaadmahmood:
Should be fixed now @DHaussermann.

DHaussermann:
Huge thanks @asaadmahmood I can no longer repro the issue where the pop-over pushes down the rest of the content when returning from admin side. 

Other testing details about the pop-over being covered are posted above
LGTM!

- Re-requested 1 code review for the last commit. 
- Please advise if https://github.com/mattermost/mattermost-webapp/pull/5712#issuecomment-644950328 requires an fix to webapp or the GitHub plugin its self

components/link_tooltip/link_tooltip.tsx
```diff
@@ -12,7 +12,10 @@ const tooltipContainerStyles: CSSProperties = {
     display: 'flex',
     flexDirection: 'column',
     alignItems: 'center',
-    zIndex: 10,
+    zIndex: 1070,
```
ali-farooq0, line 15: We should have a standardized zIndex value for all these zIndex layers :P.

asaadmahmood: Hoshiyari nai ziada :p 
This is basically the z-index our bootstrap tooltips use right now.

PR #5715, opened 2020-06-12
Add call button

larkox:
#### Summary
At the moment we have several plugins for videoconferencing. It would be great if they were all bundled in a single dropdown separated from other plugins, for instances were they have more than one videoconferencing option configured.

There are two options for the location: channel header or post action

In the channel header, the current version shows the plugin icon (as it was being shown before) if there is only one videoconferencing plugin installed. If there is more than one, a dropdown (as the one for other plugins) will show.

In the post action, the current version shows the plugin icon if there is only one videoconferencing plugin installed. If there is more than one, a dropdown will show.

I also created a new registry for these plugins, different from registerChannelHeaderButton. Depending on the final option we take, I will update the name.

I am missing a proper icon to show at the call button. Probably a phone icon or a camera icon. Position is also up to discussion.

#### Ticket Link
None

#### Related Pull Requests
None

#### Screenshots
Inspiration:
Screenshot from 2020-06-12 15-48-11

Current implementation (Option A):
Screenshot from 2020-06-12 15-50-05

Current implementation (Option B):
- WebApp desktop view
Screenshot from 2020-07-07 14-41-01
Screenshot from 2020-07-07 14-59-26
Screenshot from 2020-07-07 14-59-16
- WebApp Mobile view
Screenshot from 2020-07-07 14-41-18
Screenshot from 2020-07-07 14-58-45
Screenshot from 2020-07-07 14-58-51


jespino:
How common is to have multiple videoconference on from our clients perspective? maybe @aaronrothschild have more knowledge about that.

aaronrothschild:
Awesome stuff @larkox Multiple meeting providers sometimes happen, and this would help in that scenario.

But, the better UX prblem this solves is providing a *consistent* way to train all users "this is how you can easily start a video meeting" Click the "Call button in whatever channel you are in".  Even if there is 5 plugins installed, the call button should always be visible as a button individually.  If there are multiple call providers (Zoom and Webex say) - then users can choose from there - rather than a cluttered list of plugins. 

Many people "expect" MM to have a "calling feature" built in...this is a way to make it more obvious to the end user where they can start a call from.  

@esethna and @michaelgamble @andrewbrown00 Thoughts on this sort of functionality for plugins?

esethna:
> Even if there is 5 plugins installed, the call button should always be visible as a button individually. If there are multiple call providers (Zoom and Webex say) - then users can choose from there - rather than a cluttered list of plugins.

Like the idea, and totally agree with Aaron on this ^. This could be useful if there are multiple call options and they open in a pop-up when the video icon is clicked

plugins/channel_header_call/channel_header_call.jsx
```diff
@@ -0,0 +1,216 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+/* eslint-disable react/no-multi-comp */
```
hmhealey, line 4: Can you split this into multiple files for consistency?

plugins/channel_header_call/channel_header_call.jsx
```diff
@@ -0,0 +1,216 @@
...
+import {Dropdown, Tooltip} from 'react-bootstrap';
+import {RootCloseWrapper} from 'react-overlays';
+import {FormattedMessage} from 'react-intl';
+
+import HeaderIconWrapper from 'components/channel_header/components/header_icon_wrapper';
+import PluginChannelHeaderIcon from '../../components/widgets/icons/plugin_channel_header_icon';
+import {Constants} from 'utils/constants';
+import OverlayTrigger from 'components/overlay_trigger';
+
+class CustomMenu extends React.PureComponent {
```
hmhealey, line 17: Don't we already have a menu for the existing plugin button? Can this share code this that?

Also, to give you a heads up, Apps team is going to be working on writing a reusable menu, so hopefully we can avoid having every component implement its own menu in the future.

plugins/channel_header_call/channel_header_call.jsx
```diff
@@ -0,0 +1,216 @@
...
+            <div className='flex-child'>
+                <Dropdown
+                    id='channelHeaderPlugDropdown'
+                    onToggle={this.toggleDropdown}
+                    onSelect={this.onSelect}
+                    open={this.state.dropdownOpen}
+                >
+                    <CustomToggle
+                        dropdownOpen={this.state.dropdownOpen}
+                        bsRole='toggle'
```
hmhealey, line 164: Is the `bsRole` on these needed? I thought it's only needed if we're using bootstrap.

plugins/channel_header_call/channel_header_call.jsx
```diff
@@ -0,0 +1,216 @@
...
+                type='button'
+                onClick={this.handleClick}
+            >
+                {children}
+            </button>
+        );
+    }
+}
+
+export default class ChannelHeaderCall extends React.PureComponent {
```
hmhealey, line 85: Similarly, are we able to reuse more of the `ChannelHeaderPlug` here? I haven't looked at the code for `ChannelHeaderPlug`, but most of this looks very generic outside of the icon/tooltip used for the dropdown toggle and what's passed in as `components` by `mapStateToProps`. Perhaps they could use the same component and just pass in props for the things that are different?

hmhealey:
While I don't think supporting multiple videoconferencing plugins is something we necessarily need to do, I'm sure there's people that are going to have multiple of them enabled because one dude on their server will prefer Skype over Zoom or something. People are stubborn like that. 😛 

I do like the idea of having some way to promote some plugins to being more accessible like the video conferencing though, particularly because it does make it easier to say "click the Make A Call button" than to say "click the Zoom button which can either be on the bar or in this menu".

I'm brainstorming a bit much right now, but I could see this being extended past just videoconferencing vs other plugins where you might want to have different groups of plugins or different priorities for what should and shouldn't be grouped, but that's entirely out of scope for this.

hmhealey:
Also, to avoid rambling more, how does this look in mobile web view?

larkox:
@hmhealey Regarding mobile view, it does not change at the moment. It will still be seen in the same place as other plugins. We can add also some button for this, but since, if I am not mistaken, mobile web view may suffer some changes, I preferred to solve this only on normal view.

plugins/channel_header_plug/index.js
```diff
@@ -6,9 +6,13 @@ import {getTheme} from 'mattermost-redux/selectors/entities/preferences';
 
 import ChannelHeaderPlug from './channel_header_plug.jsx';
 
-function mapStateToProps(state) {
+function mapStateToProps(state, ownProps) {
+    let components = state.plugins.components.ChannelHeaderButton;
+    if (ownProps && ownProps.callPlugins) {
+        components = state.plugins.components.ChannelHeaderCallButton;
```
hmhealey, line 12: Not something to do now, but if the call button version of this ever starts to require more complicated to get its props, we could also consider splitting this up so that the call button has its own `mapStateToProps`, but it still shares the same underlying `ChannelHeaderPlug` component. Using a prop to switch between the two components is good enough for now though.

andrewbrown00:
Thanks @larkox for kicking this off 🎉 

@aaronrothschild and @esethna I 5/5 agree that the 'call button' should always be visible. 

I also agree with Aaron's comment:
> providing a consistent way to train all users "this is how you can easily start a video meeting"

I do want to challenge where we put the `call` icon, I'm 0/5 if it should live in the channel header vs being near the post control. Most chat products have created the mental modal for users that they either want to `send/replay to a message` or `hop on a call/video` with someone. If we apply that mental model and the proximity principle, I feel like it we could help users having the `CALL` button close to the `Post Input` at the bottom of the channel.

I definitely want to valid this hypothesis with users and this would quick to mock up the options.

larkox:
@andrewbrown00 I think it would be great to have the call button next to the input text. I am not sure if there is any concern on space, but I can try moving the call button next to the input, next to the attach and emoji buttons. It would have a similar behavior as the attach button:
Screenshot from 2020-06-29 09-56-54
Screenshot from 2020-06-29 09-57-00

Do you want me to code it down, or do you prefer to discuss this first?

plugins/channel_header_plug/channel_header_plug.jsx
```diff
@@ -169,27 +186,26 @@ export default class ChannelHeaderPlug extends React.PureComponent {
...
-                                />
-                                <span
-                                    id='pluginCount'
-                                    className='icon__text'
-                                >
-                                    {plugs.length}
-                                </span>
-                            </React.Fragment>
+                            <div>
+                                <React.Fragment>
```
jespino, line 195: You don't need the `React.Fragment` that is used to represent nodes that have no specific parent, in this case, you have a `<div>` that acts like the parent of all the children, so, you don't need it.

plugins/registry.js
```diff
@@ -132,6 +132,40 @@ export default class PluginRegistry {
         return id;
     }
 
+    // Add a "call button"" to the channel header. If there are more than one buttons registered by any
+    // plugin, a dropdown menu is created to contain all the call plugin buttons.
+    // Accepts the following:
+    // - icon - React element to use as the button's icon
+    // - action - a function called when the button is clicked, passed the channel and channel member as arguments
+    // - dropdown_text - string or React element shown for the dropdown button description
+    // - tooltip_text - string shown for tooltip appear on hover
```
jespino, line 141: nit: Should we included somewhere here the min supported version?

larkox:
@andrewbrown00 @asaadmahmood I added the call button next to the post. You can check the screenshots in the PR description.

asaadmahmood:
@andrewbrown00 I'm 5/5 we should have it in the header. Skype, Messenger, Whatsapp, Slack, any application you take, the call button would always be in the header rather than near the post input box. The input box for me translates to something you want to input in the list of messages, where as calling someone is not something you use to add something to the messages list.

So I think messaging/calling are two distinct actions, and then there's the fact that users are already are trained to respond to the call button in a specific place (in the header), so 5/5 not in favour of having the call button in the post.

hmhealey:
Re-requesting my review since we shouldn't be drastically changing a feature after dev review has completed.

In terms of the design, I agree with Asaad on this. I also think that taking up space in the post textbox for this isn't great since it feels like we want as much space there to fit a message as possible, especially on smaller screens.

larkox:
@hmhealey Sorry for not re-requesting myself, but since this is still being decided where to put the button, I did not want to bother you until we had a final decision.

hmhealey:
That makes sense. Let me know when that's decided then :)

andrewbrown00:
@larkox will you update the icon to use the `video` icon instead of the second paperclip.

**This one** 
image

**Instead of the paperclip**
image

larkox:
@andrewbrown00 Done!

andrewbrown00:
Thanks @larkox 😄  I'm going to use this PR to get some quick feedback and initial reactions, I'll follow up after with next steps on if we keep or remove the call button from the `post control`...

larkox:
After discussing with @andrewbrown00 we are moving forward with the button in the textbox (next to the attach button). If you can accept the PR so we can remove the UX Review label it would be great.

@hmhealey this is ready for re-review. Also added @nevyangelova as reviewer since `@jespino` is on PTO.

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+        const {formatMessage} = this.props.intl;
+
+        let bodyAction;
+
+        if (this.props.pluginCallMethods.length === 0) {
+            bodyAction = null;
+        } else if (this.props.pluginCallMethods.length === 1) {
+            const item = this.props.pluginCallMethods[0];
+            bodyAction = (
+                <div>
```
hmhealey, line 55: Is the outer div necessary, or can it be removed?

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+                    <button
+                        type='button'
+                        id='calldButton'
```
hmhealey, line 58: Is the id necessary? I think we've been avoiding using IDs, at least ones that are being used for styling.

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+                        className='style--none post-action icon icon--attachment'
+                        onClick={() => {
+                            if (item.action) {
+                                item.action(this.props.currentChannel, this.props.channelMember);
+                            }
+                        }}
+                        onTouchEnd={() => {
```
hmhealey, line 65: Do we need a separate `onTouchEnd` for this? `onClick` events should still be fired for touch events as far as I know.

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+                                item.action(this.props.currentChannel, this.props.channelMember);
+                            }
+                        }}
+                    >
+                        {item.icon}
+                    </button>
+                </div>
+            );
+        } else {
+            const pluginFileUploadMethods = this.props.pluginCallMethods.map((item) => {
```
hmhealey, line 76: `pluginFileUploadMethods` seems like a typo

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+                return (
+                    <li
+                        key={item.pluginId + item.id + '_callbuttonpluginmenuitem'}
```
hmhealey, line 79: Not something that needs to be changed now, but for future reference, keys in React only need to be unique within the array that they're part of, so something like `key={item.id}` would be sufficient here.

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+                        onClick={() => {
+                            if (item.action) {
+                                item.action(this.props.currentChannel, this.props.channelMember);
+                            }
+                            this.setState({menuOpen: false});
+                        }}
+                    >
+                        <a href='#'>
```
hmhealey, line 87: Because this uses an `href='#'`, the click handler should have an `e.preventDefault()` to prevent having a number sign added to the URL when this is clicked on.

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+                            <span className='mr-2'>
+                                {item.icon}
+                            </span>
+                            {item.dropdownText}
+                        </a>
+                    </li>
+                );
+            });
+            bodyAction = (
+                <div>
```
hmhealey, line 97: Same question here with this div

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+                        >
+                            {pluginFileUploadMethods}
+                        </Menu>
+                    </MenuWrapper>
+                </div>
+            );
+        }
+
+        return (
+            <div className={'style--none'}>
```
hmhealey, line 125: What does the class `style--none` do here? I think it might only be necessary on `button` elements, not divs. If that's the case, we can also just `return bodyAction;` here and remove the extra div.

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+const customStyles = {
+    left: 'inherit',
+    right: 0,
+    bottom: '100%',
+    top: 'auto',
+};
+
+class CallButton extends PureComponent {
+    static propTypes = {
```
hmhealey, line 22: Would it be too much effort to convert this file to TypeScript and specify the prop types using that instead of `propTypes`?

components/widgets/icons/camera_icon.jsx
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {FormattedMessage} from 'react-intl';
+
+export default class CameraIcon extends React.PureComponent {
```
hmhealey, line 7: Would you mind changing this to a functional component since it just has a render method? We're trying to start using functional components whenever we can since they let us use hooks and since they're supposed to start having better performance in future versions of React.

components/widgets/icons/camera_icon.jsx
```diff
@@ -0,0 +1,49 @@
...
+    render() {
+        return (
+            <span {...this.props}>
+                <FormattedMessage
+                    id='generic_icons.camera'
+                    defaultMessage='Camera Icon'
+                >
+                    {(ariaLabel) => (
```
hmhealey, line 15: Technically, this is incorrect typing since `ariaLabel` might not be a string at this point depending on the message being used here. If you convert it to a functional component though, you can use the `useIntl` hook to ensure you get a string here though. It's like `injectIntl`, but simpler.

```typescript
import {useIntl} from 'react-intl';

export default function CameraIcon(props: Props) {
    const intl = useIntl();

    return (
        <span {...props}>
            <svg aria-label={intl.formatMessage({id: 'generic_icons.camera', defaultMessage: 'Camera Icon'})}></svg>
        </span>
    );
}
```

components/call_button/call_button.jsx
larkox: This was copy pasted from the FileUpload component. After a second look, the div can be removed.

components/call_button/call_button.jsx
larkox: This was copy pasted from the FileUpload component. Not sure about the reason it was implemented like this there. If you want, I can remove it.

components/call_button/call_button.jsx
larkox: I was following the pattern on the FileUpload component, which is (I think) wrong at the moment. If a plugin was to implement to different upload methods, there would be two identical keys. I will just leave item.id here.
Do you want me to update also FileUpload with this? Or would you prefer that in a different PR?

components/call_button/call_button.jsx
larkox: Not sure at the moment what is the use, again copy pasted from FileUpload. I will double check.

components/call_button/call_button.jsx
larkox: In order to properly move this to Typescript, we would need plugin folder migrated to typescript too, to have a clear and centralized typing for the pluginCallMethods. I could type them here, but I do not think it is the best place to do it.

larkox:
@hmhealey Since this PR may be reverted eventually since we want to test this feature, let's fix that in a separate PR.

components/call_button/call_button.jsx
```diff
@@ -0,0 +1,125 @@
...
+import CameraIcon from 'components/widgets/icons/camera_icon';
+
+const customStyles = {
+    left: 'inherit',
+    right: 0,
+    bottom: '100%',
+    top: 'auto',
+};
+
+class CallButton extends PureComponent {
```
nevyangelova, line 21: Does this need to be a class component? IMO this is a good candidate for a Functional Component.

larkox: (0/5) For simplicity, I would keep this as a class component, since it does have state and some (minor) behaviour. We can improve it later.
Nevertheless, if you want it to be a functional component, I can give it a try.

nevyangelova: We want to switch to using react hooks with functional components in our codebase and in this case the state can be managed easily with hooks. I suggest the refactor because it would avoid the refactor of another component later. But if the PR is urgent I won't block it. Just a suggestion :)

larkox: Done. Actually the state was not being used, so it was simpler than I expected.

cpanato:
removed the test server, if you need again, please re-add the label

jgilliam17:
Hello @larkox - I need some guidance, not seeing the `call` button. I added Jitsi and Webex on the PR server and can see 2 separate icons in the header, nothing in the textbox.
Screen Shot 2020-08-26 at 4 15 21 PM

larkox:
@jgilliam17 The plugin has to be prepared to use this, so a special build of the plugin has to be added. I will add it to this test server.

larkox:
@jgilliam17 I added two different modified "zoom plugins" so you can disable one and see that the other one still shows as expected. Do not hesitate to contact me if you find any other trouble :smile:

larkox:
@hmhealey @jgilliam17 Heads up that I have made a small change in the CSS to fix some problems I was having when the plugin put some SVG instead of the base icon. I leave up to you whether it needs a new review or not.

hmhealey:
It still looks good from the dev side at least

jgilliam17:
Re-tested, call button as expected. Thanks @larkox

amyblais:
@larkox @jgilliam17 Does this need a Jira ticket and if yes, can you create one?

jgilliam17:
> @larkox @jgilliam17 Does this need a Jira ticket and if yes, can you create one?

I can create one and link here.

jgilliam17:
https://mattermost.atlassian.net/browse/MM-28725

PR #5724, opened 2020-06-15
MM-26008 Fix for suggestions not appearing when opening quick switcher

sudheerDev:
#### Summary
This is a regression caused by https://github.com/mattermost/mattermost-webapp/pull/5702. The linked PR prevents results change on focus to remove a flicker of results disappearing and reappearing on focus change. This caused this issue where quick switcher opens but it does not populate results because it was based on focus in. Instead of changing to a logic where only pretext change causes the computation of results change. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26008

### DIFF
--- a/components/suggestion/suggestion_box.jsx
+++ b/components/suggestion/suggestion_box.jsx
@@ -34,7 +34,7 @@ export default class SuggestionBox extends React.PureComponent {
         /**
          * Array of suggestion providers
          */
-        providers: PropTypes.arrayOf(PropTypes.object),
+        providers: PropTypes.arrayOf(PropTypes.object).isRequired,
 
         /**
          * Where the list will be displayed relative to the input box, defaults to 'top'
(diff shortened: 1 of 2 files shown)


jgilliam17:
Thank you @sudheerDev 
I'm seeing one issue on the test server
Per ticket description, I was expecting to see the `unread` channels listed as switch channel modal is open.
See gif below - It appears that Unread Channels are not shown first and populate only after user clears the search field 

Jun-17-2020 13-39-06

sudheerDev:
@jgilliam17 Hey, sorry about that, Can you give the new spinmint another try?

sudheerDev:
> code looks good, but I'm wondering why pretext is a property on the component instead of a state one, had it been a state property you wouldn't need to compare.

Ya, i don't have much context on this either and seemed like I could break this even more as this is used across the app in multiple components so reluctant to make refactor

BTW could you give another look? i had to make changes after approval

PR #5726, opened 2020-06-15
MM-26065: Revert pr 5654.

mkraft:
#### Summary

Reverts https://github.com/mattermost/mattermost-webapp/pull/5654 as an attempted fix for https://mattermost.atlassian.net/browse/MM-26065 per suggestion here: https://community-daily.mattermost.com/core/pl/r5bjwmbfrifm3djs5zeiczwy4r.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26065

#### Related Pull Requests

https://github.com/mattermost/mattermost-webapp/pull/5654

### DIFF
--- a/root.html
+++ b/root.html
@@ -17,9 +17,6 @@
     <link rel="icon" type="image/png" href="/static/images/favicon/favicon-32x32.png" sizes="32x32">
     <link rel="icon" type="image/png" href="/static/images/favicon/favicon-96x96.png" sizes="96x96">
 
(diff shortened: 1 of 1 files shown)


lieut-data:
As this was observed not to be the cause of the regression, we can, in principle, keep the gist of these changes, so long as we make sure they work with a subpath. Though I'm a little confused about the absolute `/fonts` reference here -- I don't think we serve fonts from there. Wouldn't everything be anchored under `/static`?

PR #5729, opened 2020-06-16
MM-25592: Close interactive dialog's dropdown on click

agarciamontoro:
#### Summary
In the context of interactive dialogs, users can declare one of the fields with the type `select`. This renders an input field with a list of options, filtered by the text entered by the user. In short, it's a dropdown with a filter built on top. However, there was a strange behaviour: the list did not close on click, as dropdowns usually do.

This PR solves that bug, with an additional design decision: the list closes on click only when the input field is empty. Otherwise, the behavior would be really complex to model. For example, the user may want to click on the text to correct a typo, closing the list unexpectedly, or they may double-click to select a word in the input field, which could cause a closing immediately followed by an opening, if not treated with care. As the edge cases here are multiple, I decided to isolate the change to the case where the input field is empty.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25592

#### Screenshots
dropdown


agarciamontoro:
Oops, fixing the tests. Do not review yet :)

asaadmahmood:
@agarciamontoro I think we should close the dropdown on click even if there is text inside of it.

agarciamontoro:
@asaadmahmood I know that that was what we decided yesterday, but this is the result of me trying to simplify this issue, because when working on this, I found out that the complexity of handling the editing of text *and* the behaviour we want with the click is non-trivial and needs some decisions.
Let's assume that the list is closed when the user enters text on it. Then we need to answer the following questions:
- Should we allow the user to click again to open the list?
- If we allow them to re-open it, what happens when the user double-click in a word to select it? The list closes and opens again? Or do we handle double-clicks with some timeout?
- If we go down the route of the timeout for the double click, we need to decide how large the timeout is. And regardless of our decision, we'll have non-happy users, as this is something handled by the OS and configurable by the user (accessibility is an issue here).
- If we don't allow them to re-open the list, then the user needs to press ESC and then click again, or click outside of the field and then click again on it to reopen it.

Facing these questions (although I'm sure that there are a lot more that I did not consider), and assuming that:
- the handling of text editing is already well solved by the browser (and well understood by the users),
- this is not a critical issue, and spending a lot of time designing its spec is not a priority,
then I decided to simplify this as much as possible.

Of course, I'm open to discuss this in depth if you consider it necessary, and prioritize it as needed.

What do you think?

marianunez:
@agarciamontoro Do you know which component owns the down caret on the suggestion list? I was trying to track that down to see if we could just close when you click on that and not the whole input. Maybe that would remove the concerns with the cases you brought above.

agarciamontoro:
> @marianunez Do you know which component owns the down caret on the suggestion list?

It's added with CSS, as far as I know.

> I was trying to track that down to see if we could just close when you click on that and not the whole input.

That would be a better experience, in my opinion. What do you think, @asaadmahmood?

asaadmahmood:
@agarciamontoro The double click, selecting a text functionality, etc, would not be available. Clicking and closing it would work.

We don't have paragraphs here, its not like people would be writing long sentences where they would use a lot of editing. We can consider this primarily as a dropdown, which just has a quick autocomplete filter, but it should primarily behave as a dropdown.

agarciamontoro:
> The double click, selecting a text functionality, etc, would not be available. Clicking and closing it would work.

@asaadmahmood Oh, you're suggesting that closing it would clear all the text, then?

asaadmahmood:
@agarciamontoro Nops, it just behaves like a regular dropdown that allows you to click on it to open and close it.

There is no text selection in terms of double clicking it etc.

agarciamontoro:
After a conversation in Mattermost. we decided that any click in the input field will toggle the open state of the dropdown list. We don't care for the moment for text editing features (for example, double click will close and open the list). The motivation behind this is that we don't expect the users to really edit text there, and we want the component to be more similar to a dropdown. Future feedback from users will probably change our understanding of the component; when that happens, we'll revisit this.

(Commit with the changes coming soon)

agarciamontoro:
@asaadmahmood, I've removed the check for a non-empty text. Does this look good?

dropdown

asaadmahmood:
@agarciamontoro Like a regular dropdown, the option should still remains selected.

agarciamontoro:
Closing this per the decisions in the team meeting, where we agreed that the cost/benefit ratio is high.

PR #5728, opened 2020-06-16
[MM-25834] - Add Icon Common element

nevyangelova:
#### Summary
As part of our new Menu component, and preparation for the Compass design system, this PR adds one of the subelements of the Menu, the Icon Common component. Structure is:

- glyph: the name of the desired icon glyph, using constants (eg. Icon.EMOTICON_HAPPY_OUTLINE)

- size: this desired size of the icon (xsmall, small, standard, large, xlarge), using constants (eg. Icon.SMALL)

- label: text to use for accessibility as needed, uses aria-label

Snapshots should run when we have the main element created. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25834

#### Related PR's 
https://github.com/mattermost/mattermost-webapp/pull/5730
https://github.com/mattermost/mattermost-webapp/pull/5732

PR #5730, opened 2020-06-16
[MM-25835] - Add Icon Button common component

nevyangelova:
#### Summary
As part of our new Menu component, and preparation for the Compass design system, this PR adds one of the subelements of the Menu, the Icon Button Common component. It renders the Icon Common component (to be added). Structure is:

- iconGlyph: the name of the desired icon glyph, using constants (eg. Icon.EMOTICON_HAPPY_OUTLINE)

- size: this desired size of the icon (xsmall, small, standard, large, xlarge), using constants (eg. Icon.SMALL)

- label: text to use for accessibility as needed, uses aria-label

- disabled: a boolean indicating whether the IconButton should be disabled or not

Snapshots should run when we have the main element created.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25835

#### Related PR's 
https://github.com/mattermost/mattermost-webapp/pull/5728
https://github.com/mattermost/mattermost-webapp/pull/5732


components/main_menu/elements/icon_button_common.tsx
```diff
@@ -0,0 +1,32 @@
...
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+interface Props {
+    iconGlyph: string;
+    size: string;
+    label: string;
+    disabled?: boolean;
+    additionalClass?: string;
```
hmhealey, line 11: 0/5 on this naming, but maybe something like `buttonClassName` would make sense here? Saying what is being styled might be a good idea in case we end up having custom components that support multiple additional classes. I suppose we can always change that if we run into that being a problem, but we'll want to be consistent with whatever we pick.

components/main_menu/elements/icon_button_common.tsx
```diff
@@ -0,0 +1,32 @@
...
+const IconButtonCommon: React.FC<Props> = ({iconGlyph, size, label, disabled, additionalClass}: Props) => {
+    return (
+        <div
+            className='IconButton__wrapper'
+        >
+            <button
+                className={`IconButton IconButton___${size} ${additionalClass}`}
+                disabled={disabled}
+                tabIndex={-1}
+                aria-label={label}
```
hmhealey, line 23: This will definitely be something we can add later, but I wonder how often we'll want to just directly translate this label. If so, adding in i18n support might be a good idea eventually.

sass/components/_icon-button-common.scss
```diff
@@ -0,0 +1,120 @@
...
+        left: 0;
+        bottom: 0;
+        right: 0;
+        border-radius: 4px;
+        border: solid 2px transparent;
+        z-index: 0;
+    }
+
+    &:hover,
+    &.hover {
```
hmhealey, line 40: I like this idea. Is it so that we can override the state programatically in addition to having it automatically set?

sass/components/_icon-button-common.scss
```diff
@@ -0,0 +1,120 @@
...
+        background: color(--center-channel-color-rgb, 0.08);
```
hmhealey, line 41: Is this `color` function something we built or is it built in? I was supposed to create a ticket to make a helper like this :p

nevyangelova:
> Looks good, but I have a bunch of things I wanted to ask about as usual.
> 
> In addition to those, is `components/main_menu/elements` the right place to be putting these? This seems like it would be used outside of new main menu stuff.

Yesss that's my mistake. Will move to 'components/compass/menu' :)

components/main_menu/elements/icon_button_common.tsx
nevyangelova: I agree with you, we should decide on a unified naming. My thinking was that when we render the components it will be clear what the element name is and then it's clear what the additionalClass represents. Like `<IconButtonCommon additionalClass='happy'/>` or `<IconCommon additionalClass='sad'/>`

sass/components/_icon-button-common.scss
nevyangelova: indeed, as well as adding some additional control to hover. Some browsers might not support it as expected.

sass/components/_icon-button-common.scss
nevyangelova: Whoops. It seems like it does not exist. Probably copied from figma. Will fix.

components/main_menu/elements/icon_button_common.tsx
hmhealey: As long as these components stay small, that should work, although if they start to contain more, then it might be necessary to be more specific. Like maybe we have a `List` component that needs something like `containerClassName` and `listItemClassName`.

Also, I'm leaning towards calling them `additionalClassName` over `additionalClass` just because that's what React uses elsewhere

hmhealey:
Do these need to be in menu though? I would've thought they could just live somewhere like `components/compass/icon_button`

deanwhillier:
Closing as we are restarting this effort with a slightly different approach

PR #5731, opened 2020-06-16
MM-26066 Move rudder to use CDN instead of npm

sudheerDev:
#### Summary
Remove CDN script injection of rudder and use import from MM-Redux 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26066

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1178


sudheerDev:
> I assume we'll bump `package.json` when the corresponding mattermost-redux PR is merged?

@lieut-data Yup, I will update it once I figure out rudder keys for testing changes on spinmint.

sudheerDev:
@lieut-data @grundleborg WOuld one of you give this another look? I had to change some tests, i don't know why and how they are failing 🤷‍♂️ but it is a better practice to use `runOnlyPendingTimers` so went ahead and made the change

lieut-data:
Seems ok to me!

PR #5732, opened 2020-06-16
[MM-25836] - Add common Label component

nevyangelova:
#### Summary
As part of our new Menu component, and preparation for the Compass design system, this PR adds one of the subelements of the Menu, the Label Common component. Structure is:

children: For setting in the label text we rely on the children.

size: For setting in the label size (small, standard, large). Use constants for specifying size (i.e. Label.SIZE_SMALL)

wrap:  For indicating that the label should wrap.

onMouseOver: Adds possibility to render a tooltip.

Snapshots should run when we have the main element created.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25836

#### Related PR's
#5730
https://github.com/mattermost/mattermost-webapp/pull/5728


components/compass/label/label.tsx
```diff
@@ -2,18 +2,24 @@
...
+    children: React.ReactNode;
+    size: string;
+    wrap?: boolean;
+    onMouseOver?: () => void;
 }
+
+const Label: React.FC<Props> = ({children, size, wrap, onMouseOver}: Props) => {
+    return (
+        <span
+            className={classNames([`Label Label___${size}`, {Label___wrap: wrap}])}
```
deanwhillier, line 17: We'll need a default value for size in case one isn't provided via props

components/compass/label/label.tsx
```diff
@@ -2,18 +2,24 @@
...
+            onMouseOver={onMouseOver}
```
deanwhillier, line 18: I was thinking about taking the opportunity to be smart about passing this event up to the parent. We really only need a tooltip if the label is truncated due to space restrictions so I was thinking we only need to tell the parent the user is currently moused over the label if the label's text is truncated. There's no reason to have a tooltip otherwise and requiring the parent to determine if a label is truncated does't feel like the right approach to me. The event function name might need to be customized to make sense if we take this approach.

We also need to consider how to let the parent know the user has moused away so the parent can appropriately dismiss any open tooltips.

nevyangelova:
> Some general comments:
> 
> * After thinking about this for a bit, I'm wondering if we should rename it so as not to get it confused with a form label? We almost don't need this as a component apart from it having the ability to automatically truncate itself there there isn't enough room, having the ability to toggle between single line and wrapping with minimal effort and a means of setting size based on a limited set of options.
> * We still need to add the constants for the acceptable values for size. Exported typescript string enums maybe?
> * Need to add the SCSS for this one as wll

1. I like to have this as a separate component even if it doesn't have much functionality, its possible that some will be added in the future, and keeping it separate will help with keeping it clean. As far as naming goes, to me Label sounds clear in this case, and I would name the FormLabel when we create that component, but to me general label means 'text label', thoughts? 
2. I will definitely do that, is there any reference to them apart from the ticket? 
3. This got lost with merging with Master OOOPS!

components/compass/label/label.tsx
nevyangelova: Do we know which one we want as default? I suggest 'medium'?

components/compass/label/label.tsx
nevyangelova: So we can have the 'truncated' prop passed to the parent? I think this creates a two-way flow which is an anti-pattern. Or did I misunderstand the suggestion? I think in the function we can pass 'truncated = true' when this happens render the tooltip from the parent. 

```const onMouseOver = (truncated) => {
         if (truncated) {
             ...render toolTip
         }
    }
```

deanwhillier:
Closing this as we are restarting this effort with a slightly different approach

PR #5733, opened 2020-06-16
MM-25138: CircleCI caching improvements

lieut-data:
#### Summary
Skip `restore_cache` on the `install` step, since we use `npm ci` which only turns around and throws away the restored `node_modules`. This saves time both from the restore, and the the recursive directory removal. (We still `save_cache` since we use the resulting `node_modules` in other jobs.)

Also tweak the use of the workspace persistence in the `build` and `build-storybook` to only persist the newly generated artifacts instead of the entire workspace a second time. This dramatically shortens the `Persisting to Workspace` step.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-25138

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,6 @@ jobs:
     working_directory: ~/mattermost/mattermost-webapp
     steps:
       - checkout
-      - *restore_cache
       - run: |
           npm ci
           cd node_modules/mattermost-redux && npm i && npm run build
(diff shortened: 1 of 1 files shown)


lieut-data:
Skipping `3: QA Review` since only build related.

lieut-data:
Note that `install` could be made faster by using `npm install` instead of `npm ci` and actually reusing the `node_modules` saved from past workflows. This assumes we commit to pin exact versions in `package.json` and lose the "protections" `npm ci` offers.

For now, I'm just fixing the immediate issue.

PR #5741, opened 2020-06-17
Renaming diagnostics-to-telemetry

jespino:
#### Summary
Renaming diagnostics-to-telemetry


components/root/root.jsx
```diff
@@ -158,14 +158,14 @@ export default class Root extends React.PureComponent {
         }
 
         const segmentKey = Constants.DIAGNOSTICS_SEGMENT_KEY;
```
jespino, line 160: I'm not renamign this constast because is a placeholder, but I don't know if we are changing this value somehow during the initialization.

components/root/root.jsx
```diff
@@ -199,7 +199,7 @@ export default class Root extends React.PureComponent {
         const rudderKey = Constants.DIAGNOSTICS_RUDDER_KEY;
```
jespino, line 199: Same about this constants.

components/root/root.jsx
Willyfrog: you can change it, but then you'll need to modify [here]
(https://github.com/mattermost/mattermost-shared-pipelines/blob/e8a8a685ad26000bbbf9613e779d93d896e18843/src/org/mattermost/PlatformStages.groovy#L500) as well to ensure new builds are properly managed

hmhealey: I think the key is changed by the build process if I remember correctly, so we'd need to find where that's actually done.

prapti (approved):
LGTM!
Tested with a clean local instance. The test event telemetry are showing up as expected.
Approving.

PR #5743, opened 2020-06-17
Update dependencies

crspeller:
#### Summary
Updates some of the easier to update dependencies. 

#### Ticket Link
Partially completes: https://mattermost.atlassian.net/browse/MM-24300


components/create_team/index.js
```diff
@@ -6,17 +6,16 @@ import {connect} from 'react-redux';
 import {getConfig} from 'mattermost-redux/selectors/entities/general';
 import {getCurrentChannel} from 'mattermost-redux/selectors/entities/channels';
 import {getCurrentTeam} from 'mattermost-redux/selectors/entities/teams';
-import {GlobalState} from 'mattermost-redux/types/store';
 
 import CreateTeam from './create_team';
 
-function mapStateToProps(state: GlobalState) {
+function mapStateToProps(state) {
```
hmhealey, line 12: Why are you changing this file back to JavaScript? If you're running into missing types in `GlobalState`, that import should be switched to import from the webapp's `types/store`

utils/text_formatting.tsx
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+//eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+//@ts-ignore
```
hmhealey, line 5: Why are you disabling type checking here? We should create a followup to re-enable this if it's too much to fix for now.

utils/syntax_highlighting.tsx
```diff
@@ -2,133 +2,10 @@
 // See LICENSE.txt for license information.
 
 import hlJS from 'highlight.js';
-import hljsActionscript from 'highlight.js/lib/languages/actionscript.js';
```
hmhealey, line 5: Are languages automatically enabled for highlighting now? If so, we might be able to close a help wanted ticket where we wanted to lazy load individual languages.

tests/setup.js
```diff
@@ -67,7 +67,11 @@ beforeEach(() => {
 
 afterEach(() => {
     if (logs.length > 0 || warns.length > 0 || errors.length > 0) {
-        throw new Error('Unexpected console logs' + logs + warns + errors);
+        const message = 'Unexpected console logs' + logs + warns + errors;
+        if (message.includes('componentWillReceiveProps')) {
```
hmhealey, line 71: What component is printing this error? Was it because of a `componentWillReceiveProps` or an `UNSAFE_componentWillReceiveProps`, since the latter is what we should be using now

patches/@types+marked+0.6.5.patch
```diff
@@ -1,12 +0,0 @@
-diff --git a/node_modules/@types/marked/index.d.ts b/node_modules/@types/marked/index.d.ts
```
hmhealey, line 1: Since this is no longer needed, can we remove the part of the `check-types` and `postinstall` scripts in package.json that call `patch-package`?

components/create_team/index.js
crspeller: There was some strange problem with this file. Looked like a bug in webpack but I didn't investigate further. Changing it back to js fixed the issue. It didn't seem to have anything to do with the file contents.

utils/text_formatting.tsx
crspeller: Ya the types for XRegExp are wrong. I tried reverting XRegExp, but it seems that it was the typechecker not noticing, not the types actually changing. 
Created a ticket: 
https://mattermost.atlassian.net/browse/MM-26347

utils/syntax_highlighting.tsx
crspeller: Yes they are automatically enabled now. You can still change it to a mode where the languages can be imported individually, so that ticket is probably still possible.

tests/setup.js
crspeller: Loads and loads of them. Hence me just hiding the error rather than trying to fix it.

patches/@types+marked+0.6.5.patch
crspeller: Done!

utils/syntax_highlighting.tsx
hmhealey: I wonder what the impact of that on the bundle size will be. I can't imagine it's actually too much to include some new languages even compared to entirely new libraries like we add frequently.

As for the lazy loading ticket, I think I wrote that in a way that it won't need to be updated for this, but I did link to this PR there in case it changes things.

components/create_team/index.js
hmhealey: Strange. Well, we can always see later. I doubt this component is one that we're changing enough for autocomplete and such to be really helpful.

tests/setup.js
hmhealey: I guess if it's printing errors here, that means they'll be printed in dev mode too so that we'll be more motivated to actually fix them 😛

PR #5746, opened 2020-06-18
MM-16457: Increase width of the system console logs

agnivade:
All admin console content boxes were limited to a max-width of 920px. However,
this did not make sense for console logs as users would have to frequently scroll horizontally
to see more logs.

We set an explicit id on the logs component to override this limit, while keeping the top banner
of the same width.

Other possible options without needing to add an id would have been to explicitly increase the
width of LogList component by taking it out of the DOM hierarchy and then increasing its width.
Overriding the parent div seemed like a better choice.

### Ticket link

https://mattermost.atlassian.net/browse/MM-16457

admin_console


agnivade:
@jasonblais - I kept the top banner as the same width. Let me know this is what you expected. Thanks.

jasonblais:
> I kept the top banner as the same width. Let me know this is what you expected. Thanks.

@agnivade If we were to make the top banner as wide as the logs, would this affect the width of the top banner in all System Console pages?

agnivade:
Of course not. It's just code. I can make it do whatever we want. :)

But the decision of whether it will be consistent or not is left to you. :)

jasonblais:
@asaadmahmood Would you have opinions on whether the top banner should

1) remain the same fixed length as in other pages (as shown below); or
2) have the same width as the logs table

image

I'm leaning heavily on one of these options, but would like UX confirmation :)

asaadmahmood:
Fixed width as in the other pages for now. We're looking at a system console revamp, which hopefully would cover this more holistically. Fixed width as in the other pages for now. We're looking at a system console revamp, which hopefully would cover this more holistically.

agnivade:
ping @sudheerDev @Willyfrog

agnivade:
@Willyfrog - Not sure I understood your solution properly. Could you expand a bit more?

components/admin_console/server_logs/logs.jsx
```diff
@@ -92,7 +92,10 @@ export default class Logs extends React.PureComponent {
                 />
 
                 <div className='admin-console__wrapper'>
-                    <div className='admin-console__content'>
+                    <div
+                        id='admin-logs-content'
+                        className='admin-console__content'
```
Willyfrog, line 97:
```suggestion
                        className='admin-console__content admin-logs-content'
```

sass/routes/_admin-console.scss
```diff
@@ -946,3 +946,12 @@
...
 }
+
+#admin-logs-content {
+    max-width: none;
+
+    // We keep the banner the same width and only expand the content box.
+    .banner {
+        max-width: 920px;
+    }
+}
```
Willyfrog, line 957:
```suggestion
.admin-logs-content {
    max-width: none;

    // We keep the banner the same width and only expand the content box.
    .banner {
        max-width: 920px;
    }
}
```

sass/routes/_admin-console.scss
```diff
@@ -946,3 +946,12 @@
         max-width: none;
     }
 }
+
+#admin-logs-content {
```
sudheerDev, line 950: @Willyfrog was talking about this, @agnivade instead of using id here as css selector, can you use class

```
.admin-console__content {

}
```

agnivade:
Oh I see, you want me to add another class instead of id. Got it. Yes, that works. Will change.

agnivade:
I have to use `!important` to override the other class. I hope you are fine with it.

sass/routes/_admin-console.scss
sudheerDev: Sorry important is an issue, Can you change this to instead

```
.admin-console .admin-console__content.admin-logs-content {

}

```

agnivade: Done. Thanks.

PR #5753, opened 2020-06-18
[MM-26305] migrate refs example

Willyfrog:
#### Summary

This is the sample ticket for the migration campaign. 

It migrates string refs for the checkbox defining password rules in the admin console.

#### Ticket Link

[MM-26305](https://mattermost.atlassian.net/browse/MM-26305)


Willyfrog:
> This works for an example, but in this case, the values that we're using are in `this.state`, so do we even need refs here?

@hmhealey i'll create a second ticket for cleaninig this, but keep this one as example.

PR #5755, opened 2020-06-18
MM-25640 - Leaving archived channel does not return user to the last viewed channel

cpoile:
#### Summary
- Archived channels are removed from local storage automatically.
- So, only if the current channel is /not/ an archived channel, remove the previous channel name from local storage

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25640


cpoile:
I can't figure out why the tests are saying `TypeError: (0 , _channel_utils2.isArchivedChannel) is not a function` -- do either of you guys know?

agarciamontoro:
I have absolutely no idea. I checked what the type of the symbol is with `console.log(typeof(isArchivedChannel));` and it does print `undefined`. But it's defined! If you copypaste the same function in that file and remove the import, it works smoothly. I am intrigued :thinking:

cpoile:
@saturninoabril would you have a moment to take a look at this? For some reason the function is not defined when the test runs, but it is defined and works in normal use. Any ideas?

saturninoabril:
Probably due to the mocking here - https://github.com/mattermost/mattermost-webapp/blob/master/actions/views/channel.test.js#L24 - that made it undefined. It should include the original and only mock the function as needed. I haven't tried it should be something like,
```javascript
jest.mock('utils/channel_utils.jsx', () => {
    const original = require.requireActual('utils/channel_utils.jsx');
    return {
        ...original,
        getRedirectChannelNameForTeam: () => 'town-square',
    };
});
```

cpoile:
Thank you @saturninoabril, that fixed it!

agarciamontoro:
Oh, it makes sense now. Thank you Saturnino for the explanation, I was also curious about this!

PR #5765, opened 2020-06-19
[MM-26327] use eslint to prevent string refs usage

Willyfrog:
#### Summary
To prevent the campaign growing because people might copy code from the codebase, we are adding a rule to prevent string ref usage.

#### Ticket Link

[MM-26327](https://mattermost.atlassian.net/browse/MM-26327)


Willyfrog:
> My only thought would be to use eslint-disable-line instead of the whole file? (idk, maybe more time consuming, so maybe not the way to go)

We talked this in the hangout, the reason to go with the whole file is that if someone copy-paste, it will throw the error as there is no line-disable, and if they add the whole file-disable we can spot it more easily in code review

PR #5767, opened 2020-06-20
MM-25951: Fix license id in link

sbishel:
#### Summary
Fix Id property for link in ConfigurationBar
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25951

#### Screenshots
Screen Shot 2020-06-20 at 7 18 31 AM

### DIFF
--- a/components/announcement_bar/configuration_bar/configuration_bar.jsx
+++ b/components/announcement_bar/configuration_bar/configuration_bar.jsx
@@ -41,7 +41,7 @@ class ConfigurationAnnouncementBar extends React.PureComponent {
     render() {
         // System administrators
         if (this.props.canViewSystemErrors) {
-            const renewalLink = RENEWAL_LINK + '?id=' + this.props.license.id + '&user_count=' + this.props.totalUsers;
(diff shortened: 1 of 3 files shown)


components/announcement_bar/configuration_bar/configuration_bar.jsx
```diff
@@ -41,7 +41,7 @@ class ConfigurationAnnouncementBar extends React.PureComponent {
     render() {
         // System administrators
         if (this.props.canViewSystemErrors) {
-            const renewalLink = RENEWAL_LINK + '?id=' + this.props.license.id + '&user_count=' + this.props.totalUsers;
+            const renewalLink = RENEWAL_LINK + '?id=' + this.props.license.Id + '&user_count=' + this.props.totalUsers;
```
hahmadia, line 44: Could we change this to use back ticks instead of adding them together?
`${RENEWAL_LINKI'd........`. I think the proper name is template literals instead of backticks https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals .

fmunshi, line 44:
```suggestion
            const renewalLink = `${RENEWAL_LINK}?id=${this.props.license.Id}&user_count=${this.props.totalUsers}`;
```

PR #5782, opened 2020-06-23
[MM-25785] allow pasting tables into edit post, keep caret position

Willyfrog:
#### Summary
fix adding tables into post edits
while working on it, i found it wasn't keeping caretposition when pasting tables from google docs spreadhseets, so I changed that as well

#### Ticket Link

[MM-25785](https://mattermost.atlassian.net/browse/MM-25785)

### DIFF
--- a/utils/paste.tsx
+++ b/utils/paste.tsx
@@ -53,7 +53,7 @@ function tableHeaders(row: HTMLTableRowElement): string[] {
     return Array.from(row.querySelectorAll('td, th')).map(columnText);
 }
 
-export function formatMarkdownTableMessage(table: HTMLTableElement, message?: string): string {
+export function formatMarkdownTableMessage(table: HTMLTableElement, message?: string, caretPosition?: number): string {
     const rows = Array.from(table.querySelectorAll('tr'));
 
     const headerRow = rows.shift();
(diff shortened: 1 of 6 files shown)


utils/paste.tsx
```diff
@@ -66,8 +66,14 @@ export function formatMarkdownTableMessage(table: HTMLTableElement, message?: st
     }).join('\n');
 
     const formattedTable = `${header}${body}\n`;
-
-    return message ? `${message}\n\n${formattedTable}` : formattedTable;
+    if (!message) {
+        return formattedTable;
+    }
+    if (typeof caretPosition === 'undefined') {
+        return message ? `${message}\n\n${formattedTable}` : formattedTable;
```
marianunez, line 73: This check for `message` seems redundant given the new introduced `if` above ? 
```suggestion
        return ${message}\n\n${formattedTable}`;
```

utils/paste.tsx
```diff
@@ -66,8 +66,14 @@ export function formatMarkdownTableMessage(table: HTMLTableElement, message?: st
...
+    }
+    const newMessage = [message.slice(0, caretPosition), formattedTable, message.slice(caretPosition)];
+    return newMessage.join('\n');
```
marianunez, line 76: Should this be a double `\n` to be consistent with the case above when no `caretPosition` is provided?

utils/paste.tsx
Willyfrog: you're right! changing it right away

utils/paste.tsx
Willyfrog: originally I had `\n\n` but there were a lot of spaces around and it looked weird because the table is going to render in a new line.

I thought about changing the other one as well, but it probably is there because the most likely case is that the user didn't add it, and just pasted the table in, so having that `\n` might make everything more readable.

I'm 0/5 on this, so if you think it should have the double new-line, I really don't mind pushing the change.

marianunez: 0/5, just caught my attention that it was missing the `/n`. I think it's fine that it could be slightly different in a case where it's pasted between existing text.

PR #5783, opened 2020-06-23
MM-23788 Setup for visual regression testing

saturninoabril:
#### Summary
Integrate visual regression testing using Applitools with Cypress.

#### How to run visual regression testing?
The feature is disabled by default.
To be able to run, you may get API key from Applitools and set into your env variable, and then run:
```bash
ENABLE_VISUAL_TEST=true node run_tests.js --group='@visual_regression'
```

Note: 
1. On CI, visual regression testing will run separately from functional testing.
2. It's recommended to separate visual regression tests (into a dedicated folder `/visual_regression`)  from functional tests:
- to avoid flaky tests by directly going into the UI and do snapshot instead of mixing with other tests/actions
- to not incur long duration to current overall tests since visual regression will take some time to process snapshots/results

#### How to get screenshot from the spec file?
There are three steps to get the screenshot.
1. Open the window by `cy.visualEyesOpen()`
2. Save snapshot by `cy.visualSaveSnapshot()`
3. Then close the window `cy.visualEyesClose()`

#### Ticket Link
 JIRA ticket: https://mattermost.atlassian.net/browse/MM-23788

### DIFF
--- a/e2e/applitools.config.js
+++ b/e2e/applitools.config.js
@@ -1,0 +1,19 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// See https://github.com/applitools/eyes.sdk.javascript1/tree/master/packages/eyes-cypress#advanced-configuration
+
+module.exports = {
+    appName: 'Mattermost Webapp UI',
+    accessibilityValidation: {level: 'AA', guidelinesVersion: 'WCAG_2_0'},
+    batchName: 'Webapp - master (dev)',
+    branch: 'master-dev',
+    baselineBranch: 'master-dev',
+    parentBranch: 'master-dev',
+    browser: [
+        {width: 1024, height: 768, name: 'chrome'},
+    ],
+    concurrency: 1,
+    matchLevel: 'Strict',
+    showLogs: false,
+};
--- a/e2e/cypress/integration/visual_regression/system_console/environment_section_spec.js
+++ b/e2e/cypress/integration/visual_regression/system_console/environment_section_spec.js
@@ -1,0 +1,122 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Group: @enterprise @system_console @visual_regression
+
+import * as TIMEOUTS from '../../../fixtures/timeouts';
+
+describe('System Console - Enterprise', () => {
+    const testCases = [
+        {
+            section: 'Environment',
+            header: 'Web Server',
+            sidebar: 'Web Server',
+            url: 'admin_console/environment/web_server',
+        },
+        {
+            section: 'Environment',
+            header: 'Database',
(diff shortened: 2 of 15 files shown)


e2e/applitools.config.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// See https://github.com/applitools/eyes.sdk.javascript1/tree/master/packages/eyes-cypress#advanced-configuration
+
+module.exports = {
+    appName: 'Mattermost Webapp UI',
+    accessibilityValidation: {level: 'AA', guidelinesVersion: 'WCAG_2_0'},
+    batchName: 'Webapp - master (dev)',
```
josephbaylon, line 9: Quick question: is batchName here supposed to be a default value? Reason I asked is when I don't set APPLITOOLS_BATCH_NAME, the value of the batch name in the UI reports is `undefined`.

prapti:
In review.

e2e/applitools.config.js
saturninoabril: Oh, good catch. Yes, that should be the default value. Should be fixed now, PTAL.

saturninoabril:
@josephbaylon re: NODE_TLS_REJECT_UNAUTHORIZED, I'm not particularly sure what's causing it and the implication of it. I'm seeing that too locally but not in [CircleCI](https://circleci.com/api/v1.1/project/github/saturninoabril/mattermost-cypress-docker/460/output/110/0?file=true&allocation-id=5ef300b64873eb53ef883092-0-build%2F7C27EF1E), even if both are on the same version of node `@v10.20.1`.

e2e/cypress/integration/visual_regression/system_console/environment_section_spec.js
```diff
@@ -0,0 +1,122 @@
...
+            });
+
+            // * Verify that it redirects to the URL and matches with the header
+            cy.url().should('include', testCase.url);
+            cy.get('.admin-console').should('be.visible').within(() => {
+                cy.get('.admin-console__header').should('be.visible').and(testCase.headerContains ? 'contain' : 'have.text', testCase.header);
+            });
+        });
+    });
+});
```
agarciamontoro, line 122: There is no visual snapshot in this file. Should it be outside of the `@visual_regression` group?

saturninoabril: Good catch! I also realized that this was unintentionally checked in since I have it on my on going dev but it's fine to remain as is. I will have follow up PRs and will add corresponding metadata and visual tests.

saturninoabril:
@agarciamontoro Thanks for running the tests and valuable inputs.
- __npm i__ - we have it in the [docs](https://developers.mattermost.com/contribute/webapp/end-to-end-tests/#quick-overview-on-how-to-run-e2e-testing) but maybe there's something we can improve?
- __Chromium__ - unfortunately it's not due to Cypress but Applitools.  It's not currently supported per their [list](https://github.com/applitools/eyes.sdk.javascript1/tree/master/packages/eyes-cypress#configuring-the-browser) at the moment.
- the last point (see in the comment itself)

agarciamontoro:
> npm i - we have it in the docs but maybe there's something we can improve?

Probably not, when I first run an e2e test I followed that guide and everything was clear. I added that comment in case we have different docs for the visual regression testing. But if that's the entry point, it's perfect the way it is now. Thanks!

PR #5784, opened 2020-06-23
MM-26388 - Update error handling from server

sbishel:
#### Summary
Updated to handle specific error from the server
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26388

#### Screenshots
Screen Shot 2020-06-23 at 2 44 31 PM

### DIFF
--- a/components/admin_console/group_settings/group_details/group_details.jsx
+++ b/components/admin_console/group_settings/group_details/group_details.jsx
@@ -352,8 +352,15 @@ export default class GroupDetails extends React.PureComponent {
                     result.error.server_error_id === 'api.ldap_groups.existing_user_name_error' ||
                     result.error.server_error_id === 'api.ldap_groups.existing_group_name_error') {
                     serverError = GroupNameIsTakenError;
+                } else if (result.error.server_error_id === 'model.group.name.invalid_length.app_error') {
+                    serverError = (
+                        <FormattedMessage
+                            id='admin.group_settings.group_detail.invalid_length'
+                            defaultMessage='Name must be 1 to 64 lowercase alphanumeric characters.'
+                        />
+                    );
                 } else {
-                    serverError = <FormError error={result.error?.message}/>;
+                    serverError = result.error?.message;
                 }
             }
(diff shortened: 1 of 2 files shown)


components/admin_console/group_settings/group_details/group_details.jsx
```diff
@@ -352,8 +352,15 @@ export default class GroupDetails extends React.PureComponent {
...
+                } else if (result.error.server_error_id === 'model.group.name.invalid_length.app_error') {
+                    serverError = (
+                        <FormattedMessage
+                            id='admin.group_settings.group_detail.invalid_length'
+                            defaultMessage='Name must be 1 to 64 lowercase alphanumeric characters.'
+                        />
+                    );
                 } else {
-                    serverError = <FormError error={result.error?.message}/>;
+                    serverError = result.error?.message;
```
sbishel, line 363: I went back and forth on this. I think the only change really needed is for line 363. "result.error" is returned as a AppError, so it should already be translated coming from the server. However, most of the WebApp code base and this function convert server errors, so I did both.

hahmadia: @sbishel 
I'm a little unclear but what was the reason to remove the `FormError` part of it and just make `serverError` equal to the message directly rather than passing it to the `FormError` component? I'm not sure the reason. Is `FormError` just not needed and better to display the message directly?

sbishel: A prior PR changed how the error is displayed.  It now uses a <FormError> when it is rendered here -
https://github.com/mattermost/mattermost-webapp/blob/master/components/admin_console/group_settings/group_details/group_details.jsx#L552

Because another <FormError> was getting embedded in it, that is what was was causing the improper formatting. Now serverError should simply be the message to be displayed, which is done with a <FormatMessage> in most of the function. However, this is catching unexpected errors, therefore it just gets set to the message.

hahmadia: Gotcha. Thanks for the explanation!

PR #5794, opened 2020-06-24
MM-25525 Sidebar: Event for tracking number of visible GMs and DMs in LHS

sudheerDev:
#### Summary
This is built on top of https://github.com/mattermost/mattermost-webapp/pull/5547. Specific changes to this PR are in https://github.com/mattermost/mattermost-webapp/commit/7214ce240eb49ee786f5c9387e40e1dc6598d7d5

Adding track event on page load and wait for sidebar profiles to load to get the final true value.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25525


esethna:
We want this in the v5.25 release cc// @amyblais

amyblais:
@sudheerDev Kind note that tomorrow is code complete

sudheerDev:
@amyblais Just had a quick chat with eric, pushing it to 5.26v

components/data_prefetch/index.ts
```diff
@@ -75,8 +77,9 @@ function mapStateToProps(state: GlobalState) {
 
 function mapDispatchToProps(dispatch: Dispatch<GenericAction>) {
     return {
-        actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc>, Actions>({
+        actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc>, any>({
```
nevyangelova, line 80: Won't block this PR but is it possible to add the correct type?

esethna:
@hmhealey @sudheerDev ideally this works for users with or without the new sidebar enabled. But if the effort level is large then a smaller data set would be acceptable

sudheerDev:
@hmhealey 

Would this change?

```
const DMGMCategory = categories.find((category) => category.type === CategoryTypes.DIRECT_MESSAGES);
        const channels = getChannelsForCategory(state, DMGMCategory);
```

hmhealey:
> @hmhealey
> 
> Would this change?
> 
> ```
> const DMGMCategory = categories.find((category) => category.type === CategoryTypes.DIRECT_MESSAGES);
>         const channels = getChannelsForCategory(state, DMGMCategory);
> ```

No, that wouldn't help since the categories still aren't loaded unless the new sidebar is enabled. Instead, we could either only collect this with the new sidebar enabled or we could just use the other selectors to manually filter out any hidden DMs and GMs.

sudheerDev:
@hmhealey aah, i see what you mean, @jasonblais Let me check this again

actions/user_actions.jsx
```diff
@@ -419,3 +421,11 @@ export function autoResetStatus() {
         return userStatus;
     };
 }
+
+export function trackDMGMOpenChannels() {
+    return async (doDispatch, doGetState) => {
+        const state = doGetState();
+        const channels = getDirectChannels(state);
```
sudheerDev, line 428: The selector does filter the auto closed, manual closed DM's and GM's but the comment seems to be wrong or outdated https://github.com/mattermost/mattermost-redux/blob/master/src/selectors/entities/channels.ts#L1043

hmhealey: When we're moving these into the webapp, we really need to take a look at what selectors we actually need and which aren't used any more. We've got probably 3 or 4 versions of these selectors that return channels where some are filtered in different ways and some have extra data added to the channel. That code is kind of a mess, so I'm not surprised the comments could be out of date

components/data_prefetch/data_prefetch.tsx
```diff
@@ -52,6 +53,7 @@ export default class DataPrefetch extends React.PureComponent<Props, {}> {
             queue.add(async () => this.prefetchPosts(currentChannelId));
             await loadProfilesForSidebar();
             this.prefetchData();
+            this.props.actions.trackDMGMOpenChannels();
```
hmhealey, line 56: Just to confirm, how often is this event sent out? Having extra data isn't too much of a problem since we can filter out any extra events for a given user, but we don't want to send a ridiculous amount of events

components/data_prefetch/index.ts
hmhealey: Since `Actions` was updated above, this looks like it could use the correct type instead of `any`

hmhealey:
The only thing I really want changed here is the typing [here](https://github.com/mattermost/mattermost-webapp/pull/5794#discussion_r455844627) since we should be avoiding using `any`

components/data_prefetch/data_prefetch.tsx
sudheerDev: This event is triggered on the load of the webapp alone.

sudheerDev:
> The only thing I really want changed here is the typing [here](https://github.com/mattermost/mattermost-webapp/pull/5794#discussion_r455844627) since we should be avoiding using `any`

The issue here is that the action would expect a return type if i change the type.
Untitled

The way i see to fix it is by returning something, but that seems odd considering it does not need to have

hmhealey:
Manually cherry-picked onto release-5.26

esethna:
@sudheerDev can you help submit a docs PR to add whatever flags we added here to this doc: https://docs.mattermost.com/administration/telemetry.html?highlight=telemetry

sudheerDev:
@esethna i don't see any section where i can update the event here. This is an event triggered on load of webapp. The telemetry page seems outdated and missing a lot of info in it. Like we don't use segment anymore for webapp telemetry we use rudder now.

PR #5804, opened 2020-06-25
Update NOTICE.txt

amyblais:
Matching with what is here: https://github.com/mattermost/mattermost-webapp/blob/release-5.25/package.json


NOTICE.txt
```diff
@@ -2871,6 +2855,19 @@ SOFTWARE.
...
+## semver
+
+This product contains 'semver' by Semver.
+
+"Semantic Versioning" or "SemVer" contain a set of rules and requirements that dictate how version numbers are assigned and incremented
+
+* HOMEPAGE:
+  * https://github.com/semver/semver
+
+* LICENSE:
```
amyblais, line 2867: No license

grundleborg: That repository contains no code, so I'm not really sure why it's being included in our stuff as a dependency.

grundleborg: Code appears to be unlicensed, but NPM lists it as MIT (the default).

amyblais:
@grundleborg Kind reminder to review (+merge and cherry-pick if approved)

NOTICE.txt
```diff
@@ -2871,6 +2855,19 @@ SOFTWARE.
...
+## semver
+
+This product contains 'semver' by Semver.
+
+"Semantic Versioning" or "SemVer" contain a set of rules and requirements that dictate how version numbers are assigned and incremented
+
+* HOMEPAGE:
+  * https://github.com/semver/semver
+
+* LICENSE: MIT
```
grundleborg, line 2867: @amyblais I've looked at this again, and I don't see any dependency in our package.lock.json on this repo. Are you sure it isn't `node-semver` (i.e. the one that is *removed* immediately above this block in the diff?

NOTICE.txt
grundleborg: (I think my second comment in this thread was written on the wrong pull request, sorry)

NOTICE.txt
amyblais: I can see it here: https://github.com/mattermost/mattermost-webapp/blob/release-5.25/package.json#L73. Not sure if I should just keep node-semver.

grundleborg: That's a package in npmjs called `semver`: https://www.npmjs.com/package/semver - if you look at the link to Github from there, it goes to github.com/npm/node-semver , not https://github.com/semver/semver

amyblais: Ok, removed ``semver`` and added back ``node-semver``.

PR #5805, opened 2020-06-25
Consitering strings in resolve element. 

crspeller:
#### Summary
Strings are used here in plugins and are not valid react elements. But they also can't be turned directly into components as `React.createElement("Not a component name")` thinks you mean `<Not a component name/>`


crspeller:
Merging, otherwise this breaks community.

PR #5806, opened 2020-06-26
MM-26075 Cypress/E2E: Standardize apiCreateTeam, apiDeleteTeam, apiGetTeamByName and apiGetAllTeams

saturninoabril:
#### Summary
Standardize returned data and add typings for intellisense of the ff custom API commands:
- apiCreateTeam
- apiDeleteTeam
- apiGetTeamByName
- apiGetAllTeams

#### Ticket Link
 JIRA ticket: https://mattermost.atlassian.net/browse/MM-26075


e2e/cypress/integration/enterprise/elasticsearch_autocomplete/renaming_spec.js
```diff
@@ -129,7 +129,7 @@ describe('Autocomplete with Elasticsearch - Renaming', () => {
             });
 
             // # Rename the team
-            cy.apiPatchTeam(testTeam.id, {name: 'updatedteam' + timestamp});
+            cy.apiPatchTeam(testTeam.id, {display_name: 'updatedteam' + timestamp});
```
saturninoabril, line 132: Team "name" is actually not allowed to be patched. Only the those included at https://github.com/mattermost/mattermost-webapp/compare/master...saturninoabril:MM-26075-e2e_standardize_api_commands?expand=1#diff-aa488add4a43ce7a79a0dadab2523052R57-R63 are allowed.

e2e/cypress/integration/enterprise/teams/search_teams_spec.js
```diff
@@ -40,8 +40,8 @@ describe('Search teams', () => {
         const displayName = uuidv4();
 
         // # Create a new team.
-        cy.apiCreateTeam('team-search', displayName).then((response) => {
-            this.createdTeamIDs.push(response.body.id);
+        cy.apiCreateTeam('team-search', displayName).then(({team}) => {
```
srkgupta, line 43: Wouldn't using just `team` work instead of `{team}`? Similarly using `teams` instead of `{teams}`

e2e/cypress/integration/integrations/message_to_channel_via_slash_command_spec.js
```diff
@@ -21,8 +21,8 @@ describe('Integrations', () => {
         cy.requireWebhookServer();
 
         // # Create new team and get off-topic channel
-        cy.apiCreateTeam('test-team', 'Test Team').then((teamResponse) => {
-            team = teamResponse.body;
+        cy.apiCreateTeam('test-team', 'Test Team').then(({team: newTeam}) => {
+            team = newTeam;
```
srkgupta, line 25: Shouldn't this be simply set as:
`cy.apiCreateTeam('test-team', 'Test Team').then((team) => {`

e2e/cypress/support/api/team.js
```diff
@@ -76,15 +60,23 @@ Cypress.Commands.add('apiGetTeamByName', (name) => {
...
+        let teams;
+        let totalCount;
+
+        if (includeTotalCount) {
+            teams = response.body.teams;
+            totalCount = response.body.total_count;
+        } else {
+            teams = response.body;
+        }
+        cy.wrap({teams, totalCount});
```
srkgupta, line 79: I did not see any usage of includeTotalCount parameter in any of the tests. Is this for the future?

e2e/cypress/integration/enterprise/teams/search_teams_spec.js
saturninoabril: That will work too. Yes, it could simply return [`response.body`](https://github.com/mattermost/mattermost-webapp/pull/5806/files#diff-bc91d52642ea6ce4595a81941caa05a6L60) and use directly by `.then((name-anything) => {..})` but we'll always second guess on what exactly it's being returned. `response.body` could be anything depending on API call.
The [new approach is being explicit](https://github.com/mattermost/mattermost-webapp/pull/5806/files#diff-bc91d52642ea6ce4595a81941caa05a6R48) on what it's returning and documenting as much as possible in `*.d.ts` for better dev experience.

e2e/cypress/integration/integrations/message_to_channel_via_slash_command_spec.js
saturninoabril: See above response.

e2e/cypress/support/api/team.js
saturninoabril: I understand and I'm a fan of not adding anything if not to be used. My primary goal is to match this custom API command based on corresponding [API reference](https://api.mattermost.com/#tag/teams/paths/~1teams/get). Though no usage currently, I can see that this will help other discover that there's an option like this whenever they need total count of teams. I'm fine to remove if you recommend to do so.

srkgupta: Yes, I think this would just add complexity to the test code. I am not sure of a usecase where this is essential to be validated from the test. It's required in UI, so that pagination can work, but from the test perspective, I feel its not required and will not be used.

saturninoabril: Done, PTAL again.

PR #5816, opened 2020-06-29
MM-26339 - Add context and CommandArgs to dynamic autocomplete calls

cpoile:
#### Summary
- Add what context is available to dynamic autocomplete calls. The server component of this PR (will link below) adds whatever else it can.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26339

#### Related Pull Requests
- Has server changes: https://github.com/mattermost/mattermost-server/pull/14940
- Has redux changes: https://github.com/mattermost/mattermost-redux/pull/1181


components/suggestion/command_provider.jsx
```diff
@@ -54,6 +54,17 @@ export class CommandSuggestion extends Suggestion {
 }
 
 export default class CommandProvider extends Provider {
+    constructor(props) {
```
crspeller, line 57: Maybe don't call them props. I thought this was a component for a second.

components/suggestion/command_provider.jsx
```diff
@@ -115,7 +126,14 @@ export default class CommandProvider extends Provider {
 
     handleWebapp(pretext, resultCallback) {
         const command = pretext.toLowerCase();
-        Client4.getCommandAutocompleteSuggestionsList(command, getCurrentTeamId(store.getState())).then(
+        const teamId = getCurrentTeamId(store.getState());
+
+        const args = {
+            channel_id: this.channelId,
+            ...(this.rootId && {root_id: this.rootId, parent_id: this.rootId}),
```
crspeller, line 133: This works? Maybe just make a conditional?

components/suggestion/command_provider.jsx
cpoile: @crspeller Sure, but I can't think of anything better. I'll follow your suggestion. 
(I was following https://github.com/mattermost/mattermost-webapp/blob/fca8b79c231b001aca4300846c621c564c15cbb7/components/suggestion/at_mention_provider/at_mention_provider.jsx#L18)

components/suggestion/command_provider.jsx
cpoile: It does! Isn't it awesome? At first I was adding it in a conditional, but I was like "there's got to be a better way to do this." I found it here: https://stackoverflow.com/a/40560953/1218899

components/suggestion/command_provider.jsx
crspeller: You could just de-structure them, so you don't have to name them. 
But if it's consistent with other places then I think it's probably fine.

components/suggestion/command_provider.jsx
crspeller: Never seen that before. I guess it's fine 😄

components/suggestion/command_provider.jsx
cpoile: @crspeller Yep, good idea, done. And I simplified it. No need to call a setProps function--the code is simple enough that people will know what's happening.

PR #5818, opened 2020-06-29
MM-26573 Support variations for one-byte emojis

hmhealey:
The original emoji parsing code I wrote expected unicode characters to be a fixed length (they aren't), and then when I mostly fixed that, I still left in logic that assumed that the first character of an emoji is a fixed two-byte length (it's not). This meant that most emojis with variations like skin tones were rendered correctly because most modern emoji code points (such as 👍) are two bytes long, but a few emojis came from older parts of the unicode specification (such as ✊ which is part of the Dingbats block) and those are only one byte long. For those one-byte emojis, we'd skip the second codepoint of the emoji which includes the skin colour selector in the case of ones like ✊🏿.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26573


jgilliam17:
@hmhealey Should I be able to use :+1_dark_skin_tone: on the test server and have it render correctly? All hand emojis with `dark_skin_tone` appended ? Or is it only  `:fist_raised:` emoji?

hmhealey:
Any emoji with skin tones should support those suffixes (`_light_skin_tone`, `_medium_light_skin_tone`, `_medium_skin_tone`, `_medium_dark_skin_tone`, and `_dark_skin_tone`), but those don't appear in the autocomplete. They've been around for a while, but I was just using them here to show that the unicode emojis should work the same as ones using the emoji's name.

I'm not sure why `+1` doesn't work. It exists in the emoji data file, but it's not recognized by the web app. I think it's something to do with `:+1:` normally being handled by the code that handles emoticons like `:)`. I'll file a followup

PR #5821, opened 2020-06-30
[MM-26493] Allow incoming webhooks owned by bot to override username

larkox:
#### Summary
Allow incoming webhooks owned by bot to override username

#### Ticket Link
https://mattermost.atlassian.net/secure/RapidBoard.jspa?rapidView=33&modal=detail&selectedIssue=MM-26493
Fix https://github.com/mattermost/mattermost-server/issues/14801

#### Related Pull Requests
None

#### Screenshots
None


DHaussermann:
@mickmister  or @crspeller can I please get some testing guidance on the setup. How would a bot be the owner of a webhook created on the Mattermost side?  
My undemanding is that webhooks are owned by their creators.

mickmister:
@DHaussermann This is what I'm thinking:

- start with a fresh cloud server
- make a user with the username `github`
- sign in with that user and make an incoming webhook
- convert the user to a bot with `/cloud mmcli (cloud server name) user convert github --bot`
- send a request to the webhook with `username` set in the payload, like it is set to `My fancy username` here [issue-mattermost-server-14801](https://github.com/mattermost/mattermost-server/issues/14801)
- Notice that it still says `github` as the bot's name in the resulting post, when you supply a different value in the webhook payload's `username` field

DHaussermann:
@larkox this solution does work. Webhooks owned by a bot can now user the username overwrite (both from MM and the payload). 
But, I see a regression here. Webhooks owned by bot now show duplicate `BOT` tag when the webhook posts. This does not reproduce on other branches.
ExtraBotTag

larkox:
@crspeller @mickmister Apparently, the reason not to process the bot is that the bot was already setting the bot badge by itself on the UserProfile component. The fix was simple, just don't add another bot badge if it is already a bot. Nevertheless, mentioning you in case you want to make any dev comment on the solution.

@DHaussermann This should be ready to review again.

DHaussermann:
Thanks @larkox taking a look now. 
Would the last commit also require another round of dev review?

larkox:
I personally think it is not necessary, but that is one of the reasons I mentioned both of them, in case they deem it necessary.

PR #5833, opened 2020-07-02
Select Team list container should scale in width based on browser window width #14197

sridhar02:
#### Summary
changes done for the issues mentioned in the ticket

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/14197
  JIRA https://mattermost.atlassian.net/browse/MM-23749

#### Related Pull Requests

#### Screenshots

* overflow-y addition on the mobile view 
image

* Before the change when the max-width was 400px 

image

* after the max-width is set to 800px

image

### DIFF
--- a/components/select_team/select_team.jsx
+++ b/components/select_team/select_team.jsx
@@ -336,7 +336,7 @@ export default class SelectTeam extends React.PureComponent {
                 {headerButton}
                 <div className='col-sm-12'>
                     <div
-                        className={'signup-team__container'}
+                        className={'select-team__container signup-team__container'}
                     >
                         <img
                             alt={'signup team logo'}
--- a/sass/responsive/_mobile.scss
+++ b/sass/responsive/_mobile.scss
@@ -1886,10 +1886,6 @@
         max-height: calc(75vh - (118px + 187px));
     }
     
-    .signup-team-all .infinite-scroll {
-        overflow-y: auto;
-    }
-    
 }
 
 @media screen and (max-width: 640px) {
--- a/sass/responsive/_mobile.scss
+++ b/sass/responsive/_mobile.scss
@@ -1885,6 +1885,11 @@
     .signup-team-all {
         max-height: calc(75vh - (118px + 187px));
     }
+    
+    .signup-team-all .infinite-scroll {
+        overflow-y: auto;
+    }
+    
 }
 
 @media screen and (max-width: 640px) {
--- a/components/select_team/__snapshots__/select_team.test.jsx.snap
+++ b/components/select_team/__snapshots__/select_team.test.jsx.snap
@@ -10,7 +10,7 @@ exports[`components/select_team/SelectTeam should match snapshot 1`] = `
     className="col-sm-12"
   >
     <div
-      className="signup-team__container"
+      className="select-team__container signup-team__container"
     >
       <img
         alt="signup team logo"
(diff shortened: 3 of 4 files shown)


esethna:
Thanks @sridhar02, looks good from my end! Adding @andrewbrown00 for a UX review

jgilliam17:
I noticed login screen email/password fields width has changed on the test server.
Is this intentional? 
Screen Shot 2020-07-10 at 12 16 30 PM

sridhar02:
@jgilliam17,actullay it was mentioned in the task to change the width of the teams to 800px, I have changed it and it is effecting the sign-in also as they are using the same styles for both the teams and the sign-in pages.
image

andrewbrown00:
@sridhar02 thanks for the contribution here 👍  

Changes to the `Teams` page looks good ✅ 

image

**Item to resolve**
- [ ] Revert the `Login` page back the previous width, it's too wide

Current | Expected (previous)
----- | -----
image | image

jgilliam17:
Thank you @sridhar02 
Team list page looks good.

@andrewbrown00 found two more instance where field width was increased. Should this also be reverted? Team URL screen actually looks better with the wider field, where most of the url is visible. 

Create a Team 

Screen Shot 2020-07-14 at 10 31 29 AM

Team URL screen 

Screen Shot 2020-07-14 at 10 32 12 AM

andrewbrown00:
> @andrewbrown00 found two more instance where field width was increased. Should this also be reverted? Team URL screen actually looks better with the wider field, where most of the url is visible.

Thanks @jgilliam17, yes, all non-team page instances should be reverted 🙂

sridhar02:
@jgilliam17 @andrewbrown00 I have reverted the changes and made changes that will be applicable to the team container u can check it.

andrewbrown00:
> @jgilliam17 @andrewbrown00 I have reverted the changes and made changes that will be applicable to the team container u can check it.

Looks good from my end. Thanks for quick updates 🙂 , I'll let @jgilliam17 do her review too.

components/select_team/select_team.jsx
```diff
@@ -336,7 +336,7 @@ export default class SelectTeam extends React.PureComponent {
                 {headerButton}
                 <div className='col-sm-12'>
                     <div
-                        className={'signup-team__container'}
+                        className={'select-team__container signup-team__container'}
```
calebroseland, line 339: Thanks for isolating, I've been thinking we need a unique class to distinguish some of these screens.

sass/responsive/_mobile.scss
```diff
@@ -1851,6 +1851,11 @@
     .signup-team-all {
         max-height: calc(75vh - (118px + 187px));
     }
+    
+    .signup-team-all .infinite-scroll {
+        overflow-y: auto;
+    }
```
calebroseland, line 1857: This selector doesn't seem to hit the element I presume that it targets.

```html
<div class="infinite-scroll signup-team-all">
```
?
```css
.signup-team-all.infinite-scroll {
    overflow-y: auto;
}
```

digging into this further, though.. I don't know that this is even needed, as the following already kicks in on hover inside (e.g. you start to drag on mobile)
```css
.infinite-scroll:hover {
    overflow-y: scroll;
    overflow-y: overlay;
}
```

sridhar02: yes, I have not checked in that way, I think it working on the hover, so I deleted this class now 
```js
.signup-team-all.infinite-scroll {
    overflow-y: auto;
}
```

sudheerDev:
Thank you @sridhar02

PR #5835, opened 2020-07-02
MM-26597/MM-26672/MM-26616/MM-26605 Properly forward ref to Textbox

hmhealey:
Previously, we could attach a ref to any `connect`-based component fine, but that's no longer possible since the new version of `react-redux` changed `connect` to generate a functional component, so we have to forward the ref to the underlying component to be able to do anything with it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26616
https://mattermost.atlassian.net/browse/MM-26597
https://mattermost.atlassian.net/browse/MM-26672
https://mattermost.atlassian.net/browse/MM-26605

### DIFF
--- a/components/edit_post_modal/edit_post_modal.jsx
+++ b/components/edit_post_modal/edit_post_modal.jsx
@@ -337,9 +337,7 @@ class EditPostModal extends React.PureComponent {
     }
 
     setEditboxRef = (ref) => {
-        if (ref && ref.getWrappedInstance) {
-            this.editbox = ref;
-        }
+        this.editbox = ref;
 
         if (this.editbox) {
             this.editbox.focus();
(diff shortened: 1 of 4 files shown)


hmhealey:
Can't make an omelette without breaking a few refs 😛

hmhealey:
Snuck in a couple other ref fixes since I was seeing some more warnings on master.

jgilliam17:
@hmhealey I am seeing an error when posting any message, not just ones with the emoji
Screen Shot 2020-07-02 at 2 06 00 PM
Another error when clicking on reply arrow to open RHS, combined with blank screen
Screen Shot 2020-07-02 at 2 09 40 PM

Also, this repeat error when closing edited post using Enter
Screen Shot 2020-07-02 at 2 13 54 PM

hmhealey:
The one that you saw when sending a message is from Firefox. I think it might have something to do with your password manager. I've seen it since I switched to Firefox, and that was on LastPass

hmhealey:
@jgilliam17 That error when opening the RHS should be fixed

PR #5837, opened 2020-07-02
Added Cypress test MM-T1414 and MM-T1417

srkgupta:
#### Summary
Added couple of Cypress test for Guest Account Feature

#### Ticket Link
JIRA Ticket: [MM-26072](https://mattermost.atlassian.net/browse/MM-26072)
TM4J Tickets: [MM-T1414](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T1414) and [MM-T1417](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T1417)

### DIFF
--- a/e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js
+++ b/e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js
@@ -204,7 +204,7 @@ describe('Guest Account - Guest User Experience', () => {
             cy.reload();
 
             // # As a sysadmin, add the guest user to this team
-            cy.addUserToTeam(teamTwo.id, guestUser.id).then(() => {
+            cy.externalAddUserToTeam(teamTwo.id, guestUser.id).then(() => {
                 cy.get(`#${teamTwo.name}TeamButton`).should('be.visible').click();
 
                 // * Verify if Channel Not found is displayed
--- a/e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js
+++ b/e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js
@@ -15,6 +15,17 @@
  */
 import * as TIMEOUTS from '../../../fixtures/timeouts';
 
+function demoteGuestUser(guestUser) {
+    // # Demote user as guest user before each test
+    cy.apiAdminLogin();
+    cy.apiGetUserByEmail(guestUser.email).then((emailResponse) => {
+        const user = emailResponse.body;
+        if (user.roles !== 'system_guest') {
+            cy.demoteUser(guestUser.id);
+        }
+    });
+}
+
 describe('Guest Account - Guest User Experience', () => {
     let guestUser;
 
@@ -180,5 +191,30 @@ describe('Guest Account - Guest User Experience', () => {
 
         // # Close the profile popover
         cy.get('#channel-header').click();
+    });
+
+    it('MM-T1417 Add Guest User to New Team from System Console', () => {
+        // # Demote Guest user if applicable
+        demoteGuestUser(guestUser);
+
+        // # Ceate a new team
+        cy.apiCreateTeam('test-team2', 'Test Team2').then(({team: teamTwo}) => {
+            // # Login as guest user
+            cy.apiLogin(guestUser);
+            cy.reload();
+
+            // # As a sysadmin, add the guest user to this team
+            cy.addUserToTeam(teamTwo.id, guestUser.id).then(() => {
+                cy.get(`#${teamTwo.name}TeamButton`).should('be.visible').click();
+
+                // * Verify if Channel Not found is displayed
+                cy.findByText('Channel Not Found').should('be.visible');
+                cy.findByText('Your guest account has no channels assigned. Please contact an administrator.').should('be.visible');
+                cy.findByText('Back').should('be.visible').click();
+
+                // * Verify if user is redirected to a valid channel
+                cy.findByTestId('post_textbox').should('be.visible');
+            });
+        });
     });
 });
--- a/e2e/cypress/support/api_commands.js
+++ b/e2e/cypress/support/api_commands.js
@@ -662,6 +662,20 @@ Cypress.Commands.add('promoteUser', (userId) => {
     cy.externalRequest({user: admin, method: 'post', baseUrl, path: `users/${userId}/promote`});
 });
 
+/**
+ * Add a user to a team directly via API
+ * @param {String} teamId - The team ID
+ * @param {String} userId - The user ID
+ * All parameter required
+ */
+
+Cypress.Commands.add('addUserToTeam', (teamId, userId) => {
+    const baseUrl = Cypress.config('baseUrl');
+    const admin = getAdminAccount();
+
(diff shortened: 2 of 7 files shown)


e2e/cypress/support/api_commands.js
```diff
@@ -662,6 +662,20 @@ Cypress.Commands.add('promoteUser', (userId) => {
...
 });
 
+/**
+ * Add a user to a team directly via API
+ * @param {String} teamId - The team ID
+ * @param {String} userId - The user ID
+ * All parameter required
+ */
+
+Cypress.Commands.add('addUserToTeam', (teamId, userId) => {
```
josephbaylon, line 672: Might want to update from master since this api command was added in this PR https://github.com/mattermost/mattermost-webapp/pull/5831

srkgupta: @josephbaylon this one is little different. An external request is made as sysadmin when the user is already on a page.

josephbaylon: @srkgupta I see. I think it would be a little confusing later. I did a check on the codebase and all commands/task that use `cy.externalRequest(...)` and/or `cy.task('externalRequest', ...)` require admin login. Not necessary for this PR, but might be good to start prefixing externalRequests either `adminXXX` or `externalXXX` just so it is differentiated from `apiXXX`.

saturninoabril: +1 to Joseph. I'm leaning towards `external` prefix. Also, please don't add any command to `api_commands.js`, we'll deprecate it soon. We're migrating/organizing commands into `support/api/*.js`. For this command, it might not fit to that folder and probably better to create a new one, like `support/external_commands.js`. Lastly, please add typings `support/external_commands.d.ts`

e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js
```diff
@@ -15,6 +15,17 @@
  */
 import * as TIMEOUTS from '../../../fixtures/timeouts';
 
+function demoteGuestUser(guestUser) {
+    // # Demote user as guest user before each test
+    cy.apiAdminLogin();
+    cy.apiGetUserByEmail(guestUser.email).then((emailResponse) => {
```
saturninoabril, line 21: `cy.apiAdminLogin()` and `cy.apiGetUserByEmail` are not necessary since `guestUser` is a user profile object with `id` on it.

srkgupta: @saturninoabril I am trying to fetch guestUser object again as this might have been changed within the test, especially the second test. Hence if the guestUser was promoted as Regular member in a previous test, I am demoting the user as guest again. This is intentional.

e2e/cypress/support/api_commands.js
srkgupta: Thanks @josephbaylon and @saturninoabril. There are some other APIs following the same naming conventions which might also need to be updated and is beyond the scope of this PR. Can we do the cleanup as a separate PR instead.

e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js
saturninoabril: Honestly, the conditional only complicate things IMO. I'm on the fence of not caring whatever roles it may have in the first place if all I need is to ensure that the user is guest before the test run on specific case.
```javascript
if (user.roles !== 'system_guest') {
    cy.demoteUser(guestUser.id);
}
```

srkgupta: This is handled because if we try to promote a user who is already a guest, the external request throws an error, which we do not want. Hence this specific check. 😄

e2e/cypress/support/api_commands.js
saturninoabril: I agree on other APIs and indeed it should be on a separate PR. But since you're introducing a new custom command, I suggest to start the recommended way here.

srkgupta: @saturninoabril and @josephbaylon, Updated the code as suggested. Please recheck.

e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js
saturninoabril: What error are you seeing? I tried it locally and works fine for me. Probably needs to login as admin before creating a team.

e2e/cypress/support/external_commands.d.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// ***************************************************************
+// Each command should be properly documented using JSDoc.
+// See https://jsdoc.app/index.html for reference.
+// Basic requirements for documentation are the following:
+// - Meaningful description
+// - Each parameter with `@params`
+// - Return value with `@returns`
+// - Example usage with `@example`
+// Custom command should follow naming convention of having `visual` prefix, e.g. `visualEyesOpen`.
```
saturninoabril, line 14:
```suggestion
// Custom command should follow naming convention of having `external` prefix, e.g. `externalAddUserToTeam`.
```

e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js
srkgupta: @saturninoabril This is the error the API returns if the user is already a guest user and if we try to demote the user as guest again. 

```ERRO[2020-07-03T18:38:17.126246+05:30] Unable to convert the user to guest because is already a guest.  caller="mlog/log.go:175" err_details= err_where=Api4.demoteUserToGuest http_code=501 ip_addr=127.0.0.1 method=POST path=/api/v4/users/an5hee4of7r9tdicee6yrscfsa/demote request_id=jub94jmdrtbgif9qnaawcgepzc user_id=d4ymchskjf8xjp4k5hudnn37uc```

Screenshot 2020-07-03 at 6 39 02 PM

File: e2e/cypress/support/external_commands.d.ts (hunk 3 above)
Comment (srkgupta) [reply e2e/cypress/support/external_commands.d.ts]: @saturninoabril, Good catch. Fixed this. Please recheck.

File: e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js (hunk 2 above)
Comment (saturninoabril) [reply e2e/cypress/integration/enterprise/guest_accounts/guest_experience_ui_spec.js:21]: Oh you're right. Thanks for checking. Now I got the same error as you posted. I'm good as is.

File: e2e/cypress/support/external_commands.d.ts [hunk 4]
```diff
@@ -0,0 +1,31 @@
...
+         * @param {String} teamId - The team ID
+         * @param {String} userId - The user ID
+         * All parameter required
+         *
+         * @example
+         *   cy.externalAddUserToTeam('team-id', 'user-id');
+         */
+        externalAddUserToTeam(teamId: string, userId: string): Chainable;
+    }
+}
```
saturninoabril, line 31: 🎉

josephbaylon: Thanks!

PR #5838, opened 2020-07-02
Fix random dev mode warnings

hmhealey:
Fixing some more warnings that I found when working on #5835. There are a few more warnings coming from the new sidebar code, but I don't want to change them yet since they'll likely conflict with the feature branch.

### DIFF
--- a/components/create_comment/create_comment.jsx
+++ b/components/create_comment/create_comment.jsx
@@ -196,8 +196,6 @@ class CreateComment extends React.PureComponent {
          */
         selectedPostFocussedAt: PropTypes.number.isRequired,
 
-        isMarkdownPreviewEnabled: PropTypes.bool.isRequired,
-
         /**
          * Function to set or unset emoji picker for last message
          */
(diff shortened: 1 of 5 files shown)


hmhealey:
Skipping QA review since these warnings only appear during development.

components/create_comment/create_comment.jsx
```diff
@@ -196,8 +196,6 @@ class CreateComment extends React.PureComponent {
          */
         selectedPostFocussedAt: PropTypes.number.isRequired,
 
-        isMarkdownPreviewEnabled: PropTypes.bool.isRequired,
```
calebroseland, line 199: Did this logic get moved elsewhere in the past?

hmhealey: I assume you found it since you crossed out your comment, but for anyone looking in the future, `isMarkdownPreviewEnabled` is already provided by `TextboxLinks`'s `mapStateToProps`, so this was redundant.

PR #5839, opened 2020-07-02
Smart Paste feature

jespino:
#### Summary
Initial version of the smart paste feature.


components/create_comment/create_comment.jsx
```diff
@@ -92,6 +92,16 @@ class CreateComment extends React.PureComponent {
          */
         ctrlSend: PropTypes.bool,
 
+        /**
+         *  Flag used for smart paste feature
+         */
+        smartPaste: PropTypes.bool,
```
jespino, line 98: This are the settings get from the user preferences

components/create_comment/create_comment.jsx
```diff
@@ -282,6 +293,8 @@ class CreateComment extends React.PureComponent {
         this.focusTextbox();
         document.addEventListener('paste', this.pasteHandler);
         document.addEventListener('keydown', this.focusTextboxIfNecessary);
+        document.addEventListener('keyup', this.setIsShiftPressed);
```
jespino, line 296: Detecting the shift key pressed or not to allow "not formatted paste"

components/create_comment/create_comment.jsx
```diff
@@ -360,38 +379,21 @@ class CreateComment extends React.PureComponent {
...
-        const table = getTable(clipboardData);
-        if (!table) {
+    pasteHandler = (e) => {
+        if (!e.clipboardData || !e.clipboardData.items || e.target.id !== 'reply_textbox' || this.state.isShiftPressed || (!this.props.smartPaste && !this.props.smartPasteCodeBlocks)) {
             return;
         }
 
         e.preventDefault();
-
-        const {draft} = this.state;
```
jespino, line 376: All this codes goes to the smart paste logic.

components/create_post/create_post.jsx
```diff
@@ -413,8 +429,8 @@ class CreatePost extends React.PureComponent {
             orientation = window.screen.orientation.type.split('-')[0];
         }
 
-        if (this.lastOrientation && orientation !== this.lastOrientation && (document.activeElement || {}).id === 'post_textbox') {
-            this.refs.textbox.blur();
+        if (this.textbox.current && this.lastOrientation && orientation !== this.lastOrientation && (document.activeElement || {}).id === 'post_textbox') {
```
jespino, line 432: I change this into a proper ref because was failing on master for me.

components/user_settings/advanced/user_settings_advanced.jsx
```diff
@@ -458,6 +510,129 @@ export default class AdvancedSettingsDisplay extends React.PureComponent {
...
+                                name='smartPaste'
+                                checked={smartPasteActive[1]}
+                                onChange={() => {
+                                    this.updateSetting('smart_paste', 'true');
+                                    this.updateSetting('smart_paste_code_blocks', 'false');
+                                }}
+                            />
+                            <FormattedMessage
+                                id='user.settings.advance.smartPasteOnWithoutCodeBlocks'
+                                defaultMessage='Enable smart paste without auto-detection of code blocks'
```
jespino, line 562: @esethna Around here are the tests shown in the account settings screen to configure smart paste, probably would be great if you can review them. (Probably is easier to get a spinwick running and check the texts there).

esethna:
@jespino can you please help write out the expected functions of this feature? and what should QA be on the lookout for?

jespino:
@esethna Ok, the main feature of this is coping from different places into mattermost and keeping the format. This is done through turndown (the expectation is not to have a perfect match) is more about having a good enough match. It is going to fail sometimes generating accurate representation of the pasted thing, but should be acceptable.

The feature can be enable/disable/partially-enabled, so that should be tested.

The feature support "not-formatted-text" with CTRL|CMD+Shift+V.

There is a feature of language detection that detects some languages (Java, Javascript, C, Go, Rust...) but is not a feature that is very reliable. It's a matter of best effort, and if it fails, you can always bypass it using CTRL+SHIFT+V (Or disabling it from your settings).

I think that is all, at the end of the day, the QA testing should be around changing the settings, and verify the settings changes work properly, and copy and pasting.

esethna:
@jespino 

> There is a feature of language detection that detects some languages (Java, Javascript, C, Go, Rust...) but is not a feature that is very reliable. It's a matter of best effort, and if it fails, you can always bypass it using CTRL+SHIFT+V (Or disabling it from your settings).

If it fails the language detection can we just do a normal code block (not syntax highlighted)?

jespino:
@esethna you can't do that, because we don't know what kind of text is it, you can paste a rhyme from Becker, a recipe from your grandmother or some rust code, If I detect is rust code, I'll create the code block, if not, I'll paste regular text.

esethna:
Got it, thanks, just wanted to understand the fall-back (regular text)

esethna:
@jespino when you're ready can you trigger the test server here?

components/create_comment/create_comment.jsx
Willyfrog: I think the standard for mac is different: Option-Shift-Command-V

also, if I press and release shift (or any combination with shift) and I right-click->paste, would it count as shift+paste?

jespino:
@esethna still polishing some things, but I think tomorrow is going to be ready

stylianosrigas:
I added again the label after solving the no capacity issue. Feel free to remove when you are done with your testing. Sorry for any inconvenience.

components/create_comment/create_comment.jsx
jespino: This handle the keydown and the keyup, so is going to work fine meanwhile you are not pasting and at the same time executing a combination of keystrokes that involves shift (That would be pretty hard to do at the same time).

I going to add the mac support for this.

jespino:
@esethna I added the undo feature, and added some "extensions" to better support of copy/paste inside mattermost (for example, metions, channel mentions and labels work as expected now).

esethna:
This is awesome!! @jespino can we just turn it default on and remove the account setting? Account settings are barriers to good UX becuase users rarely edit the default settings. If not, can you help me understand why it's needed and/or what we are worried about if we turned it default on and removed the setting? 

_____________________

Let's get QA to bash on it. cc// @jgilliam17 Things to please pay attention to during testing:

1. Does this break our existing feature of copying/pasting tables/excel selections
2. Copying/pasting outside Mattermost: Please test copying from various sources (ie google docs, icloud, MS word, excel, submile text etc) then pasting into Mattermost to try to break it and find limitations.
3. Copying/pasting within Mattermost: Is all the formatting retained in all cases of markdown including embedded images etc? What if I copy multiple posts? or copy partial posts? 
4. Pasting without formatting (CTRL + Shift + V)

esethna:
Discussed with @jespino. Proposal is:

1. Remove the account settings for this and default it on
2. Merge this PR to master AFTER the v5.26 branch is cut so it soaks on daily for a full two months
3. Ship in v5.28

^ Above is dependent on discussion with apps team and PR needs dev review to uncover potential corner cases etc.

esethna:
@amyblais can you please add the v5.28 milestone and then add it to this PR?

amyblais:
@metanerd Please help with ^ https://github.com/mattermost/mattermost-webapp/pull/5839#issuecomment-655807110

esethna:
@jgilliam17 See comments here about testing, let's please get this in the queue after v5.26 testing

esethna:
@jgilliam17 any estimates on when we can get this PR tested?

jgilliam17:
> @jgilliam17 any estimates on when we can get this PR tested?

It's in testing queue for this week, after 5.26 PRs are done

NOTICE.txt
```diff
@@ -1270,6 +1270,109 @@ SOFTWARE.
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+---
+
+## turndown-plugin-gfm
```
hmhealey, line 1308: Just to note, we have some small differences from GFM (most notably around list endings), but that shouldn't matter too much here

types/external/lang-detector.d.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+declare module 'lang-detector';
```
hmhealey, line 4: Are there any types we need to define here for use when the components are eventually converted to TypeScript?

types/external/turndown-plugin-gfm.d.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+declare module 'turndown-plugin-gfm';
```
hmhealey, line 4: Same here

utils/smartpaste/githubcode.tsx
```diff
@@ -0,0 +1,23 @@
...
+// See LICENSE.txt for license information.
+
+import {Rule} from 'turndown';
+
+export function githubCodeTurndownRuleBuilder(hasFirstPart: boolean, hasLastPart: boolean, text: string): Rule {
+    return {
+        filter: (node: Node): boolean => {
+            return (node.parentNode as HTMLElement)?.id === 'turndown-root' && node.nodeName === 'TABLE' && (/\b(js|blob|diff)-./).test((node as HTMLTableElement).className);
+        },
+        replacement: (): string => {
```
hmhealey, line 11: What is the `NEW_LINE_REPLACEMENT` keyword being used for? Could we maybe have a comment explaining it either here or

utils/smartpaste/githubcode.tsx
```diff
@@ -0,0 +1,23 @@
...
+            let result = '';
+            if (hasFirstPart) {
+                result += '#*#*NEW_LINE_REPLACEMENT*#*#';
```
hmhealey, line 14: Could this be turned into a named constant?

utils/smartpaste/htmlfix.tsx
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export function fixNestedLists(doc: Document): Document {
```
hmhealey, line 4: Could you also add a comment explaining what this is doing?

It looks like something isn't wrapping nested lists in `<li>` tags correctly. Is that something we're doing wrong elsewhere or is that coming from somewhere else? We don't need to fix it now, obviously, but I'm curious if that's something that our markdown code is rendering wrong.

utils/smartpaste/mattermost.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const hashtagsRule = {
+    filter: (node: Node): boolean => {
+        return node.nodeName === 'A' && (node as HTMLElement).className === 'mention-link' && Boolean((node as HTMLElement).getAttribute('data-hashtag'));
+    },
+    replacement: (content: string, node: Node): string => {
+        return (node as HTMLElement).getAttribute('data-hashtag') || '';
+    }
+};
+
+export const filePreviewButtonRule = {
```
hmhealey, line 28: What's this rule used for?

utils/smartpaste/mattermost.tsx
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Options} from 'turndown';
```
hmhealey, line 4: Do we need any special rules for elements that are replaced by `messageHtmlToComponent` like emojis and at-mentions?

utils/smartpaste/mattermost.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const codeBlockRule = {
+    filter: (node: Node) => {
+        return node.nodeName === 'CODE' && (node as HTMLElement).className.indexOf('hljs') !== -1;
+    },
+
+    replacement: (content: string, node: Node, options: Options) => {
+        const language = ((node as HTMLElement).className.match(/language-(\S+)/) || [null, ''])[1];
+        const code = node.textContent || '';
+
+        const fenceChar = options.fence?.charAt(0) || '`';
```
hmhealey, line 46: Do we need different handling for when a code block is fenced with a ~ instead of a backtick? You might need that if, for example, you're quoting code containing triple backticks like

    ~~~
    Here's how to use a code block:
    ```java
    public static void main(String[] args) {
        System.out.Println("Hello, World!");
    }
    ```
    ~~~

hmhealey: Oh, I think I see. This code is handling that by increasing the size of the fence? I didn't actually know that was possible. I thought you had to do stuff with tildes or indented code blocks for that to happen.

utils/smartpaste/githubcode.tsx
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Rule} from 'turndown';
+
+export function githubCodeTurndownRuleBuilder(hasFirstPart: boolean, hasLastPart: boolean, text: string): Rule {
+    return {
+        filter: (node: Node): boolean => {
+            return (node.parentNode as HTMLElement)?.id === 'turndown-root' && node.nodeName === 'TABLE' && (/\b(js|blob|diff)-./).test((node as HTMLTableElement).className);
```
hmhealey, line 9: Why are we checking the classname of the node before determining if we should run this rule?

utils/undoredo/undoredo.ts
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+// This code is heavily based on the undoredu.js library from
+// https://github.com/iMrDJAi/UndoRedo.js with MIT license and copyrighted by
```
hmhealey, line 5: I think this probably needs to be added to the NOTICE.txt even if we're using a modified version of the library in-line

hmhealey: Also, what are we changing here? Maybe we could make a note of it in case we ever need to update from the upstream library.

cpanato:
if need the test server again please re-add the label

utils/smartpaste/githubcode.tsx
jespino: Sure, I can add a comment, it is because the turndown conversation swallow new lines and sometimes I need to generate them in the final result.

utils/smartpaste/githubcode.tsx
jespino: Yes I'll do it

utils/smartpaste/htmlfix.tsx
jespino: No, it comes from spreadsheets mainly. Turndown doesn't support that, because is not valid, but most browsers accept it, I can add a comment explaining the why

types/external/lang-detector.d.ts
jespino: I was building a proof of concept, so I didn't defined the types.

utils/smartpaste/githubcode.tsx
jespino: This rule is a migration from the GitHub code paste support that existed before, that check is the same one that was already there but now though turndown library.

utils/smartpaste/mattermost.tsx
jespino: No, they are automatically covered (I already tested it)

NOTICE.txt
jespino: This is because it helps a lot, but we can iterate and overwrite where is needed. It is a convenient way to support more things, but we will need to iterate and fix corner cases for sure.

types/external/lang-detector.d.ts
jespino: Done

types/external/turndown-plugin-gfm.d.ts
jespino: Done

utils/smartpaste/githubcode.tsx
jespino: I added a comment in the place where the replacement of the NEW_LINE_REPLACEMENT is done.

utils/smartpaste/githubcode.tsx
jespino: Done.

utils/smartpaste/mattermost.tsx
jespino: This fix a problem related to coping and pasting markdown images from the existing posts. I added a comment in the code, to clarify that.

utils/undoredo/undoredo.ts
jespino: We are changing a big chunk of the code, I leave that message there to honor the license, I can add this to the NOTICE.txt.

jespino: Done

utils/smartpaste/mattermost.tsx
jespino: Yes, exactly.

utils/smartpaste/htmlfix.tsx
hmhealey: It's fun having to support crappy, malformed HTML all the time, isn't it? :p

jgilliam17:
Thank you @jespino 
Apologize about delay in getting to this PR
Few things I noticed:
1. Excel sheet pastes as both spreadsheet table (same as before) and as an additional screenshot/image 
Google sheets only posts as table as before 
Screen Shot 2020-08-12 at 4 40 18 PM

2. Currently on community, we are preventing some elements from being selected for copying when selecting multiple posts: Profile images, Reply count, “(edited)" text "Commented on user_names's message:" text 
Smart paste selects all  https://mattermost.atlassian.net/browse/MM-24154
Screen Shot 2020-08-12 at 6 41 12 PM

3. Copying from html email renders as table. I verified that this is the same functionality as we currently have on community servers, just noting here - maybe experience can be improved

Screen Shot 2020-08-12 at 6 48 25 PM
Screen Shot 2020-08-12 at 6 47 57 PM

4. Unable to `Undo` paste of webpage or HTML emails using cmd+z 
5. After pasting from any website, open the same post to edit, select all the text in the modal (cmd+a, followed by cmd+c) close edit post and paste the text into text box - > formats in c++ and or java sometimes Screen Shot 2020-08-12 at 6 20 56 PM
6. Repeat steps from 5. for posted snippet from the google sheet 
Formats  as ```java 
Screen Shot 2020-08-12 at 6 45 46 PM

jespino:
@jgilliam17 
1. Can you copy from excel and paste here: https://evercoder.github.io/clipboard-inspector/ and sent me an screenshot (or confirm that you are receiving an html format and an image format in the paste?
2. We can omit that cases in our smart paste, but the future plans is to allow you to have a smarter way of paste conversations from mattermost. That way we should be able to show avatars and other things when is needed. But for now we can omit what is not need. I'll add the changes.
3. This is expected, the expectation is to have a "good enough" paste, sometimes that means it is not pasting what you expected because the original generated html is weird, like missusing tables.
4. It should work, I going to investigate that.
5 and 6. This is related to the autodetect for languages, probably I have to adjust better the thresholds for what the system consider a valid guess for the language.

jgilliam17:
@jespino Here are 2 screenshots for point #1 . There's quite a bit of text after first screenshot and the last one below, I can pasted on the google doc if you need every bit of it.
Screen Shot 2020-08-14 at 10 47 56 PM


Screen Shot 2020-08-14 at 10 47 16 PM

jespino:
@jgilliam17 That is exactly what I expected, microsoft excel is providing both, a text version and an image version.

@esethna we have 4 options here:
  1. Make always html primary, so if it has html and image, I going to use the html format to smart paste it.
  2. Make always file format primary, so if it has html and image, I going to attach the image and ingore the smart paste.
  3. Keep both (I don't think that is expected).
  4. Thread this as an exception and handle the microsoft excel documents with the option 1 or the option 2. And take a decision about using 1, 2 or 3 for the rest of the documents.

The problem with the "generic" approach is we can end up finding that what is sensible for microsoft documents (probably to smart paste it) can be really weird in other scenarios (for example if some image based program adds the html with a link to the image and the image itself).

jespino:
@jgilliam17 point 4 (Undo) is working for me. Can you give me more info about your system? Ctrl-Z is for Windows and Linux, it should be Comand+Z in MacOSX

jespino:
@jgilliam17 The points 5 and 6 should be addressed now.

esethna:
> we have 4 options here:
> 
> 1. Make always html primary, so if it has html and image, I going to use the html format to smart paste it.
> 2. Make always file format primary, so if it has html and image, I going to attach the image and ingore the smart paste.
> 3. Keep both (I don't think that is expected).
> 4. Thread this as an exception and handle the microsoft excel documents with the option 1 or the option 2. And take a decision about using 1, 2 or 3 for the rest of the documents.

@jespino I think we should take the text only, not the image

jespino:
@esethna totally agree, but that is not exactly my question. I think that should be the case for excel, my question is about all the paste, approaches. If I have html format and file in the paste, what should I use? should I use the image or the html? should we make an exception for excel to that rule?

I going to make the changes to not paste any file if the html is there, and then we can see if that generate problems and where.

stylianosrigas:

jgilliam17:
@jespino 
Verified excel spreadsheet pastes as grid now only, without an additional image

1. Undo paste works for me on Safari, Chrome and Firefox. I am seeing the issue on Desktop App 4.5.3  (macOS Catalina, latest version)  Unable to undo paste using command+z on desktop app
2. Copy Invite member link and paste into text field  Pastes as Python, instead of link 
Screen Shot 2020-08-27 at 8 59 23 AM

3. Is it possible to exclude date and new message separator lines and text when copying multiple posts from Mattermost?
Screen Shot 2020-08-27 at 9 16 15 AM
4. I still see an issue when copying and pasting from Edit post modal. e.g. pasted email, copies and pastes from edit modal in ```go 
Screen Shot 2020-08-27 at 10 41 33 AM

jespino:
@jgilliam17 
1. should be handled with some kind of changes in the desktop app, I'm talking with @Willyfrog about it.
2. and 4. Are related to automatic code detection, that I have deactivated for now, looks like is a bit buggy.
3. Is fixed now.

jgilliam17:
@jespino Thank you, I'll take a look
- For @sysadmin user - not sure if it's the text left in the textbox on Town Square channel or a post on the same channel that is causing the white screen with error. Other channels seem to be ok.
Screen Shot 2020-08-31 at 4 54 51 PM

jgilliam17:
One more thing I am seeing now (new issue, since last change)
When copy/pasting conversations from Mattermost, paperclip and emoji icons get pushed to the top left corner of the textbox.
Screen Shot 2020-08-31 at 5 11 29 PM

jespino:
@jgilliam17 I'm not sure if the codeClassName is related to my changes, but I'll take a look. The paper clip is something that I can remove for sure too.

jespino:
@jgilliam17 Actually bot where my faults, now should be fixed :)

jgilliam17:
Thank you @jespino 
The textbox icons are fixed and the error and white screen are gone.
- I am seeing something new with the code paste, extra slashes are inserted in one instance (see image).  I copied this snippet from the [test case](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T1734) I was looking at in the Jira test management and the slashes were added on paste
Screen Shot 2020-09-02 at 1 37 34 PM
- When selecting and copying text from the posted code block, it posts with the text merged all into one line
Screen Shot 2020-09-02 at 2 26 46 PM


- When copy/pasting convo from Mattermost using cmd+shift+v, date and new message separator lines are still visible 
Screen Shot 2020-09-02 at 10 02 59 AM

- When copy/pasting from google form, sorry I missed this earlier,  quite a few items are omitted/missing. Google sheets, docs and slides(each separately) are copying fine.
Screen Shot 2020-09-02 at 12 19 31 PM
Screen Shot 2020-09-02 at 9 30 13 AM
- Unable to paste .png or .jpg files after using Copy image (e.g. user profile picture)
Screen Shot 2020-09-02 at 2 44 01 PM

angeloskyratzakos:
Hey all, I will remove the tag `Setup Cloud Test Server` for maintenance reasons. Feel free to re-add it when you need it. Thank you

lieut-data:

PR #5844, opened 2020-07-03
[MM-26610] Fix the crash when switching teams

larkox:
#### Summary
Fix the crash when switching teams when the RHS is open.
More info in the Jira Ticket.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26610


hmhealey:
Looks good to me. I was a bit worried that it would reopen the sidebar when switching back, but that doesn't seem to be the case.

PR #5854, opened 2020-07-03
Sidebar V2 Phase 2

hmhealey:
These are the webapp changes for the new sidebar. These changes add new features such as custom categories and drag and drop, and they also tie the contents of the sidebar to the data provided by the server. 

Note that there are outstanding issues with these features, but we've decided that they're small enough that they can be addressed once this is on master. We want to get in the core of this on Monday so that it'll have a week to soak on community-daily before the actual submission deadline.

Since this is coming from a feature branch, a thorough review is less necessary since the individual PRs going into this have already been reviewed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23730

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1185
https://github.com/mattermost/mattermost-server/pull/14967


hmhealey:
FYI, this test server uses the master branch of the server, so the sidebar will be blank when you turn on the new sidebar. I'll try to figure out why this is happening...

e2e/cypress/integration/channel_sidebar/drag_and_drop_spec.js
```diff
@@ -0,0 +1,74 @@
...
+    testWithConfig({
+        ServiceSettings: {
+            ExperimentalChannelSidebarOrganization: 'default_on',
+        },
+    });
+
+    before(() => {
+        cy.apiLogin('user-1');
+
+        cy.visit('/');
```
saturninoabril, line 30: Non-blocking and we can fix post-merge into master. Instead of using hard-coded user data like `user-1`, in the future please something like:
```javascript
cy.apiInitSetup({loginAfter: true}).then(({team}) => {
});
```

hmhealey: I'll make a note of that. I forgot that we actually had any e2e tests for this since I filed separate tickets to add tests for this anyway.

PR #5858, opened 2020-07-06
MM-26685 enable redux setting to emit rudder events when url and key is set

sudheerDev:
#### Summary
Set redux variable on client4 when url or key is set for rudder
 
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26685

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1186


components/root/root.test.jsx
```diff
@@ -167,4 +191,11 @@ describe('components/Root', () => {
         wrapper.setProps(props2);
         expect(props.history.push).toHaveBeenLastCalledWith('/signup_user_complete');
     });
+
+    test('should call for enableRudderEvents on call of onConfigLoaded if url and key for rudder is set', () => {
```
lieut-data, line 195: Can we verify the inverse as well, such that we don't enable events when the url or rudder key is not et?

sudheerDev: Did

PR #5823, opened 2020-06-30
[MM-26465] Background and general layout for cloud onboarding

devinbinnie:
#### Summary
Basic background and layout for the cloud onboarding screen for both Admin and End-User.
Currently on by default, logic to turn on based on first-use to come in a future PR. Here for UX/CSS review.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26465

PR #5861, opened 2020-07-06
[MM-26480] Card/Accordion Common Component

devinbinnie:
#### Summary
Has changes from https://github.com/mattermost/mattermost-webapp/pull/5823.
Relevant changes here: https://github.com/mattermost/mattermost-webapp/pull/5861/files/05521d800f2c00314bbd4e0d1c2b573da28cae46..60ad53a4a22c3b8f78f92a3185530f0450b2cc00

This is the Card/Accordion Common Component to be used on the next steps screen. It has some basic animations and most of the styling used. Will add additional styling more specific to the Cloud Onboarding screen in future PRs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26480

#### Related Pull Requests
Waiting on this PR: https://github.com/mattermost/mattermost-webapp/pull/5823


stevemudie:
The accordion effect works as expected.

My only concern is how tight the text is to the buttons. Or will more specific styling be taken care of later as mentioned above?
Toggle

devinbinnie:
> The accordion effect works as expected.
> 
> My only concern is how tight the text is to the buttons. Or will more specific styling be taken care of later as mentioned above?
> Toggle

Sorry I should have been more specific, the button and text placement is just a proof of concept for the common component. The actual components we use for Cloud Onboarding will look nothing like this.

matthewbirtch:
Works well @devinbinnie. Just want to make sure that for a collapsed item, clicking the entire container will open it?

devinbinnie:
> Works well @devinbinnie. Just want to make sure that for a collapsed item, clicking the entire container will open it?

Correct, when I actually go to use this component to design the "Next Step" component we're going to actually use, the entire container can be the trigger. The idea behind the common component is to be as design-agnostic as possible.

components/next_steps_view/next_steps_view.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    <div className='NextStepsView__header-logo'>
+                        <img src={this.getLogo()}/>
+                    </div>
+                </div>
+                <div className='NextStepsView__body'>
+                    <div className='NextStepsView__body-main'>
+                        <Accordion defaultExpandedKey={'Card_1'}>
+                            {(setExpanded, expandedKey) => {
+                                return (
+                                    <>
```
jwilander, line 75: Woah, first I've seen this. Is this essentially a `<React.Fragment>`?

components/card/card_body.tsx
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    componentDidMount() {
+        if (this.card.current && this.props.expanded) {
+            this.card.current.style.height = `${this.card.current.scrollHeight}px`;
+        }
+    }
+
+    componentDidUpdate(prevProps: Props) {
+        if (this.card.current) {
```
jwilander, line 29: nit: Could reduce nesting by:

```javascript
if (!this.card.current) {
    return;
}

// do other stuff
```

components/card/card.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+        expect(wrapper.instance().card.current?.style.height).toBe('');
+
+        wrapper.setProps({expanded: true});
+        expect(wrapper.instance().card.current?.style.height).not.toBe('');
+
+        wrapper.setProps({expanded: false});
+        expect(wrapper.instance().card.current?.style.height).toBe('');
+    });
+});
```
jwilander, line 42: Snapshot tests are good but I'm not seeing any cypress tests. I'm ok if we're planning on adding those later but we should make sure we have cypress testing this new onboarding flow before we merge to master

components/next_steps_view/next_steps_view.tsx
devinbinnie: Yeah! It's a shorthand they introduced in v16.2.0: https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

components/card/card_body.tsx
devinbinnie: I wanted to have it nested in case there were other statements to add to `componentDidUpdate`. But since there aren't right now, I'll make this change.

components/card/card.test.tsx
devinbinnie: Yeah I have a ticket to add those tests near the end, I'm hesitant to do them now in case the flow needs to change or something else does. Ticket here: https://mattermost.atlassian.net/browse/MM-26477

deanwhillier:
Took a look at the generated markup before I get into the actual code. Some suggestions as follows:

### CSS Naming Conventions:
I'm proposing we use [this doc](https://mattermost.atlassian.net/wiki/spaces/APPS/pages/758284290/CSS%2BNaming%2BConventions%2Bbased%2Bon%2BBEM%2Bfor%2BReact) as the basis for our naming conventions. It's absolutely open for discussion, but otherwise for consistency should be followed for new work (and cleanup as we get to it).

Apologies, but I forgot to check class naming on the other PR.

### Accordion classes:

For this markup:

```html
<div class="Accordion">
   <div class="Card">
      <div class="Card__header expanded">
         <span>Card Header 1</span>
         <button>Toggle</button>
      </div>
      <div class="Card__body expanded" style="height: 20px;">
         <div>Card Body 1</div>
      </div>
   </div>
   <div class="Card">
      <div class="Card__header">
         <span>Card Header 2</span>
         <button>Toggle</button>
      </div>
      <div class="Card__body">
         <div>Card Body 2</div>
      </div>
   </div>
   <div class="Card">
      <div class="Card__header">
         <span>Card Header 3</span>
         <button>Toggle</button>
      </div>
      <div class="Card__body">
         <div>Card Body 3<br>Bigger Card Body</div>
      </div>
   </div>
</div>
```
I would suggest these changes (see below for details):

```html
<div class="Accordion">
   <div class="Card Accordion_panel Accordion_panel___expanded">
      <div class="Card_header Accordion_panelHeader">
         <span>Card Header 1</span>
         <button>Toggle</button>
      </div>
      <div class="Card_body Accordion_panelBody" style="height: 20px;">
         <div>Card Body 1</div>
      </div>
   </div>
   <div class="Card Accordion_panel">
      <div class="Card_header Accordion_panelHeader">
         <span>Card Header 2</span>
         <button>Toggle</button>
      </div>
      <div class="Card_body Accordion_panelBody">
         <div>Card Body 2</div>
      </div>
   </div>
   <div class="Card Accordion_panel">
      <div class="Card_header Accordion_panelHeader">
         <span>Card Header 3</span>
         <button>Toggle</button>
      </div>
      <div class="Card_body Accordion_panelBody">
         <div>Card Body 3<br>Bigger Card Body</div>
      </div>
   </div>
</div>
```

**Details:**

Expanding/Collapsing of the individual panels of an accordion are part of the accordion functionality, not the cards being used as accordion panels. I would propose:

* adding an additional class (`.Accordion_panel`) to the parent element of the cards (`.Card`) to indicate that the card is acting as an accordion panel. `.Accordion_card` could be an alternative name 0/5.
* I would even go so far as to add `.Accordion_panelHeader` & `.Accordion_panelBody` to the card elements to allow you to apply styling to them as needed from the accordions perspective to keep styles between normal cards and cards used as accordion panels separate.
* moving the `.expanded` class from the individual elements of the `.Card` up to the (`.Accordion_panel`) element and changing the classname to `.Accordion_panel___expanded` to indicate the accordion panel is currently expanded and using that single class to apply styles to the individual accordion panel elements as needed would be a bit cleaner. Conceptually, it is the panel as a whole that expands/collapses even if the child elements' styles change as a result of that visual state change.
* Apply accordion styles on card elements as needed using the accordion class names instead of the card class names.

Looking at code directly next.

components/card/card.scss
```diff
@@ -0,0 +1,33 @@
...
+    & + .Card {
+        margin-top: 16px;
+    }
+}
+
+.Card__header {
+    margin: 12px 12px 0 12px;
+    padding-bottom: 12px;
+    
+    &.expanded {
```
deanwhillier, line 16: See comment above about making the `.expanded` state part of the accordion rather than the card.

components/card/card.scss
```diff
@@ -0,0 +1,33 @@
+.Card {
+    border: 1px solid var(--center-channel-color-08);
```
deanwhillier, line 2: Should we be using the new `rgba(var(--center-channel-color-rgb), 0.8)` approach here? (might have the exact syntax wrong as I'm going on memory atm)

devinbinnie: Ahh I keep using this wrong! Will fix.

jwilander:
@matthewbirtch can we handle those in a follow-up as an improvement ticket? We should avoid changing designs on PRs that are already implemented and been dev/QA/UX reviewed

PR #5865, opened 2020-07-07
[MM-26468][MM-26483] Progress Bar Common Component/Sidebar Next Steps Component

devinbinnie:
#### Summary
This is the working prototype for the Next Steps widget in the LHS.
Also included is the new progress bar common component for the webapp.
Currently there is no functionality behind this, except that the 'X' button will increment the progress so you can see how the animation is expected to work. The functionality will come in a future PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26468
https://mattermost.atlassian.net/browse/MM-26483


devinbinnie:
> Thanks @devinbinnie. It's looking great. A few things:
> 
> 1. In the sidebar, this menu item shouldn't be in the active state - maybe you already had that in mind, but just making sure. The background and highlight keyline on the left should follow the same pattern as channels in the sidebar
> 2. Can we make sure the x icon is the same size and x position as the `+` icon in the rest of the sidebar? The hover state should match for the icon as well. I realized my designs did not do this, so I've fixed that now in Figma. Apologies.
> 
> Here are a couple screenshots so you can see what I mean.
> 
> image
> 
> image
> 
> I've updated the Figma file, but let me know if you need any more specifics

Done, I've added the class for the active state of the sidebar, but won't be able to turn it on until it's merged with the other stuff. We can validate then.

devinbinnie:
> Thanks @devinbinnie. Looks like the active border is still there even when it's inactive
> 
> image
> 
> I'm also noticing that the active border in my design is wrong compared to the other sidebar items.
> 
> Also, It looks like active channels use the `sidebar-text` variable as the border color, when it should be using `--sidebar-text-active-border`. That's strange to me - is it possible to make that change across the board?
> 
> Either way, we should be using a consistent style for that active border between the channel items and the getting started sidebar item. It should be 4px wide rectangle with a 4px rounded corner radius, offset to the left by -2px so that the left rounded edge is cut off.

I've fixed the border to only show up on active state AND it's using the right shape and colour.

As for the channels, we can definitely make that change if that's the intention. We'll just want to do it as part of a separate ticket that we can merge into master irrespective of this.

PR #2301, opened 2019-01-30
MM-13862 - Updating emoji picker position

asaadmahmood:
#### Summary
MM-13862 - Updating emoji picker position

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-13862

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] Ran `make check-style` to check for style errors (required for all pull requests)
- [ ] Ran `make test` to ensure unit and component tests passed
- [ ] Added or updated unit tests (required for all new features)
- [ ] Needs to be implemented in mobile (link to PR or User Story)
- [ ] Has server changes (please link)
- [ ] Has redux changes (please link)
- [x] Has UI changes
- [ ] Includes text changes and localization file ([.../i18n/en.json](https://github.com/mattermost/mattermost-webapp/blob/master/i18n/en.json)) updates
- [ ] Touches critical sections of the codebase (auth, posting, etc.)

PR #5869, opened 2020-07-08
[MM-26673] Attach emoji picker to the `Post Edit Modal`

jfrerich:
#### Summary
This PR attaches the emoji picker to the footer of the Edit Post Modal and allows it to track with the modal as the window is resized.  The issue arose with the introduction of this PR (https://github.com/mattermost/mattermost-webapp/pull/2301).  

To fix the issue, the parent component of the `ref` used required the addition of `position: relative`
I added the `rightOffset` prop to the `EmojiPickerOverlay` which allows the parent component to specify this value. The introduction of the prop allows for future components to utilize the value and in the case of this PR determine avoid the handling logic inside the `emojiPickerPosition` function.

I placed the picker under the modal and to the right.  This could be changed if others feel a better location is appropriate.  The difficult part was just figuring out how to attach the picker to the modal.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-26673

#### Screenshots
Before:
image

After:
image


components/edit_post_modal/edit_post_modal.jsx
```diff
@@ -473,7 +476,9 @@ class EditPostModal extends React.PureComponent {
                         </div>
                     </div>
                 </Modal.Body>
-                <Modal.Footer>
+                <Modal.Footer
+                    style={{position: 'relative'}}
```
hmhealey, line 480: I imagine this'll probably move if we move the emoji picker, but wherever it ends up, can you modify the CSS to set the position instead of styling it inline? The CSS for this one should be in `_modals.scss`

components/edit_post_modal/edit_post_modal.jsx
```diff
@@ -491,10 +496,15 @@ class EditPostModal extends React.PureComponent {
                         disabled={this.isSaveDisabled()}
                         onClick={this.handleEdit}
                     >
-                        <FormattedMessage
-                            id='edit_post.save'
-                            defaultMessage='Save'
-                        />
+                        <span
+                            id='editModalFooter'
```
hmhealey, line 500: Could you remove this ID since we're not using it anywhere else? We're trying to cut down on unused IDs since people sometimes use them for CSS styling when they shouldn't.

jfrerich:
@hmhealey, I updated the screenshot image with the latest.  Turns out the picker attached to the `editPostEmoji` ref doesn't require the `position: relative` styling and the only thing needed to resolve this was to modify the `emojiPickerPosition` function.

jfrerich:
@hmhealey, @asaadmahmood ping for review.  We are trying to merge today for release

jfrerich:
@asaadmahmood, please see below for the update which moves the modal 10px left

image

jfrerich:
@DHaussermann, I've added you for QA on this PR. Could you provide a review?  This is for today's release

asaadmahmood:
Looks good for now.

DHaussermann:
Tested on desktop and browse 
Tested on webapp and mobile browser view

@jfrerich I see a couple regressions on this PR. Both visible in the mobile browser view...
1. Top of emoji category icons are now cut off
2. The `X` to close out of the emoji picker and return to regular edit is only faintly visible when you mouse over it. (Looks like emoji picker use to have it's own header above the search bar where the `X` would sit.)
Mobile-Emoji

hmhealey:
@DHaussermann I can file another ticket for those since they don't appear to be a regression. This is what I see on master
Screen Shot 2020-07-28 at 3 26 21 PM
Screen Shot 2020-07-28 at 3 25 00 PM

DHaussermann:
After testing previous versions again, it's not entirely clear that these where introduced by the PR as the emoji picker is quite broken in mobile view before this PR. 
I have created https://mattermost.atlassian.net/browse/MM-27370 to address these other 2 issues.

On that note @hmhealey do you have any thoughts on how difficult or risky it may be to merge this into 5.26? I ask because it seems that without the emoji picker is pretty unusable on mobile web view.

hmhealey:
> On that note @hmhealey do you have any thoughts on how difficult or risky it may be to merge this into 5.26? I ask because it seems that without the emoji picker is pretty unusable on mobile web view.

I have no idea. I don't trust most of the CSS that goes into the mobile web view, so I'd say it might be risky. I can check with Dean to see if he has a better idea since I think he's the one who made the emoji picker usable on mobile web in the first place.

amyblais:
@jfrerich Kind note to cherry-pick this,

jfrerich:
Thanks, @amyblais.  I just cherry-picked this commit

PR #5875, opened 2020-07-09
[MM-26594] - Don't delete channel category when there are no channels…

nevyangelova:
#### Summary
Removes the condition that removed the category when there are no channels 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26594


jgilliam17:
@nevyangelova one small issue - I see empty Favorites category at login.

devinbinnie:
Are we sure this was expected? I was specifically asked to hide the Channels category when no channels were present there. Has something changed?

nevyangelova:
> Are we sure this was expected? I was specifically asked to hide the Channels category when no channels were present there. Has something changed?

@jgilliam17 @esethna

hmhealey:
I think we might've had a misunderstanding here. I remember wanting to hide Favorites when it was empty, but I think we wanted to keep Channels around.

nevyangelova:
@jgilliam17 updated!

esethna:
@nevyangelova can you please help explain what is expected here with this change and how it affects the category types (faves, custom, channels, dms) when they are empty

jgilliam17:
@nevyangelova 
Channels category is not removed from the LHS when empty, as expected.

Some new issues were introduced
- All new categories start of as hidden, unless created from Move to>Move to new category
- Custom category disappears from the LHS when you move all the channel out of it

nevyangelova:
@jgilliam17 sorry! Paste error! Should work good now :)

nevyangelova:
> @nevyangelova can you please help explain what is expected here with this change and how it affects the category types (faves, custom, channels, dms) when they are empty

@esethna per ticket description the only change that needs to happen here is that the 'CHANNELS' category doesn't disappear when all the channels are dragged out or deleted from it. It shouldn't affect the other categories in any way.

esethna:
Great, thanks (I saw the code before your latest change and something looked off), thanks for the fix

jgilliam17:
@nevyangelova do you know if these received preferences with delay errors are related to the fix?
Screen Shot 2020-07-10 at 11 54 48 AM

nevyangelova:
> @nevyangelova do you know if these received preferences with delay errors are related to the fix?
> Screen Shot 2020-07-10 at 11 54 48 AM

No absolutely not, this doesn't touch the actions :)

components/sidebar/sidebar_category/sidebar_category.tsx
```diff
@@ -219,7 +219,7 @@ export default class SidebarCategory extends React.PureComponent<Props, State> {
             return null;
         }
 
-        if (category.type !== CategoryTypes.DIRECT_MESSAGES && category.type !== CategoryTypes.CUSTOM && !channels?.length) {
+        if (category.type !== CategoryTypes.DIRECT_MESSAGES && category.type !== CategoryTypes.CUSTOM && category.type !== CategoryTypes.CHANNELS && !channels?.length) {
```
hmhealey, line 222: Perhaps we could just change this to `category.type === CategoryTypes.FAVORITES && !channels?.length for simplicity? Otherwise, looks good.

nevyangelova: oops yes!

PR #5882, opened 2020-07-09
[MM-25264] [MM-25265] Create and apply filter component to user grids

fmunshi:
#### Summary
- Adds the newly created Filter component to the admin console 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-25264
- https://mattermost.atlassian.net/browse/MM-25265

#### Related Pull Requests
- https://github.com/mattermost/mattermost-redux/pull/1190

#### Screenshots
- Screen Shot 2020-07-09 at 1 59 55 PM


components/widgets/admin_console/admin_panel.tsx
```diff
@@ -23,7 +23,7 @@ type Props = {
 
 const AdminPanel: React.FC<Props> = (props: Props) => (
     <div
-        className={'AdminPanel ' + props.className}
+        className={'AdminPanel clearfix ' + props.className}
         id={props.id}
     >
         <div
```
fmunshi, line 29: This change was required because the current implementation of `AdminPanel` was applying a version of clearfix that caused overflow content to be hidden... with the filter component we dont want that since otherwise the filter could overflow out of the admin panel and not be visible

components/widgets/admin_console/admin_panel.scss
```diff
@@ -6,7 +6,7 @@
 .AdminPanel {
     @include border-radius($border-rad);
     @include box-shadow(0 1px 2px 0 alpha-color($black, .5));
-    @include clearfix;
+
```
fmunshi, line 9: This is replaced with the `clearfix` class from bootstrap being added to the element itself

components/textbox/index.ts
```diff
@@ -33,8 +33,8 @@ const makeMapStateToProps = () => {
         return {
             currentUserId: getCurrentUserId(state),
             currentTeamId: teamId,
-            profilesInChannel: getProfilesInChannel(state, ownProps.channelId, true),
-            profilesNotInChannel: getProfilesNotInChannel(state, ownProps.channelId, true),
+            profilesInChannel: getProfilesInChannel(state, ownProps.channelId, {active: true}),
+            profilesNotInChannel: getProfilesNotInChannel(state, ownProps.channelId, {active: true}),
```
fmunshi, line 37: I modified the signature of these two functions on redux to make the last param a `Filters` object instead of an `active` boolean

components/admin_console/team_channel_settings/channel/details/channel_members/index.ts
```diff
@@ -69,16 +79,18 @@ function makeMapStateToProps() {
             guest_count: 0,
         };
 
-        const searchTerm = state.views.search.systemUsersSearch?.term || '';
+        const searchTerm = state.views.search.userGridSearch?.term || '';
+        const filters = state.views.search.userGridSearch?.filters || {};
         let users = [];
         if (searchTerm) {
-            users = searchProfilesInChannel(state, channelId, searchTerm, false, true);
+            users = doSearchProfilesInChannel(state, channelId, searchTerm, false, {...filters, active: true});
```
fmunshi, line 86: This function was not being memoized correctly - it is fixed now

components/admin_console/filter/filter_list.tsx
```diff
@@ -0,0 +1,59 @@
...
+    optionKey: string;
+    updateValues: (values: FilterValues, optionKey: string) => void;
+}
+
+class FilterList extends React.PureComponent<Props> {
+    updateOption = async (value: boolean, key: string) => {
+        const values = {...this.props.option.values};
+        values[key].value = value;
+        await this.props.updateValues(values, this.props.optionKey);
+        this.forceUpdate(); // hack figure out a better solution?
```
fmunshi, line 21: todo: Figure out if this is still needed

fmunshi:
@michaelgamble you can login with the system admin credentials listed above and view the filter component on the team management page (https://mattermost-webapp-pr-5882.test.mattermost.cloud/admin_console/user_management/teams/kxhoegfh4jbuxpozodfrojfrfy), unfortunately you wont be able to test functionality (only the visuals of the component itself since the server branch isnt being picked up for some reason)

amyblais:
@hahmadia @mkraft Kind note for reviewers that feature complete for v5.26 is this week Friday July 17th. Let me know any concerns,

components/admin_console/filter/filter.tsx
```diff
@@ -0,0 +1,247 @@
...
+        keys.forEach((key) => {
+            const option = options[key];
+            if (option) {
+                option.keys.forEach((optionKey) => {
+                    if (!option.values[optionKey]) {
+                        error = `Invalid Filter option key: ${optionKey}, no matching value for given option key`;
+                    }
+                });
+            } else {
+                error = `Invalid Filter key: ${key}, no matching option for given key`;
```
mkraft, line 75: Are these errors OK not to be translated?

fmunshi: Actually @mkraft I initially added these as error handling for the component itself, cause the list of keys need to match the options map given (or else the page fails to render) but theres no way to enforce that, so i opted to set the options to empty and log this error

but then removed the logging of the error later (due to linting)... I think ill just remove these messages and have the component fail silently without breaking the page

components/admin_console/data_grid/data_grid.scss
```diff
@@ -138,4 +138,4 @@
             text-align: right;
         }
     }
-}
+}
```
hahmadia, line 141:
```suggestion
}

```

components/admin_console/filter/filter.tsx
```diff
@@ -0,0 +1,247 @@
...
+        const filters = this.renderFilterOptions();
+        const {filterCount} = this.state;
+
+        return (
+            <div
+                className='Filter'
+                ref={this.filterRef}
+            >
+                <button
+                    className={this.state.show ? 'Filter_button Filter__active' : 'Filter_button'}
```
nevyangelova, line 190: 3/5 prefer the use of `classnames`

components/admin_console/team_channel_settings/team/details/team_members/team_members.tsx
```diff
@@ -125,7 +143,36 @@ export default class TeamMembers extends React.PureComponent<Props, State> {
     }
 
     private search = async (term: string) => {
-        this.props.actions.setSystemUsersSearch(term);
+        this.props.actions.setUserGridSearch(term);
+    }
+
+    private onFilter = async (filterOptions: FilterOptions) => {
```
nevyangelova, line 149: Maybe we can extract this as a util function since its used in multiple components, and it seems to be the same logic?

sass/admin_console_base/_module.scss
```diff
@@ -0,0 +1,2 @@
+// Only for combining all the files in this folder
+@import 'sys_css_variables';
```
nevyangelova, line 2: nit: new line

components/admin_console/filter/filter.tsx
```diff
@@ -0,0 +1,247 @@
...
+                        error = `Invalid Filter option key: ${optionKey}, no matching value for given option key`;
+                    }
+                });
+            } else {
+                error = `Invalid Filter key: ${key}, no matching option for given key`;
+            }
+        });
+
+        if (error) {
+            options = {};
```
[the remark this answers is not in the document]
fmunshi: The reason i've done it this way is that the `options` passed in from props are used as kind of a default version of the options to revert back to when the user clicks on reset filters. The options in the state are modified by the user clicking on checkboxes in the component and then when the user clicks on reset `state.options` is set to `{...props.options}` once more 

although you are right the spread is in the wrong spot... it should be at line 63 and then there should be no spread at line 86

components/admin_console/filter/filter.tsx
```diff
@@ -0,0 +1,247 @@
...
+        this.hidePopover();
+    }
+
+    hidePopover = () => {
+        this.setState({show: false});
+        this.buttonRef?.current?.blur();
+    }
+
+    togglePopover = () => {
+        if (this.state.show) {
```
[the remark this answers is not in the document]
fmunshi: when the state is `show` = `true` that means it is currently visible so the `toggle` function needs to toggle its status back to not being visible i.e. set show to false, I dont see the confusion here

components/admin_console/filter/filter.scss
```diff
@@ -0,0 +1,122 @@
...
+        height: 32px;
+
+        background: var(--sys-center-channel-bg);
+        border-radius: 4px;
+        border: 0px;
+        color: var(--sys-button-bg);
+        font-size: 12px;
+        padding: 0 10px;
+
+        &.Filter__active, &:hover, &:focus {
```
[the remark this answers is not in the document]
fmunshi: I was just following @deanwhillier comments here and he mentioned `__active` for modifiers... https://github.com/mattermost/mattermost-webapp/pull/5707#issuecomment-646697676

Not sure what the _right_ way is

components/admin_console/filter/filter.tsx
fmunshi: Didnt know about this, is it preferable to use classnames even when not dealing with conditional statements?

actions/views/search.js
```diff
@@ -13,6 +13,26 @@ export function setModalSearchTerm(term) {
     };
 }
 
+export function setUserGridSearch(term) {
+    return async (dispatch) => {
+        dispatch({
+            type: SearchTypes.SET_USER_GRID_SEARCH,
+            data: term,
+        });
+        return {data: true};
```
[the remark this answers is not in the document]
fmunshi: Neither am I, it just seems to be a convention that all of our redux actions follow i.e. to always return _something_ for `data` so when no data is returned we just return `true` so theres a consistent way to tell if the action was successful im assuming? Might make sense for @hmhealey to shed some light on the original reasons for doing it this way

components/admin_console/team_channel_settings/team/details/team_members/team_members.tsx
fmunshi: it has slightly different logic one of them grabs roles prefixed with `team`  and one of them grabs `channel` roles technically I could extract a bit of the common logic and add a parameter for 'channel' or 'team' but i didnt think it was worth it, a little bit of duplication here is fine since we may need to modify it for each block differently in the future.

components/admin_console/filter/filter.scss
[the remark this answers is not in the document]
nevyangelova: Sorry about that its just that we are figuring out in the team which BEM notation to use :) feel free to keep that.

components/admin_console/filter/filter.tsx
nevyangelova: It's not a big deal but its good for consistency so then its easier to introduce 👍

fmunshi:
@srkgupta This is to be reviewed with https://github.com/mattermost/mattermost-server/pull/14998 & https://github.com/mattermost/mattermost-redux/pull/1190

fmunshi:
The last two issues are actually stories that are to be implemented. They will be addressed later.

https://mattermost.atlassian.net/browse/MM-26969
https://mattermost.atlassian.net/browse/MM-26970

fmunshi:
@srkgupta the issue related to guest access is now resolved

PR #5883, opened 2020-07-09
[MM-24708] Added What's New modal for Channel Sidebar Phase 2 changes

devinbinnie:
#### Summary
Copy of https://github.com/mattermost/mattermost-webapp/pull/5711 that was closed due to the feature branch being deleted.
See there for details.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24708


esethna:
Awaiting the updated GIF (blocked on working out bugs with sidebar before we capture it)

components/generic_modal.tsx
```diff
@@ -0,0 +1,151 @@
...
+        if (this.props.handleConfirm) {
+            let confirmButtonText = (
+                <FormattedMessage
+                    id='generic_modal.confirm'
+                    defaultMessage='Confirm'
+                />
+            );
+            if (this.props.confirmButtonText) {
+                confirmButtonText = (
+                    <React.Fragment>
```
hmhealey, line 71: Continuing [this discussion](https://github.com/mattermost/mattermost-webapp/pull/5711#discussion_r443620657), I did a little googling and it looks like `React.ReactNode` is the preferred type for `confirmButtonText`. I guess it's automatically typed `JSX.Element` because of how it's set above, but `React.ReactNode` supports both cases.

hmhealey: That said, I'd be fine leaving it as-is now that I understand the reasoning for this. I still wanted to call it out though since it's good to know the distinction between the two since we'll need to use `React.ReactNode` to type props like `children` elsewhere.

devinbinnie: I feel like I should know that, considering I just used that for cloud onboarding :P
I'll make the change.

components/generic_modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+    confirmButtonClassName?: string;
+    cancelButtonText?: React.ReactNode;
+    isConfirmDisabled?: boolean;
+};
+
+type State = {
+    show: boolean;
+}
+
+export default class GenericModal extends React.PureComponent<Props, State> {
```
nevyangelova, line 27: What is the benefit of a class vs. functional component here?

components/edit_category_modal/edit_category_modal.tsx
```diff
@@ -107,70 +95,32 @@ export default class EditCategoryModal extends React.PureComponent<Props, State>
...
-                restoreFocus={true}
-                role='dialog'
-                aria-labelledby='editCategoryModalLabel'
+            <GenericModal
+                onHide={this.props.onHide}
+                modalHeaderText={modalHeaderText}
+                handleConfirm={this.handleConfirm}
+                handleCancel={this.handleCancel}
+                confirmButtonText={editButtonText}
+                isConfirmDisabled={!this.state.categoryName || (Boolean(this.props.initialCategoryName) && this.props.initialCategoryName === this.state.categoryName)}
```
nevyangelova, line 104: we can extract this to a variable for clarity

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+type Props = {
+    currentUserId: string;
+    actions: {
+        savePreferences: (userId: string, preferences: PreferenceType[]) => void;
+        closeModal: (modalId: string) => void;
+    };
+}
+
+export default class SidebarWhatsNewModal extends React.PureComponent<Props> {
```
nevyangelova, line 23: Same here, can this be converted to a FC?

components/generic_modal.tsx
devinbinnie: Maybe I'm not understanding functional components properly. but this component does have some internal logic and state that need to be defined, so I didn't think that was possible with an FC.

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.tsx
devinbinnie: I don't think so, since @hmhealey had me convert this to be used by the `ModalController`, meaning it needs to be able to handle its own functionality. Again, maybe i'm not understanding FC's properly.

components/generic_modal.tsx
nevyangelova: My idea was to try move towards FC with the use of React hooks. You can achieve all that using this unless you are exposing the instance methods via refs (we should also try avoiding that btw :P). But I won't block this PR.

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.tsx
nevyangelova: Same comment applies.

components/generic_modal.tsx
devinbinnie: Do you feel like it's worth reworking this component to do that? I'm not sure about the effort since I haven't worked with hooks yet, but I guess I need to try it out at some point.

nevyangelova: No you are right, I won't block this PR but we can discuss in the Hangouts how we wan't to proceed with hooks and FC.

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.tsx
hmhealey: I've never tried using functional components with the `ModalController`, but I don't see any reason it shouldn't work assuming the types are correct for the related actions

hmhealey:
@andrewbrown00 Here's the modal with the preview gif. How's it look?
image

andrewbrown00:
Thanks @hmhealy, can you make the GIF wider to align to the right edge of the button; so the padding is equal on the left and right side of the GIF.

It’s fine if the modal get a bit taller.

hmhealey:
I removed the fixed height from the image and made sure it scaled down properly on smaller screens.

Screen Shot 2020-07-18 at 09 28 54
Screen Shot 2020-07-18 at 09 28 44
Screen Shot 2020-07-18 at 9 29 37 AM

hmhealey:
Andrew pointed out that this opens up immediately when the user enables the sidebar from Account Settings. I'm going to see if there's a way to delay it until either the Account Settings closes or the user next refreshes. Ideally, we'd show it after the modal closes, but if there's no easy way to do that, then next refresh might be okay.

prapti:
In testing queue.

hmhealey:
Having the What's New modal appear over the Account Settings modal is a bit weird, but we've decided to leave it as-is for 5.26. We've filed a follow up to possibly change it in the future: https://mattermost.atlassian.net/browse/MM-27163

hmhealey:
Manually cherry-picked onto release-5.26

PR #5888, opened 2020-07-10
MM-26075 Standardize Cypress custom commands for "api/system"

saturninoabril:
#### Summary
Standardize returned data and add type definitions for intellisense custom API system commands.

Test report - https://mm-cypress-report.s3.amazonaws.com/582-MM-26075_cypress-api-system-commands-master/mochawesome.html

#### Ticket Link
 JIRA ticket: https://mattermost.atlassian.net/browse/MM-26075


e2e/cypress/support/api/system.js
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        expect(hasLicenseKey, `No license for feature: ${key}`).to.equal(true);
+
+        return cy.wrap({license});
+    });
+});
+
+Cypress.Commands.add('requireLicense', () => {
```
saturninoabril, line 38: Need general agreement here. Note that `requireLicense` and `requireLicenseForFeature` didn't start with `api` prefix. Our guideline is to prefix with `api`. However, I feel like it's easier to understand if it starts with "require". Open to suggestion on naming and/or file location if better to move into separate file.

e2e/package.json
```diff
@@ -18,7 +18,7 @@
     "lodash.mapkeys": "4.6.0",
     "lodash.without": "4.4.0",
     "lodash.xor": "4.5.0",
-    "mattermost-redux": "5.23.0",
+    "mattermost-redux": "github:mattermost/mattermost-redux#8b003ea1aa10bf64b306a006e8a0ed7477924704",
```
saturninoabril, line 21: Good to use fix version but [some of the added types](https://github.com/mattermost/mattermost-webapp/pull/5888/files#diff-f6b9573b81480c95519c79b64ff0e029R7-R11) are not yet available to `5.23.0` nor on the latest npm release version of `5.24.2`. The commit hash here is on master, latest at the moment.

e2e/cypress/support/api/system.d.ts
```diff
@@ -0,0 +1,99 @@
...
+// ***************************************************************
+// Each command should be properly documented using JSDoc.
+// See https://jsdoc.app/index.html for reference.
+// Basic requirements for documentation are the following:
+// - Meaningful description
+// - Specific link to https://api.mattermost.com
+// - Each parameter with `@params`
+// - Return value with `@returns`
+// - Example usage with `@example`
+// Custom command should follow naming convention of having `api` prefix, e.g. `apiLogin`.
```
srkgupta, line 15: Some of the functions don't follow this pattern. For example: requireLicense(). Can we remove this?

e2e/cypress/support/api/system.js
srkgupta: I think its fine. We just need to remove the guideline only for this file.

e2e/package.json
srkgupta: Redux hash keeps changing as new commits are added. It will be tough to maintain this. And it's already a nightmare in webapp repo to maintain this. Let's not switch to the commit hash mode. Can we use "master"

saturninoabril:
@srkgupta Thanks. I ran the report earlier when I submitted this PR. Looks like no failing test were introduced by this PR. I updated the description with link to it.

e2e/package.json
saturninoabril: No, we can't use master. Let's keep an eye and switch once we have the 5.26 version of mattermost-redux.

e2e/cypress/support/api/system.js
josephbaylon: I actually prefer to prefix it so that it's consistent across the board. It will be easier to remember and order in code alphabetically if it follows convention. But I understand that would need other files to be updated. Maybe create a ticket to update it later.

e2e/package.json
josephbaylon: I believe you can use tags instead. Cleaner and easier to see which version until release version is available
https://github.com/mattermost/mattermost-redux/tags

saturninoabril: The type definition I need is on master or 5.26. Tags were currently up to 5.25 only.

josephbaylon: Got it. Might help monitor to create a ticket.

e2e/cypress/support/api/system.js
saturninoabril: Thanks for the inputs. Since the opinions are split, I have the luxury to choose (😄) - decided to add api prefix. Feel free to give feedback or change in the future.

PR #5889, opened 2020-07-10
[MM-26691] Store correct password settings values in local state

mgdelacroix:
#### Summary
The `PasswordSettings` component was incorrectly storing events in the local state. This PR modifies the `handleCheckboxChange` to store the right data in the state and trigger renders of the component when a checkbox changes (and hence, updating the error message).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26691


components/admin_console/password_settings.jsx
```diff
@@ -136,9 +136,11 @@ export default class PasswordSettings extends AdminSettings {
         this.handleChange(id, value);
     }
 
-    handleCheckboxChange = (id, value) => {
-        this.sampleErrorMsg = this.getSampleErrorMsg(this.state.passwordMinimumLength);
-        this.handleChange(id, value);
+    handleCheckboxChange = (id) => {
+        return ({target: {checked}}) => {
+            this.sampleErrorMsg = this.getSampleErrorMsg(this.state.passwordMinimumLength);
```
hmhealey, line 141: Looks like this is how it always worked before, but we're not supposed to have this component render off of something that isn't part of the state/props. Instead of storing the message as a field on `PasswordSettings`, can we instead call `this.getSampleErrorMsg` where this would be used in the `render` method?

mgdelacroix: I'm not sure I'm understanding your proposed change. You mean not having the `sampleErrorMessage` stored as part of the component and simply call `getSampleErrorMsg` during `renderSettings`? So the `handleCheckboxChange` and `handlePasswordLengthChange` would only update the state, right?

hmhealey: Yeah, that's correct. The way that this should ideally work is that the `onChange` props trigger a state change, and then the render method reads thee new state to update the displayed message.

mgdelacroix: Understood, thanks! I've updated the PR, PTAL :)

PR #5893, opened 2020-07-10
[MM-26470] Base Next Step Wizard Controller and Styling

devinbinnie:
#### Summary
This includes the logic for switching between steps and moving to the last step.
It also includes all the generic styling for the cards and buttons that are re-used.
We can save preferences by clicking 'Save Profile' (which is permanent for that user) and we can skip steps by clicking 'Skip for now'.
You can leave the Next Steps view by changing channels. It will persist on refresh if the steps haven't been completed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26470


devinbinnie:
@matthewbirtch Should be good to review everything now, the mobile view spacing was definitely off so I've fixed that.

devinbinnie:
> Hey @devinbinnie. The transition on the number looks a lot better now. Thanks!
> 
> Another pass at a review here:
> 
> **Animation for complete state**
> Can we make some changes to the transition from open to complete? It feels a bit abrupt. Something more like the animation here (click 'Save profile' in this prototype): https://www.figma.com/proto/uv9A6dZsekQHHhOhYC7yc2/Cloud-UX?node-id=2610%3A11754&viewport=541%2C490%2C0.2703074812889099&scaling=min-zoom
> 

It's hard to replicate the exact transition shown in Figma, since there doesn't seem to be any code shown behind it...
Nonetheless, I've fixed up the transition to be smoother overall and I think it looks good. Let me know your thoughts and we can touch it up as necessary.

> **One other thing to note for the behaviour:**
> Users can complete the steps in any order they wish, so if I jump right to step 3 and click 'Skip for now', I think it should jump back to Step 1 and open it instead of skipping through everything. I'm thinking 'Skip Getting Started' is the only button that skips through everything. Thoughts?

I think that would be a better UX flow in the happy path, but it could be kind of confusing. Is the idea to allow 'Skip for now' to go to the final screen if all subsequent steps have been completed? Or should 'Skip for now' never allow you to go to the final screen?

> 
> **Mobile Webview:**
> 
> * With the completed state, the left edge of the green checkmark should align with the left edge of the blue number icon and with the left edge of the body copy above
> * Space between cards should be 12px on mobile (as opposed to 16px right now)
> 
> image

This should be fixed now.

devinbinnie:
> > Is the idea to allow 'Skip for now' to go to the final screen if all subsequent steps have been completed? Or should 'Skip for now' never allow you to go to the final screen?
> 
> The idea was that 'Skip for now' is like ignoring the current step for the moment - with the ability to complete it later. So, the original intent was that it would never advance to the next screen.....only 'Skip Getting Started' would do that. Having said that, I can see how it's confusing. Let me give this some thought and I'll reply again on this PR

Okay I'll leave as-is for now, just let me know.

> 
> A few other things I noticed:
> 
> 1. For the last step completion....in this PR it doesn't animate to the complete state - it just abruptly goes to the next screen. It will need to collapse into the complete state before it advances to the next screen.

I haven't added the code for the final screen so that's why it's like that. It will be fixed in a future PR.

> 2. The shadow on the expanded card sometimes looks to be cut off-  in this case it's when step 3 is complete and step 2 is expanded.
> 
> Screen Shot 2020-07-16 at 9 28 07 AM
> 
> 1. The complete state shouldn't have a white background (this may be the cause of cutting off the shadow as well:
>    image
> 2. As discussed, I've updated the style slightly for the complete state as well. [Figma file](https://www.figma.com/file/uv9A6dZsekQHHhOhYC7yc2/Cloud-UX?node-id=34%3A4280) has been updated - here's a quick screenshot of the comparison so you can see the change. Top is the new style (consistent with the sizing and positioning of the collapsed state) and the bottom is the old style
>    image

The rest of the above is fixed.

matthewbirtch:
@devinbinnie we're thinking we can remove the 'Skip for now' from each of the individual steps. Users can click in to the other steps if they want. This should avoid all confusion. Can we just hide the button though for now in case there is push back from others and they want to bring it back?

matthewbirtch:
Also, @devinbinnie looks like some spacing updated need to be done on the complete state still:
image

devinbinnie:
@matthewbirtch Ready for review

matthewbirtch:
@devinbinnie two last things:
1. mobile webview, we need to adjust the indent now that we updated the complete style
image
2. Can we animate the transition on the shadow as well?

devinbinnie:
> @devinbinnie two last things:
> 
> 1. mobile webview, we need to adjust the indent now that we updated the complete style
>    image
> 2. Can we animate the transition on the shadow as well?

Done!

devinbinnie:
@jwilander @deanwhillier Please review ASAP, so I can avoid more messy merges :P

jwilander:
Looking shortly

components/next_steps_view/next_steps_view.tsx
```diff
@@ -45,12 +67,106 @@ export default class NextStepsView extends React.PureComponent<Props> {
...
-        return (
-            <section
-                id='app-content'
-                className='app__content NextStepsView'
+    getStartingStep = () => {
+        for (let i = 0; i < Steps.length; i++) {
+            if (!this.props.preferences.some((pref) => pref.name === Steps[i].id && pref.value)) {
+                return Steps[i].id;
+            }
+        }
```
jwilander, line 75: nit: this is a little hard to parse so it might be worth putting `preferences.some((pref) => pref.name === Steps[i].id && pref.value)` into it's own one-line function that has a clear name

components/next_steps_view/steps/complete_profile_step.tsx
```diff
@@ -0,0 +1,51 @@
...
+                <div className='NextStepsView__wizardButtons'>
+                    {/* <button
+                        className='NextStepsView__button cancel'
+                        onClick={this.onSkip}
+                    >
+                        <FormattedMessage
+                            id='next_steps_view.skipForNow'
+                            defaultMessage='Skip for now'
+                        />
+                    </button> */}
```
jwilander, line 37: All to be replaced in a follow-up PR I assume?

devinbinnie: Actually @matthewbirtch has asked that I remove it entirely for now until they work out how it affects the UX. I left it commented in case we want to bring it back.

jwilander: What about the `{'AAAAAAAA'}` :P ?

devinbinnie: Oh that's there just for me :P
That will definitely be replaced.

deanwhillier:
Some initial comments/questions in no particular order:

* On view load, first accordion animates open, is that intentional?
* Should there be a hover state on the accordion headers?
* There is a 12px gap between the edge of the header button and the visual bounds of the header. The interactive area should extend right to the edges of the visual area.
* The primary button is missing interaction states (hover, active etc)
* Are we allowing an accordion panel to be collapsed by clicking on its own header? If not, the header should not appear to be interactive when mousing over it
* Figma file was off, the accordion headers should be 52px tall (checked with Matt)
* Accordion’s should have a chevron icon on the right side

On to reviewing the code files ...

devinbinnie:
> * On view load, first accordion animates open, is that intentional?

Yes it is. I'm not sure if it was intended in the design but I think it works. I can remove if need be.

> * Should there be a hover state on the accordion headers?

@matthewbirtch ?

> * There is a 12px gap between the edge of the header button and the visual bounds of the header. The interactive area should extend right to the edges of the visual area.

Fixed!

> * The primary button is missing interaction states (hover, active etc)

I'm going to be adding hover in a follow-up ticket. Given the timing of the Button Common Component, Matt and I agreed that if I added the hover that that would be sufficient. Ticket here: https://mattermost.atlassian.net/browse/MM-27159

> * Are we allowing an accordion panel to be collapsed by clicking on its own header? If not, the header should not appear to be interactive when mousing over it

Fixed!

> * Figma file was off, the accordion headers should be 52px tall (checked with Matt)

Fixed!

> * Accordion’s should have a chevron icon on the right side

As per [this comment](https://github.com/mattermost/mattermost-webapp/pull/5861#issuecomment-659536288) I think we're going to address that in a follow-up ticket. We can do it here though if need be.

components/next_steps_view/next_steps_view.scss
```diff
@@ -54,22 +55,150 @@
...
+        cursor: default;
+    }
+
+    .NextStepsView__cardHeaderBadge {
+        color: var(--button-color);
+        background-color: var(--sidebar-text-active-border);
+    }
+}
+
+.Card.complete {
```
deanwhillier, line 159: Does it make sense to have the complete state be attached to a generic card? If you use _something like_ `.Steps` at the accordion level and `.Steps_step` for each card, then the `.complete` state can be more appropriately attached.

matthewbirtch:
> * On view load, first accordion animates open, is that intentional?

I don't think it needs to animate open on first load. The appropriate accordion should already by open

> * Should there be a hover state on the accordion headers?

Yes, sorry for not including this. Here's what I propose:
image

- border color changes to: `--center-channel-color-16` 
- chevron icon changes to `--center channel-color-64` 
- the background color on the number circle changes to `--center-channel-color-24`
- the drop shadow changes to `box-shadow: 0px 4px 6px rgba(0, 0, 0, 0.12);`
It's a subtle change, but it's enough

components/card/card.tsx
```diff
@@ -30,12 +31,12 @@ export default class Card extends React.PureComponent<Props> {
 
         return (
             <div
-                className={classNames('Card', {
+                className={classNames(`Card${this.props.className ? ` ${this.props.className}` : ''}`, {
```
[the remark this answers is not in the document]
devinbinnie: Cards on the table (pun definitely intended :P) I didn't know you could do that. Fixed.

components/next_steps_view/next_steps_view.scss
```diff
@@ -27,10 +28,10 @@
     font-size: 16px;
     line-height: 24px;
     margin-top: 8px;
-    margin-bottom: 0;
 }
 
 .NextStepsView__header-logo {
+    margin-left: auto;
```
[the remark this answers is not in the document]
devinbinnie: Apparently so...

components/next_steps_view/next_steps_view.scss
```diff
@@ -54,22 +55,145 @@
...
 
-@media screen and (max-width: 768px) {
-    #app-content.NextStepsView {
-        background-size: auto;
+.NextStepsView__body .Card__header {
+    margin: 0 12px;
+    padding: 0;
+}
+
+.NextStepsView__cardHeader {
```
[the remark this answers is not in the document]
devinbinnie: I think I fixed this already based on your previous comment.

components/next_steps_view/next_steps_view.scss
devinbinnie: I'm not seeing how this is any more or less effective, but I think it goes back to our standing argument about the separation of behaviour between the `Card` and the `Accordion`.

I think for now I'd like to keep it as is and we can revisit it when we revisit that conversation.

PR #5894, opened 2020-07-10
MM-26733 - Export Textbox so plugins can use it

cpoile:
#### Summary
- Tried to compile the Textbox from a plugin, and that was a mess.
- Instead, export the connected component into the window object.
- Now we're able to send in a custom input component to override `Textbox`'s `AutosizeTextArea`.
- In `QuickInput` we don't need the `channelId` so remove it, and if we're not using an `AutosizeTextArea` we don't need to pass in the `onHeightChange` prop (these were causing warnings went sent to a plain input, and I wanted to clean them up).
- Open to a better organization for the exported component. Maybe `window.components.Textbox`? -- done

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26733

### DIFF
--- a/components/textbox/index.ts
+++ b/components/textbox/index.ts
@@ -52,5 +52,7 @@ const connectedTextbox = connect(makeMapStateToProps, mapDispatchToProps, null, 
 export default connectedTextbox;
 
 // Attach the Textbox to the window object so that plugins can use it.
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
 window.Textbox = connectedTextbox;
 
--- a/components/textbox/index.ts
+++ b/components/textbox/index.ts
@@ -48,11 +48,5 @@ const mapDispatchToProps = (dispatch: Dispatch<GenericAction>) => ({
     }, dispatch),
 });
 
-const connectedTextbox = connect(makeMapStateToProps, mapDispatchToProps, null, {forwardRef: true})(Textbox);
-export default connectedTextbox;
-
-// Attach the Textbox to the window object so that plugins can use it.
-// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-// @ts-ignore
-window.Textbox = connectedTextbox;
+export default connect(makeMapStateToProps, mapDispatchToProps, null, {forwardRef: true})(Textbox);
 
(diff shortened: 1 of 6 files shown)


components/textbox/index.ts
```diff
@@ -48,4 +48,9 @@ const mapDispatchToProps = (dispatch: Dispatch<GenericAction>) => ({
     }, dispatch),
 });
 
-export default connect(makeMapStateToProps, mapDispatchToProps, null, {forwardRef: true})(Textbox);
+const connectedTextbox = connect(makeMapStateToProps, mapDispatchToProps, null, {forwardRef: true})(Textbox);
+export default connectedTextbox;
+
+// Attach the Textbox to the window object so that plugins can use it.
+window.Textbox = connectedTextbox;
```
scastillo, line 57: I'm new here but maybe its cleaner to add this to the plugins export.js file?

https://github.com/mattermost/mattermost-webapp/blob/master/plugins/export.js

cpoile: Thank you @scastillo, good eye! I can't believe I missed that. 🤦

plugins/export.js
```diff
@@ -15,8 +16,11 @@ window.Redux = require('redux');
...
 window.ReactRouterDom = require('react-router-dom');
-window.PostUtils = {formatText, messageHtmlToComponent};
 window.PropTypes = require('prop-types');
 window.PDFJS = require('pdfjs-dist');
+
+// Functions and components exposed on window for plugins to use.
+window.PostUtils = {formatText, messageHtmlToComponent};
 window.openInteractiveDialog = openInteractiveDialog;
 window.WebappUtils = {browserHistory};
+window.Components = {Textbox};
```
agarciamontoro, line 26: The Components namespace is nice :)

components/textbox/textbox.tsx
```diff
@@ -234,6 +235,7 @@ export default class Textbox extends React.PureComponent<Props> {
                 </div>
             );
         }
+        const inputComponent = this.props.inputComponent || AutosizeTextarea;
```
lieut-data, line 238: Should we configure this in `defaultProps` instead?

cpoile: Yes, good idea! And I took out the null as a possible value of `inputComponent` (`defaultProps` will not replace a null value).

PR #5897, opened 2020-07-13
MM-26628 Fix for empty space for new category in mobile view

sudheerDev:
#### Summary
  * Override height for small screens new category

There is a specificity override for height for smaller screens making the `height:0` in effective

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26628

#### Screenshots
Untitled

### DIFF
--- a/sass/layout/_sidebar-left.scss
+++ b/sass/layout/_sidebar-left.scss
@@ -789,6 +789,10 @@
         padding: 0;
     }
 
+    .SidebarChannelGroup .SidebarChannel.newChannelSpacer {
+        height: 0px;
+    }
+
     /* Channels */
     .SidebarChannel {
         display: block;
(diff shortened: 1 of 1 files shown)


sass/layout/_sidebar-left.scss
```diff
@@ -789,6 +789,10 @@
         padding: 0;
     }
 
+    .SidebarChannelGroup .SidebarChannel.newChannelSpacer {
```
hmhealey, line 792: Is all of this necessary, or could we get away with simplifying the selector to just `.newChannelSpacer`?

sudheerDev: we need the nesting to override the existing style.
Untitled


we need to override styles to take into account the `height: 0px` property

hmhealey: Ugh, I see. Specificity sucks to deal with.

PR #5899, opened 2020-07-14
[MM-26824] Apply filter component to channels listing and add ability to filter by team

fmunshi:
#### Summary
- Builds off of https://github.com/mattermost/mattermost-webapp/pull/5882
- Adds filtering to the channel list in the system console

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26824

#### Related Pull Requests
- Redux: https://github.com/mattermost/mattermost-redux/pull/1192
- Server: https://github.com/mattermost/mattermost-server/pull/15009

#### Screenshots
Screen Shot 2020-07-13 at 9 10 41 PM


components/admin_console/filter/team_filter_dropdown/team_filter_dropdown.tsx
```diff
@@ -0,0 +1,328 @@
...
+    page: number;
+    loading: boolean;
+    show: boolean;
+    savedSelectedTeams: Team[];
+    searchResults: Team[];
+    searchTerm: string;
+    searchTotal: number;
+}
+
+const getSelectedTeams = createSelector(
```
hahmadia, line 46: Do we want to define selectors within the component level? I feel like having these elsewhere may be a better option? I'm not sure if this is something we are already doing or not.

components/admin_console/team_channel_settings/channel/list/channel_list.tsx
```diff
@@ -60,22 +64,23 @@ export default class ChannelList extends React.PureComponent<ChannelListProps, C
     }
 
     getPaginationProps = () => {
-        const {page, term} = this.state;
-        const total = term === '' ? this.props.total : this.state.total;
+        const {page, term, filters} = this.state;
+        const isSearching = term.length > 0 || JSON.stringify(filters) !== JSON.stringify({});
```
hahmadia, line 68: Couldn't it just be '{}' ?

components/admin_console/filter/team_filter_dropdown/team_filter_dropdown.tsx
```diff
@@ -0,0 +1,328 @@
...
+    searchTeams = async (term: string, page: number) => {
+        let searchResults = [];
+        let searchTotal = 0;
+        const response = await this.props.actions.searchTeams(term, page, TEAMS_PER_PAGE);
+        if (response?.data) {
+            searchResults = page > 0 ? this.state.searchResults.concat(response.data.teams) : response.data.teams;
+            searchTotal = response.data.total_count;
+        }
+        this.setState({page, loading: false, searchResults, searchTotal, searchTerm: term});
+    }
```
mkraft, line 133: What about error handling for `loadMore` and `searchTeams`? What happens in the UI?

fmunshi: If this fails it will just show `No results found`, which is sufficient for this case i think

components/admin_console/filter/team_filter_dropdown/team_filter_dropdown.tsx
fmunshi: This selector is specifically for this component it makes most sense for it to stay here, we already use selectors in a couple of components to memoize results from computed values

components/admin_console/filter/team_filter_dropdown/team_filter_dropdown.tsx
mkraft: Recently in another PR this came up and UX was planning to do something consistent for all potential server errors I believe. Not sure if we're there yet. cc @hahmadia @michaelgamble do either of you know where that landed?

hahmadia: In my case, if the search fails, instead of showing no results, Michael had me update the ticket to something like "The search failed. Please try again" instead of no results found. That's what we did and we made a a ticket to enhance the text by making the try again clickable and it will redo the search query. But that's as far as we got, I'm not sure if Michael has other thoughts other than what we discussed at that time.

fmunshi: I'll discuss this with him in my ux review now

components/admin_console/filter/team_filter_dropdown/team_filter_dropdown.tsx
```diff
@@ -152,14 +157,24 @@ class TeamFilterDropdown extends React.PureComponent<Props, State> {
             const {data} = response;
             searchResults = page > 0 ? this.state.searchResults.concat(data.teams) : data.teams;
             searchTotal = data.total_count;
+            this.setState({page, loading: false, searchResults, searchTotal, searchTerm: term});
+            return;
         }
-        this.setState({page, loading: false, searchResults, searchTotal, searchTerm: term});
+        this.searchRetryInterval = this.searchRetryInterval * 2;
+        this.searchRetryId = window.setTimeout(this.searchTeams.bind(null, term, page), this.searchRetryInterval)
```
fmunshi, line 164: Hey @mkraft @hahmadia I added this search retry logic here which retries search if it fails (after 600ms then 1200ms then 2400ms doubling every time). After discussing with Michael this was the solution that he wanted to be implemented instead of having the user click on a retry button it would keep retrying at longer intervals and show a loading spinner in the meantime... If the search term changes then the search retry is cleared

PR #5901, opened 2020-07-14
[MM-24300/MM-28305] Upgrade dependencies 5.28

calebroseland:
#### Summary
As a follow up to #5743, upgrade the harder-to-update dependencies. (No functional changes.)

Major upgrades/changes:

- **Use Node 10.22 in CI** (thanks @metanerd!)

- **Jest `24` -> `26`**

- **ESLint `6` -> `7`**

- **Config changes and refactoring.** (details in self-review comments)
Many of these were necessary due to stricter TSC/ESLint checks or other breaking changes. 

- **React-intl `3` -> `5`** ([4x guide](https://formatjs.io/docs/react-intl/upgrade-guide-4x)) ([5x guide](https://formatjs.io/docs/react-intl/upgrade-guide-5x))

`<FormattedMessage>`/`intl.formatMessage` now accept XML/HTML placeholders to support Rich Text Formatting,
and `<FormattedHTMLMessage>` and `intl.formatHTMLMessage` have been removed.

_Aside note: if `IntlProvider`'s `wrapRichTextChunksInFragment` option is enabled in the future, <FormattedMessage> chunks would be wrapped in fragments, causing issues with `<FormattedMessage>{(thing: string) => /*...*/ }</FormattedMessage>`) usage._ 

**Breaking**: translation strings that contain `<`...`>` need to be escaped.
> XML/HTML tags are escaped using apostrophe just like other ICU constructs.

(See: http://userguide.icu-project.org/formatparse/messages#TOC-Quoting-Escaping)

This is a breaking change to some conflicting values in `i18n/*.json` and `<... defaultMessage='...'>`. 

These key/values have been escaped to prevent runtime errors in all supported langs:
```
admin.cluster.OverrideHostnameDesc
admin.gitlab.enableDescription
admin.gitlab.EnableMarkdownDesc
admin.saml.assertionConsumerServiceURLEx
admin.saml.serviceProviderIdentifierEx
admin.permissions.inherited_from
edit_post.time_limit_modal.description
marketplace_modal.list.update_confirmation.message.current_with_release_notes
marketplace_modal.list.update_confirmation.message.warning_major_version_with_release_notes
```
- [ ] (After merge) **Escape translated strings in weblate**

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-24300
  - https://mattermost.atlassian.net/browse/MM-26739
  - https://mattermost.atlassian.net/browse/MM-26735
- https://mattermost.atlassian.net/browse/MM-28305

#### Related PRs
- [x] https://github.com/mattermost/mattermost-redux/pull/1194


calebroseland:
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33822

calebroseland:
CI will fail until https://github.com/mattermost/mattermost-redux/pull/1194 is merged.

.eslintrc.json
```diff
@@ -4,10 +4,12 @@
     "plugin:cypress/recommended"
   ],
   "plugins": [
+    "babel",
```
hmhealey, line 7: At some point, these changes should be rolled back into https://github.com/mattermost/eslint-plugin-mattermost so that they can be used in mattermost-redux and the mobile app

components/activity_log_modal/components/more_info.tsx
```diff
@@ -22,7 +22,7 @@ type Props = {
     moreInfo: boolean;
 };
 
-export default function MoreInfo(props: Props) {
+export default (props: Props) => {
```
hmhealey, line 25: This should be given a name somehow since it will otherwise be unnamed in dev tools and [snapshots](https://github.com/mattermost/mattermost-webapp/pull/5901/files#diff-917a01dc9366f5c92a276e4d5572ab28R47). Maybe something like
```
const MoreInfo = () => {...};
export default MoreInfo;
```
would work?

components/admin_console/__snapshots__/elasticsearch_settings.test.jsx.snap
```diff
@@ -13,7 +13,6 @@ exports[`components/ElasticSearchSettings should match snapshot, disabled 1`] =
       <FormattedMessage
         defaultMessage="Elasticsearch"
         id="admin.elasticsearch.title"
-        values={Object {}}
```
hmhealey, line 16: Down with `defaultProps`! It's really annoying how changing it affects everywhere the component is even shallowly mounted

components/admin_console/admin_settings.tsx
```diff
@@ -55,6 +55,7 @@ export default abstract class AdminSettings extends React.PureComponent<Props, S
 
     protected abstract getStateFromConfig(config: AdminConfig): State;
 
+    // eslint-disable-next-line @typescript-eslint/ban-types
```
hmhealey, line 58: I assume the `object` type is banned by default? Would it make sense to use `any` instead?

.eslintrc.json
calebroseland: I agree 👍. I'll follow up with a sync PR (migrate relevant to `eslint-plugin-mattermost`). I was thinking to add there directly, but didn't want conflict/errors in-between merges.

components/admin_console/manage_roles_modal/manage_roles_modal.tsx
```diff
@@ -15,16 +17,17 @@ import {trackEvent} from 'actions/diagnostics_actions.jsx';
...
+export type Props = {
     show: boolean;
     user?: UserProfile;
     userAccessTokensEnabled: boolean;
 
     // defining custom function type instead of using React.MouseEventHandler
     // to make the event optional
     onModalDismissed: (e?: React.MouseEvent<HTMLButtonElement>) => void;
-    actions: { updateUserRoles: Function };
+    actions: { updateUserRoles: (userId: string, roles: string) => ActionFunc}; // TODO add generics to ActionFunc
```
hmhealey, line 30: This type should be something along the lines of `() => Promise<Something>` since `updateUserRoles` has been wrapped by `mapDispatchToProps`. If that typing isn't working out right, perhaps `mapDispatchToProps`'s type could be changed to take `dispatch: DispatchFunc` since that's the weird, overridden one we're using in mattermost-redux

components/admin_console/member_list_group/index.ts
```diff
@@ -24,8 +24,11 @@ type Props = {
 }
 
 type Actions = {
+    // eslint-disable-next-line @typescript-eslint/ban-types
```
hmhealey, line 27: What types are banned here?

components/admin_console/member_list_group/member_list_group.tsx
```diff
@@ -20,8 +20,11 @@ type Props = {
     groupID: string;
     total: number;
     actions: {
+        // eslint-disable-next-line @typescript-eslint/ban-types
```
hmhealey, line 23: Same here

components/admin_console/set_by_env.jsx
```diff
@@ -4,7 +4,7 @@
 import React from 'react';
 import {FormattedMessage} from 'react-intl';
 
-export default function() {
+export default () => {
```
hmhealey, line 7: This could also use a proper display name

components/activity_log_modal/components/more_info.tsx
calebroseland: 👍 I'll change it to this. Was because of function naming convention. Should we change `naming-convention` to allow PascalCase for functions?

components/admin_console/admin_settings.tsx
calebroseland: yes, `object` is banned. I think you're right,`any`, or maybe `Record<string, any>` would be better than an ignore comment.

components/formatted_markdown_message.jsx
```diff
@@ -21,7 +21,7 @@ export class CustomRenderer extends marked.Renderer {
             return text;
         }
         if (href[0] === TARGET_BLANK_URL_PREFIX) {
-            return `<a href="${href.substring(1, href.length)}" rel="noreferrer" target="_blank">${text}</a>`;
+            return `<a href="${href.substring(1, href.length)}" rel="noopener noreferrer" target="_blank">${text}</a>`;
```
hmhealey, line 24: Technically, `noreferrer` implies `noopener`, but no harm in including both of them.

components/help/components/attaching.tsx
```diff
@@ -7,7 +7,7 @@ import {Link} from 'react-router-dom';
 
 import FormattedMarkdownMessage from 'components/formatted_markdown_message.jsx';
 
-export default function HelpAttaching(): JSX.Element {
+export default (): JSX.Element => {
```
hmhealey, line 10: This also needs a DisplayName

components/admin_console/member_list_group/index.ts
calebroseland: `{}`
...
> Don't use `{}` as a type. `{}` actually means "any non-nullish value".

I'll change these action types up

components/html_entities/nbsp.tsx
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// https://stackoverflow.com/a/54485712
+export default () => '\u00A0';
```
hmhealey, line 5: And so does this one. Also, weird that we need to do this, but I guess it makes sense. I'm surprised we still use nbsps, but I assume a lot of them are from older code mostly.

components/intl_provider/intl_provider.jsx
```diff
@@ -69,6 +69,7 @@ export default class IntlProvider extends React.PureComponent {
                 locale={this.props.locale}
                 messages={this.props.translations}
                 textComponent='span'
+                wrapRichTextChunksInFragment={true}
```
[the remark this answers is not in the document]
hmhealey: That's strange that this isn't the default option since it seems like it should be the default

components/new_channel_flow/index.ts
```diff
@@ -28,7 +21,7 @@ function mapStateToProps(state: GlobalState) {
 
 function mapDispatchToProps(dispatch: Dispatch<GenericAction>) {
     return {
-        actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc>, Actions>({
+        actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc>, Props['actions']>({
```
hmhealey, line 24: That's a nice way to avoid duplicating this. Ideally, I don't think we'd want to specify the types here at all, but this is still a lot nicer than before.

components/toggle_modal_button_redux/toggle_modal_button_redux.jsx
```diff
@@ -9,7 +9,7 @@ import {intlShape} from 'utils/react_intl';
 
 class ModalToggleButtonRedux extends React.PureComponent {
     static propTypes = {
-        accessibilityLabel: PropTypes.string,
+        accessibilityLabel: PropTypes.node,
```
hmhealey, line 12: Is this okay to change? We pass this into an aria-label, so it might need to be a string

components/widgets/separator/basic-separator.tsx
```diff
@@ -4,7 +4,7 @@
 import React from 'react';
 import './separator.scss';
 
-export default class BasicSeparator extends React.PureComponent<React.PropsWithChildren<{}>> {
+export default class BasicSeparator extends React.PureComponent<React.PropsWithChildren<any>> {
```
hmhealey, line 7: Instead of using `React.PropsWithChildren` with an `any` in it, we could manually define our own Props type as `{children: React.ReactNode}`

tests/setup.js
```diff
@@ -75,6 +75,23 @@ afterEach(() => {
     }
 });
 
+jest.mock('react-intl', () => {
```
hmhealey, line 78: Why do we need to partly mock this? The hook isn't supported by Jest?

components/channel_header_mobile/channel_info_button/__snapshots__/channel_info_button.test.js.snap
```diff
@@ -133,10 +135,13 @@ exports[`components/ChannelHeaderMobile/ChannelInfoButton should match snapshot,
             <FormattedMessage
               defaultMessage="Info Icon"
               id="generic_icons.info"
-              values={Object {}}
             >
               <svg
-                aria-label="Info Icon"
+                aria-label={
```
hmhealey, line 140: Is it okay that we're passing an array of strings into `aria-label`?

components/toggle_modal_button_redux/toggle_modal_button_redux.jsx
calebroseland: No it's not, and thanks for catching. I believe it was a "correction" to a console error. Should've dug deeper... I think there's place higher up that accepts a React.ReactNode and eventually passes to it accessibilityLabel. I'll try to find it, or revert this if I can't.

components/channel_header_mobile/channel_info_button/__snapshots__/channel_info_button.test.js.snap
calebroseland: No. Sigh... all the `../widgets/icons/**.jsx` svg icons are use the child func syntax.
```jsx
<FormattedMessage>
    {(ariaLabel) => (
        <svg>
    )}
</FormattedMessage>
```

I'll get those all changed over.

components/widgets/separator/basic-separator.tsx
calebroseland: done

components/admin_console/member_list_group/member_list_group.tsx
calebroseland: done

components/admin_console/set_by_env.jsx
calebroseland: done

components/html_entities/nbsp.tsx
calebroseland: done

components/help/components/attaching.tsx
calebroseland: done

components/toggle_modal_button_redux/toggle_modal_button_redux.jsx
calebroseland: reverted

amyblais:
I'm unsure if this should be included in v5.26 if it doesn't make it to today's feature complete as there's lot of file changes and it appears risky. Maybe @calebroseland or @grundleborg can confirm.

grundleborg:
@calebroseland I'm not really comfortable reviewing this ticket. Could you assign someone else who is strong on the js/frontend side of things in my place?

grundleborg:
Deleted my previous comment as I'm not assigned as a reviewer :facepalm:

components/admin_console/manage_roles_modal/manage_roles_modal.tsx
calebroseland: Agreed. Updated ✓

components/activity_log_modal/components/more_info.tsx
hmhealey: Yeah, that'd be fine. If there was any way to only allow it for top level functions, that'd be nice, but I assume it's not that specific.

components/admin_console/filter/filter.tsx
```diff
@@ -188,6 +188,7 @@ class Filter extends React.PureComponent<Props, State> {
                 ref={this.filterRef}
             >
                 <button
+                    // eslint-disable-next-line @typescript-eslint/naming-convention
```
hmhealey, line 191: I missed this earlier, but this might be something to keep in mind when discussing our CSS naming conventions in the future

components/admin_console/admin_settings.tsx
hmhealey: Either should be fine here. The whole way that AdminSettings uses inheritance is already an anti-pattern for React (my bad 😛), so if we have to add some special handling to get around it, I'm fine with that.

tests/setup.js
calebroseland: Did this while testing `OverlayTrigger`, no longer needed; should've removed. (Done now)

components/channel_header_mobile/channel_info_button/__snapshots__/channel_info_button.test.js.snap
hmhealey: Ah, that's why `wrapRichTextChunksInFragment` isn't default. That answers that question then.

Thanks for changing all those. I'm also realizing we should make it so that `make i18n-extract` can recognize `useIntl` when looking for translation strings so that we don't need to `t` when using it

components/admin_console/filter/filter.tsx
calebroseland: Yeah, came in after-the-fact. Will probably be other conflicts like this as other PRs get merged.

Should I go ahead and loosen `memberLike` in eslint?

hmhealey:
I think it's fine for this to not go into the release since it's not actually fixing any bugs. In my mind, the most important part of having this done for "5.26" is just to ensure we actually do it within the 5.26 time frame, even if it doesn't go into the release itself. We'll just have to be aware that it may cause some merge conflicts when cherry-picking onto the release branch.

calebroseland:
@hmhealey, @amyblais I agree this can, and probably should not go into 5.26, as we're out of time to let things soak and there will be rolling conflicts (other PRs) that will need to be resolved.

> We'll just have to be aware that it may cause some merge conflicts when cherry-picking onto the release branch.

Cherry-pick into 5.27? Or should it continue to target master and go into 5.28?

components/channel_header_mobile/channel_info_button/__snapshots__/channel_info_button.test.js.snap
calebroseland: There's at least another ~25 instances of the FormattedMessage func child pattern that could be hiding issues. I found one that will need to be resolved before this is ready.

That'd be great! (`i18n-extract`/`useIntl`/`t()`)

amyblais:
I'd say target master and go into 5.28.

hmhealey:
> Cherry-pick into 5.27? Or should it continue to target master and go into 5.28?

I meant when cherry-picking future PRs into the release branch since they may have differences in things like snapshots or if they want to use the new hooks which aren't available in the release branch

hmhealey:
Eric is out until I believe Wednesday next week. Are we sure he needs to review this?

calebroseland:
I think we got it sorted out (wait til 5.28), I'll remove him. We can bring it up later/discuss with the team.

jgilliam17 (approved):
Thank you @calebroseland 
Tested, looks good to merge.
- Verified PDF-attachment previews are as expected.
- E2Es run was completed with few failed tests known on master - Thank You @saturninoabril 
See full report [here](https://mm-cypress-report.s3.amazonaws.com/927-webapp-pr-5901-30c163d/mochawesome.html)

PR #5919, opened 2020-07-17
[MM-26958] fix sort order of channel suggestion list

ashishbhate:
#### Summary

- Fix sort order of channel suggestion list: first DMs, then GMs, then Private channels, then Public channels

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26958
Parent https://mattermost.atlassian.net/browse/MM-15477

#### Related Pull Requests

- Has redux changes 
https://github.com/mattermost/mattermost-redux/pull/1198

#### Screenshots

image

(racerx matches because that user's nickname is "alice")


package-lock.json
```diff
@@ -14313,6 +14316,12 @@
         "postcss": "^7.0.14"
       }
     },
+    "icu4c-data": {
+      "version": "0.67.2",
+      "resolved": "https://registry.npmjs.org/icu4c-data/-/icu4c-data-0.67.2.tgz",
+      "integrity": "sha512-OIRiop+k1IVf4TBLEOj910duoO9NKwtJLwp++qWT6KT5gRziHNt+5gwhcGuTqRy++RTK2gLoAIbk8KYCNxW++g==",
+      "dev": true
+    },
```
fmunshi, line 14324: You will want to want to revert these changes to package-lock and only include the changes to the redux version

ashishbhate: @fmunshi reverted the changed to `package-lock`

hmhealey:
The code looks good, but we need to wait for the redux PR to be merged so that we can point redux back at master.

I do think it's a bit weird that we're putting DMs before public/private channels when you haven't typed an at sign, but that's something for UX to decide.

ashishbhate:
>I do think it's a bit weird that we're putting DMs before public/private channels when you haven't typed an at sign, but that's something for UX to decide.

The parent ticket has more details on how the search is expected to behave with and without the `@` 

https://mattermost.atlassian.net/browse/MM-15477

PR #5925, opened 2020-07-17
MM-26833: Lowercase login ID before submitting in LoginMFA

agarciamontoro:
#### Summary

The login controller is correctly lowercasing the input email right before submitting: https://github.com/mattermost/mattermost-webapp/blob/7d9e4f0614c05e678ed9b23c0a3d04fe3f866e76/components/login/login_controller/login_controller.jsx#L213

However, the email is passed as is to the LoginMFA component, which did not handle any case transformations: https://github.com/mattermost/mattermost-webapp/blob/7d9e4f0614c05e678ed9b23c0a3d04fe3f866e76/components/login/login_controller/login_controller.jsx#L800-L804

This caused a bug that only reproduces when the user is using multi-factor authentication, uses an email with uppercase letters and the server is using Postgres.

This PR solves the problem in the webapp, fixing this specific bug for now.

However, we need to solve this in the server. The problem there is that we do store the lowercase string in the database: the SQL store calls [PreSave on the user](https://github.com/mattermost/mattermost-server/blob/93a537a636299e1e049505bdaf072f8d3ae974a4/store/sqlstore/user_store.go#L106) before inserting into the database, which uses [NormalizeEmail](https://github.com/mattermost/mattermost-server/blob/93a537a636299e1e049505bdaf072f8d3ae974a4/model/user.go#L314-L316) to lowercase the email. However, the login process does not lowercase the input, checking two potentially different strings at the moment of the login.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26833


amyblais:
@sudheerDev @calebroseland Kind reminder to review for v5.26.

amyblais:
@agarciamontoro Jelena is off this week, maybe another QA can review this?

agarciamontoro:
Thank you for the heads-up, Amy! Assigning to @lindalumitchell because she'll know better than me who to assign this to.

lindalumitchell:
Thanks @amyblais and @agarciamontoro, I can take this one and will review after this morning's meetings.

agarciamontoro:
Thank you, Linda!

amyblais:
@agarciamontoro Kind reminder to cherry pick this,

agarciamontoro:
Thank you, Amy!

PR #5926, opened 2020-07-17
[MM-26960] Fix Archive Icon not showing up for archived channels on system console > Channels page

hahmadia:
#### Summary

Archive icon not showing up on channels page for archived channels is fixed in this PR.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26960

#### Screenshots

image


fmunshi:
Why 3 different classes applied on all 3 elements instead of just applying one class?

hahmadia:
> Why 3 different classes applied on all 3 elements instead of just applying one class?

No reason. Probably can do that ... Just didn't realise or atleast I thought this way was better haha. Will fix that up

hahmadia:
> Why 3 different classes applied on all 3 elements instead of just applying one class?

I just realised that not all of them use the 3 properties. Only some of them use some so it's easier to break it out into 3. Do you still think better if I merge together somehow?

components/admin_console/team_channel_settings/channel/list/channel_list.scss
```diff
@@ -19,6 +20,19 @@
...
             }
+
+            .channel-icon___left2Pixel {
+                left: 2px;
+            }
+
+            .channel-icon___removeLeft2Pixel {
+                left: 0px;
+            }
+
```
fmunshi, line 31: you got an extra line here

components/admin_console/team_channel_settings/channel/list/channel_list.scss
```diff
@@ -19,6 +20,19 @@
...
+            	
+            .channel-icon___increaseMargin {
+                margin: 0 10px 0 0px;
+            }
```
deanwhillier, line 35: These class names are mixing semantic and utility css techniques making them neither and loosing the benefits of both. The semantic css approach is about naming based on what an element is (not how it looks or where it's placed etc.) and the utility css approach names classes by what they do to the elements they are applied to (style, positioning etc). Combining semantic and utility naming into a single class name removes the flexibility that either approach would otherwise provide. My personal preference is to use the semantic approach by default (especially for known components and more static layouts) and only resort to utility classes where absolutely necessary (variable layouts etc.) and where semantic class names become unwieldy.

Some things to consider in this case:

* The added classes could become quite brittle long term.
* What happens if additional icons are added that have different sizes from what is already used?
* What if the existing icons are replaced and again differ in size from the original icons?
* You'll end up needing to write more code, styles and class names each time to compensate for the above.

What if instead you could generically position the icons and text in this context so that you don't need any of the utility classes and don't need additional code/styles each time an icon changes? You could base the styling on a maximum icon size (that would need to be adhered to) so that any smaller icons would automatically position/align etc.

Here is what I might do in this case (based on the generated html I see in the browser):

1. Remove the following classes entirely (from scss files and html elements):

* `.channel-icon___lowerOpacity`
* `.channel-icon___left2Pixel`
* `.channel-icon___increaseMargin`

2. Remove the following scss from `sass/layout/_channel-icon.scss`:

```scss
.channel-icon__lock,
.channel-icon__globe {
    top: 1px;
}

.channel-icon__archive,
.channel-icon__unarchive {
    left: -2px;
    top: 2px;
}
```

If an icon needs to be adjusted because there is something not quite right with the svg/image's placement/sizing of the icon itself, it should either be fixed or be adjusted with css based on the icon itself, not necessarily where it is used. If there's seemingly nothing wrong with the icon itself but instead with where it is used, I would first look to the styling around the icon before resorting to adjusting the icon's sizing/placement this specifically.

I would create this new scss:

```scss
.ChannelsList .channel-icon {
    display: inline-flex;
    justify-content: center;
    margin-right: 10px;
    opacity: 0.8;
    width: 16px;
}
```

This will generically style and position all the icons so they line up horizontally within their container, have the correct opacity, enforce spacing between the icon and the channel name text and support any icons that are 16px in width or smaller.

Note: I'm using the `.ChannelsList` class name as part of the selector to make this css more specific to this use case because I'm not familiar with all the places the `.channel-icon` class could be used so am not sure if these styles are appropriate in all cases (happy to talk more specifically about this).

3. This one is perhaps a touch more challenging depending on how the React components work. We need to apply some styles to the `.row-content` element as follows:

```scss
.row-content {
    display: flex;
    align-items: center;
}
```

This will use flexbox to vertically centre the contents of each row so that all the contents of the entire row are automatically vertically aligned to the centre of the row, icons and all. I'm 0/5 on whether you add a new class name to the existing `.row-content` element to specifically target the channel name column or just target all the columns as above.

The result of making those changes with dev tools:

image

As you can see, the single class name `.channel-icon` within the context of the `.ChannelsList` is now sufficient to properly style and position the icons and the other adjustments ensure the icons align vertically with the channel name.

Apologies for the length of this comment! Feel free to reach out if you'd like to discuss. Also, these are just suggestions at this point.

components/admin_console/team_channel_settings/channel/list/channel_list.scss
hahmadia: Removed these classes altogether

hahmadia:
> Thanks @hahmadia - this is fixed and the archived icons are showing up correctly for archived channels in System Console > Channels page.
> 
> Just a question... On the Jira ticket there is a note as follows - has the E2E cypress test been fixed too?:
> 
> > NOTE: Also please modify the archived_channels_spec.js and ensure the E2E cypress test is working fine, when this issue is fixed.

Oddly enough, I looked at that test and it has nothing to do with the system console portion or the work I'm doing. That test deals with the chat facing side of things as it's doing a `cy.visit(`/${team.name}/channels/${testChannel.name}`);` which has nothing to do with my work. I believe that test is okay to be left alone.

hahmadia:
As per @srkgupta comment on the Jira ticket, this is suppose to be cherry picked and merged into 5.26 release branch.

PR #5945, opened 2020-07-20
Cypress/E2E: Fix team members spec

josephbaylon:
#### Summary
- Fixed input locator
- Extracted method for searching

#### Ticket Link
None - both on master and release-5.26

Screen Shot 2020-07-20 at 1 11 33 PM

### DIFF
--- a/e2e/cypress/integration/enterprise/system_console/team_members_spec.js
+++ b/e2e/cypress/integration/enterprise/system_console/team_members_spec.js
@@ -11,14 +11,6 @@
 
 import * as TIMEOUTS from '../../../fixtures/timeouts';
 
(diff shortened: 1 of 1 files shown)


saturninoabril:
@srkgupta Thanks for catching! And my bad, it's introduced from my recently merged PR. Since this is to be cherry-picked into release-5.26 branch where `cy.apiAdminLogin` is still in old implementation, I suggest to do the fix on separate PR. Such separate PR will be for master only. Also need to fix the other one at https://github.com/mattermost/mattermost-webapp/blob/master/e2e/cypress/integration/enterprise/system_console/channel_members_spec.js#L28

PR #5951, opened 2020-07-21
MM-27135: Deactivated users show in channel members dropdown

agnivade:
We filter the user list by only activated users.

### Ticket link

https://mattermost.atlassian.net/browse/MM-27135

### DIFF
--- a/components/popover_list_members/popover_list_members.jsx
+++ b/components/popover_list_members/popover_list_members.jsx
@@ -9,6 +9,7 @@ import {FormattedMessage} from 'react-intl';
 
 import {browserHistory} from 'utils/browser_history';
 import {Constants, ModalIdentifiers} from 'utils/constants';
+import {isActive} from 'utils/filter_users';
 import * as Utils from 'utils/utils.jsx';
 import ChannelMembersModal from 'components/channel_members_modal';
 import OverlayTrigger from 'components/overlay_trigger';
@@ -16,8 +17,6 @@ import MemberIcon from 'components/widgets/icons/member_icon';
 import Popover from 'components/widgets/popover';
 
 import PopoverListMembersItem from 'components/popover_list_members/popover_list_members_item';
-
-const NOT_DELETED = 0;
 
 export default class PopoverListMembers extends React.PureComponent {
     static propTypes = {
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -15038,7 +15038,7 @@
     },
     "into-stream": {
       "version": "3.1.0",
-      "resolved": "http://registry.npmjs.org/into-stream/-/into-stream-3.1.0.tgz",
+      "resolved": "https://registry.npmjs.org/into-stream/-/into-stream-3.1.0.tgz",
```
Willyfrog, line 15041: can we remove the package-lock from the change? it doesn't seem related to it.

components/popover_list_members/popover_list_members.jsx
```diff
@@ -100,7 +100,8 @@ export default class PopoverListMembers extends React.PureComponent {
     render() {
         const isDirectChannel = this.props.channel.type === Constants.DM_CHANNEL;
 
-        const items = this.props.sortedUsers.map((user) => (
+        const activeUsers = this.props.sortedUsers.filter((user) => user.delete_at === 0);
```
Willyfrog, line 103: minor nitpick (non blocking): since 0 is a value used to detect a especific state and it is outside of the actual expected values, I'd change that into a named constant that expreses such meaning (something like "NOT_DELETED" or "ACTIVE")

package-lock.json
agnivade: Sure, to give some context to this, every time I make a webapp change, my npm version makes this change. I had been reverting the change since the past few months, so thought of including it this time. I will revert.

components/popover_list_members/popover_list_members.jsx
agnivade: IMO, this is fine not to be refactored. Actually "NOT_DELETED = 0" makes it even more confusing IMO.

Nevertheless, I will make the change. But I note that `delete_at === 0` is being used all over the code. We should change all of them to be consistent.

Willyfrog:
i definitely agree that we should change it throughout the code, thanks for taking one of the first steps towards that :D

amyblais:
@calebroseland Kind reminder to review for v5.26,

components/popover_list_members/popover_list_members.jsx
calebroseland: IMHO, I think a predicate function could help with clarity:

```ts
const NOT_DELETED_TIMESTAMP = 0;

export function isActive({delete_at}: User): boolean {
    return delete_at === NOT_DELETED_TIMESTAMP;
}

//...

import {isActive} from 'somewhere/utils/user';

const activeUsers = this.props.sortedUsers.filter(isActive);

```

That way any confusion behind the `0` constant contained in a clear `isActive` function.

agnivade: Done. IMO, if we are refactoring this to a separate function there wasn't a need to separately write "NEVER = 0".

PR #5952, opened 2020-07-21
Replacing blacklist with blocklist

jespino:
#### Summary
Replacing blacklist with blocklist

Jira ticket: https://mattermost.atlassian.net/browse/MM-27193

### DIFF
--- a/build/emoji/make-emojis
+++ b/build/emoji/make-emojis
@@ -70,7 +70,7 @@ end
 puts "\nThe following #{emoji_added.length} emoji were added:"
 print emoji_added
 
-blacklist = ['basecamp', 'basecampy']
+blocklist = ['basecamp', 'basecampy']
 
 categories = []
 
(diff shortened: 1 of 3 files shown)


hanzei:
Drive by comment: Is the decision to use certain terms like "Blocklist" worth documenting somewhere e.g. in https://developers.mattermost.com/contribute/getting-started/?

jespino:
@hanzei I think would be great to have some documentation around it, but we should decide what is the things that we want to try avoid in our code, blacklist was an easy one, but there are other. @jasonblais What do you think? is better to add something now about blacklist and grow from there, or is better to wait until have something more organized and write a more clear policy?

jasonblais:
I would wait until we've formalized a policy

hanzei:
I'm fine with waiting for a well defined policy as long it get added eventually.

PR #5953, opened 2020-07-21
MM-26075 Add type definitions and standard return to user/status/external api commands

saturninoabril:
#### Summary
Add type definitions and standard return to user/status/external api commands
- apiUpdateUserStatus
- apiPatchUser
- apiPatchMe
- apiRevokeUserSessions
- externalActivateUser

#### Ticket Link
Jira ticket - https://mattermost.atlassian.net/browse/MM-26075

### DIFF
--- a/e2e/cypress/support/api/user.js
+++ b/e2e/cypress/support/api/user.js
@@ -87,7 +87,7 @@ Cypress.Commands.add('apiPatchUser', (userId, userData) => {
         body: userData,
     }).then((response) => {
         expect(response.status).to.equal(200);
-        cy.wrap(response);
+        cy.wrap({user: response.body});
     });
 });
 
@@ -99,7 +99,7 @@ Cypress.Commands.add('apiPatchMe', (data) => {
         body: data,
     }).then((response) => {
         expect(response.status).to.equal(200);
-        cy.wrap(response);
+        cy.wrap({user: response.body});
     });
 });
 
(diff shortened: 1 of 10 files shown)


e2e/cypress/support/api/user.js
```diff
@@ -168,33 +168,12 @@ Cypress.Commands.add('apiCreateUser', ({prefix = 'user', bypassTutorial = true,
...
-            url: '/api/v4/users/me/status',
-            method: 'PUT',
-            body: data,
-        }).then((response) => {
-            expect(response.status).to.equal(200);
-            cy.wrap({status: response.body});
-        });
-    });
-});
-
```
srkgupta, line 197: Wondering why this API command is moved out to a separate file. I am not sure if we should break this to separate files at this granularity. I suggest to revert back and keep user related APIs into a single user file.

e2e/cypress/support/api_commands.js
```diff
@@ -661,18 +661,6 @@ Cypress.Commands.add('apiDeleteScheme', (schemeId) => {
...
- * Activate/Deactivate a User directly via API
- * @param {String} userId - The user ID
- * @param {Boolean} active - Whether to activate or deactivate - true/false
- */
-Cypress.Commands.add('apiActivateUser', (userId, active = true) => {
-    const baseUrl = Cypress.config('baseUrl');
-    const admin = getAdminAccount();
-
-    cy.externalRequest({user: admin, method: 'put', baseUrl, path: `users/${userId}/active`, data: {active}});
-});
```
srkgupta, line 674: There is a reference of this function in the test guest_identification_ui_spec. Can you please update this test

e2e/cypress/support/api/user.js
saturninoabril: Folder structure depends on how the API reference is structured - https://api.mattermost.com/#tag/status. Looking at the link, there are 3 endpoints for it. I think that's the best way to track whether we already implemented an API endpoint or not, and to prevent potential duplicates.

e2e/cypress/support/api_commands.js
saturninoabril: Thanks! I thought I changed it too but not. Now updated.

e2e/cypress/support/api/user.js
srkgupta: Got it. Thanks. That makes sense now :)

PR #5964, opened 2020-07-22
[MM-27255] Change default theme system console setting to a dropdown

hahmadia:
#### Summary

Updated default theme system console setting to a dropdown.

https://mattermost.atlassian.net/browse/MM-27255

image


components/admin_console/admin_definition.jsx
```diff
@@ -4393,7 +4393,7 @@ const AdminDefinition = {
                         label: t('admin.experimental.defaultTheme.title'),
                         label_default: 'Default Theme:',
                         help_text: t('admin.experimental.defaultTheme.desc'),
-                        help_text_default: 'Set a default theme that applies to all new users on the system.',
+                        help_text_default: 'Set a default theme that applies to all new users on the system with options "default", "organization", "mattermostDark", and "windows10".',
```
crspeller, line 4396: Presumably you can enter any theme JSON here and it will become default. Otherwise we would make this a dropdown. Maybe mention that?

hahmadia: @thefactremains ^^

hahmadia: @crspeller Actually to be honest, I think it should be a drop down. The reason for this is because if you look underneath it, we pass these `Options`
image
However, when we build the component, they aren't getting used
image

but if we look at the build drop down, options does get used
image

So gut feeling tells me that this should be a dropdown but I could be totally off here. I'm not sure it will actually accept any theme JSON.

hahmadia: @crspeller Spoke with Dennis and we also agreed a dropdown would be better as I discussed with you.

amyblais:
@hahmadia Kind note to cherry-pick this,

PR #5967, opened 2020-07-22
Cypress/E2E: Fix accessibility modals dialogs

josephbaylon:
#### Summary
- Added delay after tab to allow for css class to propagate
- Make sure row tested has drop-down actions (logged in user does not display drop-down, has to be a different user highlighted)
- Close popup modal

#### Ticket Link
None - master only

Screen Shot 2020-07-22 at 9 08 33 AM


saturninoabril:
I think this should be for master only since this was cut from 5.26.

josephbaylon:
@saturninoabril I think this isn't being run on 5.26 because it doesn't have the prod label but the behavior is the same if it's enabled there. But I changed it to master only since I haven't added the prod label

saturninoabril:
That's right. Let's only cherry-pick as needed.

PR #5968, opened 2020-07-22
[MM-26174] Automate SAML tests

hahmadia:
#### Summary

Automate SAML tests

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26174


e2e/cypress/integration/enterprise/saml/saml_automated_spec.js
```diff
@@ -0,0 +1,219 @@
...
+    } = Cypress.env();
+    const idpUrl = `${oktaBaseUrl}/app/${oktaMMAppName}/${oktaMMEntityId}/sso/saml`;
+    const idpMetadataUrl = `${oktaBaseUrl}/app/${oktaMMEntityId}/sso/saml/metadata`;
+
+    const newConfig = {
+        SamlSettings: {
+            Enable: true,
+            EnableSyncWithLdap: false,
+            EnableSyncWithLdapIncludeAuth: false,
+            Verify: true,
```
srkgupta, line 35: Currently all the test passes if Verify flag is set to false. Can you please confirm if this is a bug or if it happens on localhost because the server is not using encryption

e2e/cypress/integration/enterprise/saml/saml_automated_spec.js
```diff
@@ -0,0 +1,219 @@
...
+        });
+
+        it('SAML2 Implementation', () => {
+            cy.apiAdminLogin();
+            const test1Settings = {
+                ...newConfig,
+                ExperimentalSettings: {
+                    UseNewSAMLLibrary: true,
+                },
+            };
```
srkgupta, line 205: @saturninoabril, as per the expectations, this requires a server restart for the config to be effective. Any suggestions on how we can handle this while running our cypress test suite? My advise would be to use a Cypress env variable and run the tests with and without the NewSAML Library. This way we can have all the SAML related tests verified when this flag is turned on and off. We may not need a separate test for this. Please advise.

e2e/cypress/integration/enterprise/saml/saml_automated_spec.js
srkgupta: @hahmadia please ignore this. I just reinstalled the xmlsec library again on my localhost and it started to work fine again. Even @saturninoabril had the same issue and it worked fine once it was installed on his system.

e2e/cypress/integration/enterprise/saml/saml_automated_spec.js
srkgupta: @hahmadia can you please remove this testcase to use the newSAML Library. @saturninoabril and I had discussed this and we will now be running the SAML related cypress tests twice - i.e. with this config enabled and disabled. Since this requires a server restart for it to be effective, it needs to be controlled from outside and hence we don't need a test which changes it within the test (as it will not be effective). Thanks.

hahmadia: Removed!

hahmadia:
> @hahmadia @srkgupta Looks good to me. Is there TM4J test case/s for this? If none, I suggests to create first and and use it for each test block (same as [this](https://github.com/mattermost/mattermost-webapp/blob/master/e2e/cypress/integration/channel_settings/channel_name_validations_spec.js#L27)) so that this kind of test can be removed from future manual testing.

Not sure, I just completed what was given to me in this spreadsheet https://docs.google.com/spreadsheets/d/1uNMC4xmU4r0Z6MYgE0CpkhOQG18tU--mMfnJxeC2Kk0/edit?pli=1#gid=1301050503. 

@srkgupta Do you mind creating those tests and giving me their numbers so I can edit the tests and include it in the test title? Thank you

srkgupta:
Sure @hahmadia. Updated the testcase id in the spreadsheet. Please check and use the same. Thanks. 
https://docs.google.com/spreadsheets/d/1uNMC4xmU4r0Z6MYgE0CpkhOQG18tU--mMfnJxeC2Kk0/edit?pli=1#gid=1301050503

hahmadia:
@srkgupta @saturninoabril 
Done!

PR #5973, opened 2020-07-22
[MM-27164] Picture Selector Common Component

devinbinnie:
#### Summary
This is the profile picture selector for the profile setup step on the Cloud Onboarding screen.
Currently just setup to show the style and to be able to preview an uploaded image, and remove an existing image. Not hooked up to server yet.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27164

#### Some Considerations
- The size specified in the Figma doesn't match the usual size of image we size. We use 128px, the Figma shows a 96px box. This shouldn't be an issue as the CSS has been written to show any size in the 96px box, and the server still is responsible for resizing the image.
- There is no hover state on the select button. @matthewbirtch should there be?
- I've written a couple basic snapshot tests, but haven't gone any further as I don't have a lot of experience with Hooks yet. I will add more in the future.
- The current default profile picture puts the letter to the left, rather than centered like in the Figma. This default comes from the server. I'm thinking we address this in a follow-up ticket, as I'm guessing server work is required.


components/picture_selector.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    useEffect(() => {
+        if (!image) {
+            if (props.src) {
+                setImage(props.src);
+            } else if (props.defaultSrc) {
+                setImage(props.defaultSrc);
+            }
+        }
+    });
```
reflog, line 69: I think this hook should have a dependency parameter of [props.src, setImage]

components/picture_selector.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+type Props = {
+    src?: string;
+    defaultSrc?: string;
+    loadingPicture?: boolean;
+    onSelect: (file: File) => void;
+    onRemove: () => void;
+};
+
+const PictureSelector: React.FC<Props> = (props: Props) => {
```
reflog, line 20: 1/5 - it would be cleaner do descructure the props: `({src,defaultSrc,loadingPicture,onSelect,onRemove}:Props)`
this will clean up the function a bit

components/picture_selector.tsx
```diff
@@ -0,0 +1,124 @@
...
+                accept='.jpg,.png,.bmp'
+                type='file'
+                onChange={handleFileChange}
+                disabled={props.loadingPicture}
+                aria-hidden={true}
+                tabIndex={-1}
+            />
+            <div className='PictureSelector__imageContainer'>
+                <div
+                    className='PictureSelector__image'
```
reflog, line 103: since this div is actually an image, should probably have `aria` prop set too

components/picture_selector.tsx
devinbinnie: I like the cleaner look of the declaration personally, but I will destructure at the top of the function in a `const` so that we get basically the same effect.

components/next_steps_view/steps/complete_profile_step.tsx
```diff
@@ -21,10 +31,32 @@ export default class CompleteProfileStep extends React.PureComponent<Props, Stat
...
+                        // TODO temp for textbox demo
+                        margin: '24px',
+                        minHeight: '200px',
+                    }}
+                >
+                    <PictureSelector
+                        onSelect={this.onSelectPicture}
+                        onRemove={this.onRemovePicture}
+                        src={Utils.imageURLForUser(this.props.currentUser.id, this.props.currentUser.last_picture_update)}
+                        defaultSrc={Utils.defaultImageURLForUser(this.props.currentUser.id)}
```
nevyangelova, line 57: nit: I would extract these to a variable for readability

components/next_steps_view/next_steps_view.tsx
```diff
@@ -145,6 +146,7 @@ export default class NextStepsView extends React.PureComponent<Props, State> {
                 <Card.Body>
                     <step.component
                         id={step.id}
+                        currentUser={this.props.currentUser}
```
nevyangelova, line 149: nit: I think we can destructure props

components/picture_selector.scss
```diff
@@ -0,0 +1,54 @@
...
+    background-color: var(--center-channel-bg);
+    border: 1px solid rgba(var(--center-channel-color-rgb), 0.08);
+    box-shadow: 0px 4px 6px rgba(0, 0, 0, 0.12);
+    padding: 4px;
+    width: 28px;
+    height: 28px;
+    display: flex;
+    justify-content: center;
+
+    i {
```
nevyangelova, line 34: can we make this more specific with a class? We might end up battling it later with another icon. 🤷‍♀️

components/picture_selector.tsx
```diff
@@ -0,0 +1,127 @@
...
+    defaultSrc?: string;
+    loadingPicture?: boolean;
+    onSelect: (file: File) => void;
+    onRemove: () => void;
+};
+
+const PictureSelector: React.FC<Props> = (props: Props) => {
+    const {src, defaultSrc, loadingPicture, onSelect, onRemove} = props;
+
+    const [image, setImage] = useState<string>();
```
nevyangelova, line 23: We need to pass an initial value to `useState` so we avoid undefined on initial render.

components/picture_selector.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    useEffect(() => {
+        if (!image) {
+            if (src) {
+                setImage(src);
+            } else if (defaultSrc) {
+                setImage(defaultSrc);
+            }
+        }
+    }, [src, setImage]);
```
nevyangelova, line 71: React complains if you don’t provide dependencies for everything external. I wonder why image  isn’t in dependencies?

devinbinnie:
> @devinbinnie a few other things I noticed:
> 
> 1. 'Remove picture' button shouldn't be there until the default photo has been replaced

Fixed!

> 2. The left edge of the image needs to align with the left edge of 'Complete your profile above' (as per design below)
>    image
> 3. I'll assume the label 'Add a photo' will be in another PR

Both of these will be addressed in another PR, currently I'm just showing off that the picture selector is it's own component and works as such.

> 4. This is likely unrelated to this specific component, but there is some shifting happening on the opening and closing of the accordions and also on the transition from open to complete. Just wanted to flag it in case it hasn't been noticed. I hadn't seen in the previous PR for the accordion behavior

Ahh yes that's a consequence of me moving the `padding` for the border down into the body. Can we file a bug ticket for it? I can address it in a different PR.

components/picture_selector.scss
devinbinnie: This is already wrapped around the `PictureSelector__selectButton` class, meaning it only affects icons in that scope, and I can't see us using this class and an icon within it for something else.

components/picture_selector.tsx
devinbinnie: We want this to be undefined if there is no image present. Worse thing that happens is that the `div` using the `image` variable won't have a background image.

components/picture_selector.tsx
devinbinnie: @reflog had me add `setImage`, shouldn't that suffice?

matthewbirtch:
@devinbinnie ticket opened and assigned to you. Add whatever labels you need: https://mattermost.atlassian.net/browse/MM-27305

components/picture_selector.scss
nevyangelova: HTML elements do have more specificity and it can still clash if we are using a sibling component. I have had this issue multiple times when trying to add a new element to the RHS for example. I think we should strive to target specificity only with classes.

devinbinnie: In general, yes. But I don't think we need to be so specific with this case. There's not going to be any sibling components of `PictureSelector__selectButton`, it's pretty self-contained and I don't see any need to change to allow consumers to add their own select button.

components/picture_selector.tsx
nevyangelova: - `setImage` isn't the true state change but a method, useEffect does a comparison with the value. 
- you can definitely add it, but it would raise complexity and create new questions, because it raises the issue of comparing and wondering of the value of the function returned from the state.

devinbinnie: I'm confused, is there anything I should change here? Still new to hooks lol.

matthewbirtch:
@devinbinnie it still seems to flash the focus state.

devinbinnie:
> @devinbinnie it still seems to flash the focus state.

@matthewbirtch There, I'm blurring on click so that should stop the flash.

components/picture_selector.tsx
nevyangelova:
```suggestion
    }, [src, image]);
```

PR #5980, opened 2020-07-23
Update NOTICE.txt

amyblais:
To match with https://github.com/mattermost/mattermost-webapp/blob/release-5.26/package.json.


NOTICE.txt
```diff
@@ -392,6 +347,10 @@ THE SOFTWARE.
 
 ---
 
+## dynamic-virtualized-list
```
amyblais, line 350: Unsure if it's https://github.com/mattermost/dynamic-virtualized-list or https://github.com/dwqs/react-virtual-list

amyblais:
@grundleborg Kind reminder to review for v5.26,

NOTICE.txt
grundleborg: It's this: https://github.com/mattermost/dynamic-virtualized-list

You can find the package names in `package.json`. If they are URLs then that tells you which one it is (as in this case). If they are just names, then you can search on https://npmjs.com to find the package by it's name, and then the link to Github will be there.

amyblais: @grundleborg Thanks, added

grundleborg:
@amyblais you've got conflicts to resolve by the looks of it.

PR #5988, opened 2020-07-24
Mm 20545 migrate components/user settings/display to type script

sridhar02:
#### Summary

Migrate 'components/user_settings/display' module and associated tests to TypeScript 

#### Ticket Link

 Fixes https://github.com/mattermost/mattermost-server/issues/13696

 JIRA Ticket https://mattermost.atlassian.net/browse/MM-20545


devinbinnie:
Hey @sridhar02, thanks for the pull request!

It looks like the PR is currently failing type check, make sure to run `make check-types` before you push up your changes. It looks like some additional work will need to happen to fully convert this module over to TypeScript.

Let us know if you need any help doing so.

sridhar02:
@devinbinnie, Thanks I will ask, if there is an issue related to issue and I will try to resolve the errors as soon as possible.

devinbinnie:
> @devinbinnie, Thanks I will ask, if there is an issue related to issue and I will try to resolve the errors as soon as possible.

I'm not sure what you mean here, are you interested in other TypeScript conversion tickets/PRs?

sridhar02:
> > @devinbinnie, Thanks I will ask if there is an issue related to issue and I will try to resolve the errors as soon as possible.
> 
> I'm not sure what you mean here, are you interested in other TypeScript conversion tickets/PRs?

Sorry, my bad, I said I would ask your help if I need anything related to PR, there are a couple of tickets, I am working on , after those, I will again work on Typescript issues.

components/user_settings/display/index.ts
```diff
@@ -15,23 +16,24 @@ import {getUserCurrentTimezone} from 'mattermost-redux/utils/timezone_utils';
...
     const timezones = getTimezones(state);
     const currentUserId = getCurrentUserId(state);
     const userTimezone = getUserTimezone(state, currentUserId);
     const automaticTimezoneNotSet = userTimezone && userTimezone.useAutomaticTimezone && !userTimezone.automaticTimezone;
     const shouldAutoUpdateTimezone = !userTimezone || automaticTimezoneNotSet;
-
     const allowCustomThemes = config.AllowCustomThemes === 'true';
     const enableLinkPreviews = config.EnableLinkPreviews === 'true';
-    const defaultClientLocale = config.DefaultClientLocale;
+    const defaultClientLocale = config.DefaultClientLocale as string;
```
devinbinnie, line 32: Both this field and the one below it should already be of type `string`. How come we need the cast here ?

components/user_settings/display/index.ts
```diff
@@ -54,7 +56,13 @@ function mapStateToProps(state) {
     };
 }
 
-function mapDispatchToProps(dispatch) {
+type Actions = {
```
devinbinnie, line 59: Looks like this isn't being used, so can we remove it?

components/user_settings/display/user_settings_display.tsx
```diff
@@ -35,40 +36,92 @@ function getDisplayStateFromProps(props) {
...
+        id: string;
+        message: string;
+        moreId?: string;
+        moreMessage?: string;
+    };
+}
 
-    constructor(props) {
+type SectionProps ={
+    section: string ;
```
devinbinnie, line 50: Nitpick: extra space between the type and the semi-colon

components/user_settings/display/user_settings_display.tsx
```diff
@@ -35,40 +36,92 @@ function getDisplayStateFromProps(props) {
...
+    collapseDisplay: string;
+    linkPreviewDisplay: string;
+    handleSubmit?: () => void;
+    serverError?: string;
+    display?: string;
+}
+
+export default class UserSettingsDisplay extends React.PureComponent<Props, State> {
+    public prevSections: {
+        theme: string; // dummy value that should never match any section name
```
devinbinnie, line 116: Nitpick: This comment isn't really relevant here, it's just describing that the value for `theme` shouldn't match any section name since `theme` is always the first section.

components/user_settings/display/user_settings_display.tsx
```diff
@@ -98,7 +151,7 @@ export default class UserSettingsDisplay extends React.PureComponent {
         }
     }
 
-    componentDidUpdate(prevProps) {
+    componentDidUpdate(prevProps: any) {
```
devinbinnie, line 154: Should be `Props` instead of `any`.

components/user_settings/display/user_settings_display.tsx
sridhar02: yes, I have removed it.

components/user_settings/display/user_settings_display.tsx
sridhar02: resolved it.

components/user_settings/display/index.ts
sridhar02: yes, it can be removed and I have removed it.

components/user_settings/display/index.ts
sridhar02: there is only one line of defaultClineLocale, it is needed as I am getting an error while type checking.

components/user_settings/display/user_settings_display.tsx
sridhar02: yes resolved it.

components/user_settings/display/user_settings_display.tsx
devinbinnie: We want the comment below, not here.
The comment below describes the "magic string" of `dummySectionName`.

components/user_settings/display/user_settings_display.tsx
```diff
@@ -81,7 +136,7 @@ export default class UserSettingsDisplay extends React.PureComponent {
         }
 
         this.prevSections = {
-            theme: 'dummySectionName', // dummy value that should never match any section name
```
devinbinnie, line 84: We want to leave the comment here, not above.
The comment describes the use of `dummySectionName` as a value for this variable.

sridhar02: resolved

saturninoabril:
Hello @sridhar02 thanks for your contribution! I ran Cypress tests against this PR and it's failing to some user settings cases and looked like related to the changes here. Note that it's now failing on master. See https://mm-cypress-report.s3.amazonaws.com/709-webapp-pr-5988-327f8ca/mochawesome.html for reference. Could you please take a look at the failing tests? Let me know if you need help.

sridhar02:
@devinbinnie can you help me here, while verifying the cypress test failing , i found that unit tests are also failing in two places is user_setting_display.tsx file, in which userAutomaticTimezone is one of them, observed that it is boolean used for the input element attribute which always a boolean but in redux it defined like this, I think I need to change the redux type which will solve the error here what are your thoughts?
``` js
export type UserTimezone = {
    useAutomaticTimezone: boolean | string;
    automaticTimezone: string;
    manualTimezone: string;
};
```
https://github.com/mattermost/mattermost-redux/blob/d465f41cf007b05e95b7d2a83d1234561a87afe4/src/types/users.ts#L76

devinbinnie:
> @devinbinnie can you help me here, while verifying the cypress test failing , i found that unit tests are also failing in two places is user_setting_display.tsx file, in which userAutomaticTimezone is one of them, observed that it is boolean used for the input element attribute which always a boolean but in redux it defined like this, I think I need to change the redux type which will solve the error here what are your thoughts?
> 
> ```js
> export type UserTimezone = {
>     useAutomaticTimezone: boolean | string;
>     automaticTimezone: string;
>     manualTimezone: string;
> };
> ```
> 
> https://github.com/mattermost/mattermost-redux/blob/d465f41cf007b05e95b7d2a83d1234561a87afe4/src/types/users.ts#L76

@sridhar02 I think the `string | boolean` type is being used because our server is sending down the value as a `string` but our app only treats it as a `boolean`. There's a bit more at play here.

I think for your case, we can just force the type to be a boolean by wrapping it in a `Boolean()`.

sridhar02:
@saturninoabril @devinbinnie I have fixed all the errors and cypress tests are also rectified, can you review the PR?

PR #5990, opened 2020-07-24
MM-27167: Add undefined check

agarciamontoro:
#### Summary
Debugging the ticket linked below, we found a potentially undefined value in `DataPrefetch`. We don't know whether this will fix the bug (whose reproduction is quite inconsistent), but adding the check will not hurt.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27167


agarciamontoro:
@lindalumitchell, assigning this to you just to keep QA in the loop, but I don't think a QA review is needed here, as we just add a check on an undefined value. Furthermore, the bug is pretty inconsistent, so we'll need to merge this anyway to make sure it does not reproduce again.

PR #5995, opened 2020-07-24
[MM-26466] Close Next Steps Modal and functionality

devinbinnie:
#### Summary
This PR includes the Remove Tips and Next Steps modal that pops up when the 'x' is clicked in the sidebar element.
It also includes the logic to hide/remove the next steps view and the LHS element when this is used.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26466


matthewbirtch:
@devinbinnie I'm not seeing a sidebar element in the LHS to test this:
image

devinbinnie:
> @devinbinnie I'm not seeing a sidebar element in the LHS to test this:
> image



> @devinbinnie I'm not seeing a sidebar element in the LHS to test this:
> image

@matthewbirtch Sorry, forgot it's only on the new sidebar. Should be there now.

devinbinnie:
> Thanks @devinbinnie.
> 
> 1. The modal window should be vertically and horizontally centered on the viewport. It looks to be horizontally centered, but not vertically
> 2. The highlighted area is not positioned properly. The hand-drawn circle should be on top of the hamburger icon

I neglected to account for the team sidebar missing or not, fixed.

> 3. Looks like the highlighted area got added to the `What's New` modal as well

Bad scoping of CSS, fixed.

> 4. Depending on the state of this page, can we change the name? For the initial 3 steps, this section is called 'Getting Started' and once you've completed or skipped this, it changes to 'Tips & Next Steps'. So, the sidebar element needs to reflect that and so does the modal headline and help text next to the arrow.

Done!

> 5. Need hover states on the 'Cancel' and 'Remove' buttons - but I believe that's coming in a different PR

Yeah I can do this when I do the Hover State for Buttons ticket.

> 6. For the text next to the arrow, let's change it to "Access Tips & Next Steps any time through the Main Menu" - To be confirmed where it will appear in the menu

Done!

> 7. The `x` icon on the modal looks like it's the wrong opacity. The color is set to .56 opacity, but there must be a mother opacity on it somewhere. The same problem appears in the channel switcher modal as well, so it may be an older artifact.

Is this an issue on the Create Category modal as well? If so I should fix in another ticket for the `GenericModal` component.

matthewbirtch:
> 1. The modal window should be vertically and horizontally centered on the viewport. It looks to be horizontally centered, but not vertically

Looks like it's still not vertically centered
image

components/sidebar/sidebar_next_steps/close_next_steps_modal.scss
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.modal .CloseNextStepsModal {
+    max-width: 514px;
+
+    .GenericModal__header {
+        text-align: center;
+        padding-bottom: 8px;
+
+        h1 {
```
nevyangelova, line 28: 3/5 prefer the use of class names to avoid clashes.

devinbinnie:
> > 1. The modal window should be vertically and horizontally centered on the viewport. It looks to be horizontally centered, but not vertically
> 
> Looks like it's still not vertically centered

Sorry missed that, fixed now.

components/sidebar/sidebar_next_steps/close_next_steps_modal.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import './close_next_steps_modal.scss';
+
+type Props = {
+    screenTitle: string;
+    onConfirm: () => void;
+    onCancel: () => void;
+}
+
+export default function CloseNextStepsModal(props: Props) {
```
nevyangelova, line 20: Can we rename this to "RemoveNextStepsModal" ? Initially it was a bit confusing because it gave me the idea that it's a 'Close button' component.

components/sidebar/sidebar_next_steps/index.ts
```diff
@@ -0,0 +1,41 @@
...
+import {GlobalState} from 'types/store';
+import {Preferences} from 'utils/constants';
+
+import SidebarNextSteps from './sidebar_next_steps';
+
+function makeMapStateToProps() {
+    const getCategory = makeGetCategory();
+
+    return (state: GlobalState) => ({
+        active: state.views.nextSteps.show,
```
nevyangelova, line 23: can we rename this to 'shown' ?

components/sidebar/sidebar_next_steps/close_next_steps_modal.scss
devinbinnie: It's wrapped in a class for this modal already, so unless we have 2 `h1`s in the header, this can't clash with anything.

PR #6001, opened 2020-07-28
Cypress/E2E: Fix forgot password spec

josephbaylon:
#### Summary
- Made sure user is logged out before clearing cookies

#### Ticket Link
None - master only

Screen Shot 2020-07-28 at 11 09 49 AM

### DIFF
--- a/e2e/cypress/support/api/user.js
+++ b/e2e/cypress/support/api/user.js
@@ -35,14 +35,18 @@ Cypress.Commands.add('apiLogout', () => {
         log: false,
     });
 
-    // Ensure we clear out these specific cookies
+    // * Verify logged out
+    cy.visit('/login?extra=expired').url().should('include', '/login?extra=expired');
+
+    // # Ensure we clear out these specific cookies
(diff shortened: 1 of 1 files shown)


e2e/cypress/support/api/user.js
```diff
@@ -35,14 +35,18 @@ Cypress.Commands.add('apiLogout', () => {
         log: false,
     });
 
-    // Ensure we clear out these specific cookies
+    // * Verify logged out
+    cy.visit('/login?extra=expired').url().should('include', '/login?extra=expired');
+
```
srkgupta, line 40: Since this is a generic function and all the tests using this will be affected by this. Can we include this step specifically only in the forgot password spec?

josephbaylon: The problem is that step is in between the logout itself and deletion of cookies. Not sure how to separate that for forgot password spec. Or do you suggest I recreate a new command just for forgot password?

josephbaylon: The purpose is to verify user is logged out before cookies are deleted

PR #5659, opened 2020-06-04
MM-24361 - Updating system console save button

asaadmahmood:
#### Summary
MM-24361 - Updating system console save button

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24361

#### Screenshots
**Before:**
IMG_E3507FCCE82E-1

**After:**
IMG_CE857DC9F401-1

PR #6010, opened 2020-07-29
[MM-27168]  Set -webkit-fill-available when on iOS Safari and 100vh otherwise

fmunshi:
#### Summary
- Since webkit-touch-callout is only supported on iOS Safari (see https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout) this will make the original fix of `height: -webkit-fill-available;` apply when its needed and leave desktop chrome / other browsers unchanged.

- I dont have an iOS device to test this on so I havent actually confirmed that this works on iOS safari but I _can_ confirm that it doesnt break anything on chrome / any other desktop browsers 

- Milestone labeled as 5.26 but will also cherry-pick this to 5.25 when tested 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27168

### DIFF
--- a/sass/routes/_admin-console.scss
+++ b/sass/routes/_admin-console.scss
@@ -10,10 +10,21 @@
     height: 100%;
     padding-left: 220px;
 
+    & > div:first-child {
+        height: 100%;
+    }
+
     .wrapper--fixed {
         display: flex;
         flex-direction: column;
         height: 100vh;
+
+        // Fix for Safari on iOS MM-24361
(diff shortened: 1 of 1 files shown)


amyblais:
@asaadmahmood @calebroseland Kind note that this PR is for v5.25 dot release,

fmunshi:
tagging you @lindy65 since you reviewed the original PR here https://github.com/mattermost/mattermost-webapp/pull/5659

So I assume you have an iPad available to test with

PR #6015, opened 2020-07-30
MM-27343 Adjust sidebar channel drop IDs to account for archived channels

hmhealey:
When dropping a channel in the sidebar, the index that react-beautiful-dnd gives us only accounts for visible channels, but unfortunately that doesn't include any archived channels which may be within the category.

I added a helper function to convert the provided index to one that includes any archived channels. It does this by finding whatever channel would be above the one we've dropped and ensuring that the new index is immediately after it. The math behind it is a bit sketchy, so I left a bunch of comments which should hopefully explain what's going on.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27343


store/index.js
```diff
@@ -212,5 +212,5 @@ export default function configureStore(initialState) {
         detectNetwork: detect,
     };
 
-    return configureServiceStore({}, appReducer, offlineOptions, getAppReducer, {enableBuffer: false});
+    return configureServiceStore(initialState, appReducer, offlineOptions, getAppReducer, {enableBuffer: false});
```
hmhealey, line 215: For some reason, we never passed the preloaded state into the store before. This is only provided by test code, so it'll only affect tests using the store in this way. I'm surprised this wasn't done before, but I guess webapp tests are usually written using more mocks and less relying on real state.

amyblais:
@sudheerDev @devinbinnie Kind note to review for v5.26,

actions/views/channel_sidebar.test.ts
```diff
@@ -29,5 +34,159 @@ describe('setCategoryCollapsed', () => {
         store.dispatch(Actions.setCategoryCollapsed(category1, false));
 
         expect(isCategoryCollapsedFromStorage(getPrefix(store.getState()), store.getState().storage.storage, category1)).toBe(false);
+    }, 10000);
```
sudheerDev, line 37: What is `10000` for?

actions/views/channel_sidebar.ts
```diff
@@ -52,3 +55,48 @@ export function createCategory(teamId: string, displayName: string, channelIds?:
...
+        // The channel is being placed first, so there's nothing above that could affect the index
+        return 0;
+    }
+
+    const category = getCategory(state, categoryId);
+    const filteredChannels = makeGetChannelsForCategory()(state, category);
+
+    if (category.channel_ids.length === filteredChannels.length) {
+        // There are no archived channels in the category, so the targetIndex from react-beautiful-dnd will be correct
+        return targetIndex;
```
sudheerDev, line 80: Do we need to consider channel moving downwards with archive channels?

store/index.js
sudheerDev: I guess rehydration somehow fixes this for us?

devinbinnie: What does this solve? Just curious.

actions/views/channel_sidebar.test.ts
hmhealey: Good catch. I had some issues with the tests before I figured out the store wasn't passing along the initial state, and I thought waiting for a bit might fix it. I'll remove this.

store/index.js
hmhealey: This doesn't have anything to do with rehydration or saving redux state in the webapp. `initialState` is always an empty object when you're running the app itself, but it can be passed down with a set value for use in unit tests. A lot of the tests I've written in mattermost-redux rely on being able to set up an initial state in the store

actions/views/channel_sidebar.ts
hmhealey: Yeah, all the logic below this is about handling archived channels. If there are no archived channels, then react-beautiful-dnd handles everything itself, so targetIndex will already be correct.

amyblais:
@hmhealey Curious if anything is blocking getting this merged for RC1.

hmhealey:
Nope, it's good to merge

PR #6018, opened 2020-07-30
[MM-27406] Dont show archived channels filter if view archived config is false

fmunshi:
#### Summary
- Dont show archived channels filter if view archived channels config is false 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27406

#### Screenshots
Screen Shot 2020-07-30 at 2 24 01 PM


fmunshi:
@thefactremains Not sure if we want this config setting to apply for sys admins too... if you think it should not then we can close this PR

fmunshi:
After discussing with @wiersgallak I have confirmed that this config setting should _not_ affect system admins so I will be closing this and opening a ticket to modify the API to allow sysadmins to always be able to `include_deleted` channels from the API

PR #6019, opened 2020-07-30
MM-27358 Change how we hide unread badges on hover

hmhealey:
There's a bug in the current web app where the notification badge on the tabs in the desktop app changes its value when you hover over a channel. This is happening because we hide the badge and when that happens [the desktop app stops counting it](https://github.com/mattermost/desktop/blob/c2102bb25734a8c33fec635e0445873a23fd6559/src/browser/webview/mattermost.js#L146). I don't remember why we check this (it might be because we had multiple instances of the sidebar mounted at some point for some reason), but I've fixed it by hacking in a version of `display: none` that keeps the badge's height while still making it invisible

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27358

### DIFF
--- a/sass/layout/_sidebar-left.scss
+++ b/sass/layout/_sidebar-left.scss
@@ -924,7 +924,13 @@
         }
 
         .badge {
-            display: none;
+            // Hide the badge by giving it 0 width instead of using `display: none` since the desktop app only counts
+            // badges with a non-zero offsetHeight when counting notifications
+            margin: 0px;
+            min-width: 0px;
+            overflow: hidden;
+            padding: 0px;
+            width: 0px;
         }
     }
     .SidebarChannel.active .SidebarLink {
--- a/sass/layout/_sidebar-left.scss
+++ b/sass/layout/_sidebar-left.scss
@@ -924,13 +924,10 @@
         }
 
         .badge {
-            // Hide the badge by giving it 0 width instead of using `display: none` since the desktop app only counts
+            // Hide the badge by hiding it instead of using `display: none` since the desktop app only counts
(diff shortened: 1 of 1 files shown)


amyblais:
@deanwhillier Kind note to prioritize reviews for v5.26,

sass/layout/_sidebar-left.scss
```diff
@@ -924,7 +924,13 @@
...
 
         .badge {
-            display: none;
+            // Hide the badge by giving it 0 width instead of using `display: none` since the desktop app only counts
+            // badges with a non-zero offsetHeight when counting notifications
+            margin: 0px;
+            min-width: 0px;
+            overflow: hidden;
+            padding: 0px;
+            width: 0px;
```
deanwhillier, line 933: Nice! Interesting implementation on the Desktop app, I wonder why it is looking to ensure the badge is visible?

You can also achieve this by using:
```suggestion
            // Hide the badge by hiding it instead of using `display: none` since the desktop app only counts
            // badges with a non-zero offsetHeight when counting notifications
            position: absolute;
            visibility: hidden;
```

hmhealey: I did a bit more digging, and it's because we had two sets of badges in the app, one in the regular sidebar and another in a mobile sidebar, so we had to make sure only to count visible ones. I'm pretty sure the LHS isn't duplicated between the two any more, so that shouldn't be necessary, but that code's been in place [for 4 and a half years](https://github.com/mattermost/desktop/commit/929e5e6f88f513badddd07ef454be30cd6e1bcbb) at this point.

As for the CSS, I definitely like the simpler CSS there.

PR #6020, opened 2020-07-30
[MM-26473] Tips and Next Steps screen

devinbinnie:
#### Summary
Tips and Next Steps screen that is seen at the end of the onboarding flow.
Includes the transition screen that slides in when you complete the flow.

Some UX notes:
- Other button states (ie. hover/active/focus) will come in another PR
- The transition timings might need some tuning, let me know if they need to be tweaked
- We don't have a download flow for Linux, I think we should.
- 'Take a tour' does nothing currently as we don't have the new tour developed yet.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26473


components/card/card_body.tsx
```diff
@@ -20,6 +20,12 @@ export default class CardBody extends React.PureComponent<Props> {
     }
 
     componentDidMount() {
+        window.addEventListener('resize', this.setInitialHeight);
```
reflog, line 23: seems like we need a removeEventListener on unmount

components/next_steps_view/next_steps_view.tsx
```diff
@@ -154,11 +153,31 @@ export default class NextStepsView extends React.PureComponent<Props, State> {
...
+    renderTransitionScreen = () => {
         return (
-            <div>
-                {'Placeholder for Final Screen'}
+            <div
+                className={classNames('NextStepsView__viewWrapper NextStepsView__transitionView', {
+                    transitioning: this.state.showTransitionScreen,
+                    completed: this.state.showTransitionScreen && this.state.showFinalScreen,
+                })}
+                onTransitionEnd={this.setTimerToFinalScreen}
```
nevyangelova, line 163: is this reliable? I remember reading that `onTransitionEnd` can sometimes ignore the transitions of the child elements.

components/next_steps_view/next_steps_tips.tsx
```diff
@@ -0,0 +1,246 @@
...
+                                    defaultMessage='Take a tour'
+                                />
+                            </h3>
+                            <FormattedMessage
+                                id='next_steps_view.tips.takeATour.text'
+                                defaultMessage='Let us show you around with a guided tour of the interface.'
+                            />
+                            <button
+                                className='NextStepsView__button confirm'
+                                onClick={() => {}}
```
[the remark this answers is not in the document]
devinbinnie: Yeah, see this comment in the description:
`'Take a tour' does nothing currently as we don't have the new tour developed yet.`

components/next_steps_view/next_steps_view.tsx
devinbinnie: Yeah it should be fine, I only care about transition on that `div` anyways for that callback.

devinbinnie:
@matthewbirtch 

> **Transition from Getting Started to Next Steps**
> 
> * Overall we can speed up this transition to 500ms and apply the 'ease-in-out' function

Done!

> * Have a look at [this prototype](https://www.figma.com/proto/uv9A6dZsekQHHhOhYC7yc2/Cloud-UX?node-id=1767%3A64252&viewport=814%2C490%2C0.11302600800991058&scaling=min-zoom) for reference.
>   
>   * Notice that the computer illustration transitions to the success visual.
>   * Let me know if this is something you'd like to do together over zoom. Also, let me know if transition that illustration makes it too difficult.
>   * There is more I'd like to do with this animation, but I don't want to push too much either, so let me know what is doable.

Yeah this is probably going to be a miss. We can have a Zoom chat but this isn't what I was expecting. My bad if that's something I should have anticipated.

> **App download callout:**
> 
> * The devices illustration looks like it is getting stretched at some widths

Fixed!

> * Download Mattermost for Mac button:
>   
>   * I don't have a windows machine, but I assume this will appropriately switch to 'Download Mattermost for Windows'?

Correct, it changes based on OS. Though we don't have one for Linux, which is partially due to that we don't have a `UserAgent` method to detect that. I wonder if that's something easy to do...

>   * The background on this button should be transparent - not white

Done! Note, this is going to look funny on some resolutions as you might see some of the background shift around behind the button.

> 
> **Card responsive layout:**
> 
> * I was somehow able to get to a state where the 3rd card didn't have the same width as the others (around 1082 width)

Okay, I've gone back to having the cards be a fixed width of 310px on the desktop view. Otherwise flexbox won't cooperate.

> * The app download callout is cutoff or hidden depending on my borwser height in the layout above and I can't scroll down to see it all

This should be working fine, though in mobile view I've noticed it's cutting off incorrectly. I've added some padding to the bottom of the mobile view which should fix that.

> * As soon as the layout adjusts to this layout, mabye we collapse the extra space between the titles and the cards and the extra space between the cards and the app download callout

Done!

> * in smaller mobile size windows, the app download callout is cut off and I can't scroll down to view it all

See above.

> * when they start stacking, the space between cards increases. Can we keep it 24px? (and 16px on mobile widths)

Done!

> * Also, when the cards stack, can we change the height from a fixed width to auto? this should tighten up the space

Done!

> * Layout for the app download callout feels a bit broken between 768 and 905.
>   
>   * Title looks positioned down to far
>   * 'See all the apps' button needs to align left with the button above, add 16px space between
>   * Again, the illustration looks stretched vertically

Fixed!

> 
> **'To configure your workspace, continue on a desktop computer'**
> 
> * Are we able to do this view only if it is actually a mobile device and check the user agent?
> * I was able to get this state by first starting in a mobile sized viewport and then changing to larger screen

When you refresh the page in Chrome in 'Responsive' mode, it spoofs a mobile user agent screen. This is normal.

> 
> **Changing browser size issues:**
> 
> * When I try to resize my browser, sometimes I get elements from the previous screen peaking in

Fixed!

stevemudie:
Overall it looks great.

I did however notice the same 3rd card too wide anomaly @matthewbirtch noticed. All you need to do to see it is drag the window horizontally to make it narrow and it will change as you go.



I also noticed something completely unexpected when I used my window manager ([Rectangle](https://rectangleapp.com/)) to jump it back to full size it flashes the transition image that displayed on the way to this point.

devinbinnie:
> Overall it looks great.
> 
> I did however notice the same 3rd card too wide anomaly @matthewbirtch noticed. All you need to do to see it is drag the window horizontally to make it narrow and it will change as you go.
> 
> 
> 
> I also noticed something completely unexpected when I used my window manager ([Rectangle](https://rectangleapp.com/)) to jump it back to full size it flashes the transition image that displayed on the way to this point.
> 
> 

@stevemudie This should all be fixed now as per above. Have another look.

PR #6021, opened 2020-07-30
MM-27182 - Fix dynamic slash command autocomplete options

cpoile:
#### Summary
- The `CommandProvider` is not recreated for each channel or RHS, so the `channel_id` needs to dynamically determined.
- When used in the RHS we cannot use the current channel id (because the center could be viewing a different channel), so we have to use the channel associated with the post in the RHS. 
- Unfortunately I don't know of a way to be certain that a textbox is in the RHS vs the main channel, so I'm using a flag to indicate that this is a RHS textbox.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27182


components/suggestion/command_provider.jsx
```diff
@@ -130,14 +131,20 @@ export default class CommandProvider extends Provider {
 
     handleWebapp(pretext, resultCallback) {
         const command = pretext.toLowerCase();
-        const teamId = getCurrentTeamId(store.getState());
+
+        const selectedPost = getSelectedPost(store.getState());
+        let rootId;
+        if (this.rootId) {
+            rootId = selectedPost.root_id ? selectedPost.root_id : selectedPost.id;
+        }
```
marianunez, line 139: Do we need the `this.rootId` at all? When you are not in the RHS does `getSelectedPost` return `null`? Wondering if we can just use that as to determine if it's in the RHS

cpoile: I think we do -- when you are in a different center channel, but the RHS is open, `getSelectedPost` will return the post selected in the RHS.

marianunez: Right, how about when it's closed?

cpoile: Then it returns the `type: "system_fake_parent_deleted"` post.

marianunez: How about `getIsRhsOpen`?

cpoile: But we still wouldn't know if we're typing in the center channel textbox or the RHS checkbox.

marianunez: Sorry, I guess I'm not following how `rootId` guarantees they have not switched between the RHS and center channel textbox if the `CommandProvider` is not recreated

cpoile: It's not the same CommandProvider -- one is created for each textbox.

Do you think I should change that root_id to something like `isInRHS`? I just didn't like introducing booleans for information we already have. But I can see why that might be better.

marianunez: Yes, I think using the `this.rootId` as a flag and then using another value for it when you pass it as an args is confusing. Renaming that would help clear this a bit

cpoile: Great, done!

amyblais:
@cpoile Kind note to cherry-pick this,

PR #6026, opened 2020-07-30
MM-16998 Add eslint-plugin-jquery to block explicit usage of jquery

hmhealey:
Taking inspiration from https://github.blog/2018-09-06-removing-jquery-from-github-frontend/, this adds eslint-plugin-jQuery to prevent people from adding new references to jQuery to the code base. Any new usage of jQuery will cause an ESLint error, and any old usage of it has been marked with a comment so people know using it is wrong (morally, of course).

This'll be followed by one or more Help Wanted tickets to actually update this code to not use jQuery.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-16998

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/6025

### DIFF
--- a/components/new_channel_modal/new_channel_modal.jsx
+++ b/components/new_channel_modal/new_channel_modal.jsx
@@ -1,7 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import $ from 'jquery';
 import PropTypes from 'prop-types';
 import React from 'react';
 import {Modal} from 'react-bootstrap';
@@ -12,7 +11,6 @@ import LockIcon from 'components/widgets/icons/lock_icon';
 import LocalizedInput from 'components/localized_input/localized_input';
 import Constants from 'utils/constants.jsx';
 import {getShortenedURL} from 'utils/url';
-import * as UserAgent from 'utils/user_agent';
 import * as Utils from 'utils/utils.jsx';
 import {t} from 'utils/i18n.jsx';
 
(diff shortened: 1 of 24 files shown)


hmhealey:
Skipping QA review since this has no effect on the compiled code.

jespino:
I love it

components/new_channel_modal/new_channel_modal.jsx
```diff
@@ -110,13 +108,6 @@ export default class NewChannelModal extends React.PureComponent {
         this.displayNameInput = React.createRef();
     }
 
-    componentDidMount() {
-        // ???
-        if (UserAgent.isInternetExplorer() || UserAgent.isEdge()) {
-            $('body').addClass('browser--ie');
```
hmhealey, line 116: No idea why the NewChannelModal was responsible for this, but this logic is duplicated by the ChannelController which is a much better place for it to go

lieut-data: Should we pull this into a separate PR to keep the scope focussed on things for which we definitely don't need to have QA testing.

components/setting_upload.jsx
```diff
@@ -22,10 +21,12 @@ export default class SettingsUpload extends React.PureComponent {
 
     doFileSelect = (e) => {
         e.preventDefault();
-        var filename = $(e.target).val();
+
+        let filename = e.target.value;
```
lieut-data, line 25: Same comment as above (though I understand this should be the same).

components/new_channel_modal/new_channel_modal.jsx
hmhealey: I don't think it should be necessary for small  bits of cruft like this that shouldn't have any effect on the code. I'll make sure that the other jQuery removal PRs will include QA review when they actually have something to test though.

PR #6028, opened 2020-07-31
Cypress/E2E: Fixing search in system console for team and channel assign roles

prapti:
#### Summary
Adding fix for search in system console for team_and_channel_assign_roles_spec. There have been some changes in the scoping of the "Search" placeholder text. Ideally, I'd like to make finding the placeholder text simpler by adding a custom command since each search command needs at least 2 lines with the current setting. But need to find out all the places where the refactoring will be required. Thus checking in these changes.

(There are similar failures in `archived_channels_spec` but there are more failures than just the search component, hence taking a look at it separetely)

#### Screenshots
Screen Shot 2020-07-30 at 4 48 17 PM


josephbaylon:
@prapti There might be lint issues

PR #6031, opened 2020-07-31
Cypress/E2E: Fix marketplace spec

josephbaylon:
#### Summary
- Cypress failure is due to tab element not visible because of async loading of plugins; added scrollIntoView to marketplace tabs and plugin search elements
- Cleaned up comments
- Added missing verifications
- Disable AutomaticPrepackagedPlugins when EnableRemoteMarketplace is disabled to make sure no other plugins are loaded
- Potential cypress failure on `autofocus on search plugin input box`; will monitor if that fails on build server

#### Ticket Link
None - master only

### DIFF
--- a/e2e/cypress/integration/plugins/marketplace/marketplace_spec.js
+++ b/e2e/cypress/integration/plugins/marketplace/marketplace_spec.js
@@ -386,11 +386,13 @@ describe('Plugin Marketplace', () => {
         beforeEach(() => {
             // # Enable Plugin Marketplace
             // # Disable Plugin Remote Marketplace
+            // # Disable Automatic Prepackaged Plugins to make sure no plugins are loaded
             cy.apiUpdateConfig({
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/plugins/marketplace/marketplace_spec.js
```diff
@@ -393,31 +406,44 @@ describe('Plugin Marketplace', () => {
         });
 
         it('not display any plugins and no error bar', () => {
-            // * no plugins should be visible
+            // * Verify no plugins should be visible
             cy.get('#marketplaceTabs-pane-allPlugins').findByText('There are no plugins available at this time.');
```
josephbaylon, line 410: This case may still fail on cypress as I believe the server has prepackaged plugins loaded. So even if remote marketplace is disabled, prepackaged plugins appear

PR #6037, opened 2020-07-31
Cypress/E2E: Potential workaround to fix autocomplete shown each channel spec

josephbaylon:
#### Summary
- I was able to repro the issue intermittently. However, it only fails when running through cypress. I wasn't able to repro it manually.
- This is just a potential workaround to cypress issue - make sure elements are visible before interaction and switch to an intermediate channel before final channel for verification; after I did this, I couldn't repro the issue anymore

#### Ticket Link
None -  both on master and release-5.26

Screen Shot 2020-07-31 at 9 34 58 AM


josephbaylon:
@prapti I think the original issues was a combination of interacting with elements before fully visible and verify suggestion list before page is fully loaded. I'm hoping this fixes it on cypress server. If not, we may have to move it to unstable.

PR #6050, opened 2020-08-03
[MM-T2345] /me on RHS

clarmso:
#### Summary

The `/me` command should work on RHS (reply thread). The reply should be shown on the main channel (in a lower opacity).

#### Ticket Link

https://automation-test-cases.vercel.app/test/MM-T2345

#### Related Pull Requests

- Has server changes N/A
- Has redux changes N/A
- Has mobile changes N/A

#### Screenshots

Screen Shot 2020-08-03 at 1 24 35 AM

### DIFF
--- a/e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
+++ b/e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
@@ -11,6 +11,7 @@
 // Group: @integrations
 
 import * as TIMEOUTS from '../../../fixtures/timeouts';
+import * as MESSAGES from '../../../fixtures/messages';
 
 describe('I18456 Built-in slash commands: common', () => {
     let user1;
@@ -71,6 +72,21 @@ describe('I18456 Built-in slash commands: common', () => {
             cy.get(`#post_${postId}`).find('.user-popover').should('have.text', user1.username);
             cy.get(`#postMessageText_${postId}`).should('have.text', 'test ¯\\_(ツ)_/¯');
         });
+    });
+
+    it('MM-T2345 /me on RHS', () => {
+        loginAndVisitDefaultChannel(user1, testChannelUrl);
+        cy.postMessage(MESSAGES.MEDIUM);
+
+        // # Open RHS (reply thread)
+        cy.clickPostCommentIcon();
+
+        // # type /me test
+        cy.get('#reply_textbox').type('/me test');
+        cy.get('#addCommentButton').click();
+
+        // * Properly formatted with lower opacity
+        cy.get('div[id="postListContent"]').contains('p', 'test').should('be.visible');
     });
 });
 
--- a/e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
+++ b/e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
@@ -87,14 +87,17 @@ describe('I18456 Built-in slash commands: common', () => {
         cy.uiWaitUntilMessagePostedIncludes('test');
 
         cy.getLastPostId().then((postId) => {
-            // * Message 'test' from the current user
-            cy.apiGetPost(postId).then((body) => {
-                cy.wrap(body.user_id).should('equal', user1.id);
-                cy.wrap(body.props.message).should('equal', 'test');
+            // * Verify RHS message is from current user and properly formatted with lower opacity
+            cy.get(`#rhsPost_${postId}`).should('have.class', 'current--user').within(() => {
+                cy.get('button').should('have.text', user1.username);
+                cy.get('p').should('have.text', 'test').and('have.css', 'color', 'rgba(61, 60, 64, 0.6)');
             });
 
-            // * Message properly formatted with lower opacity
-            cy.get(`#postMessageText_${postId}`).contains('p', 'test').should('be.visible').and('have.css', 'color', 'rgba(61, 60, 64, 0.6)');
+            // * Verify message on the main channel is from current user and properly formatted with lower opacity
+            cy.get(`#post_${postId}`).should('have.class', 'current--user').within(() => {
+                cy.get('button').should('have.text', user1.username);
+                cy.get('p').should('have.text', 'test').and('have.css', 'color', 'rgba(61, 60, 64, 0.6)');
+            });
         });
     });
 });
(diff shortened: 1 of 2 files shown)


e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
```diff
@@ -72,6 +73,21 @@ describe('I18456 Built-in slash commands: common', () => {
...
+
+        // # Open RHS (reply thread)
+        cy.clickPostCommentIcon();
+
+        // # type /me test
+        cy.get('#reply_textbox').type('/me test');
+        cy.get('#addCommentButton').click();
+
+        // * Properly formatted with lower opacity
+        cy.get('div[id="postListContent"]').contains('p', 'test').should('be.visible');
```
clarmso, line 89: Could someone please give me a suggestion on how to test for opacity?

saturninoabril: The way we're setting up opacity is not via CSS `opacity` but with `color`, so one option is,
```javascript
.and('have.css', 'color', 'rgba(61, 60, 64, 0.6)');
```
(Brittle but should work before we go full into visual regression tests.)

saturninoabril, line 89: It should also verify that the last post is from the current user.

clarmso: I've added `cy.apiGetPost()` and use it in the test to verify the author of the post.

clarmso: ahhh...not ideal but ok for now.

e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
```diff
@@ -72,6 +73,21 @@ describe('I18456 Built-in slash commands: common', () => {
...
+
+    it('MM-T2345 /me on RHS', () => {
+        loginAndVisitDefaultChannel(user1, testChannelUrl);
+        cy.postMessage(MESSAGES.MEDIUM);
+
+        // # Open RHS (reply thread)
+        cy.clickPostCommentIcon();
+
+        // # type /me test
+        cy.get('#reply_textbox').type('/me test');
```
josephbaylon, line 85: Make sure reply textbox is visible before typing

```suggestion
        cy.get('#reply_textbox').should('be.visible').type('/me test');
```

e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
```diff
@@ -72,6 +73,21 @@ describe('I18456 Built-in slash commands: common', () => {
...
+        cy.get('#addCommentButton').click();
```
josephbaylon, line 86: Make sure comment button is visible before clicking

```suggestion
        cy.get('#addCommentButton').should('be.visible').click();
```

e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
saturninoabril: Okay, I think it's fine though I'd prefer verifying it via UI.

e2e/cypress/support/api_commands.js
```diff
@@ -234,6 +234,22 @@ Cypress.Commands.add('apiUnpinPosts', (postId) => {
     });
 });
 
+/**
+ * Get a post
+ * This API assume that the user is logged in and has cookie to access
+ * @param {String} post_id - The post id of the post
+ */
+Cypress.Commands.add('apiGetPost', (postId) => {
```
saturninoabril, line 242: 💯 Nice!
We're currently organizing API commands and will eventually move everything to `/support/api` folder. With that, please move this into a new file called `/support/api/post.js` and add type definition into new `/support/api/post.d.ts`. See https://github.com/mattermost/mattermost-webapp/blob/master/e2e/cypress/support/api/team.d.ts for reference.
 (We have it on item 2 in our [docs](https://developers.mattermost.com/contribute/webapp/end-to-end-tests/#writing-specs) 😄 )

e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
```diff
@@ -72,6 +73,30 @@ describe('I18456 Built-in slash commands: common', () => {
...
+        cy.uiWaitUntilMessagePostedIncludes('test');
+
+        cy.getLastPostId().then((postId) => {
```
josephbaylon, line 89: I think we need to verify the current user from UI and not from API. Since postId is the same for RHS and main channel, we can infer that the user for RHS post is the same user for main channel. This block can be re-written to,

```
            // * Verify RHS message is from current user and properly formatted with lower opacity
            cy.get(`#rhsPost_${postId}`).should('have.class', 'current--user').within(() => {
                cy.get('button').should('have.text', user1.username);
                cy.get('p').should('have.text', 'test').and('have.css', 'color', 'rgba(61, 60, 64, 0.6)');
            });

            // * Verify main channel message is properly formatted with lower opacity
            cy.get(`#postMessageText_${postId}`).contains('p', 'test').should('be.visible').and('have.css', 'color', 'rgba(61, 60, 64, 0.6)');
        });
```

e2e/cypress/support/api_commands.js
josephbaylon, line 242: Not sure this is needed if we verify the current user from UI.

e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
clarmso: Cypress should ensure the button isn't hidden and is enabled before clicking. 🤔 💭 
https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Actionability

Could someone from Mattermost confirm?

e2e/cypress/support/api_commands.js
clarmso: Ok. I've adopted your code snippet and verify the user on the UI instead of using the API.

clarmso: I've removed `apiGetPost` for now since we've decided to verify the user on the UI. I'll keep the location of the API command in mind. Now I know where to look for them if I need to use them. 😄

e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
saturninoabril: Based on my experience with Cypress, visibility check helps lessen flaky tests, especially with the kind of application we have where the UI relies heavily on async store/state, local storage, REST API calls, etc. It's an implicit wait to make sure that the UI is in ready state (not only a particular element). In this case, I'll add visibility check to `#reply_textbox` before typing since RHS is not open on initial state.
Anyway, I'm good as is (also seeing Joseph already approved).
By the way, in case you're wondering, Joseph is also a staff of Mattermost 😄

PR #6052, opened 2020-08-03
MM-20664 Add client side ping pong

sudheerDev:
#### Summary

  * Emit ping event after every 10secs
  * clear timer to not emit ping events when socket is emitting messages(Means socket is alive and does not require pings)
  * close conn if the server does not respond with a pong in due time(This is needed to close stale connections on weird network errors)

This follows the recommended approach or sockets https://github.com/websockets/ws#how-to-detect-and-close-broken-connections

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20664


client/websocket_client.jsx
```diff
@@ -201,4 +211,22 @@ export default class WebSocketClient {
         data.user_ids = userIds;
         this.sendMessage('get_statuses_by_ids', data, callback);
     }
+
+    createPingEvent() {
+        this.pingTimer = setInterval(() => {
+            this.sendMessage('ping');
```
Willyfrog, line 217: there is the option [here](https://github.com/mattermost/mattermost-webapp/blob/f4f051820cc89e66654803cbb36980a8be849a8c/client/websocket_client.jsx#L183) of not sending the message. There are also some exceptions (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/send)) we are not managing.

if any of the above cases happen, we shouldn't wait for pong and call close (or try to restore) inmediately.

client/websocket_client.test.jsx
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import WebSocketClient from './websocket_client.jsx';
+const fakeURL = 'ws://localhostt:9001';
```
Willyfrog, line 5: minor nitpick: uppercase for constants
```suggestion
const FAKE_URL = 'ws://localhostt:9001';
```

client/websocket_client.jsx
```diff
@@ -119,6 +126,9 @@ export default class WebSocketClient {
                 this.eventSequence = msg.seq + 1;
                 this.eventCallback(msg);
             }
+
+            this.clearPingPong();
+            this.createPingEvent();
```
Willyfrog, line 131: should we repeat this when we send a message? (line 182, sorry but github doesn't allow me to add the comment there)

client/websocket_client.jsx
sudheerDev: Sorry, I didn't get it. This is for closing the socket in the event when socket. does not close by itself. Exceptions are for send on the socket and not related to close. 

> there is the option here of not sending the message.

If socket conn is closed i.e this.conn === null;  then we don't emit ping messages so it does not matter all that much

client/websocket_client.jsx
sudheerDev: Good question, I didn't think this before but I would say No, because sending doesn't confirm that socket is open

stylianosrigas:

jgilliam17 (approved):
Thank you @sudheerDev 
Tested, looks good to merge.
- Verified 10 sec. ping events and disconnect banner after loss of connection. Manual test added.

PR #6062, opened 2020-08-04
MM-T1785 Toast - When marking post as unread

openmohan:
#### Summary

MM-T1785 Toast - When marking post as unread

#### Ticket Link

[Usecase](https://automation-test-cases.vercel.app/test/MM-T1785)


e2e/cypress/integration/toast/toast_spec.js
```diff
@@ -263,6 +263,45 @@ describe('toasts', () => {
             cy.get('div.toast').should('be.visible').contains('Viewing message history');
         });
     });
+
+    it('MM-T1785 Toast - When marking post as unread', () => {
+        visitTownSquareAndWaitForPageToLoad();
+
+        // # Add 40 posts to create enough space from bottom for making channel scrollable
+        for (let index = 0; index < 40; index++) {
```
saturninoabril, line 271: I'm good as is but can we reduce this to maybe 30? My only concern is about the longer duration it's taking to post 10 more posts.

openmohan: Yes, I have checked it and we are getting scrollable channel in 30 posts itself.

openmohan:
> Thank you for another great submission @openmohan!
> I am approving this PR but I see one request from @saturninoabril. Please let me know if you plan on changing the number of posts from 40 to 30. In that case, I'll wait for your change to get added before merging.
> Also, if the conflicts could be resolved, that'd be awesome. Thank you so much!

@prapti Thanks for reviewing, I have added @saturninoabril 's suggestions and conflicts are resolved now.

saturninoabril:
Thanks @openmohan!

PR #6065, opened 2020-08-04
MM-27489: fix user profile timestamp

calebroseland:
#### Summary
#5390 replaced `local_date_time` in User Profile Popover with `timestamp`, but `user.timezone` (the somewhat deceptively-named object) was mistakenly directed to the `timeZone` (string) prop, instead of the correct `userTimezone` prop.

- This PR fixes that,
```diff
- timeZone={this.props.user.timezone}
+ userTimezone={this.props.user.timezone}
```
- adds a timezone name, 
```diff
+ useTime={{hour: 'numeric', minute: 'numeric', timeZoneName: 'short'}}
```

- simplifies `semantic_time`, and
- increases test coverage.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27489

#### Screenshots
Screen Shot 2020-08-04 at 13 07 11


components/profile_popover/profile_popover.jsx
```diff
@@ -461,7 +461,6 @@ class ProfilePopover extends React.PureComponent {
                         >
                             <ToggleModalButtonRedux
                                 accessibilityLabel={addToChannelMessage}
-                                ref='addUserToChannelModalButton'
```
calebroseland, line 464: Was causing a runtime console error/warning, is not referenced anywhere, presumed safe to remove.

PR #6068, opened 2020-08-04
[MM-26972] Skip pagination to allow loading all groups on needs team

fmunshi:
#### Summary
- Since the pagination params were not being passed before it would just default to page 0 and per_page 60 on the server side, however the api endpoint was written to ignore pagination if the `per_page` value equalled `0`, this fixes that.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26972

#### Related Pull Requests
- https://github.com/mattermost/mattermost-redux/pull/1212

### DIFF
--- a/components/needs_team/needs_team.tsx
+++ b/components/needs_team/needs_team.tsx
@@ -248,7 +248,7 @@ export default class NeedsTeam extends React.PureComponent<Props, State> {
             if (team.group_constrained) {
                 this.props.actions.getAllGroupsAssociatedToTeam(team.id, true);
             } else {
-                this.props.actions.getGroups(true);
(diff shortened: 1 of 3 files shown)


srkgupta:
Hi @fmunshi 
Tested the issue on the webapp and the issue is fixed. However the issue is still seen on the mobile app. Are you planning to submit a separate PR for the mobile app or is the redux PR also supposed to fix it for mobile apps? Please let me know. Will approve the PR accordingly.

fmunshi:
Hi @srkgupta I have opened a separate PR for mobile https://github.com/mattermost/mattermost-mobile/pull/4644 but it should not block the webapp fix since mobile no longer depends on redux

PR #6070, opened 2020-08-05
[MM-27490] Fix issue with search and reloading teams

fmunshi:
#### Summary
- Fixes an issue where infinite scroll would not load more users if you loaded all the teams into the state, performed a search and then cleared the search once more.
- Also fixes an issue where clearing the search would sometimes result in the first letter reappearing due to the debounced function not being called with an empty string if search term was empty
- Also fixes `searchTeams` definition to match redux master, with searchOpts as second parameter instead of page and per_page (this change will have to be ignored when cherry-picked for 5.27)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27490


amyblais:
Perhaps Ashish or Scott can merge this while Farhan is off 🙏

PR #6072, opened 2020-08-05
MM-23409 Integrate Cypress and TM4J

saturninoabril:
#### Summary
Integrate Cypress and TM4J. The following environment variables needs to set in order to save test cases into TM4J
- TM4J_ENABLE=true|false
- TM4J_API_KEY=[api_key]
- JIRA_PROJECT_KEY=[project_key], e.g. "MM",
- TM4J_FOLDER_ID=[folder_id], e.g. 847997

See [example test run/cycle](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testPlayer/MM-R421)

#### Ticket Link
Jira ticket - https://mattermost.atlassian.net/browse/MM-23409

### DIFF
--- a/e2e/utils/test_cases.js
+++ b/e2e/utils/test_cases.js
@@ -34,22 +34,13 @@ function getStepStateResult(steps = []) {
 function getStepStateSummary(steps = []) {
     const result = getStepStateResult(steps);
 
-    let summary = '';
-    Object.entries(result).forEach(([key, value], index) => {
-        if (index) {
-            summary += ',';
-        }
-        summary += `${value} ${key}`;
-    });
-
-    return summary;
+    return Object.entries(result).map(([key, value]) => `${value} ${key}`).join(',');
 }
 
 function getTM4JTestCases(report) {
-    const allTests = getAllTests(report.results);
+    const re = /^(MM-T)\w+/g;
 
-    const re = /(MM-T)\w+/g;
-    return allTests.
+    return getAllTests(report.results).
         filter((item) => re.test(item.title)).
         map((item) => {
             return {
@@ -123,8 +114,8 @@ async function createTestExecutions(report, testCycle) {
     const startTime = startDate.getTime();
 
     const promises = [];
-    Object.entries(testCases).forEach(([key, value]) => {
-        const testScriptResults = value.
+    Object.entries(testCases).forEach(([key, steps]) => {
+        const testScriptResults = steps.
             sort((a, b) => a.title.localeCompare(b.title)).
             map((item) => {
                 return {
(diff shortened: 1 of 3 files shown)


e2e/utils/test_cases.js
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        return acc;
+    }, {});
+}
+
+function getStepStateSummary(steps = []) {
+    const result = getStepStateResult(steps);
+
+    let summary = '';
```
josephbaylon, line 37: I haven't tried but I think this can be simplified to something similar
```
    return Object.entries(result).map(([key, value]) => `${value} ${key}`).join(',');
```

e2e/utils/test_cases.js
```diff
@@ -0,0 +1,171 @@
...
+                duration: item.duration,
+                incrementalDuration: item.incrementalDuration,
+                state: item.state,
+                pass: item.pass,
+                fail: item.fail,
+                pending: item.pending,
+            };
+        }).
+        reduce((acc, item) => {
+            const key = item.title.split(' ')[0].split('_')[0];
```
josephbaylon, line 66: Quick question: I'm assuming these are the ticket numbers. Do they have underscore in them?

e2e/utils/test_cases.js
```diff
@@ -0,0 +1,171 @@
...
+        BROWSER,
+        JIRA_PROJECT_KEY,
+    } = process.env;
+
+    const testCases = getTM4JTestCases(report);
+    const startDate = new Date(report.stats.start);
+    const startTime = startDate.getTime();
+
+    const promises = [];
+    Object.entries(testCases).forEach(([key, value]) => {
```
josephbaylon, line 126: Just personal preference, may be rename `value` to `steps` for readability. I had to dig through to understand what `value` was supposed to be.

e2e/utils/test_cases.js
saturninoabril: Correct. I'll add in the description and to follow on E2E docs but for those test cases that have multiple test steps, the foreseen convention would be `[tm4j_key]_[step_number]`, e.g. `MM-T1_4`.

e2e/utils/test_cases.js
```diff
@@ -0,0 +1,171 @@
...
+        summary += `${value} ${key}`;
+    });
+
+    return summary;
+}
+
+function getTM4JTestCases(report) {
+    const allTests = getAllTests(report.results);
+
+    const re = /(MM-T)\w+/g;
```
josephbaylon, line 51: Maybe change the regex to `/^(MM-T)\w+/g;` to make it starts with ticket number since the item title split for the key below assumes it is

e2e/utils/test_cases.js
saturninoabril: Sure, will change.

e2e/utils/test_cases.js
saturninoabril: Good catch! I was switching back and forth on this earlier while testing some cases.

saturninoabril:
@josephbaylon Thanks for the reviews, now updated, PTAL again.

e2e/utils/test_cases.js
```diff
@@ -0,0 +1,163 @@
...
+
+function getStepStateSummary(steps = []) {
+    const result = getStepStateResult(steps);
+
+    return Object.entries(result).map(([key, value]) => `${value} ${key}`).join(',');
+}
+
+function getTM4JTestCases(report) {
+    return getAllTests(report.results).
+        filter((item) => /^(MM-T)\w+/g.test(item.title)).
```
saturninoabril, line 42: For some reason, the regex didn't work as expected on some cases when I assigned it initially in a variable. Directly using it did the trick.

e2e/utils/test_cases.js
```diff
@@ -0,0 +1,163 @@
...
+            actualEndDate: testScriptResults[testScriptResults.length - 1].actualEndDate,
+            executionTime: steps.reduce((acc, prev) => {
+                acc += prev.duration; // eslint-disable-line no-param-reassign
+                return acc;
+            }, 0),
+            comment: `Cypress automated test - ${getStepStateSummary(steps)}`,
+        };
+
+        // Temporarily log to verify cases that were being saved.
+        console.log(index, key); // eslint-disable-line no-console
```
saturninoabril, line 144: Intentionally added so I can verify manually the test cases being filtered and saved. Will probably remove in the future but is important to be here currently.

e2e/utils/test_cases.js
josephbaylon: I think you need to use `new RegEx(/^(MM-T)\w+/g)` for variable assignment

saturninoabril: That's interesting and so I tried. However, same result, it didn't work. When assigned to a variable, all patterns from this [json](https://mm-cypress-report.s3.amazonaws.com/703-integrate-tm4j-master/all.json) were matched except with `MM-T1784_1` up to `MM-T1784_4` which got 2 of 4 steps only. But with `/^(MM-T)\w+/g.test(item.title)`, all 4 got the match.

PR #6077, opened 2020-08-05
[MM-26469] Complete Profile Step

devinbinnie:
#### Summary
Complete Profile step (Step 1)
Hooked up and working.

Currently repeated in each step, but that will change as new steps are added.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26469


components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
```diff
@@ -0,0 +1,173 @@
...
+                            <FormattedMarkdownMessage
+                                id='next_steps_view.complete_profile_step.pictureError'
+                                defaultMessage='Photos must be in BMP, JPG or PNG format. Maximum file size is {max}.'
+                                values={{max: Utils.fileSizeToString(this.props.maxFileSize)}}
+                            />
+                        </>
+                    )}
+                </span>
+                <div className='NextStepsView__wizardButtons'>
+                    {/* <button
```
nickmisasi, line 150: Remove?

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    onRemovePicture = () => {
+        this.setState({profilePicture: undefined, profilePictureError: false, removeProfilePicture: true});
+    }
+
+    render() {
+        const {currentUser} = this.props;
+
+        // Make sure picture has been set
+        const pictureSrc = currentUser.last_picture_update ? Utils.imageURLForUser(currentUser.id, currentUser.last_picture_update) : undefined;
```
nickmisasi, line 99: Setting things to undefined can be kind of weird sometimes. I know its not usually a problem, but could this be null?

devinbinnie: In TypeScript, when defining an optional parameter, it prefers to use `undefined` as the fallback type instead of `null`.
So when you defined a parameter like this: `param?: string`, it's type evaluates to `string | undefined`.
We use `undefined` to match the type.

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
devinbinnie: This might actually come back, we removed it for UX reasons. @matthewbirtch thoughts on just blowing this out?

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
nickmisasi: I still have my JavaScript glasses on 🤓

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
nickmisasi: If that's the case I'll approve for now.

PR #6079, opened 2020-08-06
Cypress/E2E: MM-T1784 View new messages toast

saturninoabril:
#### Summary
Introduce test key of `[tm4j_id]_[step]` for test case that has multiple steps in TM4J. Such key will help us map each step with pass/fail instead of just the entire status of a test case.

We can discuss this more separately but in general, we should aim for one step only per test case. Not only due to mapping issue between Cypress and TM4J but for simplicity and being specific of a test case.

Test case - https://automation-test-cases.vercel.app/test/MM-T1784


e2e/cypress/integration/toast/helpers.js
```diff
@@ -0,0 +1,42 @@
...
+        scrollTo('bottom', {duration: TIMEOUTS.ONE_SEC}).
+        wait(TIMEOUTS.ONE_SEC);
+}
+
+export function scrollUp() {
+    // # Scroll up so bottom is not visible
+    cy.get('div.post-list__dynamic').should('be.visible').
+        scrollTo(0, '70%', {duration: TIMEOUTS.ONE_SEC}).
+        wait(TIMEOUTS.ONE_SEC);
+}
```
saturninoabril, line 42: Helper functions here can be applied to `toast_spec.js` but don't want to make changes there to avoid potential conflicts with the active PRs for Hackfest. Also, there are some tests that needs to verify individually.

josephbaylon:
@saturninoabril General question: For tickets that only have one step, does user need to add `_1` to it? If not, will your previous reporting changes handle the update of that single step or is it automatic on TM4J to update step when there's only one?

saturninoabril:
For one step, no need to add `_1`. And yes, the recent integration handles it - [testScriptResults](https://github.com/mattermost/mattermost-webapp/blob/master/e2e/utils/test_cases.js#L133) will always be 1 or more.

PR #6080, opened 2020-08-06
MM-27165 - Updating post menu css

asaadmahmood:
#### Summary
MM-27165 - Updating post menu css
MM-26860 - Bot username spacing issue

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27165
https://mattermost.atlassian.net/browse/MM-26860

#### Screenshots
**Post menu css**
Screenshot 2020-08-06 at 8 16 45 PM

**Bots in thread**
Screenshot 2020-08-20 at 3 46 10 PM

### DIFF
--- a/components/post_view/post/post.jsx
+++ b/components/post_view/post/post.jsx
@@ -214,7 +214,7 @@ class Post extends React.PureComponent {
         return false;
     }
 
-    getClassName = (post, isSystemMessage, isMeMessage, fromWebhook, fromAutoResponder) => {
+    getClassName = (post, isSystemMessage, isMeMessage, fromWebhook, fromAutoResponder, fromBot) => {
         let className = 'post';
 
         if (post.failed || post.state === Posts.POST_DELETED) {
@@ -233,6 +233,10 @@ class Post extends React.PureComponent {
             rootUser = 'same--root';
         } else {
             rootUser = 'other--root';
+        }
+
+        if (fromBot) {
+            className += ' post--bot';
         }
 
         let currentUserCss = '';
--- a/components/post_view/post/post.jsx
+++ b/components/post_view/post/post.jsx
@@ -214,7 +214,7 @@ class Post extends React.PureComponent {
         return false;
     }
 
-    getClassName = (post, isSystemMessage, isMeMessage, fromWebhook, fromAutoResponder, fromBot) => {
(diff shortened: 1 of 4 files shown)


components/post_view/post/post.jsx
```diff
@@ -229,7 +229,7 @@ class Post extends React.PureComponent {
         }
 
         let rootUser = '';
-        if (this.hasSameRoot(this.props) && !fromBot) {
```
hmhealey, line 232: This was added to fix https://mattermost.atlassian.net/browse/MM-15228 from what I can tell. Can you confirm that this isn't going to cause a regression in that ticket?

asaadmahmood: Yup, it definitely causes an issue there. I'll fix that and this ticket together.

asaadmahmood:
@hmhealey Should be fixed.
Screenshot 2020-08-26 at 3 26 43 PM

components/rhs_comment/rhs_comment.jsx
```diff
@@ -282,6 +283,7 @@ class RhsComment extends React.PureComponent {
         const isSystemMessage = PostUtils.isSystemMessage(post);
         const isMeMessage = checkIsMeMessage(post);
         const fromAutoResponder = PostUtils.fromAutoResponder(post);
+        console.log(post);
```
hmhealey, line 286: Looks like you left some debugging stuff by mistake. Otherwise, this looks good though

srkgupta (changes requested):
Hi @asaadmahmood 

The first part of the issue, i.e. `Post hover menu partially hides post text` still needs to be fixed when the Display is set as Compact View. Please check out the screenshot of the issue seen:

Screenshot 2020-09-09 at 9 12 42 PM

The second issue is fixed and is working fine now.

asaadmahmood:
@srkgupta That case is intentionally left out, because on a normal post, the reply icon sits on the right, I cannot place the post menu on top, because then there will be a shift in where the reply icon appears.

Screenshot 2020-09-10 at 8 08 22 PM

So we can accept it. Once we have a different solution for the reply icon and the number, we can move the post menu up all the time.

srkgupta (approved):
Approving the PR based on the above comment. Thanks @asaadmahmood

PR #6081, opened 2020-08-06
[MM-26970] Add filters to users to remove modal

fmunshi:
#### Summary
- Converts the existing `group_users` component to use the new DataGrid
- Adds search and membership filters

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26970

#### Screenshots
Screen Shot 2020-08-06 at 12 55 12 PM
Screen Shot 2020-08-06 at 12 56 43 PM


components/admin_console/filter/filter_checkbox.tsx
```diff
@@ -11,24 +11,38 @@ type Props = {
...
-                        checked={checked}
-                        onChange={this.toggleOption}
-                    />
+                    {checked &&
+                        <input
+                            type='checkbox'
+                            id={name}
+                            name={name}
+                            defaultChecked={true}
+                        />
```
fmunshi, line 35: This might seem weird but I was having issues where the input checkbox would not get updated in the DOM when rendered inside of the bootstrap modal... my assumption is that somewhere something was calling a `stopPropogation` or `preventDefault` on the click event, so instead of relying on the browser checkbox behaviour i've modified this to simply swap out the input box to a checked one when it is checked and the opposite when not checked

components/admin_console/team_channel_settings/group/group_users/group_users.tsx
```diff
@@ -0,0 +1,393 @@
...
+
+    async componentDidMount() {
+        const {members, total} = this.props;
+        const MEMBERSHIPS_TO_LOAD_COUNT = 100;
+        const promises = [];
+        let membershipsLoaded = 0;
+
+        // Pre-load all memberships since users are also already loaded into the state
+        while (membershipsLoaded < total) {
+            promises.push(this.loadMembersForProfilesList(members.slice(membershipsLoaded, membershipsLoaded + MEMBERSHIPS_TO_LOAD_COUNT)));
```
hmhealey, line 104: Should the loading of channel members be done lazily as more pages are viewed? Loading them all at once seems like it could cause performance issues in larger channels/teams

components/admin_console/team_channel_settings/group/group_users/group_users.tsx
```diff
@@ -0,0 +1,393 @@
...
+                    rows={rows}
+                    loading={this.state.loading}
+                    page={this.state.page}
+                    nextPage={this.nextPage}
+                    previousPage={this.previousPage}
+                    startCount={startCount}
+                    endCount={endCount}
+                    total={total}
+                    search={this.search}
+                    filterProps={filterProps}
```
hmhealey, line 386: We're constructing a new instance of `filterProps` each time this component re-renders which means that the `DataGrid` will also re-render each time, assuming it's also a PureComponent. Is that okay, or could that cause performance issues?

components/admin_console/team_channel_settings/group/group_users/group_users_groups.tsx
```diff
@@ -0,0 +1,62 @@
...
+export default class GroupUsersGroups extends React.PureComponent<GroupUsersGroupsProps, {}> {
+    renderGroupsColumn = (user: ProfileWithGroups) => {
+        const groups = user.groups || [];
+        if ((groups).length === 1) {
+            return groups[0].display_name;
+        }
+
+        const message = (
+            <FormattedMessage
+                id={t('team_channel_settings.group.group_user_row.numberOfGroups')}
```
hmhealey, line 30: We shouldn't need to use `t` here. The tooling automatically recognizes translation strings inside of `FormattedMessage`s even without it.

components/admin_console/team_channel_settings/group/group_users/group_users_groups.tsx
```diff
@@ -0,0 +1,62 @@
...
+import OverlayTrigger from 'components/overlay_trigger';
+
+type ProfileWithGroups = Partial<UserProfile & {
+    groups: Partial<Group>[];
+}>;
+
+interface GroupUsersGroupsProps {
+    user: ProfileWithGroups;
+}
+export default class GroupUsersGroups extends React.PureComponent<GroupUsersGroupsProps, {}> {
```
hmhealey, line 21: Perhaps this could be done as a functional component instead? We're trying to start using those wherever we can since they can contain hooks and since they're supposed to perform better in the future.

components/admin_console/team_channel_settings/group/group_users/group_users_role.tsx
```diff
@@ -0,0 +1,78 @@
...
+}>;
+
+interface GroupUsersRoleProps {
+    user: ProfileWithGroups;
+    membership: TeamMembership | ChannelMembership;
+    scope: 'team' | 'channel';
+}
+
+type Role = 'system_admin' | 'team_admin' | 'team_user' | 'channel_admin' | 'channel_user' | 'guest';
+export default class GroupUsersRole extends React.PureComponent<GroupUsersRoleProps, {}> {
```
hmhealey, line 25: Same question here since it's relatively simple. Plus, if you use a functional component, you can use the `useIntl` hook instead of `Utils.localizeMessage` since the latter isn't particularly React-friendly.

It'd look something like
```typescript
import {useIntl} from 'react-intl';

export default function GroupUsersRole(props: GroupUsersRoleProps) {
    const intl = useIntl();

    const role = getCurrentRole(props);

    let localizedRole;
    switch (role) {
        localizedRole = intl.formatMessage({id: 'admin.user_grid.system_admin', defaultMessage: 'System Admin'});
    }

    // ...
}
```

hmhealey: Also, the empty object for the `State` type isn't necessary. You can just omit that type parameter for class components.

components/admin_console/team_channel_settings/group/group_users/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    }>;
+    setModalSearchTerm: (term: string) => Promise<{
+        data: boolean;
+    }>;
+    setModalFilters: (filters: Filters) => Promise<{
+        data: boolean;
+    }>;
+};
+
+const searchUsers = createSelector(
```
hmhealey, line 46: This is kind of misusing `createSelector`. Instead, you can use `memoizeResult` from 'mattermost-redux/utils/helpers` to do the same thing, but in a more concise way.

Also, while you might not have multiple instances of `GroupUsers` mounted at once, it's good practice to create any memoized functions in a `makeMapStateToProps` so that multiple instances of the component don't overwrite each others' memoized results.

```typescript
import {memoizeResult} from 'mattermost-redux/utils/helpers';

    const searchUsers = memoizeResult(((users: UserProfile[], term: string, filters: Filters, memberships: Memberships): UserProfile[] => {
        // ...
    });

    return function mapStateToProps(state: GlobalState, props: Props) {
        // ...
    };
}

export default connect(makeMapStateToProps, mapDispatchToProps)(GroupUsers);
```

components/admin_console/team_channel_settings/group/group_users/group_users.tsx
fmunshi: I think DataGrid being re-rendered anytime this is component is re-rendered is kind of inevitable, what would be a better way to pass these props? Just making the object inline? like `filterProps={{options, keys, onFilter}}`?

components/admin_console/team_channel_settings/group/group_users/group_users.tsx
fmunshi: So my initial implementation had me loading the memberships lazily... 

_But_ theres no api endpoints to search channel members minus group members with the required filters too, to avoid having to do a bunch of server side work to add the support to those APIs I opted to just load all the memberships when this dialog is opened and do all the filtering locally, I think it shouldnt be too big of a deal since this modal will never really be _all_ of the channel or team members.. it just shows the members to be _removed_ when enabling group sync on a channel or team. I can test it out with a few thousand users later but it should be fine this component isnt worth the time investment at the moment I think.

components/admin_console/filter/filter_checkbox.tsx
fmunshi: yea... this.props.checked is being updated correctly but the actual checkbox itself does not get re-rendered, im not sure whats going on but this implementation if simply having two different checkboxes for a checked and unchecked state works fine... unless you got any suggestions as to why the checkbox wasnt updating before.

components/admin_console/team_channel_settings/group/group_users/group_users.tsx
hmhealey: Is there perhaps an API to get channel members by their user IDs? That said, I agree that the performance issue won't be exactly huge, so it might not be something we need to address now.

components/admin_console/team_channel_settings/group/group_users/group_users.tsx
hmhealey: You could consider saving `filterProps` in `this.state` and only modifying it when the props change. `DataGrid` re-rendering may not be the end of the world unless there's something else on this screen that's particularly performance-sensitive, but it's something to keep in mind with more complex components like it.

fmunshi:
@hahmadia retagging you for this PR since you're more familiar with the changes its making

components/admin_console/team_channel_settings/channel/details/channel_details.tsx
```diff
@@ -240,6 +240,8 @@ export default class ChannelDetails extends React.PureComponent<ChannelDetailsPr
                                 <UsersWillBeRemovedError
                                     total={usersToRemoveCount}
                                     users={result.data.users}
+                                    scope={'channel'}
```
hahmadia, line 243:
```suggestion
                                    scope='channel'
```

srkgupta (changes requested):
Hi @fmunshi 

Following issues were observed while testing the PR. Can you please fix the same:

[MM-28272](https://mattermost.atlassian.net/browse/MM-28272) - Filters are not cleared when the View Members modal is closed in System Console

[MM-28273](https://mattermost.atlassian.net/browse/MM-28273) - Message in Users to be removed modal in Channel Configuration page should indicate channel and not team

fmunshi:
@srkgupta I have fixed the two issues noted above, thanks!

srkgupta (approved):
Tested on the updated PR and all the issues are fixed now and the feature is working fine as expected. Approving the PR.

PR #6084, opened 2020-08-06
MM-23760 Allow typing into theme picker fields

hmhealey:
The main reason that you're currently unable to type into the theme picker field is because we currently reject any invalid values in the fields in the theme picker, and that validation happens every single key press, so it would often reject any changes the user tried to make (such as backspacing a single character).

The new version uses `tinycolor2` (which is already used internally by the `react-color` library that we use) to handle converting any invalid input to valid input. It runs the validation continuously to generate valid colours for use in the theming, but it only writes that colour back to the input when it's blurred. This means that the user's input is never affected by the validation.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23760


components/color_input.tsx
```diff
@@ -3,46 +3,50 @@
...
             isOpened: false,
-            hex: this.props.color.toUpperCase(),
+            value: props.value,
         };
     }
 
-    public componentWillMount() {
-        this.setHex();
+    static getDerivedStateFromProps(props: Props, state: State) {
+        if (!state.focused && props.value !== state.value) {
```
calebroseland, line 37: When dragging `<ChromePicker/>`, this currently nukes the value-change (appears frozen while dragging).

```suggestion
        if (!state.focused && !state.isOpened && props.value !== state.value) {
```

components/color_input.tsx
```diff
@@ -169,7 +164,7 @@ class ColorInput extends React.PureComponent<Props, State> {
                         id={`${id}-ChromePickerModal`}
                     >
                         <ChromePicker
-                            color={color}
+                            color={value}
                             onChange={this.handleColorChange}
```
calebroseland, line 168: Maybe only propagate at end of dragging? (debounce)

```suggestion
                            onChange={this.handleTempColorChange} // setState
                            onChangeComplete={this.handleColorChange} // props.onChange
```

components/color_input.tsx
hmhealey: This works fine for me. I'm seeing the textbox update live as you drag around
image

components/color_input.tsx
hmhealey: I was originally planning on something like that, but that's needed to update the "preview" theme as you drag (see the sidebar above). We could consider changing it at some point, but this matches current behaviour for now.

calebroseland: I didn't realize that was required (live while dragging). 

I was testing on a local dev build, in Chrome (12+ tabs) with devtools open, and I was seeing quite a bit of lag (only applying change on mouseup). I tried tried applying this (onChange + onChangeComplete), and thats what brought up the `getDerivedStateFromProps` comment/suggestion. (When opening picker, `focused` is still true until the first drag has ended.)

The production build has much better performance—especially with devtools closed 😛.

hmhealey: I think the performance issues are mostly caused by `applyTheme` being awful 😛 

Some debouncing might help that, but I'm not too concerned about it right now

PR #6088, opened 2020-08-06
MM-27574: Add Learn More subtext to Admin Advisor modal

catalintomai:
#### Summary
Add "Learn More" link and text to Admin. Advisor "Contact us" modal.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27574

#### Related Pull Requests
server: https://github.com/mattermost/mattermost-server/pull/15199


components/warn_metric_ack_modal/warn_metric_ack_modal.tsx
```diff
@@ -161,6 +162,28 @@ export default class WarnMetricAckModal extends React.PureComponent<Props, State
             />
         );
 
+        const learnMoreLink = 'https://mattermost.com/pl/default-admin-advisory';
+        const subText = (
+            <div
+                style={{display: 'flex', opacity: '0.56', flexWrap: 'wrap'}}
```
fmunshi, line 168: would make more sense if this was just added to the class? or maybe a second class added to this with the above styles, something like `help__format-subtext`? 1/5 on it though doesnt really matter too much

catalintomai: probably, though in practice we create these one-off css classes that are never re-used, so I chose to keep the overriding settings inline since they are not reused (similar to an anonymous class/inline const).

PR #6092, opened 2020-08-07
[MM-T570] Integration Page titles are bolded

djanda97:
#### Summary

Adds an E2E test to check if Integration Page titles are bolded.

#### Ticket Link

MM-T570

### DIFF
--- a/e2e/cypress/integration/integrations/integrations_spec.js
+++ b/e2e/cypress/integration/integrations/integrations_spec.js
@@ -189,7 +189,7 @@ describe('Integrations page', () => {
         // # Go to Main Menu -> Integrations
         cy.get('#sidebarHeaderDropdownButton').should('be.visible').click();
 
-        cy.get('#integrations').should('be.visible').click();
+        cy.get('.dropdown-menu').should('be.visible').findByText('Integrations').click();
 
         cy.get('.integration-option__title').contains('Incoming Webhooks').click();
 
@@ -197,10 +197,11 @@ describe('Integrations page', () => {
         integrationPageTitleIsBold('Outgoing Webhooks');
         integrationPageTitleIsBold('Slash Commands');
         integrationPageTitleIsBold('OAuth 2.0 Applications');
+        integrationPageTitleIsBold('Bot Accounts');
     });
 });
 
 function integrationPageTitleIsBold(title) {
     cy.get('.section-title__text').contains(title).click();
-    cy.get('.item-details__name').should('have.css', 'font-weight', '600');
+    cy.get('.item-details__name').should('be.visible').and('have.css', 'font-weight', '600');
 }
--- a/e2e/cypress/integration/integrations/integrations_spec.js
+++ b/e2e/cypress/integration/integrations/integrations_spec.js
@@ -13,6 +13,8 @@
 import {getRandomId} from '../../utils';
 
 describe('Integrations page', () => {
+    let testTeam;
+
     before(() => {
         // # Set ServiceSettings to expected values
         const newSettings = {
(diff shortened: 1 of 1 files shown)


djanda97:
I appreciate any feedback or tips about writing cleaner tests as I am still new to using Cypress.

e2e/cypress/integration/integrations/integrations_spec.js
```diff
@@ -178,4 +182,25 @@ describe('Integrations page', () => {
...
         cy.get('#emptySearchResultsMessage').should('be.visible').and('have.text', `No bot accounts match ${searchString}`);
     });
+
+    it('MM-T570 Integration Page titles are bolded', () => {
+        cy.visit(`/${testTeam.name}/channels/town-square`);
+
+        // # Go to Main Menu -> Integrations
+        cy.get('#sidebarHeaderDropdownButton').should('be.visible').click();
+
+        cy.get('#integrations').should('be.visible').click();
```
saturninoabril, line 192: I'm good as is but would be better if
- scoped within a menu section (though we have this kind of approach in several places) and,
- if we can leverage [API queries from testing-library](https://testing-library.com/docs/dom-testing-library/api-queries) since it works in the perspective of test user rather than the DOM structure under the hood. Something like
```suggestion
        cy.get('.dropdown-menu').should('be.visible').findByText('Integrations').click();
```

e2e/cypress/integration/integrations/integrations_spec.js
```diff
@@ -178,4 +182,25 @@ describe('Integrations page', () => {
...
+
+        cy.get('.integration-option__title').contains('Incoming Webhooks').click();
+
+        integrationPageTitleIsBold('Incoming Webhooks');
+        integrationPageTitleIsBold('Outgoing Webhooks');
+        integrationPageTitleIsBold('Slash Commands');
+        integrationPageTitleIsBold('OAuth 2.0 Applications');
```
saturninoabril, line 199: I recently updated the test to include check for Bot accounts. Could please insert,
```javascript
integrationPageTitleIsBold('Bot Accounts');
```

e2e/cypress/integration/integrations/integrations_spec.js
```diff
@@ -178,4 +182,25 @@ describe('Integrations page', () => {
...
+    });
 });
+
+function integrationPageTitleIsBold(title) {
+    cy.get('.section-title__text').contains(title).click();
+    cy.get('.item-details__name').should('have.css', 'font-weight', '600');
```
josephbaylon, line 205:
```suggestion
    cy.get('.item-details__name').should('be.visible').should('have.css', 'font-weight', '600');
```

e2e/cypress/integration/integrations/integrations_spec.js
djanda97: I'm not quite sure how to scope this within a menu section. Could you please clarify how that would look?

djanda97: Using the testing-library looks like a more maintainable approach. Thank you for introducing it to me!

e2e/cypress/integration/integrations/integrations_spec.js
djanda97: Thank you for catching this!

e2e/cypress/integration/integrations/integrations_spec.js
saturninoabril: I missed your question. `cy.get('.dropdown-menu').should('be.visible').findByText('Integrations').click()` is one way to scope which is getting the dropdown menu and then finding "Integrations" in the menu. Another is you'll find the use of `.within(() => {})`

djanda97: Thanks for the explanation, I understand it now!

PR #6095, opened 2020-08-07
MM-26053: change DM name format according to teammate name display se…

agnivade:
…tting

When the preference is set to display full_name, we display the full name first,
and then the username. This helps keep the correct order in between GMs and DMs
while searching.

This is because the GM display already changes according to that setting,
but the DM name did not. So that caused confusion when searching for a full name
because giving a space after the first name would cause the GM to jump ahead of the DM.

This keeps things consistent.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-26053


components/suggestion/switch_channel_provider.test.jsx
```diff
@@ -77,9 +77,9 @@ describe('components/SwitchChannelProvider', () => {
     };
 
     const switchProvider = new SwitchChannelProvider();
-    const mockStore = configureStore();
+    let mockStore = configureStore();
```
hmhealey, line 80: Perhaps instead of initializing `mockStore` for the whole test suite and the overriding it in the new test case, maybe we should just initialize individual versions of `mockStore` in each test case instead? As it is, I think any test cases added after your new one will also pick up the modified state which is a little weird.

components/suggestion/switch_channel_provider.jsx
```diff
@@ -332,14 +332,30 @@ export default class SwitchChannelProvider extends Provider {
...
     userWrappedChannel(user, channel) {
-        let displayName = `@${user.username}`;
-
-        if ((user.first_name || user.last_name) && user.nickname) {
-            displayName += ` - ${Utils.getFullName(user)} (${user.nickname})`;
-        } else if (user.nickname) {
-            displayName += ` - (${user.nickname})`;
-        } else if (user.first_name || user.last_name) {
-            displayName += ` - ${Utils.getFullName(user)}`;
+        const teammateNameDisplay = getTeammateNameDisplaySetting(getState());
```
hmhealey, line 335: Are we sure that we want to change what's displayed in the picker based on the name format? I don't think we do this anywhere else that we autocomplete users.

hahmadia:
Hey Agniva,

I noticed in the file you've edited, there is a function called `quickSwitchSorter`. At about line 223, there is a function called which is located in the redux repo called `sortChannelsByTypeAndDisplayName`. 

Within `sortChannelsByTypeAndDisplayName`,  we have this bit of code which determines the sort order by type 
```JS
    if (channelTypeOrder[a.type] !== channelTypeOrder[b.type]) {
        if (channelTypeOrder[a.type] < channelTypeOrder[b.type]) {
            return -1;
        }

        return 1;
    }
```
where `channelTypeOrder` is
```JS
const channelTypeOrder = {
    [General.OPEN_CHANNEL]: 0,
    [General.PRIVATE_CHANNEL]: 1,
    [General.DM_CHANNEL]: 2,
    [General.GM_CHANNEL]: 2,
};
```
Defined in `src/utils/channel_utils.ts` in mattermost-redux.

I noticed that `DM` and `GM` channel have their type order values to be the same therefore that bit of code above doesn't get executed. I was curious, would giving `GM` channels a lower priority help solve this problem without having to change the outputted text? If you keep `[General.DM_CHANNEL]: 2` and change `[General.GM_CHANNEL]: 3`, I _guess_ it will prioritize DM's over GM's and ensure those are first. 

Hope this helps.
Thanks

components/suggestion/switch_channel_provider.test.jsx
agnivade: Yeah, I thought about that and would prefer that actually. I didn't know what is the usual convention followed here. It would have just created a lot more duplicated code. Will change it. Thanks.

components/suggestion/switch_channel_provider.jsx
agnivade: Yeah, it's needed for the autocomplete to give correct results. The problem, as I outlined earlier is that GM names already respect the preference setting. But DMs don't. So if you search for "Harrison", everything is fine, but the moment you enter "Harrison H", GMs jump higher than DMs because DMs have "harrison.healy" before so it doesn't match.

You are right, and I just found that the search bar also has the same problem. We need to update that too.

agnivade:
Thanks Hossein, yes that part of the code was my first suspect. However, that isn't really the problem here. The core issue is that the names are actually different in GMs and DMs. So even if the priority becomes different, the order will still be different because what you are searching for actually does not match the DM name, but will match the GM name. And therefore GM appears higher, and DMs go way below.

Your suggestion still applies though, and probably makes sense. I'll send a PR some time.

components/suggestion/switch_channel_provider.jsx
agnivade: Ok, I checked a bit more and there are 3 places where we need to update this - UserSuggestion, SearchUserSuggestion, and AtMentionSuggestion. Combined with this change, it would change 4 places. Are we sure we want to do this? The initial ticket was meant as a bug in list ordering, but then it actually didn't turn out to be a bug, but a feature change. Now this is turning out to be an even bigger change. The only possible bug here is that `channelTypeOrder` in mattermost-redux is the same for GM and DM whereas DMs should be higher. But that isn't strictly related to this ticket.

@esethna / @wiersgallak - requesting confirmation on whether we really want to go ahead with this.

hmhealey:
I'm not familiar with where autocompleting is done for DMs/GMs in the channel switcher, but the problem likely has to do with DMs and GMs not having their `DisplayName` populated on the server. GMs have a comma-separated list of usernames and DMs have an empty one. Some of the selectors for channels in the clients populate that field with the proper names using the name format setting, but others don't since the selectors that populate it are generally less efficient.

agnivade:
> I'm not familiar with where autocompleting is done for DMs/GMs in the channel switcher, 

That's in components/suggestion/switch_channel_provider.jsx which the PR changes. Unless you are looking for somewhere else?

> Some of the selectors for channels in the clients populate that field with the proper names using the name format setting, but others don't since the selectors that populate it are generally less efficient.

I am not sure about the efficiency part. What I observed was that the name format setting was only respected for GMs and not DMs. DMs have a fixed format everywhere. 

Honestly, I am not sure what to do about this, so I will defer to whatever you think is better.

components/suggestion/switch_channel_provider.test.jsx
hmhealey: I don't think it should be too much duplicated code if we share the base state. I think calling `const mockStore = configureStore;` in each test case is important to prevent tests from interfering with others' testing conditions though. I'm also a fan of the Go idea of "a little duplication is better than a little dependency" though.

hmhealey:
I'll be honest. I'm not entirely sure what I meant yesterday. I think I meant that I know some of it is done in that file, but I couldn't remember if some of it was done by the server as well, so I think I meant the server-side code could be causing a bit of the confusion as well.

Regardless, this works as I would expect after some quick testing, so I'm not sure what I was suggesting to be changed.
Screen Shot 2020-08-12 at 9 14 32 AM

For the efficiency part, I just meant that the display_name isn't always set for DMs/GMs depending on how they're retrieved from the redux store, so if you were running into any weirdness anywhere, that might be related. If you're not running into any problems though, you don't need to worry about that.

agnivade:
The thing that is required is that the same name change format needs to be done for `UserSuggestion`, `SearchUserSuggestion`, and `AtMentionSuggestion` as well. It's doable, but it's a fairly big change. Let me know if that sounds okay to you.

agnivade:
Given the silence, I am assuming it is fine to implement this on the other components too. Please comment if you disagree.

agnivade:
Ok, there are usability problems with `AtMentionSuggestion` and `SearchUserSuggestion` if we do this. In at mentions, the text followed by `@` is always a username. So if the username is not the first thing in the autocomplete list, it will just confuse users as they need to look further to the right. I don't think we want that. Same with search users. In the test after `from:` we only take usernames. Pressing a space goes to the next item and the autocomplete bar goes away.

The only place where this can be applied is the user filter in admin console. But I don't think it even matters what name display format is chosen as it's not even a user facing feature.

So considering all of the above, I think the current change is fine.

hmhealey:
It might be better to mention them in Mattermost. I don't think either of them check notifications in GitHub very often.

agnivade:
I will. Thanks.

agnivade:
Clarified with @esethna. The current approach is OK.

@hmhealey - So nothing has really changed from the original PR. Just now the sub-tests are all self-contained as you suggested. So it should be good to go.

agnivade:
It definitely makes sense to bump up DMs higher in priority. They are already treated that way elsewhere in the codebase. I think that can be a separate PR.

agnivade:
Bump on this @hmhealey

PR #5632, opened 2020-06-01
MM-23573: Remove Segment.

grundleborg:
#### Summary
Remove Segment.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23573

#### QA Testing Steps

Start up a spin mint and check that basic functionality works (log in, post message, change channels). Then talk to Bizops Team to check that **client side** Rudder data is being received from your test server.

PR #6099, opened 2020-08-08
MM-27537 remove segment remnant (fix hosting-in-subpath issue)

calebroseland:
#### Summary

- Remove segment from meta tag in root.html

**This fix prevents a fatal web client error for installations running under a subpath.**

Background: [This mattermost-server code block](https://github.com/mattermost/mattermost-server/blob/v5.26.0-rc1/utils/subpath.go#L88-L96)  is a part of the subpath-rewriting logic, and looks for a specific meta tag to rewrite. The strings were mismatched and failed, short-circuiting the rest of the subpath rewrite process.

Notes: 
- Related PR #5632 
- We should consider allowing subpath rewriting ([here](https://github.com/mattermost/mattermost-server/blob/v5.26.0-rc1/utils/subpath.go#L144-L149)) on local development when using webapp:`make build` and server:`make run-server`, which allows for the simulation of production-like environments.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27537

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -264,7 +264,7 @@ var config = {
             meta: {
                 csp: {
                     'http-equiv': 'Content-Security-Policy',
-                    content: 'script-src \'self\' cdn.segment.com/analytics.js/ cdn.rudderlabs.com/' + CSP_UNSAFE_EVAL_IF_DEV,
+                    content: 'script-src \'self\' cdn.rudderlabs.com/' + CSP_UNSAFE_EVAL_IF_DEV,
                 },
             },
         }),


calebroseland:
cc @amyblais

amyblais:
@grundleborg @jespino Please prioritize reviews for v5.26,

lindalumitchell:
> We should consider allowing subpath rewriting (here) on local development when using webapp:make build and server:make run-server, which allows for the simulation of production-like environments.

@calebroseland I take it from ^your comment that it still isn't straightforward to set up a test server that uses a subpath, is that right? If so, I'm happy to approve this PR through and test it after it's merged.

calebroseland:
@lindalumitchell I think it may be possible to do this in a test server, but I'm not 100% sure. I'll do a quick check, but otherwise yes, it may have to be tested after merging.

calebroseland:
@lindalumitchell should be good-to-go

PR #6100, opened 2020-08-08
MM-T463: E2E to validate maximum users in group message

Niceplace:
#### Summary

- Add a few HTML id attributes to the MultiSelect component to assist with E2E tests
- Add a cypress test that validates a group message between users cannot contain more than 8 people

#### Ticket Link

JIRA Ticket: MM-T463
Related test scenario: https://automation-test-cases.vercel.app/test/MM-T463

### DIFF
--- a/e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
+++ b/e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
@@ -1,0 +1,139 @@
+
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+const createUserAndAddToTeam = (team) => {
+    cy.log(`CREATE USER AND ADD TO TEAM ${JSON.stringify(team, null, 2)}`);
+    cy.apiCreateUser().then(({user}) =>
+        cy.apiAddUserToTeam(team.id, user.id),
+    );
+};
+
+const addUsersToGMViaModal = (userCountToAdd) => {
+    // Ensure there are enough selectable users in the list
+    cy.get('#multiSelectList').
+        should('be.visible').
+        children().
+        should('have.length.gte', userCountToAdd);
+
+    // Add users 1 by 1 and ensure the help section's text updates accordingly
+    Cypress._.times(userCountToAdd, () => {
+        cy.get('#multiSelectList').
+            children().
+            first().
+            click();
+    });
+};
+
+describe('Multi-user group messages', () => {
+    /**
+ * MM-T463 GM: Maximum users
+Steps
+GM: Maximum users
+--------------------
+1. Add 7 users to the same GM (but don't Save yet) - DONE
+2. Non-RN: Verify says "You can add 0 more people" and also "You've reached the maximum number of people for this conversation. Consider creating a private channel instead."
+3. Try to click another user from the list or use arrows to highlight a user and press Enter
+4. Verify additional user is not added to the list at the top
+5. Save / Start to open GM channel
+6. Click channel name > Add Members (still says you can add zero members)
+7. Backspace or click "x" next to a username to remove it from the GM
+Expected
+After last step, message says "You can add 1 more person."
+Also says "This will start a new conversation. If you‚Äôre adding a lot of people, consider creating a private channel instead."
+
+Desktop Only:
+Ensure that "Write to..." text in main imput box is truncated and does not wrap if all of the names are too long
+
+(Can then add another user and Save to open the GM)
+ */
+    let testTeam;
+    before(() => {
+        // # Login as test user and visit town-square
+        cy.apiInitSetup().then(({team}) => {
+            testTeam = team;
+
+            // Add 10 users to the team
+            Cypress._.times(10, createUserAndAddToTeam(testTeam));
+        });
+    });
+
+    it('MM-T463 Should not be able to create a group message with more than 8 users', () => {
+        const maxUsersGMNote = "You've reached the maximum number of people for this conversation. Consider creating a private channel instead.";
+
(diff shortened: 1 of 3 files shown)


Niceplace:
@saturninoabril When removing a user from the "Direct messages" dialog, but only when accessing this dialog from "Add users" dropdown from the group message, the warning note that suggests creating a private channel instead never goes away. I tested manually further and even if I remove all the users it is still displayed. Is this expected ? Joining a gif to better explain what's going on.

mattermostMaxUsersGm_NoteNotGoingAway

e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
```diff
@@ -0,0 +1,126 @@
...
+        cy.get('#channelAddMembers').click();
+
+        // # Try to add one more user from the group message's list
+        addUsersToGMViaModal(1);
+
+        // # Check that the appropriate information & warnings are displayed in the group message's dialog
+        expectCannotAddUsersMessage(0);
+
+        // * Check that the count of selected number of users in the group message's dialogis is 7
+        cy.get('.react-select__multi-value').
```
Niceplace, line 112: Not a fan of using classnames as selectors but ReactSelect does not seem to offer a way of setting HTML id's on these items.

e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
```diff
@@ -0,0 +1,126 @@
...
+            should('be.visible').
+            and('have.length', 7);
+
+        // # Remove the last user from the group message's dialog input section
+        cy.get('.react-select__multi-value__remove').
```
Niceplace, line 117: Same comment as https://github.com/mattermost/mattermost-webapp/pull/6100/files#r467518166

saturninoabril:
Thanks for checking @Niceplace!

There are 2 messages there:
- "You've reached the maximum number of people for this conversation. Consider creating a private channel instead."
- "This will start a new conversation. If you're adding a lot of people, consider creating a private channel instead."

The issue with second message are it doesn't care
- if you're removing user/s in reference to the original members count, and
- if all users were removed.

I believe it's how it works currently but it's definitely needs to improve. I'll open up a UX discussion in this regard.

e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
```diff
@@ -0,0 +1,126 @@
...
+            Cypress._.times(10, createUserAndAddToTeam(testTeam));
+        });
+    });
+
+    it('MM-T463 Should not be able to create a group message with more than 8 users', () => {
+        // # Go to town-square channel
+        cy.visit(`/${testTeam.name}/channels/town-square`);
+
+        // # Open the 'Direct messages' dialog to create a new direct message
+        cy.get('#addDirectChannel').click();
```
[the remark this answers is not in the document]
Niceplace: I believe the explicit visibility check is not necessary when using  `click`. Cypress performs a series of default checks before doing the `click` action and visibility seems to be part of it: https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Actionability

What do you think, should we leave it as is ?

e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
```diff
@@ -0,0 +1,126 @@
...
+
+describe('Multi-user group messages', () => {
+    let testTeam;
+    before(() => {
+        // # Create a new team
+        cy.apiInitSetup().then(({team}) => {
+            testTeam = team;
+
+            // # Add 10 users to the team
+            Cypress._.times(10, createUserAndAddToTeam(testTeam));
```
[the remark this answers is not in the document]
Niceplace: > Also, after this, please login as test user instead of staying as sysadmin.

Noted ! I will apply this logic for other E2E tests moving forwad.

e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
```diff
@@ -0,0 +1,139 @@
...
+        // # Click on "Go" in the group message's dialog to begin the conversation
+        cy.get('#saveItems').click();
+
+        // * Check that the number of users in the group message is 8
+        cy.get('#channelMemberCountText').
+            should('be.visible').
+            and('have.text', '8');
+
+        // * Check that the direct message text input's placeholder is truncated if the user's list is too long
+        cy.get('[data-testid="post_textbox_placeholder"]').
```
Niceplace, line 59: This is the way I am ensuring the text input of direct messages does not wrap. I tried to assert against the `aria-label` attribute of the input since it contains the user list + 'Write to' message but even adding a newline in there changes nothing to the display.

Also, even though the E2E guide suggests favoring HTML ID's instead of `data-testid` attributes, this one was already existing and used in other tests so I decided to go with it.

What do you think ?

e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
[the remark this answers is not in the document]
saturninoabril: The visibility check coming from page load helps eliminate flakiness in tests (as an implicit wait strategy), primarily due to the async nature of our application - state, local storage, additional REST API call, etc - where the UI rendering/structure is based from. The target element might be ready but other parts if the UI might not. However, it's more of prevention only and might be too early to tell so I'm fine to leave as is.

e2e/cypress/integration/multi_team_and_dm/max_gm_members_spec.js
saturninoabril, line 59: I'm good as is. Can simply be
```suggestion
        cy.findByTestId('post_textbox_placeholder').
```

PR #6101, opened 2020-08-08
Cypress/E2E: Fix feature discovery spec

josephbaylon:
#### Summary
- Fix feature discovery spec; need to remove license for feature discovery to work
- Assume `mattermost-license.txt` is under fixtures folder; added to `.gitignore`
- Added `uiCheckLicenseExists` to `system.js` and `system.d.ts` under `supports/ui`
- Extracted common, saml, and system API files (`support/api`)
- Added `apiUploadLicense` and `apiDeleteLicense` to system API
- Added upload license block to `apiRequireLicense` and `apiRequireLicenseForFeature`

#### Ticket Link
None -  master only


e2e/cypress/support/api/saml.js
```diff
@@ -0,0 +1,42 @@
...
+        method: 'POST',
+        body: {saml_metadata_url: samlMetadataUrl},
+    }).then((response) => {
+        expect(response.status, 'Failed to obtain metadata from Identity Provider URL').to.equal(200);
+        return cy.wrap(response);
+    });
+});
+
+Cypress.Commands.add('apiUploadSAMLIDPCert', (filePath) => {
+    cy.apiUploadFile('certificate', filePath, {url: '/api/v4/saml/certificate/idp', method: 'POST', successStatus: 200, fromFixture: true});
```
saturninoabril, line 33: Maybe remnants in the past but could you check if `fromFixture` is actually needed?

e2e/cypress/support/index.js
```diff
@@ -121,6 +123,13 @@ before(() => {
             timezone: {automaticTimezone: '', manualTimezone: 'UTC', useAutomaticTimezone: 'false'},
         });
 
+        // # Upload license
+        cy.apiGetClientLicense().then(({license}) => {
+            if (license.IsLicensed === 'false' && resetBeforeTest) {
+                cy.apiUploadLicense(serverLicense.filePath);
+            }
+        });
```
saturninoabril, line 131: I suggest to do it if required by test only and not for every test. Good places are inside of `cy.apiRequireLicense` and `cy.apiRequireLicenseForFeature`. On those commands, make a note about the location of license file which should be available in `fixtures` folder. Make the call fix as `cy.apiUploadLicense('mattermost-license.txt')`

e2e/cypress/support/env.js
```diff
@@ -14,3 +14,9 @@ export function getDBConfig() {
         connection: Cypress.env('dbConnection'),
     };
 }
+
+export function getServerLicense() {
```
saturninoabril, line 18: Probably not needed.

e2e/cypress.json
```diff
@@ -19,6 +19,7 @@
         "runLDAPSync": true,
         "resetBeforeTest": false,
         "storybookUrl": "http://localhost:6006/",
-        "webhookBaseUrl": "http://localhost:3000"
+        "webhookBaseUrl": "http://localhost:3000",
+        "licenseFilePath": "mm-license.txt"
```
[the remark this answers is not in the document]
josephbaylon: Good to know. I just followed the assumptions of the naming and where the mm-license.txt from your repo. I'll make the changes and assume everything on fixtures folder.

e2e/cypress/support/api/saml.js
josephbaylon: I added that for file uploads that are not directly from fixtures folder. But since we're now assuming license is going to be from fixtures folder then I'll remove that condition.

e2e/cypress/support/api/system.js
```diff
@@ -105,3 +122,15 @@ Cypress.Commands.add('apiInvalidateCache', () => {
...
 });
+
+function uploadLicencseIfDoesNotExist() {
+    // # Upload a license if it does not exist.
+    cy.apiGetClientLicense().then(({license}) => {
+        if (license.IsLicensed === 'false') {
+            cy.apiUploadLicense('mattermost-license.txt');
+
+            // * Verify license exists via admin console
+            cy.uiCheckLicenseExists();
```
saturninoabril, line 133: I find this not following the approach we're trying to do with custom commands, especially for API commands where it should avoid mixing with UI calls.

josephbaylon: @saturninoabril Any suggestions on what the approach should be? I can just add a big timeout but I can't really test that it's failing because it's passing for me locally.

josephbaylon:
@saturninoabril Regarding the fixture license. That's the reason I originally added the condition resetBeforeTest so that existing dev won't have to deal with it. It only affects cypress. If dev removes the license manually then they'd just have to manually add it back how it is now.

e2e/cypress/support/api/common.js
```diff
@@ -0,0 +1,53 @@
...
+// See LICENSE.txt for license information.
+
+// *****************************************************************************
+// Common / Helper commands
+// *****************************************************************************
+
+Cypress.Commands.add('apiUploadFile', (name, filePath, options = {}) => {
+    const formData = new FormData();
+
+    cy.fixture(filePath, 'binary', {timeout: 1200000}).
```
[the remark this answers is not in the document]
josephbaylon: Done

e2e/cypress/support/api/system.js
saturninoabril: I played around it earlier and came up with [something like this](https://github.com/saturninoabril/mattermost-cypress-docker/compare/file-check#diff-de80b964a5e58206daa24f418a3dc1d8R121-R138). Then look for usage at `apiRequireLicense` and `apiRequireLicenseForFeature`.

josephbaylon: @saturninoabril I'm not sure that resolves the issue regarding the timing that license is available in the ui but I like the additional check for the file if it exists locally. Can you please pull the latest and then commit your changes? Looks your changes are spread out and would be easier for you to push the changes.

saturninoabril: It's not dealing on timing since it's returning a resolve function.
Yup, will send a follow up PR that's why I didn't request for changes.

saturninoabril: What I meant to say is I'm good with this PR and can merge once approved. Then I'll submit a separate PR for another round of review. But I can push my changes here too if you prefer it. Let me know.

josephbaylon: I'd prefer the latter so Prapti and Rohitesh can review in one go

saturninoabril: Done - https://github.com/mattermost/mattermost-webapp/pull/6101/commits/4a589db862a91617db1cb828eae4ca4baf2a3907. Please run on your end and let me know if you find anything.

josephbaylon: @saturninoabril Thanks! I retested and spot checked several tests and they're passing. Failing when I remove the license from UI as well as fixture, which is expected

PR #6102, opened 2020-08-08
MM-23346 MM-T123 Add e2e pinned parent post test

josephk96:
Summary

- Created a test that replies to a parent post, pins the parent post and verifies that the reply count exists and is correct

Ticket Link
Jira: https://mattermost.atlassian.net/browse/MM-23346
Test case: https://automation-test-cases.vercel.app/test/MM-T123


e2e/cypress/integration/messaging/pinned_parent_post_spec.js
```diff
@@ -0,0 +1,75 @@
...
+            cy.apiLogin(receiver);
+
+            // # Visit a test channel and post a message
+            cy.visit(`/${testTeam.name}/channels/${testChannel.name}`);
+            cy.postMessage('Hello');
+
+            // # Assign lastPostId variable to the id of the last post
+            cy.getLastPostId().then((postId) => {
+                lastPostId = postId;
+            });
```
saturninoabril, line 44: Probably best to move this into the test block itself.

e2e/cypress/integration/messaging/pinned_parent_post_spec.js
```diff
@@ -0,0 +1,75 @@
...
+        });
+    });
+
+    it('M23346 - Pinned parent post: reply count remains in center channel and is correct', () => {
+        // # Login as the other user
+        cy.apiLogin(sender);
+
+        // # Wait for three seconds (sometimes the element won't render if we don't wait)
+        cy.wait(TIMEOUTS.THREE_SEC);
```
saturninoabril, line 53: Instead of waiting, reload or visit again the test channel and then get the last post ID.

e2e/cypress/integration/messaging/pinned_parent_post_spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+        // # Click the reply button, and post a reply four times and close the thread rhs tab
+        cy.get(`#post_${lastPostId}`).trigger('mouseover');
+        cy.get(`#CENTER_commentIcon_${lastPostId}`).click({force: true});
+        for (let i = 0; i < 5; i++) {
+            cy.get('#reply_textbox').type(`Hello to you too ${i}`);
+            cy.get('#addCommentButton').click();
```
saturninoabril, line 60:
```suggestion
            cy.get('#reply_textbox').should('be.visible').type(`Hello to you too ${i}`);
            cy.get('#addCommentButton').should('be.enabled').click();
```

e2e/cypress/integration/messaging/pinned_parent_post_spec.js
```diff
@@ -0,0 +1,75 @@
...
+        }
+        cy.get('#rhsCloseButton').click();
+
+        // # Wait for three seconds (sometimes the element won't render if we don't wait)
+        cy.wait(TIMEOUTS.THREE_SEC);
+
+        // # Pin the post to the channel
+        cy.get(`#post_${lastPostId}`).trigger('mouseover');
+        cy.get(`#CENTER_button_${lastPostId}`).click({force: true});
+        cy.get(`#pin_post_${lastPostId}`).click({force: true});
```
saturninoabril, line 70: We have convenient command for this.
```suggestion
        // # Pin the post to the channel
        cy.getPostMenu(postId, 'Pin to Channel').click();
```

e2e/cypress/integration/messaging/pinned_parent_post_spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+        // * Assert that the reply count exists and is correct
+        cy.get(`#CENTER_commentIcon_${lastPostId}`).find('span').eq(0).find('span').eq(1).should('have.text', '5');
```
saturninoabril, line 73: Before this, could you add check for post pre-header that the post is actually pinned?
Screen Shot 2020-08-10 at 10 26 17 AM

e2e/cypress/integration/messaging/pinned_parent_post_spec.js
```diff
@@ -0,0 +1,75 @@
...
+            cy.postMessage('Hello');
+
+            // # Assign lastPostId variable to the id of the last post
+            cy.getLastPostId().then((postId) => {
+                lastPostId = postId;
+            });
+        });
+    });
+
+    it('M23346 - Pinned parent post: reply count remains in center channel and is correct', () => {
```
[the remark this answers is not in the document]
josephk96: Changed! In the future, should i name all the tests according to the test keys in the spreadsheet?

e2e/cypress/integration/messaging/pinned_parent_post_spec.js
josephk96: I moved it to the test block as you suggested. However, after I moved it to test block, the `lastPostId = postId;` variable assignment expression did not work as intended, so I had to wrap all the commands that use `postId` within the `cy.getLastPostId` function.

I'm not sure why the variable reassignment did not work, have you encountered this issue by any chance? I tried creating a test variable `a` in the global scope, reassigned that variable in the test block and console.log it and it was working as intended, but reassigning a variable to the value of the postId returned from the promise doesn't work in the test block, but it does in the`before` block.

saturninoabril: I think the assignment happened first on the test block before going into the before hook. That's why the value is getting is the initial value of undefined. But it's just my assumption as I'm getting similar issue in the past.

saturninoabril: But the one you implemented is the one I'm thinking actually. Wrapping everything inside the `cy.getLastPostId`.

e2e/cypress/integration/messaging/pinned_parent_post_spec.js
```diff
@@ -0,0 +1,67 @@
...
+                cy.get('#reply_textbox').should('be.visible').type(`Hello to you too ${i}`);
+                cy.get('#addCommentButton').should('be.enabled').click();
+            }
+            cy.get('#rhsCloseButton').click();
+
+            // # Pin the post to the channel
+            cy.getPostMenu(postId, 'Pin to Channel').click();
+
+            // # Find the 'Pinned' span in the post pre-header to verify that the post was actually pinned
+            cy.findByText('Pinned').should('exist');
```
saturninoabril, line 61: Since this is not scoped to the last post, I suggest to add the ff:
```suggestion
            cy.get(`#post_${postId}`).findByText('Pinned').should('exist');
```

e2e/cypress/integration/messaging/pinned_parent_post_spec.js
josephk96: Ah, thanks for the explanation Saturn!

PR #6117, opened 2020-08-11
MM-23357 - Add e2e test for issue MM-23357

josephk96:
Summary

Created a test where three users are added to a GM and search for a certain user by a substring of their name which is not their display name.

Ticket Link
Jira: https://mattermost.atlassian.net/browse/MM-23357
Test case: https://automation-test-cases.vercel.app/test/MM-T1226


josephk96:
This was pretty tricky because the selector looked like this `id="switchChannel_usere95217, userfefd0b"` but the Id is dynamic, so sometimes the order of the users would be switched, which would cause the test to fail. For some reason, testing library's findByText method also failed for finding the span text that contained the username of the user we searched for, so I had to select the element by class and make sure the div contained the username text of the user we searched for. Please let me know if I should refactor this :)

e2e/cypress/integration/messaging/ctrl_cmd_k_find_gm_by_matching_name_spec.js
```diff
@@ -0,0 +1,59 @@
...
+
+            // # Type either cmd+K / ctrl+K depending on OS and type in the first 7 of the second user's last name
+            cy.get('#post_textbox').cmdOrCtrlShortcut('K');
+            cy.focused().type(`${secondUser.last_name.slice(0, 7)}`);
+
+            // * The suggestion for the GM channel with the user that was searched for should show the username of the user
+            cy.get('.suggestion--selected').should('exist').and('contain.text', secondUser.username);
+
+            // * The suggestion for the GM channel with the user that was searched for should show the 'G' text
+            cy.get('#suggestionList').find('div').eq(1).findByText('G').should('exist');
```
saturninoabril, line 56: I'm good as is but would be nice to combine this to above so that it's clear that it refers to the same selected suggestion, like:
```javascript
cy.get('.suggestion--selected').should('exist').
    and('contain.text', secondUser.username).
    findByText('G').should('exist');
```

e2e/cypress/integration/messaging/ctrl_cmd_k_find_gm_by_matching_name_spec.js
```diff
@@ -0,0 +1,59 @@
...
+
+describe('Messaging', () => {
+    let testTeam;
+    let firstUser;
+    let secondUser;
+    let thirdUser;
+
+    before(() => {
+        // # Login as test user
+        cy.apiInitSetup().then(({team, user}) => {
```
josephbaylon, line 20: Just for readability, would recommend to keep names consistent,
firstUser = user1
secondUser = user2
thirdUser = user3

josephk96:
I implemented the suggested changes :)
Thanks @saturninoabril and @josephbaylon for taking the time to review this!

e2e/cypress/integration/messaging/ctrl_cmd_k_find_gm_by_matching_name_spec.js
josephk96: This is much more concise. Thanks pare!

e2e/cypress/integration/messaging/ctrl_cmd_k_find_gm_by_matching_name_spec.js
josephk96: Thanks Joseph, this does make more sense haha :)

PR #6119, opened 2020-08-11
MM-23623: Make the validation of what is considered a post delete con…

mkraft:
…sistent between backend and frontend.

#### Summary

Updates the validation of what is considered a "deleted" post so that the front-end and backend are using the same logic.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23623

#### Screenshots

For a user who does not have permissions to delete their own posts, whitespace only don't enable the save button...

Screen Shot 2020-08-11 at 2 16 07 PM

... but non-whitespaces do enable it.

Screen Shot 2020-08-11 at 2 16 15 PM

### DIFF
--- a/components/edit_post_modal/edit_post_modal.jsx
+++ b/components/edit_post_modal/edit_post_modal.jsx
@@ -353,7 +353,7 @@ class EditPostModal extends React.PureComponent {
             return !this.props.canEditPost;
         }
 
-        if (this.state.editText !== '') {
+        if (this.state.editText.trim() !== '') {
             return !this.props.canEditPost;
         }
 


mkraft:
> Hi @mkraft
> This also needs to be fixed in an another place, i.e. where user tries to edit other's post and then truncates the post and tries to save it. Can you please fix it there as well.
> 
> Steps to reproduce:
> 
> 1. Ensure user has permission to edit other's post and does not has permission to Delete Other's post.
> 2. Now go to other user's post and edit and remove all the characters and try to save.
> 3. It prompts the user to delete the post and upon save, the same issue will be encountered.
> 
> Video demonstrating this issue:
> https://community-release.mattermost.com/files/ocsp7bbwd3dommhrsx81b3trea/public?h=xBxxXVD2gzURfWStpavD6OtvOcmzSy36TPrkxdajgSk

Can you please check again to verify that you had the permissions configured the way you think. In my tests this bug doesn't reproduce.

PR #6121, opened 2020-08-12
MM-T459 E2E Test for Group Messaging: Add first user

Niceplace:
#### Summary

In the dialog to create a new direct message (or group message), ensure that the user list is always sorted alphabetically, even after selecting a first user.

- Added e2e test with cypress
- Updated util function to render username span with unique html ID, this helps with a more precise selection of the user's display name + display username in the direct message dialog
- Updated snapshot for unit test that depends on the modification of the previous step

Related test case: [MM-T459](https://automation-test-cases.vercel.app/test/MM-T459)
Related ticket: [MM-1816](https://mattermost.atlassian.net/browse/MM-18186)

### DIFF
--- a/components/more_direct_channels/__snapshots__/more_direct_channels.test.jsx.snap
+++ b/components/more_direct_channels/__snapshots__/more_direct_channels.test.jsx.snap
@@ -190,9 +190,9 @@ exports[`components/MoreDirectChannels should match renderOption snapshot 1`] = 
       className="more-modal__name"
     >
       <span
-        id="displayedUserNameundefined"
+        id="displayedUserNameusername1"
       >
-        @undefined
+        @username1
         <span
           className="light"
         >
--- a/components/more_direct_channels/__snapshots__/more_direct_channels.test.jsx.snap
+++ b/components/more_direct_channels/__snapshots__/more_direct_channels.test.jsx.snap
@@ -189,7 +189,9 @@ exports[`components/MoreDirectChannels should match renderOption snapshot 1`] = 
     <div
       className="more-modal__name"
     >
-      <span>
+      <span
+        id="displayedUserNameundefined"
+      >
         @undefined
         <span
           className="light"
--- a/e2e/cypress/integration/multi_team_and_dm/gm_add_first_user_spec.js
+++ b/e2e/cypress/integration/multi_team_and_dm/gm_add_first_user_spec.js
@@ -64,7 +64,7 @@ describe('Multi-user group messages', () => {
             and('have.length', 1);
 
         // * Expect search bar to have focus
-        cy.get('#selectItems input[type="text"]').should('have.focus');
+        cy.get('.react-select__multi-value').next().find('input').should('have.focus');
 
         // * Expect previous search term to be cleared from search bar
         cy.get('#selectItems input[type="text"]').should('have.text', '');
@@ -102,15 +102,13 @@ const expectUserListSortedAlphabetically = (filterString, excludeFilter = false)
                 expect(stringComparison, `${currentChildText} should be before ${siblingText}`).to.be.lte(0);
             }
 
-            if (filterString) {
-                excludeFilter ?
+            excludeFilter ?
 
                 // * If a user is selected, ensure remaining users list does not contain the selected user
-                    expect(currentChildText).to.not.contain(filterString) :
+                expect(currentChildText).to.not.contain(filterString) :
 
                 // * Optionally, ensure all usernames match the autocomplete input
-                    expect(currentChildText).to.contain(filterString);
-            }
+                expect(currentChildText).to.contain(filterString);
         });
 };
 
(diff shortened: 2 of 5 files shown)


e2e/cypress/integration/multi_team_and_dm/gm_add_first_user_spec.js
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+/* eslint-disable jquery/no-text */
+/* eslint-disable jquery/no-find */
+
+describe('Multi-user group messages', () => {
+    const userPrefix1 = '0aadam';
+    const userPrefix2 = '0aabadam';
+    const userPrefix3 = 'beatrice';
+    const searchTerm = '0aa';
```
Niceplace, line 10: In the CI, what dataset is originally present in the env before running the cypress tests. Is it the same data set as `make test-data` from `mattermost-server` ? Just wondering how this can affect the display order of the users that I am creating here.

e2e/cypress/integration/multi_team_and_dm/gm_add_first_user_spec.js
```diff
@@ -0,0 +1,124 @@
...
+                expectUserListSortedAlphabetically(selectedUserName, true);
+            });
+
+        // * Expect to have a remove button ('x') present next to the selected username (previous step)
+        cy.get('.react-select__multi-value__remove').
+            should('be.visible').
+            and('have.length', 1);
+
+        // * Expect search bar to have focus
+        cy.get('#selectItems input[type="text"]').should('have.focus');
```
Niceplace, line 67: I was not able to match specific assertion`to the left of the cursor` in 
> Selected username appears in the search box to the left of the cursor 

as per https://automation-test-cases.vercel.app/test/MM-T459

There might be a way to get the cursor position with the `selectionStart` property of HTMLInputElement but this ReactComponent seems to do some tricks by hiding the input and displaying stuff in a `div` instead so my initial tests were inconclusive. I was not sure if it was worth pursuing so I decided to focus on the other assertions needed for this test instead. 

Should I try to make it work anyways ?

e2e/cypress/integration/multi_team_and_dm/gm_add_first_user_spec.js
saturninoabril: Before running the entire, no data is present in the DB. We're creating all the test data as needed by the test (including sysadmin user), and gets accumulated as we run each test.

components/more_direct_channels/__snapshots__/more_direct_channels.test.jsx.snap
```diff
@@ -189,7 +189,9 @@ exports[`components/MoreDirectChannels should match renderOption snapshot 1`] =
     <div
       className="more-modal__name"
     >
-      <span>
+      <span
+        id="displayedUserNameundefined"
```
saturninoabril, line 193: Could you please update [related unit test for snapshot](https://github.com/mattermost/mattermost-webapp/blob/master/components/more_direct_channels/more_direct_channels.test.jsx#L175) so it will not reflect as undefined?  Thanks

```diff
- expect(wrapper.instance().renderOption({id: 'user_id_1', delete_at: 0}, true, jest.fn())).toMatchSnapshot();
+ expect(wrapper.instance().renderOption({id: 'user_id_1', username: 'username1', delete_at: 0}, true, jest.fn())).toMatchSnapshot();
```

e2e/cypress/integration/multi_team_and_dm/gm_add_first_user_spec.js
Niceplace: Yes ! Awesome it works

components/more_direct_channels/__snapshots__/more_direct_channels.test.jsx.snap
Niceplace: Good catch ! Thanks ! I'll be mindful of these types of issues in the future.

Niceplace:
Thanks for the feedback @saturninoabril I appreciate your advice ! :) 
I think I'm good for a second pass.

Cheers !

Niceplace:
@saturninoabril I'm afraid I do not see any additionnal comment :/

e2e/cypress/integration/multi_team_and_dm/gm_add_first_user_spec.js
```diff
@@ -0,0 +1,122 @@
...
+        // * Expect to have a remove button ('x') present next to the selected username (previous step)
+        cy.get('.react-select__multi-value__remove').
+            should('be.visible').
+            and('have.length', 1);
+
+        // * Expect search bar to have focus
+        cy.get('.react-select__multi-value').next().find('input').should('have.focus');
+
+        // * Expect previous search term to be cleared from search bar
+        cy.get('#selectItems input[type="text"]').should('have.text', '');
```
saturninoabril, line 68: Oh sorry I thought I had it. Here it is for consistency. Thanks for checking @Niceplace!
```suggestion
        cy.get('#selectItems input').should('have.text', '');
```

PR #6123, opened 2020-08-12
MM-27507: Show errors due to rate limiting while inviting

agnivade:
The exception handler wasn't really handling error responses
sent from the server. We had to explicitly check the error field
for that and update the notSent array appropriately.

https://mattermost.atlassian.net/browse/MM-27507

#### Related Pull Requests

- Has server changes (https://github.com/mattermost/mattermost-server/pull/15230)

#### Screenshots

ratelimit


agnivade:
Some QA testing would be preferred here since it's a UI affecting change.

PR #6127, opened 2020-08-13
MM-23350 Add e2e test for MM-23350

josephk96:
Summary

Created a test where the test user posts a provided Markdown text containing an external link and asset that the post contains -
1. The correct url
2. The correct image

and that the link is valid. I tried to click on the link, but Cypress doesn't allow you to open a site not within the same domain, so I followed this guide https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__tab-handling-links/cypress/integration/tab_handling_anchor_links_spec.js  and made sure that the link in the a tag, when requesting it, is valid and provides html. However, please let me know if you would like me to test the url in a different way!

Below is another resource I used that said clicking on a tags is not recommended

https://stackoverflow.com/questions/50495996/cypress-e2e-testing-how-to-get-around-cross-origin-errors

Ticket Link
Jira: https://mattermost.atlassian.net/browse/MM-23350
Test case: https://automation-test-cases.vercel.app/test/MM-T188

### DIFF
--- a/e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
+++ b/e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
@@ -1,0 +1,60 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Group: @messaging
+
+describe('Messaging', () => {
+    let testTeam;
+    let testChannel;
+    let testUser;
+
+    before(() => {
+        // # Login as test user and visit the newly created test channel
+        cy.apiInitSetup().then(({team, channel, user}) => {
+            testUser = user;
+            testTeam = team;
+            testChannel = channel;
+
+            cy.apiLogin(testUser);
+
+            // # Visit a test channel and post a message
+            cy.visit(`/${testTeam.name}/channels/${testChannel.name}`);
+        });
+    });
+
+    it('MM-T188 - Inline markdown image that is a link, opens the link', () => {
+        // # Enable 'Show markdown preview option in message input box' setting in Account Settings > Advanced
+        cy.get('.sidebar-header-dropdown__icon').click();
+        cy.findByText('Account Settings').should('be.visible').click();
+        cy.get('#advancedButton').should('be.visible').click();
+        cy.get('#advancedPreviewFeaturesEdit').should('be.visible').click();
+        cy.get('#advancedPreviewFeaturesmarkdown_preview').should('be.visible').check();
+        cy.get('#saveSetting').should('be.visible').click();
+        cy.reload();
+
+        // # Post the provided Markdown text in the test channel
+        cy.focused().type('[![Build Status](https://travis-ci.org/mattermost/platform.svg?branch=master)](https://travis-ci.org/mattermost/platform)').type('{enter}');
+
+        cy.getLastPostId().then((postId) => {
+            cy.get(`#postMessageText_${postId}`).find('a').then(($a) => {
+                // * Check that the newly created post contains an a tag with the correct href link
+                cy.wrap($a).should('have.attr', 'href', 'https://travis-ci.org/mattermost/platform');
+
+                // # Check that the newly created post has the correct AltText
+                cy.wrap($a).findByAltText('Build Status').should('exist');
+
+                // # Assign the value of the a tag href to the 'href' variable and assert the link is valid
+                // eslint-disable-next-line jquery/no-prop
+                const href = $a.prop('href');
+                cy.request(href).its('body').should('include', '</html>');
+            },
+            );
+        });
+    });
+});
--- a/e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
+++ b/e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
@@ -55,7 +55,6 @@ describe('Messaging', () => {
                     and('have.attr', 'target', '_blank');
 
                 // * Check that the newly created post has an image
-
                 cy.wrap($a).find('img').should('be.visible').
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
```diff
@@ -0,0 +1,60 @@
...
+        cy.apiInitSetup().then(({team, channel, user}) => {
+            testUser = user;
+            testTeam = team;
+            testChannel = channel;
+
+            cy.apiLogin(testUser);
+
+            // # Visit a test channel and post a message
+            cy.visit(`/${testTeam.name}/channels/${testChannel.name}`);
+        });
```
saturninoabril, line 28: Looks like declared variables are not used elsewhere. Also login to test user if no other admin function is needed after init, can simply use `loginAfter: true`.
```suggestion
        cy.apiInitSetup({loginAfter: true}).then(({team, channel, user}) => {
            // # Visit a test channel and post a message
            cy.visit(`/${team.name}/channels/${channel.name}`);
        });
```

e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+    it('MM-T188 - Inline markdown image that is a link, opens the link', () => {
+        // # Enable 'Show markdown preview option in message input box' setting in Account Settings > Advanced
+        cy.get('.sidebar-header-dropdown__icon').click();
+        cy.findByText('Account Settings').should('be.visible').click();
+        cy.get('#advancedButton').should('be.visible').click();
+        cy.get('#advancedPreviewFeaturesEdit').should('be.visible').click();
+        cy.get('#advancedPreviewFeaturesmarkdown_preview').should('be.visible').check();
+        cy.get('#saveSetting').should('be.visible').click();
+        cy.reload();
```
saturninoabril, line 39: We can do this via API using `cy.apiSaveShowMarkdownPreviewPreference()` but since we don't have current test for such UI flow, I'm fine to do it this way. Just few comments.
- should scope elements check of advance settings within `#accountSettingsModal`
- instead of `cy.reload()` which is expensive reloading the page again, just close the account settings modal.

e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+        // # Post the provided Markdown text in the test channel
+        cy.focused().type('[![Build Status](https://travis-ci.org/mattermost/platform.svg?branch=master)](https://travis-ci.org/mattermost/platform)').type('{enter}');
```
saturninoabril, line 42: This is intermittently failing (missed typing first few characters into the textbox) as it seemed needed to wait some time. I suggest to use `cy.postMessage` instead.

e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+                // # Check that the newly created post has the correct AltText
+                cy.wrap($a).findByAltText('Build Status').should('exist');
+
+                // # Assign the value of the a tag href to the 'href' variable and assert the link is valid
+                // eslint-disable-next-line jquery/no-prop
+                const href = $a.prop('href');
+                cy.request(href).its('body').should('include', '</html>');
+            },
+            );
```
saturninoabril, line 57: May put on the same line.

saturninoabril:
I'm good with the approach on link. The only thing that's missing there for me is whether there's an image actually rendered in a post. Could you please add verification on it.  Thanks @josephk96!

josephk96:
@saturninoabril , thanks for the quick review! I'll implement the changes when I get home :)

josephk96:
I implemented the changes that you requested, @saturninoabril  :) Thanks again for the review!

e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+        cy.getLastPostId().then((postId) => {
+            cy.get(`#postMessageText_${postId}`).find('a').then(($a) => {
+                // * Check that the newly created post contains an a tag with the correct href link
+                cy.wrap($a).should('have.attr', 'href', 'https://travis-ci.org/mattermost/platform');
+
+                // # Check that the newly created post has an image
+                cy.wrap($a).findByAltText('Build Status').should('have.attr', 'src').then(($src) => {
+                    expect($src).to.not.be.empty;
+                });
```
saturninoabril, line 53: Maybe we can improve this by targeting the `img` directly and specifically validating the src attribute. Also, add check to target attribute of anchor element.
```javascript
const linkUrl = 'https://travis-ci.org/mattermost/platform';
const imageUrl = 'https://travis-ci.org/mattermost/platform.svg?branch=master';
const label = 'Build Status';
cy.postMessage(`[![${label}](${imageUrl})](${linkUrl})`);

/** code in between */

// * Check that the newly created post contains an a tag with the correct href link and target
cy.wrap($a).
    should('have.attr', 'href', linkUrl).
    and('have.attr', 'target', '_blank');

// * Check that the newly created post has an image
const baseUrl = Cypress.config('baseUrl');
cy.wrap($a).find('img').should('be.visible').
    and('have.attr', 'src', `${baseUrl}/api/v4/image?url=${encodeURIComponent(imageUrl)}`).
    and('have.attr', 'alt', label);
```

josephk96:
@saturninoabril Thank you pare, the test looks more solid now

e2e/cypress/integration/messaging/inline_markdown_image_link_open_link_spec.js
```diff
@@ -0,0 +1,70 @@
...
+
+        cy.getLastPostId().then((postId) => {
+            cy.get(`#postMessageText_${postId}`).find('a').then(($a) => {
+                // * Check that the newly created post contains an a tag with the correct href link and target
+                cy.wrap($a).
+                    should('have.attr', 'href', linkUrl).
+                    and('have.attr', 'target', '_blank');
+
+                // * Check that the newly created post has an image
+
```
josephbaylon, line 58: Minor: remove extra line
```suggestion
```

josephk96:
Thanks @josephbaylon !

PR #6130, opened 2020-08-13
Cypress/E2E: Update TM4J test keys

saturninoabril:
#### Summary
Update TM4J test keys

### DIFF
--- a/e2e/cypress/integration/messaging/message_parse_spec.js
+++ b/e2e/cypress/integration/messaging/message_parse_spec.js
@@ -18,7 +18,7 @@ describe('Messaging', () => {
         });
     });
 
-    it('M17444 - correctly parses "://///" as Markdown and does not break the channel', () => {
+    it('MM-T196 Markdown correctly parses "://///" and doesn\'t break the channel', () => {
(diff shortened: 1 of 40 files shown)


e2e/cypress/integration/messaging/message_parse_spec.js
```diff
@@ -18,7 +18,7 @@ describe('Messaging', () => {
         });
     });
 
-    it('M17444 - correctly parses "://///" as Markdown and does not break the channel', () => {
+    it('MM-T196 Markdown correctly parses "://///" and doesn\'t break the channel', () => {
```
josephbaylon, line 21: I haven't tried but maybe use backtick to avoid escaping `'`
```suggestion
    it(`MM-T196 Markdown correctly parses "://///" and doesn't break the channel`, () => {
```

saturninoabril: Yup, that will work but I'm good as is.

josephbaylon: Actually lint would fail on backtick, so nevermind 😄

PR #6131, opened 2020-08-13
[MM-24919] - Add back line breaks on channel header and fix block quotes 

ghost:
#### Summary
This PR adds the breaks back in the channel header popover in order to display block quotes correctly. Since this will change the existing behaviour I need help with testing (again) in order to determine the desired behaviour. We might need to avoid using markdown here if we can't match all use cases. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24919


hmhealey:
Using this description
Screen Shot 2020-08-14 at 4 37 33 PM

---

On the test server, it appears as
Screen Shot 2020-08-14 at 4 34 36 PM
Screen Shot 2020-08-14 at 4 34 39 PM

---
And on master on my local instance, it looks like
Screen Shot 2020-08-14 at 4 35 03 PM
Screen Shot 2020-08-14 at 4 35 05 PM

---

I forget if we wanted to preserve newlines in the expanded view, and if so, are we okay that the newlines change? I know we had a "bug" reported way back when we started to respect newlines, but there was someone with a channel header set up this way (with all the newlines and pipes), so I don't know if that's a case that's really valid. I just used it here for illustration purposes.

jgilliam17:
Thank you @nevyangelova 
The issue originally reported in the linked ticket is resolved.
I can also see the issue Harrison mentioned and some glitching when the channel header text has few links, formatted text flows out, See `test4` channel on the test server and screenshot below
Screen Shot 2020-08-18 at 10 56 39 AM

nevyangelova:
@jgilliam17 @hmhealey thanks. The word wrapping should be fairly easy to fix. But the question "Do we want the line breaks?" remains. Because we cannot have both, its a prop passed to markdown, if we would like to solve what the ticket requires we need the breaks. But then we will have the visual that Harrison shared above. @esethna can advice?

esethna:
> We can remove the line breaks and fix the position of the block quote IMO, that would be the best middle ground solution.

@nevyangelova that works for me. Key here for me is that we don't want any breaking changes to how peoples channel headers would appear before vs after this change.

stylianosrigas:

nevyangelova:
PR updated @esethna @jgilliam17

esethna:
Still seeing the same issue: 
image

esethna:
@nevyangelova should we consider closing this ticket? It's not externally reported (just found via automated tests), and this component seems flakey?

nevyangelova:
> @nevyangelova should we consider closing this ticket? It's not externally reported (just found via automated tests), and this component seems flakey?

Sorry I forgot to remove the word wrapping!! We can proceed with merging as discussed previously. 😅

esethna (approved):

jgilliam17:
Thank you @nevyangelova 
The wrapping issue is fixed.
Can you check the original issue from the ticket with the quote, I am seeing the new line missing again
Screen Shot 2020-09-09 at 8 26 57 PM

jgilliam17:
@esethna 
It seems that issue reported in MM-24919 won't be fixed in order for us not to see [an issue Harrison mentioned here](https://github.com/mattermost/mattermost-webapp/pull/6131#issuecomment-674263884)
This was internally reported/found during RC and will probably get solved with common components in the near future (per discussion with Nevy).
I can take the test case out of our test cycle rotation until common component is implemented so we don't keep reopening/failing. Any thoughts?

esethna:
@jgilliam17 that's fine, especially if this was reported internally,

hmhealey (approved):

jgilliam17 (approved):
Thank you @nevyangelova 
Tested, looks good to merge.
Note: I took the test case out of our test cycle rotation until common component is implemented since the original issue reported won't be fixed by this PR.

amyblais:
@sudheerDev Kind reminder to review for v5.28.

sudheerDev (approved):

PR #6133, opened 2020-08-13
[MM-27132] Get user group mention keys for channel

fmunshi:
#### Summary
- Modify user mention keys selector to look at group mention keys associated to that specific channel
- Move group mentions permission checks out of redux as certain selectors require the user to get the list of groups even if they cannot mention (eg. highlighting)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27132

#### Related Pull Requests
- https://github.com/mattermost/mattermost-redux/pull/1214


components/create_post/index.js
```diff
@@ -93,6 +91,7 @@ function makeMapStateToProps() {
         });
         const channelMemberCountsByGroup = selectChannelMemberCountsByGroup(state, currentChannel.id);
         const currentTeamId = getCurrentTeamId(state);
+        const groupsWithAllowReference = useGroupMentions ? new Map() : getAssociatedGroupsForReferenceByMention(state, currentTeamId, currentChannel.id);
```
catalintomai, line 94: Farhan, can you help me understand this? If we do not have `groupMentions` set, then we call `getAssociatedGroupsForReferenceByMention`?

fmunshi: This was incorrect... good catch @catalintomai 😅

components/create_comment/index.js
```diff
@@ -76,6 +74,7 @@ function makeMapStateToProps() {
             permission: Permissions.USE_GROUP_MENTIONS,
         });
         const channelMemberCountsByGroup = selectChannelMemberCountsByGroup(state, ownProps.channelId);
+        const groupsWithAllowReference = useGroupMentions ? getAssociatedGroupsForReferenceByMention(state, channel.team_id, channel.id) : new Map();
```
hmhealey, line 77: We can't pass in a new object when we're not using group mentions here or else CreateComment will constantly re-render unnecessarily. It did that before with the previous code as well, but we should fix it while we're in here. Instead, we should pass in `null` and handle it in the component.

components/create_post/index.js
```diff
@@ -93,6 +91,7 @@ function makeMapStateToProps() {
         });
         const channelMemberCountsByGroup = selectChannelMemberCountsByGroup(state, currentChannel.id);
         const currentTeamId = getCurrentTeamId(state);
+        const groupsWithAllowReference = useGroupMentions ? getAssociatedGroupsForReferenceByMention(state, currentTeamId, currentChannel.id) : new Map();
```
hmhealey, line 94: Same here

components/textbox/index.ts
```diff
@@ -30,12 +32,20 @@ const makeMapStateToProps = () => {
 
     return (state: GlobalState, ownProps: Props) => {
         const teamId = getCurrentTeamId(state);
+        const license = getLicense(state);
+        const useGroupMentions = license?.IsLicensed === 'true' && license?.LDAPGroups === 'true' && haveIChannelPermission(state, {
+            channel: ownProps.channelId,
+            team: teamId,
+            permission: Permissions.USE_GROUP_MENTIONS,
+        });
+        const autocompleteGroups = useGroupMentions ? getAssociatedGroupsForReference(state, teamId, ownProps.channelId) : [];
```
hmhealey, line 41: Same here. You can also make a selector to return a memoized copy of [] instead

components/create_comment/index.js
mkraft: Updated. Thanks.

components/create_post/index.js
mkraft: Updated.

components/textbox/index.ts
mkraft: Updated.

mkraft:
Taking over this PR for @fmunshi while he's on PTO.

fmunshi:
@srkgupta I have fixed the issue mentioned above.

amyblais:
Rohitesh is off this week so I've swapped the QA reviewer to Ogi,

fmunshi:
Hi @ogi-m, In order to test this you will need a user that belongs to two different ldap groups (both with allow_reference set to true), and a group synced channel that only has _one_ of those two groups synced to it. 

Once you have that the expected scenario can be reproduced, when the synced group is mentioned it should be highlighted and when the non synced group is @ mentioned it should just appear as a blue keyword and not have any highlight

ogi-m (approved):
Tested and passed - outside group is highlighted in blue 👍

PR #6137, opened 2020-08-14
MM-T453 E2E tests to validate multiple scenarios of redirection with direct messages

Niceplace:
#### Summary
 Add e2e tests to validate redirection scenarios with direct messages:

- Closing a direct message should redirect to last viewed channel
- Closing a direct message when vieweing another direct message should remain on the present direct message
- When in a direct message, if the URL is re-written to the root and launched, it should redirect to the direct message

#### Ticket Link

Related test case: https://automation-test-cases.vercel.app/test/MM-T453

### DIFF
--- a/e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
+++ b/e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
@@ -37,7 +37,7 @@ describe('Direct messages: redirections', () => {
         cy.visit(offTopicChannelUrl);
     });
 
-    it('MM-T453 Closing a direct message should redirect to last viewed channel', () => {
+    it('MM-T453_1 Closing a direct message should redirect to last viewed channel', () => {
         // # From the 'Direct Messages' menu, find a specific user and send 'hi'
         sendDirectMessageToUser(firstDMUser, 'hi');
 
@@ -60,7 +60,7 @@ describe('Direct messages: redirections', () => {
         expectActiveChannelToBe('Off-Topic', offTopicChannelUrl);
     });
 
-    it('MM-T453 Closing a different direct message should not affect active direct message', () => {
+    it('MM-T453_2 Closing a different direct message should not affect active direct message', () => {
         // # Send a direct message to a first user
         sendDirectMessageToUser(firstDMUser, 'hi first');
 
@@ -74,7 +74,7 @@ describe('Direct messages: redirections', () => {
         expectActiveChannelToBe(secondDMUser.username, `/messages/@${secondDMUser.username}`);
     });
 
-    it('MM-T453 Changing URL to root url when viewing a direct message should redirect to direct message', () => {
+    it('MM-T453_3 Changing URL to root url when viewing a direct message should redirect to direct message', () => {
         // # Send a direct message to a first user
         sendDirectMessageToUser(firstDMUser, 'hi');
 
--- a/e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
+++ b/e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
@@ -26,10 +26,10 @@ describe('Direct messages: redirections', () => {
                 secondDMUser = createdUser;
                 cy.apiAddUserToTeam(testTeam.id, secondDMUser.id);
             });
-        });
 
-        // # Login as test user
-        cy.apiLogin(testUser);
+            // # Login as test user
+            cy.apiLogin(testUser);
+        });
     });
 
     beforeEach(() => {
--- a/e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
+++ b/e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
@@ -27,15 +27,17 @@ describe('Direct messages: redirections', () => {
                 cy.apiAddUserToTeam(testTeam.id, secondDMUser.id);
             });
         });
+
+        // # Login as test user
+        cy.apiLogin(testUser);
+    });
+
+    beforeEach(() => {
+        // # View 'off-topic' channel
+        cy.visit(offTopicChannelUrl);
     });
 
     it('MM-T453 Closing a direct message should redirect to last viewed channel', () => {
-        // # Login as test user
-        cy.apiLogin(testUser);
-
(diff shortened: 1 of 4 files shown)


e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
```diff
@@ -0,0 +1,154 @@
...
+    cy.get('#selectItems input').should('be.enabled').type(`@${user.username}`, {force: true});
+
+    // * Expect user count in the list to be 1
+    cy.get('#multiSelectList').
+        should('be.visible').
+        children().
+        should('have.length', 1);
+
+    // # Select first user in the list
+    cy.get('body').
```
Niceplace, line 123: I have to use the keyboard here because everytime I try to click on the item, even after asserting its visibility, it fails because Cypress thinks it's detached from the DOM

e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
```diff
@@ -0,0 +1,154 @@
...
+    // * Expect url to match url passed in argument
+    cy.url().should('contain', url);
+};
+
+const sendDirectMessageToUser = (user, message) => {
+    // # Open a new direct message with firstDMUser
+    cy.get('#addDirectChannel').click();
+
+    // # Type username
+    cy.get('#selectItems input').should('be.enabled').type(`@${user.username}`, {force: true});
```
Niceplace, line 114: When typing, the input suddenly seems to lose focus after a few characters and becomes overlapped by another element. Using force fixes this issue.

e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
```diff
@@ -0,0 +1,154 @@
...
+        const channels = response.body;
+
+        // Get the name of the channel to build the CSS selector for that specific DM link in the sidebar
+        const channelDmWithFirstUser = channels.find((channel) =>
+            channel.type === 'D' && channel.name.includes(recipient.id),
+        );
+
+        // # Close the DM via 'x' button next to username in direct message list
+        cy.get(`#sidebarItem_${channelDmWithFirstUser.name} .btn-close`).
+            click({force: true});
```
Niceplace, line 152: I have to use force here because the 'x' is hidden (via CSS), I tried `invoke('show').click()` as suggested in some github threads but had mixed result. This seems to work everytime.

e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
```diff
@@ -0,0 +1,154 @@
...
+            });
+        });
+    });
+
+    it('MM-T453 Closing a direct message should redirect to last viewed channel', () => {
+        // # Login as test user
+        cy.apiLogin(testUser);
+
+        // # View 'off-topic' channel
+        cy.visit(offTopicChannelUrl);
```
saturninoabril, line 37: Same here. Could be move to `beforeEach`?

e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
```diff
@@ -0,0 +1,154 @@
...
+        sendDirectMessageToUser(secondDMUser, 'hi second');
+
+        // # Close the direct message previously opened with the first user
+        closeDirectMessageViaXButton(testUser, firstDMUser, testTeam);
+
+        // * Expect channel title and url to secondDMUser's username
+        expectActiveChannelToBe(secondDMUser.username, `/messages/@${secondDMUser.username}`);
+    });
+
+    it('MM-T453 Changing URL to root url when viewing a direct message should redirect to direct message', () => {
```
saturninoabril, line 81:
```suggestion
    it('MM-T453_3 Changing URL to root url when viewing a direct message should redirect to direct message', () => {
```

e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
```diff
@@ -0,0 +1,154 @@
...
+
+            // # Create a second test user
+            cy.apiCreateUser().then(({user: createdUser}) => {
+                secondDMUser = createdUser;
+                cy.apiAddUserToTeam(testTeam.id, secondDMUser.id);
+            });
+        });
+    });
+
+    it('MM-T453 Closing a direct message should redirect to last viewed channel', () => {
```
saturninoabril, line 32: Looks like this is combination of step 1 and step 2 which I think makes sense. I'll revise the test case. Not documented yet but in case of multiple steps, we need to change test keys into `[test_case_id]_[step_number]`.
```suggestion
    it('MM-T453_1 Closing a direct message should redirect to last viewed channel', () => {
```

e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
```diff
@@ -0,0 +1,154 @@
...
+        cy.get('#channelHeaderDropdownButton').click();
+
+        // # Click on 'Close direct message' menu item
+        cy.get('#channelCloseMessage').click();
+
+        // * Expect to be redirected to off-topic channel, check channel title and url
+        expectActiveChannelToBe('Off-Topic', offTopicChannelUrl);
+    });
+
+    it('MM-T453 Closing a different direct message should not affect active direct message', () => {
```
saturninoabril, line 61:
```suggestion
    it('MM-T453_2 Closing a different direct message should not affect active direct message', () => {
```

e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
saturninoabril: Thanks, looks good to me.

e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js
josephbaylon: I agree. We should put this block to beforeEach and remove from individual tests, unless we have other test cases to be added that won't comply to the pattern, but we can refactor later.

Niceplace:
Changes applied, tested locally and all seems :heavy_check_mark: ! Thanks for the feedback :)

PR #6143, opened 2020-08-14
[MM-26472] Invite Members step

devinbinnie:
#### Summary
Invite Members step for the Cloud Onboarding flow.

NOTES:
- The new `multi_input` component is not ready for general use, so I've added a TODO to finish it at a later date. It works well for the current flow.
- Erroneous emails will show the red border as soon as they are added instead of after you click 'Send'
- Our current `isEmail` function recognizes emails of the format: `name@host`, and doesn't require the TLD, since you could technically have an email on a local domain. We can change this if needed since in Cloud this won't occur. There's actually an email regex here that i'd like to try: https://emailregex.com/

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26472


components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
```diff
@@ -0,0 +1,288 @@
...
+                <div className='NextStepsView__wizardButtons'>
+                    {/* <button
+                        className='NextStepsView__button cancel'
+                        onClick={this.onSkip}
+                    >
+                        <FormattedMessage
+                            id='next_steps_view.skipForNow'
+                            defaultMessage='Skip for now'
+                        />
+                    </button> */}
```
jwilander, line 274: Meant to be commented? Can we remove it?

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
```diff
@@ -0,0 +1,288 @@
...
+        textField.innerText = this.getInviteURL();
+        textField.style.position = 'fixed';
+        textField.style.opacity = '0';
+
+        document.body.appendChild(textField);
+        textField.select();
+
+        try {
+            this.setState({copiedLink: document.execCommand('copy')});
+        } catch (err) {
```
nevyangelova, line 156: Do we need to provide user feedback if there is an error?

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
```diff
@@ -0,0 +1,288 @@
...
+        if (this.state.emailInput === value) {
+            return;
+        }
+
+        if (this.state.emails.length >= 10) {
+            this.setState({emailError: Utils.localizeMessage('next_steps_view.invite_members_step.tooManyEmails', 'Invitations are limited to 10 email addresses.')});
+            return;
+        }
+
+        if (value.indexOf(' ') !== -1 || value.indexOf(',') !== -1) {
```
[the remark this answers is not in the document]
devinbinnie: I'm not sure what you mean, can you elaborate?

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
devinbinnie: I think we're okay here, there is user feedback when the link is successfully copied, which is how most sites seem to do it.

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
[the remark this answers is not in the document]
devinbinnie: We do a `split` on the value of the textbox, so if there's no space it should just take the value of the textbox.

PR #6152, opened 2020-08-15
[MM-T2370] Default images to collapsed

djanda97:
#### Summary

Adds an E2E test that switches the account settings for the test user to have images collapsed by default. The test checks that the posted image remains collapsed even after reloading the page.

#### Ticket Link

MM-T2370

### DIFF
--- a/e2e/cypress/integration/scroll/default_images_collapsed_spec.js
+++ b/e2e/cypress/integration/scroll/default_images_collapsed_spec.js
@@ -1,0 +1,43 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+describe('Scroll', () => {
+    let testTeam;
+
+    before(() => {
+        // # Create new team and new user and visit Town Square channel
+        cy.apiInitSetup({loginAfter: true}).then(({team}) => {
+            // # Switch the account settings for the test user to have images collapsed by default
+            cy.apiSaveCollapsePreviewsPreference('true');
+
+            testTeam = team;
+            cy.visit(`/${testTeam.name}/channels/town-square`);
+        });
+    });
+
+    it('MM-T2370 Default images to collapsed', () => {
+        const filename = 'huge-image.jpg';
+
+        // # Post an image in center channel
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/scroll/default_images_collapsed_spec.js
```diff
@@ -0,0 +1,43 @@
...
+    let testTeam;
+
+    before(() => {
+        // # Create new team and new user and visit Town Square channel
+        cy.apiInitSetup({loginAfter: true}).then(({team}) => {
+            // # Switch the account settings for the test user to have images collapsed by default
+            cy.apiSaveCollapsePreviewsPreference('true');
+
+            testTeam = team;
+            cy.visit(`/${testTeam.name}/channels/town-square`);
```
saturninoabril, line 20: I think no need to declare `testTeam` since it's only used here where `team` can be used directly.
```suggestion
            cy.visit(`/${team.name}/channels/town-square`);
```

saturninoabril: Though not mentioned, the test assumed several posts before the collapsed image. I'll update the test case but could you please add below after visiting town-square channel,
```javascript
Cypress._.times(10, (index) => cy.postMessage(index));
```

e2e/cypress/integration/scroll/default_images_collapsed_spec.js
```diff
@@ -0,0 +1,43 @@
...
+
+        cy.get('.post-image').should('be.visible');
+
+        cy.postMessage('{enter}');
+
+        // * Observe image preview is collapsed
+        cy.findByLabelText('file thumbnail').should('not.exist');
+
+        // # Refresh the browser
+        cy.reload();
```
saturninoabril, line 38: "Scroll pop" might be a term we used and understood internally but it refers to a sudden change of height between initial load and final render of an element (or post in this case) causing an observed jank or "pop up") in the UI.
With that, we should compare the rendered height between before and after reload. Open to suggestion but I'm good as suggested below even if it's not that reliable due to some Cypress internals in targeting the element. At least a good starting point moving forward. 
```suggestion
        // # Save height of the last post
        cy.getLastPost().then((lastPost) => {
            cy.wrap(parseInt(lastPost[0].clientHeight, 10)).as('lastPostHeight');
        });

        // # Refresh the browser
        cy.reload();

        // * Verify that the last post is still the same after reload
        cy.getLastPost().then((lastPost) => {
            cy.get('@lastPostHeight').then((lastPostHeight) => {
                expect(parseInt(lastPost[0].clientHeight, 10)).to.equal(lastPostHeight);
            });
        });
```

PR #6161, opened 2020-08-16
MM-T580 : Custom slash command auto-complete displays trigger word and not command name, MM-T2834 Slash command help stays visible for system slash command

M-ZubairAhmed:
#### Summary
- MM-T580 : Wrote the test to check if created slash commands autocomplete can show trigger word instead of command title
- MM-T2834 : 

#### Ticket Link
- MM-T580
- MM-T2834

### DIFF
--- a/e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
+++ b/e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
@@ -244,7 +244,7 @@ describe('I18456 Built-in slash commands: common', () => {
         });
     });
 
-    it.only('MM-T2834 Slash command help stays visible for system slash command', () => {
+    it('MM-T2834 Slash command help stays visible for system slash command', () => {
         // # Login as user 1 and visit default channel
         loginAndVisitDefaultChannel(user1, testChannelUrl);
 
--- a/e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
+++ b/e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
@@ -243,6 +243,24 @@ describe('I18456 Built-in slash commands: common', () => {
                 });
         });
     });
+
+    it.only('MM-T2834 Slash command help stays visible for system slash command', () => {
+        // # Login as user 1 and visit default channel
+        loginAndVisitDefaultChannel(user1, testChannelUrl);
+
+        // # Type the rename slash command in textbox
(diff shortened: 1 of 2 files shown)


e2e/cypress/integration/integrations/integrations_spec.js
```diff
@@ -199,6 +200,93 @@ describe('Integrations page', () => {
...
+        // # Enter a request url for custom slash command
+        cy.findByLabelText('Request URL').should('exist').scrollIntoView().type('https://example.com');
+
+        // # Check the option of autocomplete
+        cy.findByLabelText('Autocomplete').should('exist').scrollIntoView().click();
+
+        // # Hit save to save the custom slash command
+        cy.findByText('Save').should('exist').scrollIntoView().click();
+
+        // * Verify we are at setup successfull URL
```
saturninoabril, line 252:
```suggestion
        // * Verify we are at setup successful URL
```

e2e/cypress/integration/integrations/builtin_commands/common_commands_spec.js
```diff
@@ -243,6 +243,24 @@ describe('I18456 Built-in slash commands: common', () => {
...
+
+    it('MM-T2834 Slash command help stays visible for system slash command', () => {
+        // # Login as user 1 and visit default channel
+        loginAndVisitDefaultChannel(user1, testChannelUrl);
+
+        // # Type the rename slash command in textbox
+        cy.get('#post_textbox', {timeout: TIMEOUTS.HALF_MIN}).should('be.visible').clear().type('/rename ');
+
+        // # Scan inside of suggestion list
+        cy.get('#suggestionList').should('exist').and('be.visible').within(() => {
```
josephbaylon, line 255: Minor: For positive assertion chains, `exist` is redundant and can be removed if `be.visible` is present. Also prefer `be.visible` over `exist` on the other verifications.
```suggestion
        cy.get('#suggestionList').should('be.visible').within(() => {
```

M-ZubairAhmed: Hi,

I believe both should be present as `exist` determines if it exists in the dom and `visible` checks if it visible to user and not scrolled hidden. Hence i have written both the assertions.

C.C @saturninoabril

josephbaylon: Hi @M-ZubairAhmed , I think you can infer implicit existence on visibility (positive assertion), i.e. an element cannot be visible if it doesn't exist. Existence check is required on negative assertions though. It's not a big deal, I'll go ahead and approve the PR. Thanks for the contribution!!

PR #6170, opened 2020-08-17
MM-26730 Don't increment mention count for unloaded channels

hmhealey:
This bug was caused by a race condition where the NEW_POST_RECEIVED websocket event for the "You have been added to the channel" post would sometimes come in while we were still handling the USER_ADDED (to channel) event. When this happened, we wouldn't have the ChannelMember loaded by the time we called `handleNewPost`, so we'd make a request to fetch that (line 41 of `actions/post_actions.jsx` in `handleNewPost`) and then we'd proceed to increment the mention count (line 106 of `actions/post_utils.js` in `setChannelReadAndViewed`).

The problem with that is that the newly fetched ChannelMember would have the correct mention count for the channel, and then we'd proceed to increment it again, throwing it off by one. This fixes that bug by skipping the incrementing if we just fetched the channel member object fresh from the server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26730


amyblais:
@sudheerDev Kind reminder to review for v5.27,

sudheerDev (approved):
LGTM

PR #6182, opened 2020-08-18
[MM-27907] Fixed input focus colour and fixed placeholder focus on multi_input

devinbinnie:
#### Summary
This PR fixes a couple issues with the new inputs:
- Fixed the text input so that the legend doesn't use the focus colour when not focused.
- Fixed the `multi_input` so that the placeholder shows up when no values are present on blur.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27907

### DIFF
--- a/components/input.css
+++ b/components/input.css
@@ -79,6 +79,10 @@
     color: var(--button-bg);
 }
 
+.Input_fieldset:focus-within .Input_legend {
+    color: var(--button-bg);
+}
+
 .Input_fieldset___legend {
 }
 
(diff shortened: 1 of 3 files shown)


components/input.css
```diff
@@ -109,15 +113,15 @@
     padding: 0 4px;
     white-space: nowrap;
     border: none;
+    color: rgba(var(--center-channel-color-rgb), 0.64);
```
marianunez, line 116: Here could we just use `var(--center-channel-color-64)`?

devinbinnie: We are moving away from using the static variables provided by `applyTheme`, this way we don't need to have a different CSS variable every time we want a slightly lighter/darker version of one of the theme colours.

marianunez: Good to know. Thanks!

matthewbirtch:
Also, the focus border on buttons should be on the inside, rather than the outside

image

I think you should be able to use `inset` on the shadow definition

`box-shadow: 0 0 0 2px inset var(--sidebar-text-active-border)`

PR #6189, opened 2020-08-18
[MM-27159] Button States for Cloud Onboarding

devinbinnie:
#### Summary
This PR adds the missing hover, active, focus and disabled states for all buttons used by the Cloud Onboarding flow.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27159


matthewbirtch:
Can we make the focus border on buttons on the inside rather than the outside?
image

I think you should be able to use the `inset` setting on the shadow definition

`box-shadow: 0 0 0 2px inset var(--sidebar-text-active-border)`

devinbinnie:
> Can we make the focus border on buttons on the inside rather than the outside?
> 
> I think you should be able to use the `inset` setting on the shadow definition
> 
> `box-shadow: 0 0 0 2px inset var(--sidebar-text-active-border)`

Done!

PR #6191, opened 2020-08-18
MM-27202 Add channels created by other sessions to the new sidebar

hmhealey:
The new sidebar requires that channels are explicitly added to it, and I previously forgot to do that in the case of the channel_created websocket event.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27202


e2e/cypress/integration/websocket/channel_created/new_sidebar_spec.js
```diff
@@ -0,0 +1,41 @@
...
+        // # Start with a new team
+        const teamName = `team-${getRandomId()}`;
+        cy.createNewTeam(teamName, teamName);
+
+        // * Verify that we've switched to the new team
+        cy.get('#headerTeamName').should('be.visible').should('contain', teamName);
+
+        // # Create a new channel from another session
+        const channelName = `channel-${getRandomId()}`;
+        cy.getCurrentTeamId().then((currentTeamId) => {
```
ali-farooq0, line 34: would it be worth verifying `#sidebarItem_${channelName}` does _not_ exist before creating it? I know chances are low that the channel would already exist, but an extra assertion of it not existing wouldn't hurt.

e2e/cypress/integration/websocket/channel_created/old_sidebar_spec.js
```diff
@@ -0,0 +1,41 @@
...
+        cy.get('#headerTeamName').should('be.visible').should('contain', teamName);
+
+        // # Create a new channel from another session
+        const channelName = `channel-${getRandomId()}`;
+        cy.getCurrentTeamId().then((currentTeamId) => {
+            cy.apiCreateChannel(currentTeamId, channelName, channelName, 'O', '', '', false);
+        });
+
+        // Verify that the new channel is in the sidebar
+        cy.get(`#sidebarItem_${channelName}`).should('be.visible');
```
ali-farooq0, line 39: same comment as above.

e2e/cypress/integration/websocket/channel_created/new_sidebar_spec.js
hmhealey: I don't think that should be necessary in this case since we've created a new team for each of these tests, so there won't be any channels in the team to begin with.

ali-farooq0: Ah true good point.

PR #6197, opened 2020-08-19
MM-27899 Show loader only when loading posts

sudheerDev:
#### Summary
   * Hide the loader when posts are not loading to improve performance for linux distros

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27889


sass/components/_post.scss
```diff
@@ -652,11 +652,8 @@
     }
     .loading-screen {
         position: relative;
-        height: 40px;
+        height: calc(40px + 1em);
```
sudheerDev, line 655: I needed to add 1em to not cause a jump when hiding the loader, but I didn't need to conditionally add it 🤷‍♂️

components/post_view/post_list_row/post_list_row.jsx
```diff
@@ -105,12 +119,16 @@ export default class PostListRow extends React.PureComponent {
             );
         }
 
-        if (listId === PostListRowListIds.OLDER_MESSAGES_LOADER || listId === PostListRowListIds.NEWER_MESSAGES_LOADER) {
+        const isOlderMessagesLoader = listId === OLDER_MESSAGES_LOADER;
+        const isNewerMessagesLoader = listId === NEWER_MESSAGES_LOADER;
+        if (isOlderMessagesLoader || isNewerMessagesLoader) {
+            const shouldHideAnimation = (isOlderMessagesLoader && !loadingOlderPosts) || (isNewerMessagesLoader && !loadingNewerPosts);
```
Willyfrog, line 125: you could simplify this as you already checked it should show the three dots
```suggestion
            const shouldHideAnimation = !loadingOlderPosts && !loadingNewerPosts;
```

Willyfrog (commented):
initially I'd say that it looks good (I made a minor comment to reduce code, non blocking), but when using the test server it didn't feel right, seems like sometimes we don't get to see the animation (it might have already fetched and hasn't yet displayed the info?)
I'm not really sure if it is a subjective appreciation or that's actually happening, can you check if you feel it different from community?

sudheerDev:
@Willyfrog seems to be fine for me
Aug-19-2020 17-31-21

Willyfrog:
then it probably was something acting on my end :)

Willyfrog (approved):

PR #6199, opened 2020-08-19
[MM-22161] - Improve accessibility of channel switcher

nevyangelova:
#### Summary
Improved accessibility for channel switcher, added description for modal and actions as well as listing the channels provided.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22161


components/suggestion/switch_channel_provider.jsx
```diff
@@ -150,7 +150,10 @@ class SwitchChannelSuggestion extends Suggestion {
                 {...Suggestion.baseProps}
             >
                 {icon}
-                <span>{displayName}</span>
+                <span
+                    aria-label={displayName}
```
hmhealey, line 154: Shouldn't the contents of the span be read instead of needing a label here?

components/suggestion/switch_channel_provider.jsx
```diff
@@ -150,7 +150,10 @@ class SwitchChannelSuggestion extends Suggestion {
...
+                    role='img'
```
hmhealey, line 155: Why set the role to `img`?

components/suggestion/switch_channel_provider.jsx
hmhealey: Maybe we could look at changing that instead? I think using the visible data for accessibility is preferred over adding the aria attributes whenever possible.

jgilliam17:
Thank you @nevyangelova 
Tested using macOS accessibility VoiceOver 
- Verified that dialog opens and indicates that it’s a dialog, text field, and reads the help text: "Type to find a channel. Use UP/DOWN to browse, ENTER to select, ESC to dismiss."

Screen Shot 2020-08-26 at 2 15 29 PM

- Verified using up/down arrows I can navigate through results and read channel names

I am seeing few smaller issues:
- Text was changed from "Type to find a channel. Use 🔼 🔽 to browse,ENTER to select, ESC to dismiss." to "Type to find a channel. Use UP/DOWN to browse,ENTER to select, ESC to dismiss."
Screen Shot 2020-08-26 at 3 29 59 PM
Screen Shot 2020-08-26 at 3 31 27 PM

- Unable to scroll up/down and read Unread channel names that are present when the modal is first open
Screen Shot 2020-08-26 at 3 24 30 PM

- When typing a search word e.g. `test`, help text (see below) is read again before top search result is read
Screen Shot 2020-08-26 at 2 33 13 PM

- VoiceOver is unable to read channels from the bottom of the list, the ones user does not belongs to (under `other channels')
Screen Shot 2020-08-26 at 3 25 10 PM

jgilliam17:
Correction: 
Out of scope - Verified points  2. and 4. on community, release and daily and it works the same as on the PR test server. Channel names in Unread Channels and Other Channels sections are not read by the VoiceOver when using up/down arrows.
@esethna Should we fix here as part of this PRI or create a new issue for follow up? Thanks.

components/suggestion/switch_channel_provider.jsx
devinbinnie: @nevyangelova Where is the `aria-hidden` prop being added? I can't find it myself.

esethna:
@jgilliam17 we should fix this as part of this PR. Can you also help text on JAWS Windows? Please ask Linda for a license if you don't have it. Thanks.

components/suggestion/switch_channel_provider.jsx
nevyangelova: Correction, I am not able to find a hidden prop either so in this case I cannot debug why its not working without adding the label.

nevyangelova:
Fixed, @esethna @jgilliam17 

1. If I don't update the text to 'up/down' the screen reader won't read the symbols, I checked on mobile web and the same thing has been done there so it will be read out.

2. You probably need to wait for the screen reader to read the hint first. Fixed.

3. This is default browser behaviour for Chrome and its not repeated in Firefox for example. If you check Google or Stack overflow or any other site with a search box you will see the same behaviour.

4. Fixed.

jgilliam17:
Thank you @nevyangelova 
Using Jaws on Windows
- Verified that dialog opens and indicates that it’s a dialog, text field, and reads the help text: "Type to find a channel. Use UP/DOWN to browse, ENTER to select, ESC to dismiss."
- Verified using up/down arrows I can navigate through results and read channel names, including unread, and other results.

I am seeing a new issues
- Using arrows does not scroll the list results up, shows red outline outside the modal
- Hitting Enter does not select the channel

jgilliam17:
@nevyangelova  I retested on daily for comparison 
> * Using arrows does not scroll the list results up, shows red outline outside the modal

^ This is out of scope, works this way on daily. We can create a new issue for follow up.

> * Hitting Enter does not select the channel

This is a regression on this PR, on daily, hitting Enter selects the channel you navigated to using arrows

jgilliam17:
Thank you @nevyangelova 
Tested, looks good to merge.
- Verified dialog help text and up/down arrow navigation, Enter and ESC selection works as expected.

jgilliam17 (approved):

amyblais:
This is in progress and will miss code complete, so I'll push it to v5.29 for now. This is not a regression.

nevyangelova:
@amyblais its not it just needs reviews

components/suggestion/switch_channel_provider.jsx
```diff
@@ -142,15 +142,17 @@ class SwitchChannelSuggestion extends Suggestion {
                 onClick={this.handleClick}
                 onMouseMove={this.handleMouseMove}
                 className={className}
+                role='list'
```
hmhealey, line 145: Perhaps this should be the `listitem` instead of the span within it? Does that make this get read properly?

components/suggestion/switch_channel_provider.jsx
```diff
@@ -142,15 +142,17 @@ class SwitchChannelSuggestion extends Suggestion {
...
                 ref={(node) => {
                     this.node = node;
                 }}
                 id={`switchChannel_${channel.name}`}
                 data-testid={channel.name}
+                aria-label={channel.name}
```
hmhealey, line 151: Shouldn't this be the display name?

hmhealey (changes requested):

components/suggestion/switch_channel_provider.jsx
nevyangelova: Otherwise it doesn't read the other channels

components/suggestion/switch_channel_provider.jsx
nevyangelova: same here

hmhealey: Oh, they don't have their `displayName` set because that's injected into the "My Channels" by those selectors I hate 😛. Maybe we could do `displayName || channel.name` here instead?

components/suggestion/switch_channel_provider.jsx
hmhealey: Maybe `role='list'` should be set on the `SuggestionList` instead since we're not using a proper list element?

hmhealey (approved):

PR #6209, opened 2020-08-19
Reset after patch

hahmadia:
#### Summary

Reset `hasAllowReferenceChanged` and `hasGroupMentionNameChanged` to false after a successful patch. Otherwise, keep it at the value it is.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27738

### DIFF
--- a/components/admin_console/group_settings/group_details/group_details.jsx
+++ b/components/admin_console/group_settings/group_details/group_details.jsx
@@ -372,7 +372,7 @@ export default class GroupDetails extends React.PureComponent {
             });
         }
 
-        return true;
+        return serverError ? false : true;
     };
 
     handleRolesToUpdate = async () => {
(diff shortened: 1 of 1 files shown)


srkgupta (changes requested):
Hi @hahmadia 

The issue is partially fixed now. Can you please check and fix this. 

Here's what does not works:
1. Link the Group for first time and then click on Edit. 
2. Now enable the Group Mentions and save. 
3. Now try to add a team/channel and then save again. 
Notice the same invalid error is displayed. 

Here's when this fix works:
1. Click on Edit for a Group for which Group mention is already enabled. 
2. Add a Team/Channel and save. 
Notice: There are no errors displayed this time after this PR.

srkgupta (approved):
Sorry @hahmadia. For some reason it did not work fine on my local instance previously. I tested this again on the cloud test instance and the issue is working fine there. The issue is fixed and is working fine. Approving the PR.

PR #6211, opened 2020-08-19
[MM-27535] Error states in invite flow when inviting more users than allowed

nickmisasi:
#### Summary
In Mattermost Cloud we will be enforcing user limits in the free tier. This PR adds some functionality to the invite flow for **admins only**
If you attempt to invite more users than you have left in your free tier, you will be given a warning:

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27535

#### Related Pull Requests
- Has server changes https://github.com/mattermost/mattermost-server/pull/15197
- Has redux changes https://github.com/mattermost/mattermost-redux/pull/1217

#### Screenshots
image


components/invitation_modal/invitation_modal_members_step.jsx
```diff
@@ -118,6 +124,27 @@ class InvitationModalMembersStep extends React.PureComponent {
         this.props.onSubmit(users, emails, this.state.usersInputValue);
     }
 
+    shouldShowPickerError = () => {
+        const state = getState();
+
+        const userLimit = getConfig(state).ExperimentalCloudUserLimit;
```
devinbinnie, line 130: This needs to be passed in as a prop to the component in order to ensure the component re-renders if the config were to change. You'll want to wrap this component in a redux connector and pass your config item down through there.

components/invitation_modal/invitation_modal_members_step.jsx
```diff
@@ -118,6 +124,27 @@ class InvitationModalMembersStep extends React.PureComponent {
...
+        const currentUsers = state.entities.admin.analytics.TOTAL_USERS;
+        const userIsAdmin = isAdmin(getMyTeamMember(state, this.props.currentTeamId).roles);
```
devinbinnie, line 132: Same with this guy here (`getMyTeamMember` that is)

components/invitation_modal/invitation_modal_members_step.jsx
```diff
@@ -214,36 +242,62 @@ class InvitationModalMembersStep extends React.PureComponent {
...
+                            )}
                             onChange={this.onChange}
                             value={this.state.usersAndEmails}
-                            validAddressMessageId={t('invitation_modal.members.users_emails_input.valid_email')}
+                            validAddressMessageId={t(
+                                'invitation_modal.members.users_emails_input.valid_email',
+                            )}
                             validAddressMessageDefault='Invite **{email}** as a team member'
+                            errorMessageId={
+                                'invitation_modal.invite_members.hit_cloud_user_limit'
```
devinbinnie, line 257: You'll want to add `t` to this id as well to make sure our marker function picks it up as a translation.

components/invitation_modal/invitation_modal_members_step.jsx
```diff
@@ -214,36 +242,62 @@ class InvitationModalMembersStep extends React.PureComponent {
...
+                            }
+                            errorMessageDefault={
+                                'You have reached the user limit for your tier'
+                            }
+                            errorMessageValues={{
+                                text:
+                                    getConfig(getState()).
+                                        ExperimentalCloudUserLimit -
+                                    getState().entities.admin.analytics.
```
devinbinnie, line 266: You'll want to use a selector connected to redux to get this value.

components/widgets/inputs/users_emails_input.jsx
```diff
@@ -258,32 +264,54 @@ export default class UsersEmailsInput extends React.PureComponent {
...
-                loadingMessage={this.loadingMessage}
-                onInputChange={this.handleInputChange}
-                inputValue={this.props.inputValue}
-                openMenuOnFocus={true}
-                onFocus={this.onFocus}
-                tabSelectsValue={true}
-                value={values}
-                aria-label={this.props.ariaLabel}
-            />
+            <div>
```
devinbinnie, line 267: You can avoid adding the `div` here and instead use `React.Fragment` so that it renders the block without an unnecessary `div`. There's a shorthand you can use here that I've suggested below:
```suggestion
            <>
```

components/widgets/inputs/users_emails_input.jsx
```diff
@@ -258,32 +264,54 @@ export default class UsersEmailsInput extends React.PureComponent {
...
+                        >
+                            {(message) => (
+                                <components.NoOptionsMessage>
+                                    {message}
+                                </components.NoOptionsMessage>
+                            )}
+                        </FormattedMarkdownMessage>
+                    </div>
+                )}
+            </div>
```
devinbinnie, line 314: Continued from above:
```suggestion
            </>
```

components/widgets/inputs/users_emails_input.scss
```diff
@@ -203,3 +225,23 @@
         margin-left: 8px;
     }
 }
+
+.InputErrorBox {
+    margin-top: 2px;
+    width: 100%;
+    height: 60px;
+    background: rgba(197, 67, 72, 0.1);
```
devinbinnie, line 233: Does this correspond to a specific colour or should this be themed? Usually most colours correspond to a theme variable we have.

components/widgets/inputs/users_emails_input.scss
```diff
@@ -203,3 +225,23 @@
...
+    align-items: center;
+
+    span {
+        height: 20px;
+        position: relative;
+        font-size: 16px;
+        line-height: 20px;
+        margin: auto;
+        text-align: center;
+        color: #FD5960
```
devinbinnie, line 245: Same with this colour.

package.json
```diff
@@ -36,7 +36,7 @@
     "localforage-observable": "2.0.1",
     "mark.js": "8.11.1",
     "marked": "github:mattermost/marked#87769262aa02e1784570f61f4f962050e07cc335",
-    "mattermost-redux": "github:mattermost/mattermost-redux#47b470894fc7c0cac8181a04b348295b4bf8855b",
+    "mattermost-redux": "git://github.com/mattermost/mattermost-redux.git#MM-27535",
```
devinbinnie, line 39: This should be using the most recent commit hash that includes your redux changes.

webpack.config.js
```diff
@@ -420,7 +420,7 @@ if (targetIsDevServer) {
             }],
             port: 9005,
             watchContentBase: true,
-            writeToDisk: false,
+            writeToDisk: true,
```
devinbinnie, line 423: Was this change intentional?

package-lock.json
```diff
@@ -8245,7 +8245,7 @@
         },
         "p-cancelable": {
           "version": "0.4.1",
-          "resolved": "https://registry.npmjs.org/p-cancelable/-/p-cancelable-0.4.1.tgz",
```
devinbinnie, line 8248: We usually avoid pushing up changes to `package-lock.json` except in the case where `mattermost-redux` needs to be updated. Can we revert the changes that aren't to the `mattermost-redux` repo?

nickmisasi: Yeah sorry, included this for testing only

webpack.config.js
nickmisasi: Yes, in order to get the debugger working with VSCode, but I don't think I should slip this in a PR about a different feature, so I'll take it out.

components/widgets/inputs/users_emails_input.scss
nickmisasi: Fixed all instances of these to point to a variable `var(--error-text)` since it already existed.

components/widgets/inputs/users_emails_input.scss
nickmisasi: This is a specific colour for the background of this error box. I created a variable for the colour so it can be used/overridden if needed.

components/invitation_modal/invitation_modal_members_step.test.jsx
```diff
@@ -5,7 +5,7 @@ import React from 'react';
 
 import {shallowWithIntl} from 'tests/helpers/intl-test-helper';
 
-import InvitationModalMembersStep from './invitation_modal_members_step.jsx';
+import InvitationModalMembersStep from './invitation_modal_members_step/invitation_modal_members_step.jsx';
```
devinbinnie, line 8: You might want to move this test file into the folder you've created so that everything is organized nicely.

components/invitation_modal/invitation_modal_members_step/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import {GlobalState} from 'types/store';
+
+import {isAdmin} from 'utils/utils.jsx';
+
+import InvitationModalMembersStep from './invitation_modal_members_step';
+
+function mapStateToProps(state: GlobalState) {
+    return {
+        userLimit: getConfig(state).ExperimentalCloudUserLimit,
+        currentUsers: state!.entities!.admin!.analytics!.TOTAL_USERS,
```
devinbinnie, line 17: I don't think we need assertions for each of these fields. At minimum `state.entities.admin` should always exist.

components/invitation_modal/invitation_modal_members_step/invitation_modal_members_step.jsx
```diff
@@ -118,6 +121,23 @@ class InvitationModalMembersStep extends React.PureComponent {
         this.props.onSubmit(users, emails, this.state.usersInputValue);
     }
 
+    shouldShowPickerError = () => {
+        const {userLimit, currentUsers, userIsAdmin} = this.props;
+        const admin = userIsAdmin(this.props.currentTeamId);
```
devinbinnie, line 126: Since `userIsAdmin` is only ever called here using a prop passed into the component, can we just calculate this in the connector instead? You have access to passed in props as part of the second argument to `mapStateToProps`

`function mapStateToProps(state: GlobalState, ownProps: OwnProps)`

package-lock.json
```diff
@@ -8251,7 +8251,7 @@
...
+          "resolved": "http://registry.npmjs.org/p-cancelable/-/p-cancelable-0.4.1.tgz",
```
devinbinnie, line 8254: A few more `package-lock.json` changes to remove.

components/invitation_modal/invitation_modal_members_step/invitation_modal_members_step.jsx
nickmisasi: Good point.

components/invitation_modal/invitation_modal_members_step/index.ts
nickmisasi: I actually did this to appease the red squiggly's, but it appears that's not needed anymore. I'll leave it on TOTAL_USERS because things will probably break if that's undefined.

nickmisasi:
@matthewbirtch that phrasing change can be found here: https://github.com/mattermost/mattermost-server/pull/15320

components/invitation_modal/invitation_modal_members_step/__snapshots__/invitation_modal_members_step.test.jsx.snap
```diff
@@ -0,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`components/invitation_modal/InvitationModalMembersStep should match the snapshot 1`] = `
+<ContextConsumer>
```
devinbinnie, line 4: Usually this isn't a good sign when a snapshot goes from 279 lines to 13.
This is probably indicative of a bad test setup.
Looking down it looks like you're importing the connected redux component to the test. You'll want to import just the raw unconnected component to make sure the snapshots look right.

nickmisasi: Cool - imported the component directly and updated the snapshot, which has it now back to its master state.

components/invitation_modal/invitation_modal_members_step/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type OwnProps = {
+    currentTeamId: string;
+}
+
+function mapStateToProps(state: GlobalState, ownProps: OwnProps) {
+    return {
+        userLimit: getConfig(state).ExperimentalCloudUserLimit,
+        currentUsers: state.entities.admin.analytics!.TOTAL_USERS,
+        userIsAdmin: isAdmin(getMyTeamMember(state, ownProps.currentTeamId).roles),
```
marianunez, line 22: Should we use here [`getCurrentUser`](https://github.com/mattermost/mattermost-redux/blob/4e60437ef540f030a5fdb9e921451f80f2f2c1ef/src/selectors/entities/common.ts#L42) to use for the `isAdmin` check? It seems more clear on what we are doing that `getMyTeamMember`

components/invitation_modal/invitation_modal_members_step/invitation_modal_members_step.jsx
```diff
@@ -118,6 +121,22 @@ class InvitationModalMembersStep extends React.PureComponent {
         this.props.onSubmit(users, emails, this.state.usersInputValue);
     }
 
+    shouldShowPickerError = () => {
+        const {userLimit, currentUsers, userIsAdmin} = this.props;
+        if (userLimit === '0' || !userIsAdmin) {
```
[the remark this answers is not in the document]
nickmisasi: Good call. The license stuff didn't actually exist when I wrote this, so I was using the userlimit being 0 as "not cloud". I've added the license check now.

sass/base/_css_variables.scss
```diff
@@ -15,7 +15,6 @@
     --mention-highlight-link: #166de0;
     --new-message-separator: #ff8800;
     --online-indicator: #06d6a0;
-    --sidebar-bg: #145dbf;
```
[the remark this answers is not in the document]
nickmisasi: Probably not. I added it back.

components/invitation_modal/invitation_modal_members_step/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import {GlobalState} from 'types/store';
+
+import {isAdmin} from 'utils/utils.jsx';
+
+import InvitationModalMembersStep from './invitation_modal_members_step';
+
+function mapStateToProps(state: GlobalState) {
+    return {
+        userLimit: getConfig(state).ExperimentalCloudUserLimit,
+        currentUsers: state.entities.admin.analytics!.TOTAL_USERS,
```
marianunez, line 20: I see you have a null check there but what happens if the analytics is not yet in redux? 

We should call [`getStandardAnalytics`](https://github.com/mattermost/mattermost-redux/blob/master/src/actions/admin.ts#L432) if `state.entities.admin.analytics` when the component mounts to guarantee that we are always checking against a correct current user count.

PR #6214, opened 2020-08-20
Cypress/E2E: Fix high scoped scheme spec

josephbaylon:
#### Summary
- Added a wait for location path to end with `/admin_console/user_management/channels` after saving to ensure url has redirected before continuing with the rest of the actions
- I tried to make a separate cypress run but for some reason it's timing out;  it passed on my local though (see `high_scoped_scheme_spec` below)

#### Ticket Link
None -  master only

Screen Shot 2020-08-19 at 9 48 12 PM


srkgupta (approved):
Thanks @josephbaylon. All tests passed. Approving the PR.

saturninoabril (approved):
Thanks, tested and passed.

PR #6223, opened 2020-08-20
MM-23736 Rewrite line numbering for code blocks

hmhealey:
The current code that adds line numbers to code blocks actually mangles the HTML in the code blocks which breaks some syntax highlighting and causes some issues with whitespace as in the ticket.
Screen Shot 2020-08-20 at 12 04 35 PM
The whitespace issue there is obvious, but the multi-line string should also be highlighted.

So instead of adding line numbers by splitting up the highlighted HTML, we're going to instead add a separate div next to the code containing line numbers. Since the code doesn't wrap, the line numbers will just line up with the highlighted text. This is actually how we did line numbers for code file previews in the past, so I consolidated that code when I remember it existed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23736

#### Screenshots
Screen Shot 2020-08-20 at 12 00 02 PM
Screen Shot 2020-08-20 at 12 00 24 PM
Screen Shot 2020-08-20 at 12 01 59 PM


utils/markdown/index.test.js
```diff
@@ -107,6 +105,14 @@ this is long text this is long text this is long text this is long text this is
 ~~~`);
 
         expect(output).toContain('<span class="post-code__language">TeX</span>');
-        expect(output).toContain('<code class="hljs hljs-ln">');
+    });
+
+    test('should add line numbers to code', () => {
```
ali-farooq0, line 110: should we add a test to make sure it fixes the reported case?
````
```python
    op.execute("""
        UPDATE events.settings
        SET name = 'paper_review_conditions'
        WHERE module = 'editing' AND name = 'review_conditions'
    """)
```
````

File: components/code_preview.jsx [hunk 2]
```diff
@@ -129,17 +118,11 @@ export default class CodePreview extends React.PureComponent {
                 <span className='post-code__language'>
                     {`${this.props.fileInfo.name} - ${language}`}
                 </span>
-                <div className='post-code__container'>
-                    <code className='hljs'>
-                        <table>
-                            <tbody>
-                                <tr>
-                                    <td className='post-code__lineno'>{strlines}</td>
```
Willyfrog, line 137: if we are not using this class anywhere else, we should remove it from CSS as well

sass/layout/_markdown.scss
```diff
@@ -196,48 +186,27 @@ h6 {
...
-        color: #ccc;
-        border-right: 1px solid #CCC;
-        vertical-align: top;
-        padding-right: 5px;
-        vertical-align: top;
-        padding-right: 4px;
+    .post-code__line-numbers {
+        @include user-select(none);
+        border-right: 1px solid #CCCCCC;
+        color: #CCCCCC;
```
Willyfrog, line 192: should any of the theme's variable color be considered for this one? it might not be visible on some themes otherwise

Willyfrog (approved):
Looking good, two non-blocking issues

components/code_preview.jsx
hmhealey: Good catch. I thought I'd removed that, but obviously, I didn't.

sass/layout/_markdown.scss
hmhealey: These colours are a bit different since the background comes from the user's code theme which has 4 presets and no colour options. The two dark themes are visible in the screenshots above, but here's the light themes:
Screen Shot 2020-08-24 at 9 25 47 AM
Screen Shot 2020-08-24 at 9 25 59 AM
It might be a bit light, but I'd prefer to address that separately since we can't just do it based on the user's theme.

utils/markdown/index.test.js
```diff
@@ -107,7 +105,31 @@ this is long text this is long text this is long text this is long text this is
...
+        SET name = 'paper_review_conditions'
+        WHERE module = 'editing' AND name = 'review_conditions'
+    """)
+\`\`\``);
+
+        const div = document.createElement('div');
+        div.innerHTML = output;
+
+        // The HTML is valid as long as no new HTML tags have been injected. Unescaped characters are fine though.
+        expect(div.innerHTML).toBe(output.replace(/&quot;&quot;&quot;/g, '"""').replace(/&#x27;/g, '\''));
```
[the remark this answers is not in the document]
hmhealey: That'd also be an option, but I think this is fine as well. They both check to make sure that the hljs-string span doesn't get mangled in slightly different ways.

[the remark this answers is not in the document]
ali-farooq0: Will this fail if the said bug was reintroduced? If so then 👌.

jgilliam17:
@hmhealey adding a comment here per our discussion 
There's an issue with alignment of line numbers and text on file preview
You mentioned scrolling through code preview with the mouse wheel is stuck on the first try. I wasn't able to verify on my end, works fine with the track pad. I will re-test with the mouse shortly.

Also confirmed working as expected:
- Line numbers line up with the text in the code block and are not selectable;  appear correctly in search results when highlighted.
- sh formatting has been corrected as mentioned in https://mattermost.atlassian.net/browse/MM-27856

hmhealey:
@jgilliam17 I've fixed the misaligned numbers, and I'm not seeing the scrolling issue any more, so this should be good for a second round of tests.
Screen Shot 2020-08-27 at 2 22 07 PM
Screen Shot 2020-08-27 at 2 22 13 PM

PR #6147, opened 2020-08-14
MM-27559 - Clicking 'Create a playbook' button does nothing on MM-Des…

ali-farooq0:
#### Summary
Allowing any internal-mattermost link to open within the app.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-27559

#### Screenshots
2020-08-14 at 1 52 PM

PR #6240, opened 2020-08-21
MM-28042,MM-28046 - Web/Desktop: Relative links open in a new tab/window

ali-farooq0:
#### Summary
Restored the regex used to detect internal links

#### Ticket Link
Fixes:
https://mattermost.atlassian.net/browse/MM-28042
https://mattermost.atlassian.net/browse/MM-28046


sudheerDev (approved):
LGTM

sudheerDev:
@ali-farooq0 Linked another ticket and assigned it to you as it solves that as well.

hmhealey:
What about https://github.com/mattermost/mattermost-webapp/pull/6147? Perhaps we can just check for if the link starts with the site URL or `/` instead?

ali-farooq0:
I'm afraid that with the relative link it might be possible to abuse it in someway which is why i kept the regex in. Also I do check if the link starts with siteURL or the origin regex `/(...admin...|pl...)` so  #6147 is still works.

hmhealey:
Yeah, that makes sense. I was worried that might be an issue when I was suggesting it. I don’t think that’s actually the case, but it’s good to be safe. As long as there’s no regressions here, this looks good to me.

ali-farooq0:
@stevemudie would you be able help QA this PR? Thanks!

PR #6245, opened 2020-08-22
MM-T250 Mark as unread in the RHS

openmohan:
#### Summary

MM-T250 Mark as unread in the RHS

#### Ticket Link

[MM-T250](https://automation-test-cases.vercel.app/test/MM-T250)


saturninoabril (approved):
Thanks @openmohan. Tested and passed.

PR #6249, opened 2020-08-24
MM-T452 Add E2E test to validate user can join an open channel via a link in DM

Niceplace:
#### Summary

E2E Test for Hackfest : Join open team from link in DM

This test validates that a user which is not member of an open team (a team that any user on the server can join without an invite) can automatically join this open team by using the URL any of  the team's public channels.

#### Ticket Link

[MM-T452](https://automation-test-cases.vercel.app/test/MM-T452])

### DIFF
--- a/e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
+++ b/e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
@@ -1,0 +1,123 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// Stage: @prod
+// Group: @multi_team_and_dm
+
+describe('Join an open team from a direct message link', () => {
+    let openTeam;
+    let testUserInOpenTeam;
+
+    let secondTestTeam;
+    let testUserOutsideOpenTeam;
+
+    before(() => {
+        cy.apiCreateTeam('mmt452-second-team', 'mmt452-second-team', 'I', true).then(({team}) => {
+            secondTestTeam = team;
+
+            // # Create pool of users for tests
+            cy.apiCreateUser().then(({user}) => {
+                testUserOutsideOpenTeam = user;
+                cy.apiAddUserToTeam(secondTestTeam.id, testUserOutsideOpenTeam.id);
+            });
+        });
+
+        cy.apiCreateTeam('mmt452-open-team', 'mmt452-open-team', 'O', true).then(({team}) => {
+            openTeam = team;
+
+            // # Allow any user with an account on this server to join this team
+            cy.apiPatchTeam(openTeam.id, {
+                allow_open_invite: true,
+            });
+
+            // # Create pool of users for tests
+            cy.apiCreateUser().then(({user}) => {
+                testUserInOpenTeam = user;
+                cy.apiAddUserToTeam(openTeam.id, testUserInOpenTeam.id);
+
+                // # Login as test user
+                cy.apiLogin(testUserInOpenTeam);
+            });
+        });
+    });
+
+    it('MM-T452 User with no teams should be able to join an open team from a link in direct messages', () => {
+        // # View 'off-topic' channel
+        cy.visit(`/${openTeam.name}/channels/off-topic`);
+
+        // * Expect channel title to match title passed in argument
+        cy.get('#channelHeaderTitle').
+            should('be.visible').
+            and('contain.text', 'Off-Topic');
+
+        // # Copy full url to channel
+        cy.url().then((openTeamChannelUrl) => {
+            // # From the 'Direct Messages' menu, find a specific user and send 'hi'
+            sendDirectMessageToUser(testUserOutsideOpenTeam, openTeamChannelUrl);
+
+            // # Logout as test user
+            cy.apiLogout();
+
+            // # Login as user outside team
+            cy.apiLogin(testUserOutsideOpenTeam);
+
+            // # Open direct message from the user in the open team (testUserInOpenTeam)
+            cy.get(`a[href="/${secondTestTeam.name}/messages/@${testUserInOpenTeam.username}"]`).
+                click();
+
+            // * Expect channel title to contain the username (ensures we opened the right DM)
+            cy.get('#channelHeaderTitle').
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// Stage: @prod
```
saturninoabril, line 4: Please remove. We'll add it in the future once we found it to be stable.
```suggestion
```

e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+                // # Login as test user
+                cy.apiLogin(testUserInOpenTeam);
+            });
+        });
+    });
+
+    it('MM-T452 User with no teams should be able to join an open team from a link in direct messages', () => {
+        // # View 'off-topic' channel
+        cy.visit(`/${openTeam.name}/channels/off-topic`);
```
saturninoabril, line 46: Instead of default off-topic channel, please create a new public channel and use it to verify whether the test user also joined the said public channel.

e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
```diff
@@ -0,0 +1,125 @@
...
+        // # Copy full url to channel
+        cy.url().then((openTeamChannelUrl) => {
+            // # From the 'Direct Messages' menu, send the URL of the public channel to the user outside the team
+            sendDirectMessageToUser(testUserOutsideOpenTeam, openTeamChannelUrl);
+
+            // # Logout as test user
+            cy.apiLogout();
+
+            // # Login as user outside team
+            cy.apiLogin(testUserOutsideOpenTeam);
```
saturninoabril, line 62: I suggests to do `cy.reload()` after logging in as another user to take effect the state related to such current session.

e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+            // # Open direct message from the user in the open team (testUserInOpenTeam)
+            cy.get(`a[href="/${secondTestTeam.name}/messages/@${testUserInOpenTeam.username}"]`).
```
saturninoabril, line 65: We don't normally target an element using this approach but I'm fine as is since it also validates whether the test user is currently viewing the `secondTestTeam`.

e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
```diff
@@ -0,0 +1,125 @@
...
+                click();
+
+            // * Expect channel title to contain the username (ensures we opened the right DM)
+            cy.get('#channelHeaderTitle').
+                should('be.visible').
+                and('contain.text', `${testUserInOpenTeam.username}`);
+
+            // # Click on URL sent by the user in the open team
+            cy.get('[data-testid="postContent"]').
```
saturninoabril, line 74: Can simply be,
```suggestion
            cy.findByTestId('postContent').
```

saturninoabril (changes requested):
Thanks @Niceplace! Please see comments.

e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
Niceplace: I understand. In that scenario my alternative was to fetch the unique ID of the channel via API calls like I previously did here: https://github.com/mattermost/mattermost-webapp/blob/a7e0ce41af85b67429ad53e3d7acbb17ef9e627d/e2e/cypress/integration/multi_team_and_dm/close_current_dm_redirects_spec.js#L135

This method requires sysadmin token, which means I had to login as the sysadmin to make the request and that seemed complicated to do in the middle of a test. I wish there was a way to make API calls and supply the token instead of having to rely on the current user session's authentication to do it.
I will keep as is but if I stumble upon a similar scenario in the future I will try to find a better way to do this.

e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+    // # Click on "Go" in the group message's dialog to begin the conversation
+    cy.get('#saveItems').click();
+
+    // * Expect the channel title to be the user's username
+    // In the channel header, it seems there is a space after the username, justifying the use of contains.text instead of have.text
+    cy.get('#channelHeaderTitle').should('be.visible').and('contain.text', user.username);
+
+    // # Type message and send it to the user
+    cy.get('#post_textbox').
```
josephbaylon, line 129: Can be replaced with postMessage to post message to textbox,
```
cy.postMessage(message);
```

e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
josephbaylon: `cy.reload()` might be required in this case because it the test fails for me without it.

e2e/cypress/integration/multi_team_and_dm/join_open_team_from_dm_spec.js
josephbaylon: Another alternative is to just visit the dm page directly,
```
cy.visit(`/${secondTestTeam.name}/messages/@${testUserInOpenTeam.username}`);
```

Niceplace: Yeah Sweet ! There is always a simpler solution :D

Niceplace:
Thanks for the comments @josephbaylon' ! I updated based on your suggestions, cheers !

saturninoabril (approved):
@Niceplace Thanks, LGTM!

PR #6256, opened 2020-08-24
MM-27432 Hide notification badge on channel when menu is open

hmhealey:
To hide the badge when the menu is open, the parent of the menu (`SidebarChannelLink`) has to know whether or not the menu is open. Since `SidebarCategory` already knows whether or not its menu is open, I made it so that the parent of a `SidebarMenu` is always the one in charge of its open state which requires passing around some props and `onToggleMenu` listeners.

The other major change because of this is moving the calculation of the menu position to happen in `componentDidUpdate` after the menu is opened instead of happening as a side effect of the `setState` caused by clicking the toggle button.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27432

### DIFF
--- a/components/sidebar/sidebar_category/__snapshots__/sidebar_category.test.tsx.snap
+++ b/components/sidebar/sidebar_category/__snapshots__/sidebar_category.test.tsx.snap
@@ -53,7 +53,8 @@ exports[`components/sidebar/sidebar_category should match snapshot 2`] = `
           "user_id": "",
         }
       }
-      onToggle={[Function]}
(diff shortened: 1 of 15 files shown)


devinbinnie:
@hmhealey You have a bunch of type check errors to resolve.

hmhealey:
Note that SidebarMenu now appears to be controlled, but it isn't completely controlled because `MenuWrapper` controls its own state and doesn't have an option to pass in the open state as props. I'm not wanting to make further changes there at this time.

PR #6258, opened 2020-08-25
MM-28005-don't allow edit if ldap picture set on sync

sbishel:
#### Summary
Ldap Picture should not be allowed to be set by the user if it is set during login via LDAP for both LDAP and SAML login types. This PR implements for SAML.

In addition, the property "LdapPictureAttributeSet" is no longer getting set in index.ts. This was removed when the filg was converted to typescript. I'm guessing this was due to it missing in the Config datatype. Related PR adds this property.

- [ ] Update Redux before merging

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28005

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1222

#### Screenshots
Screen Shot 2020-08-24 at 7 48 33 PM


srkgupta (approved):
Tested the PR on localhost and there were no regressions found. In order to test the actual issue, it needs to be tested on the QA test instance with ADFS configured. Hence the actual issue will be tested post merge. Approving the PR as is.

PR #6266, opened 2020-08-25
[MM-27812]: e2e cypress test for TM4J test MM-T551

ashishbhate:
#### Summary

Automates MM-T551

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27812


e2e/cypress/integration/notifications/not_unless_Im_mentioned_spec.js
```diff
@@ -0,0 +1,140 @@
...
+            return true;
+        };
+
+        win.Notification = Notification;
+
+        cy.spy(win, 'Notification').as('notifySpy');
+    });
+
+    // Verify that we now have a Notification property
+    cy.window().should('have.property', 'Notification');
```
saturninoabril, line 62: We're using this in another place. I'd suggest to extract this into a common function, probably at a new `cypress/integration/support/notification.js`, then maybe call as `spyNotificationAs('notifySpy')`. (Also please update `at_mentions_spec.js` to use such exported function.)

e2e/cypress/integration/notifications/not_unless_Im_mentioned_spec.js
```diff
@@ -0,0 +1,140 @@
...
+}
+
+describe('reply-notifications', () => {
+    function ignoreUncaughtException() {
+        cy.on('uncaught:exception', (err) => {
+            expect(err.message).to.include('.close is not a function');
+
+            return false;
+        });
+    }
```
saturninoabril, line 72: Same here, please extract to `cypress/integration/support/notification.js`. I think we missed to export it before but I'm requesting to make it happen here.  Thanks.

e2e/cypress/integration/notifications/not_unless_Im_mentioned_spec.js
```diff
@@ -0,0 +1,140 @@
...
+    });
+
+    it('MM-T551 not unless I\'m mentioned', () => {
+        ignoreUncaughtException();
+
+        // # Set users notification settings
+        setReplyNotificationsSetting('#notificationCommentsNever');
+
+        // # Navigate to town square channel
+        cy.get(`#sidebarItem_${'town-square'}`).scrollIntoView().click({force: true});
```
saturninoabril, line 116: Call to `scrollIntoView` here and other places seemed to be not necessary since element are not out of view.

e2e/cypress/integration/notifications/not_unless_Im_mentioned_spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+        // # post a message
```
saturninoabril, line 118: Nit: Please capitalize comments for consistency. Thanks

e2e/cypress/integration/notifications/not_unless_Im_mentioned_spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+        // * Verify stub was not called
+        cy.get('@notifySpy').should('be.not.called');
+
+        // * Verify unread mentions badge does not exist
+        cy.get('#publicChannel').scrollIntoView();
+        cy.get('#sidebarItem_town-square').
+            scrollIntoView().
+            find('#unreadMentions').
+            should('be.not.visible');
```
saturninoabril, line 138: This can simply,
```javascript
cy.get('#sidebarItem_town-square').find('#unreadMentions').should('be.not.visible');
```
or preferably,
```suggestion
        cy.findByLabelText('town square public channel unread').should('be.visible');
```

e2e/cypress/integration/notifications/not_unless_Im_mentioned_spec.js
```diff
@@ -0,0 +1,140 @@
...
+    });
```
saturninoabril, line 139: I think the test is missing whether the notification really works. So, I suggest to post another reply mentioning the receiver and verify whether the notification is called/received. Something like,
```javascript
            // # Switch again to other channel
            cy.get(`#sidebarItem_${otherChannel.name}`).click({force: true});

            // # Reply to a post as another user mentioning the receiver
            cy.postMessageAs({sender, message: `Hello @${receiver.username}`, channelId: townsquareChannelId, rootId: postId});

            // * Verify stub was called
            cy.get('@notifySpy').should('be.called');

            // * Verify unread mentions badge exists
            cy.findAllByLabelText('town square public channel 1 mention').should('be.visible');
```
I'll update the test case in TM4J as well to reflect this new request.

saturninoabril (changes requested):
Thanks @ashishbhate! Tested and passed but left few suggestions.

e2e/cypress/integration/notifications/not_unless_Im_mentioned_spec.js
ashishbhate: I've applied the first suggestion (dropping `scrollIntoView`).

In this test we're checking the absence of a badge. Although the second suggestion is technically correct, to me it feels unintuitive because we're testing the absence of something by checking the presence of something else.

I think the intent original code is more easy to understand.

saturninoabril: Thanks for checking and totally makes sense.

saturninoabril (approved):
@ashishbhate Thanks! Tested and passed.

ogi-m (approved):
LGTM 👍

PR #6271, opened 2020-08-26
MM-T12 Loading indicator when posting images

openmohan:
#### Summary

MM-T12 Loading indicator when posting images

#### Ticket Link

[MM-T12](https://automation-test-cases.vercel.app/test/MM-T12)

### DIFF
--- a/e2e/cypress/integration/files_and_attachments/upload_files_spec.js
+++ b/e2e/cypress/integration/files_and_attachments/upload_files_spec.js
@@ -84,18 +84,18 @@ describe('Upload Files', () => {
         // * Verify the image loading component is visible
         cy.get('.image-container').should('be.visible').find('.image-loading__container').should('be.visible');
 
-        // # OtherUser creates posts in the channel
-        cy.postMessageAs({
-            sender: otherUser,
-            message: 'post1',
-            channelId: testChannel.id,
-        });
+        Cypress._.times(5, () => {
+            // # OtherUser creates posts in the channel
+            cy.postMessageAs({
+                sender: otherUser,
+                message: 'message',
+                channelId: testChannel.id,
+            });
 
-        for (let i = 0; i < 5; i++) {
             // * Verify image is not loading for each posts
             cy.get('.image-container').should('be.visible').find('.image-loading__container').should('not.exist');
 
             cy.wait(TIMEOUTS.HALF_SEC);
-        }
+        });
     });
 });
--- a/e2e/cypress/integration/files_and_attachments/upload_files_spec.js
+++ b/e2e/cypress/integration/files_and_attachments/upload_files_spec.js
@@ -12,12 +12,24 @@
 
 describe('Upload Files', () => {
     let testTeam;
+    let testChannel;
+    let otherUser;
 
     before(() => {
         // # Create new team and new user and visit Town Square channel
-        cy.apiInitSetup({loginAfter: true}).then(({team}) => {
+        cy.apiInitSetup().then(({team, channel}) => {
             testTeam = team;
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/files_and_attachments/upload_files_spec.js
```diff
@@ -52,4 +64,26 @@ describe('Upload Files', () => {
             });
         });
     });
+
+    it('MM-T12 Loading indicator when posting images', () => {
+        const filename = 'huge-image.jpg';
```
saturninoabril, line 69: I think this needs to visit again the channel since the previous test left the RHS open.

e2e/cypress/integration/files_and_attachments/upload_files_spec.js
```diff
@@ -52,4 +64,26 @@ describe('Upload Files', () => {
...
+
+        // # Post an image in center channel
+        cy.get('#centerChannelFooter').find('#fileUploadInput').attachFile(filename);
+        cy.get('.post-image').should('be.visible');
+        cy.postMessage('{enter}');
+
+        // # Login as otherUser
+        cy.apiLogin(otherUser);
```
saturninoabril, line 77: Better to reload the page after login and verify that the `.image-loading__container` is visible.

e2e/cypress/integration/files_and_attachments/upload_files_spec.js
```diff
@@ -52,4 +64,26 @@ describe('Upload Files', () => {
...
+
+        // # OtherUser creates posts in the channel
+        cy.postMessageAs({
+            sender: otherUser,
+            message: 'post1',
+            channelId: testChannel.id,
+        });
+
+        // * Verify image is not loading for each posts
+        cy.get('.image-loading__container').should('not.be.visible');
```
saturninoabril, line 87: Actually, I'm not sure if this will going to be effective but I'll leave it as is.

Maybe do this several times to make sure that loading indicator is not showing up.

saturninoabril (changes requested):
Thanks @openmohan, please see comments.

e2e/cypress/integration/files_and_attachments/upload_files_spec.js
josephbaylon, line 87: Since the parent of the loading container is `image-container`, maybe this can be improved to,

```suggestion
        cy.get('.image-container').should('be.visible').find('.image-loading__container').should('not.exist');
```

But yeah, as @saturninoabril suggested, verify first that loading indicator shows up when posting several images.

e2e/cypress/integration/files_and_attachments/upload_files_spec.js
openmohan: I will change `before` to `beforeEach`, so the test case will be more predictable.

openmohan:
@saturninoabril @josephbaylon Thanks for the reviews. I have updated the PR with suggestions.

e2e/cypress/integration/files_and_attachments/upload_files_spec.js
```diff
@@ -52,4 +66,36 @@ describe('Upload Files', () => {
...
+
+        // # OtherUser creates posts in the channel
+        cy.postMessageAs({
+            sender: otherUser,
+            message: 'post1',
+            channelId: testChannel.id,
+        });
+
+        for (let i = 0; i < 5; i++) {
+            // * Verify image is not loading for each posts
```
saturninoabril, line 95: Should the `cy.postMessageAs` inside the loop (or better `Cypress._.times()`)?

saturninoabril (approved):
Thanks @openmohan! Tested and passed.

PR #6273, opened 2020-08-26
MM-27825: Add Cypress Test Case for MM-T495

grundleborg:
#### Summary
Adds cypress test for MM-T495 Desktop Notifications - Can set to DND and no notification fires on DM

*Note*: I have a small concern that the mocking of granting permissions for notifications is not actually working but I can't seem to figure this out. Can anyone who hasn't already manually granted the permissions please take a look and check if it works as intended.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27825

### DIFF
--- a/e2e/cypress/integration/notifications/desktop_notifications_spec.js
+++ b/e2e/cypress/integration/notifications/desktop_notifications_spec.js
@@ -36,19 +36,19 @@ describe('Desktop notifications', () => {
             cy.apiCreateDirectChannel([testUser.id, user.id]).then((res) => {
                 const channel = res.body;
 
-                // # Ensure notifications are set up to fire a desktop notification if you receive a DM
+                // Ensure notifications are set up to fire a desktop notification if you receive a DM
                 cy.apiPatchUser(user.id, {notify_props: {...user.notify_props, desktop: 'all'}});
 
                 // Visit the MM webapp with the notification API stubbed.
                 stubNotificationAs('withNotification', `/${testTeam.name}/channels/town-square`, 'granted');
 
-                // # Post the following: /dnd
+                // Make sure user is marked as online.
                 cy.get('#post_textbox').clear().type('/online{enter}');
 
-                // # Have another user send you a DM
+                // Have another user send you a DM to trigger a Desktop Notification.
                 cy.postMessageAs({sender: testUser, message: MESSAGES.TINY, channelId: channel.id});
 
-                // * Desktop notification is not received
+                // Desktop notification should be received.
                 cy.wait(TIMEOUTS.HALF_SEC);
                 cy.get('@withNotification').should('have.been.calledOnce');
             });
--- a/e2e/cypress/integration/notifications/desktop_notifications_spec.js
+++ b/e2e/cypress/integration/notifications/desktop_notifications_spec.js
@@ -78,6 +78,10 @@ describe('Desktop notifications', () => {
                 // * Desktop notification is not received
                 cy.wait(TIMEOUTS.HALF_SEC);
                 cy.get('@withoutNotification').should('not.have.been.called');
+
+                // * Verify that the status indicator next to your name has changed to "Do Not Disturb"
+                cy.get('button[aria-label="set status"] > span > svg').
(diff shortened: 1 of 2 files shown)


e2e/cypress/integration/notifications/desktop_notifications_spec.js
```diff
@@ -0,0 +1,58 @@
...
+            cy.apiCreateDirectChannel([testUser.id, user.id]).then((res) => {
+                const channel = res.body;
+
+                // Mock window.Notification to check if desktop notifications are triggered.
+                cy.visit(`/${testTeam.name}/channels/town-square`, {
+                    onBeforeLoad(win) {
+                        cy.stub(win.Notification, 'permission', 'granted');
+                        cy.stub(win, 'Notification').as('Notification');
+                    },
+                });
```
saturninoabril, line 41: To ensure that the mocking works, we probably need to check the initial state where the user should receive a notification. With that, I'd suggest to move this into a function so we can reuse stubbing for with and without notification.
So for initial state, it can be,
```javascript
// # Ensure notifications are set up to fire a desktop notification if you receive a DM
cy.apiPatchUser(user.id, {notify_props: {...user.notify_props, desktop: 'all'}});

// Mock window.Notification to check if desktop notifications are triggered.
stubNotificationAs('withNotification');
cy.wait(TIMEOUTS.FIVE_SEC);

// # Have another user send you a DM
cy.postMessageAs({sender: testUser, message: MESSAGES.TINY, channelId: channel.id});

// * Desktop notification is received
cy.get('@withNotification').should('have.been.calledWithMatch', 'Direct Message', (args) => {
    expect(args.body).to.equal(`@${testUser.username}: ${MESSAGES.TINY}`);
    return true;
});
```

Then for `/dnd`
```javascript
stubNotificationAs('withoutNotification');
cy.wait(TIMEOUTS.FIVE_SEC);
...
cy.get('@withoutNotification').should('not.have.been.called');
```

saturninoabril (changes requested):
Thanks @grundleborg. Please see suggestion on how to ensure that the mocking works.  

Btw, I've clarified the second expectation in the test case which originally reads "Turn off Do Not Disturb by using the status indicator drop-down next to your name at top of LHS" and QA team agreed that it's not correct and it should be "Verify that the status indicator next to your name has changed to Do Not Disturb". I've updated the TM4J test case. Please help add the last verification.

e2e/cypress/integration/notifications/desktop_notifications_spec.js
grundleborg: @saturninoabril makes sense, just wondering, do you see some way of injecting the notification stub outside of the `cy.visit` call like I've done it in the PR so far? Otherwise the `stubNotificationAs` function is going to have to do the `visit` call inside it I think.

grundleborg:
@saturninoabril addressed your review comments. Please take a look.

e2e/cypress/integration/notifications/desktop_notifications_spec.js
saturninoabril: I'm good as is but maybe it can be done thru `cy.window`, like
```javascript
function stubNotificationAs(name) {
    cy.window().then((win) => {
        function Notification(title, opts) {
            this.title = title;
            this.opts = opts;
        }

        Notification.requestPermission = () => 'granted';
        Notification.close = () => true;

        win.Notification = Notification;

        cy.stub(win, 'Notification').as(name);
    });
}
```

Then, 
```javascript
cy.visit(`/${testTeam.name}/channels/town-square`);
stubNotificationAs('withoutNotification');
```

grundleborg: I'll give it a try...

grundleborg: @saturninoabril updated as per your suggestion.

saturninoabril (approved):
Thank you @grundleborg! Tested and passed.

isacikgoz (approved):
LGTM 👍

lindy65 (approved):
Thanks @grundleborg - looks good to me 👍

PR #6282, opened 2020-08-26
E2E: Always use UTC timezone during local execution

calebroseland:
#### Summary
During execution of `messaging/local_date_time_spec.js`, an assertion ~~will~~ may throw the following error.
> Timed out retrying: expected '<time.post__time>' to have text '4:30 AM', but the text was '10:30 PM'

The '10:30 PM' part is variable based on the timezone of the machine it is running on. 

As it already exists for the Jest unit tests of this repo, this PR specifies UTC as the timezone to eliminate the above variation.


saturninoabril (approved):
Thanks for fixing @calebroseland!

ashishbhate (approved):

PR #6288, opened 2020-08-27
MM-T402 - Finish Tutorial

josephk96:
This e2e test does the following -
1. Using the sysadmin account, create a team and make sure that anyone can join the team
2. Logout from the sysadmin account and create a new account/user
3. Finish the initial tutorial
4. Complete the following three tutorials by clicking on the pulsating tooltips - Message box, Town Square, Main Menu
5. Reload the page without cache
6. Ensure that when the page is reloaded, the tutorial does not appear

I disabled LDAP (not sure if necessary, but just in case) for this test. After clicking the pulsating tooltip(s), I initially attempted to verify that the correct text was being shown via Testing Library's `findByText` method, but it was failing either due to ambiguous text being displayed or because the text was being displayed in many html sub-elements, so I used `contains` and checked for a part of each tutorial paragraph.

I did not save the text I tested in variables because they were not being reused, but let me know if you think that I should :)

Test case: https://automation-test-cases.vercel.app/test/MM-T402


e2e/cypress/integration/onboarding/finish_tutorial_spec.js
```diff
@@ -0,0 +1,113 @@
...
+
+// Group: @onboarding
+
+import * as TIMEOUTS from '../../fixtures/timeouts';
+import {generateRandomUser} from '../../support/api/user';
+
+describe('Onboarding', () => {
+    let testTeam;
+    const user = generateRandomUser();
+    const {username, email, password} = user;
```
saturninoabril, line 19:
```suggestion
    const {username, email, password} = generateRandomUser();
```

saturninoabril (approved):
Thanks @josephk96! Tested and passed. LGTM, except for last comment. Approving the PR.

PR #6296, opened 2020-08-27
[MM-27526] Onboarding Flow for End Users

nickmisasi:
#### Summary
When an end user logs in to mattermost for the first time, we want to greet them with an onboarding flow. This PR adds a few steps to Devin's work that reveals cards with end-user specific actions on them. 

Setting up your profile
Enable desktop notifications
Set preferences

Once these are complete (or skipped) the user will be shown some end-user specific tips. Currently, these are:
Browse Channels
See who's here
Keyboard shortcuts

Each of these cards has buttons that will open a modal for each action.

*note:* When using the Turn on notifications button, this will only request access on your first try for your entire browser. It is not per user. If you have already enabled notifications you will see a notification, rather than having permission requested.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27526

#### Related Pull Requests
- Has server changes n/a
- Has redux changes n/a
- Has mobile changes n/a

#### Screenshots

image
image
image
image
image


devinbinnie:
I'm noticing that the sidebar element is show 5 steps when it should be 3. You'll probably need to filter the total number of steps on user as well

image

devinbinnie:
Looks like the margin is off on steps 2 and 3 by a few pixels:

image

components/next_steps_view/next_steps_view.tsx
```diff
@@ -183,8 +184,15 @@ export default class NextStepsView extends React.PureComponent<Props, State> {
         );
     }
 
+    // Filter the steps shown by checking if our user has any of the required roles for that step
+    filterSteps = (step: StepType) => {
+        return isStepForUser(step, this.props.currentUser.roles);
+    }
+
+    steps = Steps.filter(this.filterSteps);
```
devinbinnie, line 192: This is the sort of thing that should be handled by a redux selector. We should try and avoid adding global constants to a class like this.

components/next_steps_view/steps.ts
```diff
@@ -6,45 +6,93 @@ import {getLicense} from 'mattermost-redux/selectors/entities/general';
...
+        roles: [],
     },
     {
         id: RecommendedNextSteps.TEAM_SETUP,
-        title: localizeMessage('next_steps_view.titles.teamSetup', 'Name your team'),
+        title: localizeMessage(
+            'next_steps_view.titles.teamSetup',
+            'Name your team'
+        ),
+        roles: ['system_admin', 'system_user'],
```
devinbinnie, line 55: Shouldn't this one and the one below it only be `system_admin`?

components/next_steps_view/steps/setup_preferences_step/setup_preferences_step.tsx
```diff
@@ -0,0 +1,40 @@
...
+                    'You can change how you receive notifications, update your profile, customize display settings and more. Preferences can be accessed through the Main Menu.'
+                }
+                buttonMessageId={
+                    'next_steps_view.preferenceSetup.setPreferences'
+                }
+                buttonDefaultMessage={'Set Preferences'}
+                onClick={onClick}
+            />
+            {shouldShowSettings && (
+                <UserSettingsModal onHide={onFinish}/>
```
devinbinnie, line 35: This modal is hooked up to redux, so we should have it open with `openModal` as it is a better practice to make use of the `ModalController`

components/next_steps_view/steps.ts
nickmisasi: The way the role system returns your roles is you have your role, and all roles underneath you. So an admin has roles: `system_admin system_user`, while a user has just `system_user`. There's a comment here: https://github.com/mattermost/mattermost-webapp/blob/0ce08414374cc97dc109f64e4717e7b970107da5/components/next_steps_view/steps.ts#L33 about it. 

I was trying to find a clean solution to this for a while, and this way meant the least conditional logic, otherwise it got weird when there were certain cases of a step that needed to be for BOTH admins and normal users (ie, the complete profile step). A simple isAdmin boolean wouldn't work for those cases.

components/next_steps_view/next_steps_view.tsx
nickmisasi: Good callout. This is my redux noobiness showing. Much cleaner with the selector.

components/next_steps_view/steps/setup_preferences_step/setup_preferences_step.tsx
nickmisasi: I cannot for the life of me get the UserSettingsModal to call the onHide prop I pass it through `dialogProps` in openModal. I've tried both with connecting the component, and using the useDispatch hook. No matter what I do, my passed onHide function is never called. Will need some time to dig into this one more

nickmisasi:
The 2 UX issues you mentioned should be fixed in the next spinwick @devinbinnie

components/next_steps_view/steps.ts
devinbinnie: Gotcha, I wasn't aware how the roles worked. Thanks for the insight :)

components/next_steps_view/steps/setup_preferences_step/setup_preferences_step.tsx
devinbinnie: Let me know if I can be of some help :)

nickmisasi: Should be good in the next spinwick.

package-lock.json
```diff
@@ -18690,8 +18672,8 @@
       }
     },
     "mmjstool": {
-      "version": "github:mattermost/mattermost-utilities#beffd8b4c5b985ef1bfb500e06f3306d40a67599",
-      "from": "github:mattermost/mattermost-utilities#beffd8b4c5b985ef1bfb500e06f3306d40a67599",
+      "version": "github:mattermost/mattermost-utilities#519b99a4e51e6c67a0dbd46a6efdff27dc835aaa",
+      "from": "github:mattermost/mattermost-utilities#519b99a4e51e6c67a0dbd46a6efdff27dc835aaa",
```
devinbinnie, line 18676: Does this need to be upgraded here?

components/next_steps_view/steps.test.tsx
```diff
@@ -15,6 +15,12 @@ describe('components/next_steps_view/steps', () => {
                 preferences: {
                     myPreferences: {},
                 },
+                users: {
+                    currentUserId: 'current_user_id',
+                    profiles: {
+                        current_user_id: {roles: 'system_role'},
+                    },
+                },
```
marianunez, line 23: Could we add a test for the steps for the case of non-admin ?

components/sidebar/sidebar_next_steps/sidebar_next_steps.tsx
```diff
@@ -21,7 +26,9 @@ type Props = {
     active: boolean;
     showNextSteps: boolean;
     currentUserId: string;
+    currentUser: UserProfile;
```
marianunez, line 29: Is this being used?

nickmisasi: Nope. Removed.

components/next_steps_view/steps.test.tsx
nickmisasi: I've added 2 small tests to make sure the filter steps logic is working based on a user's role.

PR #6304, opened 2020-08-28
revert cypress to 4.8.0

Willyfrog:
#### Summary
Downgrade cypress to 8.4.0 to have more stable tests

while trying to create [this test](https://github.com/Willyfrog/mattermost-webapp/blob/MM-27207_e2e-archived-channels-I/e2e/cypress/integration/channel/archived_channels_spec.js#L108) i was running into many different problems as either it was taking longer to load some elements or it break while typing/waiting for no reason. Downgrading seems to get stable results.

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -8,7 +8,7 @@
     "chai": "4.2.0",
     "chalk": "4.1.0",
     "cross-env": "7.0.2",
-    "cypress": "4.12.1",
+    "cypress": "^4.8.0",
     "cypress-file-upload": "4.0.7",
     "cypress-multi-reporters": "1.4.0",
     "cypress-plugin-tab": "1.0.5",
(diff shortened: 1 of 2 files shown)


e2e/package.json
```diff
@@ -8,7 +8,7 @@
     "chai": "4.2.0",
     "chalk": "4.1.0",
     "cross-env": "7.0.2",
-    "cypress": "4.12.1",
+    "cypress": "^4.8.0",
```
saturninoabril, line 11: Should this be
```suggestion
    "cypress": "4.8.0",
```

saturninoabril: Also please change title and description to `4.8.0`

saturninoabril (approved):
Thanks @Willyfrog! Please change title and description to 4.8.0.

PR #6312, opened 2020-08-28
MM-27819 MM-T471 Add unit test for message reactions in a GM channel

hmhealey:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27819


e2e/cypress/integration/messaging/message_reaction_gm_spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+            // # Choose "slightly_frowning_face" emoji
+            // delaying 500ms in case of lag
+            cy.get('.emoji-picker__items #emoji-1f641').wait(500).click();
+
+            // * The number shown on the reaction is incremented by 1
+            cy.get(`#postReaction-${postId}-slightly_frowning_face .Reaction__number--display`).
+                should('have.text', '1').
+                should('be.visible');
+        });
```
srkgupta, line 62: The cypress test is missing the following assertion specified in the TM4J test:

`Narrow the Desktop app window and verify plus sign appears next to existing reaction (hover not needed in narrow mobile view)`

Suggested changes:

```
            // # resize window to mobile view
            cy.viewport('iphone-6');

            // # Focus on last postReaction - Workaround for hover not needed in Narrow Desktop view
            cy.get('#post_textbox').type(' testing').focus().tab().tab({shift: true}).tab({shift: true});

            // * Verify if Add Reaction button is displayed
            cy.findByLabelText('Add a reaction').should('be.visible');
```

srkgupta (changes requested):
Hi @hmhealey 
Thanks for adding the cypress test. The test on an overall looks good. Just added a couple of suggestions.

e2e/cypress/integration/messaging/message_reaction_gm_spec.js
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Stage: @prod @smoke
```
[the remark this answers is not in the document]
hmhealey: Will do. Is there something in Jira that says what the Group is supposed to be for a given test? I hadn't really paid attention to this heading before since I've just been copying it between files.

e2e/cypress/integration/messaging/message_reaction_gm_spec.js
hmhealey: Actually, this seems to be failing currently based on those steps since it says the button should appear without hovering in mobile view. Should I submit a fix for that at the same time?

hmhealey: I fixed the visibility of that button. Some CSS got renamed when the animations were added, and that broke the visibility in mobile view.

srkgupta (approved):
Thanks @hmhealey. The changes looks good and is working fine. Approving the PR.

e2e/cypress/integration/messaging/message_reaction_gm_spec.js
```diff
@@ -0,0 +1,85 @@
...
+            cy.get('#channelIntro').click();
+
+            // * Verify that the Add Reaction button is no longer visible
+            cy.findByLabelText('Add a reaction').should('not.be.visible');
+
+            // # Resize window to mobile view
+            cy.viewport('iphone-6');
+
+            // * Verify that the Add Reaction button is once again visible
+            cy.findByLabelText('Add a reaction').should('be.visible');
```
saturninoabril, line 82: @hmhealey We had some discussions about adopting written scripts against the original test case in TM4J, especially on how to break the test into multiple steps or extract into its own test case. Just like to chime in here before updating the TM4J. Would it be possible to combine above steps in the first `it` block? If no, can we run the second `it` block in isolation by running `it.only`?  Thanks

hmhealey: Yeah, I can combine the two, but would it be better to maybe repeat the setup steps for the second one and keep them isolated? I was thinking of the two as covering different pieces of functionality, but the second one certainly won't run without the first. I'm not sure when we prefer to reuse test setup to save time and when we want the tests to remain isolated since I think we tend towards reuse normally

saturninoabril: Yeah, I think it's fine and I agree on keeping test isolated. I'll go on test isolation first then reusability. It makes it easier to deal with during test maintenance.  Thanks Harrison!

hmhealey: I ended up just combining the test cases for now since I've been focused on some other stuff. When I get back to unit tests, I might look at splitting the tests up again.

jgilliam17 (approved):
Thank you @hmhealey 
Looks good to merge.
- Test case details, steps and links updated.

sudheerDev (approved):
LGTM

sudheerDev:
@saturninoabril Not sure if you have any pending comments.

saturninoabril:
Nothing. I think all are good now. Feel free to merge anytime.

PR #6314, opened 2020-08-29
MM-T1794 Permalink post view combined with New Message toast

josephk96:
Summary

I created a test to make sure that after clicking on a permalink and posting a new message as a second user and clicking on the 'Jump to New Messages' button, the new message is visible

Test case: https://automation-test-cases.vercel.app/test/MM-T1794


saturninoabril (approved):
Thanks @josephk96, LGTM!

PR #6319, opened 2020-08-29
MM-27861 signal successful login to other tabs

bhargav50:
#### Summary
Signals login event to other tabs on successful login

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15302

### Changes introduced
- calls signalLogin in componentDidUpdate of logged_in component
- reloads another tab on focus after it receives signalLogin event

### Unit tests added
- logged_in component should call signalLogin once loggedIn state is received from redux store
- root component should auto login on focus after it receives signalLogin


jasonblais:
Thank you @bhargav50!

components/root/root.jsx
```diff
@@ -118,15 +122,19 @@ export default class Root extends React.PureComponent {
                 console.log('detected logout from a different tab'); //eslint-disable-line no-console
                 GlobalActions.emitUserLoggedOutEvent('/', false, false);
             }
+            if (e.originalEvent.key === StoragePrefixes.LOGIN && isNewLocalStorageEvent(e)) {
+                const isLoggedIn = getCurrentUser(store.getState());
 
-            if (e.originalEvent.key === StoragePrefixes.LOGIN && e.originalEvent.storageArea === localStorage && e.originalEvent.newValue) {
                 // make sure it isn't this tab that is sending the logout signal (only necessary for IE11)
-                if (BrowserStore.isSignallingLogin(e.originalEvent.newValue)) {
+                if (BrowserStore.isSignallingLogin(e.originalEvent.newValue) || isLoggedIn) {
```
hmhealey, line 129: Why make this change to look if we have a current user? We might not actually need this if block at all since we've dropped IE11 support unless you needed this on other browsers as well.

components/root/root.jsx
```diff
@@ -118,15 +122,19 @@ export default class Root extends React.PureComponent {
...
                     return;
                 }
 
-                console.log('detected login from a different tab'); //eslint-disable-line no-console
-                location.reload();
+                // detected login from a different tab
+                function onVisibilityChange() {
+                    location.reload();
+                }
+                document.addEventListener('visibilitychange', onVisibilityChange, false);
```
hmhealey, line 137: Why change this to only trigger when the window comes back into view?

components/root/root.test.jsx
```diff
@@ -79,6 +79,7 @@ describe('components/Root', () => {
 
         wrapper.instance().onConfigLoaded();
         expect(props.history.push).toHaveBeenCalledWith('/signup_user_complete');
+        wrapper.unmount();
```
hmhealey, line 82: Why are you unmounting the wrapper at the end of these tests? I didn't think that was necessary to clean up Enzyme.

components/root/root.test.jsx
```diff
@@ -201,5 +208,31 @@ describe('components/Root', () => {
...
+
+        window.dispatchEvent(loginSignal);
+        document.dispatchEvent(new Event('visibilitychange'));
+        expect(window.location.reload).toBeCalledTimes(1);
+        wrapper.unmount();
+    });
+
+    afterEach(() => {
+        if (window.location && window.location.reload && window.location.reload.mockRestore) {
+            window.location.reload.mockRestore();
```
hmhealey, line 235: I think this happens already with how we have Enzyme configured.

components/root/root.test.jsx
bhargav50: This was added because location reload was called 9 times as there are 9 tests. If I don't call unmount on the wrapper then componentWillUnmount won't exexcuted and so window event listener wasn't removed. So there will be 9 listeners because each test case is adding the listener on storage event

components/root/root.jsx
bhargav50: This was written in a comment on Jira: https://mattermost.atlassian.net/browse/MM-27861?focusedCommentId=96638

components/root/root.jsx
bhargav50: If I don't check for current user while reloading, one tab will send loginSignal to another, another tab will again send to source tab and all tabs will keep on reloading....
So the check is performed to only reload if the tab is not logged in previously

bhargav50: I have removed checking for IE 11 specific constraints in latest changes

components/root/root.test.jsx
bhargav50: This is also only to support requirements to only auto login on focus, if we just do immediate login then these unmount calls are not required.

components/root/root.jsx
hmhealey: Oh right, I forgot about that discussion. That works for me then.

hmhealey (approved):
Thanks for answering my questions. This looks good to me then!

nevyangelova (approved):

nevyangelova:
LGTM can you resolve the conflict and sync with master? Thanks for the contribution 👍

jgilliam17 (approved):
Thank you @bhargav50 
Tested, looks good to merge.
- Verified login is signaled to other tabs (Windows, macOS and Linux browsers). 
After user has been logged in, login on other tabs completes when focus is placed on those tabs and not immediately at the time of initial login.
@nevyangelova Can you please help merge? Thanks.

bhargav50:
I have resolved conflicts with master and pushed but one check is still failing. Can you help me figuring out what is this actually suggesting?

jasonblais:
@nevyangelova @bhargav50  The failed build check was for an outdated branch, so I ran `/update-check` to rebase against master. It looks like one of the tests are failing now (though there weren't any code changes due to the rebase: https://github.com/mattermost/mattermost-webapp/pull/6319/commits/3d7e5b037bbe7f125bfe615fffed011afeb249f2).

hmhealey:
I've updated again to fix a merge conflict. Hopefully the build failure was just due to a random failure. I'll keep an eye on it to ensure it passes.

PR #6324, opened 2020-08-30
MM-T550 Words that trigger mentions - @-icon still shows mentions list if all triggers deselected

josephk96:
Summary

As a new user, I deselected all 'Words that trigger mentions', and with the sysadmin account I made sure that typing '@' in town square would still show the user's name, and typed a message tagging the user. I also verified that the @ button is visible, and clicking it would show the recent mentions in the RHS, along with the username in the search bar.
 
Test case: https://automation-test-cases.vercel.app/test/MM-T550


e2e/cypress/integration/notifications/at_icon_still_shows_mentions_list_with_deactivated_triggers_spec.js
```diff
@@ -0,0 +1,90 @@
...
+        cy.apiLogout();
+
+        // # Login as otherUser and visit the team
+        cy.apiLogin(otherUser);
+        cy.visit(`${testTeam.name}`);
+
+        // # Click on the @ button
+        cy.get('#channelHeaderMentionButton', {timeout: TIMEOUTS.HALF_MIN}).should('be.visible').click();
+
+        // * Ensure that the user's name is in the search box after clicking on the @ buton
```
saturninoabril, line 81:
```suggestion
        // * Ensure that the user's name is in the search box after clicking on the @ button
```

saturninoabril (changes requested):
@josephk96 LGTM, except that I don't see the test user added to private channel. Please take a look again.  Thanks!

prapti (changes requested):
@josephk96 There seems to be a mismatch in the title and test case linked to this PR [MM-T550](https://automation-test-cases.vercel.app/test/MM-T550), and the test that has been implemented. Also the `it` statement says 'MM-T546'. Could you please recheck these?
@saturninoabril The test steps for MM-T550 where it says "Deselect all words that trigger mentions" -- I understood this test as going into Account Settings > Notifications > Words That Trigger Mentions and unselecting all the options. Then verifying that clicking the '@' button by the search box still showed the mentions. Do you think that's true or should we clarify the test case with someone?

josephk96:
Hi @prapti , you are absolutely right, I used the wrong test case in the `it` statement. I'll fix it when I get home :) Thanks for pointing this out!

saturninoabril:
@prapti Yeah, I think it's true but feel to clarify. And thanks for catching up the mismatch. That explains why I'm not seeing user being added to private channel 'cause I'm looking at the wrong MM-T546 test case :)

josephk96:
I fixed the test case description. Sorry again for the mistake

saturninoabril (approved):
Thanks @josephk96! LGTM, tested and passed.

prapti (approved):
Thank you both for revisiting!
Looks great! Approving 🎉

PR #6325, opened 2020-08-30
MM-T546 Words that trigger mentions - Deselect username, still get mention when added to private channel

josephk96:
Summary

This test checks that when a user deselects 'username' in 'Words That Trigger Mentions' and is added to a private channel, the user will still get a mention.

Test case: https://automation-test-cases.vercel.app/test/MM-T546


saturninoabril (approved):
Thanks @josephk96, LGTM!

PR #6327, opened 2020-08-30
MM-T562 New message bar - Message posted while scrolled up in same channel

josephk96:
Summary

This test checks that when scrolled to the top of a channel and there are new messages, the 'Jump To New Messages' button will appear and that clicking on the button will direct to the bottom of the channel so that the new message would be visible and the button would not be visible anymore after clicking.

Test case: https://automation-test-cases.vercel.app/test/MM-T562


srkgupta (approved):
Thanks @josephk96. The test looks great and is working fine. Approving the PR. Great work.

saturninoabril (approved):
Thanks @josephk96, LGTM!

PR #6335, opened 2020-08-31
[MM-27813] Add Cypress Test Case for MM-T552

ashishbhate:
#### Summary

- Adds cypress test for MM-T552 Reply Notifications - threads that I start 
- Minor refactoring to use library functions and better test file names

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-27813


ashishbhate:
@josephbaylon thanks for catching that. 

I've reverted the change that was causing that failure. I thought my refactoring precluded the need for `ignoreUncaughtException` but it appears to still be required.

ogi-m (approved):
LGTM 👍

PR #6337, opened 2020-08-31
[MM-27274] Enforce convert_public_channel permission for converting public channels to private

fmunshi:
#### Summary
- Adds permissions `convert_public_channel_to_private` and uses it to enforce converting a public channel to private
- Adds permission `convert_private_channel_to_public`
- These permissions have been default added to roles that contain `manage_team` which ensures the existing behaviour will not change since `team_admin` and `system_admin` roles had that permission before and now only those two roles will have this new permission by default.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27274

#### Related Pull Requests
- Has server changes: https://github.com/mattermost/mattermost-server/pull/15370
- Has redux changes: https://github.com/mattermost/mattermost-redux/pull/1226
- Related mobile PR: https://github.com/mattermost/mattermost-mobile/pull/4756

#### Screenshots
Screen Shot 2020-09-03 at 1 04 33 AM


components/admin_console/permission_schemes_settings/permissions_tree/permissions_tree.jsx
```diff
@@ -79,6 +79,7 @@ export default class PermissionsTree extends React.PureComponent {
                         ],
                     },
                     Permissions.DELETE_PUBLIC_CHANNEL,
+                    Permissions.CONVERT_PUBLIC_CHANNEL,
```
jespino, line 82: CONVERT_PUBLIC_CHANNEL for me sounds not very explicit. I would prefer something like `CONVERT_PUBLIC_CHANNEL_TO_PRIVATE` or something like that.

jespino (changes requested):
I tiny detail about the name, otherwise, looks good to me. If it is just me being picky with the name, we can move forward, but I think we can do better with the permission name.

jespino (approved):

fmunshi:
@michaelgamble can you confirm that the wording is what you expect

e2e/cypress/integration/channel/convert_channel_to_private_spec.js
```diff
@@ -0,0 +1,165 @@
...
+    let testTeam;
+
+    before(() => {
+        cy.apiInitSetup().then(({team, user}) => {
+            testUser = user;
+            testTeam = team;
+        });
+    });
+
+    it('MM-28274 - Convert to private channel should only be shown to users with permission', () => {
```
fmunshi, line 23: @srkgupta I have added E2E tests here, it tests 4 different things
- Regular system user with the new permission _can_ convert public channels to private
- Regular system user _without_ the new permission _cannot_ convert public channels to private
- Team admin without the permission _cannot_ convert public channels to private
- Team admin with the permission _can_ convert public channels

srkgupta: Awesome. Thanks @fmunshi

srkgupta, line 23: Can you please rename the test id from `MM-28274` to `MM-T3348`, since now we have to map to the TM4J test ID instead of JIRA ticket ID.

srkgupta (approved):
Awesome. Thanks @fmunshi. Tested the PR and the Cypress test and it is working fine. Just have a minor suggestion in the Cypress test ID. Can you please change the same before you merge the PR. Approving the PR.

e2e/cypress/integration/channel/convert_channel_to_private_spec.js
fmunshi: This is done now, just waiting on @michaelgamble to review then I can merge...

michaelgamble (approved):
looks good :)

PR #6344, opened 2020-08-31
MM-T555 Notification Preferences do not save when modal is closed without saving

josephk96:
Summary

This test clicks on Account Settings -> Notifications -> Email Notifications -> Send Email Notifications -> Never without saving, closes the Account Settings Modal, reopens it, and verifies that the setting was not saved.

Test case: https://automation-test-cases.vercel.app/test/MM-T555


saturninoabril (approved):
Thanks @josephk96! Tested and passed. LGTM.

@josephbaylon The one you're seeing is valid bug with fix https://github.com/mattermost/mattermost-webapp/pull/6355

PR #6356, opened 2020-09-02
Cypress/E2E: AddTM4J ID to common slash commands

saturninoabril:
#### Summary
AddTM4J ID to common slash commands - MM-T573 and MM-T574


srkgupta (approved):
LGTM. Thanks @saturninoabril

PR #6358, opened 2020-09-02
[MM-27207] e2e archived channels (I)

Willyfrog:
#### Summary
5 of the test for archived channels (I), I realized late that it was specified to send 1 PR per TM4J

#### Ticket Link

[MM-27207](https://mattermost.atlassian.net/browse/MM-27207)

Tests:

1. [MM-T1670](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T1670)
1. [MM-T1671](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T1671)
1. [MM-T1672](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T1672)
1. [MM-T1673](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T1673)
1. [MM-T1674](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T1674)


e2e/cypress/integration/channel/archived_channels_spec.js
```diff
@@ -55,5 +60,159 @@ describe('Leave an archived channel', () => {
...
+            // # Search for a post in an archived channel
+            cy.get('#searchBox').focus().clear();
+
+            cy.get('#searchBox').should('be.visible').type(`${testArchivedMessage}{enter}`);
+
+            // # Open the archived channel by selecting Jump from search results and then selecting the link to move to the most recent posts in the channel
+            cy.get('#searchContainer').should('be.visible');
+
+            cy.get('a.search-item__jump').first().click();
+            cy.wait(TIMEOUTS.ONE_SEC); // let the page load
```
hmhealey, line 128: The waits in these tests don't seem necessary. On the next check, Cypress will wait until that sidebar item is visible, so we shouldn't need an explicit wait.

e2e/cypress/support/ui_commands.js
```diff
@@ -432,3 +446,54 @@ Cypress.Commands.add('checkRunLDAPSync', () => {
         }
     });
 });
+
+Cypress.Commands.add('createArchivedChannel', (options, messages, memberIds) => {
```
hmhealey, line 450: Maybe this could be split into two commands for `uiCreateChannel` and `uiArchiveChannel`? I'm surprised we don't actually have a `uiCreateChannel` already

hmhealey (changes requested):

e2e/cypress/support/ui_commands.js
saturninoabril: Ah, yeah I think it makes sense to add those two commands. Also, I'd suggest to make it specific with the UI flow so it can be used in other places. If it's a combination of everything - UI flow and several API calls - and to be used by particular spec file then I'd suggest not to make `createArchivedChannel` as command but a helper function in the spec file itself.

e2e/cypress/integration/channel/archived_channels_spec.js
Willyfrog: originally there wasn't any, but if the page doesn't fully load, it gets into a race condition where it is still loading but interacting with part of it, so I ended up with ugly waits, I can remove them, but it might fail sometimes.

e2e/cypress/support/ui_commands.js
Willyfrog: most likely since there is an apiCreate, most people just go with that. But sure, I'll recreate this as a helper function in the spec and have uiCreateChannel and uiArchiveChannel as commands.

e2e/cypress/integration/channel/archived_channels_spec.js
hmhealey: Instead of a wait for a time, you could wait for a certain element to be visible, like I'm waiting in my tests for the sidebar to be open with `cy.get('#sidebarContainer').should('be.visible');`

Willyfrog:
@hmhealey @saturninoabril I moved  everything into separate ui actions and added a helper function on the file. Let me know if that's ok.

Regarding `createChannel`, I decided to let the tester decide if it should be using the new or the old sidebar (it changes only how to start) as he is the one to know which one to use. guessing was getting complicated and might lead to bad results.

e2e/cypress/integration/channel/archived_channels_spec.js
saturninoabril: I agree with HH. Also the succeeding ``cy.get(`#sidebarItem_${testChannel.name}`)`` will fail after 20s (far greater than 1s) which is the [default command timeout](https://github.com/mattermost/mattermost-webapp/blob/master/e2e/cypress.json#L4). But I understand that there are some cases where explicit wait is really needed.

e2e/cypress/support/api/user.js
```diff
@@ -256,3 +256,10 @@ Cypress.Commands.add('apiPromoteGuestToUser', (userId) => {
         return cy.apiGetUserById(userId);
     });
 });
+
+Cypress.Commands.add('apiGetPreferences', (userId) => {
```
saturninoabril, line 260: I don't see this used elsewhere but may help in the future. So, I'm fine to include as is. However please move into `support/api/preference.js` as we're currently arranging API commands and would like to match the structure per [API reference](https://api.mattermost.com/). Also, please add type at `preference.d.ts`.

e2e/cypress/support/ui_commands.js
```diff
@@ -432,3 +447,49 @@ Cypress.Commands.add('checkRunLDAPSync', () => {
         }
     });
 });
+
+Cypress.Commands.add('createChannel', (name, isPrivate, purpose, header, isNewSidebar) => {
```
saturninoabril, line 451: Please change to `uiCreateChannel` and move to appropriate file at `/support/ui/` folder. (Not sure if better to create a new `/support/ui/channel.js` so I'll defer to you).

Note that the `ui` prefix was agreed upon to easily identify that a command is done via UI flow and to easily discover via autocomplete/intellisense.

Same comment for other UI commands.

saturninoabril (changes requested):
Thanks @Willyfrog! All tests passed and matched with those in TM4J 🎉 . I'm all good except for few suggestions.

e2e/cypress/support/api/user.js
Willyfrog: originally I was trying to get if the new sidebar was active to choose one or the other, but ended up rejecting the idea as it was getting complicated to detect and I thought it would be better to specify in the test if that was desired.

But as you, I thought it might come useful in the future and didn't delete it

e2e/cypress/integration/channel/archived_channels_spec.js
Willyfrog: i'll try to remove them, and see if it now passes (the test has changed a lot since I added them so some parts of it might not fail now)

Willyfrog: so, I finally figured out what was causing the problem. Once we use the search and jump from the results, we get a permalink, after some time it changes into the actual channel url, since tests sometimes go fast, there was a race condition when I closed the archived channel, sometimes it would have changed the url to the one after close and sometimes the timeout to change the url to the channel would come after.

I've added a check to wait for url change before doing anything else and it seems to have fixed the problem.

hmhealey: Yeah, that URL change was causing some issues in my tests as well. I wonder if we'll want to wait for a scroll or something to trigger that since just having it change automatically might make stuff flaky.

hmhealey (approved):

saturninoabril (approved):
Thanks @Willyfrog! LGTM, tested and passed. Approving now but left comment to fix typo.

sudheerDev (approved):
LGTM

PR #6369, opened 2020-09-03
Add eslint-plugin-no-only-tests

saturninoabril:
#### Summary
On local dev, running test with `.only` or `.skip` saves time in iterating over a test. However, sometimes it's being submitted as is and causes other tests being skipped if missed during reviews. This PR adds eslint rule that fails when found during `npm run check`. I've seen it happened both with Jest unit test and Cypress E2E.

Screen Shot 2020-09-03 at 3 00 51 PM

#### Ticket Link
none


hmhealey (approved):
I didn't notice we lost the no-exclusive-tests ESLint rule when we switched away from Mocha. Looks good to me.

josephbaylon (approved):
LGTM

PR #6378, opened 2020-09-03
Allow enable_notification_step to be completed in private browsers

nickmisasi:
#### Summary
When in a private browser, Notification.permission is automatically set to denied. This means that if a user is onboarding in an incognito window, clicking the `Enable Notifications` button will do nothing. Wrapping in a try catch allows the step to be completed once the button has been pressed. 

A more permanent solution here would be to detect private browsers and not show the option somehow - this is not trivial and will need more investigation.

#### Ticket Link
n/a, cloud bug bash

#### Related Pull Requests
- Has server changes n/a
- Has redux changes n/a
- Has mobile changes n/a

#### Screenshots
image


marianunez (changes requested):
I am hesitant to show the `Turn Notifications` step completed given that it gives the impression that we actually did turn them on. 

As discussed offline with @nickmisasi, we can completely omit the step for this border case for an initial solution and possibly in the future have some clear UX indication that the step was skipped due to browser restrictions? 

Any thought on this @matthewbirtch ?

nickmisasi:
@matthewbirtch (sorry for the double mention but I already had this typed out) 

Looking for some feedback here. When the user has either 
a) already disabled notifications
b) logged into mattermost via incognito or other private browser
The enable notifications button in its current state will not do anything. There are a couple options on what to do here:
1. This PR, which just moves on if they've clicked the button, marking it as completed (regardless of whether they've actually been enabled or not)
2. We could tweak the step for enabling notifications to not show if the user has already disabled them, or is in incognito
3. Figure out a way to "disable" the step with some sort of colour fade, would need design input on this.

matthewbirtch:
Hey @nickmisasi and @marianunez. Here are my thoughts on the following cases:

a) Notifications previously disabled: I think enabling is important enough that we should try to ask users to re-enable them again. So, in this case, I think we should still show the step. Assuming it's possible to trigger the browser notification dialog again if they've previously disabled?

b) Incognito mode: Can we still show the step, but change the messaging to "Desktop notifications cannot be enabled in private browsing mode." and change the button to read 'skip for now'. Then, if they return in regular browser mode, we could show them a global banner encouraging them to enable notifications (similar to [this screen](https://www.figma.com/file/uv9A6dZsekQHHhOhYC7yc2/Cloud-UX?node-id=39%3A9168) I designed)

Thoughts?

nickmisasi:
Thanks for the input @matthewbirtch I have a few comments below:

a) Once a user has denied/disabled notifications in the browser, you cannot get the browser to ask for permission again. The only way for the user to enable notifications after that is through the browsers settings for that environment: 
image

b) From some quick research, there is no officially supported way of detecting if a browser is a private one or not. In fact, [Google actively removes](https://www.cnet.com/news/chrome-76-adds-dark-mode-removes-incognito-mode-detection/#:~:text=Google%20released%20Chrome%20version%2076,privacy%20feature%20is%20being%20used.) workarounds for detecting chrome incognito. [There are still workarounds](https://gist.github.com/jherax/a81c8c132d09cc354a0e2cb911841ff1) but as you can see in the comments on that page, these workarounds could break for any supported browser at any time. The way the browser acts in the case of it being incognito would be the same as if the user had previously denied notifications. Our request for permission immediately returns as denied. 

I'm not sure if this information changes your proposed designs. Some sort of hybrid approach where the step is shown, but it's greyed out would be possible, but the differentiation between denied and incognito is where it would get tricky to support long-term.

matthewbirtch:
Thanks @nickmisasi that all makes sense. I wouldn't propose two different solutions then. I think we would probably want to update the messaging to say something like:

**Browser**
"Notifications were previously disabled or you are browsing in private mode. View your browser's notifications settings to re-enable notifications." 
[view browser settings] (opens chrome notifications settings if possible)

**Desktop App**
"Notifications were previously disabled. View your operating system's notifications settings to re-enable notifications." [view system settings] (opens OS notification settings if possible)

Would this work?

nickmisasi:
@matthewbirtch Looks like there isn't a way to open the Chrome settings from inside a web page: https://security.stackexchange.com/questions/171572/why-cant-i-open-the-chrome-settings-page-from-javascript 

I've verified this with a quick HTML page:
`<a href="chrome://settings" target="_blank">open settings</a>` displays the following error in your console: 
image
and am seeing similar results for firefox and safari.


With the desktop app, I believe we control things so it might be possible. I'll need to test this out and verify first!

matthewbirtch:
@nickmisasi I wondered about that. I guess we'll have to modify to something like this then:

**Browser**
"Notifications were previously disabled or you may be browsing in private mode. You'll need to open your browser settings or turn off private mode to enable notifications."
No button

**Desktop App**
"Notifications were previously disabled. View your operating system's notifications settings to re-enable notifications." [View System Settings] (opens OS notification settings if possible)

marianunez (approved):
Thanks @nickmisasi!

stevemudie (approved):
LGTM

matthewbirtch (approved):
Looks good! Thanks @nickmisasi

PR #6391, opened 2020-09-07
Add custom slash command tests

iomodo:
#### Summary
This PR will add e2e tests for custom slash commands.

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-28258


e2e/cypress/integration/integrations/custom_slash_commands/slash_commands_spec.js
```diff
@@ -0,0 +1,213 @@
...
+    });
+});
+
+function addNewCommand(trigger) {
+    // # Add new command
+    cy.get('#addSlashCommand').click();
+
+    // # Type a trigger word, url and display name
+    cy.get('#trigger').type(`${trigger}`);
+    cy.get('#url').type('http://hidden-peak-21733.herokuapp.com/test_inchannel');
```
saturninoabril, line 184: I didn't know that there's a live URL like this but I suggest not to use it for Cypress testing. We usually spin a local node server for webhook testing by `npm run start:webhook`.

Then, url could be generated by,
```javascript
cy.apiGetChannelByName(testTeam.name, 'town-square').then((res) => {
    const url = `${Cypress.env('webhookBaseUrl')}/send_message_to_channel?channel_id=${res.body.id}`;
});
```

Be sure to add verification in before hook so it'll fail immediately when no node server is running/accessible.
```javascript
before(() => {
    cy.requireWebhookServer();
});
```

saturninoabril (changes requested):
Thanks @iomodo! Overall looks good and all tests passed. But left a comment to use local node server instead of remote one as a test URL indicated in the test case.

e2e/cypress/integration/integrations/custom_slash_commands/slash_commands_spec.js
```diff
@@ -0,0 +1,226 @@
...
+        // * Verify that we get the error message
+        cy.findByText('A trigger word cannot begin with a /').should('exist').and('be.visible').scrollIntoView();
+    });
+
+    it('MM-T691 Error: trigger word required', () => {
+        // # Add new command
+        cy.get('#addSlashCommand').click();
+
+        // # Do not input trigger word
+        cy.get('#url').type('http://hidden-peak-21733.herokuapp.com/test_ephemeral');
```
saturninoabril, line 59: One last request. Could you change all reference to this URL, maybe use `http://example.com` instead?  Thanks

saturninoabril (approved):
Thanks @iomodo! LGTM, tested and passed, except for one last request. Approving now.

jfrerich (approved):

DHaussermann (approved):
Tested and passed.
- Reviewed test to ensure guiltiness and practices are followed
- Ran tests locally - All ran green
LGTM!

PR #6403, opened 2020-09-08
[MM-28207] Require channel id for textbox and post markdown components

fmunshi:
#### Summary
- Fixes broken preview post functionality where channel id was not passed and modified existing component definitions to require a channelId as all these components always have a channelId (unless im missing something)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28207


hmhealey:
Could we pass the channel ID to the post preview instead? I think the `CreatePost` component and co should all be attached to a channel.

fmunshi:
@hmhealey I modified a couple of component definitions to make `channelId` a required prop here

I think I got it all but let me know if you think I missed something...

isacikgoz (approved):
Looks good from my side 👍

hmhealey (approved):

stevemudie (approved):
LGTM

PR #6384, opened 2020-09-03
[MM-28328][MM-28329] Click "Getting Started" in lower LHS to go back to Next Steps View

nickmisasi:
#### Summary
When clicking away from the getting started or tips & next steps experiences the user couldn't return to it by clicking the Getting Started or Tips and Next Steps section in the lower LHS. You can now do this by clicking in that area.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28328
https://mattermost.atlassian.net/browse/MM-28329
#### Related Pull Requests
n/a
#### Screenshots
Mousing over once clicking away should turn to a cursor (I can't screenshot a cursor)
image

PR #6407, opened 2020-09-08
[MM-28328] Move call from render function to componentDidMount to prevent race condition

nickmisasi:
#### Summary
This if statement was run in a render component which meant there could be a race condition preventing the animation from playing when the transition happens between next_steps and next_steps_tips. Moving into componentDidMount remedies this problem.

#### Ticket Link
Fix related to #6384 

#### Related Pull Requests
#6384 
- Has server changes n/a
- Has redux changes n/a
- Has mobile changes n/a

#### Screenshots
n/a

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13297,6 +13297,14 @@
       "dev": true,
       "requires": {
         "icu4c-data": "0.64.2"
+      },
+      "dependencies": {
+        "icu4c-data": {
+          "version": "0.64.2",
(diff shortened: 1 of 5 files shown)


stevemudie (approved):
LGTM

package-lock.json
```diff
@@ -14304,12 +14312,6 @@
         "postcss": "^7.0.14"
       }
     },
-    "icu4c-data": {
```
ali-farooq0, line 14307: should this change be reverted?

ali-farooq0 (approved):

devinbinnie (approved):

PR #6418, opened 2020-09-09
Cypress/E2E: Fix TM4J ID for MM-T2500

saturninoabril:
#### Summary
Fix TM4J ID for MM-T2500

Note that no test script change here other than removing the other `describe` block and adding TM4J ID with step number for each `it` block.


josephbaylon (approved):
LGTM. Tests passed.

prapti (approved):
LGTM! Tests passed and the test case id changes look good.

PR #6305, opened 2020-08-28
Cypress/E2E: Fix failing tests due to new preference introduced by cloud onboarding

saturninoabril:
#### Summary
Fix failing tests due to new preference introduced by cloud onboarding

Other:
- hide what's new modal via API

PR #6427, opened 2020-09-11
Cypress/E2E: Fix e2e new sidebar

saturninoabril:
#### Summary
Fix failing E2E test in new sidebar

### DIFF
--- a/e2e/cypress/integration/websocket/channel_created/new_sidebar_spec.js
+++ b/e2e/cypress/integration/websocket/channel_created/new_sidebar_spec.js
@@ -26,6 +26,9 @@ describe('Handle removed user - new sidebar', () => {
         const teamName = `team-${getRandomId()}`;
         cy.createNewTeam(teamName, teamName);
 
+        // # Close "What's new" modal
+        cy.uiCloseWhatsNewModal();
+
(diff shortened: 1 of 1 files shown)


srkgupta (approved):
Changes LGTM and test passed. @saturninoabril my only question is why isn't this change also pushed to master and only to 5.27?

saturninoabril:
It's already fixed on master - https://github.com/mattermost/mattermost-webapp/pull/6305 - and it's done via API command. However, lots have changed on master already and others might not be applicable to 5.27 (given it's cut from 5.26). I think it would be better to do quick fix with minimum code change as much as possible.

srkgupta:
> It's already fixed on master - #6305 - and it's done via API command. However, lots have changed on master already and others might not be applicable to 5.27 (given it's cut from 5.26). I think it would be better to do quick fix with minimum code change as much as possible.

Got it. Thanks. Makes sense

josephbaylon (approved):
LGTM. Although the latest run on Cypress passed. @saturninoabril I'll let you decide whether to merge.

saturninoabril:
It passed because I added this in advance 😄

PR #6428, opened 2020-09-11
[MM-27397] Add marketplace search term to telemetry event

hanzei:
#### Summary
Send the search term as part of the telemetry event. This allows for better analytics on which plugins are requested.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27397


larkox (approved):

components/plugin_marketplace/marketplace_modal.test.js
```diff
@@ -140,5 +150,17 @@ describe('components/marketplace/', () => {
...
             expect(wrapper).toMatchSnapshot();
         });
+
+        test('Should call for track event when searching', () => {
+            const wrapper = shallow(
+                <MarketplaceModal {...baseProps}/>,
+            );
+
+            wrapper.setState({filter: 'nps'});
+            wrapper.instance().doSearch();
```
hanzei, line 160: This isn't a pretty way to test as it's doesn't test the surface of a component.

I tried triggering the search button directly, which worked, but `refs` /seem/ to be `nil` for `wrapper`. That caused https://github.com/mattermost/mattermost-webapp/blob/dd32c9b68e4b54565ba4118e1be67c3d597a0bdf/components/plugin_marketplace/marketplace_modal.js#L151 to fail.

Any advice is appreciated.

lieut-data (approved):

hanzei:
@DHaussermann I'm not sure if you can test this PR pre-merge. Maybe it's easier to test post-merge in RC testing.

amyblais:
@DHaussermann Kind reminder to review for v5.28.

DHaussermann:
@hanzei  I'm confused by this PR. 
1. On this cloud server I have spun up, will it contain the QA Rudder key or should I do this on a server where I can set the key and data plane. 
2. Is there a corresponding market place build for this?

DHaussermann:
@hanzei this is testable on the server above server ID `3d1iiyb19fdsbx3mo5rrmk7y7a`. I can  see events coming back but there seems to be an issue with `ui_marketplace_search`

I can see on this same server:
```
ui_marketplace_configure
ui_marketplace_download
ui_marketplace_opened
ui_marketplace_closed 
ui_marketplace_download_update    
```

But I see no corresponding search events (with or without search terms) on this server at all. Just to be clear, I did install several plugins using the search term to filter them from the list. I think this is the event we're looking for.
Screen Shot 2020-09-21 at 4 47 44 PM

Thoughts?

DHaussermann:
Disregard previous comment. the cloud server data is nor appearing on the QA dashboard. I was mistaken. 

I now have data from my local showing in Looker and I'm trying again to test this.

DHaussermann (approved):
Tested and passed
Search events now contain a `FILTER` field which will show the text value when not `nps`

LGTM!

DHaussermann:
@hanzei I missed something about this yesterday. Are we making an exception from the term `nps` here:
`expect(trackEvent).toHaveBeenCalledWith('plugins', 'ui_marketplace_search', {filter: 'nps'});`
If so, It's not clear to me why. Note that the "NPS" plugin is called "Users Satisfaction Surveys" and will not be returned when filter for "nps"

Can you please explain what's happening on this line and why it provides value?

hanzei:
@DHaussermann That line ist just an example used in a `*.test.js`. It's not part of the production code.

PR #6430, opened 2020-09-11
[MM-28458] Only show Cloud Admin Onboarding to the First Admin User

nickmisasi:
#### Summary
Some weirdness was happening with the admin onboarding when you had a system_user, and then promoted that user to admin. Through discussion on how to fix this we realized it wouldn't make sense to give every admin the ability to rename the team or generally go through the admin onboarding. This PR makes it so that only the first Admin sees the admin onboarding flow. This includes both the next steps, and the next tips views. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28458

#### Related Pull Requests

- Has server changes n/a
- Has redux changes n/a
- Has mobile changes n/a

#### Screenshots
n/a


ali-farooq0 (approved):
Nice 👍

components/next_steps_view/steps.ts
```diff
@@ -86,10 +86,34 @@ export const Steps: StepType[] = [
     },
 ];
 
+export const isFirstAdmin = createSelector(
+    (state: GlobalState) => getCurrentUser(state),
+    (state: GlobalState) => getUsers(state),
+    (currentUser, users) => {
+        const userIds = Object.keys(users);
+        for (const userId of userIds) {
```
devinbinnie, line 94: This feels a bit inefficient, especially on installations with 1000+ users.
Can we maybe filter down by users with the `system_admin` role and then sort them by their `create_at` date?
Then it's a simple O(1) check to see if the current user id == the first user id.

devinbinnie (changes requested):

components/next_steps_view/steps.ts
nickmisasi: This for loop is O(n), whereas filter + sort would be O(n) + O(nlogn) (and nlogn could be worse depending on sorting algorithm) I could change the for loop to be an array.prototype.filter, but then you would lose the early exit condition of finding a newer admin as the second, third, etc user in the list.

Is there a selector similar to getUsers that is only getAdmins? Or an entity that doesn't show users and only admins? That's the only thing I could see speeding this up - and it'd still be O(n) (short of SQL filters in the backend, but we're trying to avoid server changes)

stevemudie (approved):
LGTM

components/next_steps_view/steps.ts
devinbinnie: I don't think there is, but you're right that'd be the only thing to speed this up. I'd like to see a `filter` here personally instead, but I guess if the performance is the same there's no need to change it.

devinbinnie (approved):

PR #6442, opened 2020-09-14
[MM-28483] Surface custom about modal for cloud instances

nickmisasi:
#### Summary
The default about modal reveals a lot of information that a Cloud user will not care about (and probably shouldn't see for security reasons). This PR adds a custom about modal that appears for Cloud licenses. It utilizes the default AboutCloudModal as an entry-point.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28483

#### Related Pull Requests
n/a

#### Screenshots

image


components/about_build_modal/about_build_modal.jsx
```diff
@@ -12,6 +12,8 @@ import Nbsp from 'components/html_entities/nbsp';
 
 import {AboutLinks} from 'utils/constants';
 
+import AboutBuildModalCloud from './about_build_modal_cloud/about_build_modal_cloud';
```
devinbinnie, line 15: For future reference, if you're not connecting the component to redux, you don't need to create a subfolder for it.

components/about_build_modal/about_build_modal.jsx
```diff
@@ -59,6 +61,16 @@ export default class AboutBuildModal extends React.PureComponent {
     render() {
         const config = this.props.config;
         const license = this.props.license;
+        const isCloud = license.Cloud === 'true';
+
+        if (isCloud) {
```
devinbinnie, line 66: Can we move the boolean statement here, just to reduce code bloat?

devinbinnie (approved):
LGTM, just a couple small comments.

matthewbirtch (changes requested):
Nice work @nickmisasi. Two small things I noticed:
- Can we change the padding on `.modal-body` to be:  `padding: 15px 64px`
- If I switch to a dark theme, the line "High trust messaging for the enterprise" is hard to see. It needs to use the proper theme color for the center channel text color
image

Other than that, I think it's good to go!

components/about_build_modal/about_build_modal_cloud/about_build_modal_cloud.scss
```diff
@@ -0,0 +1,72 @@
...
+
+            }
+            .subtitle {
+                width: 294px;
+                height: 24px;
+                left: 0px;
+                top: 32px;
+                font-size: 16px;
+                line-height: 24px;
+                color: #3D3C40;
```
marianunez, line 46: We should use the theme color here and may be what Matt was referring to in dark mode.

components/about_build_modal/about_build_modal_cloud/about_build_modal_cloud.tsx
```diff
@@ -0,0 +1,122 @@
...
+    );
+
+    //
+    const licensee = (
+        <div className='form-group'>
+            <FormattedMessage
+                id='about.licensed'
+                defaultMessage='Licensed to:'
+            />
+            {'\u00a0' + license.Company}
```
marianunez, line 51: For company we want to show the company information when they signed up for cloud and not the company of the license because in this case is always Mattermost. 

We can add this later when we have the endpoints to get Company information from CWS which we will have for Billing in the System Console Soft GA. Created a ticket https://mattermost.atlassian.net/browse/MM-28698

marianunez (approved):
Looks great @nickmisasi! Non-blocking comment below

components/about_build_modal/about_build_modal_cloud/about_build_modal_cloud.scss
nickmisasi: Yeah definitely the culprit

nickmisasi:
Fixed the text, added the padding as requested @matthewbirtch 

With that new padding it doesn't necessarily match the design, distance between the top of the mattermost svg and the header is 46 pixels in figma, but now with 15 px padding on top its 61. Let me know what you think.

image
image

matthewbirtch:
> but now with 15 px padding on top its 61. Let me know what you think.

Ya, you're right. With the 15px on top, it feels off balance. I think you can make that top padding 0 actually. That should do the trick

nickmisasi:
@matthewbirtch back to 0px padding at the top on the next spinwick update. Let me know what you think!

stevemudie (approved):
LGTM

matthewbirtch (approved):
Looks good @nickmisasi. Thanks!

PR #6445, opened 2020-09-15
Possible fix for accessibility test for support in buttons in channel header

prapti:
The error is happening because of the new plugin icon that is recently added. Since this is an accessibility test, the best way to test is using keyboard tab to verify that the buttons in channel header can be accessed as such. Therefore, instead of adding any complex solution, only adding a tab to verify that the pin button can be accessed as expected.

#### Screenshots
Screen Shot 2020-09-14 at 6 16 14 PM

### DIFF
--- a/e2e/cypress/fixtures/partial_default_config.json
+++ b/e2e/cypress/fixtures/partial_default_config.json
@@ -473,6 +473,9 @@
         "PluginStates": {
             "com.mattermost.nps": {
                 "Enable": false
+            },
+            "com.mattermost.plugin-incident-response": {
+                "Enable":false
             }
         },
         "EnableMarketplace": true,
--- a/e2e/cypress/fixtures/partial_default_config.json
+++ b/e2e/cypress/fixtures/partial_default_config.json
@@ -475,7 +475,7 @@
                 "Enable": false
             },
             "com.mattermost.plugin-incident-response": {
-                "Enable":false
+                "Enable": false
             }
         },
         "EnableMarketplace": true,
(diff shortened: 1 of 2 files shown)


saturninoabril (changes requested):
Thanks @prapti. For me, the proper fix here is to disable incident response plugin by default at `partial_default_cofig.json`. This was recently enabled by https://github.com/mattermost/mattermost-server/pull/15360 (same as you mentioned earlier). There are several places where this affected some test, like in the main menu where it adds a plugin option -- made the cloud onboarding test failed too and probably other accessibility/permission tests there.

We can always do `cy.apiUpdateConfig` to enable the plugin on demand depending on test requirement.

prapti:
@saturninoabril I did think about that as well that the plugin would cause other failures, but the incident response plugin is going to be installed and enabled by default in the cloud servers from 5.28 going forward (because of the PR you mentioned). So I was thinking that the adjustments needed to be made to the other tests instead.
Thoughts?

saturninoabril:
My worry here is the difference between local development and CI. It's pre-packaged on compiled version but when running on source, like most contributors do, it needs to install manually. Not to mention but this will add overhead to external contributors since they are running without EE. That's why I think this should be enabled on demand, similar to NPS. Besides, there's a separate E2E for incident response to take care of tests and impact to core. But feel free to evaluate risk/impact and let us know.

prapti:
Thank you for putting that into perspective @saturninoabril. Yeah, in that case I think it only makes sense to disable the plugin esp. since the on-prem behavior is different from the cloud version :)
I'll go ahead and make those changes.

e2e/cypress/fixtures/partial_default_config.json
```diff
@@ -473,6 +473,9 @@
         "PluginStates": {
             "com.mattermost.nps": {
                 "Enable": false
+            },
+            "com.mattermost.plugin-incident-response": {
+                "Enable":false
```
saturninoabril, line 478:
```suggestion
                "Enable": false
```

saturninoabril (approved):
Thanks Prapti for updating and considering my suggestion. Left non-blocking comment.

srkgupta (approved):
Thanks @prapti. The changes LGTM

PR #6447, opened 2020-09-15
Unit Test Execution, Performance Tuning

calebroseland:
#### Summary

This PR tunes unit test scripts to make better use of processing power available (in both developer machine and CI environments).

`npm run`:
- `test`: remove `--forceExit --detectOpenHandles` and set `--maxWorkers` to 50%, which in most cases will allow up to the number of physical CPU cores available (due to hyper/multi threading). In my testing, more-workers-than-cores didn't yield significantly better performance. But, for those who want...
- `test:speed` will leave maxWorkers alone. [Jest docs](https://jestjs.io/docs/en/cli.html#--maxworkersnumstring) say # _cores_ available minus one, but I was observing # of _threads_ available minus one. <sub><sup>Maybe don't run this one while zoom is open 😅; your cooling fans may ramp up.</sub></sup>
- newly added `test:debug` retains `--forceExit --detectOpenHandles` for debugging unit tests.
- and finally, `test-ci` discards `--forceExit --detectOpenHandles`, gets `--ci` to skip writing forgotten snapshots, and `--maxWorkers=8` to make CI unit test execution roughly 5-6x faster, and about 1.7-2x faster for the CI pipeline as a whole. (This PR: `2m 17s` and `9m 34s` respectively.)

You can append to `npm test "search_pattern" --detectOpenHandles` (and/or other args) to retain if running subsets of tests.

Reasoning:

`--forceExit` and `--detectOpenHandles` are useful for debugging purposes, but I believe they should be used only when necessary (e.g. `test:debug`), largely because `--detectOpenHandles` "has a significant performance penalty and should only be used for debugging."

This is because`--detectOpenHandles` implies `--runInBand`, which executes tests serially (effectively passing `--maxWorkers=1`), which is a shame considering our [xlarge (CircleCI) executor has 8 vCPUs](https://circleci.com/docs/2.0/configuration-reference/#docker-executor), and most modern laptops/desktops have at least 4 (if not 6 or 8) cores available.

If all else fails in CI—and Jest hangs—CircleCI's default `10m` output timeout will kick in around the same time, if not a little quicker than a normal pass.


metanerd (approved):
This sounds like a great improvement!

saturninoabril (approved):
Great, looks good to me.

hmhealey (approved):

PR #6460, opened 2020-09-16
MM-28716: expose textbox's openWhenEmpty

lieut-data:
#### Summary
Incident response is using the exported textbox to allow slash command autocomplete in the playbook backstage. The underlying component has support for automatically opening, but requires exposing `openWhenEmpty` to do so.

While in here, I hardened the component against use in a context where there is no active channel defined, namely if the playbook backstage is loaded directly (and thus we have not yet "entered" the center channel space).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28716

### DIFF
--- a/components/__snapshots__/textbox.test.tsx.snap
+++ b/components/__snapshots__/textbox.test.tsx.snap
@@ -1,5 +1,122 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`components/TextBox should match snapshot with additional, optional props 1`] = `
+<div
+  className="textarea-wrapper textarea-wrapper--preview"
+>
+  <SuggestionBox
+    channelId="channelId"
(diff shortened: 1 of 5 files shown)


hmhealey (approved):

agarciamontoro (approved):
Nice!

As an aside, the SuggestionBox component seems to have way too many options.

hmhealey:
Yeah, it wasn't particularly well designed to begin with, and then it got reused in a bunch of different places requiring a bunch of extra props to add slightly different behaviours. I know Nevy really wants to rewrite it when she gets a chance.

prapti (approved):
Autocomplete suggestion gets automatically trigged when focusing on the slash command text box in playbook creation page. Also did some sanity test around text box in center channel and RHS.
LGTM!

PR #6475, opened 2020-09-17
MM-28695 - Hard break when editing a post and sending from preview mode

cpoile:
#### Summary
- Since we only hide the `EditPostModal`, `TextBox` and `PostMarkdown` will often be called with a null `post` and an undefined `channel`. So it can't be required, and we need to test for it in the `PostMarkdown` map state to props. 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-28695

#### Related Pull Requests
- https://github.com/mattermost/mattermost-webapp/pull/6403


cpoile:
@fmunshi Usually I do, but this has happened on the last couple webapp PRs -- if it keeps updating package-lock, and I'm on master, then shouldn't we commit the new package-lock?

fmunshi:
yep, happens to me too so not sure 🤷 , but last time i committed a small package-lock change similar to this I broke one of the steps in our CI...

cpoile:
I think we should check it in -- it's needs to be correct for the CI cache to work correctly. If it breaks the CI somewhere else, then we should look into that because it's passing the CI here. :)

fmunshi:
Sounds good to me

calebroseland (approved):
LGTM ✓

DHaussermann (approved):
Tested and passed. 
- Can send post from preview mode with and without markdown
- Can edit a post and send from preview with and without markdown
- Tested this on all channel types including DM
LGTM!

Added test coverage here: https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T3374

amyblais:
@cpoile This is ready to merge.

amyblais:
@cpoile Kind reminder that this is ready to merge.

cpoile:
@amyblais Thanks, must have missed this. :)

PR #6476, opened 2020-09-17
[MM-28790] allow plugins to open in a different tab

Willyfrog:
#### Summary

Allow plugins to open oauth, while letting the incidence response still open within the app

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28790


crspeller (approved):

hmhealey (approved):

hmhealey (changes requested):
Actually, can you add a unit test for this case?

utils/markdown/renderer.tsx
```diff
@@ -207,8 +207,10 @@ export default class Renderer extends marked.Renderer {
 
         output += `" href="${outHref}" rel="noreferrer"`;
 
+        const pluginURL = `${this.formattingOptions.siteURL}/plugins`;
+
         // Any link that begins with siteURL should be opened inside the app
```
lieut-data, line 212:
```suggestion
         // Any link that begins with siteURL should be opened inside the app, except when rooted
         // at /plugins, which is logically "outside the app" despite being hosted by a plugin.
```

utils/markdown/renderer.tsx
```diff
@@ -207,8 +207,10 @@ export default class Renderer extends marked.Renderer {
 
         output += `" href="${outHref}" rel="noreferrer"`;
 
+        const pluginURL = `${this.formattingOptions.siteURL}/plugins`;
```
lieut-data, line 210: To double check, do we normalize `siteURL` to remove trailing slashes that might be configured?

lieut-data (approved):

utils/markdown/renderer.tsx
Willyfrog: from what I see, we get the siteurl from https://github.com/mattermost/mattermost-webapp/blob/b99ee5511507678edc1e3adb7d813a9f5ae139fd/utils/url.tsx#L36-L37

hmhealey: And if it's coming from the server instead of the window, that also trims it
https://github.com/mattermost/mattermost-server/blob/master/config/client.go#L18

hmhealey (approved):

jgilliam17:
@Willyfrog 
When installing more than 1 plugin, there is an error on the Plugin Marketplace screen and also 502 bad gateway on refresh. 
Steps:
1) Open plugin marketplace
2) Click on install button for the first plugin
3) After first plugin installs click on the second plugin to install (observe error text, see attached image)
4) Refresh the page
Observed: Second plugin fails to install with `received invalid response from server`, on refresh displays 502 error
Note: Not able to reproduce on master.test

Screen Shot 2020-09-21 at 12 55 09 PM
Screen Shot 2020-09-21 at 1 29 20 PM

hanzei:
@jgilliam17 Good catch :+1: That is an unrelated issue. I've filled https://mattermost.atlassian.net/browse/MM-28949 to track it.

jgilliam17 (approved):
Thank you @Willyfrog 
Tested, looks good to merge.
-Verified plugin oauth opens in a separate tab, while allowing incident response to open in the app.
Also, thank you @DHaussermann for the plugin install/testing setup

PR #6483, opened 2020-09-17
Remove cloud billing from beta

devinbinnie:
#### Summary
Remove cloud billing entirely from beta branch


marianunez (approved):
We could've just removed the setting from the System Console but I guess this works too. Thanks!

PR #6495, opened 2020-09-18
MM-28365 Avoid crashing PostUtil.formatText invoked from plugin

kaakaa:
#### Summary
Add existing check for emojiMap in `formatText`, because emojiMap never be set when invoked from plugins.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15365
https://mattermost.atlassian.net/browse/MM-28365


hmhealey (approved):
Thanks for the PR, and extra thanks for adding a test as well!

nevyangelova (approved):

prapti (approved):
Looks good to me! No error seen while running a JS command in the webapp console with special chars or emojis. Also tested using Matterpoll as @hanzei suggested (thank you so much for the clarification again!). Works as expected.
Approving.

PR #6497, opened 2020-09-18
[MM-28727] UX changes for Cloud onboarding flow

nickmisasi:
#### Summary
This PR contains a bunch of flow changes for cloud onboarding flow. Highlight of what's changed:
```
If a user has the ‘Getting Started’ item visible in the LHS and they click the 'x' icon in it, the modal shows letting the user know they can get back to “Getting Started” through the Main Menu

“Getting Started” is removed from the LHS if they confirm to remove it in the modal

Main Menu should remain “Getting Started”
```
```
If a user has the 'Getting Started' section active and clicks “Skip Getting Started”, this transitions to become “Tips & Next Steps” and there is now no way back to the Getting Started Steps since the user has explicitly chosen to skip it

LHS Menu Item changes to “Tips & Next Steps”

Main Menu changes from “Getting Started” to “Tips & Next Steps”
```
```
If a user has “Tips & Next Steps” item in the LHS and they click the 'x' icon on it, the modal shows letting the user know then can get back to “Tips & Next Steps” through the Main Menu

“Tips & Next Steps” is removed from the LHS if they confirm to remove it in the modal

Main Menu shows “Tips & Next Steps”
```

Various conditional changes are included in this PR to allow the concept of showing next steps, or next step tips depending on state.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28727

#### Related Pull Requests
n/a

#### Screenshots
n/a


matthewbirtch (approved):
@nickmisasi works great! 

Once small nit (non-blocking): the height of the sidebar menu item changes height when we switch between 'Getting Started' and 'Tips & Next Steps'. Can we animate that height transition? No big deal if not, but wanted to check

stevemudie (approved):
LGTM

components/next_steps_view/__snapshots__/next_steps_view.test.tsx.snap
```diff
@@ -4,109 +4,5 @@ exports[`components/next_steps_view should match snapshot 1`] = `
 <section
   className="app__content NextStepsView"
   id="app-content"
->
```
devinbinnie, line 7: You may want to update the test to set the `show` state variable to `true`, otherwise this snapshot doesn't really have any meaning.

components/next_steps_view/next_steps_view.tsx
```diff
@@ -309,7 +325,8 @@ export default class NextStepsView extends React.PureComponent<Props, State> {
             <section
                 id='app-content'
                 className='app__content NextStepsView'
-            >
+            >{this.state.show &&
```
devinbinnie, line 328: Nit: code style, should be on the next line

components/next_steps_view/steps.ts
```diff
@@ -135,11 +155,35 @@ export const showNextSteps = createSelector(
...
 export const nextStepsNotFinished = createSelector(
     (state: GlobalState) => getCategory(state, Preferences.RECOMMENDED_NEXT_STEPS),
     (state: GlobalState) => getCurrentUser(state),
-    (stepPreferences, currentUser) => {
-        const checkPref = (step: StepType) => stepPreferences.some((pref) => (pref.name === step.id && pref.value === 'true') || !isStepForUser(step, currentUser.roles));
+    (state: GlobalState) => isFirstAdmin(state),
+    (stepPreferences, currentUser, firstAdmin) => {
+        let roles = currentUser.roles;
+        if (!firstAdmin) {
+            roles = 'system_user';
```
devinbinnie, line 184: It feels like using the roles isn't as reliable a way of determining system admin at this point.
It's not something we have to do now, but refactoring to work off of a single selector to check for `isFirstAdmin` would be ideal.

devinbinnie (approved):
A few non-blocking comments.

components/next_steps_view/__snapshots__/next_steps_view.test.tsx.snap
nickmisasi: Good point. Fixed.

components/next_steps_view/steps.ts
nickmisasi: isFirstAdmin is a single selector, the line you're referring to here basically changes which steps are shown based on that isFirstAdmin selector. I agree we could probably refactor this in the future to be a bit better - as is though, isFirstAdmin is a stop-gap, so I don't think worth fixing as part of this PR.

ali-farooq0 (approved):

PR #6505, opened 2020-09-20
[MM-27437][MM-28250][MM-28250] - Fix sync between emoji reactions from RHS and emoji picker

nevyangelova:
#### Summary
This introduced a fix to the sync between emoji reactions coming from RHS and posts, as well as selected from emoji picker

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27988
https://mattermost.atlassian.net/browse/MM-28250
https://mattermost.atlassian.net/browse/MM-27437


components/post_view/post_reaction/index.js
```diff
@@ -3,17 +3,30 @@
...
+import {addReaction, removeReaction} from 'actions/post_actions.jsx';
 
 import PostReaction from './post_reaction';
 
+function mapStateToProps(state, ownProps) {
+    const getReactionsForPost = makeGetReactionsForPost();
+    const currentUserId = getCurrentUserId(state);
+
+    return {
+        reactions: Object.values(getReactionsForPost(state, ownProps.postId) || {}),
```
hmhealey, line 18: This is going to pass a new array into `PostReaction` every time `mapStateToProps` is called. We should pass in this object into `PostReaction` and have it call `Object.values` on it itself.

Alternatively, instead of making `PostReaction` be in charge of deciding between `addReaction` and `removeReaction`, we could make a new `toggleReaction` action that gets these values from the redux state so that we can avoid passing them into the component in the first place. That could also be used in both of these components.

components/post_view/reaction/reaction.jsx
```diff
@@ -116,13 +116,16 @@ export default class Reaction extends React.Component {
...
-        if (nextProps.reactionCount !== prevState.displayNumber && nextProps.sortedUsers.currentUserReacted === prevState.userReacted) {
+
+        if (nextProps.reactionCount !== prevState.displayNumber) {
             // set counts relative to current user having reacted
             if (prevState.userReacted) {
                 return {
                     displayNumber: nextProps.reactionCount,
                     reactedNumber: nextProps.reactionCount,
                     unreactedNumber: nextProps.reactionCount - 1,
+                    userReacted: nextProps.sortedUsers.currentUserReacted,
```
hmhealey, line 127: This doesn't need to be changed now, but I still don't like that these are stored in state instead of just being derived from props when rendered.

components/post_view/reaction_list/index.js
```diff
@@ -34,8 +37,9 @@ function mapDispatchToProps(dispatch) {
     return {
         actions: bindActionCreators({
             addReaction,
+            removeReaction,
         }, dispatch),
     };
 }
 
-export default connect(makeMapStateToProps, mapDispatchToProps)(ReactionList);
+export default connect(makeMapStateToProps(), mapDispatchToProps)(ReactionList);
```
hmhealey, line 45: `makeMapStateToProps` shouldn't be called by us or else you end up in the same situation as just having a single `mapStateToProps` function where selectors are overwriting each others' results.

hmhealey (changes requested):

jgilliam17:
Thank you @nevyangelova 
I see a small glitch when removing a reaction (not reproducing on master). See gif.
e.g. When removing an emoji with the current count of 2, I expected the count to roll back to 1, instead count appears to jump to 3 before it rolls down to 1.
emoji counter remove

components/post_view/post_reaction/index.js
nevyangelova: I also thought about doing this but for some reason decided to do a quick fix 🤷  I will do it now :)

components/post_view/reaction/reaction.jsx
nevyangelova: Same thoughts here. We can open a follow up ticket coz there will be some thinking.

nevyangelova:
@esethna @jgilliam17 this component is tricky because it used double buffer between state and props. This is particularly bad in this case because two places are updating the reaction count and they aren't synchronised between each other. So this is why we got the initial bug. By removing one of the conditions I managed to update the sync but then we get the 'jumping' bug that Jelena found. This is because the values are being updated in multiple places as well and not programmatically. I talked with @hmhealey and @sudheerDev and we tried to come up with a way to avoid both bugs and we came to the conclusion that the only way this will work is by a thorough refactor and moving everything to be driven by redux state. Which won't make friday's code complete.

Options here are 1. We ship the current fixes as I got them to work somewhat clear. And then refactor as soon as possible (I already have an idea how) 
2. We revert the emoji reaction animation and implement the component properly.

Let me know what you think.

esethna:
@nevyangelova see comment on this ticket which I think should just be closed as won't fix: https://mattermost.atlassian.net/browse/MM-27988

As for the glitchiness in the picker, I tested the spinwick and this change breaks the smoothness of the reaction animation:
https://i.gyazo.com/e56683716e2a67f071adbb4746dd7a6c.gif

I propose we queue the refactor ticket needed to fix this properly.

nevyangelova:
OK I will close this and update the ticket statuses :)

PR #6509, opened 2020-09-21
MM-28891 Comment out in product notices e2e test

sudheerDev:
#### Summary
Commenting out e2e tests till we have a repo to server notices for e2e tetsts or mock the response.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28891

### DIFF
--- a/e2e/cypress/integration/modals/product_notice_spec.js
+++ b/e2e/cypress/integration/modals/product_notice_spec.js
@@ -9,95 +9,95 @@
 
 // Group: @modals
 
-describe('MM-T3361 In product notices', () => {
-    let testTeam;
-    let testUser;
-
-    before(() => {
-        cy.apiInitSetup().then(({team, user}) => {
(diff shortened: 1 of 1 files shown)


hmhealey (approved):
Perhaps a longer term approach would be some mechanism on the server to override where those notices are coming from? I know it kind of goes against the purpose of the feature, but letting people intercept and potentially send their own upgrade notices might be nice for people using whitelabeled versions of MM.

nevyangelova (approved):

sudheerDev:
Skipping QA review as this just is commenting out a pieces of code

sudheerDev:
> Perhaps a longer term approach would be some mechanism on the server to override where those notices are coming from? I know it kind of goes against the purpose of the feature, but letting people intercept and potentially send their own upgrade notices might be nice for people using whitelabeled versions of MM.

That would be nice, cc @esethna

PR #6514, opened 2020-09-21
Show Create channels tip to admin

nickmisasi:
#### Summary
UX for when a first user went through tips was weird because you could "Browse Channels" but there wouldn't be any there. The first admin will now have the option to "Create a channel" instead - button takes you directly to the new channel flow.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28806
#### Related Pull Requests
n/a
#### Screenshots

image


stevemudie (approved):
LGTM

matthewbirtch (approved):
Looks good! Thanks @nickmisasi

components/next_steps_view/next_steps_tips.tsx
```diff
@@ -309,6 +314,63 @@ export default function NextStepsTips(props: Props) {
         );
     }
 
+    let channelsSection;
+    if (props.isFirstAdmin) {
+        channelsSection = (
```
ali-farooq0, line 319: Could just create two tiny components, would be much easier to read the whole thing as opposed to this giant if/else block :)

ali-farooq0 (approved):

ethervoid (approved):
LGTM same as Ali mentioned about splitting into two smaller components

components/next_steps_view/next_steps_tips.tsx
nickmisasi: I think this is a good idea, but want to do it to the entire component since all of these tips are just a card with title + description + action button. 

I've simplified the if statements to set an object that renders the one card instead of 2 possible JSX fragments. This is much more readable in the interim. 

I am going to open a ticket to actually refactor the whole page into using something similar to the [TextCardWithAction](https://github.com/mattermost/mattermost-webapp/tree/master/components/next_steps_view/steps/text_card_with_action) component I built for the next_steps_view

ali-farooq0: Sounds good! Thanks for making the change @nickmisasi! 🎉

ali-farooq0 (approved):
Double appppproved ✅

PR #6518, opened 2020-09-22
[MM-27708] Disable Plugin on Removal

hanzei:
#### Summary
This PR adds an E2E for https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T39.

I've opted for using the API as the test case doesn't test the UI and rather on the server functionality. There is already a test related to plugnn installs in `upgrade_spec.js` so I decided it's a good place to put this test too.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27708


e2e/cypress/integration/plugins/upgrade_spec.js
```diff
@@ -42,7 +42,7 @@ describe('Plugin remains enabled when upgraded', () => {
         cy.apiRemovePluginById(pluginIdDemo);
     });
 
-    it('MM-T40 Plugin remains enabled when upgraded', () => {
+    xit('MM-T40 Plugin remains enabled when upgraded', () => {
     // * Upload Demo plugin from the browser
         const fileName1 = 'com.mattermost.demo-plugin-0.1.0.tar.gz';
         const fileName2 = 'com.mattermost.demo-plugin-0.2.0.tar.gz';
```
DHaussermann, line 48: Seems like here we are not using `cy.apiInstallPluginFromUrl` because the point is deliberately to test the UI. 
This means the test will never be able to run green without human intervention. 

@saturninoabril not sure how to solve this problem. Can we commit the tar.gz files containing the binaries directly to the repo so they already exist in `fixtures`? This would work but seems not ideal.

DHaussermann (changes requested):
- 2 small changes needed to make the tests work as expected 
- Not sure if we need to make it so the test can run without manually moving the files int the `fixture` folder. Pinged @saturninoabril with a question on this.

e2e/cypress/integration/plugins/upgrade_spec.js
```diff
@@ -42,7 +42,7 @@ describe('Plugin remains enabled when upgraded', () => {
         cy.apiRemovePluginById(pluginIdDemo);
     });
 
-    it('MM-T40 Plugin remains enabled when upgraded', () => {
+    xit('MM-T40 Plugin remains enabled when upgraded', () => {
```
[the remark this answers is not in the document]
hanzei: Good catch :+1: That is an unrelated change.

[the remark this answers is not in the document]
hanzei: Good catch, I disabled that anciently.

e2e/cypress/integration/plugins/upgrade_spec.js
hanzei: My main motivation behing using  `cy.apiInstallPluginFromUrl` is that we don't have to commit the binaries then.

saturninoabril: We're not actually including plugin binaries to the repo. On top of this file, there's an instruction on how to download and save the binaries for local testing/dev. In CI, there's a separate step/job that handles automatic downloads of required binaries.

e2e/cypress/integration/plugins/upgrade_spec.js
```diff
@@ -99,6 +99,52 @@ describe('Plugin remains enabled when upgraded', () => {
...
+            true,
+        ).then((response) => {
+            manifest = response.body;
+        });
+
+        // * Confirm demo plugin is not enabled
+        cy.apiGetAllPlugins().then((response) => {
+            const {active, inactive} = response.plugins;
+            cy.log('inactive', inactive);
+            cy.log('manifest', manifest);
```
saturninoabril, line 129: Maybe these were used during debugging?

e2e/cypress/integration/plugins/upgrade_spec.js
```diff
@@ -99,6 +99,52 @@ describe('Plugin remains enabled when upgraded', () => {
         // * Verify v0.2.0 of plugin
         cy.findByText(/0.2.0/).scrollIntoView().should('be.visible');
     });
+
+    it('MM-T39 Disable Plugin on Removal', () => {
+        // # Install demo plugin and enable it
+        cy.apiAdminLogin();
+        cy.apiInstallPluginFromUrl(
+            'https://github.com/mattermost/mattermost-plugin-demo/releases/download/v0.1.0/com.mattermost.demo-plugin-0.1.0.tar.gz',
```
saturninoabril, line 107: Maybe put this in a constant since it's used in two places?

saturninoabril (approved):
Thanks @hanzei! Feels like this should have done via UI, but I'm good as is. Left non-blocking comments.

e2e/cypress/integration/plugins/upgrade_spec.js
hanzei: Good suggestion :+1:

e2e/cypress/integration/plugins/upgrade_spec.js
hanzei: Good catch :+1:

hanzei:
@DHaussermann Please take another look

e2e/cypress/integration/plugins/upgrade_spec.js
```diff
@@ -25,13 +25,14 @@ import * as TIMEOUTS from '../../fixtures/timeouts';
 
 describe('Plugin remains enabled when upgraded', () => {
     const pluginIdDemo = 'com.mattermost.demo-plugin';
+    const demoPluginURL = 'https://github.com/mattermost/mattermost-plugin-demo/releases/download/v0.1.0/com.mattermost.demo-plugin-0.1.0.tar.g';
```
DHaussermann, line 28: @hanzei the 2nd test was failing to download the plugin even after your last merge. 

Seems there is an error in the URL the `z` is missing here in `com.mattermost.demo-plugin-0.1.0.tar.g`
Adding the `z` to the end fixed it for me.

DHaussermann (changes requested):
@hanzei - the last merged fixed the 1st tests but the 2nd test is failing for me as it can't download the plugin

e2e/cypress/integration/plugins/upgrade_spec.js
saturninoabril: Ah, I missed that. Nice catch, Dylan!

hanzei: Good catch @DHaussermann :+1:

DHaussermann (approved):
Tested and passed
- Test steps are valid
- Both tests now run green
LGTM!

Set corresponding YM4J test Cypress filed to `update`

PR #6525, opened 2020-09-22
Possible fix for the failing long draft test in messaging

prapti:
Took a stab at this test since I was seeing it last week as well. Wasn't sure about what was causing the failure since it passes as expected in every run on my local server, thus increased the wait from half a second to 3 seconds for assertions after the draft is typed. Hopefully this fixes the flakiness.
Also demoted the test from prod. Will keep an eye out to see if this fixes the issue before promoting it back to prod.


josephbaylon (approved):
LGTM. @prapti I added cherry pick label and v5.28 milestone as this looks like an issue on 5.28 as well.

srkgupta (approved):
Tested and passed.

PR #6464, opened 2020-09-16
[MM-28726] Add X to top right of Next Steps Tips for easy back-out

nickmisasi:
#### Summary
When in the Next Steps and Tips view, there was no easily noticeable way to back-out of it. This PR adds an X to the top right corner. When that X is clicked, it's the same behaviour as if the user clicked the X on the LHS - a modal appears stating that it will be added to the main menu.

Actually having it added to the main menu will come in a future PR, as part of ticket MM-28727 (which will build on this one)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28726

#### Related Pull Requests
n/a
#### Screenshots
image

PR #6527, opened 2020-09-22
[MM-28878] Revert style change to remove next steps arrow

devinbinnie:
#### Summary
The `left` offset of the arrow used to indicate the hamburger menu on the Remove Next Steps modal was changed to work with the older sidebar. This PR resets it to work with the new sidebar.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28878

### DIFF
--- a/components/sidebar/sidebar_next_steps/remove_next_steps_modal.scss
+++ b/components/sidebar/sidebar_next_steps/remove_next_steps_modal.scss
@@ -1,7 +1,7 @@
 .RemoveNextStepsModal__helpBox {
     position: fixed;
     top: 16px;
-    left: 181px;
+    left: 201px;
     z-index: 1050;
 }
 


marianunez:
@nickmisasi would this break the modal interaction to exit out of the `Tips and Next Steps` from  https://github.com/mattermost/mattermost-webapp/pull/6464 where the change was introduced?

devinbinnie:
> @nickmisasi would this break the modal interaction to exit out of the `Tips and Next Steps` from #6464 where the change was introduced?

Nope, it's just a style change. Just moves the arrow over.

nickmisasi:
@marianunez @devinbinnie I think I made this change because I was fixing the "bug" that it wouldn't line up when the new sidebar wasn't enabled. My bad.

Reverting this change will bring it back to the way it was on initial release (ie, is in the right position if you have the new sidebar enabled), but I suspect that if you disable the new sidebar, the circle won't line up in the other direction (it'll be too far to the right). If we don't care about the old sidebar, then this is fine. But if we do plan to support the old sidebar this won't work for both.

devinbinnie:
> @marianunez @devinbinnie I think I made this change because I was fixing the "bug" that it wouldn't line up, but when I fixed it I didn't have the new sidebar enabled. My bad.
> 
> Reverting this change will bring it back to the way it was on initial release (ie, is in the right position if you have the new sidebar enabled), but I suspect that if you disable the new sidebar, the circle won't line up in the other direction (it'll be too far to the right). If we don't care about the old sidebar, then this is fine. But if we do plan to support the old sidebar this won't work for both.

It's not your fault, don't worry about it. I'm personally of the belief that we shouldn't be spending time coding for the old sidebar, it will be disappearing eventually.

marianunez (approved):

nickmisasi (approved):

stevemudie (approved):
LGTM

amyblais:
@devinbinnie Maria mentioned this should be added in the next beta release, let me know if this needs cherry-picking.

devinbinnie:
> @devinbinnie Maria mentioned this should be added in the next beta release, let me know if this needs cherry-picking.

I've already manually cherry-picked into the `cloud-beta` branch, so unless it needs to go into a release branch, I think we're good.

marianunez:
> I've already manually cherry-picked into the `cloud-beta` branch, so unless it needs to go into a release branch, I think we're good.

@devinbinnie I am not able to see the commit for this change in the `cloud-beta` branch. Maybe still needs to be pushed?

devinbinnie:
> > I've already manually cherry-picked into the `cloud-beta` branch, so unless it needs to go into a release branch, I think we're good.
> 
> @devinbinnie I am not able to see the commit for this change in the `cloud-beta` branch. Maybe still needs to be pushed?

Oh whoops, this is a different PR. So many PRs to cherry-pick :P will do this one as well.

PR #6543, opened 2020-09-23
[MM-28199] Dropdown Input Common Component

devinbinnie:
#### Summary
Dropdown Input using `react-select`, needed for Cloud Billing.
Demo in System Console > Biling > Subscription

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28199


stevemudie (approved):
LGTM

matthewbirtch (commented):
Hey @devinbinnie, this is looking good. 

I believe we called this out before, but the focused state stroke is supposed to be on the inside of the input box, not the outside. I thought we resolved this with the text input components, but I'm seeing it again here (and on the inputs in the Getting Started screen.

It might be hard to see, but it does cause the border radius to look larger and does cause the actual size of the input box to be larger in the focus state.

Frame 6

ethervoid:
Given my little expertise with React, I think @ali-farooq0 is better suited to give you a better review here

matthewbirtch:
As mentioned in our chat @devinbinnie, I'm now seeing a shift in the other direction when I focus
image

matthewbirtch (approved):
Thanks @devinbinnie! Looks great

components/dropdown_input.scss
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.DropdownInput__option.selected > div {
+    background-color: rgba(var(--button-bg-rgb), 0.08);
+    color: var(--center-channel-color);
+}
+
+.DropdownInput__option.focused > div {
+    background-color: rgba(var(--center-channel-color-rgb), 0.08);
+}
```
ali-farooq0, line 49: nit: empty line

components/dropdown_input.tsx
```diff
@@ -0,0 +1,143 @@
...
+        const {onBlur} = props;
+
+        setFocused(false);
+
+        if (onBlur) {
+            onBlur(event);
+        }
+    };
+
+    let inputClass = className ? `Input ${className}` : 'Input';
```
ali-farooq0, line 102: Would it be better to use pkg Classnames here? might help inline these as opposed to using the ternary operator.

ali-farooq0 (approved):
Nice work @devinbinnie!

components/dropdown_input.tsx
devinbinnie: Yeah, this was pulled from CWS where I don't think we're using this. Should be a quick refactor.

components/admin_console/billing/billing_subscriptions.tsx
```diff
@@ -142,8 +145,18 @@ const BillingSubscriptions: React.FC<Props> = () => {
                         className='BillingSubscriptions__topWrapper'
                         style={{marginTop: '20px'}}
                     >
-                        <div style={{border: '1px solid #000', width: '568px'}}>
+                        <div style={{border: '1px solid #000', width: '568px', padding: '8px', backgroundColor: '#fff'}}>
```
deanwhillier, line 148: Are you planning on moving these styles out to a css module or global styles at some point?

devinbinnie: These are temporary just so I can get an idea of the layout.
All of these in-line styles will be going away and replaced with actual components or `div`s with classes attached.

deanwhillier (approved):
LGTM, again with the understanding that these components will be iterated on as part of the upcoming common components initiative.

@devinbinnie, I chatted briefly with @matthewbirtch and we were thinking that we should have the mouse cursor change to show the interaction hand when mousing over the input field and dropdown options.

marianunez:
@devinbinnie just a reminder to cherry-pick this to `cloud-ga`

marianunez:
@devinbinnie Ignore my comment above, this is already in `cloud-ga` 🤦‍♀️

PR #6544, opened 2020-09-23
MM-28992 Remove unused notPresent and mustBePresent props from MultiSelectSetting

hmhealey:
It looks like this prop used to be needed at some point, but I can't find anywhere that it's used in the current code base. The MultiSelectSetting also doesn't appear to be exposed to plugins, so we don't need to be worried about them either.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28992


enahum (approved):

jespino (approved):
Lgtm

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified available languages multi-selector setting in system console - works as expected.

PR #6546, opened 2020-09-24
[MM-28965] - Prevent link overlap in channel header

nevyangelova:
#### Summary
Quick fix for links that overlap in channel header

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28965

Fix v5.29


components/markdown_image.jsx
```diff
@@ -113,9 +113,7 @@ export default class MarkdownImage extends React.PureComponent {
                                 rel='noopener noreferrer'
                                 target='_blank'
                                 title={this.props.title}
-                            >
-                                {alt}
-                            </a>
+                            />
```
hmhealey, line 116: I don't know if this needs to be changed. If someone links to an unsafe image, it won't show anything now whereas it made a link to that image before.

sass/layout/_headers.scss
```diff
@@ -269,6 +269,7 @@
             .emoticon {
                 position: relative;
                 top: -1px;
+                overflow: hidden;
```
hmhealey, line 272: I didn't know the header had its own styling for emojis. I'm not at all surprised this broke then

hmhealey (changes requested):

components/markdown_image.jsx
nevyangelova: Oh I see, thanks for clarifying

hmhealey (approved):

calebroseland (approved):

amyblais:
@nevyangelova From your investigation do you know if this is a regression in v5.27?

nevyangelova:
@amyblais it isn't these styles were implemented a year ago.

ThiefMaster:
I'm relatively sure that this problem only started showing up during one of the last two releases. I had the header that's mentioned in the JIRA ticket for a long time and clicking the link there worked fine.

nevyangelova:
@ThiefMaster it's possible that some styles were removed and revealed this bug but this is the proper fix and it wasn't implemented correctly a year ago.

ThiefMaster:
I just wanted to point out that it'd probably be a good candidate for a bugfix release if there's going to be one since it's a small but annoying bug.

nevyangelova:
@ThiefMaster makes sense, @amyblais should this go in 5.28?

amyblais:
I think v5.29 is ok.

jgilliam17 (approved):
Thank you @nevyangelova 
Tested, looks good to merge.
- Verified header description links are clickable and they are not overlapping.
Note: The line that appears on mouse hover over linked emoji can't be removed currently and will have to be addressed separately if we decide to change/remove how markdown is processed.

PR #6552, opened 2020-09-24
Remove usage of t function with useIntl

hmhealey:
The `t` function is only necessary when the i18n tools aren't able to identify a translation key themselves, but it can recognize keys in calls to `formatMessage`, so using the `t` function in all of these is unnecessary.

Skipping QA review because there's no functional changes from this.


nevyangelova (approved):
LGTM found no errors on the branch, still this might need some extra QA ?

jespino (approved):
LGTM. Thanks!

PR #6565, opened 2020-09-26
Add file search results item widget

jespino:
#### Summary
Add file search results item widget

#### Screenshots


jespino:
This is going to be part of other PR

PR #6567, opened 2020-09-27
[MM-28680] rm usage of jQuery from theme settings

lipmem:
#### Summary
Removed the usage of jquery from:
- components/user_settings/display/user_settings_theme/custom_theme_chooser.jsx
- components/user_settings/display/user_settings_theme/premade_theme_chooser/premade_theme_chooser.jsx
- utils/utils.jsx

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15161


components/user_settings/display/user_settings_theme/custom_theme_chooser.jsx
```diff
@@ -214,30 +216,34 @@ export default class CustomThemeChooser extends React.PureComponent {
...
     }
 
     toggleSection(node) {
-        if (UserAgent.isIos()) {
-            // iOS doesn't support jQuery animations
-            $(node).toggleClass('open'); // eslint-disable-line jquery/no-class
+        node.classList.toggle('open');
+
+        // set overflow after animation, so the colorchooser is fully shown
+        if (node.classList.contains('open')) {
```
hmhealey, line 241: Could the overflow-y always be hidden? I'm not sure if setting it to visible after the animation changes anything since an opened section should be big enough that nothing overflows.

sass/routes/_settings.scss
```diff
@@ -356,13 +356,18 @@
...
                         @include border-radius(0 0 3px 3px);
                         @include legacy-pie-clearfix;
-                        @include alpha-property(background-color, $white, .05);
-                        display: none;
+                        @include alpha-property(background-color, $white, 0.05);
+
                         margin: 0 20px 0 0;
-                        padding: 24px 0 0 24px;
+                        padding: 0 0 0 24px;
+                        transition: all 0.5s ease-in-out;
```
hmhealey, line 363: Could you shorten this slightly to 0.4s? That's the default duration used by jquery's `slideToggle`

sass/routes/_settings.scss
```diff
@@ -356,13 +356,18 @@
...
+                        max-height: 0;
+                        overflow-y: hidden;
 
                         &.open {
-                            display: block;
+                            max-height: 1000px;
```
calebroseland, line 368: This appears to cause an issue with the theme selector in mobile view (`@media`/`max-width: 768px`)

components/user_settings/display/user_settings_theme/custom_theme_chooser.jsx
```diff
@@ -214,30 +216,34 @@ export default class CustomThemeChooser extends React.PureComponent {
...
+            setTimeout(() => {
+                node.style.overflowY = 'inherit';
```
calebroseland, line 243: If you close a section while it's still animating, the contents will overflow/be visible. (Mobile view and tablet/desktop viewports)

hmhealey (changes requested):
Thanks for the PR, @lipmem! I have a few suggestions to make, but otherwise, this looks good to me

sass/routes/_settings.scss
hmhealey: I was looking at that as well since the arbitrary max-height seemed odd to me. It seems to be needed for the animation to behave smoothly for some reason. I tried setting it to other values like `100%`, but the animation is only smooth when it's set to some arbitrary pixel value. As seen in your gif, we'll have to increase it a bit since that section becomes longer than 1000px when the screen is small enough.

calebroseland: While animating max-height can work, IMHO I think it would be better to use `react-transition-group` hooks (that might look something [like this](https://github.com/calebroseland/calebroseland-com/blob/master/src/components/transitions/components/CssTransition.vue#L40-L86))

Otherwise, Account Settings is slated to get a visual overhaul, so we could maybe switch over to a flexbox layout during that? (https://css-tricks.com/using-css-transitions-auto-dimensions/#bonus-technique-flexbox)

@hmhealey do you have any thoughts on this?

calebroseland (changes requested):
Left a couple of comments on the slideToggle replacement, but otherwise looks good!

components/user_settings/display/user_settings_theme/custom_theme_chooser.jsx
lipmem: I tried it without.
But if you open the color picker for lowest select, the color picker is cut off due its absolute placement.

components/user_settings/display/user_settings_theme/custom_theme_chooser.jsx
calebroseland: Maybe an [ontransitionend](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ontransitionend) handler would be more appropriate?

lipmem: I changed it to the ontransitionend handler :)

sass/routes/_settings.scss
lipmem: `react-transition-group` hooks would still need to use css transitions. As I understand it?
So I think I do not come around animating max-height. :/

sass/routes/_settings.scss
lipmem: Done :)

lipmem:
@hmhealey @calebroseland 
Hey, thank you for your feedback. I committed a bit to implement your suggestions

components/user_settings/display/user_settings_theme/custom_theme_chooser.jsx
hmhealey: Ah, I missed that when fiddling with it locally. Good catch!

hmhealey (approved):
Looks good to me! We might be able to do something to remove the refs entirely from the component, but we're planning on a larger refactor of the settings modal, so we can fix any quirks with the animation then.

calebroseland (approved):

jgilliam17 (approved):
Thank you @lipmem 
Tested, looks good to merge.
- Verified changing custom theme settings and theme application throughout the app.

@calebroseland Can you please help merge? Thanks.

PR #6574, opened 2020-09-28
[MM-20599] Migrated select_team component to Typescript

diode:
#### Summary
Select Team Component - TypeScript Migration

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/13475


components/select_team/select_team.tsx
```diff
@@ -190,7 +201,7 @@ export default class SelectTeam extends React.PureComponent {
                 </div>
             );
         } else {
-            let joinableTeamContents = [];
+            let joinableTeamContents: any = [];
```
calebroseland, line 204: I think a `ReactNode` would be more appropriate here, but I guess `any` might be okay for now as the array/forEach logic would have to be replaced?

calebroseland (changes requested):
Thanks for the PR, @diode! Overall looking good, just a few comments/suggestions.

components/select_team/select_team.tsx
```diff
@@ -137,21 +148,21 @@ export default class SelectTeam extends React.PureComponent {
         }
     };
 
-    handleLogoutClick = (e) => {
+    handleLogoutClick = (e: any): void => {
```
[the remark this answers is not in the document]
calebroseland: `{MouseEvent} from 'react'`

components/select_team/components/select_team_item.tsx
```diff
@@ -79,8 +86,8 @@ export default class SelectTeamItem extends React.PureComponent {
                 {this.renderDescriptionTooltip()}
                 <a
                     href='#'
-                    id={Utils.createSafeId(team.display_name)}
-                    onClick={canJoin ? this.handleTeamClick : null}
+                    id={this.getId(team.display_name)}
```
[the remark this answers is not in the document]
diode: Thanks. This change will be done.

components/select_team/index.ts
```diff
@@ -16,9 +16,9 @@ import {getCurrentUser} from 'mattermost-redux/selectors/entities/users';
 import {addUserToTeam} from 'actions/team_actions';
 import {isGuest} from 'utils/utils';
 
-import SelectTeam from './select_team.jsx';
+import SelectTeam from './select_team';
 
-function mapStateToProps(state) {
+function mapStateToProps(state: any) {
```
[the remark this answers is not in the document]
diode: Thanks. This change will be done.

components/select_team/components/select_team_item.test.tsx
```diff
@@ -22,44 +22,58 @@ describe('components/select_team/components/SelectTeamItem', () => {
 
     test('should match snapshot, on public not joinable', () => {
         const props = {...baseProps, canJoinPublicTeams: false};
-        const wrapper = shallow(<SelectTeamItem {...props}/>);
+        const wrapper: ShallowWrapper<any, any, SelectTeamItem> = shallow(
```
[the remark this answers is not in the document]
diode: Thanks. This change will be done.

components/select_team/select_team.tsx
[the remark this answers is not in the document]
diode: Initially tried this. But from test, this method is directly called with a mock event. So had to use **any**
```javascript
 wrapper.instance().handleLogoutClick({preventDefault: jest.fn()});
```

components/select_team/select_team.tsx
```diff
@@ -137,21 +148,21 @@ export default class SelectTeam extends React.PureComponent {
...
         e.preventDefault();
         trackEvent('select_team', 'click_logout');
         emitUserLoggedOutEvent('/login');
     };
 
-    clearError = (e) => {
+    clearError = (e: any): void => {
```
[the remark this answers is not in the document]
diode: Initially tried this. But from test, this method is directly called with a mock event. So had to use **any**
```javascript
 wrapper.instance().clearError({preventDefault: jest.fn()});
```

components/select_team/select_team.tsx
diode: `joinableTeamContents` is used both as an array

```javascript
joinableTeamContents.push(
   key={'team_' + listableTeam.name}
   team={listableTeam}
   ...                        
);                    
```

and to store JSX

```javascript
  joinableTeamContents = (
     <div className='signup-team-dir-err'>
         ...
     </div>
  );                           
```

components/select_team/select_team.tsx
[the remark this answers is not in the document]
calebroseland: I think I've seen it coerced to any in some places, would something like that work?
```ts
wrapper.instance().clearError({preventDefault: jest.fn()} as any);
```

components/select_team/select_team.tsx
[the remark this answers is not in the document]
diode: Made this change, Thanks

components/select_team/select_team.tsx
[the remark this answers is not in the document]
diode: Made this change, Thanks

components/select_team/components/select_team_item.tsx
```diff
@@ -12,21 +11,21 @@ import TeamInfoIcon from 'components/widgets/icons/team_info_icon';
...
-    static propTypes = {
-        team: PropTypes.object.isRequired,
-        onTeamClick: PropTypes.func.isRequired,
-        loading: PropTypes.bool.isRequired,
-        canJoinPublicTeams: PropTypes.bool.isRequired,
-        canJoinPrivateTeams: PropTypes.bool.isRequired,
-    };
+type Props = {
+    team: any,
+    onTeamClick: (team: any) => void,
```
hmhealey, line 16: These can both use the `Team` type from `mattermost-redux/types/teams`. That can also be used in a few places in `SelectTeam` as well

components/select_team/components/select_team_item.tsx
```diff
@@ -79,8 +78,8 @@ export default class SelectTeamItem extends React.PureComponent {
                 {this.renderDescriptionTooltip()}
                 <a
                     href='#'
-                    id={Utils.createSafeId(team.display_name)}
-                    onClick={canJoin ? this.handleTeamClick : null}
+                    id={Utils.createSafeId(team.display_name) ?? undefined}
```
hmhealey, line 81: I haven't seen this pattern before where we coalesce a value to undefined. Do we need this because `createSafeId` can return null, and React complains if we pass a null id? I haven't seen this anywhere before, but I assume it's because we haven't used this in a TS file before.

Maybe we could change the return type of `createSafeId` to return undefined instead of null in the error case to save people from having to do this elsewhere?

components/select_team/index.ts
```diff
@@ -12,13 +12,14 @@ import {Permissions} from 'mattermost-redux/constants';
 import {haveISystemPermission} from 'mattermost-redux/selectors/entities/roles';
 import {getSortedListableTeams, getTeamMemberships} from 'mattermost-redux/selectors/entities/teams';
 import {getCurrentUser} from 'mattermost-redux/selectors/entities/users';
+import {GlobalState} from 'mattermost-redux/types/store';
```
hmhealey, line 15: This should actually be imported from `types/store` in the web app. That type extends the `mattermost-redux` one to add some app-specific state.

components/select_team/select_team.tsx
```diff
@@ -29,35 +28,45 @@ import FormattedMarkdownMessage from 'components/formatted_markdown_message';
...
+type Actions = {
+    getTeams: any,
+    loadRolesIfNeeded: any,
+    addUserToTeam: any,
+}
 
-    constructor(props) {
+type Props = {
+    currentUserId: string,
+    currentUserRoles?: string,
```
hmhealey, line 44: Some of these like `currentUserRoles` can actually be made required since they're always passed in by `mapStateToProps`. Maybe you could make those required to avoid some ugly null checks below?

components/select_team/select_team.tsx
```diff
@@ -29,35 +28,45 @@ import FormattedMarkdownMessage from 'components/formatted_markdown_message';
...
-        siteURL: PropTypes.string,
-        actions: PropTypes.shape({
-            getTeams: PropTypes.func.isRequired,
-            loadRolesIfNeeded: PropTypes.func.isRequired,
-            addUserToTeam: PropTypes.func.isRequired,
-        }).isRequired,
-        totalTeamsCount: PropTypes.number.isRequired,
-    };
+type Actions = {
+    getTeams: any,
```
hmhealey, line 37: Could you add some proper types for these actions?

hmhealey (changes requested):
Thanks for the PR, @diode! I have a few more suggestions for where we can be clearer about some types.

components/select_team/select_team.tsx
hmhealey: I think `ReactNode` can also take an array as well as JSX, but I guess that doesn't work when we call `.length` on it below

components/select_team/components/select_team_item.tsx
diode: Thanks. Made this change.

components/select_team/components/select_team_item.tsx
diode: Thanks. Made this change.

components/select_team/index.ts
diode: Thanks. Made this change.

components/select_team/select_team.tsx
diode: Thanks. Made this change.

hmhealey:
Thanks for making those changes. Those look good to me, although there are some unintended changes included on the branch now. Would you be able to remove those yourself? If not, I can do some git magic to get rid of them.

diode:
@hmhealey  Thanks. I will remove those commits.

diode:
@hmhealey Please check now.

hmhealey (approved):
Thanks for sorting that out!

calebroseland (approved):

jgilliam17 (approved):
Thank you @diode 
Tested, looks good to merge.
- Verified select team modal - as before, no change in functionality. 

@hanzei Can you please help merge? Thanks. 🙂

hanzei:
Thanks for the contribution @diode :+1:

PR #6577, opened 2020-09-29
[MM-28944] Don't stop searching until no more results returned

mgdelacroix:
#### Summary
We've had problems when a post gets returned by a search engine but it's marked as deleted or absent in the database, the search endpoint returns 19 instead of 20 (for example) returns, and the frontend marks the search as ended, having more results in the following pages.

This PR and it's redux counterpart only mark the search as ended when no more results are returned by the search endpoint, to fix this bug.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28944

#### Related Pull Requests
- [Has redux changes](https://github.com/mattermost/mattermost-redux/pull/1248)


package.json
```diff
@@ -33,7 +33,7 @@
     "localforage-observable": "2.0.1",
     "mark.js": "8.11.1",
     "marked": "github:mattermost/marked#87769262aa02e1784570f61f4f962050e07cc335",
-    "mattermost-redux": "github:mattermost/mattermost-redux#7a241e2a4f56217320ae79facca5a3a3fb83e95d",
+    "mattermost-redux": "github:mattermost/mattermost-redux#af5a84470d7a103162725b7e3c5e28caa38774f0",
```
mgdelacroix, line 36: This link needs to be updated before merging the PR

components/search_results/search_results.jsx
```diff
@@ -179,6 +180,13 @@ class SearchResults extends React.Component {
         const noResults = (!results || results.length === 0);
         const searchTerms = this.props.searchTerms;
 
+        // to avoid loading icon showing infinitely, if the first page
+        // has results but no scroll, trigger the second page search
+        // to mark search as ended
+        if (this.props.searchPage === 0 && !noResults) {
```
jespino, line 186: What happens if there are 0 results in the page 0?

mgdelacroix: In that case the `isEnd` is set to `true` directly and the loading icon doesn't show at all

isacikgoz (approved):
Looks good to me 👍

components/search_results/search_results.jsx
jespino: Ok, this is super hard to track, and is using very distant code to archive this. If we change that behavior in the mattermost-redux repo, expecting to have it solved in the origin, we can break it without noticing it.

My suggestion is to add the `!this.props.isSearchAtEnd` check here too.

jespino (approved):
LGTM, some concerns, not extrictly related with the fix done here, more about the architecture of this.

components/search_results/search_results.jsx
mgdelacroix: Added!

jespino: Thanks! :)

lindy65 (approved):
Thanks @mgdelacroix - tested on /cloud server `elasticsearch` pointing to this PR's hash and did not find any issues.  👍

PR #6581, opened 2020-09-29
Cypress/E2E: Add cy.uiClickPostDropdownMenu command

saturninoabril:
#### Summary
Add `cy.uiClickPostDropdownMenu` command.

Test report: https://mm-cypress-report.s3.amazonaws.com/1024-cypress-post-dropdown-menu-master/mochawesome.html (failing test are not related to change)

### DIFF
--- a/e2e/cypress/integration/mark_as_unread/channel_unread_spec.js
+++ b/e2e/cypress/integration/mark_as_unread/channel_unread_spec.js
@@ -11,7 +11,7 @@
 
 import {beRead, beUnread} from '../../support/assertions';
 
-import {markAsUnreadByPostIdFromMenu, verifyPostNextToNewMessageSeparator, switchToChannel} from './helpers';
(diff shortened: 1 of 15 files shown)


josephbaylon (approved):
LGTM. Changes look straightforward. Wondering, though, if we should pull out constants for the menu texts like 'Mark as Unread' etc so that when those change there's only one place to change, or pull out helper functions specific to that menu item, e.g. cy.uiClickPostMenuMarkAsUnread.

saturninoabril:
I'm okay with the use of constants, feel free to submit PR once this is merged. I'm a bit off to very targeted commands like `cy.uiClickPostMenuMarkAsUnread`. It will end up with huge overhead, especially for those new to our codebase. Right now, there are too many Cypress commands and I'd prefer to limit it as much as possible.

srkgupta (approved):
Thanks for the report @saturninoabril. The changes LGTM

PR #6583, opened 2020-09-29
[MM-27142] In-Web-App purchase of Mattermost Cloud 

nickmisasi:
#### Summary
We want Mattermost Cloud users to be able to purchase cloud licenses without leaving the webapp. This PR adds a modal that can take a credit card, and send payments to a Customer Web Server instance.

In order to see the purchase modal, you can access it through the sidebar main menu, via the button "Purchase Cloud"
When at or over the user limit for the spinwick, as an administrator you will see a banner (either blue or red) with a button to Upgrade Mattermost Cloud. Clicking this will take you to the modal. 
When at or over the user limit, as an administrator, clicking "Invite People" from the main menu will surface a modal, with an "Upgrade Mattermost Cloud" button. Clicking this will take you to the purchase modal.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27142

#### Related Pull Requests
- Has redux changes https://github.com/mattermost/mattermost-redux/pull/1251

#### Screenshots
image
image
image


actions/cloud.tsx
```diff
@@ -0,0 +1,98 @@
...
+                            postal_code: billingDetails.postalCode,
+                        },
+                    },
+                },
+            },
+        );
+
+        if (!result) {
+            console.error('Stripe confirm card setup failed.'); //eslint-disable-line no-console
+            return 'Stripe confirm card failed';//'intl.formatMessage({id: 'errors.generic_payment_failure'})'};
```
marianunez, line 69: We still need to add localization for these error messages in this function

components/cloud/types/customer.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const SET_CUSTOMER = 'SET_CUSTOMER';
+
+export type SetCustomerAction = {
+    type: typeof SET_CUSTOMER;
+    customer: Customer;
+};
+
+//export type CustomerActionType = SetCustomerAction | LogoutAction;
```
marianunez, line 39: I believe this overlaps with @devinbinnie [`Customer` type definition](https://github.com/devinbinnie/mattermost-redux/blob/727f469505d95b70060c93374adbfe878ab8e951/src/types/cloud.ts#L24) on this Redux PR https://github.com/mattermost/mattermost-redux/pull/1249 . We should use that instead and avoid duplicates in the webapp

components/main_menu/main_menu.jsx
```diff
@@ -341,6 +343,13 @@ class MainMenu extends React.PureComponent {
                             dialogType={MarketplaceModal}
                             text={formatMessage({id: 'navbar_dropdown.marketplace', defaultMessage: 'Plugin Marketplace'})}
                         />
+                        <Menu.ItemToggleModalRedux
+                            id='cloudPurchaseModal'
+                            modalId={ModalIdentifiers.CLOUD_PURCHASE}
+                            show={!this.props.mobile}
+                            dialogType={PurchaseModal}
+                            text={'Purchase Cloud'}
+                        />
```
marianunez, line 352: This should be removed before merge. This was just a easy access for dev/testing

components/payment_form/card_input.css
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.StripeElement:focus::placeholder {
+    color: 'transparent';
+}
+
+.StripeElement:focus {
+    border-color: transparent;
+    box-shadow: 0 0 0 2px var(--button-bg);
+}
```
marianunez, line 37: Nit: missing new line

components/payment_form/payment_form.scss
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    .PaymentForm-change {
+        font-family: Source Sans Pro;
+        font-weight: 600;
+        font-size: 16px;
+        line-height: 20px;
+        color: #0058CC;
+    }
+}
```
marianunez: Nit: missing new line

components/payment_form/payment_form.scss
```diff
@@ -0,0 +1,93 @@
...
+        opacity: 0.4;
+        margin-bottom: 16px;
+    }
+
+    .PaymentForm-saved-card {
+        font-family: Source Sans Pro;
+        font-size: 16px;
+        line-height: 24px;
+        color: #22406D;
+        border-bottom: 1px solid rgba(0, 0, 0, 0.12);
```
marianunez, line 76: We need to change have this use the theme colors but can make it part of the saved card ticket: https://mattermost.atlassian.net/browse/MM-29096

components/purchase_modal/icon_message.tsx
```diff
@@ -0,0 +1,96 @@
...
+                <FormattedMessage
+                    id={title}
+                />
+            </h3>
+            <div className={classNames('IconMessage-sub', error || '')}>
+                {subtitle ? <FormattedMessage id={subtitle}/> : null}
+            </div>
+            {button}
+            {link}
+            {footer}
```
marianunez, line 88: Do we need to add the error message here? I think it originally had it and I removed it at some point

components/purchase_modal/process_payment_setup.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    icon={processSvg}
+                    footer={progressBar}
+                />
+            );
+        case ProcessState.SUCCESS:
+            return (
+                <IconMessage
+                    title={t('admin.billing.subscription.upgradedSuccess')}
+                    subtitle={
+                        'Starting August 8, 2020 you will be charged based on the number of enabled users'
```
marianunez, line 144: We need to change this date to use the first date of the next month as we are doing in the modal here https://github.com/mattermost/mattermost-webapp/blob/46247a6f71a288694af34dfa9348d7c2f506536f/components/purchase_modal/purchase_modal.tsx#L83

components/purchase_modal/purchase.scss
```diff
@@ -0,0 +1,165 @@
...
+        .blue-dots {
+            position: absolute;
+            left: 0;
+            right: 0;
+            bottom: 0;
+            top: 0;
+            z-index: -1;
+
+            // this opacity makes it invisible
+            // opacity: 0.16;
```
marianunez, line 108: Should we remove this?

components/purchase_modal/purchase.scss
```diff
@@ -0,0 +1,165 @@
...
+            opacity: 0.6;
+            font-size: 16px;
+            line-height: 24px;
+            margin-bottom: 42px;
+        }
+
+        .SignUpRHS-link {
+            color: #22406D;
+            margin-left: 32px;
+        }
```
marianunez, line 156: I believe we can remove this, this was part of the migration from CWS but not used here

images/cloud/upgrade.svg
```diff
@@ -0,0 +1,30 @@
+<svg width="267" height="163" viewBox="0 0 267 163" fill="none" xmlns="http://www.w3.org/2000/svg">
```
marianunez, line 1: Is this the same one currently used in the user limit modal? It seems we are adding it again

images/onboarding-bg.svg
```diff
@@ -0,0 +1,143 @@
+<svg width="1366" height="767" viewBox="0 0 1366 767" fill="none" xmlns="http://www.w3.org/2000/svg">
```
marianunez, line 1: Are these onboarding images intended to be part of this PR? Maybe a bad merge?

marianunez (changes requested):
@nickmisasi awesome work getting this PR complete! Overall looks great! Some comments below

actions/cloud.tsx
nickmisasi: I thought we were saving these for the console log, and showing a generic error to the user? I don't think it's super helpful to tell the end user that we're using stripe. Should I make them more generic? Or just show the same generic error?

components/cloud/types/customer.ts
devinbinnie: For future reference @nickmisasi, we should try to keep all our types in `mattermost-redux`, especially if they represent data coming from the server.

devinbinnie:
Looks like we have some duplicate work done with this PR here: https://github.com/mattermost/mattermost-webapp/pull/6543

Not sure which we should keep, but I was hoping to introduce these dropdowns as common components.

components/purchase_modal/icon_message.tsx
nickmisasi: Instead of having discrete subtitle and errors in the DOM I just figured an error is a subtitle, but colour it red.

marianunez:
> Mattermost test server created! 🎉
> 
> Access here: [mattermost-webapp-pr-6583.test.mattermost.cloud](https://mattermost-webapp-pr-6583.test.mattermost.cloud)
> 
> Account Type	Username	Password
> Admin	sysadmin	Sys@dmin123
> User	user-1	User-1@123

@matthewbirtch gentle ping, this is ready for review ASAP given that some items in this PR is blocking some other work on the System Console.

actions/cloud.tsx
marianunez: The console error logs is fine but if we are returning an error here and displaying it somewhere then we need to localize that. If not, then does it make sense to return an error at all? We can return a generic error like the one commented next to it.

components/purchase_modal/icon_message.tsx
marianunez: Can you post a screenshot of a failure so we can have a UX review on that as well?

actions/cloud.tsx
nickmisasi: I've changed this function up. It logs the error if it runs into one, but displays generic error to the user. This now returns boolean true/false for success/failure respectively. Let me know what you think now!

matthewbirtch:
Apologies for the dealy @marianunez @nickmisasi. I'll be reviewing tonight

matthewbirtch (changes requested):
Hey @nickmisasi, first pass review here.

1. Looks like the background image isn't spanning the width of larger screens and is getting cut off on the right
Screen Shot 2020-09-30 at 8 25 14 PM

2. The x icon appears to be slightly off position - it should be equally positioned 12px from the top and right. Right now it is positioned 10px from the top and 20px from the right

3. A Scrollbar seems to be appearing even if it's not needed
image

4. For the illustration element in the left column, I think we need to work on getting the theme colors for the svgs. I believe Asaad had a solution for this and I think we did this for the onboarding stuff too? Right now there are definitely contrast issues with these illustrations
image

4. For the middle column (form content), rather than setting the width on the form to 75% and max-width of 448px in this case, let's remove those and just add left and right padding of 96px

5. A number of issues with the input fields UI. I just want to make sure you're using the work Devin had done?
    - The credit card input doesn't seem to be using proper theme colors
image
    - Inputs are supposed to be 40px in height, but the `Name on card`, `Address`, `Address 2`, `City`, `Zip Code` are 36px (or less in the case of the Zip code input)
    - the border color on the select input isn't the same as the text inputs
    - the down arrow icon in the select input doesn't match what Devin did and what's in the designs
    - You'll notice in the dark theme screenshot above, select inputs don't appear to be following theme colors.
image
    - the labels are not positioned correctly on all inputs
    - the placeholder text color on the inputes seems too light compared to the designs and makes it seem disabled. Although the credit card field placeholder text color seems more right
    - the corner radius doesn't seem quite right on the focus state because the thicker border appears on the outside rather than the inside - can you check with Devin on this? He recently fixed this problem
**Implemented**
image
**Designed**
image

6. The country input don't have the label
image

7. For select inputs, when the dropdown menu is open, I'm used to being able to type the first few characters of my provice or country and the menu jumps down to it. These select inputs don't do this standard behaviour

8. Need a space in front of 'See how billing works'
image

9. In the right column summary, can we add `line-height:1` to the `.price-text` line? This should reduce the spacing here to match closer with design

10. Is it possible to anchor the Mattermost Cloud logo to the bottom?

11. Can we use a proper theme colors for the logo as well? We should use the Button BG Color for Mattermost and 48% Center Channel color for the 'Cloud' portion of the logo

12. For the processing screen
    - we need to vertically center the content within the modal
    - we will need to look a theme colors for illustrations and the progress bar
    - it should say "**Verifying** your payment information". Right now it says "Verify..."
image

13. For the success screen
    - the computer screen illustration should be horizontally centered. It looks like the image is causing a visual off-centered imbalance because of the leaves as part of the image.    
    - it should also be consistently positioned with the previous processing state. It looks like it shifts
    - we need to vertically center the content within the modal
    - we will need to look a theme colors for illustrations
image

7. We'll need to work on the views for smaller screens. A mobile view was designed initially. You can see it here, but it might be beneficial to work together on this. Here's the link to the mobile design for the payment screen: https://www.figma.com/file/uv9A6dZsekQHHhOhYC7yc2/Cloud-UX?node-id=3273%3A106295

This is probably enough for now. Let me know if you would like to hop on a zoom call to go through any of it.

components/cloud/types/customer.ts
marianunez: FYI, that PR is merged so we can use that customer type here instead

components/purchase_modal/purchase_modal.tsx
```diff
@@ -0,0 +1,233 @@
...
+import {getNextBillingDate} from 'utils/utils';
+
+import PaymentForm from '../payment_form/payment_form';
+
+import './purchase.scss';
+import 'components/payment_form/payment_form.scss';
+import ProcessPaymentSetup from './process_payment_setup';
+
+const STRIPE_CSS_SRC = 'https://fonts.googleapis.com/css?family=Open+Sans:400,400i,600,600i&display=swap';
+const STRIPE_PUBLIC_KEY = 'pk_test_ttEpW6dCHksKyfAFzh6MvgBj';
```
marianunez, line 27: @joram this is the same as in CWS. Do we need to change this in the build process for the webapp?

nickmisasi:
Thanks again for the detailed evaluation @matthewbirtch 

I believe I've addressed everything in your comment, with the exception of things to do with themeing colours and responsive design. We do plan to do this before GA, through [this ticket](https://mattermost.atlassian.net/browse/MM-29214), but there are a few other core mechanics that are going to take priority. I'll address each of your comments explicitly, to help make sure I've gotten to them all:


> 1. Looks like the background image isn't spanning the width of larger screens and is getting cut off on the right

Should be spanning the entire width of the display now. Just noticed as I posted this screenshot that it looks like since the LHS is 25% of the screen, the middle column is farther away from it. Let me know what you'd like me to do here.
image


> 2. The x icon appears to be slightly off position - it should be equally positioned 12px from the top and right. Right now it is positioned 10px from the top and 20px from the right

Should be good here (see image above)

> 3. A Scrollbar seems to be appearing even if it's not needed

Wasn't really able to reproduce this, but I added a CSS attribute to make sure it won't happen anyways



> 5. For the middle column (form content), rather than setting the width on the form to 75% and max-width of 448px in this case, let's remove those and just add left and right padding of 96px

Done




> A number of issues with the input fields UI. I just want to make sure you're using the work Devin had done?

Didn't actually have Devin's dropdowns, since they weren't yet merged. He has merged them now, and I've integrated them for the country and state selector.



> * the labels are not positioned correctly on all inputs
> * the placeholder text color on the inputes seems too light compared to the designs and makes it seem disabled. Although the credit card field placeholder text color seems more right
> * the corner radius doesn't seem quite right on the focus state because the thicker border appears on the outside rather than the inside - can you check with Devin on this? He recently fixed this problem

Think I've got the labels in the correct position now. A little hard to decipher exactly where to place from Figma. Let me know.
I believe the placeholder text is good to go
Corner radius should be fixed by the inset borders



> 7. The country input don't have the label
> 8. For select inputs, when the dropdown menu is open, I'm used to being able to type the first few characters of my provice or country and the menu jumps down to it. These select inputs don't do this standard behaviour

Fixed by integrating Devin's dropdown

> 9. Need a space in front of 'See how billing works'

Pesky HTML. Fixed.

> 10. In the right column summary, can we add line-height:1 to the .price-text line? This should reduce the spacing here to match closer with design

Added. I didn't see any visual change though.

> 11. Is it possible to anchor the Mattermost Cloud logo to the bottom?

Anchored.







> For the processing screen

> * we need to vertically center the content within the modal

Centered

> * we will need to look a theme colors for illustrations and the progress bar

(will be done later)

> * it should say "**Verifying** your payment information". Right now it says "Verify..."

Fixed



> For the success screen

> * the computer screen illustration should be horizontally centered. It looks like the image is causing a visual off-centered imbalance because of the leaves as part of the image.

This one was a little challenging because of the leaves. Let me know how it looks, I can tweak it slightly if you think it needs it.

> * it should also be consistently positioned with the previous processing state. It looks like it shifts

I think I managed this as part of the above.

> * we need to vertically center the content within the modal

Should be good here

components/cloud/types/customer.ts
nickmisasi: Should be good here. I moved the sku's into `types/cloud/sku.ts` but removed this file in favour of Devin's mattermost-redux changes

matthewbirtch (approved):
@nickmisasi there are a few tweaks to make follow-up tickets for (see below). However, I'm going to approve as they are not blockers.

1. Credit card input doesn't seem to have a focus state
2. Credit card input label animates off strangely when I leave focus on the field

3. The border seems to shift in the focus state (I had this issue in one of Devin's PRs too) and it seems to be adding a white border around the outside (you can see it more against the wave background image. 

image
4. The underline includes the space on the 'See how billing works' hover state
image
5. Not sure there will be anything we can do about this, but when I use Chrome's autofill on the form, it highlights the fields that it fills, but the highlight doesn't seem to span the full input box oddly enough
image

components/purchase_modal/purchase_modal.tsx
marianunez: (Once again I mentioned wrong, apologies `@joram`!)

@jwilander see above

nickmisasi:
@matthewbirtch I've fixed 4, and I think I fixed 3 and 5 also. 1, and 2, I can note down and address as part of the theming, the Stripe card input is very hard to get styles to apply to.

components/purchase_modal/purchase_modal.tsx
```diff
@@ -0,0 +1,233 @@
...
+                            {this.purchaseScreen()}
+                            <div>
+                                <img
+                                    className='waves'
+                                    src={wavesBackground}
+                                />
+                                <img
+                                    className='blue-dots'
+                                    src={blueDotes}
+                                />
```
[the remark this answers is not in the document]
nickmisasi: These are the blue dots in the top left of the modal: 
image

I don't see dots in the bottom right on the designs - could you link?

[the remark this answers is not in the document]
marianunez: Screen Shot 2020-10-05 at 2 20 02 PM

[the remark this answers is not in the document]
nickmisasi: Totally couldn't see that in the designs until I zoomed in

I've added them - the positioning figma spit out wasn't very good, and the opacity there of 0.08 made it invisible on my screen - so I tweaked it slightly. Hope that's ok!

components/payment_form/state_selector.tsx
```diff
@@ -0,0 +1,65 @@
...
+    if (stateList.length > 0) {
+        return (
+            <DropdownInput
+                onChange={onStateSelected}
+                value={props.state ? {value: props.state, label: props.state} : undefined}
+                options={stateList.map((stateCode) => ({
+                    value: stateCode.code,
+                    label: stateCode.name,
+                }))}
+                legend={'State/Province'}
```
devinbinnie, line 45: These should be translated.

components/purchase_modal/icon_message.scss
```diff
@@ -0,0 +1,94 @@
...
+
+        .IconMessage-img {
+        }
+
+        .IconMessage-progress {
+            margin: 0 auto;
+            margin-top: 90px;
+            height: 4px;
+            width: 387px;
+            background: rgba(34, 64, 109, 0.12);
```
devinbinnie, line 55: Themeing?

components/purchase_modal/index.js
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
devinbinnie, line 1: This should be a `.ts` file.

components/purchase_modal/process_payment.css
```diff
@@ -0,0 +1,19 @@
+.ProcessPayment-progress {
+    margin: 0 auto;
+    margin-top: 60px;
+    height: 4px;
+    width: 387px;
+    background: rgb(34,64,109, 0.12);
```
devinbinnie, line 6: Does this need themeing?

components/purchase_modal/purchase.scss
```diff
@@ -0,0 +1,153 @@
...
+                padding: 32px 0px;
+            }
+        }
+
+        .central-panel {
+            width: 50%;
+        }
+
+        .full-width {
+            border-color: var(--center-channel-color-16)
```
devinbinnie, line 57: Same here with the use of `--center-channel-color-rgb`

images/onboarding-bg.svg
devinbinnie: Yeah looks like this overwrites the work we did to theme this.

images/onboarding-success.svg
```diff
@@ -0,0 +1,40 @@
+<svg width="444" height="313" viewBox="0 0 444 313" fill="none" xmlns="http://www.w3.org/2000/svg">
```
devinbinnie, line 1: Same with this file

utils/utils.jsx
```diff
@@ -1895,3 +1897,8 @@ export function adjustSelection(inputBox, e) {
         setSelectionRange(inputBox, selectionStart + 1, selectionEnd - 1);
     }
 }
+
+export function getNextBillingDate() {
+    const nextBillingDate = moment().add(1, 'months').startOf('month');
```
devinbinnie, line 1902: If this is always the start of the month, wouldn't that make the date always the 1st?
If so, couldn't we just use the normal `Date` object?
```suggestion
    const nextBillingDate = new Date();
    nextBillingDate.setMonth(nextBillingDate.getMonth() + 1);
    nextBillingDate.setDay(1);
```

I only say that because you generally don't need `moment.js` for most things. They even say that themselves:
https://momentjs.com/docs/#/-project-status/

devinbinnie (changes requested):

utils/utils.jsx
nickmisasi: Couple reasons I'd prefer to keep moment as is:
1. Formatting the date is cleaner
2. Moment usually does some magic in the background around localization with its formatting. Things beyond just "translate December to french", but also how the month/day/year is ordered

components/payment_form/card_image.tsx
```diff
@@ -0,0 +1,54 @@
...
+            />
+        );
+    }
+
+    return null;
+}
+
+function getCardImage(brand: string): string {
+    switch (brand) {
+    case 'amex':
```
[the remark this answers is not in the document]
nickmisasi: It's not returning a string, but a component (which is the cards image). '' means you can check for if it exists or not, and conditionally render the card (or not)

If you returned brand, this check would break. It could be anything falsey, but can't be truthy (ie, brand)

components/payment_form/card_input.tsx
```diff
@@ -0,0 +1,185 @@
...
+} & OwnProps;
+
+type State = {
+    focused: boolean;
+    error: string;
+    empty: boolean;
+    complete: boolean;
+}
+
+const REQUIRED_FIELD_TEXT = 'This field is required';
```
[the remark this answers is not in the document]
nickmisasi: These actually are localized before being displayed

components/payment_form/card_input.tsx
```diff
@@ -0,0 +1,185 @@
...
+    private validateInput = () => {
+        const {required} = this.props;
+        const {empty, complete} = this.state;
+        let error = '';
+
+        this.setState({error: ''});
+        if (required && empty) {
+            error = REQUIRED_FIELD_TEXT;
+
+            //t('shortcuts.header'),
```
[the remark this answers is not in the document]
nickmisasi: Nope

components/purchase_modal/icon_message.scss
nickmisasi: Theming will be fixed as part of: https://mattermost.atlassian.net/browse/MM-29214

components/purchase_modal/process_payment.css
nickmisasi: Same here

components/purchase_modal/purchase.scss
nickmisasi: fixed

components/cloud/utils/states.ts
```diff
@@ -0,0 +1,79 @@
+
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export interface StateCode {
+    code: string;
+    name: string;
+}
+
+export const US_STATES = [
```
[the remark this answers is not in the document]
nickmisasi: There was no folder for `constants` in utils, so I moved the `states.ts` file into `utils` folder. Couldn't put in utils/constants.jsx because typescript and

components/payment_form/payment_form.tsx
```diff
@@ -0,0 +1,334 @@
...
+        super(props);
+
+        this.cardRef = React.createRef<CardInputType>();
+
+        this.state = this.getResetState(props);
+    }
+
+    public componentDidUpdate(prevProps: Props) {
+        if (prevProps.paymentMethod == null && this.props.paymentMethod != null) {
+            this.setState(this.getResetState()); //eslint-disable-line react/no-did-update-set-state
```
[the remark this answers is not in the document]
nickmisasi: getDerivedStateFromProps doesn't have access to the previous props, which is needed in these cases. React docs for getDerivedStateFromProps say:
```
If you need to perform a side effect in response to a change in props, use componentDidUpdate lifecycle instead.
```
I think the side-effect here is clearing the form. 

I've tweaked how this is so we aren't eslint ignoring things anymore - let me know what you think!

actions/cloud.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export function getProductPrice() {
+    return async () => {
+        let cloudProducts = [] as Product[];
+
+        try {
+            cloudProducts = await Client4.getCloudProducts();
+        } catch (error) {
+            // eslint-disable-next-line no-console
+            console.error(`Error fetching cloud products: ${error}`);
```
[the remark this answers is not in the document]
nickmisasi: These were actually debugging ones - I don't think these should be shared to the console/user since its around payments. You're right, they'll be put on the redux error stack.

actions/cloud.tsx
```diff
@@ -0,0 +1,99 @@
...
+import {getCode} from 'country-list';
+
+import {Client4} from 'mattermost-redux/client';
+import {Product} from 'mattermost-redux/types/cloud';
+
+import {getConfirmCardSetup} from 'components/payment_form/stripe';
+
+import {StripeSetupIntent, BillingDetails} from 'types/cloud/sku';
+
+export function getProductPrice() {
```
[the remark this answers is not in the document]
nickmisasi: Good callout. Should be fixed in newest diff

utils/utils.jsx
devinbinnie: Okay, we can leave as-is for now. Might work on removing `moment.js` throughout the app at some point. But i'm only 1/5 on that.

components/payment_form/payment_form.tsx
```diff
@@ -0,0 +1,331 @@
...
+import StateSelector from './state_selector';
+import CardInput, {CardInputType} from './card_input';
+import CardImage from './card_image';
+
+import './payment_form.scss';
+import 'components/input.css';
+
+type Props = {
+    className: string;
+    initialBillingDetails?: BillingDetails | null;
```
devinbinnie, line 26: Since, this is an optional parameter, the type evaluates to `BillingDetails | null | undefined`.
For simplicity, could we remove the null and just check for `undefined` or just do `!this.props.initialBillingDetails`

components/purchase_modal/purchase_modal.tsx
```diff
@@ -0,0 +1,252 @@
...
+const stripePromise = loadStripe(STRIPE_PUBLIC_KEY);
+
+type Props = {
+    show: boolean;
+    isDevMode: boolean;
+    products?: Dictionary<Product>;
+    actions: {
+        closeModal: () => void;
+        getCloudProducts: () => void;
+        completeStripeAddPaymentMethod: (stripe: Stripe, billingDetails: BillingDetails, isDevMode: boolean) => any;
```
devinbinnie, line 42: The return type should be something like `boolean | ServerError`

devinbinnie (approved):
Just a couple TS nit-picks but otherwise this is looking good!
Nice work @nickmisasi!

marianunez:
@stevemudie this is ready for `QA Review`, I've added some additional QA steps in the epic: https://mattermost.atlassian.net/browse/MM-27142

components/purchase_modal/purchase_modal.tsx
jwilander: Yes, we'll need to replace this when building releases

stevemudie (approved):
Worked with Nick and muscled through the tests we could do on it.
Looks good and will of course test again once it's fully in place.

components/purchase_modal/purchase_modal.tsx
nickmisasi: Think we're good here now!

components/purchase_modal/purchase_modal.tsx
ali-farooq0: is this a real key? ahh should this even be here? I propose removing it from the webapp @marianunez

PR #5855, opened 2020-07-04
Limit the entires to 20 and changed the order of display in the more …

pradeepmurugesan:
…DM dialog

#### Summary
Right now, on clicking the  `more` under `Direct Messages` opens a modal and lists users in alphabetical order followed by the group channels.

This PR limits the entries to 20 and orders the entries by recent direct messages, users and then group channels.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/14198

#### Related Pull Requests

- Has server changes - NO
- Has redux changes - NO
- Has mobile changes -NO 

#### Screenshots

Screenshot 2020-07-10 at 11 22 45

Screenshot 2020-07-10 at 11 22 27

Screenshot 2020-07-10 at 11 22 08

PR #6585, opened 2020-09-30
MM-27903, MM-27899: Folded Reply Threads: ThreadItem and ThreadFooter

calebroseland:
#### Summary

View in storybook to view all states (hover, selected, following, etc.). 

Only `<Timestamp/>` and `<Avatar/>` changes affect production codepaths. All other components just in storybook for now.

- **[MM-18107](https://mattermost.atlassian.net/browse/MM-18107): Folded Reply Threads, Components/Building Blocks**
  - **[MM-27903](https://mattermost.atlassian.net/browse/MM-27903): Global Threads: `<ThreadItem/>`**
  - **[MM-27899](https://mattermost.atlassian.net/browse/MM-27900): Channel Threads: `<ThreadFooter/>`**
  - Supporting components: 
    -  [MM-27899](https://mattermost.atlassian.net/browse/MM-27899): `<Avatars/>`
    - `<Avatar/>` (improvements to existing)
      - new size opts: 
        - `xxs`: 16
        - `xs`: 20 (previously 16; 16 and its utilizations move to `xxs`)
        - other sizes unchanged `sm`, `md`, `lg`, `xl`, `xxl`
    - `<Button>`
    - `<FollowButton>`
    - `<ThreadMenu>`
    - `<SimpleTooltip>`
- `<Timestamp/>` improvements
  - Standard Units (`RangeDescriptor`s) for easier relative time usage
  - Relative time `style` now defaults to `long`, rendering '25 seconds' instead of '25 sec.'
  - More consistent `useSemanticOutput` behavior; passing children no longer bypasses `<SemanticTime/>` wrapping
  - Storybook stories to experiment with relative time formats
- Storybook improvements
  - Nav/org restructuring
  - `<StoryGrid>`/`<StoryBox>`
  - Minimal state mocking (should be refactored as/if it scales)

#### Screenshots
Screen Shot 2020-09-30 at 03 02 25
Screen Shot 2020-09-30 at 02 33 20
Screen Shot 2020-09-30 at 16 37 22
Screen Shot 2020-09-30 at 02 10 22
Screen Shot 2020-09-30 at 03 11 51
Screen Shot 2020-09-30 at 16 37 41
Screen Shot 2020-09-30 at 02 42 08


.eslintrc.json
```diff
@@ -58,7 +58,10 @@
       }
     ],
     "react/no-string-refs": 2,
-    "no-only-tests/no-only-tests": ["error", {"focus": ["only", "skip"]}]
+    "no-only-tests/no-only-tests": ["error", {"focus": ["only", "skip"]}],
+    "react/style-prop-object": [2, {
+      "allow": ["Timestamp"]
```
calebroseland, line 63: Allow `style` prop on `<Timestamp/>`

.eslintrc.json
```diff
@@ -58,7 +58,10 @@
...
+    }]
```
calebroseland, line 64: Allow `style` prop on `<Timestamp/>`, of `Intl.RelativeTimeFormat`

components/widgets/icons/icons.stories.js
```diff
@@ -77,7 +78,7 @@ const icons = [
     AccordionToggleIcon, AlertIcon, ArchiveIcon, AtIcon, ArrowRightIcon,
     AttachmentIcon, BackIcon, BotIcon, CheckboxCheckedIcon,
     CheckboxPartialIcon, CloseCircleIcon, CloseCircleSolidIcon, CloseIcon,
-    DownloadIcon, DraftIcon, EmailIcon, EmojiIcon, FaAddIcon, FaBackIcon, FaDropdownIcon,
+    DotsHorizontal, DownloadIcon, DraftIcon, EmailIcon, EmojiIcon, FaAddIcon, FaBackIcon, FaDropdownIcon,
```
calebroseland, line 81: DotsHorizontal was feeling left out

components/widgets/menu/menu_items/menu_item.tsx
```diff
@@ -13,7 +13,7 @@ export default function menuItem(Component: React.ComponentType<any>) {
         icon?: React.ReactNode;
         text?: React.ReactNode;
     }
-    class MenuItem extends React.PureComponent<Props|React.ComponentProps<typeof Component>> {
+    class MenuItem extends React.PureComponent<Props & React.ComponentProps<typeof Component>> {
```
calebroseland, line 16: The union was hiding `id`, `show`, etc.. from type inference downstream

components/threading/common/button.scss
```diff
@@ -0,0 +1,86 @@
...
+            width: 1px;
+            height: 15px;
+            position: absolute;
+            left: -.5px;
+            background: rgba(var(--center-channel-color-rgb), .16);
+            content: '';
+            display: inline-block;
+        }
+    }
+}
```
calebroseland, line 77: This is what creates the separator when two `.Button`s or `.Separated`s are next to each other.

calebroseland:
@jespino do you know if/where the CI build publishes the storybook build?

components/threading/global_threads/thread_item/thread_item.scss
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    p {
+        text-overflow: ellipsis;
+        overflow: hidden;
+
+        // cross-browser multi-line text-overflow support
+        display: -webkit-box;
+        -webkit-line-clamp: 2;
+        -webkit-box-orient: vertical;
```
calebroseland, line 104: yes, `-webkit-` prefix even in FF

jespino:
It is here for master branch: https://alpha-storybook.devops.mattermost.com/master/index.html
and here for the PR: https://alpha-storybook.devops.mattermost.com/feature/threading-overhaul-comps-a/index.html

components/simple_tooltip.tsx
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jespino, line 1: Should it be in the widgets?

components/threading/channel_threads/channel_threads.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+    'day',
+    'week',
+    'month',
+    'quarter',
+    'year',
+];
+
+const users = [
+    {
+        url: '/api/v4/users/1/image?_=0',
```
jespino, line 31: Can we change the urls to `./api/v4/users/...`, that would make it work in the local and online storybook.

components/threading/channel_threads/thread_footer/index.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jespino, line 1: Creating a folder, normally is the approach that we follow for connected components, I'm not against using it here, but I want to be sure you did it on purpose. I think we can decide to follow this approach always, or not follow the approach at all, but having it sometimes and sometimes not can be a bit confusing. What do you think @deanwhillier? What do you think @calebroseland?

components/threading/channel_threads/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,108 @@
...
+    isFollowing,
+    actions: {
+        follow,
+        unfollow,
+        requestOpenThread: open,
+    },
+}: Props) {
+    return (
+        <div className='ThreadFooter'>
+            {Boolean(newReplies) && <>
```
jespino, line 48: Why the Fragment (`<>`) for a single component?

components/threading/channel_threads/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,108 @@
...
+                </SimpleTooltip>
+            </>}
+
+            <Avatars
+                users={participants}
+                size='sm'
+            />
+
+            <Button
+                onClick={() => open()}
```
jespino, line 71: This is creating a new function every render, can we just pass here `open`? as far as open is not handling any parameter, should be ok.

components/threading/channel_threads/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,108 @@
...
+                <FormattedMessage
+                    id='threading.numReplies'
+                    defaultMessage='{totalReplies, plural, =0 {Reply} =1 {# reply} other {# replies}}'
+                    values={{totalReplies}}
+                />
+            </Button>
+
+            <FollowButton
+                isFollowing={isFollowing}
+                start={() => follow()}
```
jespino, line 83: Same here with the follow/unfllow.

components/threading/channel_threads/thread_footer/index.ts
jespino: Ok, I guess this is going to be eventually connected as least to pass the follow/unfollow actions, right?

components/threading/global_threads/global_thread.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+    'day',
+    'week',
+    'month',
+    'quarter',
+    'year',
+];
+
+const users = [
+    {
+        url: '/api/v4/users/1/image?_=0',
```
jespino, line 30: Same here, add the `.` at the beginning to make it work locally and in the CI deployed version.

components/threading/global_threads/thread_item/thread_item.tsx
```diff
@@ -0,0 +1,132 @@
...
+    totalReplies: number,
+
+    isFollowing: boolean,
+    isSaved: boolean,
+    isSelected: boolean,
+
+    actions: {
+        select: () => void,
+    },
+} & Pick<ComponentProps<typeof ThreadMenu>, 'actions'>;
```
jespino, line 36: Nice one

components/threading/global_threads/thread_menu/thread_menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+                        text={formatMessage({
+                            id: t('threading.threadMenu.save'),
+                            defaultMessage: 'Save',
+                        })}
+                    />
+                )}
+                <Menu.ItemAction
+                    onClick={copyLink}
+                    text={formatMessage({
+                        id: t('threading.threadMenu.copy'),
```
jespino, line 136: I think with the formatMessage you don't need the `t` function there.

components/threading/global_threads/thread_menu/thread_menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+                content={formatMessage({
+                    id: 'threading.threadItem.menu',
+                    defaultMessage: 'Actions',
+                })}
+            >
+                {children}
+            </SimpleTooltip>
+
+            <Menu
+                ariaLabel={''}
```
jespino, line 68: empty aria label?

components/timestamp/timestamp.tsx
```diff
@@ -142,7 +148,7 @@ class Timestamp extends PureComponent<Props, State> {
 
         // relative
         numeric: 'auto',
-        style: 'narrow',
+        style: 'long',
```
jespino, line 151: Changing this default can affect other usages of this component, are we checking that in other places?

components/widgets/users/avatar/index.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jespino, line 1: Ok, here is clearly not waiting for future component connections, is this the new pattern for components? always in its own folder? @calebroseland @deanwhillier 
I think the the proliferation of other files around the components like the `scss`, `stories.tsx`, `test.tsx`, `__snapshots`... probably makes a lot of sense to make it this way.

jespino (changes requested):
Some changes needed, but looks pretty good to me. Great work @calebroseland! :confetti_ball:

components/simple_tooltip.tsx
calebroseland: I guess it could be, but I thought it was a bit more generic that what most of the widgets+stories are.

components/threading/channel_threads/thread_footer/thread_footer.tsx
calebroseland: I think a remnant of an earlier version, will change 👍

components/threading/channel_threads/thread_footer/thread_footer.tsx
calebroseland: We could, but I usually like to keep event handlers and actions/functionality separate to make it easier to handle interactions a little easier and in this case prevent future type collisions if the actions ever get parameters.

Open to thought. Should I switch it to a class component and put the event handlers there?

components/threading/global_threads/thread_menu/thread_menu.tsx
calebroseland: Menu requires it, waiting to to accessibility dev/work until integration stage.

components/timestamp/timestamp.tsx
calebroseland: Yes, though I don't think relative time is being used anywhere else yet. There's a contributor PR that is using it, but it's still in review and [specifies long](https://github.com/mattermost/mattermost-webapp/pull/5855/files/bfc2293620b64418f327462f10ecf7f52f7709b0#diff-035f90f92c7c4207c4bbca0cc40c402eR19).

components/threading/global_threads/thread_menu/thread_menu.tsx
calebroseland: It'll work without it, yes, but I think `t` is mainly for i18n tooling? (e.g. `make i18n-extract`)

matthewbirtch (changes requested):
@calebroseland this is AWESOME!

A number of minor things I notices...this probably looks worse than it is.

## Thread Footer
1. When hovering over the avatars, the tooltip doesn't appear to be positioning properly (as you noted in your description)
2. Also, when hovering, there should be a subtle shadow on the hovered avatar (see below)

image
Should be like this:
image

3. When the unread dot is not showing, everything shifts left. The plan was for the avatars to always stay positioned statically and aligned to the left edge of the post text content. They shouldn't shift when the unread dot is hidden
image

4. The tooltip on the unread dot is also not positioned quite right (as you noted in your description)
image

4. The color of the replies and follow text should be the Center Channel Color at 64% opacity. Looks like it's using #333333 at full opacity

5. The hover state for the thread footer area should look like the below. Curently is using blue (I changed that a while ago). 
image

6. The hover state for the follow and replies buttons should look like the following
image

7. For the last reply date, I can't seem to get it to show a time less than a day. If I select 4 minutes, it still says 'Last reply today'

8. If I change the number of participants to more than 7, the `+4` doesn't seem to update. We will want to have some kind of limit here though. We don't have space to show more than 2 digits, so we'll want to limit this to say 99+ if the remaining users is more than 99

## Thread List Item
1. The `•••` icon button isn't quite aligned with equal margin on top and right. See below for how it's supposed to be positioned
**Designed**
image
**Implemented**
image

2. It looks like the time stamp may also need to nudge to the right a bit more as well

3. Text color is #333333, the default theme's text color is #3D3C40

4. When I click the `•••` icon it shifts to the left

5. The popover menu from the `•••` button goes off screen. Can we align the menu with the right side of the button?
image

5. If I turn off 'is following', the popover menu changes to say `Keep Following`. I know we would never really see this state int he threads list, but it should say 'Follow thread - You will get notified about replies'

5. Can the popover menu be a bit wider so the help text doesn't wrap on two lines? Not a big deal, but wanted to check

5. The tooltips on the `•••` icon button isn't positioned correctly (as you noted in your description)
image

7. We'll need to truncate the channel name here. I can get into a scenario where it wraps on two lines. we won't want that. We may need to do that on the name as well

8. For mentions, if there are more than 9, maybe we should just show 9+? Or we can go with a 2-digit limit and show 99+ for another over 99, but that might break things

components/threading/global_threads/thread_menu/thread_menu.tsx
jespino: Yes, t is for translations extractions when you can't infer it from the code, in this case formatmessage is enough for the extraction tool to know that is a translatable string.

components/simple_tooltip.tsx
jespino: I think, for the widgets, the simpler the better, widgets shouldn't have anything specially complex.

components/threading/channel_threads/thread_footer/index.ts
jespino: Lets wait until Dean give his opinion, probably we can create some help wanted tickets around moving existing components with sass file aside, into directory components.

components/threading/channel_threads/thread_footer/thread_footer.tsx
jespino: as a functional component there is no much to do, probably creating the function and store it in a module level variable, but create it in the context of the function, something like:

```js
var open = null;
function Component() {
  open = open || () => {
    this.props.actions.whatever()
  }
}
```

components/threading/global_threads/thread_menu/thread_menu.tsx
calebroseland: Ahh, I didn't realize that. I thought it only picked it up `<FormattedMessage/>`.

calebroseland:
@matthewbirtch

### Thread Footer
1. Found/fixed tooltip-in-SB issue:
a. Alignment
b. Added immediate transition between `<Avatar/>`s (`500ms` open delay on first-enter, `0ms` between and exit)

2. Elevations added to:
a. `<Avatar/>` (only in `<Avatars/>`)
b. `<Tooltip/>` (globally)
3. Indicator no longer collapses (so others doesn't shift)
4. (SB-tooltip fixed)
5. Color updated
6. Blue-hover removed
7. Updated hover state button color
8. Positive numbers went forward in time; switched knob to absolute datetime to be clearer
9. There were only 7 mock users available. It was more designed to show smaller number UX (e.g. 2, 3, 4 etc.)
I updated the logic here quite a bit, so named can be up to 200 (Alan Arkin repeated 196x), and unnamed ("and ### others") is unlimited.

### Thread List Item
1. Alignment fixed. Unintentionally I had aligned it with the timestamp it visually replaces, but aligning with the corner makes perfect sense, too


2. Fixed, padding-right was off
3. Text color updated
4. Can't reproduce this, what browser/viewport/zoom settings/etc. do you have?
5. At desktop viewports it is opening left, that looks like a bug in `<Menu/>` at mobile viewports, should fix in another ticket IMO (as it may affect other areas)
6. Wording updated
7. Width increased
8. (SB-tooltip fixed)
9. Added `nowrap` to entire line (name, team, timestamp)
10. '99+' is possible if we drop from font-size 10px to 8px. Up to you.
Screen Shot 2020-10-07 at 16 18 59Screen Shot 2020-10-07 at 16 19 18



- Also updated `<Avatars/>` in ThreadItem to be 20px as per spec/design; had to move existing size `xs` (was 16px, now 20px) to `xss` (new, 16px)
- On the topic, I assume `<Avatar/>`s should behave as `right-over/covers-left`? (is mixed in design)

matthewbirtch:
Awesome work @calebroseland. Answers to your questions below:

> * Added `nowrap` to entire line (name, team, timestamp). What should the cutoff width/ratios be for name/team?

By `team` I assume you mean `channel`, but do you need character counts or pixel widths? Based on the designed width of the inbox (400px), I would go with these numbers for max widths to truncate at:
- Name: 148px (roughly 18 characters)
- Channel name: 120px (roughly 18 characters)
- Timestamp: 64px (roughly 12 characters)
- can we use shortforms for the time stamp? ("5 mins ago", "5 hrs ago", "5 days ago")


> 10\. '99+' is possible if we drop from font-size 10px to 8px. Up to you.

Let's do that. It's very tiny, but the chances of seeing this are quite slim I think. Also,wWe should do the same thing for the extra participants circle too if it goes beyond 99. 

> * On the topic, I assume `<Avatar/>`s should behave as `right-over/covers-left`? (is mixed in design)

Oh, you're right. I didn't even realize that. I think how you have it is fine as-is. Consistent between the two with the right side covering the left. Good catch

components/threading/channel_threads/thread_footer/index.ts
deanwhillier: Sorry for the delay. We talked about this on my team/guild and decided to start putting single components in folders with index.js files whether connected or not.

components/widgets/users/avatar/index.ts
deanwhillier: Yes, this will be the new pattern moving forward. :)

calebroseland:
- ~~Team~~ Channel—thanks for calling that out, staring at 'Enterprise Team' worked itself pretty deep, apparently! (clarified in code)

I think we could skip the truncating for the time being (can come back to it). Timestamp can use shortform, yes. Perhaps later during integration phase we can add responsiveness to toggle between short/normal.

Sounds good on the 99+. The extra participants is doing that already, I'll go `+1` ... `+99` then `99+`.

components/simple_tooltip.tsx
calebroseland: moved to widgets ✓

components/threading/channel_threads/thread_footer/thread_footer.tsx
calebroseland: ✓ Dropped the inline functions, and where necessary used `useCallback`

matthewbirtch (approved):
This is good to go @calebroseland. Awesome stuff

components/threading/global_threads/global_threads.tsx
```diff
@@ -0,0 +1,3 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
```
jespino, line 3: Why this empty file?

jespino (approved):
LGTM

components/threading/global_threads/global_threads.tsx
calebroseland: Just forgot to delete after init scaffolding, deleted now

components/threading/global_threads/thread_item/thread_item.tsx
```diff
@@ -0,0 +1,136 @@
...
+}: Props) => {
+    return (
+        <div
+            className={classNames('ThreadItem', {
+                'has-unreads': newReplies,
+                'is-selected': isSelected,
+            })}
+            tabIndex={0}
+            onClick={actions.select}
+        >
```
deanwhillier, line 67: Here's another case I'd considering using an `<article>` tag. That would also allow you to 'reset' the heading level to `<h1>` and not have to worry about counting heading levels to try to get the correct semantic level here.

components/threading/global_threads/thread_item/thread_item.tsx
```diff
@@ -0,0 +1,136 @@
...
+            <h4>
+                {Boolean(newMentions || newReplies) && (
+                    <div className='indicator'>
```
deanwhillier, line 70: Ditto on `indicator`.

components/widgets/badges/badge.tsx
```diff
@@ -1,32 +1,37 @@
...
-}
+    return (
+        <div className='Badge'>
+            <div
+                {...attrs}
+                role={attrs.onClick ? 'button' : undefined}
+                className={'Badge__box ' + className}
+            >
+                {children}
+            </div>
```
deanwhillier, line 32: If we are going to make this optionally clickable, I'd prefer to use an actual button or link element instead of the role attribute unless it absolutely doesn't make sense or for legacy reasons just isn't currently possible.

components/widgets/badges/badge.scss
```diff
@@ -13,6 +13,10 @@
     line-height: 14px;
     background: var(--center-channel-color-16);
 
+    &[role='button'] {
+        cursor: pointer;
+    }
+
```
deanwhillier, line 19: See comment about using an actual button.

components/widgets/users/avatar/avatar.scss
```diff
@@ -0,0 +1,74 @@
+.Avatar {
+    border-radius: 50%;
+    -webkit-user-select: none; /* Chrome all / Safari all */
+    -moz-user-select: none;    /* Firefox all */
+    -ms-user-select: none;     /* IE 10+ */
+    user-select: none;
```
deanwhillier, line 6: Purely as an aside, we really need to get our tooling to handle this! 😛

deanwhillier (changes requested):
Awesome work so far @calebroseland! Some items to consider below and:

- It might be worth considering updating html element semantics for the posts used as part of the new collapsed thread view. I was contemplating making each parent post an `<article>` and then using `<header>`, `<footer>`, etc. elements as appropriate inside. Not something for this PR necessarily, but perhaps something to consider as you start integrating the components into the actual webapp. Happy to chat about this at any point if you'd like.

components/threading/channel_threads/thread_footer/thread_footer.scss
```diff
@@ -0,0 +1,54 @@
+.ThreadFooter {
+    display: inline-flex;
+    align-items: center;
+
+    padding: 5px 10px;
+
+    font-family: 'Open Sans';
```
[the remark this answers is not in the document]
calebroseland: That was probably a leak while copying a few lines from the Figma spec, will remove

components/threading/channel_threads/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,116 @@
...
+                    content={
+                        <FormattedMessage
+                            id='threading.numNewMessages'
+                            defaultMessage='{newReplies, plural, =0 {no unread messages} =1 {one unread message} other {# unread messages}}'
+                            values={{newReplies}}
+                        />
+                    }
+                >
+                    <div
+                        className='indicator'
```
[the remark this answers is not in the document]
calebroseland: Over in `<ThreadItem/>`, it contains either a `.dot-unreads` or a `.dot-mentions`, so it was more for brevity I guess, given that in-context (`.dot-unreads` below, `newReplies ? ...` above) seems quite clear I think? I'm fine with either way.

components/threading/channel_threads/thread_footer/thread_footer.scss
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    .Avatars {
+        margin-right: 4px;
+        font-weight: 600;
+    }
+
+    .alt-visible {
+        display: none;
+    }
```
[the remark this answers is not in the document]
calebroseland: It means this class will be alternatively visible, e.g. `:hover`, `:focus-within`, etc.

Inverse class names might be something like:

```diff
- .alt-visible {}
+ .normally-hidden {}

- .alt-hidden {}
+ .normally-visible {}
```

components/threading/channel_threads/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+            <Avatars
+                users={participants}
+                totalUsers={totalParticipants}
+                size='sm'
+            />
+
+            <Button
+                onClick={open}
+                prepend={<ReplyIcon className='Icon'/>}
```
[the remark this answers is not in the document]
calebroseland: Ahh, yes—we should. I saw it in `widgets/icons` and just reached for this one first. Is there a central/common Icon component in the works that renders font-based icons? Or just use a good ol' `i`/`span`?

components/widgets/badges/badge.tsx
calebroseland: I agree, and I can switch it out; was trying to keep changes to other codepaths as minimal as possible for the time being, but I guess it wouldn't be that large of an impact.

components/widgets/users/avatar/avatar.scss
calebroseland: Absolutely! Only showing in the diff here cause I moved the component/files into a folder.

calebroseland: And perhaps we don't even need them now that we've dropped IE and all the other browsers should be respecting this (any many other previously prefixed properties) now.

components/threading/channel_threads/thread_footer/thread_footer.tsx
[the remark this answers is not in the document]
deanwhillier: Ah yes, I see. Naming is probably one of the hardest things about software development 😛.

I can't think of an obvious replacement. I guess what threw me off initially was the empty div with that className applied when there are no unreads/mentions. In the ThreadItem you remove the div entirely, but leave it empty in the ThreadFooter. When the div is empty, the className doesn't speak to what it is an indicator name for. Unless you can think of something simple, probably not worth worrying about this. :)

components/threading/channel_threads/thread_footer/thread_footer.scss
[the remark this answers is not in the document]
deanwhillier: Okay, I see. That's more utility than semantic then. I haven't seen that before. If you think this is readily understood, then I'm okay to ignore this comment. :)

components/threading/channel_threads/thread_footer/thread_footer.tsx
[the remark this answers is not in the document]
deanwhillier: There will be a common component for the icon font, but it doesn't exist yet. Thinking about it a bit more, maybe we just leave this for now so we don't do more work than necessary. Apologies for the noise! :P

components/widgets/users/avatar/avatar.scss
deanwhillier: Ya, we will need to remove them all at some point ... don't worry about now unless my mentioning it is now bugging you! :P

components/threading/channel_threads/thread_footer/thread_footer.tsx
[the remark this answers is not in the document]
calebroseland: Nah, not noise—didn't take long to swap out :)

deanwhillier (approved):
👍

ogi-m (approved):
Done smoke tests around date separators, post times and avatars, LGTM 👍

PR #6591, opened 2020-09-30
Migrate team_settings_modal to typescript

sowmiyamuthuraman:
Fixes https://github.com/mattermost/mattermost-server/issues/15477

JIRA: https://mattermost.atlassian.net/browse/MM-20463


components/team_settings_modal/index.ts
```diff
@@ -6,9 +6,9 @@ import {connect} from 'react-redux';
 import {ModalIdentifiers} from 'utils/constants';
 import {isModalOpen} from 'selectors/views/modals';
 
-import TeamSettingsModal from './team_settings_modal.jsx';
+import TeamSettingsModal, {State} from './team_settings_modal';
 
-function mapStateToProps(state) {
+function mapStateToProps(state:State) {
```
sudheerDev, line 11: Shouldn't this be `GlobalState` from `import {GlobalState} from 'types/store';`?

components/team_settings/team_settings.jsx
```diff
@@ -61,7 +61,7 @@ TeamSettings.propTypes = {
     updateSection: PropTypes.func.isRequired,
     closeModal: PropTypes.func.isRequired,
     collapseModal: PropTypes.func.isRequired,
-    team: PropTypes.object.isRequired,
+    team: PropTypes.object,
```
sudheerDev, line 64: Why are we removing required here?

sudheerDev (changes requested):
Hey, @sowmiyamuthuraman Thank you. I have a couple of questions here.

components/team_settings/team_settings.jsx
sowmiyamuthuraman: @sudheerDev **team** prop is not passed to TeamSettings component used  in the team_settings_modal.

hahmadia: It's not being directly passed as prop into the component however it is being passed by the redux as a prop which is why you cannot remove this. You can see it in `components/team_settings/index.js` in the `mapStateToProps` function.

components/team_settings_modal/index.ts
hahmadia: ^^ Agreed

components/team_settings_modal/team_settings_modal.tsx
```diff
@@ -3,22 +3,29 @@
...
+const SettingsSidebar = React.lazy(() => import('components/settings_sidebar'));
 
 import TeamSettings from 'components/team_settings';
 
-export default class TeamSettingsModal extends React.PureComponent {
-    static propTypes = {
-        onHide: PropTypes.func,
-    };
+type Props = {
+    onHide: () => any
```
hahmadia, line 16: Let's follow the pattern how it's being elsewhere. I'd suggest it to be
```suggestion
    onHide: () => void;
```

hahmadia (changes requested):
Along with suggestion Sudheer made, I have one small change. Thanks!

components/team_settings/team_settings.jsx
sowmiyamuthuraman: @hahmadia how can we fix the typescript error (ie) team prop is required?

hahmadia: @sowmiyamuthuraman Can you do
`getCurrentTeam(state) || {}` in https://github.com/mattermost/mattermost-webapp/blob/master/components/team_settings/index.js#L12 and see if that fixes your problem? Let me know! Thanks

sowmiyamuthuraman: @hahmadia Still facing the same issue

```Property 'team' is missing in type '{ activeTab: string; activeSection: string; updateSection: (section: string) => void; closeModal: () => void; collapseModal: () => void; }' but required in type 'Pick<Pick<Pick<{ activeTab: any; activeSection: any; updateSection: any; closeModal: any; collapseModal: any; team: any; }, "updateSection" | "activeSection" | "closeModal" | "collapseModal" | "activeTab">, never> & Pick<...>, "team" | ... 2 more ... | "collapseModal">'```

Comment (hahmadia) [reply components/team_settings/team_settings.jsx:64]: In that case I'm not sure what to say then. I assume this change is ok then to be made. @sudheerDev do you have any suggestions?

Comment (hahmadia) [reply components/team_settings/team_settings.jsx:64]: I looked into it further and I believe you need to change it to optional only cause you aren't directly passing it as a prop and therefore typescript is gonna complain. Unless Sudheer has a solution, I think it's fine to leave it as it is.

**Review by hahmadia — APPROVED** [review]
With the exception of the `isRequired` comment from Sudheer which I don't have a solution for (I believe it's fine to leave out the isRequired since we aren't directly passing in team prop and therefore typescript complains), the rest of the code looks good to me. Thanks for the contribution!
THANKYOU

File: components/team_settings/team_settings.jsx (hunk 2 above)
Comment (sudheerDev) [reply components/team_settings/team_settings.jsx:64]: That is weird, but i am gonna guess it is because this is not a ts file, and it is weird we say required and we have a check for team object at the start, 

```
if (!team) {
        return null;
    }
```

Lets go ahead with the changes as this is a minor nit tbh

sudheerDev (approved):
Thank you @sowmiyamuthuraman

stevemudie (approved):
LGTM

hanzei:
Thanks for the contribution @sowmiyamuthuraman :+1:

PR #6594, opened 2020-09-30
MM-27498 Implement new unread filter

hmhealey:
This PR implements the new version of the unread filter as specced out [here](https://www.figma.com/proto/zVNYaShAXDK8i1RdpuEkr8/Channel-Sidebar---All-Phases?node-id=2185%3A1327&viewport=-5533%2C4423%2C1&scaling=min-zoom). Instead of the unread filter behaving the same as temporarily collapsing all of the channel categories, it now condenses all channels into a single category similar to the old Unreads category that existed in prior to the new sidebar.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27498

#### Screenshots
Screen Shot 2020-09-30 at 3 33 43 PM
Screen Shot 2020-09-30 at 3 33 51 PM


components/sidebar/sidebar_category_header.tsx
```diff
@@ -0,0 +1,56 @@
...
+    displayName: string;
+    dragHandleProps?: DraggableProvidedDragHandleProps;
+    isCollapsed?: boolean;
+    isCollapsible?: boolean;
+    isDragging?: boolean;
+    isDraggingOver?: boolean;
+    onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+}
+
+const SidebarCategoryHeader = React.forwardRef((props: Props, ref?: React.Ref<HTMLDivElement>) => {
```
hmhealey, line 21: I pulled this out of the `SidebarCategory` into a reusable component so that I didn't need to repeat the styles in the `UnreadChannels` component

components/sidebar/sidebar_category_header.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return (
+        <div className='SidebarChannelGroupHeader'>
+            <div
```
hmhealey, line 24: This has changed from a button to a clickable div. I hoped to switch between the two components based on whether or not the header should be clickable, but I ran into typing issues because of the ref. This seems to work well enough though since the header can still be tabbed through as before.

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -128,40 +140,63 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
...
+                isCollapsed={this.isCollapsed(this.props)}
+                channel={channel}
+                currentTeamName={currentTeamName}
+            />
         );
+
+        let wrappedComponent: React.ReactNode;
+
+        if (isDraggable) {
+            wrappedComponent = (
```
hmhealey, line 154: It might be nice if we had some sort of `ConditionalDraggable` component that could toggle this behaviour via a prop, but I wasn't able to do that at this time.

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -410,7 +409,44 @@ export default class SidebarCategoryList extends React.PureComponent<Props, Stat
 
     render() {
         const {categories} = this.props;
-        const renderedCategories = categories.map(this.renderCategory);
+
+        let channelList: React.ReactNode;
+        if (this.props.isUnreadFilterEnabled) {
```
hmhealey, line 414: Swapping `UnreadChannels` for a set of `ChannelCategory` components gives us things like the `More Unreads` indicators for free as long as we still set the refs for each channel correctly.

components/sidebar/unread_channels/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import {getUnreadChannels} from 'selectors/views/channel_sidebar';
+
+import {GlobalState} from 'types/store';
+
+import UnreadChannels from './unread_channels';
+
+function mapStateToProps(state: GlobalState) {
+    return {
+        unreadChannels: getUnreadChannels(state),
```
hmhealey, line 14: This could theoretically be passed from the `SidebarChannelList`, but I decided to call the selector again for the sake of clarity.

sass/layout/_sidebar-left.scss
```diff
@@ -599,26 +599,24 @@
         .SidebarMenu:not(.menuOpen) {
             display: none;
         }
+    }
 
-        &.draggingOverHeader:hover {
-            //background: radial-gradient(60.83% 3421.87% at 50% 50%, rgba(255, 255, 255, 0.4) 0%, rgba(255, 255, 255, 0) 100%);
-            box-shadow: inset -1px 0 0 rgba(var(--sidebar-text-rgb), 0.60), inset 0 -1px 0 rgba(var(--sidebar-text-rgb), 0.6), inset 1px 0 0 rgba(var(--sidebar-text-rgb), 0.6), inset 0 1px 0 rgba(var(--sidebar-text-rgb), 0.6);
-            border-radius: 4px;
+    .SidebarChannelGroup:not(.dmCategory) .draggingOver .SidebarChannelGroupHeader:hover {
```
hmhealey, line 604: The previous code applied the `draggingOverHeader` class only sometimes depending on the type of the category. Since the UNREAD CHANNELS header doesn't have a category, I chose to always apply the class but add an additional class to the category to identify the DMs and let the CSS determine what styles DMs need vs other categories.

selectors/views/channel_sidebar.ts
```diff
@@ -76,6 +74,67 @@ export function makeGetCurrentlyDisplayedChannelsForTeam(): (state: GlobalState,
             }).flat();
         },
     );
+})();
+
+// getUnreadChannels returns an array of all unread channels on the current team for display with the unread filter
+// enabled. Channels are sorted by recency with channels containing a mention grouped first.
+export const getUnreadChannels = (() => {
```
hmhealey, line 81: Some of this logic could have been reused from the old sidebar, but that has a lot of cruft in it, so I chose to rewrite what I could. In particular, the sorting logic is copied from the old sidebar, but this one has a lot fewer dependencies than that one does, and it doesn't unnecessarily rewrite channel objects for no reason.

selectors/views/channel_sidebar.ts
```diff
@@ -1,60 +1,61 @@
...
 }
 
 export function isUnreadFilterEnabled(state: GlobalState) {
     return state.views.channelSidebar.unreadFilterEnabled;
 }
 
-export function makeGetCollapsedStateForAllCategoriesByTeam(): (state: GlobalState, teamId: string) => Record<string, boolean> {
+// getChannelsInCategoryOrder returns an array of channels on the current team that are currently visible in the sidebar.
+// Channels are returned in the same order as in the sidebar.
+export const getChannelsInCategoryOrder = (() => {
```
hmhealey, line 35: I realize that previously passing in the team ID as a parameter might have been good for letting us reuse selectors, but it also made them much more complicated since you had to worry about making a selector factory to avoid mucking up memoization.

esethna (approved):
@hmhealey looks good, only tweak would be the text should say "All Unreads" when the filter is enabled

image

devinbinnie (approved):

hmhealey:
As I just found, there's a potential bug here about channels moving when you click on them and they become current. I need to go see how the old Unreads category behaved since it will likely be something similar

esethna:
@andrewbrown00 @hmhealey how are muted channels with unreads or mentions expected to sort here?

hmhealey:
I believe I made them sort like the Unread category previously did where muted channels are treated the same as unmentioned ones for sorting

esethna:
@andrewbrown00 thoughts on this? I'm wondering if we want to surface muted channels at all in the unread filter? Or alternatively at the bottom of the list of unreads when the filter is enabled. Thoughts?

andrewbrown00:
> @andrewbrown00 thoughts on this? I'm wondering if we want to surface muted channels at all in the unread filter? Or alternatively at the bottom of the list of unreads when the filter is enabled. Thoughts?

@esethna @hmhealey let's move muted channels with user mentions/direct messages below all unmuted channels/direct messages when the filter is enabled.

hmhealey:
@andrewbrown00 @esethna I've updated it to sort muted channel mentions beneath other unread channels now, and I've fixed the bug I found in R&D where switching to a channel with a mention while the unread filter was enabled would cause it to move in the RHS.
Screen Shot 2020-10-07 at 5 42 41 PM

@deanwhillier @devinbinnie  I've fixed the issue where we weren't using semantic markup for the header button. I've also made some other changes as mentioned above, so it's ready for a re-review now.

types/store/index.ts
```diff
@@ -47,7 +47,7 @@ export type GlobalState = BaseGlobalState & {
             };
             focusedPostId: string;
             mobileView: boolean;
-            keepChannelIdAsUnread: Channel | null;
+            lastUnreadChannel: (Channel & {hadMentions: boolean}) | null; // Actually only an object with {id: string, hadMentions: boolean}
```
hmhealey, line 50: I forget why the name of this field in the store was different than what we called it literally everywhere else, so I renamed it.

The ugly type is because I didn't want to go change this all throughout the channel selectors in mattermost-redux where it's used. We can do that cleanup once the old sidebar is removed and most of those selectors are removed.

selectors/views/channel_sidebar.ts
```diff
@@ -76,6 +74,84 @@ export function makeGetCurrentlyDisplayedChannelsForTeam(): (state: GlobalState,
...
+        getMyChannelMemberships,
+        getLastPostPerChannel,
+        (state: GlobalState) => state.views.channel.lastUnreadChannel,
+        (channels, myMembers, lastPosts, lastUnreadChannel) => {
+            function isMuted(channel: Channel) {
+                return isChannelMuted(myMembers[channel.id]);
+            }
+
+            function hasMentions(channel: Channel) {
+                if (lastUnreadChannel && channel.id === lastUnreadChannel.id && lastUnreadChannel.hadMentions) {
```
hmhealey, line 108: This `lastUnreadChannel.hadMentions` logic keeps track of the current channel and whether or not it had mentions when we switched to it, and we can use that to sort the current channel as if it did have mentions. It was part of a previous attempt for this exact same thing with the Unreads section of the old sidebar (thanks, past me!), but it mysteriously isn't used in the current iteration of that code, so it took some time to find.

esethna (approved):

andrewbrown00 (approved):
Looks good @hmhealey thanks!

image

amyblais:
@devinbinnie @deanwhillier Kind reminder to review this PR Eric wanted to aim this for v5.29.

deanwhillier:
Thanks @amyblais. Was talking with Harrison about this earlier today. There is one more update needed based on our conversation and then I'll do my final.

hmhealey:
@devinbinnie @deanwhillier This should be good to review now

devinbinnie (approved):

deanwhillier (commented):
Quick comment: looks like the unread header needs a bit of CSS to compensate for the space the expand/collapse arrow was taking up in the expandable header.

image

components/sidebar/sidebar_category_header.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const SidebarCategoryHeader = React.forwardRef((props: Props, ref?: React.Ref<HTMLDivElement>) => {
+    return (
+        <div className='SidebarChannelGroupHeader'>
+            <div
+                ref={ref}
+                className={classNames('SidebarChannelGroupHeader_groupButton', {
+                    dragging: props.isDragging,
+                })}
+                aria-label={props.displayName}
```
deanwhillier, line 26: This isn't needed on the static header.

components/sidebar/sidebar_category_header.tsx
```diff
@@ -0,0 +1,78 @@
...
+            >
+                <div {...props.dragHandleProps}>
+                    {wrapEmojis(props.displayName)}
+                </div>
```
deanwhillier, line 30: Do we need the `dragHandleProps` code here given it's a static header?

components/sidebar/sidebar_category_header.tsx
```diff
@@ -0,0 +1,78 @@
...
+    isDraggingOver?: boolean;
+}
+
+export const SidebarCategoryHeader = React.forwardRef((props: Props, ref?: React.Ref<HTMLDivElement>) => {
+    return (
+        <div className='SidebarChannelGroupHeader'>
+            <div
+                ref={ref}
+                className={classNames('SidebarChannelGroupHeader_groupButton', {
+                    dragging: props.isDragging,
```
deanwhillier, line 24: Do we need the `dragging` class code here given it's a static header?

sass/layout/_sidebar-left.scss
```diff
@@ -599,26 +599,24 @@
...
+
     .SidebarChannelGroup .SidebarChannelGroupHeader_groupButton {
         height: 32px;
         padding: 0;
         font-size: 12px;
         font-family: 'Open Sans', sans-serif;
         font-weight: 600;
         text-transform: uppercase;
-        overflow: hidden;
         text-overflow: ellipsis;
```
deanwhillier, line 620: If you are removing `overflow: hidden;` here, the corresponding `text-overflow: ellipsis` style will have no effect anymore so you should probably remove it as well. Neither appears to be necessary for truncating text here anyways as the child div css appears to be handling it.

```suggestion
```

deanwhillier (changes requested):
A few items to consider below.

Also, I noticed that the 'NEW' label seems to have decided it doesn't want to align properly anymore 😛:

image

components/sidebar/sidebar_category_header.tsx
hmhealey: Oh static is a much better name for it. I would've removed those both from the static one, but I couldn't come up with a better name than something like `SidebarCategoryHeaderNotCollapsibleOrDraggable` 😛 . Static vs not static is much better though since it also lets the collapsible/draggable one be the primary one

deanwhillier (approved):
👍

esethna:
@jgilliam17 this one is also a priority for v5.29 :)

jgilliam17:
@hmhealey Can you please clarify if Muted unread channels are also supposed to surface on the LHS when filter for All Unreads is enabled, or only Muted channels with mentions? Thanks

hmhealey:
Muted channels are only marked as unread when they have a mention, so you can't actually have a muted channel without a mention.

esethna:
@jgilliam17 just muted channels with mentions should appear when filtered.

e2e/cypress/integration/channel_sidebar/unread_filter_spec.js
```diff
@@ -0,0 +1,170 @@
...
+                ExperimentalChannelSidebarOrganization: 'default_on',
+            },
+        });
+
+        cy.apiInitSetup({loginAfter: true}).then(() => {
+            cy.visit('/');
+        });
+    });
+
+    it('should change the filter label when the unread filter changes state', () => {
```
jgilliam17, line 30:
```suggestion
    it('MM-T3441 should change the filter label when the unread filter changes state', () => {
```

e2e/cypress/integration/channel_sidebar/unread_filter_spec.js
```diff
@@ -0,0 +1,170 @@
...
+        cy.get('.SidebarFilters:contains(Unread)').should('be.visible');
+
+        disableUnreadFilter();
+
+        // * Verify that the unread filter is back in all channels state
+        cy.get('.SidebarFilters:contains(VIEWING:)').should('be.visible');
+        cy.get('.SidebarFilters:contains(All channels)').should('be.visible');
+    });
+
+    it('should not persist the state of the unread filter on reload', () => {
```
jgilliam17, line 48:
```suggestion
    it('MM-T3442 should not persist the state of the unread filter on reload', () => {
```

e2e/cypress/integration/channel_sidebar/unread_filter_spec.js
```diff
@@ -0,0 +1,170 @@
...
+
+        // # Reload the page
+        cy.reload();
+
+        // * Verify that all categories are visible again
+        cy.get('.SidebarChannelGroupHeader:contains(CHANNELS)').should('be.visible');
+        cy.get('.SidebarChannelGroupHeader:contains(DIRECT MESSAGES)').should('be.visible');
+    });
+
+    it('should only show unread channels with filter enabled', () => {
```
jgilliam17, line 63:
```suggestion
    it('MM-T3443 should only show unread channels with filter enabled', () => {
```

e2e/cypress/integration/channel_sidebar/unread_filter_spec.js
```diff
@@ -0,0 +1,170 @@
...
+        // * Verify that the unread channel is still visible
+        cy.get(`#sidebarItem_${unreadChannelName}`).should('be.visible').should(beUnread);
+
+        disableUnreadFilter();
+
+        // * Verify that the read channel has reappeared
+        cy.get(`#sidebarItem_${readChannelName}`).should('be.visible').should(beRead);
+    });
+
+    it('should always show the current channel, even if it is not unread', () => {
```
jgilliam17, line 93:
```suggestion
    it('MM-T3444 should always show the current channel, even if it is not unread', () => {
```

e2e/cypress/integration/channel_sidebar/unread_filter_spec.js
```diff
@@ -0,0 +1,170 @@
...
+
+        enableUnreadFilter();
+
+        // * Verify that the Town Square is still visible
+        cy.get('#sidebarItem_town-square').should('be.visible').should(beRead);
+
+        disableUnreadFilter();
+    });
+
+    it('should hide channels once they have been read', () => {
```
jgilliam17, line 112:
```suggestion
    it('MM-T3445 should hide channels once they have been read', () => {
```

jgilliam17 (commented):
Thank you @hmhealey 
Tested, looks good to merge.
-Verified, when enabled, new filter by unread now compress all unread channels into a single category All Unreads
- Verified Unread filter disables on refresh
- Verified muted unread channels with mentions are sorted at the bottom of All Unreads
- Than you for adding E2Es 🎉 I added TM4J test cases and updated with steps, details and links. Could you add one more for muted unread channel with mention?

PR #6603, opened 2020-10-01
[MM-29183] - fix(post_info): add id prop to ToolTip

Phizzard:
## Summary

Fixes console error raising accessibility concerns with React-Bootstrap's <ToolTip /> component missing a value from it's Id prop.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15690


components/post_view/post_info/post_info.jsx
```diff
@@ -307,7 +307,7 @@ export default class PostInfo extends React.PureComponent {
                     delayShow={Constants.OVERLAY_TIME_DELAY}
                     placement='top'
                     overlay={
-                        <Tooltip>
+                        <Tooltip id='additional-information-overlay'>
```
hanzei, line 310: What happens if there are multiple posts with a tooltip?

hanzei (changes requested):
Thanks for the contribution:+1: One question below

components/post_view/post_info/post_info.jsx
Phizzard: Good question! I believe the `id` prop does get used as an `id` HTML attribute, which needs to be unique.

Should we add a new required prop to the `PostInfo` component for this value?

hanzei: Could we generate a unique id using e.g. the post id?

iomodo (approved):
LGTM! Thanks for the contribution @Phizzard

hanzei (approved):

components/post_view/post_info/post_info.jsx
```diff
@@ -307,7 +307,7 @@ export default class PostInfo extends React.PureComponent {
                     delayShow={Constants.OVERLAY_TIME_DELAY}
                     placement='top'
                     overlay={
-                        <Tooltip>
+                        <Tooltip id={`additional-information-overlay-${post.id}`}>
```
saturninoabril, line 310: Could we change to
```suggestion
                        <Tooltip id={`viewAdditionalInfo-${post.id}`}>
```

saturninoabril (changes requested):
Thanks @Phizzard! Left a comment and I'm all good.

saturninoabril (approved):
Thanks @Phizzard for updating. LTGM

PR #6608, opened 2020-10-02
[MM-20474] Migrate 'components/post_view/commented_on' module and associated tests to TypeScript

didithilmy:
#### Summary

Migrates `components/post_view/commented_on` modules to TypeScript.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15467

#### Related Pull Requests

None

#### Screenshots

None


components/post_view/commented_on/index.ts
```diff
@@ -29,9 +43,9 @@ function mapStateToProps(state, ownProps) {
     };
 }
 
-function mapDispatchToProps(dispatch) {
+function mapDispatchToProps(dispatch: Dispatch<GenericAction>) {
     return {
-        actions: bindActionCreators({
+        actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc>, Actions>({
             showSearchResults,
             updateSearchTerms,
```
didithilmy, line 50: Hi @devinbinnie, I need help to type this one, `updateSearchTerms` does not have the signature of a typical action so this line results in a type error. I'm not sure how to type it.

Btw, I don't see any of the actions being used or propagated to the children. Is it safe to just remove it?

devinbinnie: Huh, weird. Seems like we would have removed those if they weren't used, but you're right there's no reference to them as far as I can see. Feel free to remove them :)

Is there any other assistance you need now that those can be removed?

didithilmy: No thanks, shouldn't be a problem.

devinbinnie (approved):
LGTM! Thanks @didithilmy!

srkgupta (approved):
Did some sanity testing and there were no issues found. Detailed testing will be done with the daily Cypress E2E tests. Approving the PR.

sudheerDev (approved):
LGTM, Thanks @didithilmy

PR #6611, opened 2020-10-02
Keyboard shortcut to go directly to the search bar with the in: channel search modifier pre-filled

MathewtheCoder:
#### Summary
Keyboard shortcut CMD/Ctrl + Shift + F, pre-fills the search box in channel header with in: {current channel name}

#### Ticket Link
Fixes: https://github.com/mattermost/mattermost-server/issues/15639
JIRA: https://mattermost.atlassian.net/browse/MM-19327

#### Screenshots
Town-Square---Sample-Mattermost


esethna (changes requested):
Thanks @MathewtheCoder! Few tweaks:

1) Let's do this for group or direct messages as well
2) On windows it seems it's using the Windows key rather than CTRL to execute this behaviour
3)  On the desktop app, we want to override the current CTRL/CMD + F to have this same behaviour. Currently CTRL/CMD + F opens the browser type find function. I'm not sure if this will require changes on the desktop app repository or not. @Willyfrog would you know?

MathewtheCoder:
@esethna 

> 1. Let's do this for group or direct messages as well

Can you please clarify what exactly should be done for group and direct messages ?

> 2. On windows it seems it's using the Windows key rather than CTRL to execute this behaviour ?

It's an issue with the hot key check. I was only able to test it in Mac only. Will fix it.

> 3. On the desktop app, we want CTRL/CMD + F to have this same behaviour. Currently CTRL/CMD + F opens the browser type find function on the desktop app. I'm not sure if this will require changes on the desktop app repository or not.

Do we have a separate repo for the electron app if not how can i test the desktop version from react codebase ?

Willyfrog:
> 3)  On the desktop app, we want CTRL/CMD + F to have this same behaviour. Currently CTRL/CMD + F opens the browser type find function on the desktop app. I'm not sure if this will require changes on the desktop app repository or not. @Willyfrog would you know? 

Yes, it is mapped to use chromium's search. That means older desktop versions will shadow the shortcut

esethna:
@MathewtheCoder 

1. It should add the same "in:" formatting for the current direct message or group message the user is in when they hit the shortcut:

image

2. Thanks

3. @Willyfrog what changes are needed on desktop to change the shortcut behaviour to what's described in the ticket? @MathewtheCoder are you able to make the change for the desktop app as well?

Willyfrog:
just removing the entry from the menu would be enough to prevent that:
https://github.com/mattermost/desktop/blob/master/src/main/menus/app.js#L122

also associated code if we are not going to use the in-browser search:
https://github.com/mattermost/desktop/blob/a30233f9139f803f5cfcd5121abf2ed63a8951d8/src/browser/components/MainPage.jsx#L328

esethna:
Thanks Guillermo, @MathewtheCoder let us know if you have any questions on the above ^ That part will have to be submitted to the desktop repo: https://github.com/mattermost/desktop/

MathewtheCoder:
@esethna I will try to implement from the above comments but is there any way to seed fake profiles so i can create user groups and test the user group scenario. Currently i only have me as the user.

Willyfrog:
> @esethna I will try to implement from the above comments but is there any way to seed fake profiles so i can create user groups and test the user group scenario. Currently i only have me as the user.

usually `make test-data` generates a ton of users and channels, not sure about GMs

esethna:
@MathewtheCoder  you can also use `/test users 10` will create 10 fake profiles ^

esethna:
@MathewtheCoder how's this going?

MathewtheCoder:
@esethna I had fixed the hotkey issue and added the logic to profile the search input the user name when hotkey is triggered in user or groups. But the search suggestions were only working when i used the **from** prefix for users and groups not **in** so i have currently made it profile from + {current username/groupname}.

esethna:
@MathewtheCoder the in: modifier should work for users and groups
image

esethna:
@MathewtheCoder are you still working on this issue? Are you planning on making the changes to desktop app as well as usggested by Guillermo above?

MathewtheCoder:
@esethna I am still working on it. Will fix the @in for groups and users

esethna:
@MathewtheCoder please let us know if you need any help or guidance with this?

MathewtheCoder:
Replicated the logic into functional component and its getting pre-filled correctly. But only issue is after filling the focus is lost from the search bar so i am looking into it.

esethna:
@MathewtheCoder let us know if you need any help here

MathewtheCoder:
@esethna Yes i am facing some issue with the focus on the input field it seems the event call back is called twice and i am getting two suggestion box on CMD+Shift + F. I am looking into that.

hmhealey:
I hope it's not still the case, but I know in the past that we had two copies of the search box components. There's a chance that that's why you're having those called twice

MathewtheCoder:
@hmhealey @esethna I fixed the issue of event callback calling twice but i am still seeing two search boxes. I looked at the html code due to what @hmhealey said i found two instances of the search hint there. Should I remove second one or is there any use for it ?

esethna:
@hmhealey or @nevyangelova can you provide some guidance here? This PR is potentially blocking another PR that's ready to go in: https://github.com/mattermost/mattermost-webapp/pull/6896

hmhealey:
I don't know if it's still the case, but we at least used to have two separate search boxes, one that was part of the ChannelHeader and another that is part of the RHS. When the RHS is closed, the user would type in the ChannelHeader one, and then when they go to search, it would be swapped for the one in the RHS. That might be the two search boxes you're seeing now.

I tried this out on the test server, and while it fills in the search box, it doesn't switch focus correctly unless the RHS is already open. When that happens, the search hint is in a weird spot, so it might be trying to switch to the wrong search box.
Screen Shot 2021-01-26 at 2 31 26 PM
I did manage to get it in a state where it doesn't fill in the search box as well when using it multiple times in a row, but I don't know if I can reproduce it consistently.

hmhealey:
@MathewtheCoder Is the information I posted helpful? I know the search box can be a bit weird to work with at times.

TheDarkestDay:
Hey, guys. I have been doing some research about this double suggestion box issue. Can't say that I've found a lot, but at least I can work through it one way or another. 

So, this is right - we always have exactly two instances of `Search` component in the app markup. First one belongs to `ChannelHeader` component (and I assume this is the one we would like to enhance with all this shortcut stuff) and yet another one for `SidebarRight`. I haven't managed yet to discover reasoning behind existence of the latter one but I believe there should be something.

Both of these components use `SearchBar` component with this logic for handling shortcut. So, that is why we have this kind of conflict between two search bars fighting for keyboard events. Therefore, I would like to ask the following questions:

1. What is the actual purpose of this second `SearchBar` inside `SidebarRight`? Do we need to facilitate the same feature for it as well?
2. If not, can we probably introduce some sort of simple flag or introduce some other way to distinguish these two search bars, to let only required one capture and react to keyboard events?

Not sure whom to ping here. Probably @MathewtheCoder @hmhealey @esethna

esethna:
@nevyangelova might have some feedback or insight here ^

nevyangelova:
> Hey, guys. I have been doing some research about this double suggestion box issue. Can't say that I've found a lot, but at least I can work through it one way or another.
> 
> So, this is right - we always have exactly two instances of `Search` component in the app markup. First one belongs to `ChannelHeader` component (and I assume this is the one we would like to enhance with all this shortcut stuff) and yet another one for `SidebarRight`. I haven't managed yet to discover reasoning behind existence of the latter one but I believe there should be something.
> 
> Both of these components use `SearchBar` component with this logic for handling shortcut. So, that is why we have this kind of conflict between two search bars fighting for keyboard events. Therefore, I would like to ask the following questions:
> 
> 1. What is the actual purpose of this second `SearchBar` inside `SidebarRight`? Do we need to facilitate the same feature for it as well?
> 2. If not, can we probably introduce some sort of simple flag or introduce some other way to distinguish these two search bars, to let only required one capture and react to keyboard events?
> 
> Not sure whom to ping here. Probably @MathewtheCoder @hmhealey @esethna

Hey @TheDarkestDay that is indeed an issue we are dealing with but mostly because of the way SuggestionBox handles passing around refs. The reason why we are rendering another "Search" component was to reduce some of the two-directional data flow that was happening. I don't know what you mean by a second "SearchBar" though? The searching logic is needed to be replicated in the sidebar for tablet and mobile screens (if you shrink your resolution you will see the search icon pop up and will be able to access the search in RHS). Ideally we would have the same feature for both components. Hope that helps!

hmhealey:
I thought we had removed the second `SearchBar`, but I guess it still exists for mobile view (shown when the window is reduced to below a certain width). If you look at them in the DOM, one has the id `searchBar` and the other has the ID `sbrSearchBar`.
Screen Shot 2021-02-24 at 5 00 39 PM
Screen Shot 2021-02-24 at 5 01 02 PM
There's only ever one visible at once, so perhaps there's something we can do to detect which one is visible and only focus that one? I'm not entirely sure if that's possible since it's just hidden by CSS, but that might be a possibility.

esethna:
@MathewtheCoder let us know if you're still working on this? If not, perhaps @asaadmahmood can help finish it up and resolve the issues?

esethna:
@MathewtheCoder @TheDarkestDay just a note that @devinbinnie is going to take this over so closing this PR. Thanks for the head start here :)

PR #6631, opened 2020-10-05
[MM-20478] Migrate post_header module to TypeScript

abdusabri:
#### Summary

Migrates `post_header` component to TypeScipt.

The component didn't have its own tests.

Please refer to the thread for why `post_info` connected props were made optional.

https://community-daily.mattermost.com/core/pl/emepbkuprj8y3nw1x8oxsxe4ya

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/15463, https://mattermost.atlassian.net/browse/MM-20478


components/post_view/post_header/post_header.tsx
```diff
@@ -1,96 +1,97 @@
...
+
+    /**
+     * Set to mark post as being hovered over
+     */
+    hover: boolean;
+
+    /*
+    * Set to render the post time when not hovering
+    */
+    showTimeWithoutHover: boolean;
```
abdusabri, line 61: This was originally optional. However, I've traced it down in the whole project, and found that it will always have a Boolean value. So, I think making it required is better than checking for `undefined` in the code (if left optional) or using `as Boolean`

nevyangelova (approved):

components/post_view/post_header/post_header.tsx
nevyangelova: @abdusabri thanks! I agree that if it always has a value it will be best to have it as required.

calebroseland (approved):

jgilliam17 (approved):
Thank you @abdusabri 
Tested, looks good to merge.
- Verified post header and post menu options - no change after migration.

PR #6645, opened 2020-10-05
[MM-20470] Migrate emoji page and any team permission gate to typescript #15470

vanya829:
#### Summary

This PR migrate to typescript next components:

- emoji page (was initially in the scope of the ticket)
- any team permission gate (caused ts errors on the emoji page)

#### Related Pull Requests


jasonblais:
Ticket link: https://github.com/mattermost/mattermost-server/issues/15470

components/permissions_gates/any_team_permission_gate/any_team_permission_gate.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Permissions enough to pass the gate (binary OR)
+     */
+    permissions: string[],
+
+    /**
+     * Has permission
+     */
+    hasPermission?: boolean,
```
devinbinnie, line 16: This prop is marked as `isRequired` in the original file. Is there a reason it was changed to optional here?

components/permissions_gates/any_team_permission_gate/any_team_permission_gate.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Invert the permission (used for else)
+     */
+    invert?: boolean,
```
devinbinnie, line 21: Same with this one.

devinbinnie (commented):
Thanks for your contribution @vanya829!
I've left a couple comments for your review.

components/permissions_gates/any_team_permission_gate/any_team_permission_gate.tsx
vanya829: This one is optional because it has a default in the original component. The default makes no sense if the prop is required.

components/permissions_gates/any_team_permission_gate/any_team_permission_gate.tsx
vanya829: The reason is that ts complains about this prop because it's added via `connect` in `/any_team_permission_gate_index`. I tried a few ways of typing it but didn't manage to fix this issue. Maybe you know how to deal with that?

devinbinnie: Where is it complaining? Usually the TS compiler is able to detect props added by `redux`.
Can you post the error you're getting?

vanya829: image
Here is an example of `npm run check-types` with `hasPermission` required

devinbinnie: Ahh alright, so the test files do not import the redux-connected component. We expect to mock out those properties ourselves.

So in the case of this test, if `hasPermission` is missing it would evaluate to `false` within the component. Meaning you'd want to add `hasPermission={false}` to that particular test.

vanya829: image
It's not only the test problem. Here is an example of another component with an error, and you can see, that it imports the connected component. Will try to find the root cause of the issue.

sudheerDev: I hit the same issue in another PR, in that case, I thought it is because the parent component is in js 🤷

sudheerDev (approved):
LGTM Thank you @vanya829

devinbinnie:
@vanya829 Can you resolve the conflicts on the PR? After that I can take another look.

components/permissions_gates/any_team_permission_gate/any_team_permission_gate.tsx
vanya829: @devinbinnie @sudheerDev fixed this issue by setting explicit types in the `connect`

vanya829: `any_team_permission_gate.tsx` was converted to `tsx` in other MR, so I reverted my changes from those files

vanya829:
@devinbinnie conflicts resolved

devinbinnie (approved):
LGTM! Thanks @vanya829!

ogi-m (approved):
Smoke tested the custom emoji page, no issues 👍

PR #6666, opened 2020-10-06
Migrate system notice to typescript

vanya829:
#### Summary

Moved system notice component to Typescript.
Not all props we inline with their definitions, so I was forced to use `any` definition it a few places:

- `show` method in notice type optional, because it missing in one of the notice lists in the tests
- `config` param inside `show` method should be `Partial<ClientConfig>` based on selectors' definition, but it expects `InstallationDate` that is not a part of this type. That is why it's `any`
- `license` param inside `show` method should be `ClientLicense`  based on selectors' definition, but it expects `IsLicensed` that is not a part of this type. That is why it's `any`
- `dismissedNotices` prop in `SystemNotice` component should be `boolean` based on selectors definition, but in component, it's used as an object with boolean props.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/13477


hahmadia (approved):
I looked and thought about the doubts you've raised and I honestly think it's okay to leave it as it is. Thank you for bringing them up and after thinking over them, I feel that what you've done is okay and there is no need for any changes. Thanks again!

gabrieljackson (approved):
Looks great! Thanks for the contribution.

srkgupta:
Hello @gabrieljackson or @hahmadia 
Can you let me know which screens are affected because of this change so that it can be tested accordingly.

hanzei:
@srkgupta This PR affects the System Notice shown on the top. I contains messages like https://github.com/vanya829/mattermost-webapp/blob/e8fd6355be72eba359829244c1d67df5e9bf0ffd/components/system_notice/notices.tsx#L93.

srkgupta (approved):
Thanks @vanya829 and @hanzei. Did some basic testing on the webapp and there were no issues/JS errors shown. Approving the PR.

PR #6669, opened 2020-10-07
[MM-20494] Migrate 'components/post_view/post_attachment_opengraph' module and associated tests to TypeScript

KuSh:
#### Summary
Migrate 'components/post_view/post_attachment_opengraph' module and associated tests to TypeScript

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15447
JIRA: https://mattermost.atlassian.net/browse/MM-20494


KuSh:
I choose to simplify getNearestPoint() function as extra parameters configuring used attributes were only used differently by tests

KuSh:
I added the necessary definitions in mattermost-redux : https://github.com/mattermost/mattermost-redux/pull/1252

components/post_view/post_attachment_opengraph/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    const config = getConfig(state);
+
+    return {
+        currentUserId: getCurrentUserId(state),
+        enableLinkPreviews: config.EnableLinkPreviews === 'true',
+        openGraphData: getOpenGraphMetadataForUrl(
+            state,
+            ownProps.postId,
+            ownProps.link,
+        ) as OpenGraphMetadata,
```
hmhealey, line 29: I assume this can be changed once mattermost-redux is updated?

components/post_view/post_attachment_opengraph/post_attachment_opengraph.tsx
```diff
@@ -18,52 +22,23 @@ const DIMENSIONS_NEAREST_POINT_IMAGE = {
...
-        actions: PropTypes.shape({
-            editPost: PropTypes.func.isRequired,
-        }).isRequired,
-    }
+export type Props = {
+    postId: string;
+    link: string;
+    currentUserId?: string;
+    post: Post;
+    openGraphData?: OpenGraphMetadata | null;
```
hmhealey, line 30: Can this just be `openGraphData?: OpenGraphMetadata`? I'm not sure if the selector ever returns null, and the unit tests for this can be updated to pass `undefined` instead.

Up until recently, we've been using null for empty values, but we've since learned undefined is easier to use in most cases.

KuSh: Yeah sure, I'll do that !

components/post_view/post_attachment_opengraph/index.ts
KuSh: If I add every missing required property yes !

hmhealey (approved):
Thanks! It's still a bit weird for me to see undefined everywhere, but that definitely makes the type definitions look a lit nicer.

KuSh:
> Thanks! It's still a bit weird for me to see undefined everywhere, but that definitely makes the type definitions look a lit nicer.

Undefined is the new null ^^

nevyangelova (approved):

stevemudie (approved):
LGTM

hanzei:
Thanks for the great PR @KuSh :tada:

PR #6679, opened 2020-10-07
[MM-28213] Free Tier Overage Warning Panel

nickmisasi:
#### Summary
In the System Console subscription/billing section we want to give the user a heads up when they're close to the threshold of the user count in the free tier. This PR adds a banner when they are within 3 users of the limit. It should not appear when they are over the limit. The panel is permanently dismissable by each individual user (ie, one user dismissing does not affect another).

Keep in mind, this displays on a page that is not fully finished. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28213?filter=15333

#### Related Pull Requests
N/A

#### Screenshots
image


components/admin_console/billing/billing_subscriptions/index.ts
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
devinbinnie, line 1: You should be using hooks to connect `redux` to this component. Adding a connector to a functional component is an anti-pattern as far as I know.

I'm cool with using a connector but then we should convert the component to a class component, which is also fine.

package-lock.json
```diff
@@ -12559,11 +12559,15 @@
       "resolved": "https://registry.npmjs.org/full-icu/-/full-icu-1.3.1.tgz",
       "integrity": "sha512-VMtK//85QJomhk3cXOCksNwOYaw1KWnYTS37GYGgyf7A3ajdBoPGhaJuJWAH2S2kq8GZeXkdKn+3Mfmgy11cVw==",
       "dev": true,
+      "requires": {
```
devinbinnie, line 12562: We can revert these `package-lock.json` changes

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
```diff
@@ -1,24 +1,43 @@
...
 // See LICENSE.txt for license information.
 
-import React, {useState} from 'react';
-import {FormattedMessage} from 'react-intl';
+import React, {useState, useEffect} from 'react';
+import {FormattedMessage, useIntl} from 'react-intl';
 import {Tooltip} from 'react-bootstrap';
 
+import {PreferenceType} from 'mattermost-redux/types/preferences';
+
```
devinbinnie, line 9: Nit: extra spacing. We can group all the `mattermost-redux` imports together

devinbinnie (changes requested):

components/admin_console/billing/billing_subscriptions/index.ts
ali-farooq0: +1 to Devin's point. Recommend using hooks and getting rid of this connector component.

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
```diff
@@ -107,12 +129,40 @@ const BillingSubscriptions: React.FC<Props> = () => {
...
+        if (!props.analytics) {
+            (async function getStandardAnalytics() {
+                await props.actions.getStandardAnalytics();
+            }());
+        }
+    }, []);
+
+    const shouldShowInfoBanner = (): boolean => {
+        const {analytics, userLimit, isCloud, preferences} = props;
+        if (!analytics || !isCloud || !userLimit || preferences.some((pref) => pref.name === CloudBanners.HIDE && pref.value === 'true')) {
```
ali-farooq0, line 146: do we even need this check? since by default this function returns `false`. maybe this only needs the check that returns `true` and return `false` for everything else? would simplify the code here.

matthewbirtch (approved):
Looks good to me!

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
nickmisasi: I'm not sure why VSCode always does this to me. Fixed.

stevemudie (approved):
LGTM

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
nickmisasi: Things like isCloud are probably not entirely necessary here, but the checks like !analytics, !userLimit, etc are needed because at this point in time they might not be loaded (see useEffect). I left the rest in because I think it's helpful to immediately see explicitly which reasons to not show the banner.

components/admin_console/billing/billing_subscriptions/index.ts
nickmisasi: I've converted it to using hooks, let me know what you think.

ali-farooq0 (approved):

components/admin_console/admin_definition.jsx
```diff
@@ -52,7 +52,7 @@ import CustomTermsOfServiceSettings from './custom_terms_of_service_settings';
 import SessionLengthSettings from './session_length_settings';
 import LDAPFeatureDiscovery from './feature_discovery/ldap.tsx';
 import SAMLFeatureDiscovery from './feature_discovery/saml.tsx';
-import BillingSubscriptions from './billing/billing_subscriptions';
+import BillingSubscriptions from './billing/billing_subscriptions.tsx';
```
devinbinnie, line 55: Shouldn't need the `.tsx` in the import here.

devinbinnie (approved):

PR #6680, opened 2020-10-07
[MM-29399] Remove an unused changeCSS() style

KuSh:
#### Summary

Remove an unused changeCSS() style

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15837
JIRA ticket: https://mattermost.atlassian.net/browse/MM-29399


deanwhillier:
QA review will be done when merging the branch into Master.

hmhealey (approved):
Good catch that that rule isn't actually used. It looks like it's being overwritten by the `.app__body .modal .settings-modal .nav-pills>li.active button:active` rule on like 539 of utils.jsx. The git blame looks like that's been the case for a few years now, so I think it's safe just to remove instead of potentially treating that as a bug

nevyangelova (approved):

PR #6709, opened 2020-10-08
[MM-28221] Payment Info Edit/View

devinbinnie:
#### Summary
Payment Information View.

We'll just do a developer review now and review UX/QA in a bigger PR later on.

NOTES:
- Wasn't able to fully test the credit card update, but the billing address was properly update so I'm assuming it works.
- Remove Payment Info isn't currently implemented, I'm not sure it's needed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28221

#### Screenshots
Screen Shot 2020-10-08 at 5 23 17 PM
Screen Shot 2020-10-08 at 5 23 32 PM


components/admin_console/billing/payment_info_edit.tsx
```diff
@@ -1,17 +1,77 @@
...
+import {GlobalState} from 'types/store';
+import {browserHistory} from 'utils/browser_history';
 
-type Props = {
+import './payment_info_edit.scss';
+import AlertBanner from 'components/alert_banner';
 
-};
+const STRIPE_CSS_SRC = 'https://fonts.googleapis.com/css?family=Open+Sans:400,400i,600,600i&display=swap';
+const STRIPE_PUBLIC_KEY = 'pk_test_ttEpW6dCHksKyfAFzh6MvgBj';
```
marianunez, line 27: Given that we need this in two places we should move it to the `payment_form/stripe.ts` file so we just need to update it one place there for production

marianunez (approved):
Thanks @devinbinnie! Just a comment below

gabrieljackson (approved):
👍

components/admin_console/billing/payment_info_edit.tsx
ali-farooq0: curious as to where are we getting this key from? I would refrain from using this key altogether and have it be an ENV variable. High risk if this is our key and we're sharing with the world.

ali-farooq0 (approved):

components/admin_console/billing/payment_info_edit.tsx
marianunez: This is not our key, is a public Stripe test key. Our key is not on the public repo and is replaced only for production build.

ali-farooq0: 👍

PR #6711, opened 2020-10-08
Remove Cloud billing flag from System Console

marianunez:
#### Summary

Removing Cloud billing flag from System Console **only on `cloud-ga` branch**. Leaving in `master` for now for development purposes.


ali-farooq0 (approved):

jwilander (approved):
These should make it into `master` too right?

marianunez:
> These should make it into `master` too right?

Yes, but keeping it out of `master` right now because it is being used for quicker development. There is still a couple of more features to complete. We can remove after that.

PR #6728, opened 2020-10-10
📦 added jest watch typeahead

paulussujono:
#### Summary
Added `jest-watch-typeahead` plugin to show which files are matched by the pattern when running jest in watch mode.
https://github.com/jest-community/jest-watch-typeahead

##### Before
image

##### After
image
image
image


saturninoabril (approved):
Awesome, LGTM, thanks @paulussujono!

ali-farooq0 (approved):
Nice work @paulussujono 🎉 !!

hmhealey:
Hi @paulussujono, thanks for the PR. I want to check with the team before approving this since we want to avoid adding too many dependencies that will later need to be maintained since we already have quite a lot of them. This one only affects development and won't affect end users at all, but we want to make sure we'll actually make use of it in case it slows down the development workflow at all. We should be meeting to discuss that tomorrow.

hmhealey:
We chatted about this, and some people hadn't actually seen Jest watch before, so they're looking forward to trying it out and using it with this plugin as well. Thanks again!

hmhealey (approved):

PR #6734, opened 2020-10-11
[GH-15891] Migrate changeCSS() to CSS variable in utils/utils.jsx, ln. 629

sudiptog81:
#### Summary

Replaced JavaScript variable `theme.centerChannelBg` with CSS variable `--center-channel-bg` for `.app__body .popover.bottom>.arrow:after`.

#### Ticket Link

Fixes [mattermost/mattermost-server#15891](https://github.com/mattermost/mattermost-server/issues/15891)

### DIFF
--- a/sass/components/_popover.scss
+++ b/sass/components/_popover.scss
@@ -75,6 +75,7 @@
 
             &:after {
                 border-color: transparent;
+                border-bottom-color: v(center-channel-bg);
             }
         }
     }
--- a/utils/utils.jsx
+++ b/utils/utils.jsx
@@ -626,7 +626,6 @@ export function applyTheme(theme) {
         changeCss('.app__body .date-separator .separator__text, .app__body .new-separator .separator__text', 'background:' + theme.centerChannelBg);
         changeCss('.app__body .search-help-popover .search-autocomplete__divider span', 'background:' + theme.centerChannelBg);
         changeCss('.app__body .dropdown-menu, .app__body .popover, .app__body .tip-overlay', 'background:' + theme.centerChannelBg);
-        changeCss('.app__body .popover.bottom>.arrow:after', 'border-bottom-color:' + theme.centerChannelBg);
         changeCss('.app__body .popover.right>.arrow:after, .app__body .tip-overlay.tip-overlay--sidebar .arrow, .app__body .tip-overlay.tip-overlay--header .arrow', 'border-right-color:' + theme.centerChannelBg);
         changeCss('.app__body .popover.left>.arrow:after', 'border-left-color:' + theme.centerChannelBg);
         changeCss('.app__body .popover.top>.arrow:after, .app__body .tip-overlay.tip-overlay--chat .arrow', 'border-top-color:' + theme.centerChannelBg);


nevyangelova (approved):

hahmadia (approved):

sass/components/_popover.scss
```diff
@@ -75,6 +75,7 @@
 
             &:after {
                 border-color: transparent;
+                border-bottom-color: v(center-channel-bg);
```
hahmadia, line 78: Sorry! I believe I approved a little pre-maturely. Are we able to change the `v` to `var` and add `--` to the front of the variable. Thanks. (This is copying the pattern done elsewhere. For example https://github.com/mattermost/mattermost-webapp/blob/master/components/generic_modal.scss#L106 )
```suggestion
                border-bottom-color: var(--center-channel-bg);
```

@hmhealey or @deanwhillier 
Can you guys please confirm the requested change is corrected. This is after reviewing the sample PR. I just want to double check or if I should approve this right away. Thank you

hahmadia (changes requested):
LGTM for the most part. A small change (from how I've seen it done in other parts of our code). Thanks!

sass/components/_popover.scss
nevyangelova: Nice catch @hahmadia I guess I also was in a spur of approvals

sudiptog81: I will make this change real quick! Should I also change this thing in other PRs?

hahmadia (approved):
LGTM! I guess it's not wrong cause I searched the codebase and found it being done that way in other areas. However, just going purely off of the help wanted ticket and the example PR attached, I feel it's best to follow that pattern set by the creators for the sake of consistency. 

Thanks for the quick change 👍

deanwhillier:
Thanks all! Yes, both approaches work. We are preferring the recommended approach though, so thanks for changing!

sass/components/_popover.scss
```diff
@@ -75,6 +75,7 @@
 
             &:after {
                 border-color: transparent;
+                border-bottom-color: var(--center-channel-bg);
```
deanwhillier, line 78: So, @sudiptog81, this one appears to be a bit of a trick and as a result, we will need to make a bit of a change to where you have placed this line.

Have you ever used the technique where you use thick borders (3 transparent, 1 coloured) on a zero-width/height element to create visual arrows? It's actually a fairly old technique (a may not be so common anymore) so no worries if not!

Basically, you set the entire border to fully transparent and then add a colour to the opposite border from the direction you want the arrow to point and then position it.

What that means for this PR is that you actually need to move your added line out of the the `bottom`/`right`/`top`/`left` shared selector and into one that is only `bottom` like this:

```scss
&.bottom {
    > .arrow::after {
         border-bottom-color: var(--center-channel-bg);
    }
}
```

There is already a solo `&.bottom` selector, so you can just add to that one.

All that said, it looks like the arrows have actually been permanently hidden so this line may not even be necessary. I need to look into that though, so maybe just go ahead and make this change and we can remove later as needed.

deanwhillier (changes requested):
Noticed something tricky below. 😛

Have a look and let me know if you have any questions!

deanwhillier:
QA for this PR will be batched when the campaign branch is merged to master.

sass/components/_popover.scss
sudiptog81: Sure, I'll go ahead and make the changes.

deanwhillier (approved):
LGTM, thanks so much @sudiptog81!

nevyangelova (approved):

PR #6736, opened 2020-10-11
[GH-15885] Migrate changeCSS() to CSS variable in utils/utils.jsx, ln. 613

tony-ho:
#### Summary

Remove changeCss() for `.nav-pills__unread-indicator svg`.
This is no longer required as CSS variable `--mention-color` is already used.

https://github.com/mattermost/mattermost-webapp/blob/c3f01ee92ca30d6bedf62aa93b6e117efd37f9a4/sass/layout/_sidebar-left.scss#L421

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15885

### DIFF
--- a/sass/layout/_legacy-sidebar-left.scss
+++ b/sass/layout/_legacy-sidebar-left.scss
@@ -181,6 +181,10 @@
             @include transition-delay(0s);
             @include transition-timing-function(ease-in, step-end);
         }
+
+        svg {
+            fill: var(--mention-color);
+        }
     }
 
     .nav-pills__unread-indicator-top {
--- a/utils/utils.jsx
+++ b/utils/utils.jsx
@@ -610,7 +610,6 @@ export function applyTheme(theme) {
     }
 
     if (theme.mentionColor) {
-        changeCss('.sidebar--left .nav-pills__unread-indicator svg', 'fill:' + theme.mentionColor);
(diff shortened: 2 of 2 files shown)


calebroseland:
Thanks for this, @tony-ho! You're right that its already in `_sidebar-left.scss`, but would you also mind adding it to `_legacy-sidebar-left.scss`? If you disable the new sidebar (Account Settings > Sidebar), it won't color the icon properly 
Screen Shot 2020-10-12 at 15 16 55

calebroseland (changes requested):
Thanks for this, @tony-ho! You're right that its already in `_sidebar-left.scss`, but would you also mind adding it to `_legacy-sidebar-left.scss`? If you disable the new sidebar (Account Settings > Sidebar), it won't color the icon properly 
Screen Shot 2020-10-12 at 15 16 55

tony-ho:
Thanks @calebroseland, good pickup. I've added the `fill` property in `_sidebar-left.scss`.

deanwhillier:
QA for this PR will be done when the campaign branch is merged to master.

hmhealey (approved):
Thanks for helping with this! I'm looking forward to when we can get rid of the old sidebar so that we won't need to repeat these styles between both of them

PR #6738, opened 2020-10-11
MM-20492 Migrate 'components/post_view/post_message_view' module and …

pikami:
#### Summary
Migrate 'components/post_view/post_message_view' module and associated tests to TypeScript

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15449


components/post_view/post_message_view/post_message_view.test.tsx
```diff
@@ -49,32 +49,35 @@ describe('components/post_view/PostAttachment', () => {
...
         expect(wrapper).toMatchSnapshot();
     });
 
     test('should match checkOverflow state on handleHeightReceived change', () => {
         const wrapper = shallow(<PostMessageView {...baseProps}/>);
-        const instance = wrapper.instance();
-        instance.checkOverflow = jest.fn();
+        const instance = wrapper.instance() as PostMessageView;
+
+        //instance.checkOverflow = jest.fn();
```
jespino, line 80: This is clearly not needed, looks like some obsolete code, so we can remove the comment also.

components/post_view/post_message_view/post_message_view.tsx
```diff
@@ -13,63 +13,35 @@ import PostMarkdown from 'components/post_markdown';
...
+type Props = {
+    post: Post; /* The post to render the message for */
+    enableFormatting?: boolean; /* Set to enable Markdown formatting */
+    options?: any; /* Options specific to text formatting */
+    compactDisplay?: boolean; /* Set to render post body compactly */
+    isRHS?: boolean; /* Flags if the post_message_view is for the RHS (Reply). */
+    isRHSOpen?: boolean; /* Whether or not the RHS is visible */
+    isRHSExpanded?: boolean; /* Whether or not the RHS is expanded */
+    theme: any; /* Logged in user's theme */
+    pluginPostTypes?: any; /* Post type components from plugins */
```
jespino, line 25: The type of tihs is already defined in `types/store/plugins.ts` in the mattermost-webapp repo. You can pick that same types and use it here, I think it would be `{[postType: string]: PostPluginComponent}`.

components/post_view/post_message_view/post_message_view.tsx
```diff
@@ -13,63 +13,35 @@ import PostMarkdown from 'components/post_markdown';
...
-    };
+type Props = {
+    post: Post; /* The post to render the message for */
+    enableFormatting?: boolean; /* Set to enable Markdown formatting */
+    options?: any; /* Options specific to text formatting */
+    compactDisplay?: boolean; /* Set to render post body compactly */
+    isRHS?: boolean; /* Flags if the post_message_view is for the RHS (Reply). */
+    isRHSOpen?: boolean; /* Whether or not the RHS is visible */
+    isRHSExpanded?: boolean; /* Whether or not the RHS is expanded */
+    theme: any; /* Logged in user's theme */
```
jespino, line 24: The theme itself is defined in mattermost-redux `types/preferences.ts` you can import that and use it here.

components/post_view/post_message_view/post_message_view.tsx
```diff
@@ -13,63 +13,35 @@ import PostMarkdown from 'components/post_markdown';
...
-        /*
-         * Post type components from plugins
-         */
-        pluginPostTypes: PropTypes.object,
-        currentRelativeTeamUrl: PropTypes.string.isRequired,
-    };
+type Props = {
+    post: Post; /* The post to render the message for */
+    enableFormatting?: boolean; /* Set to enable Markdown formatting */
+    options?: any; /* Options specific to text formatting */
```
jespino, line 19: I think this is `TextFormattingOptions` types from `utils/text_formatting` (you can see an example in `components/markdown/markdown.tsx`

jespino (changes requested):
Some types improvements needed, but otherwise looks good to me. Thanks @pikami! :tada:

components/post_view/post_message_view/post_message_view.tsx
pikami: I need some advice on this, the `PostPluginComponent` type has the `component` property witch is not of a correct type (`React.Component` instead of `React.ComponentType`) so I cannot use it, should I make a PR to the `mattermost-redux` repo with a fix?

components/post_view/post_message_view/post_message_view.test.tsx
pikami: Comment removed.

components/post_view/post_message_view/post_message_view.tsx
pikami: Updated.

components/post_view/post_message_view/post_message_view.tsx
pikami: Updated.

components/post_view/post_message_view/post_message_view.tsx
jespino: Would be great but for now, lets keep it here as any.

jespino (approved):
Now looks good to me. Thanks @pikami! :tada:

nickmisasi (approved):
Sorry for the delay! Everything looks good here. Thanks @pikami! 🎉

hanzei:
Thank you for the contribution @pikami :+1:

PR #6755, opened 2020-10-13
[MM-27834] Collapse on more than 5 plugin buttons cypress tests

iomodo:
#### Summary
This PR will Add a cypress test and check that channel header buttons are being collapsed when there are more than 5.

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-27834


e2e/cypress/integration/plugins/plugin_buttons_spec.js
```diff
@@ -0,0 +1,69 @@
...
+        // # Try to remove demo plugin, just in case
+        cy.apiRemovePluginById('com.mattermost.demo-plugin');
+
+        // # Upload and enable test plugin with 5 channel header icons
+        cy.apiUploadPlugin('com.mattermost.test-plugin-0.1.0.tar.gz').then(() => {
+            cy.apiEnablePluginById('com.mattermost.test-plugin');
+        });
+
+        // * Validate that there are 10 channel header icons
+        cy.get('.channel-header__icon').should('have.length', 10);
```
larkox, line 56: It is not obvious that apart of the plugin icons, there are other 5 icons on the channel header. Also, it may become quite brittle if any of those icons is removed from the channel header. Is there any better way to check this?

larkox (approved):
LGTM. Just a small consideration, in case it is possible.

e2e/cypress/integration/plugins/plugin_buttons_spec.js
iomodo: There are two ways we could improve this:
1. Add dynamically calculated id-s to the channel header buttons which is not trivial amount of work.
2. Do not validate exact number of icons like 10, but verify the difference of 4 when adding an extra icon.
I'm 0/5 if we should do either of them. Maybe @DHaussermann will have some insight

e2e/cypress/integration/plugins/plugin_buttons_spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+        // # Upload and enable demo plugin with one additional channel header icon
+        cy.apiUploadPlugin('com.mattermost.demo-plugin-0.2.0.tar.gz').then(() => {
+            cy.apiEnablePluginById('com.mattermost.demo-plugin');
+        });
+
+        // * Validate that channel header icons collapsed and there are only 6
+        cy.get('.channel-header__icon').should('have.length', 6);
```
mickmister, line 64: How did we go from 10 icons to 6 icons after enabling the demo plugin?

mickmister: Do you have any thoughts on this @iomodo?

DHaussermann:
Hi @iomodo I can't see this test pass.
I think that it want's 10 items in the drop down. Stagily my local has a couple extra in addition to those provided by the test plugin but I'm still not at 10. 
Screen Shot 2020-11-03 at 9 19 55 AM
If this is in-fact the problem, can we change the test to be less dependent on existing data and have the test plugin create all the objects needed?

e2e/cypress/integration/plugins/plugin_buttons_spec.js
mickmister: @DHaussermann Can you take a look at the code near this section to see if the issue is here?

DHaussermann:
@iomodo any thoughts on this?
The test seems to expect 10 plugins but your test plugin only adds 5.

e2e/cypress/integration/plugins/plugin_buttons_spec.js
iomodo: @mickmister whenever there are more than 5 plugin icons header icons collapse and we get 1 containing all 5. That is from 10 icons total to 6.

iomodo:
> @iomodo any thoughts on this?
> The test seems to expect 10 plugins but your test plugin only adds 5.

@DHaussermann updated the tests, by removing all the plugins installed beforehand.

DHaussermann (approved):
Tested and passed
Thanks @iomodo this change resolves the issue I was seeing.
This test matches the TM4J task
LGTM!

Updated cypress status in TM4J

e2e/cypress/integration/plugins/plugin_buttons_spec.js
mickmister: Can we add a css class to the plugin buttons and count them that way?

DHaussermann:
@iomodo This has been through 2 reviews. Please merge this when you get a chance.

PR #6764, opened 2020-10-13
[MM-29559][MM-29558] Company Info Fixes

devinbinnie:
#### Summary
Two fixes:
- Fixed the Company Info flickering by making sure the component doesn't render until customer is loaded
- Fixed the missing address when `company_address` is not null but rather everything is blank, so now it checks if `line1` is blank or not.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29558
https://mattermost.atlassian.net/browse/MM-29559

### DIFF
--- a/components/admin_console/billing/company_info_display.tsx
+++ b/components/admin_console/billing/company_info_display.tsx
@@ -59,51 +59,52 @@ const noCompanyInfoSection = (
 const CompanyInfoDisplay: React.FC = () => {
     const companyInfo = useSelector((state: GlobalState) => state.entities.cloud.customer);
 
-    let body = noCompanyInfoSection;
+    if (!companyInfo) {
+        return null;
+    }
 
-    if (companyInfo) {
-        const address = companyInfo.company_address || companyInfo.billing_address;
-        body = (
-            <div className='CompanyInfoDisplay__companyInfo'>
-                <div className='CompanyInfoDisplay__companyInfo-text'>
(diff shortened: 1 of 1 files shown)


components/admin_console/billing/company_info_display.tsx
```diff
@@ -59,51 +59,52 @@ const noCompanyInfoSection = (
...
+                        id='admin.billing.company_info.companyAddress'
+                        defaultMessage='Company Address'
+                    />
+                </div>
+                <div className='CompanyInfoDisplay__companyInfo-address'>
+                    <div>{address.line1}</div>
+                    {address.line2 && <div>{address.line2}</div>}
+                    <div>{`${address.city}, ${address.state}, ${address.postal_code}`}</div>
+                    <div>{address.country}</div>
+                </div>
```
marianunez, line 94: Should we include a case here if this address is empty to show something instead of empty commas? "Not available"?

marianunez: Yes, as soon as they sign up they only have Company Name filled in (it's part of the workspace creation) but they don't have anything else. They only will have the billing address once they upgrade. So they do have a company information with name but no address

marianunez (approved):

marianunez:
This will be tested post-merge given that we still can't connect to CWS from PR test servers. Removing `QA Label`

nickmisasi (approved):

PR #6794, opened 2020-10-14
MM-28833 Remove computed details from getChannel selector

hmhealey:
As I mentioned earlier this week, `getChannel` is problematic because its a selector that takes an argument without being behind a factory function, so if you have two or more components both using it, they'll constantly be overwriting each others' results, leading to those components re-rendering constantly. The main reason that this causes problems is that `getChannel` returns a new object each time for DMs and GMs because it adds on some additional derived fields such as the dynamic display_name for DM and GM channels. On the current version of master, this leads to things like parts of the new sidebar and all of the RHS re-rendering constantly.

I changed `getChannel` to be a plain function that doesn't compute any data so that it can be re-used easily wherever we don't need the computed fields, and I've used `makeGetChannel` to generate one-off selectors for everywhere that does need that data.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28833

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1257


components/post_view/index.js
```diff
@@ -53,7 +53,6 @@ function makeMapStateToProps() {
         return {
             lastViewedAt,
             channelLoading,
-            channel,
```
hmhealey, line 56: This isn't actually used by the component

components/rhs_thread/index.ts
```diff
@@ -27,11 +26,10 @@ function makeMapStateToProps() {
         const selected = getSelectedPost(state);
         const socketStatus = getSocketStatus(state);
 
-        let channel = null;
+        const channel = getSelectedChannel(state);
```
hmhealey, line 29: We can get the selected channel directly from the store instead of having to infer it from the selected post to simplify things slightly

nevyangelova (approved):
Thanks! Ran with your redux change and it does rerender less 🎉

sudheerDev (approved):

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified DM & GM headers, DMs in channels switcher, channel names in RHS and search results and GM channel names in desktop notifications.

PR #6796, opened 2020-10-14
Mm 29111 search bar and search results as fc

michelengelen:
#### Summary
- Refactored `SearchResults` and `SearchBar` component to functional components using TypeScript.
- Disconnected both components from redux-store and simplified component structure.
- Implemented new connected component `Search` which handles state updates and renders `SearchResults` (when not in `ChannelHeader`) and `SearchBar`.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15852
Fixes https://github.com/mattermost/mattermost-server/issues/15854

Note to @nevyangelova: Please be mild! ✨


components/search/search.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+        if (_searchTerms.trim() === '') {
+            return searchHintOptions;
+        }
+
+        const pretextArray = _searchTerms.split(/\s+/g);
+        const pretext = pretextArray[pretextArray.length - 1];
+        const penultimatePretext = pretextArray[pretextArray.length - 2];
+
+        const shouldShowHintOptions = penultimatePretext ? !searchHintOptions.some(({searchTerm}) => penultimatePretext.toLowerCase().endsWith(searchTerm.toLowerCase())) : !searchHintOptions.some(({searchTerm}) => _searchTerms.toLowerCase().endsWith(searchTerm.toLowerCase()));
```
nevyangelova, line 66: Could be like:

```suggestion
const shouldShowHintOptionsPredicate = penultimatePretext ? 
    ({searchTerm}) => penultimatePretext.toLowerCase().endsWith(searchTerm.toLowerCase()) : 
    ({searchTerm}) => _searchTerms.toLowerCase().endsWith(searchTerm.toLowerCase());

const shouldShowHintOptions = !searchHintOptions.some(shouldShowHintOptionsPredicate);
```

components/search/search.tsx
```diff
@@ -0,0 +1,391 @@
...
+        actions.updateSearchTerms(term);
+    };
+
+    const updateHighlightedSearchHint = (indexDelta: number, changedViaKeyPress = false): void => {
+        let newIndex = highlightedSearchHintIndex + indexDelta;
+
+        switch (indexDelta) {
+        case 1:
+            // KEY.DOWN
+            // is it at the end of the list?
```
nevyangelova, line 132: 👏 very nice comments overall

components/search/search.tsx
```diff
@@ -0,0 +1,391 @@
...
+    };
+
+    const handleSearch = async (): Promise<void> => {
+        const terms = searchTerms.trim();
+
+        if (terms.length === 0) {
+            return;
+        }
+
+        if (terms.length) {
```
nevyangelova, line 182: if we already checked for 0 terms isn't this redundant? Aren't we meaning to check for more than 1 term?

components/search/search.tsx
```diff
@@ -0,0 +1,391 @@
...
+                        onOptionSelected={handleAddSearchTerm}
+                        onMouseDown={handleSearchHintSelection}
+                        highlightedIndex={highlightedSearchHintIndex}
+                        onOptionHover={setHoverHintIndex}
+                    />
+                </Popover>
+            );
+        }
+
+        return <></>;
```
nevyangelova, line 299: Instead we can do an early exit (this is kind of the practice at Mattermost) - flip the condition and return empty first.

components/search_bar/search_bar.tsx
```diff
@@ -0,0 +1,140 @@
...
+            e.preventDefault();
+            props.updateHighlightedSearchHint(-1, true);
+        }
+
+        if (Utils.isKeyPressed(e, KeyCodes.ENTER)) {
+            props.handleEnterKey(e);
+        }
+    };
+
+    const getSearch = (node: SuggestionBox): void => {
```
nevyangelova, line 81: > note: doesn't need to be change as part of this PR but we need to file a separate ticket to get rid of this ref function or come up with another solution.

This is what the comment was referring to. This should be added to the redux state.

components/search_results/search_results.tsx
```diff
@@ -0,0 +1,297 @@
...
+    <div
+        {...props}
+        className='scrollbar--vertical'
+    />
+);
+
+export const shouldRenderFromProps = (props: Props, nextProps: Props): boolean => {
+    // Shallow compare for all props except 'results'
+    for (const key in nextProps) {
+        // if (!nextProps.hasOwnProperty(key) || key === 'results') {
```
nevyangelova, line 52: nit: remove comment

components/search_results/search_results.tsx
```diff
@@ -0,0 +1,297 @@
...
+        isMentionSearch,
+        isOpened,
+        updateSearchTerms,
+    } = props;
+
+    const noResults = (!results || !Array.isArray(results) || results.length === 0);
+    const isLoading = isSearchingTerm || isSearchingFlaggedPost || isSearchingPinnedPost || !isOpened;
+    const showLoadMore = !isSearchAtEnd && !isFlaggedPosts && !isPinnedPosts;
+
+    let ctls;
```
nevyangelova, line 141: nit: better naming?

components/search_results/search_results.tsx
```diff
@@ -0,0 +1,297 @@
...
+    let loadingMorePostsComponent;
+
+    let sortedResults = results;
+
+    const titleDescriptor: Record<string, string> = {};
+    const noResultsprops: NoResultsProps = {
```
nevyangelova, line 147: nit: camel case

components/search_results/search_results.tsx
```diff
@@ -0,0 +1,297 @@
...
+                >
+                    {ctls}
+                    {loadingMorePostsComponent}
+                </div>
+            </Scrollbars>
+        </div>
+    );
+};
+
+React.memo(SearchResults, shouldRenderFromProps);
```
nevyangelova, line 293: 👏

components/sidebar_right/sidebar_right.jsx
```diff
@@ -154,92 +152,47 @@ export default class SidebarRight extends React.PureComponent {
...
             previousRhsState,
             searchVisible,
             isPluginView,
+            isOpen,
             isExpanded,
         } = this.props;
 
         let content = null;
-        let expandedClass = '';
+        let isSbrExpanded = false;
```
nevyangelova, line 166: nit: better naming

components/sidebar_right/sidebar_right.jsx
```diff
@@ -154,92 +152,47 @@ export default class SidebarRight extends React.PureComponent {
...
+        case postCardVisible:
+            isSbrExpanded = isExpanded;
+            content = <RhsCard previousRhsState={previousRhsState}/>;
+            break;
+        case isPluginView:
+            isSbrExpanded = isExpanded;
+            content = <RhsPlugin/>;
+            break;
+        case searchVisible:
+            isSbrExpanded = isExpanded;
```
nevyangelova, line 189: Note: switch(true) is nice, but I think it will be more clear for the `isSbrExpanded` to be assigned more explicitly, so I would extract it.

nevyangelova (changes requested):
Thanks @michelengelen ! Great work 🎉

components/search/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+} from 'actions/views/rhs';
+import {autocompleteChannelsForSearch} from 'actions/channel_actions';
+import {autocompleteUsersInTeam} from 'actions/user_actions';
+import {getRhsState, getSearchTerms, getIsSearchingTerm, getIsRhsOpen, getIsRhsExpanded} from 'selectors/rhs';
+import {RHSStates} from 'utils/constants';
+import {GlobalState} from 'types/store';
+
+import Search from './search';
+
+type OwnProps = {
```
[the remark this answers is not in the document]
michelengelen: sadly not ... only some are defined, but most Props from the state are calculated ones.

components/search/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    isSideBarRight?: boolean,
+    isSideBarRightOpen?: boolean,
+    isFocus: boolean,
+    channelDisplayName?: string,
+    getFocus?: (searchBarFocus: () => void) => void,
```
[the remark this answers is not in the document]
michelengelen: absolutely

components/search/search.tsx
```diff
@@ -0,0 +1,391 @@
...
+import Popover from 'components/widgets/popover';
+import UserGuideDropdown from 'components/channel_header/components/user_guide_dropdown';
+
+import SearchBar from 'components/search_bar/search_bar';
+import SearchResults from 'components/search_results/index';
+import SearchDateProvider from 'components/suggestion/search_date_provider';
+import SearchChannelProvider from 'components/suggestion/search_channel_provider';
+import SearchUserProvider from 'components/suggestion/search_user_provider';
+
+import type {Props} from './';
```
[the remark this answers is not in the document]
michelengelen: good idea ... will do!

components/search/search.tsx
michelengelen: thanks! ✨

components/search/search.tsx
michelengelen: true ... will change!

components/sidebar_right/sidebar_right.jsx
michelengelen: done!

components/sidebar_right/sidebar_right.jsx
michelengelen: done!

components/search_results/search_results.tsx
michelengelen: done! (dunno why I overlooked that!) 😅

components/search_results/search_results.tsx
michelengelen: done

components/search_results/search_results.tsx
michelengelen: done!

components/search/search.tsx
michelengelen: done

deanwhillier (commented):
Thanks for the contribution @michelengelen, much appreciated!

After a quick scan I have a couple initial comments:

1. `components/search_results/search_results.tsx`: our webpack configuration currently prefers `.js` & `.jsx` over `.ts` & `tsx` files meaning that when I run these changes locally, `search_results.jsx` is being loaded instead of `search_results.tsx`, which is causing issues with loading search results. Are you seeing the same thing on your end? Can we remove `search_results.jsx`?
2. See below.

components/search_bar/search_bar.tsx
```diff
@@ -0,0 +1,140 @@
...
+        >
+            <form
+                role='application'
+                className={searchFormClass}
+                onSubmit={props.handleSubmit}
+                style={style.searchForm}
+                autoComplete='off'
+                aria-labelledby='searchBox'
+            >
+                <div className='search__font-icon TEST'>
```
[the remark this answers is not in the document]
michelengelen: no, absolutely not! 😅

michelengelen:
> 1. `components/search_results/search_results.tsx`: our webpack configuration currently prefers `.js` & `.jsx` over `.ts` & `tsx` files meaning that when I run these changes locally, `search_results.jsx` is being loaded instead of `search_results.tsx`, which is causing issues with loading search results. Are you seeing the same thing on your end? Can we remove `search_results.jsx`?

No, it does not cause any errors on my side. I just missed it, because Webstorm groups files with the same name, but different endings (`.js` & `.ts` and `.jsx` & `.tsx`), which is kinda nice, but sometimes leads to these cases. 

I will remove it!

nevyangelova (approved):
LGTM!

nevyangelova (changes requested):
I approved a bit early. The search behaviour is not correct, which is expected with such a significant change. When I try to search for messages from users, search results component appears over the hint. 

Screenshot 2020-10-16 at 17 06 28

Padding seems a bit more than on master

Screenshot 2020-10-16 at 17 07 08

Search results header appears on the popover and on master its not shown
Screenshot 2020-10-16 at 17 08 40

Cursor moves out of the quotes when searching for phrases, with `" "`
Screenshot 2020-10-16 at 17 10 22

@jgilliam17 can you help provide additional QA feedback?

components/search_results/search_results.tsx
```diff
@@ -0,0 +1,296 @@
...
+);
+
+const renderThumbVertical = (props: Record<string, unknown>): JSX.Element => (
+    <div
+        {...props}
+        className='scrollbar--vertical'
+    />
+);
+
+export const shouldRenderFromProps = (props: Props, nextProps: Props): boolean => {
```
hmhealey, line 49: Nothing to change right now, but we may not actually need this going forward. The previous `SearchResults` was written before we were using `PureComponent` everywhere, and I don't think we understood `shouldComponentUpdate` at the time either

components/search_results/search_results.tsx
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+
+    return false;
+};
+
+interface NoResultsProps {
+    variant: NoResultsVariant,
+    titleValues?: Record<string, JSX.Element|string>
+    subtitleValues?: Record<string, JSX.Element|string>
```
hmhealey, line 83: These types can be `Record<string, React.ReactNode>`

components/search_results/search_results.tsx
```diff
@@ -0,0 +1,296 @@
...
+    let loadingMorePostsComponent;
+
+    let sortedResults = results;
+
+    const titleDescriptor: Record<string, string> = {};
+    const noResultsProps: NoResultsProps = {
+        variant: NoResultsVariant.ChannelSearch,
+    };
+
+    switch (true) {
```
hmhealey, line 150: Can you replace this block with an if-else instead? I can understand what it does after looking at it for a minute, but I prefer to be more obvious with what code does since other people will have to read and understand it as well.

Same with where you're doing that below

components/search_results/index.tsx
```diff
@@ -69,12 +73,5 @@ function makeMapStateToProps() {
...
-    return {
-        actions: bindActionCreators({
-            getMorePostsForSearch,
-        }, dispatch),
-    };
-}
-
-export default connect(makeMapStateToProps, mapDispatchToProps)(SearchResults);
+// eslint-disable-next-line @typescript-eslint/ban-types
+export default connect<StateProps, {}, OwnProps, GlobalState>(makeMapStateToProps)(injectIntl<'intl', Props>(SearchResults));
```
hmhealey, line 77: Instead of injectIntl, we can use `useIntl` in SearchResults as well

components/sidebar_right/sidebar_right.jsx
```diff
@@ -249,7 +196,15 @@ export default class SidebarRight extends React.PureComponent {
                     className='sidebar--right__bg'
                 />
                 <div className='sidebar-right-container'>
-                    {content}
+                    <Search
```
hmhealey, line 199: What's the reason for wrapping `content` in the new `Search` component now? We don't necessarily need to change it, but it seems like this might have scaling problems if we were to change the other sidebar components to wrap each other like this.

hmhealey (changes requested):

components/search_results/search_results.tsx
nevyangelova: I agree with most of your points but I personally find the switch more readable, instead of chaining 5-6 if else statements like it was. I know there might be some performance issues but not in the case of this component.

components/search/index.tsx
```diff
@@ -1,49 +1,52 @@
...
             autocompleteChannelsForSearch,
             autocompleteUsersInTeam,
             updateRhsState,
+            getMorePostsForSearch,
         }, dispatch),
     };
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(SearchBar);
+export default connect<StateProps, DispatchProps, OwnProps, GlobalState>(mapStateToProps, mapDispatchToProps)(Search);
```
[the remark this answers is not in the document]
michelengelen: It is not ultimately necessary for typescript, but it leaves a vulnerability if not.

I was having your (https://developers.mattermost.com/contribute/webapp/migrating-to-typescript/)[migrating to typescript] guide in mind

components/search/search.tsx
```diff
@@ -0,0 +1,392 @@
...
+    // generate intial component state and setters
+    const [focussed, setFocussed] = useState<boolean>(false);
+    const [keepInputFocussed, setKeepInputFocussed] = useState<boolean>(false);
+    const [indexChangedViaKeyPress, setIndexChangedViaKeyPress] = useState<boolean>(false);
+    const [highlightedSearchHintIndex, setHighlightedSearchHintIndex] = useState<number>(-1);
+    const [visibleSearchHintOptions, setVisibleSearchHintOptions] = useState<SearchHintOption[]>(
+        determineVisibleSearchHintOptions(searchTerms),
+    );
+
+    const suggestionProviders = [
```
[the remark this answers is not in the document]
michelengelen: I will add `useRef`, so that it does not get re-initiated on every render. Nice catch! ✨

components/search/search.tsx
```diff
@@ -0,0 +1,392 @@
...
+            });
+        } catch {
+            newVisibleSearchHintOptions = [];
+        }
+    }
+
+    return newVisibleSearchHintOptions;
+};
+
+const Search: React.FC<Props> = (props: Props): JSX.Element => {
```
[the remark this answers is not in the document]
michelengelen: I don't think that would be a good idea. Again, in the developer guide for migrating to typescript it says every component should be as well defined as possible.

components/search/search.tsx
```diff
@@ -0,0 +1,392 @@
...
+    };
+
+    const updateHighlightedSearchHint = (indexDelta: number, changedViaKeyPress = false): void => {
+        let newIndex = highlightedSearchHintIndex + indexDelta;
+
+        switch (indexDelta) {
+        case 1:
+            // KEY.DOWN
+            // is it at the end of the list?
+            newIndex = newIndex === visibleSearchHintOptions.length ? 0 : newIndex;
```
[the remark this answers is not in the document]
michelengelen: I don't see where this could help here?
It is a modulo function, that will return `NaN` when `visibleSearchHintOptions.length === 0`.
Or did you mean to use it another way?

components/search/search.tsx
```diff
@@ -0,0 +1,392 @@
...
+    return newVisibleSearchHintOptions;
+};
+
+const Search: React.FC<Props> = (props: Props): JSX.Element => {
+    const {actions, searchTerms} = props;
+
+    // generate intial component state and setters
+    const [focussed, setFocussed] = useState<boolean>(false);
+    const [keepInputFocussed, setKeepInputFocussed] = useState<boolean>(false);
+    const [indexChangedViaKeyPress, setIndexChangedViaKeyPress] = useState<boolean>(false);
```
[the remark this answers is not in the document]
michelengelen: It is not ultimately affecting rendering, but since it does not trigger unnecessary re-renders it could stay like this.
We could otherwise utilise `useRef`, but it is less elegant

components/search/search.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+    const renderSearchBar = ():JSX.Element => (
+        <>
+            <div className='sidebar-collapse__container'>
+                <div
+                    id={props.isSideBarRight ? 'sbrSidebarCollapse' : 'sidebarCollapse'}
+                    className='sidebar-collapse'
+                    onClick={handleClose}
+                >
+                    <FormattedMessage
```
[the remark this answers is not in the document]
michelengelen: done ... did not know of that. Thanks for the catch!

components/search_bar/search_bar.tsx
```diff
@@ -0,0 +1,140 @@
...
+    };
+
+    const getSearch = (node: SuggestionBox): void => {
+        searchRef.current = node;
+        if (props.getFocus) {
+            props.getFocus(props.handleFocus);
+        }
+    };
+
+    const searchFormClass = `search__form${isFocussed ? ' search__form--focused' : ''}`;
```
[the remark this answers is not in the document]
michelengelen: done

components/search_bar/search_bar.tsx
```diff
@@ -0,0 +1,140 @@
...
+                <SuggestionBox
+                    ref={getSearch}
+                    id={props.isSideBarRight ? 'sbrSearchBox' : 'searchBox'}
+                    tabIndex='0'
+                    className='search-bar a11y__region'
+                    containerClass='w-full'
+                    data-a11y-sort-order='9'
+                    aria-describedby={props.isSideBarRight ? 'sbr-searchbar-help-popup' : 'searchbar-help-popup'}
+                    aria-label={Utils.localizeMessage('search_bar.search', 'Search')}
+                    placeholder={Utils.localizeMessage('search_bar.search', 'Search')}
```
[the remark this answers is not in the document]
michelengelen: done

components/search_results/search_results.tsx
michelengelen: done

components/search/search.tsx
```diff
@@ -0,0 +1,392 @@
...
+    };
+
+    const handleSearch = async (): Promise<void> => {
+        const terms = searchTerms.trim();
+
+        if (terms.length === 0) {
+            return;
+        }
+
+        const {error} = await actions.showSearchResults?.(Boolean(props.isMentionSearch));
```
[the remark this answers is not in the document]
michelengelen: done

components/search_results/search_results.tsx
michelengelen: I do agree with @nevyangelova here

components/search_results/index.tsx
michelengelen: done

components/sidebar_right/sidebar_right.jsx
michelengelen: We did that because we wanted to have just one connected component instead of several others down the dependency tree. I understand the reasoning and I do think that you have a point (especially with the ref handling), but this should be handled in another issue (refactoring `SuggestionBox` #15855 is a big step forward)

There is still the `SearchResultItem` which is also connected, but again: this should be handled in a different issue

michelengelen:
@nevyangelova I did provide some fixes for the comments you made on the PR. Could you re-check again? 

Thanks for the great feedback so far!

> search results component appears over the hint.
> 
> Screenshot 2020-10-16 at 17 06 28

**FIXED**

> Padding seems a bit more than on master
> 
> Screenshot 2020-10-16 at 17 07 08

I did not change any styles. I could not notice any changes when comparing local, test-server and production.

> Search results header appears on the popover and on master its not shown
> Screenshot 2020-10-16 at 17 08 40

I do not really know what you mean by that. Could you elaborate a bit? that would be really helpful! 🙇‍♂️ 

> Cursor moves out of the quotes when searching for phrases, with `" "`
> Screenshot 2020-10-16 at 17 10 22

**FIXED!** (that was a nasty bug 😓 )

components/sidebar_right/sidebar_right.jsx
nevyangelova: I agree that those should be handled in the next steps, we need to first see how this and SuggestionBox will come together.

nevyangelova (approved):

components/search/index.tsx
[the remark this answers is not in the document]
hmhealey: Fair enough. I don't think we need to require them since it adds more overhead to the TypeScript migration on top of what people already need to understand about it, but it certainly doesn't hurt to have here.

components/search/search.tsx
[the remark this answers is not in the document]
hmhealey: Interesting. I didn't know that's how you simulate instance variables with functional components. That naming sounded like we were misusing something at first, so I was surprised to see that that's the recommended way to do that.

components/search/search.tsx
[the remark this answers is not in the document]
hmhealey: I meant that it would handle keeping the index between 0 and the length like this is doing but in a more concise way. I didn't realize that there's a case where the length would be 0 though.

components/search_results/search_results.tsx
hmhealey: As someone who now knows what this does, I don't see much difference between
```javascript
if (a) {
} else if (b) {
} else {
}
```
and
```javascript
switch(true) {
case a:
    break;
case b:
    break;
default:
}
```
readability-wise other than the switch being slightly longer because of the breaks (which does seem unnecessary), but I'm more concerned with readability for people who aren't familiar with the code. Since this is an open source project, we need to be careful to make sure that others can understand and work with the code, and that means we have to avoid "clever" stuff that isn't immediately obvious to the reader when there's an easier way to write it.

components/sidebar_right/sidebar_right.jsx
hmhealey: I'm not sure I understand. I was meaning the structure changes going from something like
```javascript
let content;
if (threadVisible) {
    content = <Thread/>;
} else if (searchVisible) {
    content = <Search/>;
}

return content;
```
to
```javascript
let content;
if (threadVisible) {
    content = <Thread/>;
}

return <Search searchVisible={searchVisible}>{content}</Search>;
```
looked a bit odd to me since if we were following the new pattern for each RHS state (thread, search, plugin-overridden, etc), we'd end up with a more deeply nested set of components instead of only mounting the one that is needed at any given time.

components/search/search.tsx
```diff
@@ -0,0 +1,392 @@
...
+        // when focus is released from the search box.
+        setTimeout((): void => {
+            if (keepInputFocussed) {
+                setKeepInputFocussed(false);
+            } else {
+                setFocussed(false);
+            }
+        }, 0);
+
+        updateHighlightedSearchHint(0);
```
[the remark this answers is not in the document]
michelengelen: I refactored the function and when calling without parameters it resets the highlights

components/search/search.tsx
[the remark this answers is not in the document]
michelengelen: fair enough ...

components/search_results/search_results.tsx
michelengelen: mmh, i know what you mean, but in my opinion this is a very basic pattern in javascript and increases readability by a fair amount. This is cited from an SO-post about performance comparison of `if/else` and `switch`: 

> That's micro optimization and premature optimization, which are evil. Rather worry about readabililty and maintainability of the code in question. If there are more than two `if/else` blocks glued together or its size is unpredictable, then you may highly consider a `switch` statement.

So performance wise there is no difference (`switch` is slightly faster, but almost unmeasurable). And readability is improved with a `switch` statement.

Any opinions on that @nevyangelova?

components/sidebar_right/sidebar_right.jsx
michelengelen: It is a decision born from integration issues with the `SearchBar` component (once with `SearchResults` in the `SidebarRight` and once without in `ChannelHeader`). Do you have another idea how this could be handled in y more "clever" way?

components/search_results/search_results.tsx
nevyangelova: I still hold my opinion about keeping the `switch` statement. The switch statement is faster in most cases when compared to if-else, but significantly faster only when the number of conditions is large. The primary difference in performance between the two is that the incremental cost of an additional condition is larger for if-else than it is for switch. Therefore, our natural inclination to use if-else for a small number of conditions and a switch statement for a larger number of conditions is exactly the right advice when considering performance.

nevyangelova:
kind reminder for @jgilliam17 to check for additional UX bugs 🎉

jgilliam17:
Thank you @michelengelen 
Tested search and search results - looks good
Noticed one issue with exclude search terms
See screenshot 
e.g. I searched for `from:sysadmin -goldfinch`
Expected for roughly all the posts to display in the search results except for 1st post on Town Square that contains `goldfinch` and maybe one more 
Observed: `-` fails to return any results, excludes all when used as the only search option or in combination with `from:`

michelengelen:
> Thank you @michelengelen
> Tested search and search results - looks good
> Noticed one issue with exclude search terms
> See screenshot
> e.g. I searched for `from:sysadmin -goldfinch`
> Expected for roughly all the posts to display in the search results except for 1st post on Town Square that contains `goldfinch` and maybe one more
> Observed: `-` fails to return any results, excludes all when used as the only search option or in combination with `from:`

Does not seem to be a problem with my refactoring. i did not change anything on the submitting part. I will have a look anyways and come back to you with my findings!

jgilliam17:
> Does not seem to be a problem with my refactoring. i did not change anything on the submitting part. I will have a look anyways and come back to you with my findings!

@michelengelen Thanks for checking! We can definitely address this separately if it's out of scope (I can create a follow up issue). I ran few search tests to check Search Results as style/padding was mentioned in previous comments, and noticed `-` wasn't producing results I expected.

michelengelen:
@jgilliam17 I did check on local, test-server and production that using `-` behaves differently.

First of all: requests are the same on every environment: Status 200, request params are OK as well. The only difference is that the response is not correct on the test-server and on local. Any idea what can cause this?

jgilliam17:
I checked the system console and test-server search queries are using database search, Bleve and Elasticsearch are not enabled. I'd have to re-test and see if that makes a difference in using `-`.

jgilliam17:
Thanks @michelengelen 
Re-tested - Looks good to merge.

Note: I enabled Bleve and `-` worked as expected, this must be a database search issue 
For some reason I am now seeing an issue with `from:` it does not offer sysadmin as an option or even auto-complete when typing `sys`.  I'll log that separately, Ogi confirmed it's happening on his test server too unrelated to this PR.

jgilliam17 (approved):

michelengelen:
@hmhealey kind reminder to review/approve. ✨

components/search_results/search_results.tsx
hmhealey: Sorry, I meant to continue this conversation sooner since I'm 5/5 that this is the wrong way to do this. I know it's a silly thing to be arguing about, but my main problem with this is that it's making a simple piece of code less obvious for people working on this code, especially those who may be seeing it for the first time. We'll have a lot of new people in that situation as well since this is open source. Switch statements are intended to be used when you're checking if something is one of a number of different values, and the thing where I get caught on is that we're using it to check many different conditions here. It is technically correct this way, but I don't want people to have to look at StackOverflow to understand a conditional statement.

hmhealey: I forgot to reply with a review so that this thread doesn't get lost, but I still would like that changed. I'd be up to chat more about it though if you'd like.

hmhealey (changes requested):

components/search_results/search_results.tsx
michelengelen: I can totally understand your reasoning behind it, but I also think that working on open source software should be seen as opportunities to learn new things. I am aligning with the coding guidelines as well. Simple checks (1-2 easy conditions) are best handled with `if-else`, more discrete values should require the usage of a `switch` statement.

It is a different story when you do have very complex conditions, but with these simple ones (all except one are single boolean values) a `switch` should be the obvious choice

michelengelen: @hmhealey i did change it to a if/else block. I can relate to your reasoning and don't think arguing about it will bring us anywhere, since it is mostly a matter of personal coding style.

components/search_results/search_results.tsx
calebroseland: @michelengelen, I think the return of `React.memo` is what needs to be used here. E.g.:
```tsx
export default React.memo(SearchResults, shouldRenderFromProps);
```

calebroseland (changes requested):

components/search_results/search_results.tsx
michelengelen: you are absolutely right ... nice catch

michelengelen:
@hanzei I did already address all requested changes from @hmhealey and @calebroseland ... is there more to do atm?

calebroseland (approved):
Thanks, @michelengelen!

hmhealey (approved):

hmhealey:
Looks good now! Thanks for changing that

nevyangelova:
@michelengelen @jgilliam17 @hmhealey there are errors in the search logic:
1. Search is stuck on 'Searching...' even when results are present.
2. Search terms aren't updated on second click.
3. Search terms when clicked from rhs don't display the suggestions, for example clicking on 'from:' doesn't return a list of users.

Reverting this PR for now as we will need a deeper investigations to fix those issues. I don't know how we missed those during all of our testing but it's a learning mistake for our team. Sorry Michel!

Michel would you like to pair on this and talk about the potential fixes when you have some time?

michelengelen:
@nevyangelova sure thing. Maybe we should discuss a bigger refactoring as well. I don't know how this could happen as well.

PR #6801, opened 2020-10-15
MM-20578 Migrate 'components/admin_console/system_user_detail' module…

pikami:
#### Summary
Migrate 'components/admin_console/system_user_detail' module and associated tests to TypeScript

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/13494


components/admin_console/system_user_detail/index.ts
```diff
@@ -23,10 +27,10 @@ function mapStateToProps(state, ownProps) {
         mfaEnabled: config.EnableMultifactorAuthentication === 'true',
         navigationBlocked: getNavigationBlocked(state),
         showNavigationPrompt: showNavigationPrompt(state),
-    };
+    } as any;
```
jespino, line 30: Why this is any? souldn't it be Props or at least `Partial<Props>`

components/admin_console/system_user_detail/system_user_detail.tsx
```diff
@@ -68,99 +91,100 @@ export default class SystemUserDetail extends React.PureComponent {
...
+            } as UserProfile,
             addTeamOpen: false,
             refreshTeams: true,
+            error: '',
         };
     }
 
-    setTeamsData = (teams) => {
-        const teamIds = teams.map((team) => team.team_id);
+    setTeamsData = (teams: Record<string, any>): void => {
```
jespino, line 101: Shouldn't it be Record<string, Team>?

components/admin_console/system_user_detail/system_user_detail.tsx
```diff
@@ -451,21 +475,18 @@ export default class SystemUserDetail extends React.PureComponent {
...
                         className='error-message'
                         ref='errorMessage'
-                        onMouseOver={this.openTooltip}
-                        onMouseOut={this.closeTooltip}
                     >
                         <FormError error={this.state.serverError}/>
                     </div>
                     <Overlay
                         show={this.state.errorTooltip}
-                        delayShow={Constants.OVERLAY_TIME_DELAY}
```
jespino, line 468: Why this is removed?

components/admin_console/system_user_detail/system_user_detail.tsx
```diff
@@ -451,21 +475,18 @@ export default class SystemUserDetail extends React.PureComponent {
                 <div className='admin-console-save'>
                     <SaveButton
                         saving={this.state.saving}
-                        disabled={!this.state.saveNeeded || (this.canSave && !this.canSave())}
+                        disabled={!this.state.saveNeeded}
```
jespino, line 478: Nice caches around not-existing fields/functions

components/admin_console/system_user_detail/system_user_detail.tsx
```diff
@@ -68,99 +91,100 @@ export default class SystemUserDetail extends React.PureComponent {
...
         this.setState({refreshTeams: false});
     }
 
-    openAddTeam = () => {
+    openAddTeam = (): void => {
         this.setState({addTeamOpen: true});
     }
 
-    addTeams = (teams) => {
+    addTeams = (teams: Array<any>): void => {
```
jespino, line 112: Shouldn't it be `Array<Team>`?

components/admin_console/system_user_detail/system_user_detail.tsx
```diff
@@ -33,26 +38,44 @@ import SheidOutlineIcon from 'components/widgets/icons/shield_outline_icon.jsx';
...
+        updateUserActive: (userId: string, active: boolean) => Promise<ActionResult | ActionResult[]>;
+        setNavigationBlocked: (blocked: boolean) => void;
+        addUserToTeam: (teamId: string, userId?: string) => Promise<{data: TeamMembership; error?: any}>;
     }
+}
 
-    static defaultProps = {
+export type State = {
+    teams: Record<string, any> | null;
+    teamIds: Array<string> | null;
```
jespino, line 54: Maybe we can be more expressive here and use `Array<$ID<Team>>`

components/admin_console/system_user_detail/system_user_detail.tsx
```diff
@@ -33,26 +38,44 @@ import SheidOutlineIcon from 'components/widgets/icons/shield_outline_icon.jsx';
...
+    actions: {
+        updateUserActive: (userId: string, active: boolean) => Promise<ActionResult | ActionResult[]>;
+        setNavigationBlocked: (blocked: boolean) => void;
+        addUserToTeam: (teamId: string, userId?: string) => Promise<{data: TeamMembership; error?: any}>;
     }
+}
 
-    static defaultProps = {
+export type State = {
+    teams: Record<string, any> | null;
```
jespino, line 53: This `any` should be `Team`, I think.

components/admin_console/system_user_detail/system_user_detail.test.tsx
```diff
@@ -24,7 +25,7 @@ describe('components/admin_console/system_user_detail', () => {
             setNavigationBlocked: jest.fn(),
             addUserToTeam: jest.fn(),
         },
-    };
+    } as any;
```
jespino, line 28: We can use `Props` or `Partial<Props>` here.

jespino (changes requested):
Some changes needed, but is looking pretty good. :)

components/admin_console/system_user_detail/system_user_detail.tsx
pikami: Overlay doesn't have a property `delayShow` so type-check marks this as wrong

components/admin_console/system_user_detail/system_user_detail.tsx
```diff
@@ -68,99 +91,100 @@ export default class SystemUserDetail extends React.PureComponent {
...
 
-    handleDeactivateMember = () => {
+    handleDeactivateMember = (): void => {
         this.props.actions.updateUserActive(this.props.user.id, false).
             then(this.onUpdateActiveResult);
         this.setState({showDeactivateMemberModal: false});
     }
 
-    onUpdateActiveResult = ({error}) => {
+    onUpdateActiveResult = ({error}: any): void => {
```
devinbinnie, line 160: Does this have a type? It could be `ServerError` or something like that.

devinbinnie (changes requested):
Thanks for the PR @pikami!
I've added a few comments for you to review.

components/admin_console/system_user_detail/system_user_detail.tsx
pikami: This function was required in `team_list.tsx` so I assumed it had  the correct type and added it here, turns out that wasn't the case, updated both places with `GroupTeam[]` type

components/admin_console/system_user_detail/system_user_detail.tsx
pikami: Fixed.

components/admin_console/system_user_detail/system_user_detail.tsx
pikami: Updated.

components/admin_console/system_user_detail/system_user_detail.tsx
pikami: Updated.

components/admin_console/system_user_detail/index.ts
```diff
@@ -2,19 +2,23 @@
...
+import {GlobalState} from 'mattermost-redux/types/store';
+
 import {setNavigationBlocked, deferNavigation, cancelNavigation, confirmNavigation} from 'actions/admin_actions.jsx';
 import {getNavigationBlocked, showNavigationPrompt} from 'selectors/views/admin';
 
-import SystemUserDetail from './system_user_detail.jsx';
+import SystemUserDetail, {Props} from './system_user_detail';
 
-function mapStateToProps(state, ownProps) {
+function mapStateToProps(state: GlobalState, ownProps: Props & { match: any }) {
```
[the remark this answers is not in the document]
pikami: Updated, couln't use the `Match` type because it's `params` does not have `user_id`

components/admin_console/system_user_detail/system_user_detail.tsx
```diff
@@ -33,26 +38,44 @@ import SheidOutlineIcon from 'components/widgets/icons/shield_outline_icon.jsx';
...
-            email: null,
-        },
+            email: '',
+        } as UserProfile,
         mfaEnabled: false,
-    }
+    } as Props;
 
-    constructor(props) {
+    constructor(props: Props & RouteComponentProps) {
```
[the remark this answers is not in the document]
pikami: The `super()` function needs to take the same props that the component has.

components/admin_console/system_user_detail/system_user_detail.test.tsx
pikami: >We can use Props or Partial<Props> here.

This would require me to mock the whole `RouteComponentProps` interface which I don't think is needed

>Shouldn't need any casting here, if the props are truly optional then they shouldn't be required in the component. Otherwise, they should be added here.

The tests only cover a part of the properties of `User` prop so I don't think we need to add them all, and there's also the problem of `RouteComponentProps` being required, unless I'm missing something

components/admin_console/system_user_detail/index.ts
pikami: Type checking is complaining about actions being incompatible and I was not able to figure out how to fix that.
Could you take a look?

components/admin_console/system_user_detail/system_user_detail.tsx
pikami: `ActionResult` type has error set as `any` type,
By the previous code the state `error` was set to an empty string so I'm assuming this is a string.
Updated.

pikami: Updated.

components/admin_console/system_user_detail/system_user_detail.test.tsx
devinbinnie: What's required from `RouteComponentProps`? If we're not consuming any of the fields in the component then we shouldn't even need to use it.

If we do, we could add those fields to `Props` instead to avoid mocking out the whole `RouteComponentProps` interface.

devinbinnie:
@pikami Could you resolve the conflicts on this PR for us? After that I can have another look.

pikami:
@devinbinnie conflicts resolved.

components/admin_console/system_user_detail/index.ts
```diff
@@ -2,43 +2,43 @@
...
-            getProfiles,
             updateUserActive,
             setNavigationBlocked,
-            deferNavigation,
-            cancelNavigation,
-            confirmNavigation,
             addUserToTeam,
         }, dispatch),
-    };
+    } as any;
```
devinbinnie, line 41: This still shouldn't require an any. If you are getting an error here you may have to use the `ActionCreatorsMapObject` and manually define the types for the actions specified. Have a look at this file for an example: https://github.com/mattermost/mattermost-webapp/blob/master/components/sidebar/index.ts

devinbinnie (changes requested):
Thanks for your hard work @pikami. Just one comment to resolve.

components/admin_console/system_user_detail/index.ts
pikami: I removed the any, but my solution is not elegant.
`setNavigationBlocked` is typed differently from `updateUserActive` and `addUserToTeam` and this results in the `ActionCreatorsMapObject` not being able to return correct types.
I binned these groups of actions separately and then assigned them to the props.

components/admin_console/system_user_detail/index.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+type Actions = {
+    updateUserActive: (userId: string, active: boolean) => Promise<{error: ServerError}>;
+    setNavigationBlocked: (blocked: boolean) => void;
+    addUserToTeam: (teamId: string, userId?: string) => Promise<{data: TeamMembership; error?: any}>;
+}
+
+function mapDispatchToProps(dispatch: Dispatch<GenericAction>) {
+    const apiActions = bindActionCreators<ActionCreatorsMapObject<ActionFunc>, Actions>({
```
devinbinnie, line 45: I don't think we need to break these up into `apiActions` and `uiActions`, seems like unnecessary code bloat.

devinbinnie (approved):
Looks a lot better @pikami, thanks!
One non-blocking comment.

components/admin_console/system_user_detail/index.ts
pikami: I wasn't able to find a way to put them both into one object

jespino (approved):
Looks good to me. Thanks @pikami! :tada:

lindy65 (approved):
Thanks @pikami @jespino - looks good to me 👍

hanzei:
@pikami Could you please resolve the merge conflict? Then the PR is ready to go.

pikami:
@hanzei Merge conflict resolved.

hanzei:
Awesome, thanks @pikami :+1:

PR #6610, opened 2020-10-02
[MM-27188] E2E Tests Automated - Cypress/E2E: Automate backlogs - System Console (11 test cases)

hahmadia:
#### Summary

Automating tests - Cypress/E2E: Automate backlogs - System Console (11 test cases)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27188

PR #6802, opened 2020-10-15
Cypress/E2E: Fix require license on some system console specs

josephbaylon:
#### Summary
- Moved apiRequireLicense only to tests that require it (feedback from https://github.com/mattermost/mattermost-webapp/pull/6610)
- Fixed issues with site statistics

#### Ticket Link
NA

### DIFF
--- a/e2e/cypress/integration/system_console/main_menu_spec.js
+++ b/e2e/cypress/integration/system_console/main_menu_spec.js
@@ -17,8 +17,6 @@ const goToAdminConsole = () => {
 
 describe('Main menu', () => {
     before(() => {
-        cy.apiRequireLicense();
-
         goToAdminConsole();
 
         // # Open the hamburger menu
@@ -59,6 +57,9 @@ describe('Main menu', () => {
     });
 
     it('MM-T913 About opens About modal', () => {
+        // * Check if server has license
+        cy.apiRequireLicense();
+
         // # click to open about modal
         cy.findByText('About Mattermost').click();
 
@@ -79,6 +80,8 @@ describe('Main menu', () => {
     });
 
     it('MM-T899 - Edition and License: Verify Privacy Policy link points to correct URL', () => {
+        // * Check if server has license
+        cy.apiRequireLicense();
         goToAdminConsole();
 
         // * Find privacy link and then assert that the href is what we expect it to be
--- a/e2e/cypress/integration/system_console/reporting/site_statistics_spec.js
+++ b/e2e/cypress/integration/system_console/reporting/site_statistics_spec.js
@@ -20,6 +20,11 @@ const goToAdminConsole = () => {
 describe('System Console > Site Statistics', () => {
     let testTeam;
 
+    afterEach(() => {
+        // # Reset locale
+        cy.apiPatchMe({locale: 'en'});
+    });
+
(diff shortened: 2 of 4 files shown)


e2e/cypress/integration/system_console/reporting/site_statistics_spec.js
```diff
@@ -245,5 +243,8 @@ describe('System Console > Site Statistics', () => {
                 });
             });
         });
+
+        // # Reset locale
+        cy.apiPatchMe({locale: 'en'});
```
srkgupta, line 248: This needs to be added in the after block because if the testcase fails in between, this step might not be executed and all subsequent steps will start to fail.

e2e/cypress/integration/system_console/search_box_spec.js
```diff
@@ -18,11 +18,10 @@ const goToAdminConsole = () => {
 };
 
 describe('System console', () => {
-    before(() => {
+    it('MM-T897_1 - Focus should be in System Console search box on opening System Console or refreshing pages in System Console', () => {
+        // * Check if server has license
         cy.apiRequireLicense();
-    });
```
srkgupta, line 23: This step is only executed on the first testcase. The last test `MM-T1634 - Search box should remain visible / in the header as you scroll down the settings list in the left-hand-side` is also dependent on EE license.

srkgupta (changes requested):
Thanks @josephbaylon. Couple of suggestions.

e2e/cypress/integration/system_console/reporting/site_statistics_spec.js
josephbaylon: Moved the reset to afterEach instead since more tests can be added after this test cases which can be affected if it fails prematurely.

e2e/cypress/integration/system_console/search_box_spec.js
josephbaylon: I added a `apiDeleteLicense` locally on `MM-T1634` and it doesn't seem to require a license. Does it fail for you when you add `apiDeleteLicense`

srkgupta: Sorry. My apologies. I thought Bleve Indexing is only available for EE. Just found that its also available for TE. Please ignore.

srkgupta (approved):
Thanks @josephbaylon. Changes LGTM

e2e/cypress/integration/system_console/main_menu_spec.js
```diff
@@ -59,6 +57,9 @@ describe('Main menu', () => {
     });
 
     it('MM-T913 About opens About modal', () => {
+        // * Check if server has license
+        cy.apiRequireLicense();
```
saturninoabril, line 61: I realized that my thoughts in our chat which was "add inside the `it` block if wanted to maintain test grouping" is not accurate. I think the correct way is to move test that requires license in a separate file and then add metadata of `@enterprise` so that we'll be able to filter out those tests whenever we'll run Cypress in TE.
But I'm good as is. It's only something we should aim for so it's easier to implement TE and EE tests in the future.

saturninoabril (approved):
Thanks @josephbaylon!

PR #6806, opened 2020-10-15
[MM-28218] Edit Company Information screen

devinbinnie:
#### Summary
Edit Company information screen.

You should be able to review the look and feel of the Edit screen, but no information will show up, and you won't be able to Save. That review will need to come later.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28218

#### Related Pull Requests
Enterprise: https://github.com/mattermost/enterprise/pull/802
CWS: https://github.com/mattermost/customer-web-server/pull/253
Redux: https://github.com/mattermost/mattermost-redux/pull/1259
Server:  https://github.com/mattermost/mattermost-server/pull/15974


devinbinnie:
@matthewbirtch @stevemudie Looks like I'll need to mock out some data to make sure it displays properly. Will ping when ready.

nickmisasi (approved):

devinbinnie:
@matthewbirtch @stevemudie I've added some mock data which should allow you to see what the screens look like, but they won't be functional. We'll have to verify the rest of the functionality once we have a hook into the CWS.

matthewbirtch (changes requested):
Looks great @devinbinnie. 
1. If we don't have any billing address and I check 'Same as Billing Address', the billing address below shows as 'undefined, undefined, undefined' like the screenshot below. If we don't have a billing address, I suggest we don't show the 'Same as Billing Address' checkbox and show the fields for the address.
image

2. Can we make the state/postal code inputs span the same width as the inputs above?
image

3. For the hover state of the edit button, can we do this?
image
```
color: rgba(var(--sys-center-channel-color-rgb),.72);
background: rgba(var(--sys-center-channel-color-rgb),.08);
border-radius: 4px;
```
- The height and width of the button should be 28px x 28px as well
- Also, we'll want to apply this to the credit card payment edit button too

components/admin_console/billing/company_info_edit.tsx
```diff
@@ -28,16 +200,62 @@ const CompanyInfoEdit: React.FC<Props> = () => {
...
+                                    <input
+                                        type='checkbox'
+                                        checked={sameAsBillingAddress}
+                                        onChange={(event) => setSameAsBillingAddress(event.target.checked)}
+                                    />
+                                    <FormattedMessage
+                                        id='admin.billing.company_info_edit.sameAsBillingAddress'
+                                        defaultMessage='Same as Billing Address'
+                                    />
+                                </label>
```
marianunez, line 247: Should we show `Same as Billing Address` only if they have a billing address? If it's empty this option will not have anything to save 

Screen Shot 2020-10-16 at 12 36 13 PM

stevemudie (approved):
LGTM

components/admin_console/billing/company_info_edit.tsx
```diff
@@ -1,17 +1,186 @@
...
+                state: '',
+            }));
+        } else {
+            await dispatch(updateCloudCustomerAddress({
+                city: city || '',
+                country: country || '',
+                line1: address || '',
+                line2: address2 || '',
+                postal_code: postalCode || '',
+                state: state || '',
```
marianunez, line 102: Are we checking that these fields are required? I believe they all should be expect `line2`

devinbinnie: They should all be required and we're checking as part of the validation step, but because their definition in `useState` has them potentially null (since both `billing_address` and `company_address` can potentially be null), each of those types are `string | undefined` and need to be null checked. So this is more done for the sake of TypeScript.

components/admin_console/billing/company_info_edit.tsx
```diff
@@ -28,16 +197,64 @@ const CompanyInfoEdit: React.FC<Props> = () => {
...
+                                    name='companyName'
+                                    type='text'
+                                    value={companyName}
+                                    onChange={updateState(setCompanyName)}
+                                    placeholder={Utils.localizeMessage('admin.billing.company_info.companyName', 'Company name')}
+                                    required={true}
+                                />
+                            </div>
+                            <div className='form-row'>
+                                <Input
```
marianunez, line 219: @matthewbirtch Should we hide the `numEmployees` tag under the name until they have added something? Doesn't seem to provide any value showing `0 employees` there from the start

marianunez (approved):
Thanks @devinbinnie! Just have a non-blocking comment below

components/admin_console/billing/company_info_edit.tsx
devinbinnie: Makes sense to me.

matthewbirtch: yep, makes sense to me too

matthewbirtch (approved):
Looks good!

components/admin_console/billing/company_info_edit.tsx
marianunez: @devinbinnie could we include as part of this PR?

PR #6809, opened 2020-10-16
Migrate changeCSS() use of sidebarBg to CSS variable in utils/utils.jsx

ng29:
Migrate changeCSS() to CSS variable: --sidebar-bg

#### Summary
Migrate changeCSS() to CSS variable in utils/utils.jsx, ln. 504.
Change use of js variable heme.sidebarBg to css variable --sidebar-bg.

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/15826
Jira https://mattermost.atlassian.net/browse/MM-29387

### DIFF
--- a/sass/responsive/_mobile.scss
+++ b/sass/responsive/_mobile.scss
@@ -16,17 +16,18 @@
                 border: none;
             }
         }
-        .modal{
-            .settings-modal{
-                &.display--content{
-                    .modal-content{
-                        background: var(--sidebar-bg);
-                    }
-                }
-            }
+     .modal {
+         .settings-modal {
+             &:not(.settings-modal--tables) {
+                 &:not(.display--content) {
+                     .modal-content {
+                         background: var(--sidebar-bg);
+                     }
+                 }
+             }
+         }
         }
     }
-
     .post-right__container {
         height: 100%;
 
--- a/sass/responsive/_mobile.scss
+++ b/sass/responsive/_mobile.scss
@@ -2182,6 +2182,7 @@
                 &.display--content {
                     .modal-body {
                         max-height: 90%;
+                        background: var(--sidebar-bg);
                     }
                 }
 
--- a/sass/responsive/_mobile.scss
+++ b/sass/responsive/_mobile.scss
@@ -16,6 +16,15 @@
                 border: none;
             }
         }
+        .modal{
+            .settings-modal{
+                &.display--content{
+                    .modal-content{
+                        background: var(--sidebar-bg);
+                    }
+                }
+            }
+        }
     }
 
     .post-right__container {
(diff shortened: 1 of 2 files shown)


asaadmahmood (changes requested):
Thanks for the contribution @ng29.
One thing, the rule applied in the css file is on the 480px media query, whereas it was 768 in utils file.

Plus the rules are a bit more elaborate excluding specific classes as well.

ng29:
> Thanks for the contribution @ng29.
> One thing, the rule applied in the css file is on the 480px media query, whereas it was 768 in utils file.
> 
> Plus the rules are a bit more elaborate excluding specific classes as well.

Hey @asaadmahmood 

Could you relate it more specifically as i just started CSS and unable to find 480px in the mobile.scss file ??

asaadmahmood:
@ng29 The css you've included in inside: `@media screen and (max-width: 480px) {` tag.
And since the media query used in the utils file pointed to 768px, we should add this rule inside the `@media screen and (max-width: 768px) {` wrapper.

ng29:
> @ng29 The css you've included in inside: `@media screen and (max-width: 480px) {` tag.
> And since the media query used in the utils file pointed to 768px, we should add this rule inside the `@media screen and (max-width: 768px) {` wrapper.

Ahh,  mah bad @asaadmahmood now i get the correct position and just cross checking that it should be implemented in line 409 block, correct me if i am wrong? 

and one more doubt that there is not such hierarchy as i implemented previously from util

asaadmahmood:
@ng29 You should push a commit, and define that same hierarchy in the css file if its not already available in the 768px rule.

ng29:
> @ng29 You should push a commit, and define that same hierarchy in the css file if its not already available in the 768px rule.

Thanks for the info @asaadmahmood , i have made the changes as per my understanding in the 768px rule. 

Kindly review and merge if no changes are required.

asaadmahmood (changes requested):
Based on the code in the utils file.
I think the code should be:

```
        .modal {
            .settings-modal {
                &:not(.settings-modal--tables) {
                    &:not(.display--content) {
                        .modal-content {

                        }
                    }
                }
            }
        }
```

ng29:
> Based on the code in the utils file.
> I think the code should be:
> 
> ```
>         .modal {
>             .settings-modal {
>                 &:not(.settings-modal--tables) {
>                     &:not(.display--content) {
>                         .modal-content {
> 
>                         }
>                     }
>                 }
>             }
>         }
> ```

Got it @asaadmahmood , thanks for correcting. I have made the requested changes kindly review again 👍

asaadmahmood (approved):
Looks good, but I think there's an extra bracket. Just verify.

ng29:
> Looks good, but I think there's an extra bracket. Just verify.

I reckon thats closure of app body, so no extra ones

asaadmahmood (approved):

sudheerDev (approved):
LGTM @ng29 Thank you

PR #6813, opened 2020-10-16
[MM-26334] Fixes direct messages loading screen misalignment

ZombiMigz:
#### Summary

  Reduces the height of the loading screen so it is more centered and/or unintrusive(small screens) enough to remove the scroll bar during loading.

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/15926

#### Screenshots
Capture
Capture2

This is my first time working on a big project like this so please tell me if I did anything wrong or could improve the PR. tysm


calebroseland:
Hi @ZombiMigz, thanks for the PR and the work on this! 

I think the root of the problem may be upstream a bit. 

`multiselect_list` is used in a lot of these `.more-modal`s, and I think fixing there might be the way to go (See first two here for what I think is most of it)

https://github.com/mattermost/mattermost-webapp/blob/master/components/searchable_channel_list.jsx#L150
```diff
-    listContent = <LoadingScreen style={{marginTop: '50%'}}/>;
+    listContent = <LoadingScreen/>;
```

https://github.com/mattermost/mattermost-webapp/blob/master/components/multiselect/multiselect_list.tsx#L167-L168
```diff
<LoadingScreen
-    position='absolute'
```
(^ this seems to be the main culprit)

https://github.com/mattermost/mattermost-webapp/blob/master/sass/components/_modal.scss#L593
```diff
+   .more-modal .loading-screen{
+        padding: 1.9em 1em;
+        font-size: 1.25em;
+        height: auto;
+   }
```
(One possibility: this styling places 'loading' next to the no results message, preventing the top-center-top jumping that currently occurs going between these two messages)

(Otherwise, to keep centered, probably a `height: 100%` on the parent `div` above `<LoadingScreen/>` would do.)

calebroseland (commented):
Hi @ZombiMigz, thanks for the PR and the work on this! 

I think the root of the problem may be upstream a bit. 

`multiselect_list` is used in a lot of these `.more-modal`s, and I think fixing there might be the way to go (See first two here for what I think is most of it)

https://github.com/mattermost/mattermost-webapp/blob/master/components/searchable_channel_list.jsx#L150
```diff
-    listContent = <LoadingScreen style={{marginTop: '50%'}}/>;
+    listContent = <LoadingScreen/>;
```

https://github.com/mattermost/mattermost-webapp/blob/master/components/multiselect/multiselect_list.tsx#L167-L168
```diff
<LoadingScreen
-    position='absolute'
```
(^ this seems to be the main culprit)

https://github.com/mattermost/mattermost-webapp/blob/master/sass/components/_modal.scss#L593
```diff
+   .more-modal .loading-screen{
+        padding: 1.9em 1em;
+        font-size: 1.25em;
+        height: auto;
+   }
```
(One possibility: this styling places 'loading' next to the no results message, preventing the top-center-top jumping that currently occurs going between these two messages)

(Otherwise, to keep centered, probably a `height: 100%` on the parent `div` above `<LoadingScreen/>` would do.)

ZombiMigz:
@calebroseland Thank you so much for the response. Do you have a preference on whether the `<LoadingScreen/>` is centered or aligned to the top? If you would like it to be centered, I can't seem to get `<LoadingScreen/>` to inherit `height: 100%` from the multiselect_list. Any ideas or advice would be amazing.

image
(`<LoadingScreen/>` I'm trying to fix the height of this. Once the height is fixed it should center properly)

image
(parent element)

image
(grandparent element. I don't know why but it seems to be used to calculate `height: 100%` using this rather than the parent)

If it doesn't need to be centered I'll implement the other changes you suggested and that should be it. Thanks again!

calebroseland:
@ZombiMigz my preference would be to see 'Loading...' move to the top to eliminate the top-center-top jumping (2/5). @esethna / @michaelgamble what do you think?

calebroseland:
@ZombiMigz `<LoadingScreen/>` in the DM more modal here has that `position: absolute` that I mentioned earlier, that causes it to defer to the nearest `position: relative` parent (`.modal-body`) for its `height: 100%` and other things.

esethna:
@calebroseland agree with it moving to the top ^

calebroseland:
@ZombiMigz do you need any help with anything? If you use the css/changes I suggested a while back, that should bring this close to where it needs to be. (Might still need to update unit testing snapshots.)

ZombiMigz:
@calebroseland Ah sorry it seems I forgot to push changes. I believe everything should be good now, just put in the changes you suggested.

josephbaylon:
@ZombiMigz Please fix failing test. Thanks!

jasonblais:
@ashishbhate @calebroseland ready for your reviews

josephbaylon (approved):
LGTM. Verified Loading indicator is now moved to top center.

calebroseland (approved):
LGTM, thanks @ZombiMigz!

ashishbhate (approved):

ashishbhate:
Thanks @ZombiMigz !

PR #6815, opened 2020-10-16
MM-29082: SuggestionBox refactorization

lantrungseo:
## Summary

This PR refactor SuggestionBox implementation.
Tasks:
- [x] Migrate to __functional components__ & __React Hooks__.
- [x] Maintain old API in `QuickInput` & `SuggestionBox` components
 + Keep the old props and state.
 + All methods accessed via `refs` of `PureComponent` is now replaced by `forwardRef` and `useImperativeHandle`
- [x] Introduce `QuickInputState`: a class that control `QuickInput` current data (`value`, `selectionStart`, `selectionEnd`, `isFocused`,...).  It allows `SuggestionBox` to retrieve and manipulate the state of `QuickInput` without passing a `ref` to it, thus reduce the abusal of `ref` and make logic in `SuggestionBox` pure and clean. This solution resembles the one used in [Facebook DraftJS](https://github.com/facebook/draft-js).
- [x] Extract `SuggestionDate` component from `SuggestionBox`.
- [ ] Write tests
## Ticket Link

Jira ticket: [MM-20982](https://mattermost.atlassian.net/browse/MM-29082)
Github issue: Fixes https://github.com/mattermost/mattermost-server/issues/15855

## Note:
Please squash the commits before merging this PR.


components/quick_input/quick_input_state.ts
```diff
@@ -0,0 +1,45 @@
...
+        this.textboxState.value = initialState.value || defaultTextboxState.value;
+        this.textboxState.selectionStart = initialState.selectionStart || defaultTextboxState.selectionStart;
+        this.textboxState.selectionEnd = initialState.selectionEnd || defaultTextboxState.selectionEnd;
+        this.textboxState.isFocused = initialState.isFocused || defaultTextboxState.isFocused;
+    }
+
+    static createNew(initialState: Partial<ITextboxState> = defaultTextboxState){
+        return new QuickInputState(initialState);
+    }
+    static replaceNewState(initialState: Partial<ITextboxState> = defaultTextboxState){
```
nevyangelova, line 27: I don't see this being used anywhere.

components/quick_input/clearable_tooltip.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import {FormattedMessage} from 'react-intl';
+import {Tooltip} from 'react-bootstrap';
+import OverlayTrigger from 'components/overlay_trigger';
+import Constants from 'utils/constants.jsx';
+
+interface IClearableTooltipProps{
```
nevyangelova, line 7: nit: empty space between the name and bracket, I saw this in several places, would you be able to use our eslint configuration?

components/quick_input/clearable_tooltip.tsx
```diff
@@ -0,0 +1,55 @@
...
+        }
+        return (
+            <FormattedMessage
+                id={'input.clear'}
+                defaultMessage='Clear'
+            />
+        )
+    }, [props.clearableTooltipText]);
+
+    if(props.clearable && props.value && props.onClear){
```
nevyangelova, line 27: same here.

components/quick_input/clearable_tooltip.tsx
```diff
@@ -0,0 +1,55 @@
...
+                    <i className='icon icon-close-circle'/>
+                </span>
+            </OverlayTrigger>
+        </div>
+        )
+    }
+    return null;
+};
+
+export default ClearableTooltip;
```
nevyangelova, line 55: nit: new line

components/quick_input/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+export default QuickInput;
+
+export {QuickInputState};
+
+QuickInput.defaultProps = {
+    delayInputUpdate: false,
+    value: '',
+    clearable: false,
+    quickInputState: QuickInputState.createNew()
+};
```
nevyangelova, line 116: nit: new line

components/quick_input/use_ref_handler.ts
```diff
@@ -0,0 +1,23 @@
+import {useImperativeHandle} from 'react';
+
+export const useRefHandler = (
+    // reference pass from parents
+    ref: ((instance: unknown) => void) | React.MutableRefObject<unknown> | null,
+    inputRef: React.MutableRefObject<HTMLInputElement | undefined>
+) => {
+    useImperativeHandle(ref, () => ({
```
nevyangelova, line 8: I never considered using this, but I like the approach so far 👏

nevyangelova (commented):
Thanks @lantrungseo looks really promising to me on first iteration 👏 🎉 is there a reason why this is still a draft PR or we can proceed to a more in-depth review and QA?

lantrungseo:
@nevyangelova hi there, sorry it took me so long since I've been busy a lot.  There are still one issue I need to resolve to fully finish this ticket. 
[GH-15855](https://github.com/mattermost/mattermost-server/issues/15855)

Here it suggests splitting `SuggestionBox ` into 2 components for separate purposes: one for rendering users, messages,... list, another specifically for date suggestion. But since the logic in both component is basically takes after each other, and as I considered moving to **functional components & hooks**, I could not and would not recommend using inheritance here.

I suggest delegating `SuggestionDate` and `SuggestionList` to the scope of `Providers`, which means we only needs to pass `providers` as a props in `SuggestionBox`, and it will automatically render the proper suggestion list component for us. What do you think?

P/S: Sorry but I haven't linting the code yet. I'll fix it later. 
P/S 2: I suggest using husky to create pre-commit hook to prettify and lint-fixing the code.

jasonblais:
Hi @nevyangelova, what do you think about the above suggestion from @lantrungseo?

nevyangelova:
Hey @jasonblais sorry, me and @lantrungseo discussed this in chat already but I forgot to add the comment in the PR. He will proceed with the suggestions.

jasonblais:
@lantrungseo when you're ready for the pull request to be reviewed, let us know :)

lantrungseo:
@jasonblais oh sorry, pls review! Usually I have the habit of drafting the PR for development, after that I will un-draft it to be ready for review.

jasonblais:
Thank you @lantrungseo! It looks like there are a few build failures. Would you be open to reviewing and resolving them, after which we'll start reviews? If you need any help, don't hesitate to let us know.

lantrungseo:
@jasonblais  there are 2 test suites and 3 test case failed, all non-related to my current code? Is that normal? Please check the log in CircleCI

jasonblais:
Thank you for the heads up. @hmhealey or @nevyangelova would you be open to checking the logs on why test test suites are failing, unrelated to the code changes?

lantrungseo:
> Thank you for the heads up. @hmhealey or @nevyangelova would you be open to checking the logs on why test test suites are failing, unrelated to the code changes?
Hold on, please wait for the latest commit. I think I found the issue behind the test case. Thanks a lot =))

lantrungseo:
@jasonblais please start the review. I have fixed all the issues in the build & test CI

components/quick_input/clearable_tooltip.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+interface IClearableTooltipProps{
+    clearableTooltipText?: string | JSX.Element;
+    clearable?: boolean;
+    onClear?: (e: React.MouseEvent) => void;
+    value: string;
+}
+
+const ClearableTooltip = (props: IClearableTooltipProps) => {
+    const clearableTooltipText = React.useMemo(() => {
```
nevyangelova, line 18: Do we need this hook for the tooltip component?

components/quick_input/quick_input_state.ts
```diff
@@ -0,0 +1,48 @@
...
+    selectionEnd: number | null;
+}
+
+const defaultTextboxState: ITextboxState = {
+    value: '',
+    selectionEnd: null,
+    selectionStart: null,
+};
+
+export class QuickInputState {
```
nevyangelova, line 15: Looking at it, I think we might be able to get rid of Quick Input all together and all associated state with it. I think it was used for a bug in an old version of Microsoft Edge that we no longer support. Would you be open to replace it with a HTML input instead?

components/quick_input/quick_input_textbox.tsx
```diff
@@ -0,0 +1,109 @@
...
+        {
+            ...remainedProps,
+            defaultValue: defaultValueRef.current,
+            ref,
+
+            // add value as props (controlled components)
+            value: quickInputState?.getValue() || value,
+
+            // override default onChange event
+            onChange: (e: React.ChangeEvent<HTMLInputElement | any>) => {
```
nevyangelova, line 61: You could pass the whole event target and filter the props needed in the quickinput state.replace

components/quick_input/quick_input_textbox.tsx
```diff
@@ -0,0 +1,109 @@
...
+                }
+
+                //old API
+                if (onCompositionUpdate) {
+                    onCompositionUpdate(e);
+                }
+            },
+
+            // override onSelect event
+            onSelect: (e: any) => {
```
nevyangelova, line 91: > You could pass the whole event target and filter the props needed in the quickinput state.replace

same here

components/suggestion/suggestion_box/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+    const preventSuggestionListCloseFlag = React.useRef<boolean>(false);
+    const timeoutIdRef = React.useRef<NodeJS.Timeout >();
+
+    const suggestionHandlerRef = useSuggestionHandler(
+        pretextRef, timeoutIdRef, props, suggestionBoxState,
+        updateSuggestionBoxState, setQInputState, qInputModifier, qInputRefHandler,
+    );
+
+    /*eslint max-nested-callbacks: [0]*/
+    React.useEffect(() => {
```
nevyangelova, line 55: Seems like the state setting is convoluted, perhaps replacing with the HTML input will help in this case too?

components/suggestion/suggestion_box/use_container_ref.ts
```diff
@@ -0,0 +1,101 @@
...
+import React from 'react';
+
+import {QuickInputState} from '../../quick_input';
+
+import * as UserAgent from 'utils/user_agent';
+
+import {ISuggestionBoxProps, ISuggestionBoxState} from './types';
+import {SuggestionHandler} from './use_suggestion_handler';
+
+export const useContainerRef = (
```
nevyangelova, line 12: Ideally we would move away from using refs as SuggestionBox heavily relies on them at the moment. Handling focus with the SuggestionBox is weird, and we have to use focusin/focusout to get access to e.relatedTarget. React 17 is going to be switching the event handlers it uses internally, so we won't need the refs once that comes out. But if we can remove the need for them all together it will be great :)

components/suggestion/suggestion_box/use_suggestion_handler.ts
```diff
@@ -0,0 +1,269 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import React, {useMemo} from 'react';
```
nevyangelova, line 3: I like having separate handlers it makes it readable and testable, I wonder if we can go even further in detecting unnecessary functionality and stripping it?

components/quick_input/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+    quickInputState?: QuickInputState;
+    onTextboxChange?: (qInputState: QuickInputState) => void;
+    onTextboxCompositionUpdate?: (
+        qInputState: QuickInputState, composingData: string | null | undefined
+    ) => void;
+    onTextboxSelect?: (qInputState: QuickInputState) => void;
+}
+
+const QuickInput = React.forwardRef((props: IQuickInputProps, ref) => {
+    //reference to quick input text box
```
nevyangelova, line 65: Our convention for /index.tsx files is to connect the component to redux but not render it within

nevyangelova (changes requested):
Thanks @lantrungseo for taking this challenging task and providing a creative solution. Overall it’s a big change in a good direction. Some of the logic is nicely separated in the specific hooks. I know I requested some big changes on here, let me know if this is OK with you and if you need any more guidance. Also adding some unit tests would be a great bonus 🎉

components/quick_input/clearable_tooltip.tsx
lantrungseo: I suggest splitting it into separate components. The quick input component files is confusing enough =))

components/quick_input/quick_input_state.ts
lantrungseo: Which means the component quick input is the same, but we do not support rendering custom textbox? Just rendering input only?

lantrungseo: That could be a challenge to remove `QuickInput`, since some of `SuggestionBox` in the webapp render `QuickInput` with customized components such as `AutosizedTextarea`?

components/quick_input/quick_input_textbox.tsx
lantrungseo: I don't think that is recommendable. The point of event handler is to handle new state changed of the component (for`QuickInput` is the `QuickInputState`, or for `input` is the `React.ChangeEvent`). Passing the whole event target to a component controlled by `QuickInputState` is not really logical, I think.

components/quick_input/quick_input_textbox.tsx
lantrungseo: Yeah, same above.

components/suggestion/suggestion_box/index.tsx
lantrungseo: Can you specify here?

components/quick_input/index.tsx
lantrungseo: Oh ok, this can be solved easily.

components/suggestion/suggestion_box/use_suggestion_handler.ts
lantrungseo: Well this can be solved.

components/suggestion/suggestion_box/use_container_ref.ts
lantrungseo: Lemme check about this.

components/quick_input/clearable_tooltip.tsx
nevyangelova: Sounds good 👍

components/suggestion/suggestion_box/index.tsx
nevyangelova: I don't really have a concrete suggestion but would be nice to reduce the amount of callbacks. Also I was referring to getting rid of QuickInput here as well

components/quick_input/quick_input_state.ts
nevyangelova: Good question! I think this is one of the functionalities that QuickInput shouldn't be responsible for tbh as we already handle some of the logic in Textbox.tsx. @hmhealey can you advice if it's safe to remove QuickInput?

lantrungseo: So if it's possible I would like to:

1, Remove `QuickInput` from `SuggestionBox` (please advise whether it's safe or not).
2, Remove logic in `QuickInput` that allows rendering custom textbox rather than `input`? (Again, please advise whether it's safe to do so).

Question: Do I need to keep `QuickInputState` as a controlled state of `QuickInput` ? (Personally I like that approach better than passing `onChange(targetEvent)` directly. This practice is used by `Draft-js` of Facebook as well!!)

Suggestion: I would like to convert `SuggestionBox` to controlled component using a controlled state class `SuggestionBoxState`, where it controlled `value`, `suggestions`, `terms` of `SuggestionBox`. With this approach I think we can safely reduce the impact scope of `Providers`. Please lemme know about your opinions.

nevyangelova: Sounds good to me! It should be safe to remove it as quick input was only implemented for a specific bug in an old version in Microsoft edge. If we run into issues while implementing we can reconsider the approach. 

Regarding quickInputState I don't think it will be necessary as we won't need to store any input specific props? IMO the components state can be handled in SuggestionBox. 

I like the suggestion you can probably store the input/textbox specific props there too.

lantrungseo: Okay, so far I need to render `inputComponent` inside `SuggestionBox` without going through `QuickInput`, right? This is managable.

components/quick_input/quick_input_textbox.tsx
lantrungseo: Okay, I think I should revert `QuickInput` to its original API, since `SuggestionBox` is on its way of getting rid of `QuickInput` as its child. Refactoring `QuickInput` should be on a separate PR.

components/quick_input/quick_input_state.ts
nevyangelova: Correct. And you can go ahead and removing the state as well :)

nevyangelova:
Hey @lantrungseo let us know if everything was clear or you need additional clarification. Thanks for your amazing help 🎉

jasonblais:
@lantrungseo Thanks for your contribution! Let us know if we can help with the above feedback, would be happy to :

nevyangelova:
Closing this as a stale PR

PR #6816, opened 2020-10-16
[MM-28802] Limit team and full name length in getting started

ethervoid:
#### Summary

This PR adds length limits to the full name and team name fields in the getting started module that matches the length we have defined in the models for those entities: 64 for team name and 128 for the user's full name

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28802

#### Related Pull Requests

#### Screenshots

Peek 2020-10-16 12-17
Peek 2020-10-16 12-07


matthewbirtch:
@ethervoid the screenshots look good, but I'm not getting the 'Getting Started' screen when use the SpinWick link above (for admin or end user) so I can't test it

ethervoid:
> but I'm not getting the 'Getting Started' screen when using the SpinWick link above (for admin or end-user) so I can't test it

I'll update the preferences for the admin in order to allow you to see the getting started part

> Also, we should create a follow-up ticket for the name limit. 128 seems really long. Any reason we couldn't shorten it to 64 characters as well?

In Mattermost we have a limit of 64 for the name and 64 for the surname so in this case, we're merging both of them so to keep consistency we should allow 128

devinbinnie (approved):

ethervoid:
@matthewbirtch the test server is ready and the changes in the error messages too

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
```diff
@@ -64,7 +66,10 @@ export default class CompleteProfileStep extends React.PureComponent<Props, Stat
     private handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         let fullNameError;
         if (!event.target.value) {
-            fullNameError = Utils.localizeMessage('next_steps_view.complete_profile_step.nameCannotBeBlank', 'Your name can’t be blank');
+            fullNameError = Utils.localizeMessage('next_steps_view.complete_profile_step.fullNameCannotBeBlank', 'Your full name cannot be blank');
+        }
+        if (event.target.value.length > MAX_FULL_NAME_LENGTH) {
```
marianunez, line 71: Should we either null check here or do and `else if` here to avoid a null reference error?
```suggestion
        if (event.target.value?.length > MAX_FULL_NAME_LENGTH) {
```

components/next_steps_view/steps/team_profile_step/team_profile_step.tsx
```diff
@@ -64,7 +66,10 @@ export default class TeamProfileStep extends React.PureComponent<Props, State> {
     private handleNameInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         let teamNameError;
         if (!event.target.value) {
-            teamNameError = Utils.localizeMessage('next_steps_view.team_profile_step.nameCannotBeBlank', 'Team name can’t be blank');
+            teamNameError = Utils.localizeMessage('next_steps_view.team_profile_step.nameCannotBeBlank', 'Team name cannot be blank');
+        }
+        if (event.target.value.length > MAX_TEAM_NAME_LENGTH) {
```
marianunez, line 71: Same as above 

```suggestion
        if (event.target.value?.length > MAX_TEAM_NAME_LENGTH) {
```

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
```diff
@@ -19,6 +19,8 @@ import {StepComponentProps} from '../../steps';
 
 import './complete_profile_step.scss';
 
+const MAX_FULL_NAME_LENGTH = 128;
```
marianunez, line 22: Should we also limit each first/last string to 64 individually when it's split out (line 88)?

marianunez (changes requested):

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
ethervoid: Could be confusing given that we're using just one input

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
ethervoid: You're right. I'll use the else if

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
marianunez: What is the server behavior if over 64 chars the first/last name? If they input something where the first name is 100 chars and the last name 15, the split will still send the first name too long and will behave the same as the original report for team. 

Maybe we can set a limit when we trim the first/last name by just taking the first 64 of each?

ethervoid: Changed. Now is validating first and last name individually

matthewbirtch:
@ethervoid a couple comments/questions:

1. shouldn't the team name be truncated in the sidebar? I'm getting this when I update the team name:
image

2. for the name, the error messages shows "Your first name must be less than 65 characters"
- this is confusing since it's just one 'name' field, not two separate first and last name fields
- Also, how do we know it's a first or last name? Sometimes a first name has a space in it or sometimes a last name has a space in it, so splitting the name up based on the space character might not always work
image

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
marianunez: > 2. for the name, the error messages shows "Your first name must be less than 65 characters"
> 
> * this is confusing since it's just one 'name' field, not two separate first and last name fields
> * Also, how do we know it's a first or last name? Sometimes a first name has a space in it or sometimes a last name has a space in it, so splitting the name up based on the space character might not always work
>   image

We are currently taking everything after the first space as the last name. Given the nature of the `Full Name` field we have to make some assumptions. 

I agree that this validation it's confusing and I suggest that we keep the full length restriction as originally implemented and silently trim the first/last names if they exceed 64 chars. 

Thoughts?

marianunez: > 2. for the name, the error messages shows "Your first name must be less than 65 characters"
> 
> * this is confusing since it's just one 'name' field, not two separate first and last name fields
> * Also, how do we know it's a first or last name? Sometimes a first name has a space in it or sometimes a last name has a space in it, so splitting the name up based on the space character might not always work
>   image

We are currently taking everything after the first space as the last name. Given the nature of the `Full Name` field we have to make some assumptions. 

I agree that this validation it's confusing and I suggest that we keep the full length restriction as originally implemented and silently trim the first/last names if they exceed 64 chars. 

Thoughts?

marianunez:
> 2. for the name, the error messages shows "Your first name must be less than 65 characters"
> 
> * this is confusing since it's just one 'name' field, not two separate first and last name fields
> * Also, how do we know it's a first or last name? Sometimes a first name has a space in it or sometimes a last name has a space in it, so splitting the name up based on the space character might not always work
>   image

We are currently taking everything after the first space as the last name. Given the nature of the `Full Name` field we have to make some assumptions. 

I agree that this validation it's confusing and I suggest that we keep the full length restriction as originally implemented and silently trim the first/last names if they exceed 64 chars. 

Thoughts?

components/next_steps_view/steps/complete_profile_step/complete_profile_step.tsx
marianunez: See my comment below. I was suggesting trimming when setting the values and not change validation of the full name.

ethervoid:
Trimming could be confusing as well I rather prefer to inform about a problem with the name than silently remove part of with. What about having two inputs as we have in MM?

ethervoid:
Ok I'll do that then

ethervoid:
Changes are done. @matthewbirtch I've also made the change to the team name in the sidebar. Hopefully, it will not have any side-effect, I've check where are we using that CSS class and is only there:

Screenshot from 2020-10-19 21-27-12

matthewbirtch (approved):
Thanks @ethervoid. Looks good

marianunez (approved):
Thanks @ethervoid!

stevemudie (commented):
This tested fine except I noticed that if you go in to edit the team name you're limited to 15 characters. Should this be addressed in this PR or a new one?

As an aside, the truncated team name at the top left works but as a user I'd expect it to show the full name in a tooltip when hovering over it. Should I write up a story ticket for that?

ethervoid:
> Should I write up a story ticket for that?
IMHO we should move that to another ticket with the change for the team input from 15 to 64 which could involve more than just change that

stevemudie (approved):
Worked as expected!

PR #6821, opened 2020-10-16
[MM-29689, MM-29690, MM-29691, MM-29692] Deprecate calls to changeCSS()

rohanrk:
#### Summary

Deprecates changeCSS() calls. I've tested a bit with the channel text color and I cannot see any difference between before and after removing the css calls. I don't think it's necessary to put the calls in the scss file but I'll defer to a reviewers judgement on that. I have that commit ready just in case. 

I'll test more as well, but I don't see the class names in the inspector when searching.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15964
Fixes https://github.com/mattermost/mattermost-server/issues/15965
Fixes https://github.com/mattermost/mattermost-server/issues/15966
Fixes https://github.com/mattermost/mattermost-server/issues/15967


asaadmahmood (approved):

Willyfrog (approved):
LGTM! thanks!

PR #6853, opened 2020-10-19
[MM-24963] Fix inconsistent hover/tooltips UX

calebroseland:
#### Summary

A number of `<OverlayTrigger/>`s are inside their corresponding buttons/links. This can cause a discrepancy between the CSS hover state and activation of said trigger (some imperceivable, others noticeable, depending on the padding/styling). This PR inverts these so the buttons/links are inside `<OverlayTrigger/>`. 

- [MM-24963](https://mattermost.atlassian.net/browse/MM-24963) - Fix hover box/tooltip target for following interactive elements: 
  - Header
    - Channel Members
    - Help Menu/Dropdown
  - LHS/Sidebar (new)
    - Filter by unread
    - New Category
    - Category/Channel Menu
    - DMs: Sort By and New
  - Center Channel
    - Post Reaction (original in ticket)
  - RHS
    - Info Card "back to search results" button (when opened from search results)
- `<OverlayTrigger/>`: new `disabled` prop to disable overlay content from being displayed via `visibility: hidden` (prevents positional glitches vs. wholly removing from DOM)
    e.g. turning off the tooltip of a button that also has a popover/content overlay.
    The overriding `show=false` in `react-bootstrap@1.3+` could replace this if/when upgraded.


components/overlay_trigger.tsx
```diff
@@ -11,10 +11,16 @@ import store from 'stores/redux_store.jsx';
...
-type Props = OverlayTriggerProps;
+type Props = OverlayTriggerProps & {
+    disabled?: boolean;
+};
 
 const OverlayTrigger = React.forwardRef((props: Props, ref?: React.Ref<BaseOverlayTrigger>) => {
-    const {overlay, ...otherProps} = props;
+    const {
+        overlay,
+        disabled,
```
hmhealey, line 21: I'm slightly hesitant about adding new props to this component since it was supposed to be a transparent replacement for the one provided by react-bootstrap until we upgraded it (or removed it), but I suppose we could always keep our wrapper around it or just remove it in the future if necessary. This isn't a particularly complicated new prop to maintain anyway.

hmhealey (approved):

jfrerich (approved):
All of the inversions look good to me.

jgilliam17 (approved):
Thank you @calebroseland 
Tested, looks good to merge.
- Verified tooltip appears on hover as cursor changes to pointer. Verified the list of items listed in accompanying jira ticket.

components/overlay_trigger.tsx
calebroseland: Completely agree. We can eventually unwrap it and use `show=false`—I tried `OverlayTrigger` from 1.3 and it works as-is (though ref-less, it does forward contexts properly).

PR #6854, opened 2020-10-20
[MM-26503] Category names not outlining in Safari on channel drop

lestgabo:
#### Summary
- hover does not trigger in Safari while dragging
- just added similar class without the hover
- because the class only appears when a channel is "dragged" over

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15903
JIRA - https://mattermost.atlassian.net/browse/MM-26503

#### Related Pull Requests
n/a


hmhealey:
Thanks for the PR, @lestgabo. Unfortunately, this does change the behaviour slightly since the current version only highlights the category heading when you're actually hovering over it.
ezgif-2-9afa64db19c7
I didn't know that the reason this was broken on Safari is because they specifically don't support `:hover` styles while dragging. I suppose we could implement our own hovered state for those with JS, although that doesn't feel great since CSS is the correct way to do this and adding workarounds for one browser never feels good.

If you can find a simpler way to get it to only highlight the names when you're over them that would be great. I'll also see if anyone on the team has any ideas of how we should do this going forward.

lestgabo:
I spent some more time on this and just couldn't find a way for both conditions:
1. outline **only** on header hover
2. Safari to outline **only** on header hover

Since my PR cannot cover the 2 conditions above I am handing this over to you guys 8)

hmhealey:
Closing this in favour of a new PR since the branch seems to have gotten a bit mixed up.

PR #6858, opened 2020-10-20
MM-29609 Reuse UnreadChannelIndicator styling for both sidebars

hmhealey:
The styles on the Unread Channel indicator in the sidebar diverged between the old and new sidebar, so this makes them both the same. I also took the opportunity to move the component into its own folder with its own styles next to it and do some minor cleanup (using CSS variables and removing unnecessary mixins).

This is based on #6594, so the changes specific to it can be found here: https://github.com/mattermost/mattermost-webapp/compare/MM-27498...MM-29609_unread-indicator-styling

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29609

#### Screenshots
Before:
Screen Shot 2020-10-20 at 10 47 30 AM

After: 
Screen Shot 2020-10-20 at 10 47 00 AM


components/sidebar/unread_channel_indicator/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
devinbinnie, line 1: What purpose does this file serve?

devinbinnie (approved):

components/sidebar/unread_channel_indicator/index.ts
hmhealey: We had a discussion last week about moving components into their own folders when they started including multiple files like snapshots and tests and such, so I was doing that here. This was my first instinct to keep the import path simple for a non-connected component, but I'm 0/5 about whether the component lives in index in that case. This is kind of nice thoughsince it keeps the file named after the component instead of having to look for `unread_channel_indicator/index.tsx` in your editor at least.

amyblais:
@calebroseland Kind reminder to review for v5.29.

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -128,40 +140,63 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
...
-                            <ChannelComponent
-                                isCollapsed={this.isCollapsed(this.props)}
-                                channel={channel}
-                                currentTeamName={currentTeamName}
-                            />
-                        </li>
-                    );
-                }}
-            </Draggable>
+        const component = (
```
calebroseland, line 143:
```suggestion
        const channel = (
```
?
(feels more natural; not important, just a name)

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -128,40 +140,63 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
...
+            <ChannelComponent
+                isCollapsed={this.isCollapsed(this.props)}
+                channel={channel}
+                currentTeamName={currentTeamName}
+            />
         );
+
+        let wrappedComponent: React.ReactNode;
```
calebroseland, line 151: Does this need a variable? Would a return ternary work?

types/store/index.ts
```diff
@@ -47,7 +47,7 @@ export type GlobalState = BaseGlobalState & {
             };
             focusedPostId: string;
             mobileView: boolean;
-            keepChannelIdAsUnread: Channel | null;
+            lastUnreadChannel: (Channel & {hadMentions: boolean}) | null; // Actually only an object with {id: string, hadMentions: boolean}
```
calebroseland, line 50:
```suggestion
            lastUnreadChannel: (Pick<Channel, 'id'> & {hadMentions: boolean}) | null; // Actually only an object with {id: string, hadMentions: boolean}
```
If it's just the `id`?

calebroseland (approved):
LGTM, just a minor or two.

components/sidebar/sidebar_channel/sidebar_channel.tsx
hmhealey: It would feel a bit nicer, but that conflicts with `const channel = this.props.channel` above

components/sidebar/sidebar_channel/sidebar_channel.tsx
hmhealey: I'm not a fan of ternaries for something this big. We do early returns in some other places, but I was brought up in the land of Java and functions with only a single return, so I tend to favour that pattern myself.

types/store/index.ts
hmhealey: I tried changing the type on this to match the actual type, but it ran into some other issues because everything else expects `lastUnreadChannel` to be a full Channel, including some code in mattermost-redux, and I wanted to avoid having to make a bunch of changes to both repos to clean that up. We can address that more easily once they're eventually in the same repo.

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified style of More Unreads bubble on the new sidebar, matches mobile and old sidebar.

amyblais:
@hmhealey This needs to be cherry picked manually.

PR #6861, opened 2020-10-20
[MM-29468] Fix color picker to use the same one we use in custom theme picker on chat facing side

hahmadia:
#### Summary

Changes the colour picker on system console to  be same one as we use in the custom theme picker on chat facing side.
This not only fixes the bug in the ticket but also uses only one implementation of a colour picker rather than the 2 we currently have. 

In addition, I noticed some settings where `TEXT` when they should really be `COLOR` so I've updated those too.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29468

#### Screenshots

**DISABLED**
image

**ENABLED**
image


hmhealey (approved):

lindy65:
Hi @hahmadia - thanks, I've tested on the spinwick and, while the picker color issue is fixed, I'm seeing that I have to refresh the page before changes to banner text or picker colors are effected.  This was not previously the case.  Is this something that you can fix or should I create a separate ticket?

sbishel (approved):
Nice changes.

lindy65:
Hi @hahmadia - is this ready for re-testing based on my findings above?

hahmadia:
@lindy65 
I tried locally and on the above hosted server but I wasn't able to reproduce your issue? Everything seemed to work fine for me. Please retry and let me know if you still experience those problems? Thanks!

Gif

lindy65 (approved):
Hi @hahmadia - I retested on the updated spinwick and I am no longer able to reproduce the issues I previously saw. Looks good to me 👍

PR #6874, opened 2020-10-21
[MM-28363] Add asset for user limit overage emails

nickmisasi:
#### Summary
This asset is needed for an upcoming PR around user limit overage emails. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28363?filter=15333

#### Related Pull Requests
N/A (yet)

#### Screenshots

image


webpack.config.js
```diff
@@ -290,6 +290,7 @@ var config = {
                 {from: 'images/warning.png', to: 'images'},
                 {from: 'images/logo-email.png', to: 'images'},
                 {from: 'images/browser-icons', to: 'images/browser-icons'},
+                {from: 'images/cloud', to: 'images'},
```
marianunez, line 293: Should this be limited to the specific png we want?

devinbinnie, line 293: Does this break any of the other images we're using?

devinbinnie (approved):

webpack.config.js
nickmisasi: Thinking here was that any future assets could be added without messing with webpack. I don't think there's any significant negative impact, but I'm happy to change to be explicit - let me know what you think!

@devinbinnie (will continue your discussion here) This shouldn't break anything, simply marks it in webpack to be servedby URL /static/images/<png>

devinbinnie: Okay, as long as there's no explicit reference to the images (eg. in CSS files)

marianunez (approved):

webpack.config.js
nickmisasi: Since those are included via imports, they should be fine. I will be sure to give one last check before merging.

nickmisasi: Confirmed, good to go!

PR #6878, opened 2020-10-21
MM-28490 Increase the number of new posts received before the events being batched

hmhealey:
Jason reported an issue where some new DMs weren't causing the channel to become visible in the sidebar, and we think this is only happening to him because he's in so many channels, so he's more likely to run into the batched handling of new posts. The batched post handling skips some steps of receiving new posts such as counting mentions and unread messages, and because the DM is not becoming unread, it's not appearing in the sidebar.

We'd obviously like to improve the handling of multiple new posts at once, but for the time being, we're going to see if decreasing the chance that this logic triggers helps to begin with.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28490

### DIFF
--- a/actions/websocket_actions.jsx
+++ b/actions/websocket_actions.jsx
@@ -546,7 +546,7 @@ function debouncePostEvent(wait) {
     };
 
     return function fx(msg) {
-        if (timeout && count > 2) {
+        if (timeout && count > 4) {
             // If the timeout is going this is the second or further event so queue them up.
             if (queue.push(msg) > 200) {
                 // Don't run us out of memory, give up if the queue gets insane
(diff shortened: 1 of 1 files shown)


crspeller (approved):

nevyangelova (approved):

hmhealey:
@jgilliam17 I'm not sure what the best way to test this is since it previously required getting two posts within 100ms of each other to cause the bad behaviour. Maybe I'll try to force it with an E2E test since it might be difficult to test manually

jgilliam17:
Thank you @hmhealey 
I ran a quick test on the PR server, didn't see any missing DMs. 
I agree, using E2Es to test going forward is a good idea and checking after merge to see if it reproduces for Jason and other reporters. 
Will you be adding E2Es here on this PR before merge? Let me know. Thanks.

hmhealey:
I've spent the day trying to write an E2E test to trigger the problem behaviour, but I haven't been able to do it in Cypress. I did manage to do it by capturing the new post request and resending it 8 times in a row, but I can't get that to work in Cypress for some reason. I think our best bet is to just merge this and see if it fixes anything.

jgilliam17 (approved):
Thank you @hmhealey 
QA will re-test after merge.

PR #6879, opened 2020-10-21
[MM-28216] Hookups to back-end for Billing History section

devinbinnie:
#### Summary
This contains all the hookups to Redux to pull down billing history. It also contains a virtual paging implementation that can be converted once the server supports actual paging.

UX has already been reviewed by Matt.
Will need QA once hooked up to CWS.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28216


components/admin_console/billing/billing_history.tsx
```diff
@@ -94,62 +63,90 @@ const getPaymentStatus = (status: string) => {
...
+    };
 
     useEffect(() => {
+        dispatch(getCloudProducts());
+        dispatch(getCloudSubscription());
+        dispatch(getInvoices());
         pageVisited('cloud_admin', 'pageview_billing_history');
     }, []);
 
+    useEffect(() => {
```
ethervoid, line 123: Why use `useEffect` two times?

ethervoid (commented):
One question

components/admin_console/billing/billing_history.tsx
```diff
@@ -186,43 +183,59 @@ const BillingHistory: React.FC<Props> = () => {
...
+                                        values={{
+                                            fullUsers,
+                                            partialUsers,
+                                        }}
+                                    />
+                                </div>
+                            </td>
+                            <td className='BillingHistory__table-total'>
+                                <FormattedNumber
+                                    value={invoice.total}
```
ethervoid, line 218: Invoice total comes as an int where the last two digits are the decimal so we need

components/admin_console/billing/billing_history.tsx
devinbinnie: First `useEffect` only runs once as there are no dependent parameters (see the second argument), so it acts like `componentDidMount`.

Second `useEffect` updates other state when the two dependent parameters change.

ethervoid: Cool, good to know :)

ethervoid (approved):
LGTM!

components/admin_console/billing/billing_history.tsx
```diff
@@ -94,62 +63,90 @@ const getPaymentStatus = (status: string) => {
...
+        dispatch(getCloudProducts());
+        dispatch(getCloudSubscription());
+        dispatch(getInvoices());
         pageVisited('cloud_admin', 'pageview_billing_history');
     }, []);
 
+    useEffect(() => {
+        if (invoices && Object.values(invoices).length) {
+            const invoicesByDate = Object.values(invoices).sort((a, b) => b.period_start - a.period_start);
+            setBillingHistory(invoicesByDate.slice(firstRecord - 1, (firstRecord - 1) + PAGE_LENGTH));
```
marianunez, line 126: Wondering if making `firstRecord` zero-based would simplify some of these calculations and below when setting the paging indicators.

marianunez (approved):
Great work, thanks @devinbinnie!

components/admin_console/billing/billing_history.tsx
devinbinnie: We could, but then we'd need to do the same in reverse for displaying the record counts. Might reduce some code bloat but it'd be negligible. 0/5

marianunez: Agreed

PR #6888, opened 2020-10-22
[MM-29844] Fixed email input to allow multiple emails to be typed and trigger validation on more cases

devinbinnie:
#### Summary
When copying and pasting emails over the limit, the control wouldn't allow the email to be added, but would still disable the Send button, requiring the user to delete an email unnecessarily.

This PR allows the users to enter as many emails as they want, but will trigger validation on each addition and on typing to ensure that more than 10 cannot be sent out.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29844

### DIFF
--- a/components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
+++ b/components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
@@ -103,11 +103,6 @@ export default class InviteMembersStep extends React.PureComponent<Props, State>
             return;
         }
 
-        if (this.state.emails.length >= 10) {
-            this.setState({emailError: Utils.localizeMessage('next_steps_view.invite_members_step.tooManyEmails', 'Invitations are limited to 10 email addresses.')});
-            return;
-        }
-
         if (value.indexOf(' ') !== -1 || value.indexOf(',') !== -1) {
             const emails = value.split(/[\s,]+/).filter((email) => email.length).map((email) => ({label: email, value: email, error: !isEmail(email)}));
             const newEmails = [...this.state.emails, ...emails];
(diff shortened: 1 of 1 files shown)


stevemudie (approved):
Worked as expected!

nickmisasi (approved):

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
```diff
@@ -141,7 +136,8 @@ export default class InviteMembersStep extends React.PureComponent<Props, State>
     onBlur = () => {
         if (this.state.emailInput) {
             const emails = this.state.emailInput.split(/[\s,]+/).filter((email) => email.length).map((email) => ({label: email, value: email, error: !isEmail(email)}));
-            this.setState({emails: [...this.state.emails, ...emails], emailInput: '', emailError: undefined});
+            const newEmails = [...this.state.emails, ...emails];
+            this.setState({emails: newEmails, emailInput: '', emailError: newEmails.length > 10 ? Utils.localizeMessage('next_steps_view.invite_members_step.tooManyEmails', 'Invitations are limited to 10 email addresses.') : undefined});
```
marianunez, line 140: Actually, given that they already have the admin, it's should probably be limited to 9. Should we do the same we do in the invite modal and rely on the actual count of available users in the free tier?

devinbinnie: Seems kinda late to make this change to be honest, we should discuss it with the rest of the team first and I think it should be a separate ticket.

marianunez: Sure, could you create a follow up ticket?

marianunez (approved):

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
marianunez: https://mattermost.atlassian.net/browse/MM-29958

PR #6890, opened 2020-10-22
[MM-29914] Fixed the initial value of the country on company info edit

devinbinnie:
#### Summary
When editing company information where an existing company address was present, the country field wasn't being populated. This required the user to have to select the country a second time before validation allowed the Save button to be enabled.

This PR fixes the initial value of the country so that the Save button is enabled immediately when opening the screen, such that editing other non-required fields (ie. number of employees) to function without issue.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29914

### DIFF
--- a/components/admin_console/billing/company_info_edit.tsx
+++ b/components/admin_console/billing/company_info_edit.tsx
@@ -39,7 +39,7 @@ const CompanyInfoEdit: React.FC<Props> = () => {
     const [address2, setAddress2] = useState(companyInfo?.company_address?.line2);
     const [city, setCity] = useState(companyInfo?.company_address?.city);
     const [postalCode, setPostalCode] = useState(companyInfo?.company_address?.postal_code);
-    const [country, setCountry] = useState(getName(companyInfo?.company_address?.country || 'US'));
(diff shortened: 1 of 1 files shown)


ethervoid (approved):
LGTM

marianunez:
Devin, it seems the ticket linked is not related to this PR?

marianunez:
Nvm, got it!

marianunez (commented):
Would it be more intuitive to enable the Save button when there is a change in the num of employees field instead?

devinbinnie:
> Would it be more intuitive to enable the Save button when there is a change in the num of employees field instead?

It would actually require more logic to do that since it's not required for validation, so what you would be asking for is to enable the Save button based on if the existing information had changed. We can do that if you want, but this solves the issue in the same way without that extra complication.

1/5 on how we proceed.

marianunez:
> It would actually require more logic to do that since it's not required for validation, so what you would be asking for is to enable the Save button based on if the existing information had changed. We can do that if you want, but this solves the issue in the same way without that extra complication.

Sure, makes sense.

marianunez (approved):

PR #6908, opened 2020-10-23
[MM-28222][MM-28215] Payment Warnings

devinbinnie:
#### Summary
Banners to display when the credit card is about to expire, is expired and the most recent invoice failed.

NOTES:
- Used <10 days as the 'about to expire' metric, can be easily changed.
- Using the last invoice displayed on the Subscription screen to show if the most recent payment failed, not sure if it needs other logic.

Will need UX/QA review after merge.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28222
https://mattermost.atlassian.net/browse/MM-28215


components/admin_console/billing/billing_subscriptions.tsx
```diff
@@ -51,12 +52,27 @@ const BillingSubscriptions: React.FC<Props> = () => {
...
 
     const contactSalesLink = useSelector((state: GlobalState) => getCloudContactUsLink(state, InquiryType.Sales));
 
+    const isCreditCardExpired = () => {
+        if (!customer) {
+            return false;
+        }
+
+        // Will developers ever learn? :D
+        const expiryYear = customer.payment_method.exp_year + 2000;
```
marianunez, line 68: Hehehe, in defense of devs, last two digits of CC is probably standard

components/admin_console/billing/billing_subscriptions.tsx
```diff
@@ -51,12 +52,27 @@ const BillingSubscriptions: React.FC<Props> = () => {
...
+        const lastExpiryDate = new Date(expiryYear, customer.payment_method.exp_month - 1, 1);
+        lastExpiryDate.setMonth(lastExpiryDate.getMonth() + 1);
```
marianunez, line 70: Why the -1 for the month and then add it back?

components/admin_console/billing/billing_subscriptions.tsx
```diff
@@ -200,10 +249,25 @@ const BillingSubscriptions: React.FC<Props> = () => {
...
-                        className='BillingSubscriptions__topWrapper'
-                        style={{marginTop: '20px'}}
-                    >
+                    {showCreditCardBanner && (
+                        <AlertBanner
+                            mode='info'
+                            title={
+                                <FormattedMessage
+                                    id='admin.billing.payment_info.creditCardAboutToExpire'
+                                    defaultMessage='Your credit card is about to expire'
```
marianunez, line 258: Would this be `your credit card has expired` given that we are checking with today's date directly?

marianunez (approved):
Great work @devinbinnie! Minor comments below

components/admin_console/billing/billing_subscriptions.tsx
devinbinnie: You'd think we'd have learned after the whole Y2K fiasco :P

components/admin_console/billing/billing_subscriptions.tsx
devinbinnie: when you define a `Date` object, the month value starts at 0 for some reason, so instead of 1 being January, 0 is January. But we store January as 1, so that's why we need the -1.

But then we have to increment it to get the actual expiry date of the card, since the card is usually good until the end of the month. So if your card expiry day is October 2021, you should be able to use it until October 31st, 2021.

I know it's redundant, but it's clear in terms of what operations are happening. I can remove the redundancy if you'd like and just add a comment stating the above.

marianunez: Comment would be great. Thanks!

devinbinnie: Done! Also moved out the expired function to a utils file.

PR #6920, opened 2020-10-26
[MM-29903] - Check if current user is on last post when scrolling in rhs

nevyangelova:
#### Summary
Having the check for current user caused weird scrolling behaviour in RHS when another user is replying to the thread. The comment box would disappear out of sight. Removing the check seems to work for both cases.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29903

#### Screenshots
4199edfd-7e83-42d7-9163-6e0e2f744554

### DIFF
--- a/components/rhs_thread/rhs_thread.tsx
+++ b/components/rhs_thread/rhs_thread.tsx
@@ -128,7 +128,7 @@ export default class RhsThread extends React.Component<Props, State> {
 
         const curLastPost = curPostsArray[0];
 
-        if (curLastPost.user_id === this.props.currentUserId) {
+        if (this.userIsAtLastPost() || curLastPost.user_id === this.props.currentUserId) {
             this.scrollToBottom();
         }
     }
@@ -170,6 +170,15 @@ export default class RhsThread extends React.Component<Props, State> {
         if (UserAgent.isMobile() && document!.activeElement!.id === 'reply_textbox') {
             this.scrollToBottom();
         }
+    }
+
+    private userIsAtLastPost = () => {
+        const postReplies = (this.rhspostlistRef.current as HTMLElement).childNodes;
+        const rhsPostListRect = (this.rhspostlistRef.current as HTMLElement).getBoundingClientRect();
+        const offset = 24;
+        const postRepliesArray = [...postReplies];
+
+        return ((postRepliesArray[postRepliesArray.length - 1] as HTMLElement).offsetTop + offset) + rhsPostListRect.top === rhsPostListRect.bottom;
     }
 
     private handleCardClick = (post: Post) => {
--- a/components/rhs_thread/rhs_thread.tsx
+++ b/components/rhs_thread/rhs_thread.tsx
@@ -118,11 +118,10 @@ export default class RhsThread extends React.Component<Props, State> {
     private resizeRhsPostList() {
         const containerHeight = this.containerRef.current?.getBoundingClientRect().height;
         const createContainerHeight = this.postCreateContainerRef.current?.getBoundingClientRect().height;
-        const lastPost = this.getLastPost()?.getBoundingClientRect();
-        const bottomMargin = 8;
-        if (containerHeight && createContainerHeight && lastPost) {
+        const headerHeight = 56;
+        if (containerHeight && createContainerHeight) {
             this.setState({
-                postsContainerHeight: (containerHeight - (createContainerHeight + lastPost.height)) + bottomMargin,
+                postsContainerHeight: containerHeight - (createContainerHeight + headerHeight),
             });
         }
     }
--- a/components/rhs_thread/rhs_thread.tsx
+++ b/components/rhs_thread/rhs_thread.tsx
@@ -70,11 +70,15 @@ type State = {
     openTime: number;
     postsArray?: Record<string, any>[];
     isBusy?: boolean;
+    postsContainerHeight: number;
+    userScrolledToBottom: boolean;
 }
 
 export default class RhsThread extends React.Component<Props, State> {
     private scrollStopAction: DelayedAction;
     private rhspostlistRef: React.RefObject<HTMLDivElement>;
+    private containerRef: React.RefObject<HTMLDivElement>;
+    private postCreateContainerRef : React.RefObject<HTMLDivElement>;
 
     public static getDerivedStateFromProps(props: Props, state: State) {
         let updatedState: Partial<State> = {selected: props.selected};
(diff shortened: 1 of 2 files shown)


hmhealey:
I think the ticket for this might be incorrect. If the user has scrolled up in the RHS, I wouldn't expect to be scrolled to the bottom.

nevyangelova:
Perhaps but this is the ticket that has all the correspondence and is scheduled for the release so I think we can go ahead with it

components/rhs_thread/rhs_thread.tsx
```diff
@@ -172,6 +172,19 @@ export default class RhsThread extends React.Component<Props, State> {
...
+        const postReplies = (this.rhspostlistRef.current as HTMLElement)?.childNodes;
+        const rhsPostListRect = (this.rhspostlistRef.current as HTMLElement)?.getBoundingClientRect();
+        const offset = 24;
+
+        if (postReplies) {
+            const postRepliesArray = Array.from(postReplies);
+            return ((postRepliesArray[postRepliesArray.length - 1] as HTMLElement).offsetTop + offset) + rhsPostListRect.top === rhsPostListRect.bottom;
+        }
+
+        return false;
```
Willyfrog, line 185: if there are no replies, shouldn't it be true that user is at last post?

components/rhs_thread/rhs_thread.tsx
```diff
@@ -172,6 +172,19 @@ export default class RhsThread extends React.Component<Props, State> {
...
     }
 
+    private userIsAtLastPost = () => {
+        const postReplies = (this.rhspostlistRef.current as HTMLElement)?.childNodes;
+        const rhsPostListRect = (this.rhspostlistRef.current as HTMLElement)?.getBoundingClientRect();
+        const offset = 24;
+
+        if (postReplies) {
+            const postRepliesArray = Array.from(postReplies);
+            return ((postRepliesArray[postRepliesArray.length - 1] as HTMLElement).offsetTop + offset) + rhsPostListRect.top === rhsPostListRect.bottom;
```
Willyfrog, line 182: I'm not sure about equality here, if the list is scrolled a few pixels up it shouldn't scroll although the last post is almost entirely visible? 0/5 on this, but I think we should at least consider it :)

Willyfrog (changes requested):
I left some comments but it is ok not to change anything if that's the behaviour we want, let me know if you think otherwise and I'll aprove it :)

components/rhs_thread/rhs_thread.tsx
nevyangelova: Makes sense to address it with UX with the new implementation but for this I guess the user will be scrolled down just a little bit when the new message is incoming? I did a few tests by tweaking the values and it all seemed ok to me but additional opinions are welcome!

components/rhs_thread/rhs_thread.tsx
nevyangelova: Yes

nevyangelova: Sorry actually no, in this case we should keep it false, otherwise it will scroll every time, since it's in an `||` condition in the scrollToBottom logic

components/rhs_thread/rhs_thread.tsx
```diff
@@ -172,6 +172,19 @@ export default class RhsThread extends React.Component<Props, State> {
         }
     }
 
+    private userIsAtLastPost = () => {
+        const postReplies = (this.rhspostlistRef.current as HTMLElement)?.childNodes;
```
hmhealey, line 176: Why do we need to cast this to an HTMLElement? Shouldn't an HTMLDivElement already have these fields?

Maybe this is related to how I keep having issues treating HTMLElement as a base type for the other HTML*Element types though

hmhealey: Also, I think `children` might be preferable to `childNodes` here

components/rhs_thread/rhs_thread.tsx
```diff
@@ -172,6 +172,19 @@ export default class RhsThread extends React.Component<Props, State> {
...
+        const rhsPostListRect = (this.rhspostlistRef.current as HTMLElement)?.getBoundingClientRect();
+        const offset = 24;
+
+        if (postReplies) {
+            const postRepliesArray = Array.from(postReplies);
```
hmhealey, line 181: Is `Array.from` necessary here? `childNodes` isn't an array, but I think it still supports `.length` and indexing

components/rhs_thread/rhs_thread.tsx
hmhealey, line 182: I'm not sure I understand this logic here. It doesn't look like it's checking if the RHS is scrolled to the bottom since there would still be the comment box below `rhspostlistRef`. Could you explain what we're trying to check for here?

hmhealey (commented):
Also, not to invalidate your work here, but it might be a bit easier to use `getSnapshotBeforeUpdate` to check if we were at the bottom before the post was made and use that to keep us at the bottom.

hmhealey:
I think this method is having some issues with shorter posts. It seems to only scroll once every 3 or 4 posts I'm not sure if that's coming from the offset or something
ezgif-1-eb1b438c8f60

components/rhs_thread/rhs_thread.tsx
nevyangelova: Yeah it's definitely the offset for shorter posts, basically I check if the offset of the last post + its height minus the hidden part of the rhs post list container is equal to the visible distance.

components/rhs_thread/rhs_thread.tsx
nevyangelova: Sorry I left that change because I was using a filter before. I will update.

components/rhs_thread/rhs_thread.tsx
```diff
@@ -102,6 +102,14 @@ export default class RhsThread extends React.Component<Props, State> {
         this.rhspostlistRef = React.createRef();
     }
 
+    public getSnapshotBeforeUpdate(prevProps: Props) {
+        if (prevProps.posts.length < this.props.posts.length) {
+            const list = this.rhspostlistRef.current;
+            return list ? list.scrollHeight - list.scrollTop : null;
```
hmhealey, line 108: I think it might be a bit nicer to return an object like `{atBottom: true}` here instead of just returning a number.

hmhealey (changes requested):

components/rhs_thread/rhs_thread.tsx
hmhealey: Also, it looks like the type of the snapshot is the 3rd type parameter like `RhsThread extends React.PureComponent<Props, State, Snapshot>`

jwilander:
Test server update is failing on:

```
{"level":"info","ts":1604160553.778792,"caller":"app/server.go:226","msg":"Server is initializing..."}
{"level":"info","ts":1604160553.7791555,"caller":"app/web_hub.go:83","msg":"Starting websocket hubs","number_of_hubs":4}
{"level":"warn","ts":1604160553.779271,"caller":"app/server.go:235","msg":"Sentry reporting is enabled, but SENTRY_DSN is not set. Disabling reporting."}
{"level":"debug","ts":1604160553.7792916,"caller":"app/web_hub.go:413","msg":"Hub is starting","index":1}
{"level":"debug","ts":1604160553.7794287,"caller":"app/web_hub.go:413","msg":"Hub is starting","index":0}
{"level":"debug","ts":1604160553.7794766,"caller":"app/web_hub.go:413","msg":"Hub is starting","index":2}
{"level":"debug","ts":1604160553.7795568,"caller":"app/web_hub.go:413","msg":"Hub is starting","index":3}
{"level":"info","ts":1604160553.7802284,"caller":"utils/i18n.go:90","msg":"Loaded system translations","for locale":"en","from locale":"/mattermost/i18n/en.json"}
{"level":"debug","ts":1604160553.7809167,"caller":"utils/html.go:30","msg":"Parsing server templates","templates_directory":"/mattermost/templates"}
{"level":"info","ts":1604160553.7844265,"caller":"sqlstore/supplier.go:235","msg":"Pinging SQL","database":"master"}
{"level":"info","ts":1604160553.7996995,"caller":"sqlstore/supplier.go:235","msg":"Pinging SQL","database":"replica-0"}
{"level":"error","ts":1604160554.0809183,"caller":"sqlstore/supplier.go:909","msg":"Failed to create index","error":"pq: relation \"idx_threads_channel_id\" does not exist","error":"pq: column \"channelid\" does not exist"}
```

I'm going to re-create it

jgilliam17:
Thank you @nevyangelova 
Re-tested and looks good on big screens now too.
I'll approve after dev reviews, just in cases I need to test additional changes.

Willyfrog (approved):
LGTM!

hmhealey (approved):

PR #6933, opened 2020-10-27
[MM-29836] Migrate changeCSS() to CSS variable in utils/utils.jsx, ln. 693

kcc343:
#### Summary
Removes ChangeCSS() for `.app__body .shadow--2',` selectors with `'box-shadow: 0 20px 30px 0' <ins> changeOpacity(theme.centerChannelColor, 0.1) </ins> ', 0 14px 20px 0 ' <ins> changeOpacity(theme.centerChannelColor, 0.1)` properties

I am assuming `.shadow--2` does not affect any elements on the app so I am just deleting the line if that is right

#### Ticket Link
  Github Issue: Fixes https://github.com/mattermost/mattermost-server/issues/16022
  JIRA Ticket: Fixes https://mattermost.atlassian.net/browse/MM-29836

PR #6938, opened 2020-10-27
Apply Theme center-channel-color to .shadow--2

reginaalyssa:
#### Summary
Removes `changeCss()` for `.shadow--2` `-moz-box-shadow` and `-webkit-box-shadow` (currently only used in .app__body .system-notice.shadow--2)
Removes `.shadow--2` class for `SystemNotice` component
No additions were made to SCSS files because `SystemNotice` already has `@include box-shadow(0 20px 30px alpha-color($black, .07), 0 14px 20px alpha-color($black, .07));` set in `sass/components/_system-notice.scss` 

#### Ticket Link
Fixes [mattermost/mattermost-server#16023](https://github.com/mattermost/mattermost-server/issues/16023)
Fixes [mattermost/mattermost-server#16024](https://github.com/mattermost/mattermost-server/issues/16024)

#### Related Pull Requests
Seems like the non `-moz` and non -`webkit` version of this similar change was tackled in https://github.com/mattermost/mattermost-webapp/pull/6933

#### Screenshots
Currently retains its previous look as shown below:
image

### DIFF
--- a/components/system_notice/__snapshots__/system_notice.test.jsx.snap
+++ b/components/system_notice/__snapshots__/system_notice.test.jsx.snap
@@ -2,7 +2,7 @@
 
 exports[`components/SystemNotice should match snapshot for admin, admin notice 1`] = `
 <div
-  className="system-notice bg--white shadow--2"
+  className="system-notice bg--white"
 >
   <div
     className="system-notice__header"
@@ -69,7 +69,7 @@ exports[`components/SystemNotice should match snapshot for admin, admin notice 1
 
 exports[`components/SystemNotice should match snapshot for admin, regular notice 1`] = `
 <div
-  className="system-notice bg--white shadow--2"
+  className="system-notice bg--white"
 >
   <div
     className="system-notice__header"
@@ -119,7 +119,7 @@ exports[`components/SystemNotice should match snapshot for admin, regular notice
 
 exports[`components/SystemNotice should match snapshot for regular user, admin and regular notice 1`] = `
 <div
-  className="system-notice bg--white shadow--2"
+  className="system-notice bg--white"
 >
   <div
     className="system-notice__header"
@@ -177,7 +177,7 @@ exports[`components/SystemNotice should match snapshot for regular user, no noti
 
 exports[`components/SystemNotice should match snapshot for regular user, regular notice 1`] = `
 <div
-  className="system-notice bg--white shadow--2"
+  className="system-notice bg--white"
 >
   <div
     className="system-notice__header"
@@ -229,7 +229,7 @@ exports[`components/SystemNotice should match snapshot for show function returni
 
 exports[`components/SystemNotice should match snapshot for show function returning true 1`] = `
 <div
-  className="system-notice bg--white shadow--2"
+  className="system-notice bg--white"
 >
   <div
     className="system-notice__header"
@@ -279,7 +279,7 @@ exports[`components/SystemNotice should match snapshot for show function returni
 
 exports[`components/SystemNotice should match snapshot for with allowForget equal false 1`] = `
 <div
-  className="system-notice bg--white shadow--2"
+  className="system-notice bg--white"
 >
   <div
     className="system-notice__header"
--- a/components/system_notice/system_notice.jsx
+++ b/components/system_notice/system_notice.jsx
@@ -116,7 +116,7 @@ export default class SystemNotice extends React.PureComponent {
 
         return (
             <div
-                className='system-notice bg--white shadow--2'
+                className='system-notice bg--white'
             >
                 <div className='system-notice__header'>
                     <div className='system-notice__logo'>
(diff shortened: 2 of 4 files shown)


reginaalyssa:
Please let me know if we should instead modify `@include box-shadow(0 20px 30px alpha-color($black, .07), 0 14px 20px alpha-color($black, .07));` from `.system-notice` such that `alpha-color($black, .07)` is replaced by either:

- `rgba(var(--center-channel-color-rgb), 0.1)` (opacity taken from the original `changeCss()`)
- `rgba(var(--center-channel-color-rgb), 0.07)` (opacity taken from the current CSS for `system-notice`)

Thank you!

hmhealey:
I think it should probably be the 0.1 since the rules applied by changeCss override the ones in the plain CSS.

reginaalyssa:
I can update accordingly if needed but when I checked the CSS prior to any
changes made (removing changeCss), Inspect Element showed that the element
with the class of .system-notice.shadow—2 was still following the declared
.system-notice box-shadow from the SCSS file instead of the .shadow—2
changeCss()

Please feel free to let me know if the same behavior happens on your end
and what actions need to be done for this PR. Thank you! :)

hmhealey:
On master, I'm seeing the box shadow being applied by changeCss taking effect here.

Screen Shot 2020-10-28 at 1 25 01 PM

Are you perhaps testing on the PR branch? Those lines have been removed here

reginaalyssa:
Ohh, I might have to double check again but I remember trying it out on this PR branch where the changeCss lines were still present/not removed. In any case, thanks for confirming the behavior on your end!

Will revisit this again and I’ll update the PR with the 0.1 opacity. Do you recommend replacing the box shadow for system-notice directly or would we prefer to keep the .shadow—2 class and add the box-shadow there instead?

hmhealey:
I like adding it to the `system-notice` class. I'm not sure if we were planning on using the `shadow--2` class elsewhere before, but I don't see any reason to keep extra utility classes around that we aren't using.

reginaalyssa:
Screen Shot 2020-10-27 at 11 26 37 PM

Attaching the updated look of the System Notice after applying the change of using box-shadow based on the center-channel-color with opacity of 0.1

reginaalyssa:
Updated the SCSS file to reflect the changes already. 

Just wanted to give an update on the double checking. I checked on my local for Safari and Chrome for the `master` branch, it's still appearing as if the `.shadow--2` has no effect on the component/element on my end. Not sure what's happening, I also restarted the `make run` and the box-shadow from `.system-notice` is still the only one showing up on Inspect Element. Maybe I needed to configure something else to replicate the setup. :(

But pushed the changes already as shown in my previous comment. Let me know if there's anything else that needs to be done!

Screen Shot 2020-10-29 at 11 18 11 AM
Screen Shot 2020-10-29 at 11 09 09 AM

hmhealey (changes requested):
I wouldn't worry too much about the class seemingly not being used before. I wouldn't be surprised if it was broken at some point and nobody noticed.

Your PR is currently failing because the test snapshots for those components are expecting the shadow--2 class to still be there, so those probably just need to be updated. If you run `npm test -- -u`, it'll update them for you, so as long as they match what your changes did, you can just commit them as-is.

reginaalyssa:
> I wouldn't worry too much about the class seemingly not being used before. I wouldn't be surprised if it was broken at some point and nobody noticed.
> 
> Your PR is currently failing because the test snapshots for those components are expecting the shadow--2 class to still be there, so those probably just need to be updated. If you run `npm test -- -u`, it'll update them for you, so as long as they match what your changes did, you can just commit them as-is.

Hello! I tried running the command mentioned but there weren't any changes when I ran `git status` after running `npm test -- -u`. I also tried manually changing the contents of `components/system_notice/__snapshots__/system_notice.test.jsx.snap` but they weren't appearing in the changes I could commit. Hope it's okay to ask for help on whether these files are included in `.gitignore` or if there's an added step I need to do. Thank you!

hmhealey:
Nope, that should be it. You would see a summary with something like "5 snapshots updated from 2 test suites" if something changed as well though since that's what CI is failing on. I can push the required changes if you'd like since it should only take a minute.

reginaalyssa:
> Nope, that should be it. You would see a summary with something like "5 snapshots updated from 2 test suites" if something changed as well though since that's what CI is failing on. I can push the required changes if you'd like since it should only take a minute.

Yes please, that would be helpful as the changes aren't showing up on my end. Thank you!

In the meantime, I've resolved the merge conflicts for this branch.

hmhealey (approved):
Snapshots are updated, so the build is now passing

reginaalyssa:
Thanks @hmhealey!

sudheerDev (approved):
Thank you @reginaalyssa

jgilliam17 (approved):
Thank you @reginaalyssa 
Approving based on your attached screenshot. I wasn't able to manually test by triggering the admin only notice in sys. console.

hmhealey:
Thanks again for the PR, @reginaalyssa. This'll be in for 5.30 releasing in December.

PR #6939, opened 2020-10-27
[MM-29120/MM-27904, MM-29121] Collapsed Reply Threads: ThreadList and ThreadHeader

calebroseland:
This is part 2 (a continuation) of #6585. Once part 1 is merged, the diff will be smaller e.g. https://github.com/mattermost/mattermost-webapp/compare/feature/threading-overhaul-comps-a...feature/threading-comps-b.

Storybook-only code. Nothing really to QA.

- **[MM-18107](https://mattermost.atlassian.net/browse/MM-18107)**: 
  - [MM-29121](https://mattermost.atlassian.net/browse/MM-29121): `<ThreadHeader>` ([storybook link](https://alpha-storybook.devops.mattermost.com/feature/threading-comps-b/index.html?path=/story/features-threading-global-threads--threadheader))
  - [MM-29120](https://mattermost.atlassian.net/browse/MM-29120)/[MM-27904](https://mattermost.atlassian.net/browse/MM-27904): `<ThreadList>` ([storybook link](https://alpha-storybook.devops.mattermost.com/feature/threading-comps-b/index.html?path=/story/features-threading-global-threads--threadlist))

#### Screenshots

Screen Shot 2020-11-02 at 16 58 38
Screen Shot 2020-11-02 at 16 58 56


matthewbirtch (changes requested):
Great work @calebroseland. Looks great. This first pass is focused on the Thread Header. I'll move on to the others when I have a moment.

## Thread Header
1. The Channel Name hover state needs to increase the text's opacity `color: rgba(var(--center-channel-color-rgb), 0.72);`

2. Is it possible to make this menu a touch wider so that the help text fits on one line in English? Looks like a width of 248px would do the trick

3. For the 'Following' button, we need a hover state for when it's followed. We could just add `background: rgba(var(--button-bg-rgb), 0.16);` to the hover state for when the following button is active.

4. The items on the left side of the thread header don't appear to be quite aligned with the items on the right side


5. The height of the header container appears to be much taller than it should be. It's measuring at 86px, but should be 56px.
image

6. The 'Thread' title should be using the `center-channel-color` variable. It appears to be using `#333333`

matthewbirtch (approved):
And here's the review for the Thread List. Nice work!

## Thread List
1. Can we remove the left margin on the 'All your threads' button? (but keep the 4px margin between the two buttons)
2. Now that I'm seeing the thread list header and the list items together, I'm wondering if we need to move the ••• hover button to align with the 'mark as read' icon. It feels odd that they're not aligned. Maybe we need to do something like this:

3. For the timestamp, can we remove the `.` character and make it plural (when needed)? e.g. `5 hrs ago` or `10 mins ago` or `2 weeks ago` (not sure of wks is a commonly used shortform)
4. Also, I think we are going to want to use the month shortform too when the date is showing on the timestame. e.g. `Oct 1` (and don't use `01` for October 1)
5. We should capitalize 'Yesterday'
6. `This wk.` should be `This week' if possible

matthewbirtch (changes requested):
Meant to choose 'request changes' not 'approve' in the last comment. See above

jespino:
I going to be some days out, so I've unassigned myself from this review.

calebroseland:
@matthewbirtch 

**Thread Header**
1–6 fixed ✓

**Thread List**
1. Done
2. Done
3. We're using a batteries-included browser API for these, so fine tuning low-level punctuation/plurals would be quite a bit of work (well-written customization methods/pipelines and i18n work). I think we should maybe stick with `style='long'` (`Last week`, `5 hours ago`) for now. 
 
This was technically a part of the previous PR, and some of these timestamp wording/formats could span multiple areas. Like with date separators, they currently have the leading zero for days. ThreadFooter would be in center channel, where current timestamps don't use relative time. 

4. I set ThreadFooter and ThreadItem to standard numeric day (no leading zero).

**Should we write up a separate story to address/synchronize time formats in Threads and Channels?** 

5. Capitalization applied in ThreadItem (`Now`, `Yesterday`, `Last week`, and so on)

6. Mixed short-vs-long (depending on unit) would be possible, but for reasons in 3, maybe we should just stick with `long` for now?

matthewbirtch:
> **Should we write up a separate story to address/synchronize time formats in Threads and Channels?**

@calebroseland ya, we might want to if we find that the long format is causing collisions with the username and channel name

matthewbirtch (approved):
Looks awesome. Great work! We can address some of the time format issues later if they come up as more significant problems

esethna:
@matthewbirtch perhaps we should keep track of anything that we are pushing to address later? Propose maybe opening a new page in the confluence site for tracking that stuff to revisit?

matthewbirtch:
> @matthewbirtch perhaps we should keep track of anything that we are pushing to address later? Propose maybe opening a new page in the confluence site for tracking that stuff to revisit?

Done: https://mattermost.atlassian.net/wiki/spaces/Threads/pages/1157791745/Items+to+Revisit

calebroseland:
Closing PR as it is continued in `feature/collapsed-reply-threads` and merging would produce conflicts down the line.

PR #6956, opened 2020-10-28
MM-29986 Don't show archived channels with unread filter enabled

hmhealey:
This is coming in a bit late, but with the new unread filter enabled, we forgot to filter out archived channels when the user isn't looking at them.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29986

### DIFF
--- a/selectors/views/channel_sidebar.ts
+++ b/selectors/views/channel_sidebar.ts
@@ -84,9 +84,11 @@ export const getUnreadChannels = (() => {
         (state: GlobalState) => getUnreadChannelIds(state),
         getCurrentChannelId,
         (allChannels, unreadChannelIds, currentChannelId) => {
-            const unreadChannels = unreadChannelIds.map((channelId) => allChannels[channelId]);
+            const unreadChannels = unreadChannelIds.
+                map((channelId) => allChannels[channelId]).
+                filter((channel) => channel.delete_at === 0);
 
-            if (unreadChannelIds.indexOf(currentChannelId) === -1) {
+            if (unreadChannels.findIndex((channel) => channel.id === currentChannelId) === -1) {
                 unreadChannels.push(allChannels[currentChannelId]);
             }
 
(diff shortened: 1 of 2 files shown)


selectors/views/channel_sidebar.ts
```diff
@@ -84,9 +84,11 @@ export const getUnreadChannels = (() => {
         (state: GlobalState) => getUnreadChannelIds(state),
         getCurrentChannelId,
         (allChannels, unreadChannelIds, currentChannelId) => {
-            const unreadChannels = unreadChannelIds.map((channelId) => allChannels[channelId]);
+            const unreadChannels = unreadChannelIds.
+                map((channelId) => allChannels[channelId]).
+                filter((channel) => channel.delete_at === 0);
```
Willyfrog, line 89: you can do it in just one go, to prevent going through the list multiple times:

```suggestion
            const unreadChannels = unreadChannelIds.
                reduce((result, channelId) => {
                    const channel = allChannels[channelId];

                    // remove archived channels from the list unless it is the one we are currently reading
                    if (channel.delete_at !== 0 || channel.id === currentChannelId) {
                        result.push(channel);
                    }
                    return result;
                 });
```

you can also use a foreach if you think it  would be more clear.

Willyfrog (commented):
code looks good, but I added a small performance optimization

devinbinnie (approved):

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified archived channels do not show on the LHS when unread filter is enabled. An archived channel is visible on the LHS with the filter enabled only when user is currently viewing it.

selectors/views/channel_sidebar.ts
hmhealey: I'm curious what the actual performance difference is here. I tend to use multiple calls for clarity, but that might be because I'm thinking too much about the big-O notation performance where looping through an array once is treated the same as doing it 100,000 times 😛.

I think I'll go for a for loop. I always have trouble wrapping my head around reduce statements for some reason, and I've heard for-in is substantially faster than forEach

Willyfrog: yes it is the same in big-O notation :) and most likely there is a negligible performance gain, but I guess my functional programmer ODT sees a map+filter and needs to convert it to a reduce 😅

Willyfrog (approved):

PR #6965, opened 2020-10-29
[MM-20428] Migrate terms of service to Typescript

Saucistophe:
#### Summary
Migrate terms of service to Typescript.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16146
https://mattermost.atlassian.net/browse/MM-20428


components/terms_of_service/terms_of_service.test.tsx
```diff
@@ -62,29 +64,29 @@ describe('components/terms_of_service/TermsOfService', () => {
 
     test('should call updateTermsOfServiceStatus on registerUserAction', async () => {
         const wrapper = shallow(<TermsOfService {...baseProps}/>);
-        await wrapper.instance().registerUserAction(true, jest.fn());
+        await (wrapper.instance() as TermsOfService).registerUserAction(true, jest.fn());
```
hmhealey, line 67: I think it might be preferable to instead add a type parameter to `shallow` instead of using `as` to cast the results of `wrapper.instance()` later on like
```
const wrapper = shallow<TermsOfService>(<TermsOfService {...baseProps}/>);
await wrapper.instance().registerUserAction(true, jest.fn());
```
I think it ends up being slightly easier if you call `wrapper.instance()` multiple times

components/terms_of_service/terms_of_service.tsx
```diff
@@ -19,19 +19,40 @@ import {browserHistory} from 'utils/browser_history';
...
+    emojiMap: EmojiMap;
+}
+
+interface TermsOfServiceState {
+    customTermsOfServiceId: string;
+    customTermsOfServiceText: string;
+    loading: boolean;
+    loadingAgree: boolean;
+    loadingDisagree: boolean;
+    serverError: JSX.Element | null;
```
hmhealey, line 49: This can be made a bit simpler by using `React.ReactNode` instead.
```suggestion
    serverError: React.ReactNode;
```

hmhealey (changes requested):
Thanks for the PR, @Saucistophe!

I just had a couple suggestions of things that could be slightly simplified, and could you also migrate `components/terms_of_service/index.js` to TypeScript as well?

components/terms_of_service/terms_of_service.test.tsx
Saucistophe: I was trying to do something like that, but did not realize the generic parameter was not enough.
Thanks!

components/terms_of_service/terms_of_service.tsx
Saucistophe: Done

Saucistophe:
There you go

hmhealey (approved):
Thanks for changing those. Looks good to me now!

jfrerich (commented):
LGTM!  Thanks, @Saucistophe!

jfrerich (approved):

jgilliam17 (approved):
Thank you @Saucistophe 
Tested, looks good to merge.
- Verified Terms of service - no change in functionality.

hmhealey:
Merged. Thanks again for the PR, @Saucistophe. This'll be in the 5.30, coming in December.

PR #6966, opened 2020-10-30
Mirgrate password reset form to ts

morganrconnolly:
#### Summary

Updates `components/password_reset_form/` to typescript

#### Ticket Link

fixes https://github.com/mattermost/mattermost-server/issues/16147

#### Related Pull Requests


morganrconnolly:
I'm stuck on fixing the last linting error on this PR
```
components/password_reset_form/password_reset_form.test.tsx:42:28 - error TS2339: Property 'passwordInput' does not exist on type 'Component<{}, {}, any>'.

42         wrapper.instance().passwordInput.current!.value = 'PASSWORD';
```
I'm not sure how to declare a more specific type for `wrapper` because it's generated using `mountWithIntl`, specifically
```       
 const wrapper = mountWithIntl(<PasswordResetForm {...props}/>);
```

components/password_reset_form/index.ts
```diff
@@ -1,22 +1,30 @@
...
 import {resetUserPassword} from 'mattermost-redux/actions/users';
 
+import {ServerError} from 'mattermost-redux/types/errors';
+import {GenericAction} from 'mattermost-redux/types/actions';
+import {GlobalState} from 'types/store';
 import PasswordResetForm from './password_reset_form';
 
-const mapStateToProps = (state) => {
+type Actions = {
+    resetUserPassword: (token: any, newPassword: string) => any;
```
Willyfrog, line 15: seems like function signature does not match the one in https://github.com/mattermost/mattermost-redux/blob/master/src/actions/users.ts#L1193
can you take a look at it?

components/password_reset_form/password_reset_form.tsx
```diff
@@ -1,39 +1,39 @@
...
 import {FormattedMessage} from 'react-intl';
 
+import {ServerError} from 'mattermost-redux/types/errors';
 import {browserHistory} from 'utils/browser_history';
 import Constants from 'utils/constants';
 import LocalizedInput from 'components/localized_input/localized_input';
 
 import {t} from 'utils/i18n.jsx';
+interface Props {
+  location: {search: any};
```
Willyfrog, line 14: why any? wouldn't search be a string?

components/password_reset_form/password_reset_form.tsx
```diff
@@ -1,39 +1,39 @@
...
+  siteName: string | undefined;
```
Willyfrog, line 15: I think the convention is to use `?`
```suggestion
  siteName?: string;
```

components/password_reset_form/password_reset_form.tsx
```diff
@@ -1,39 +1,39 @@
...
+  actions: {
+    resetUserPassword: (token: any, newPassword: string) => any;
```
Willyfrog, line 17: see comment above

Willyfrog (changes requested):
Hello and thanks for working on this!
I've added a few changes on types that might need to be addressed, can you take a look and share your thoughts on it?

components/password_reset_form/index.ts
morganrconnolly: `token: any` was an attempt to fix this `make check-types` error.
image

Let me know if there's anything else I should do to address the above error

morganrconnolly:
@Willyfrog Thanks for the review!

Willyfrog (changes requested):
can you remove the package-lock.json from the PR? I didn't notice that it was in the PR but if we are not changing the package.json file we shouldn't modify this one, otherwise we'll get a lot of conflicts

Also, it seems the test is failing because it is sending a null instead of the expected string. So either the original signature is wrong (which I doubt) or the test is not passing the needed data, can you review the tests and fix them?

Thanks for the changes, btw!

nickmisasi (commented):
Thanks @morganrconnolly! +1 on what @Willyfrog mentioned. Feel free to re-request a review from me once these issues have been fixed!

components/password_reset_form/index.ts
Willyfrog: sorry, I missed this comment!
so the problem seems to be that on line 53 it tries to get the token from the url search params:
`        const token = (new URLSearchParams(this.props.location.search)).get('token');`
but that might not exist. if this token that is required doesn't exist, what would be the next step?

For me, it basically means that either we are doing the types wrong (we are asking it to be required but it is not) or we ar not being consistent with it (like exiting or reporting an error if the conditions doesn't match the requirements)

Willyfrog:
hello @morganrconnolly seems like the tests are failing regarding types, can you take a look? let us know if you need any help with that :)

morganrconnolly:
@Willyfrog I'm stuck on the following make check-types error:  
image
Let me know what you think. 

I looked at `resetPassword` in `mattermost-server/api4/user.go` to affirm that `token` is required for the call to `resetUserPassword` in `password_reset_form.tsx`.  I added a cast to resolve the `string | null` token error . Additionally, an error is now thrown if `token === null` . 
Any feedback is welcome c:

Willyfrog:
> @Willyfrog I'm stuck on the following make check-types error:
> image
> Let me know what you think.

Hi there!!

Taking a quick look to some similar file (`password_reset_link.test.tsx`) I'd say this should help:

```ts
       const wrapper = mountWithIntl(<PasswordResetForm {...props}/>);
        (wrapper.instance() as PasswordResetForm).passwordInput.current!.value = 'PASSWORD';
```

> I looked at `resetPassword` in `mattermost-server/api4/user.go` to affirm that `token` is required for the call to `resetUserPassword` in `password_reset_form.tsx`. I added a cast to resolve the `string | null` token error . Additionally, an error is now thrown if `token === null` .
> Any feedback is welcome c:

the `if` statement is using `typeof token` instead of the actual token, this is asking if its type is null, which is not as it has a type. Instead, you need to ask if the value is `null`, so the compiler can remove that from the list of possible values. 

Something like the next snippet should work:
```ts
        if (token === null) {
            throw new Error('token must be a string');
        }
        const safeToken: string = token;
```

devinbinnie:
@morganrconnolly Do you still plan to finish up this PR?

AshishDhama:
Closing as done via https://github.com/mattermost/mattermost-webapp/pull/10027

PR #6967, opened 2020-10-30
MM-30139 Cypress/E2E: Force retry when test executions failed on internal server error

saturninoabril:
#### Summary
Force retries (until CI timeout) when test executions failed on internal server error

#### Ticket Link
Jira ticket - https://mattermost.atlassian.net/browse/MM-30139

#### Screenshots
Screen Shot 2020-10-30 at 3 51 26 PM

### DIFF
--- a/e2e/utils/test_cases.js
+++ b/e2e/utils/test_cases.js
@@ -1,9 +1,12 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+/* eslint-disable no-console */
+
 // See reference: https://support.smartbear.com/tm4j-cloud/api-docs/
 
 const axios = require('axios');
+const chalk = require('chalk');
 
 const {getAllTests} = require('./report');
 
@@ -74,7 +77,7 @@ function saveToEndpoint(url, data) {
         },
         data,
     }).catch((error) => {
-        console.log('Something went wrong:', error.response.data.message); // eslint-disable-line no-console
+        console.log('Something went wrong:', error.response.data.message);
         return error.response.data;
     });
 }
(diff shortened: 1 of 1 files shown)


saturninoabril:
@amyblais I'm requesting to cherry-pick this into release-5.29 so that we'll be able to consistently save test results from Cypress to TM4J. No side-effect to product.

amyblais:
Sounds good, thanks @saturninoabril!

srkgupta (changes requested):
Hi @saturninoabril 
Can we add a max retry of upto 3 or 5 times. Currently this goes on forever and it might be stuck if there is something wrong somewhere. 

Oops. Looking at the screenshot, it takes around 16 retries for it to succeed. Please ignore my previous suggestion of 3-5 retries. But can we at least try for a definite amount of retries? Pl

saturninoabril:
Unfortunately, I ended up a force retry like "retry until fail due to timeout in CI, usually up to 10 minutes" and try to save as much as possible.

srkgupta:
> Unfortunately, I ended up a force retry like "retry until fail due to timeout in CI, usually up to 10 minutes" and try to save as much as possible.

Each test taking upto 10 mins (in worst case) sounds a lot. Can we retry max of say 25 or 40 times to begin with?

saturninoabril:
Ah, I see where you're going to and I'd say it's not how it works. Saving test results has its own job/step with 10 mins default timeout. Worst case is 10 mins for all the tests (not per test) but as I observed it, its max saving time is 20 seconds. My suggestion is to go on max timeout and limit retry count in the future only if needed. Currently, I'm 3/5 to have it as is.

srkgupta (approved):
Thanks for the clarification @saturninoabril. That's fine then. LGTM

josephbaylon (approved):
LGTM. Ah this only retries saving of test executions and not retrying test execution itself. The title got me.

PR #6973, opened 2020-10-30
[MM-20407] Converted js to ts

saltshaiker:
#### Summary
Migrates js/jsx files to ts/tsx in `components/analytics/team_analytics`. However, running into the following error on several lines in `team_analytics.tsx`:
`Property 'team' does not exist on type 'Readonly<{}>'. ts(2339)`

#### Ticket Link
(Attempting to) Fix [https://github.com/mattermost/mattermost-server/issues/16165](https://github.com/mattermost/mattermost-server/issues/16165)

#### Related Pull Requests
N/A

#### Screenshots
N/A

### DIFF
--- a/components/analytics/team_analytics/index.ts
+++ b/components/analytics/team_analytics/index.ts
@@ -1,20 +1,20 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {connect} from 'react-redux';
-import {bindActionCreators} from 'redux';
-import {getTeams} from 'mattermost-redux/actions/teams';
-import {getProfilesInTeam} from 'mattermost-redux/actions/users';
-import {getTeamsList} from 'mattermost-redux/selectors/entities/teams';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import { getTeams } from 'mattermost-redux/actions/teams';
+import { getProfilesInTeam } from 'mattermost-redux/actions/users';
+import { getTeamsList } from 'mattermost-redux/selectors/entities/teams';
 
 import BrowserStore from 'stores/browser_store.jsx';
-import {getCurrentLocale} from 'selectors/i18n';
+import { getCurrentLocale } from 'selectors/i18n';
 
-import TeamAnalytics from './team_analytics.jsx';
+import TeamAnalytics from './team_analytics';
 
 const LAST_ANALYTICS_TEAM = 'last_analytics_team';
 
-function mapStateToProps(state) {
+function mapStateToProps(state: any) {
     const teams = getTeamsList(state);
     const teamId = BrowserStore.getGlobalItem(LAST_ANALYTICS_TEAM, null);
     const initialTeam = state.entities.teams.teams[teamId] || (teams.length > 0 ? teams[0] : null);
@@ -27,7 +27,7 @@ function mapStateToProps(state) {
     };
 }
 
-function mapDispatchToProps(dispatch) {
+function mapDispatchToProps(dispatch: any) {
     return {
         actions: bindActionCreators({
             getTeams,
--- a/components/analytics/team_analytics/team_analytics.tsx
+++ b/components/analytics/team_analytics/team_analytics.tsx
@@ -5,6 +5,8 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 import { General } from 'mattermost-redux/constants';
+import { Team, TeamStats } from 'mattermost-redux/types/teams';
+import { RelationOneToOne } from 'mattermost-redux/types/utilities';
 
 import LoadingScreen from 'components/loading_screen';
 
@@ -25,20 +27,21 @@ import { formatPostsPerDayData, formatUsersWithPostsPerDayData } from '../format
 const LAST_ANALYTICS_TEAM = 'last_analytics_team';
 
 //TODO find out where the types for these are initialized if you do end up using these
-// type Props = {
-//     actions: any;
-//     teams: object[];
-//     stats: object;
-//     locale: string;
-// }
+type Props = {
+    actions: any;
+    teams: Team[];
+    stats: RelationOneToOne<Team, TeamStats>;
+    locale: string;
+}
 
-// type State = {
-//     team: any;
-//     recentlyActiveUsers: [];
-//     newUsers: [];
-// }
+type State = {
+    team: Team | null;
+    recentlyActiveUsers: [];
+    newUsers: [];
+    id: string;
+}
 
-export default class TeamAnalytics extends React.PureComponent {  //? do I have to add the <P,S> parameters here?
+export default class TeamAnalytics extends React.PureComponent<Props, State> {  //? do I have to add the <P,S> parameters here?
     static propTypes = {
 
         /*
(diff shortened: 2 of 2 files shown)


jasonblais:
@d3fau7t Thanks! Let us know if you need any help resolving the error?

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -25,20 +27,21 @@ import { formatPostsPerDayData, formatUsersWithPostsPerDayData } from '../format
...
+}
+
+type State = {
+    team: Team | null;
+    recentlyActiveUsers: [];
+    newUsers: [];
+    id: string;
+}
+
+export default class TeamAnalytics extends React.PureComponent<Props, State> {  //? do I have to add the <P,S> parameters here?
```
d3fau7t, line 44: I've made a couple changes after finding the docs on converting the types, but there's still some ambiguity about some of the types and how the function arguments should be formatted.  I referenced another file: `needs_team.tsx` but there aren't some one-to-one conversions that are obvious to me.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -145,8 +153,8 @@ export default class TeamAnalytics extends React.PureComponent {  //? do I have
         }
 
         const stats = this.props.stats[this.state.team.id];
-        const postCountsDay = formatPostsPerDayData(stats[StatTypes.POST_PER_DAY]);
-        const userCountsWithPostsDay = formatUsersWithPostsPerDayData(stats[StatTypes.USERS_WITH_POSTS_PER_DAY]);
+        const postCountsDay = formatPostsPerDayData(stats[StatTypes.POST_PER_DAY], stats); //! need to sort out the arguments for this function
+        const userCountsWithPostsDay = formatUsersWithPostsPerDayData(stats[StatTypes.USERS_WITH_POSTS_PER_DAY], stats);
```
d3fau7t, line 157: This is one of the instances where I'm a little confused with which parameter should be set as the second argument.  I believe it's currently defined as:
`formatUsersWithPostsPerDayData(labels: string[], data: any)`
`data` as type `any` is ambiguous for me presently

d3fau7t (commented):
Did some updates after discovering the files laying out the types used by y'all, but still have a couple of questions on some aspects I pointed out

components/analytics/team_analytics/team_analytics.tsx
devinbinnie: Yes, the generics you've used here are correct.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -3,28 +3,45 @@
...
 
-import {formatPostsPerDayData, formatUsersWithPostsPerDayData} from '../format';
+import { formatPostsPerDayData, formatUsersWithPostsPerDayData } from '../format';
 
 const LAST_ANALYTICS_TEAM = 'last_analytics_team';
 
-export default class TeamAnalytics extends React.PureComponent {
+//TODO find out where the types for these are initialized if you do end up using these
+type Props = {
+    actions: any;
```
devinbinnie, line 31: You need to explicitly define the actions passed into the component here. The `propTypes` member object below contains the names of all of those, and you can have a look at the redux connector to find where the actual functions are imported from to get the types.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -3,28 +3,45 @@
...
+
+type State = {
+    team: Team | null;
+    recentlyActiveUsers: [];
+    newUsers: [];
+    id: string;
+}
+
+export default class TeamAnalytics extends React.PureComponent<Props, State> {  //? do I have to add the <P,S> parameters here?
     static propTypes = {
```
devinbinnie, line 45: `propTypes` should be removed altogether since TypeScript will handle the type-checking for us.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -43,24 +60,29 @@ export default class TeamAnalytics extends React.PureComponent {
...
+            // getProfilesInTeam: PropTypes.func.isRequired;
+        }
     }
 
-    constructor(props) {
+    constructor(props: Props) {
         super(props);
 
         this.state = {
+            id: this.props.initialId,
```
devinbinnie, line 85: I don't see an `initialId` prop on this component, so this won't pass type-check. If it did, it would change the behaviour of the component which is not our aim here, so we should remove this.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -75,46 +97,46 @@ export default class TeamAnalytics extends React.PureComponent {
...
-        this.props.teams.forEach((t) => {
+        this.props.teams.forEach((t: Team) => {
             if (t.id === teamId) {
                 team = t;
             }
         });
 
         this.setState({
-            team,
+            team: null, //! double check this
```
devinbinnie, line 131: In ES6, if the variable name matches the key name in an object declaration, we can simply pass the key name.
This works out to the same as:
```
this.setState({
            team: team,
});
```

components/analytics/team_analytics/team_analytics.tsx
devinbinnie: So, we shouldn't be adding additional arguments to functions in a TypeScript migration, as this will change the behaviour of the component.

If the function expects a second argument, you can pass in `undefined`.
If the function expects the second argument to be an actual object, let me know as that means our type definitions will need to be updated.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -323,13 +345,13 @@ export default class TeamAnalytics extends React.PureComponent {
     }
 }
 
-export function formatRecentUsersData(data, locale) {
+export function formatRecentUsersData(data: any, locale: any) {
```
devinbinnie, line 348: Looks like `data` in this case is `UserProfile[]` and `locale` is `string`.

Couple helpful hints:
- We should be avoiding using `any` in almost all circumstances, as that doesn't enforce any strict typing. `any` is okay in some of the test files.
- We have a library called `mattermost-redux` that contains the type definitions for most of the types we use in the application, so if you find yourself needing `any` check that library first.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -323,13 +345,13 @@ export default class TeamAnalytics extends React.PureComponent {
...
     if (data == null) {
         return [];
     }
 
-    const formattedData = data.map((user) => {
-        const item = {};
+    const formattedData = data.map((user: any) => {
+        const item: any = {};
```
devinbinnie, line 354: Might need to define a `{name: string, value: string}` type here.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -323,13 +345,13 @@ export default class TeamAnalytics extends React.PureComponent {
...
 
-export function formatRecentUsersData(data, locale) {
+export function formatRecentUsersData(data: any, locale: any) {
     if (data == null) {
         return [];
     }
 
-    const formattedData = data.map((user) => {
-        const item = {};
+    const formattedData = data.map((user: any) => {
```
devinbinnie, line 353: Should be `UserProfile`

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -350,13 +372,13 @@ export function formatRecentUsersData(data, locale) {
     return formattedData;
 }
 
-export function formatNewUsersData(data, locale) {
+export function formatNewUsersData(data: any, locale: any) {
```
devinbinnie, line 375: Same here as above.

components/analytics/team_analytics/team_analytics.tsx
```diff
@@ -350,13 +372,13 @@ export function formatRecentUsersData(data, locale) {
...
     if (data == null) {
         return [];
     }
 
-    const formattedData = data.map((user) => {
-        const item = {};
+    const formattedData = data.map((user: any) => {
```
devinbinnie, line 380: Same here.

components/analytics/team_analytics/index.ts
```diff
@@ -1,20 +1,20 @@
...
-import {getCurrentLocale} from 'selectors/i18n';
+import { getCurrentLocale } from 'selectors/i18n';
 
-import TeamAnalytics from './team_analytics.jsx';
+import TeamAnalytics from './team_analytics';
 
 const LAST_ANALYTICS_TEAM = 'last_analytics_team';
 
-function mapStateToProps(state) {
+function mapStateToProps(state: any) {
```
devinbinnie, line 17: We have a type definition for this called `GlobalState`, found in the `types/store` module

components/analytics/team_analytics/index.ts
```diff
@@ -27,7 +27,7 @@ function mapStateToProps(state) {
     };
 }
 
-function mapDispatchToProps(dispatch) {
+function mapDispatchToProps(dispatch: any) {
```
devinbinnie, line 30: Usually this is either `Dispatch` or `Dispatch<GenericAction>`.
Have a look at some other `index.ts` if you need some ideas of what to use here.

devinbinnie (changes requested):
Thanks for the PR @d3fau7t!
I've added a bunch of comments that I hope will answer all of your questions. Let me know if you have any more.

jfrerich:
Thanks for the PR, @d3fau7t!  Could you please address the feedback from @devinbinnie and feel free to ask for additional assistance!

devinbinnie:
Closing this PR as inactive, @saltshaiker let us know if you plan on finishing this up.

PR #6979, opened 2020-10-30
[MM-27931][MM-30158] Multi-selection and dragging of channels

devinbinnie:
#### Summary
This PR enables multi-selection of channels for dragging and dropping between channels. 
You should be able to click on a channel to select it, Ctrl/Cmd + Click a channel to add to your selection, and Shift+Click to select a grouping of channels.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27931
https://mattermost.atlassian.net/browse/MM-30158


devinbinnie:
Currently a WIP, just showing off the selection process before the drag and drop portion is added.

matthewbirtch:
@devinbinnie I'm thinking we drop the opacity down as you suggested. Let's go with `rgba(var(--sidebar-text-active-border-rgb), 0.24)` for the selected state. It seems to work best for other themes too

devinbinnie:
> @devinbinnie I'm thinking we drop the opacity down as you suggested. Let's go with `rgba(var(--sidebar-text-active-border-rgb), 0.24)` for the selected state. It seems to work best for other themes too

@matthewbirtch Done!

matthewbirtch:
@devinbinnie I think we need a different hover state for items that are selected though. Something like:
```
#SidebarContainer .SidebarLink.selected:hover {
    background: rgba(var(--sidebar-text-active-border-rgb), 0.32);
}
```

devinbinnie:
> @devinbinnie I think we need a different hover state for items that are selected though. Something like:
> 
> ```
> #SidebarContainer .SidebarLink.selected:hover {
>     background: rgba(var(--sidebar-text-active-border-rgb), 0.32);
> }
> ```

@matthewbirtch 👍

esethna:
@matthewbirtch should we allow "esc" to clear the selection?

matthewbirtch:
> @matthewbirtch should we allow "esc" to clear the selection?

@esethna I'm not sure about that. I suppose we could, but it doesn't seem to be a common pattern with multi-select experiences. Just looked at Figma's layers pannel, Mac OS Finder, Evernote and the ESC key doesn't appear to clear the selection in those cases.

Having said that, it might be nice to be able to clear the selection with one key and the ESC key would be an appropriate key. I suppose we could try it and see if we run in to any problems with it.

esethna:
@devinbinnie @matthewbirtch Agree, I think having an easy way to clear the selection is important ^

esethna (changes requested):
@devinbinnie a few things I just noticed:

1. I think there's a bug where shift selection is not working after refresh. ie refresh the page, then hold shift and click a channel, notice that it does a ctrl selection instead of a shift selection. If I change channels in the lhs then use shift select it works as expected.

image

2. As discussed above I'm 4/5 we should add `esc` to clear the selection given these selectable items are also "buttons"

hmhealey:
Initial dev review of this looks good to me

devinbinnie:
> @devinbinnie a few things I just noticed:
> 
> 1. I think there's a bug where shift selection is not working after refresh. ie refresh the page, then hold shift and click a channel, notice that it does a ctrl selection instead of a shift selection. If I change channels in the lhs then use shift select it works as expected.
> 
> image
> 

@esethna This is something I discussed with Matt. What you're expecting is that the Shift select should work off of the active channel, and that the active channel should be selected. That's no longer the case, you'll have to first select a channel by clicking on it such that it changes to the selected colour (it's something like the sidebar border colour) and then Shift+Click another channel to do the shift select.

> 1. As discussed above I'm 4/5 we should add `esc` to clear the selection given these selectable items are also "buttons"

Sure, I can do that.

devinbinnie:
@esethna @matthewbirtch @hmhealey @nevyangelova Ready for review!

matthewbirtch (changes requested):
Nice work @devinbinnie. Feels pretty great to me.

1. For the "N selected" tag, rather than a theme color, I'm thinking we just use the same color we use for tooltips (#000000). I've updated the Figma file to reflect this. Right now it doesn't work well on dark themes.
image

2. As I play with it, there are two things that throw me off sometimes:
  a) The currently active channel looks like it's also selected in my multi-select selection. The states might be too similar and we might want to make the selection style more obvious in the future, but I'm ok with testing this out and looking at a possible enhancement in the future.
  b) When I shift-click, for some reason I expect that the active channel should be part of the selection and a shift-click should select everything between the active channel and the channel I shift-click. I believe @esethna brought this up previously. 

I'm not entire sure how we solve the above, but wanted to flag them as possible issues that users might identify

devinbinnie:
@matthewbirtch 

> 1. For the "N selected" tag, rather than a theme color, I'm thinking we just use the same color we use for tooltips (#000000). I've updated the Figma file to reflect this. Right now it doesn't work well on dark themes.

I think this is because we're mixing styles between the sidebar and the center channel, which don't always line up in any theme. I've changed the text `span` to use `center-channel-bg`, so that it will be the inverse of the center channel and provide a decent contrast. Let me know if you think that works, otherwise we can use the default but it might stand out kinda funny on higher contrast themes.

> 2. As I play with it, there are two things that throw me off sometimes:
>    a) The currently active channel looks like it's also selected in my multi-select selection. The states might be too similar and we might want to make the selection style more obvious in the future, but I'm ok with testing this out and looking at a possible enhancement in the future.

Understandable, perhaps we need to think of a better indicator that a channel is "selected for multiselect" vs. "is the active channel".

>    b) When I shift-click, for some reason I expect that the active channel should be part of the selection and a shift-click should select everything between the active channel and the channel I shift-click. I believe @esethna brought this up previously.

Okay we can do this. My question though, is the expectation to always have the active channel be part of the selection? Because it's currently possible to unselect all by pressing escape or clicking somewhere that isn't one of the channels.

I think this ties into a) a fair bit and perhaps if we can come up with a better way of discerning 'selected' from 'active' then we can avoid this.

matthewbirtch:
@devinbinnie, I'm thinking we could try making it super obvious for the selected state. Here's what I'm thinking:

```
#SidebarContainer .SidebarLink.selected {
    background: rgba(var(--sidebar-text-active-border-rgb), 1);
    color: rgba(var(--sidebar-text-rgb),1);
}

#SidebarContainer .SidebarLink.selected:hover {
    background: rgba(var(--sidebar-text-active-border-rgb), 0.88);
}

#SidebarContainer .SidebarChannel.active .SidebarLink.selected {
    background: rgba(var(--sidebar-text-active-border-rgb), 1);
}

#SidebarContainer .SidebarChannel.active .SidebarLink.selected:hover {
    background: rgba(var(--sidebar-text-active-border-rgb), 0.88);
}

```

devinbinnie:
> @devinbinnie, I'm thinking we could try making it super obvious for the selected state. Here's what I'm thinking:
> 
> ```
> #SidebarContainer .SidebarLink.selected {
>     background: rgba(var(--sidebar-text-active-border-rgb), 1);
>     color: rgba(var(--sidebar-text-rgb),1);
> }
> 
> #SidebarContainer .SidebarLink.selected:hover {
>     background: rgba(var(--sidebar-text-active-border-rgb), 0.88);
> }
> 
> #SidebarContainer .SidebarChannel.active .SidebarLink.selected {
>     background: rgba(var(--sidebar-text-active-border-rgb), 1);
> }
> 
> #SidebarContainer .SidebarChannel.active .SidebarLink.selected:hover {
>     background: rgba(var(--sidebar-text-active-border-rgb), 0.88);
> }
> ```

@matthewbirtch Done, have a look. Though personally I think we should tone it down a little bit.

matthewbirtch:
> @matthewbirtch Done, have a look. Though personally I think we should tone it down a little bit.

Fair point. Maybe adjust:

```
#SidebarContainer .SidebarLink.selected {
    background: rgba(var(--sidebar-text-active-border-rgb), .64);
    color: rgba(var(--sidebar-text-rgb),1);
}

#SidebarContainer .SidebarLink.selected:hover {
    background: rgba(var(--sidebar-text-active-border-rgb), 0.72);
}

#SidebarContainer .SidebarChannel.active .SidebarLink.selected {
    background: rgba(var(--sidebar-text-active-border-rgb), 1);
}

#SidebarContainer .SidebarChannel.active .SidebarLink.selected:hover {
    background: rgba(var(--sidebar-text-active-border-rgb), 0.88);
}
```

actions/views/channel_sidebar.ts
```diff
@@ -93,10 +97,74 @@ export function adjustTargetIndexForMove(state: GlobalState, categoryId: string,
...
+        // Nothing already selected
+        if (!selectedChannelIds.length) {
+            return dispatch({
+                type: ActionTypes.MULTISELECT_CHANNEL,
+                data: channelId,
+            });
+        }
+
+        const allChannelsIdsInOrder = getChannelsInCategoryOrder(state).map((channel) => channel.id);
+        const indexOfNew: number = allChannelsIdsInOrder.indexOf(channelId);
```
nevyangelova, line 145: Could we check `channelId === lastSelected` here instead?

actions/views/channel_sidebar.ts
```diff
@@ -93,10 +97,74 @@ export function adjustTargetIndexForMove(state: GlobalState, categoryId: string,
...
+        // multi selecting in the same column
+        // need to select everything between the last index and the current index inclusive
+
+        // nothing to do here
+        if (indexOfNew === indexOfLast) {
+            return null;
+        }
+
+        const isSelectingForwards: boolean = indexOfNew > indexOfLast;
+        const start: number = isSelectingForwards ? indexOfLast : indexOfNew;
```
nevyangelova, line 157: could this be `start = Math.min(indexOfLast, indexOfNew)` ?

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -396,8 +409,24 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
 
         if (result.reason === 'DROP' && result.destination) {
             if (result.type === 'SIDEBAR_CHANNEL') {
-                this.props.actions.moveChannelInSidebar(result.destination.droppableId, result.draggableId, result.destination.index);
-                trackEvent('ui', 'ui_sidebar_dragdrop_dropped_channel');
+                let channelsToMove = [result.draggableId];
+                if (this.props.selectedChannelIds.length) {
+                    // Filter out channels that can't go in the category specified
+                    const targetCategory = this.props.categories.find((category) => category.id === result.destination?.droppableId);
+                    channelsToMove = [...this.props.selectedChannelIds.filter((channelId) => {
```
nevyangelova, line 416: do we need this spread?

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -396,8 +409,24 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
...
+                        const selectedChannel = this.props.displayedChannels.find((channel) => channelId === channel.id);
+                        const isDMGM = selectedChannel?.type === General.DM_CHANNEL || selectedChannel?.type === General.GM_CHANNEL;
+                        return targetCategory?.type === 'custom' || (isDMGM && targetCategory?.type === 'direct_messages') || (!isDMGM && targetCategory?.type !== 'direct_messages');
+                    })];
+
+                    // Reorder such that the channels move in the order that they appear in the sidebar
+                    const displayedChannelIds = this.props.displayedChannels.map((channel) => channel.id);
+                    channelsToMove = channelsToMove.sort((a, b) => displayedChannelIds.indexOf(a) - displayedChannelIds.indexOf(b));
```
nevyangelova, line 424: Don't need to assign, sort will mutate the original array.

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -396,8 +409,24 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
...
+                }
+                this.props.actions.moveChannelsInSidebar(result.destination.droppableId, channelsToMove, result.destination.index, result.draggableId);
+
+                // Remove selection from channels that were moved
+                channelsToMove.forEach((channelId) => this.props.actions.multiSelectChannelAdd(channelId));
```
nevyangelova, line 429: Do we have an action that takes an array instead?

reducers/views/channel_sidebar.ts
```diff
@@ -75,8 +75,62 @@ export function newCategoryIds(state: string[] = [], action: GenericAction): str
...
+        if (state.length > 1) {
+            return [action.data];
+        }
+
+        // task was previously selected but not in a group
+        // we will now clear the selection
+        return [];
+    case ActionTypes.MULTISELECT_CHANNEL_ADD:
+        // if not selected - add it to the selected items
+        if (state.indexOf(action.data) === -1) {
```
nevyangelova, line 98: The state of this could be a javascript Set so then we don't have to do the checks

nevyangelova (approved):
LGTM just a few non-blocking comments :)

actions/views/channel_sidebar.ts
devinbinnie: We're getting the index here for selection, I don't think that will help :P

actions/views/channel_sidebar.ts
devinbinnie: Ooh, that's a nice optimization!

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
devinbinnie: Nope!

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
devinbinnie: We don't, this does the same thing as an action like that would do.
I would still have to remove each element from the array individually.

reducers/views/channel_sidebar.ts
devinbinnie: This feels like something we could actually change across the board if it makes sense. Want to talk about it in hangouts or something?

actions/views/channel_sidebar.ts
```diff
@@ -58,33 +58,37 @@ export function createCategory(teamId: string, displayName: string, channelIds?:
...
         return 0;
     }
 
     const category = getCategory(state, categoryId);
     const filteredChannels = makeGetChannelsForCategory()(state, category);
+    const filteredChannelIds = filteredChannels.map((channel) => channel.id);
+
+    // When dragging multiple channels, we don't actually remove all of them from the list as react-beautiful-dnd doesn't support that
+    // Account for channels removed above the insert point, except the one currently being dragged which is already accounted for by react-beautiful-dnd
+    const removedChannelsAboveInsert = filteredChannelIds.filter((channel, index) => channel !== draggableChannelId && channelIds.indexOf(channel) !== -1 && index <= targetIndex);
```
hmhealey, line 81: So targetIndex includes the other channels being moved from above this because react-beautiful-dnd doesn't know we removed them, so we do this to update the `targetIndex` to account for that, correct?

actions/views/channel_sidebar.ts
```diff
@@ -58,33 +58,37 @@ export function createCategory(teamId: string, displayName: string, channelIds?:
...
 
     if (category.channel_ids.length === filteredChannels.length) {
         // There are no archived channels in the category, so the targetIndex from react-beautiful-dnd will be correct
-        return targetIndex;
+        return targetIndex - removedChannelsAboveInsert.length;
```
hmhealey, line 85: 0/5 but perhaps we can give this a new name since it's repeated quite a few times. Maybe we rename `targetIndex` and just this `targetIndex` going forward to reduce the diff.

actions/views/channel_sidebar.ts
```diff
@@ -93,10 +97,73 @@ export function adjustTargetIndexForMove(state: GlobalState, categoryId: string,
...
     if (sourceIndex !== -1 && sourceIndex < newIndex) {
         newIndex -= 1;
     }
 
-    return newIndex;
+    return Math.max(newIndex - removedChannelsAboveInsert.length, 0);
+}
+
+export function clearChannelSelection() {
+    return (dispatch: DispatchFunc) => {
```
hmhealey, line 109: Can this just return the action without dispatching it or does that run into some weird typing issue?

reducers/views/channel_sidebar.ts
```diff
@@ -75,8 +75,62 @@ export function newCategoryIds(state: string[] = [], action: GenericAction): str
     }
 }
 
+export function selectedChannelIds(state: string[] = [], action: GenericAction): string[] {
```
hmhealey, line 78: 0/5, but we're running into a bit of a naming conflict here since we have a concept of a "selected post/channel" which are for the post visible in the RHS similar to the actions. `multiselectedChannelIds` might be a better name if this is only used when actually multiselecting

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -151,6 +159,19 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
         let wrappedComponent: React.ReactNode;
 
         if (isDraggable) {
+            let selectedCount: React.ReactNode | undefined;
```
hmhealey, line 162: I think the `| undefined` might be unnecessary. I think ReactNode is nullable on its own

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -362,6 +365,10 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
         const droppable = [...document.querySelectorAll<HTMLDivElement>('[data-rbd-droppable-id*="droppable-categories"]')];
         droppable[0].style.height = `${droppable[0].scrollHeight}px`;
 
+        if (!this.props.selectedChannelIds.find((id) => before.draggableId === id)) {
```
hmhealey, line 368: I was going to suggest `some` since it returns a boolean, but it turns out they perform the same. Nothing to change here, but I thought that might be interesting to know.

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -370,8 +377,14 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
         if (this.props.categories.some((category) => category.id === before.draggableId)) {
             draggingState.type = DraggingStateTypes.CATEGORY;
         } else {
-            const draggingChannel = this.props.displayedChannels.find((channel) => channel.id === before.draggableId);
-            draggingState.type = (draggingChannel?.type === General.DM_CHANNEL || draggingChannel?.type === General.GM_CHANNEL) ? DraggingStateTypes.DM : DraggingStateTypes.CHANNEL;
+            const draggingChannels = this.props.displayedChannels.filter((channel) => this.props.selectedChannelIds.indexOf(channel.id) !== -1 || channel.id === before.draggableId);
```
hmhealey, line 380: 0/5, but is there maybe a way to do this without looping through the array so many times?

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -396,8 +409,24 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
...
-                this.props.actions.moveChannelInSidebar(result.destination.droppableId, result.draggableId, result.destination.index);
-                trackEvent('ui', 'ui_sidebar_dragdrop_dropped_channel');
+                let channelsToMove = [result.draggableId];
+                if (this.props.selectedChannelIds.length) {
+                    // Filter out channels that can't go in the category specified
+                    const targetCategory = this.props.categories.find((category) => category.id === result.destination?.droppableId);
+                    channelsToMove = this.props.selectedChannelIds.filter((channelId) => {
+                        const selectedChannel = this.props.displayedChannels.find((channel) => channelId === channel.id);
+                        const isDMGM = selectedChannel?.type === General.DM_CHANNEL || selectedChannel?.type === General.GM_CHANNEL;
+                        return targetCategory?.type === 'custom' || (isDMGM && targetCategory?.type === 'direct_messages') || (!isDMGM && targetCategory?.type !== 'direct_messages');
```
hmhealey, line 419: I think we have constants defined for these

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -396,8 +409,24 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
 
         if (result.reason === 'DROP' && result.destination) {
             if (result.type === 'SIDEBAR_CHANNEL') {
-                this.props.actions.moveChannelInSidebar(result.destination.droppableId, result.draggableId, result.destination.index);
-                trackEvent('ui', 'ui_sidebar_dragdrop_dropped_channel');
+                let channelsToMove = [result.draggableId];
+                if (this.props.selectedChannelIds.length) {
+                    // Filter out channels that can't go in the category specified
+                    const targetCategory = this.props.categories.find((category) => category.id === result.destination?.droppableId);
```
hmhealey, line 415: Perhaps this logic could be handled by `moveChannelsInSidebar`? I wonder if we might have easier access to some of this data in the `action`. Also, it might be easier to test if we do that since we can just give `moveChannelsInSidebar` an arbitrary list of IDs and see what happens.

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -396,8 +409,24 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
 
         if (result.reason === 'DROP' && result.destination) {
             if (result.type === 'SIDEBAR_CHANNEL') {
-                this.props.actions.moveChannelInSidebar(result.destination.droppableId, result.draggableId, result.destination.index);
-                trackEvent('ui', 'ui_sidebar_dragdrop_dropped_channel');
```
hmhealey, line 400: This telemetry event looks like it's been removed by mistake

reducers/views/channel_sidebar.ts
```diff
@@ -75,8 +75,62 @@ export function newCategoryIds(state: string[] = [], action: GenericAction): str
...
+    switch (action.type) {
+    case ActionTypes.MULTISELECT_CHANNEL:
+        // Task was not previously selected
```
hmhealey, line 81: Can you update these comments to refer to channels? It's a bit confusing to see task-related stuff that I assume that came from some demo code

hmhealey (changes requested):

reducers/views/channel_sidebar.ts
hmhealey: I'm trying to remember if we use Set or Map anywhere else in the redux code. I think we might have tried in redux or the mobile app, but we ran into issues serializing the redux state for storage on the mobile app.

Another thing to consider is that I think it's easier to treat literal arrays/objects as "immutable" than it is with Set since you can just do things like `[...arr, 123]` to add an item to an array, but it's a few more steps to do "immutably" with a Set.

esethna:
@matthewbirtch 3/5 but it feels pretty odd to me that the active channel is not part of the selection, since I initiate the action using a shift-click or ctrl-click (meaning I have the intention to multi-select). So if it doesn't multiselect it feels broken to me

actions/views/channel_sidebar.ts
```diff
@@ -58,33 +58,37 @@ export function createCategory(teamId: string, displayName: string, channelIds?:
...
+    return (dispatch: DispatchFunc, getState: () => GlobalState) => {
+        const newIndex = adjustTargetIndexForMove(getState(), categoryId, channelIds, targetIndex, draggableChannelId);
 
-        return dispatch(moveChannelToCategory(categoryId, channelId, newIndex));
+        return dispatch(moveChannelsToCategory(categoryId, channelIds, newIndex));
     };
 }
 
-export function adjustTargetIndexForMove(state: GlobalState, categoryId: string, channelId: string, targetIndex: number) {
+export function adjustTargetIndexForMove(state: GlobalState, categoryId: string, channelIds: string[], targetIndex: number, draggableChannelId: string) {
```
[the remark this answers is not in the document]
devinbinnie: Correct.

actions/views/channel_sidebar.ts
devinbinnie: Correct.

actions/views/channel_sidebar.ts
devinbinnie: What would you suggest?

actions/views/channel_sidebar.ts
devinbinnie: Weird typing issue, because it expects the object returned to be of type: `{type: string, data: any}`
Since there's no `data` here, I have to wrap it.

reducers/views/channel_sidebar.ts
devinbinnie: Yeah I think that makes sense. Will change.

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
devinbinnie: I do enjoy facts :D

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
devinbinnie: We could do a `reduce` to compile an object such that the call to `indexOf` can be replaced with an object lookup, but I don't think it's a huge performance gain.

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
devinbinnie: Yeah I think that's a smart refactor, there's a lot of logic in the component here. I'll make the change.

matthewbirtch (approved):
I think this is working really well now @devinbinnie. Great job

esethna (approved):
Great, thanks guys

esethna:
@matthewbirtch @devinbinnie what about the ... menu items once multi-selected? I think we proposed limiting the options and having the move to > move all the channels? https://docs.google.com/document/d/17Uw_HMptsXADgganXffjB1k8_WuAn5nyOU5JrqZj_AA/edit#bookmark=id.3idkksu9xrrp

image

actions/views/channel_sidebar.ts
hmhealey: Strange. I'm surprised undefined doesn't satisfy any. I guess it just doesn't satisfy it implicitly.

hmhealey (approved):

matthewbirtch:
> @matthewbirtch @devinbinnie what about the ... menu items once multi-selected? I think we proposed limiting the options and having the move to > move all the channels? https://docs.google.com/document/d/17Uw_HMptsXADgganXffjB1k8_WuAn5nyOU5JrqZj_AA/edit#bookmark=id.3idkksu9xrrp

Crap, I totally missed this (sorry, sort of catching up from the stuff Andrew did still). @devinbinnie do we want to do this work in another PR? Might be best at this point, but I'll leave that up to you

devinbinnie:
> > @matthewbirtch @devinbinnie what about the ... menu items once multi-selected? I think we proposed limiting the options and having the move to > move all the channels? https://docs.google.com/document/d/17Uw_HMptsXADgganXffjB1k8_WuAn5nyOU5JrqZj_AA/edit#bookmark=id.3idkksu9xrrp
> 
> Crap, I totally missed this (sorry, sort of catching up from the stuff Andrew did still). @devinbinnie do we want to do this work in another PR? Might be best at this point, but I'll leave that up to you

@matthewbirtch @esethna 
We actually have a completely separate ticket for the menu items: https://mattermost.atlassian.net/browse/MM-30159

PR #6982, opened 2020-10-31
GH-16153: Typescript components/logged in

swatikothari:
Fixes https://github.com/mattermost/mattermost-server/issues/16153

### DIFF
--- a/components/logged_in/index.ts
+++ b/components/logged_in/index.ts
@@ -1,0 +1,56 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {connect} from 'react-redux';
+import {bindActionCreators, Dispatch} from 'redux';
+
+import {autoUpdateTimezone} from 'mattermost-redux/actions/timezone';
+import {getCurrentChannelId} from 'mattermost-redux/selectors/entities/channels';
+import {getLicense, getConfig} from 'mattermost-redux/selectors/entities/general';
+import {getCurrentUser, shouldShowTermsOfService} from 'mattermost-redux/selectors/entities/users';
+import {GenericAction} from 'mattermost-redux/types/actions';
+import {Channel} from 'mattermost-redux/types/channels';
+
+import {browserHistory} from 'utils/browser_history';
+import {checkIfMFARequired} from 'utils/route';
+import {getChannelURL} from 'utils/utils';
+
+import {GlobalState} from 'types/store';
+
+import LoggedIn from './logged_in';
+
+export declare type GetStateFunc = () => GlobalState;
+
+type Props = {
+    match: {
+        url: string;
+    }
+}
+
+function mapStateToProps(state: GlobalState, ownProps: Props) {
+    const license = getLicense(state);
+    const config = getConfig(state);
+    const showTermsOfService = shouldShowTermsOfService(state);
+
+    return {
+        currentUser: getCurrentUser(state),
+        currentChannelId: getCurrentChannelId(state),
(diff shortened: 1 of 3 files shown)


hanzei:
@swatikothari Also thanks for this PR :+1: 

There is one small linter issue:
```
/root/mattermost/mattermost-webapp/components/logged_in/logged_in.tsx
  20:1  error  More than 1 blank line not allowed  no-multiple-empty-lines
```

and another type issue:
```

components/logged_in/logged_in.tsx:8:20 - error TS7016: Could not find a declaration file for module 'semver'. '/root/mattermost/mattermost-webapp/node_modules/semver/index.js' implicitly has an 'any' type.
  Try `npm install @types/semver` if it exists or add a new declaration (.d.ts) file containing `declare module 'semver';`

8 import semver from 'semver';
                     ~~~~~~~~
```

Could you please take a look at them?

components/logged_in/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import {browserHistory} from 'utils/browser_history';
+import {checkIfMFARequired} from 'utils/route';
+import {getChannelURL} from 'utils/utils';
+
+import {GlobalState} from 'types/store';
+
+import LoggedIn from './logged_in';
+
+export declare type GetStateFunc = () => GlobalState;
```
hmhealey, line 22: This type already exists in `mattermost-redux/types/actions`, so it can be imported from there

components/logged_in/logged_in.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+import LoggedIn from 'components/logged_in/logged_in';
+import BrowserStore from 'stores/browser_store';
+
+jest.mock('actions/websocket_actions.jsx', () => ({
+    initialize: jest.fn(),
+}));
+
+BrowserStore.signalLogin = jest.fn();
+
+type Props = {
```
hmhealey, line 18: Instead of repeating this type definition, you can export it from `logged_in.tsx` and re-import it here

components/logged_in/logged_in.tsx
```diff
@@ -0,0 +1,209 @@
...
+import BrowserStore from 'stores/browser_store';
+
+
+const dispatch = store.dispatch;
+const getState = store.getState;
+
+const BACKSPACE_CHAR = 8;
+
+type Props = {
+    currentUser?: UserProfile | Record<string, unknown> | null,
```
hmhealey, line 27: This should just be `currentUser?: UserProfile` if possible. I don't think `getUserProfile` can return null, and we ideally shouldn't support arbitrary objects here. If you run into issues because the tests are trying to pass an empty object, it'd be better to have the test use `{} as UserProfile` so that the non-test code will be more strictly typed.

components/logged_in/logged_in.tsx
```diff
@@ -0,0 +1,209 @@
...
+    currentChannelId?: string,
+    children?: JSX.Element,
```
hmhealey, line 29: Unless we have a reason not to, `children` should generally be typed as `React.ReactNode` since that supports strings and functions as children.

hmhealey (changes requested):
Thanks for the PR, @swatikothari! I just have a couple small suggestions for changes, but everything looks good for the most part.

Could you also remove the existing `js` files that these are replacing? I think our webpack prefers js files right now, so they might be in use when we're testing.

Willyfrog (approved):
other than @hmhealey comments, looks good

Willyfrog:
hey @swatikothari , let us know if there is anything we can do to help move this forward or clarify any of the comments

devinbinnie:
Closing this PR as inactive, @swatikothari let us know if you plan on finishing up this PR.

PR #6992, opened 2020-11-01
[MM-30174] Migrate changeCSS() to CSS variable in utils/utils.jsx, ln. 708

IanvsPoplicola:
#### Summary
Migrate a CSS variable for popover from `utils/utils.jsx` into `sass/components/_popover.scss`

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16178

### DIFF
--- a/sass/components/_popover.scss
+++ b/sass/components/_popover.scss
@@ -72,6 +72,7 @@
     &.left {
         > .arrow {
             display: none;
+            border-left-color: rgba(var(--center-channel-color), 0.25);
 
             &:after {
                 border-color: transparent;
--- a/sass/components/_popover.scss
+++ b/sass/components/_popover.scss
@@ -72,7 +72,6 @@
     &.left {
         > .arrow {
             display: none;
-            border-left-color: rgba(var(--center-channel-color-rgb), 0.25);
 
             &:after {
                 border-color: transparent;
@@ -82,6 +81,12 @@
 
     &.bottom {
         margin-top: 10px;
+    }
+
+    &.left {
+        > .arrow {
+            border-left-color: rgba(var(--center-channel-color-rgb), 0.25);
+        }
     }
 
     ul + p,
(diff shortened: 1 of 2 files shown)


sass/components/_popover.scss
```diff
@@ -72,6 +72,7 @@
     &.left {
         > .arrow {
             display: none;
+            border-left-color: rgba(var(--center-channel-color), 0.25);
```
asaadmahmood, line 75: Please use the `--center-channel-color-rgb` variable.

asaadmahmood (changes requested):

sass/components/_popover.scss
```diff
@@ -72,6 +72,7 @@
     &.left {
         > .arrow {
             display: none;
+            border-left-color: rgba(var(--center-channel-color-rgb), 0.25);
```
hmhealey, line 75: This is adding a border-left-color to elements that didn't have it before since only `.left > .arrow` had it before, but I'm not sure if that actually causes any problems here.

@asaadmahmood Are you aware of why we specified different borders for each version of the arrow before? Perhaps we could just set `border-color` for all of them at once.

If we can't do that, then this should be moved into its own rule to only target `.left > .arrow`

IanvsPoplicola: The CSS style is now moved to the correct place, only targeting `.left > .arrow`. I can also change it to set `border-color` once if that's preferred!

asaadmahmood (approved):

agarciamontoro (approved):
Thank you for your contribution, @IanvsPoplicola! This looks good to me :tada:

hmhealey (approved):

sass/components/_popover.scss
hmhealey: I think it's safe to leave it as-is. I'm a bit worried about changing to just `border-color` since this is quite old CSS. Thanks for moving it though!

PR #6993, opened 2020-11-01
[MM-30173] Migrate changeCSS() to CSS variable in utils/utils.jsx, ln. 707

IanvsPoplicola:
#### Summary
Migrate a CSS variable for popover from `utils/utils.jsx` into `sass/components/_popover.scss`

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16177

### DIFF
--- a/sass/components/_popover.scss
+++ b/sass/components/_popover.scss
@@ -70,7 +70,7 @@
 
     &.right {
         > .arrow {
-            border-right-color: rgba(var(--center-channel-color), 0.25);
+            border-right-color: rgba(var(--center-channel-color-rgb), 0.25);
         }
     }
     &.top,
--- a/sass/components/_popover.scss
+++ b/sass/components/_popover.scss
@@ -67,12 +67,7 @@
     }
 
     &.bottom,
-
-    &.right {
-        > .arrow {
-            border-right-color: rgba(var(--center-channel-color-rgb), 0.25);
-        }
-    }
+    &.right,
     &.top,
     &.left {
         > .arrow {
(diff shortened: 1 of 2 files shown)


sass/components/_popover.scss
```diff
@@ -67,7 +67,12 @@
     }
 
     &.bottom,
-    &.right,
+
+    &.right {
+        > .arrow {
+            border-right-color: rgba(var(--center-channel-color), 0.25);
```
asaadmahmood, line 73: Please use the --center-channel-color-rgb variable.

asaadmahmood (changes requested):

crspeller (approved):
Other than @asaadmahmood comment.

IanvsPoplicola:
@asaadmahmood Thank you for the review! I've updated the PR as per your comment :)

sass/components/_popover.scss
```diff
@@ -67,7 +67,12 @@
     }
 
     &.bottom,
-    &.right,
+
+    &.right {
```
hmhealey, line 71: I think this might result in similar behaviour to the things I've commented on in your other two PRs. Unless we can just set a `border-color` for all of these rules, we'll have to add separate rules for `.bottom`, `.right`, and so on.

IanvsPoplicola: Yup totally, I've updated all 4 PRs to place the CSS styles in the correct place. Thank you for the sharp spotting!

asaadmahmood (approved):

asaadmahmood (approved):

crspeller (approved):

hmhealey (approved):

PR #6997, opened 2020-11-02
Migrate components/analytics/system_analytics module to TypeScript

Ampit:
#### Summary

This converts the system_analytics component/module in web app to TypeScript.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/16164


components/analytics/system_analytics/system_analytics.tsx
```diff
@@ -313,7 +315,7 @@ export default class SystemAnalytics extends React.PureComponent {
                     />
                 }
                 icon='fa-globe'
-                count={stats[StatTypes.TOTAL_PUBLIC_CHANNELS] + stats[StatTypes.TOTAL_PRIVATE_GROUPS]}
+                count={Number(stats[StatTypes.TOTAL_PUBLIC_CHANNELS]) + Number(stats[StatTypes.TOTAL_PRIVATE_GROUPS])}
```
Willyfrog, line 318: since stats can be either a number or a analitycsRow, won't you get a NaN out of it?
this applies to other changes as well, but this might be more problematic as you might end up with NaN+NaN

Willyfrog (changes requested):
Thanks for working in this PR, I've added a question on how to handle the different options of the declared type.

components/analytics/system_analytics/system_analytics.tsx
Ampit: I'm not really sure how to tackle this.

Any suggestions would be appreciated ?

Thanks for your input.

Willyfrog: I'd probably use a helper function to ensure it extracts the number out of it wether it is one type or the other.

Doing a quick search I found [this post](https://thoughtbot.com/blog/the-case-for-discriminated-union-types-with-typescript) that might help you. But let me know either here or in the comunity server if you want to discuss it further :)

Ampit: So check if it is a number then use that otherwise if it is an analyticsRow extract the value out of that object { name: string, value: number} and use that ?

Also looking at the StatisticCount component it expects the count prop to be a number and if it's not a number it just renders loading message. So i guess it never was using the analyticsrow type... I'm not sure adding those checks makes any difference ?

does that makes sense or am i completely off here.. please let me know.

Willyfrog: > So check if it is a number then use that otherwise if it is an analyticsRow extract the value out of that object { name: string, value: number} and use that ?

yes, I think that would be the approach here.

> 
> Also looking at the StatisticCount component it expects the count prop to be a number and if it's not a number it just renders loading message. So i guess it never was using the analyticsrow type... I'm not sure adding those checks makes any difference ?

I think the difference would be that the signature `Dictionary<number | AnalyticsRow[]>` allows it to happen, and if it happens it'll work, instead of showing an eternal "loading..." message that might confuse users. originally it was simply an object with no definition and as such no real constraint for what is accepted. What made you choose `Dictionary<number | AnalyticsRow[]>` as the type for stats?

Ampit: I saw it with those types in other components. Also while inspecting it.. data looks similar to that type.

Willyfrog: I'm fine with it, but then we have to be able to handle the info :)

so your approach would probably be the way to move forward

Ampit: // Helper Function
private checkStat(stat: number | AnalyticsRow[]):number {
    if (typeof stat === 'number') {    
            return stat;    
        }    
        return stat[0].value;    
}

// Usage

count={this.checkStat(stats[StatTypes.TOTAL_PUBLIC_CHANNELS]) + this.checkStat(stats[StatTypes.TOTAL_PRIVATE_GROUPS])}

I came up with this ?

Willyfrog: yep, that should work!
I'd probably change the name to `getStatValue` or something along the lines to give a more meaningful name.

Ampit: Done

Willyfrog (approved):
nice!

nevyangelova (commented):
LGTM !

nevyangelova (approved):

jasonblais:
@jgilliam17 friendly reminder to help with a QA review :)

jgilliam17:
Thank you @Ampit and apologize for delay in getting to this PR
I am seeing an error on Site Statistics screen after upgrading license to Enterprise Edition. Can you please take a look? Thank you. 
Screen Shot 2020-12-01 at 8 36 53 AM

jgilliam17:
One more thing - I am not seeing all the values in the System Statistics  
I compared with master.test server (first image), only top 6 values display on the PR server (bottom image)
master test

Screen Shot 2020-12-01 at 10 12 59 AM

Ampit:
@jgilliam17 Thanks for your review.

I'm not sure how i can check my code for enterprise edition on my local dev setup ?

jgilliam17:
> I'm not sure how i can check my code for enterprise edition on my local dev setup ?

@nevyangelova or @Willyfrog would you be able to give guidance here? Thanks.

components/analytics/system_analytics/system_analytics.tsx
```diff
@@ -43,8 +45,15 @@ export default class SystemAnalytics extends React.PureComponent {
...
     }
 
-    render() {
-        const stats = this.props.stats;
+    private getStatValue(stat: number | AnalyticsRow[]):number {
+        if (typeof stat === 'number') {
+            return stat;
+        }
+        return stat[0].value;
+    }
```
Willyfrog, line 53: The problem is not with the enterprise code, but that we are assuming there is a first element always and it might be undefined.

```suggestion
    private getStatValue(stat: number | AnalyticsRow[]):number? {
        if (typeof stat === 'number') {
            return stat;
        }
        if (!stat) {
            return undefined;
        }
        return stat[0].value;
    }
```
by using undefined, you'll have to deal with it when presenting the data as being "not available".

This is one possible approach, but there are other options, so don't feel forced to use this one if you come up with some other idea.

Willyfrog (changes requested):

components/analytics/system_analytics/system_analytics.tsx
```diff
@@ -313,7 +322,7 @@ export default class SystemAnalytics extends React.PureComponent {
                     />
                 }
                 icon='fa-globe'
-                count={stats[StatTypes.TOTAL_PUBLIC_CHANNELS] + stats[StatTypes.TOTAL_PRIVATE_GROUPS]}
+                count={this.getStatValue(stats[StatTypes.TOTAL_PUBLIC_CHANNELS]) + this.getStatValue(stats[StatTypes.TOTAL_PRIVATE_GROUPS])}
             />
         );
```
Ampit, line 328: @Willyfrog 
WIth that i have to do something like this to resolve this ?

        const totalPublicChannelsCount = this.getStatValue(stats[StatTypes.TOTAL_PUBLIC_CHANNELS]);
        const totalPrivateGroupsCount = this.getStatValue(stats[StatTypes.TOTAL_PRIVATE_GROUPS]);
        const totalChannelCount = () => {
            if (totalPublicChannelsCount && totalPrivateGroupsCount) {
                return totalPublicChannelsCount + totalPrivateGroupsCount;
            } else if (!totalPublicChannelsCount && totalPrivateGroupsCount) {
                return totalPrivateGroupsCount;
            } else if (totalPublicChannelsCount && !totalPrivateGroupsCount) {
                return totalPublicChannelsCount;
            }
            return undefined;
        };
        const channelCount = (
            <StatisticCount
                id='totalChannels'
                title={
                    <FormattedMessage
                        id='analytics.system.totalChannels'
                        defaultMessage='Total Channels'
                    />
                }
                icon='fa-globe'
                count={totalChannelCount()}
            />
        );

Willyfrog: yes, and also make sure that `StatisticCount` can handle undefined

Ampit: Alright will do

Ampit: StatisticCount does handle undefined count values.

I have pushed new changes.

components/analytics/system_analytics/system_analytics.tsx
```diff
@@ -43,8 +45,18 @@ export default class SystemAnalytics extends React.PureComponent {
         }
     }
 
-    render() {
-        const stats = this.props.stats;
+    private getStatValue(stat: number | AnalyticsRow[]):number | undefined {
+        if (typeof stat === 'number') {
+            return stat;
+        }
+        if (!stat) {
```
Willyfrog, line 52:
```suggestion
        if (!stat || stat.length === 0) {
```

Willyfrog (changes requested):
adding a small extra, but looking good

Willyfrog (approved):

Willyfrog:
looking good! thanks @Ampit :)

I think it is good to test again @jgilliam17

components/analytics/system_analytics/system_analytics.tsx
```diff
@@ -43,8 +45,18 @@ export default class SystemAnalytics extends React.PureComponent {
         }
     }
 
-    render() {
-        const stats = this.props.stats;
+    private getStatValue(stat: number | AnalyticsRow[]):number | undefined {
```
Willyfrog, line 48:
```suggestion
    private getStatValue(stat: number | AnalyticsRow[]): number | undefined {
```

Willyfrog: this should fix the linting problem

jgilliam17 (approved):
Thank you @Ampit 
Tested, looks good to merge.
- Verified System Analytics module in system console - as expected.

PR #6998, opened 2020-11-02
MM-30095/MM-30195 Re-add missing elements to RHS

hmhealey:
Some recent RHS changes accidentally removed some stuff in the RHS that was actually important. There's still an outstanding issue with the search not working any more, but Nevy is looking into that.

One small thing to note, but there's a slight difference in the DOM from before the RHS refactoring since the `.sidebar-right-container` div now wraps the `.post-right__container` one. Before, I think we'd have only one of those at once. I don't think this causes any different behaviour, but I thought I'd note it since the CSS in the RHS is complicated.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30095
https://mattermost.atlassian.net/browse/MM-30195

### DIFF
--- a/components/sidebar_right/sidebar_right.jsx
+++ b/components/sidebar_right/sidebar_right.jsx
@@ -9,6 +9,7 @@ import {trackEvent} from 'actions/telemetry_actions.jsx';
 import Constants from 'utils/constants';
 import * as Utils from 'utils/utils.jsx';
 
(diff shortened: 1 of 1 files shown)


devinbinnie (approved):

nevyangelova (approved):

hmhealey:
Closing this since we're reverting the changes that caused this because of the aforementioned search issue.

PR #7012, opened 2020-11-03
[MM-30190] Show spinner when loading invoices instead of 'no invoices found'

nickmisasi:
#### Summary
The billing history screen would flash "No invoices found" (or something similar) until the invoices endpoint returned something (regardless of if it was going to return something or not. 

invoices starts out null, and then is defined when the endpoint returns, even if the endpoint returns nothing. So we just need to know whether its null or not, before handing over the render to the rest of the logic.

I also added a spinner because I thought it looked good. Spoke with Matt, and he agrees.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30190

#### Related Pull Requests
n/a

### DIFF
--- a/components/admin_console/billing/billing_history.scss
+++ b/components/admin_console/billing/billing_history.scss
@@ -99,6 +99,11 @@
     text-align: right;
 }
 
+.BillingHistory__spinner {
+    margin: auto;
+    text-align: center;
+}
+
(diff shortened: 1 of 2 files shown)


devinbinnie (approved):
👌

marianunez (approved):

PR #7020, opened 2020-11-04
[MM-30183] - No more text decoration on hover and focus 

hahmadia:
#### Summary

No more text decoration on hover and focus on mobile webb app view

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30183

#### Screenshots

image


amyblais:
@hahmadia I saw discussion on this in the Developers:Webapp channel, but can you help confirm that this is not a recent regression?

hahmadia:
Yo @amyblais 🌞 !

So I checked the blame of the files and the CSS class `post_permalink` has been applied forever and it hasn't changed since 3 years which leads me to the fact that this is **not** a recent regression and has been existing forever. Somehow, it was discovered only recently through regression testing. 

QA can double check me on this but I'm confident this is not a regression and has been there for a long time. Just not something anyone has probably noticed.

devinbinnie (approved):

calebroseland (approved):

furqanmlk (approved):

PR #7024, opened 2020-11-04
MM-30090 Open managed resource links in a new tab 

hmhealey:
Due to some changes in 5.28 around the internal/external link handling in the web app, we broke the newly added [Desktop Managed Resources](https://docs.mattermost.com/install/desktop-managed-resources.html) feature which basically lets non-MM services host stuff at certain routes under the MM server.

Since most paths in the server could be a team URL, the new link handling is correct for anyone not using that feature, so instead of reverting the link handling changes, we're adding a way to the server to manually specify these paths that should not be handled by the web app.

I rewrote the handling of managed vs unmanaged links (which is a better term than internal vs external since plugins and public file links would technically be external), and I've tried to add test cases to cover all of these recent changes we've made that got us in this situation to begin with. In particular, I'm trying to also avoid regressions for these tickets
- https://mattermost.atlassian.net/browse/MM-29049
- https://mattermost.atlassian.net/browse/MM-28790
- https://mattermost.atlassian.net/browse/MM-28042
- https://mattermost.atlassian.net/browse/MM-27559

Outside of the managed path changes, there are a couple small changes to treat more relative links as internal like we previously did for all absolute links, but none that were covered by previous test cases.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30090

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/16213
https://github.com/mattermost/mattermost-redux/pull/1282

### DIFF
--- a/utils/markdown/index.test.js
+++ b/utils/markdown/index.test.js
@@ -169,4 +169,48 @@ this is long text this is long text this is long text this is long text this is 
         const plugin = format('[plugin](/reiciendis-0/plugins/example))', {siteURL: 'http://localhost'});
         expect(plugin).toContain('<a class="theme markdown__link" href="/reiciendis-0/plugins/example" rel="noreferrer" data-link="/reiciendis-0/plugins/example">plugin</a>');
     });
+
+    describe('should correctly open links in the current tab based on whether they are handled by the web app', () => {
+        for (const testCase of [
+            {name: 'regular link', link: 'https://example.com', handled: false},
+            {name: 'www link', link: 'www.example.com', handled: false},
+
+            {name: 'link to a channel', link: 'http://localhost/team/channels/foo', handled: true},
+            {name: 'link to a DM', link: 'http://localhost/team/messages/@bar', handled: true},
+            {name: 'link to the system console', link: 'http://localhost/admin_console', handled: true},
+            {name: 'permalink', link: 'http://localhost/reiciendis-0/pl/b3hrs3brjjn7fk4kge3xmeuffc', handled: true},
+            {name: 'link to a specific system console page', link: 'http://localhost/admin_console/plugins/plugin_com.github.matterpoll.matterpoll', handled: true},
+
+            {name: 'relative link', link: '/', handled: true},
+            {name: 'relative link to a channel', link: '/reiciendis-0/channels/b3hrs3brjjn7fk4kge3xmeuffc', handled: true},
+            {name: 'relative link to a DM', link: '/reiciendis-0/messages/b3hrs3brjjn7fk4kge3xmeuffc', handled: true},
+            {name: 'relative permalink', link: '/reiciendis-0/pl/b3hrs3brjjn7fk4kge3xmeuffc', handled: true},
+            {name: 'relative link to the system console', link: '/admin_console/', handled: true},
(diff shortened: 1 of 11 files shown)


utils/markdown/index.test.js
```diff
@@ -169,4 +169,48 @@ this is long text this is long text this is long text this is long text this is
...
+
+            {name: 'link to a plugin-handled path', link: 'http://localhost/plugins/example', handled: false},
+            {name: 'link to a file attachment public link', link: 'http://localhost/files/o6eujqkmjfd138ykpzmsmc131y/public?h=j5nPX8JlgUeNVMOB3dLXwyG_jlxlSw4nSgZmegXfpHw', handled: false},
+
+            {name: 'relative link to a plugin-handled path', link: '/plugins/example', handled: false},
+            {name: 'relative link to a file attachment public link', link: '/files/o6eujqkmjfd138ykpzmsmc131y/public?h=j5nPX8JlgUeNVMOB3dLXwyG_jlxlSw4nSgZmegXfpHw', handled: false},
+
+            {name: 'link to a managed resource', link: 'http://localhost/trusted/jitsi', options: {managedResourcePaths: ['trusted']}, handled: false},
+            {name: 'relative link to a managed resource', link: '/trusted/jitsi', options: {managedResourcePaths: ['trusted']}, handled: false},
+            {name: 'link that is not to a managed resource', link: 'http://localhost/trusted/jitsi', options: {managedResourcePaths: ['jitsi']}, handled: true},
```
hmhealey, line 199: @dpanic To make sure we're on the same page with how this should work, do these cases look correct to you? If you specify `trusted` as a managed resource path, `http://mymattermostserver.com/trusted/jitsi` and `/trusted/jitsi` should both be opened in a new window, but `http://mymattermostserver.com/somethingelse/trusted` should not, correct?

dpanic: Yes that was expected behaviour.

jespino (approved):
LGTM

dpanic (approved):
LGTM

Willyfrog (approved):
LGTM
added the docs needed tag so we don't forget to update https://docs.mattermost.com/install/desktop-managed-resources.html with info about setting this parameter

hmhealey:
@jgilliam17 This will have to wait for the server changes to be merged for the setting to function. Those are approved, so I just need to wait for the build to pass for them

hmhealey:
@justinegeffen Adding you for review on the help text I added. It's pretty simple since it just refers back to the docs [here](https://docs.mattermost.com/install/desktop-managed-resources.html)
image

hmhealey:
@jgilliam17 This is now testable

justinegeffen:
Thanks @hmhealey! That looks good to me. :)

justinegeffen (approved):

amyblais:
@hmhealey Does this setting need to be added to diagnostics/telemetry?

hmhealey:
@amyblais Good catch. I submitted a PR for that

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified Managed Resource Paths section in sys console and 4 linked tickets for regression, as expected.

PR #7025, opened 2020-11-05
[MM_20423] 'external_image'module to ts (#16151)

dizkek:
#### Summary

Converted 'external_image' module to ts

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16151


components/external_image/external_image.tsx
```diff
@@ -1,20 +1,25 @@
...
-export default class ExternalImage extends React.PureComponent {
-    static propTypes = {
-        children: PropTypes.func.isRequired,
-        enableSVGs: PropTypes.bool.isRequired,
-        hasImageProxy: PropTypes.bool.isRequired,
-        imageMetadata: PropTypes.object,
-        src: PropTypes.string.isRequired,
-    };
+interface ImageMetaData {
+    format?: string;
```
jespino, line 9: I think this `ImageMetaData` it is actually a `PostImage` object type from `mattermost-redux/types/posts` (that doesn't have the `format` field defined. So I think we should add that field into `mattermost-redux` repo.

In any case, I think the fields are not optional. The image metadata can exists or not, but if it exists, it should have all the fields. And actually, there is another field that is not shown here that is the `frame_count` for gif images.

dizkek: I checked `mattermost-redux/types/posts` and I reckon I can add format property in `PostImage ` and can import from there. Thanks for that information.

I reckon I can change width and height fields are not optional since type error only happens in `external_image_test` file when I change those properties are not optional. Hence, I can change teseCase.

But I am so confused whether I have to add format field as optional or not because I checked `post_attatchment_opengraph`(which uses`PostImage` type) but I cannot find any code has format filed for this type.

hahmadia: It's not optional. I checked our backend code and we have this structure defined there as well `PostImage` and you can see it has the following fields defined
image
This shows that the we need to add a `format: string` and `frameCount: number` to the PostImage type in redux.

The redux object should be mapping to the object defined in the backend (or in go terms, the struct)

components/external_image/external_image.tsx
```diff
@@ -1,20 +1,25 @@
...
+    width?: number;
+    height?: number;
+}
+
+interface Props {
+    children: (src : string) => React.ReactNode;
+    enableSVGs: boolean;
+    hasImageProxy: boolean;
+    imageMetadata?: ImageMetaData,
```
hahmadia, line 18: Import `PostImage` from redux and use it here.
You can create a new PR which adds those above fields to the `PostImage` object in the redux repository.

hahmadia (commented):
Looks good! Just some changes that need to be made in addition to comments Jesus made. Thanks!

components/external_image/external_image.tsx
dizkek: I changed my code and pushed again. Actually, I changed types as you advised me in the last comment. I checked type-test locally and it worked okay. But the problem is I changed types in mattermost-redux which is belongs to node_modules. Hence, its changes has not been updated in this commit and type-checks errors are keep failing.

dizkek: I requested PR in redux-repository and waiting for the merge.

hahmadia: The build failure is because the redux hash needs to be update with latest commit in your redux PR. 
Once your redux PR gets merged, then we can update the redux hash with the latest in master redux before merging this PR into master.

hahmadia:
The type check failure you are seeing is not anything to do with your PR. The latest redux changes have not had their web app component updated so it's failing. I'll submit a separate PR for this and once it's merged into master, we can merge it into your PR and we should be good! Sorry about the inconvenience.

hahmadia:
Awesome! Everything is good now. Approving :)

hahmadia (approved):

package.json
```diff
@@ -37,7 +37,7 @@
     "localforage-observable": "2.0.1",
     "mark.js": "8.11.1",
     "marked": "github:mattermost/marked#87769262aa02e1784570f61f4f962050e07cc335",
-    "mattermost-redux": "github:mattermost/mattermost-redux#40368bb807b8b5a3dc0f4ce4a5a76eabc5c9de54",
+    "mattermost-redux": "github:mattermost/mattermost-redux#495176305523322d09706c0416d3acd42e56a716",
```
jespino, line 40: You can update this now to the master branch head in the mattermost-redux repo. Other than that, looks good to me. Thanks @dizkek!

jespino (changes requested):
I tiny change required (related to the process), once is done, we can merge it.

jespino:
The QA can be done now because the changes needed shouldn't affect the behavior.

saturninoabril (approved):
Thanks @dizkek for your contribution! LGTM, tested and passed.

package.json
dizkek: I changed but there is still some type errors....

hahmadia:
To help push this over the finish line, I helped sort out the merge problems. The redux did not need to be updated as it's already been updated in the latest master.

hahmadia:
@jespino Please review when you have time. Thanks

jespino (approved):
Looks good to me. Thanks @tbalthazar! :tada:

dizkek:
Thanks so much guys for my first PR!

PR #5828, opened 2020-07-01
[MM-24436]- Add a threshold from bottom for new messages toast

sridhar02:
#### Summary
Changed the buffer value to be considered from bottom for the toast message

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/14904
JIRA https://mattermost.atlassian.net/browse/MM-24436

PR #7028, opened 2020-11-06
MM-29929 Fix for unread messages toast on marking channel as unread

sudheerDev:
#### Summary
This was a regression caused here https://github.com/mattermost/mattermost-webapp/pull/5828.

   * The initOffset we use for displaying toast should be only used mount as that is not a scroll position that is updated after mount. It. can only be used for determining the offset when mounting like unread toast but not after. scrolling a bit and marking channel as unread. We are not tracking the real-time scroll offsets to do this atm.

   * Changing this to show unread toast on all occasions of marking the channel as
     unread other than channel at the bottom

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29929

### DIFF
--- a/components/toast_wrapper/toast_wrapper.jsx
+++ b/components/toast_wrapper/toast_wrapper.jsx
@@ -98,15 +98,17 @@ class ToastWrapper extends React.PureComponent {
                 showMessageHistoryToast = props.initScrollOffsetFromBottom > THRESHOLD_FROM_BOTTOM || !props.atLatestPost;
             }
 
-            // show unread toast when a channel is marked as unread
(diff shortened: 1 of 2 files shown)


Willyfrog (approved):

amyblais:
@devinbinnie Please help review today if you can for today's v5.29 code freeze.

deanwhillier (approved):

jgilliam17 (approved):
Thank you @sudheerDev 
Tested, looks good to merge.
- Verified unread toast is present when manually marking the channel as
unread in all cases other than when the user is at channel bottom.

PR #3252, opened 2019-07-25
MM-17226 Reduce unnecessary rendering in search results (and flagged/pinned) sidebar

jwilander:
#### Summary
This pull request will reduce search result sidebar rendering in two components:

##### Case 1: SearchResults (first commit)
If the search results sidebar is open and any changes to any posts occurs (new posts, edited posts, etc.) then the SearchResults component would re-render one or more times. The cause of this is that SearchResults was a PureComponent and the selector for the array of `results` relied on the posts in the Redux state. If that Redux state for posts changed at all, the memoization for the selector would miss and a new `results` array would be returned, causing a render.

The solution is not make SearchResults a PureComponent, perform our own shallow compare for all props/state except for `results`. For `results` we do a slightly deeper compare to check if any of the children have actually changed.

##### Case 2: SearchResultsItem (second commit)
If the search results sidebar is open, any item in the results would re-render if the channel that item belongs to received a new post. The cause of this was SearchResultsItem received the full channel object as a prop. Channel objects contain a count of the number of messages in the channel and this count updating was causing the unnecessary re-renders of SearchResultsItem.

The solution was to pass in only the parts of the channel object we care about as their own primitive type props.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-17226

PR #7029, opened 2020-11-06
MM-29111: search bar and search results as functional components

michelengelen:
fixed a bug with not rendering the search-results when the props changed (new results are pushed into the component)

It is a re-PR for #6796 (all other infos are found there! :) ✨ 

@nevyangelova


jgilliam17:
Thank you @michelengelen 
On the first pass, I noticed some missing results when searching for hashtags from the posted text, and also when using `-` and `""` Is this within scope? I'm not seeing the same issue on master.test

michelengelen:
hi @jgilliam17 ... i suppose the missing results when using `-` terms is similar to what happened the last time with my original PR. It was happening because the `Bleve` (i am not 100% certain on this) service is not started on the test-server. I c an only suggest that the other missing results have the same cause. Could you re-check with activated Bleve service?

nevyangelova:
@michelengelen when you try selecting the search hint from the RHS it doesn't focus the input and the suggestions are not displayed. Otherwise LGTM

components/search_results/search_results.tsx
```diff
@@ -0,0 +1,263 @@
...
+        isSearchingPinnedPost,
+        isSideBarExpanded,
+        isMentionSearch,
+        isOpened,
+        updateSearchTerms,
+    } = props;
+
+    const noResults = (!results || !Array.isArray(results) || results.length === 0);
+    /* eslint-disable */
+    console.log('####### results: ', results)
```
nevyangelova, line 110: we forgot to remove the logs :D

nevyangelova (changes requested):

jgilliam17:
I enabled Bleve and still seeing the issue. Master.test is not using Bleve or Elasticsearch and I am getting the results for -, "" and # search there.
Not sure what could be causing this issue on the PR server. @nevyangelova @hmhealey thoughts?

hmhealey:
I've never used Bleve, so I don't know if that's having an effect on this. The server determines which search results are returned to the client. It sounds like the server isn't returning something here, but that shouldn't be affected by the PR.
The client only displays the results and highlights the text.

@michelengen Are the changes in this PR compared to the last one just https://github.com/mattermost/mattermost-webapp/pull/7029/commits/2aa85bcfbb087d72b41913ac218fd31e2ec36e5c?

Do you know why that was making it not re-render? That code looks like it shouldn't be necessary at first glance, but I'm not familiar with the PR that added it (https://github.com/mattermost/mattermost-webapp/pull/3252).

@jwilander Do you happen to remember how big the effects of https://github.com/mattermost/mattermost-webapp/pull/3252 were? I'm wondering if they're big enough that we can easily tell if that's still needed or if it's safe to remove.

michelengelen:
@hmhealey I could investigate why the memo was preventing a render. i suppose the checks were not sufficient.

jwilander:
@hmhealey I recall it impacting performance quite a bit. That would be backed up by the fact that we cherry-picked it into a release branch. I don't remember the specifics though

michelengelen:
Ok @jwilander i will investigate tomorrow and provide a solution of possible. 🦈

michelengelen:
Hi @jwilander ... I am very sorry. I had not the time to pursue this lately, but did run some tests yesterday and found that we were using the React.memo function parameter (props-checking) in a wrong way. It is kinda hostorically grown from the `shouldComponentUpdate` method.

We were simply returning the wrong Boolean 😄 

`shouldComponentUpdate` targets component and the checking function passed to `React.memo` is best described as `arePropsEqual`. I will provide a solution for it today or tomorrow!

components/search_results/search_results.tsx
michelengelen: Done! ✨

nevyangelova (changes requested):
Still the same comment as before, when clicking on the search hints in RHS (minimise screen to view) the field doesn't focus and the suggestions are not displayed, otherwise LGTM!

michelengelen:
> Still the same comment as before, when clicking on the search hints in RHS (minimise screen to view) the field doesn't focus and the suggestions are not displayed, otherwise LGTM!

Shucks! I did miss that completely. I thought it is just the logs we forgot to remove! 

I will have a look tomorrow! 🙏🏻

jwilander:
> Hi @jwilander ... I am very sorry. I had not the time to pursue this lately, but did run some tests yesterday and found that we were using the React.memo function parameter (props-checking) in a wrong way. It is kinda hostorically grown from the `shouldComponentUpdate` method.
> 
> We were simply returning the wrong Boolean 😄
> 
> `shouldComponentUpdate` targets component and the checking function passed to `React.memo` is best described as `arePropsEqual`. I will provide a solution for it today or tomorrow!

Ah, nice catch. Thanks @michelengelen !

nevyangelova (approved):
Perfect! 👍

hmhealey (approved):
Looks good to me.

I'm not particularly surprised it was an issue with memoization. Back when we used to use `shouldComponentUpdate` everywhere prior to switching to `PureComponent`, we constantly ran into issues with it. It was even worse at the time since we were frequently mutating objects instead of treating them as immutable, and we were also passing around a lot more instead of connecting child components to redux. Hooks seem to handle that sort of thing more nicely though.

jgilliam17:
Thanks @michelengelen 
Looks good. 
Just one more focus issue similar to mobile view issue Navy mentioned.
After opening search in mobile view and selecting a hint e.g exclude - or quotes, focus is not placed in the search box. If you start typing right after selecting a hint, text is added to main channel text box.

jespino:
I have some concerns about this PR. This is migrating a class based component into a functional component, but is not really converting it to a "funtional component approach" it is just putting everything together in the same function. From my perspective the functional components approach is great because tent to create smaller, well defined and well scoped components, in this case we have a huge (330 LOC) function with 20 inline functions defined and 5 different state variables handled. For me looks like an anti-pattern. I would suggest to divide this component in multiple small functional components well defined and well scoped, and then migrate the component using this small building blocks.

michelengelen:
@jespino i know where you are getting at. I already discussed this with @nevyangelova and we came to the conclusion that the whole quickinput, providers (I know that these will be deprecated and replaced soonish) search and search results should be re-done since a lot of it are kinda anti-patterns and massively are over-engineered in some regards.

I will have a look at a rebuild over the Christmas break and see how much easier this can be build.

We decided to leave it as is for the time being since this is a highly requested feature.

WDYT?

jespino:
I'm perfectly ok on merging this and splitting it later, but I just wanted to raise my concerns about the approach. I don't think this is a proper functional component, and we should do this all the way around, split and migrate, not migrate and split.

michelengelen:
I do understand you're concerns and am in the same boat as you. I originally planned on simplifying the component massively. It did not turn out as I expected, but the learning from this is what I wrote above. 

So kindly stay tuned and wait for the rebuild organism. ✨

michelengelen:
Hi @jgilliam17 I updated the focussing behaviour and merged the upstream/master branch into mine, so that no conflicts arise! Could you have a look again?

michelengelen:
@jespino  FYI: I started on a new implementation of this in a side-branch. It will fully utilise FC structures and best practices. ✨

jespino:
@michelengelen Glad to read that, thanks!

jgilliam17 (approved):
Thank you @michelengelen 
Tested, looks good to merge.
- Verified search bar and results. Also thanks for fixing the focus issues raised in mobile web view.

jasonblais:
@michelengelen Thanks for your contribution!

PR #7031, opened 2020-11-06
[MM-30226][MM-39255] Allow a post to be sent when you do CMD + ENTER on Mac with 'Allow all messages' to be sent via CMD + ENTER

hahmadia:
#### Summary

This PR fixes when trying to edit a post and using `CMD + ENTER` on mac for all messages to send the message. Previously, it did nothing. Now, it sends the message. This matches the behaviour of when creating a post.

Also addresses the problem where if you are within a code block when editing a post and setting is set to send messages only for code blocks, if you hit enter it should move the text to a new line and not submit. If you type `CMD + ENTER` then it will edit the post.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30226
https://mattermost.atlassian.net/browse/MM-30255

### DIFF
--- a/components/edit_post_modal/edit_post_modal.jsx
+++ b/components/edit_post_modal/edit_post_modal.jsx
@@ -297,7 +297,7 @@ class EditPostModal extends React.PureComponent {
 
     handleKeyDown = (e) => {
         const {ctrlSend, codeBlockOnCtrlEnter} = this.props;
-        
+
         const ctrlOrMetaKeyPressed = e.ctrlKey || e.metaKey;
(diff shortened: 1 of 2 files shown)


crspeller (changes requested):
Can we add a test for this case?

hahmadia:
> Can we add a test for this case?

@crspeller 
Yes we can! As the great Rudy Tomajnovich once said "Don't ever underestimate the heart of a CHAMPION". As long as you believe it can be done, it will be done! 😄 

spongebobthumpsup

crspeller (approved):

nickmisasi (approved):
LGTM! Sorry for the delay.

furqanmlk (approved):
Working as expected.

PR #7037, opened 2020-11-08
[MM-8497] Ability to set Do Not Disturb for a specified period of time 

darkLord19:
Client side changes for mattermost/mattermost-server#15640

Server side changes at mattermost/mattermost-server#16067


jfrerich:
Thanks, @darkLord19! Let us know when this PR is ready for a full review and we'll add reviewers!

esethna:
Adding @jespino to help with the issue you described https://community-daily.mattermost.com/core/pl/5j3ty1zjspydibp7ccew86j38y

components/status_dropdown/status_dropdown.jsx
```diff
@@ -116,11 +144,54 @@ export default class StatusDropdown extends React.PureComponent {
...
+                val2: 1000,
+            },
+        ];
+
+        const dndSubMenuItems = times.map((tm) => {
+            return {
+                id: `dndTime-${tm.val}`,
+                direction: 'right',
+                text: tm.text,
+                action: this.setStatusDND(tm),
```
jespino, line 185: You are setting the action here to the result of calling `setStatusDND` instead of setting the action to a function that do that, probably you can solve this just using `() => this.setStatusDND(tm)`. Anyway you are passing the whole `tm` object to the `setStatusDND` when you are using only the `val2` attribute, probably makes sense to pass the `val2` attribute only instead. And by the way, val1 and val2 are awful variable names, probably you should go for sometihng else or even remove the `val` variable if is not strictly neded, you can use the position in the list as key, because this is not going to be modified in runtime.

darkLord19: Thanks for the help @jespino. I agree val1 and val2 are awful names but this tm object is not final. I was just trying to get the feature working. I will be changing the representation once I get the idea of implementation details. I pushed it here to get the help.

esethna:
@darkLord19 how's this going? Can we help at all?

darkLord19:
@jespino now it's almost ready but only one issue, on time picker menu item click, page is getting refreshed, not seeing any errors in console.

jespino:
@darkLord19 the problem can be solved stopping the propagation at the menu wrapper level. for that you can add to the `MenuWrapper` node the `stopPropagationOnToggle={true}` attribute.

esethna:
@darkLord19 let us know how else we can help. Apologies for the delay over the holidays!

darkLord19:
> @darkLord19 let us know how else we can help. Apologies for the delay over the holidays! If we can get your PRs merged by Jan 19th we can likely ship them in our Jan 26th cloud release and following February 16th self-hosted release.

Oh cool. thanks for the heads up. Only one minor issue is remaining so I will try to get it done before the weekend if possible so there is some time for proper review and testing.

darkLord19:
@jespino i have binded setStatus to props.actions in dnd_custom_time_picker_modal but i keep getting this error " Uncaught TypeError: this.props.actions is undefined". Not sure what is the issue here.

components/status_dropdown/status_dropdown.jsx
```diff
@@ -17,12 +17,16 @@ import StatusAwayIcon from 'components/widgets/icons/status_away_icon';
 import StatusOnlineIcon from 'components/widgets/icons/status_online_icon';
 import StatusDndIcon from 'components/widgets/icons/status_dnd_icon';
 import StatusOfflineIcon from 'components/widgets/icons/status_offline_icon';
+import DndCustomTimePicker from 'components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal';
```
jespino, line 20: Here is the problem, you are importing the component directly instead of the "connected component", you have to remove the last `dnd_custom_time_picker_modal` part.

jespino:
@darkLord19 found the problem with the setStatus, added a comment in the code :)

esethna:
@darkLord19 hope you're well! Anything else we can help with?

darkLord19:
> @darkLord19 hope you're well! Anything else we can help with?

Hey @esethna this is all done but blocked by the server tests failures now.

components/status_dropdown/index.jsx
```diff
@@ -21,11 +25,15 @@ function mapStateToProps(state) {
...
+    const etz = areTimezonesEnabledAndSupported(state);
     return {
         userId,
         profilePicture: Client4.getProfilePictureUrl(userId, currentUser.last_picture_update),
         autoResetPref: get(state, Preferences.CATEGORY_AUTO_RESET_MANUAL_STATUS, userId, ''),
         status: getStatusForUserId(state, userId),
+        userTimezone: ctz,
+        enableTimezone: etz,
     };
 }
```
michelengelen, line 38: there is no need to instantiate variables for the values.

They can immediately saved to the corresponding properties:

```suggestion
    return {
        userId: currentUser.id,
        profilePicture: Client4.getProfilePictureUrl(userId, currentUser.last_picture_update),
        autoResetPref: get(state, Preferences.CATEGORY_AUTO_RESET_MANUAL_STATUS, userId, ''),
        status: getStatusForUserId(state, userId),
        userTimezone: getUserTimezone(state, userId),
        enableTimezone: areTimezonesEnabledAndSupported(state),
    };
}
```

components/status_dropdown/index.jsx
```diff
@@ -9,8 +9,12 @@ import {getCurrentUser, getStatusForUserId} from 'mattermost-redux/selectors/ent
 import {Preferences} from 'mattermost-redux/constants';
 import {get} from 'mattermost-redux/selectors/entities/preferences';
 
+import {getUserTimezone} from 'mattermost-redux/selectors/entities/timezone';
+
 import {openModal} from 'actions/views/modals';
 
+import {areTimezonesEnabledAndSupported} from 'selectors/general';
+
 import StatusDropdown from 'components/status_dropdown/status_dropdown.jsx';
```
michelengelen, line 18: please group imports logically/contextually (all mattermost-redux imports in one block and so on)

components/status_dropdown/status_dropdown.jsx
```diff
@@ -17,12 +17,16 @@ import StatusAwayIcon from 'components/widgets/icons/status_away_icon';
...
+import DndCustomTimePicker from 'components/dnd_custom_time_picker_modal';
+import {getCurrentDateTimeForTimezone} from 'utils/timezone';
 
 export default class StatusDropdown extends React.PureComponent {
     static propTypes = {
         style: PropTypes.object,
         status: PropTypes.string,
         userId: PropTypes.string.isRequired,
+        userTimezone: PropTypes.object,
+        enableTimezone: PropTypes.bool,
```
michelengelen, line 29: I am not entirely sure if this wording is the best we can use for that. Since it is a boolean it should probably be something like this: `isTimezoneEnabled`

WDYT @jespino ?

components/status_dropdown/status_dropdown.jsx
```diff
@@ -37,35 +41,101 @@ export default class StatusDropdown extends React.PureComponent {
...
+            if (tz.useAutomaticTimezone) {
+                currentDate = getCurrentDateTimeForTimezone(tz.automaticTimezone);
+            } else {
+                currentDate = getCurrentDateTimeForTimezone(tz.manualTimezone);
+            }
+        }
+        return currentDate;
+    }
+
+    dndTimes = ['30 mins', '1 hour', '2 hours', 'Today', 'Tomorrow', 'Custom']
```
michelengelen, line 65: We could probably move this to a constant file. I am 1/5 this would be a good addition in `UserStatuses`

components/status_dropdown/status_dropdown.jsx
```diff
@@ -37,35 +41,101 @@ export default class StatusDropdown extends React.PureComponent {
...
+            break;
+        case 3:
+            // set hours of date to point to last moment of the day
+            endTime.setHours(23, 59, 59, 999);
+            break;
+        case 4:
+            // set hours of date to point to last moment of the day
+            // and add 24 hours to it to point to last moment of tomorrow
+            currentTime.setHours(23, 59, 59, 999);
+            endTime = new Date(currentTime + (48 * 30 * 60000));
```
michelengelen, line 118: using `48 * 30 * 60000` might be misleading here ... Please use `24 * 60 * 60 * 1000` (h * m * s * ms).
It is the more approach most devs use and know.

components/status_dropdown/status_dropdown.jsx
```diff
@@ -115,6 +193,24 @@ export default class StatusDropdown extends React.PureComponent {
...
+                    action: () => setCustomTimedDnd(),
+                };
+            }
+            return {
+                id: `dndTime-${time}`,
+                direction: 'right',
+                text: time,
+                action: () => setDnd(event, index),
+            };
+        });
```
michelengelen, line 213: this can be simplified with using a ternary in `action` property:

```suggestion
        const dndSubMenuItems = this.dndTimes.map((time, index) => ({
                id: `dndTime-${time}`,
                direction: 'right',
                text: time,
                action: index === 5 ? () => setCustomTimedDnd() : () => setDnd(event, index),
            });
        });
```

components/widgets/menu/menu.scss
```diff
@@ -14,7 +14,7 @@
 
         ul {
             margin: 0;
-            padding: 0;
+            padding: 8px 0;
```
michelengelen, line 17: This should be handled by component SCSS, since changing the base-component has side-effects for other implementations

components/widgets/menu/menu_header.scss
```diff
@@ -6,15 +6,15 @@
         clear: both;
         color: inherit;
         cursor: default;
-        display: block;
+        display: flex;
```
michelengelen, line 9: This should be handled by component SCSS, as well, since changing the base-component will certainly have side-effects

components/widgets/menu/menu_header.scss
```diff
@@ -6,15 +6,15 @@
...
         font-weight: normal;
         padding: 0 32px 0 24px;
         text-align: left;
         white-space: nowrap;
         width: 100%;
         overflow: hidden;
         text-overflow: ellipsis;
-        line-height: 28px;
         font-weight: $font-weight--semibold;
+        height: 36px;
```
michelengelen, line 18: same here

components/widgets/menu/menu_items/menu_item.scss
```diff
@@ -20,8 +20,8 @@
         }
 
         span.icon {
-            width: 28px; // less 12px padding is 16px
-            height: 16px; // less 3px top padding is 16px;
+            width: 26px; // less 12px padding is 16px
+            height: 14px; // less 3px top padding is 16px;
```
michelengelen, line 24: please dont change the base-component SCSS ... alter values in the implementing component

sass/components/_day-picker.scss
```diff
@@ -19,7 +19,16 @@
         }
 
         .DayPicker-Day {
-            border-radius: $border-rad * 2;
+            min-width: 32px;
```
michelengelen, line 22: same here ... IMO changin the base component is pretty unsafe, sice it might have a big impact on other components

sass/components/_dnd-pickers-dropdown-menu.scss
```diff
@@ -0,0 +1,23 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+
+    .btn {
+        font-size: 14px;
+        font-weight: 600;
+        padding: 0 2rem;
+        height: 4rem;
+        border-radius: 4px;
```
michelengelen, line 21: please use `$border-rad` variable (with int multipliers if needed)

components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.scss
```diff
@@ -0,0 +1,118 @@
...
+    .modal-footer {
+        display: none;
+    }
+
+    .btn-primary {
+        font-size: 14px;
+        font-weight: 600;
+        padding: 0 2rem;
+        height: 4rem;
+        border-radius: 4px;
```
michelengelen, line 13: please use `$border-rad` variable (with int multipliers if needed)

michelengelen (changes requested):
There are a lot of changes in base-component SCSS which should either be moved to the component SCSS as modifiers. if not possible the impact on other components should be monitored closely!

Any additional thoughts on this @jespino ?

components/generic_modal.scss
```diff
@@ -1,10 +1,18 @@
 @charset 'UTF-8';
 
 .GenericModal {
+    width: 692px;
+
```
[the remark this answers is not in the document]
jespino: Yes, I prefer to have an extra class for the specific modal.

components/status_dropdown/status_dropdown.jsx
jespino: Yes, I agree with that, `enableTimezone` looks like an action. I think isTimezoneEnable is clearer.

components/status_dropdown/status_dropdown.jsx
```diff
@@ -115,6 +193,24 @@ export default class StatusDropdown extends React.PureComponent {
         const setDnd = needsConfirm ? () => this.showStatusChangeConfirmation('dnd') : this.setDnd;
         const setAway = needsConfirm ? () => this.showStatusChangeConfirmation('away') : this.setAway;
         const setOffline = needsConfirm ? () => this.showStatusChangeConfirmation('offline') : this.setOffline;
+        const setCustomTimedDnd = needsConfirm ? () => this.showStatusChangeConfirmation('dnd') : this.setCustomTimedDnd;
+
+        const dndSubMenuItems = this.dndTimes.map((time, index) => {
+            if (index === 5) {
+                return {
+                    id: `dndTime-${time}`,
```
jespino, line 201: time has spaces and things like that. I would suggest to convert `time` to some kind of ids, and then use that id without spaces here. And also for the `text` that should be translated.

components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+    setTimeMenuList = () => {
+        const timeMenuItems = [];
+        let h = 0;
+        let m = 0;
+        const curr = this.props.getCurrentDateTime(this.props.userTimezone, this.props.enableTimezone);
+        if (curr.toDateString() === this.state.dateString) {
+            h = curr.getHours();
+            m = curr.getMinutes();
+            if (m > 20) {
```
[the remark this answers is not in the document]
darkLord19: No. So, the logic here is if it is already > 12.20 then in dropdown don't show item for 12.30 but directly show for 1.00.

components/generic_modal.scss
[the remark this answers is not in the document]
darkLord19: ++ @asaadmahmood

components/widgets/menu/menu.scss
darkLord19: @asaadmahmood

components/widgets/menu/menu_header.scss
darkLord19: ++ @asaadmahmood

components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.tsx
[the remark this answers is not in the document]
michelengelen: Ok, then that's fine! Thanks for clarifying! 🙇🏼‍♂️

components/status_dropdown/status_dropdown.jsx
darkLord19: how about splitting time at space and only using first element and also adding index of array?

darkLord19:
@asaadmahmood hey! can you go through review comments about scss things and see if you can help out! I'd probably mess up even more so pinging you!

asaadmahmood:
@darkLord19 I might be able to look at this next week. A bit tied with some other stuff right now.

darkLord19:
@asaadmahmood no worries! I will try to address them in the meanwhile and you can help out with the remaining things if any when you are free.

asaadmahmood:
Sounds good.

components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.tsx
```diff
@@ -162,7 +162,9 @@ export default class DndCustomTimePicker extends React.PureComponent<Props, Stat
                 <div className='DndModal__content'>
                     <div>
                         <div className='DndModal__input DndModal__input--no-border'>
-                            <div className='DndModal__input__label'>{'Date'}</div>
+                            <div className='DndModal__input__label'>
+                                {localizeMessage('dnd_custom_time_picker_modal.date','Date')}
+                            </div>
```
michelengelen, line 167: Could you please use FormattedMessage here as well, so we stay consistent? We could also remove `localizeMessage` from imports, since this is the last time it gets used.

components/status_dropdown/index.jsx
```diff
@@ -25,15 +27,14 @@ function mapStateToProps(state) {
     }
 
     const userId = currentUser.id;
-    const ctz = getUserTimezone(state, userId);
-    const etz = areTimezonesEnabledAndSupported(state);
+
     return {
         userId,
```
michelengelen, line 32: userId can be simplified as well.

```
userId: currentUser.id
```

components/status_dropdown/status_dropdown.jsx
```diff
@@ -196,19 +196,11 @@ export default class StatusDropdown extends React.PureComponent {
...
-                };
-            }
             return {
-                id: `dndTime-${time}`,
+                id: `dndTime-${index}-${time.split(' ')[0]}`,
                 direction: 'right',
-                text: time,
-                action: () => setDnd(event, index),
+                text: localizeMessage('status_dropdown.dnd_sub_menu_item.time',time),
+                action: () => index === 5 ? () => setCustomTimedDnd() : () => setDnd(event, index),
```
michelengelen, line 203: that way a function will be returned, that does again return a function.
just remove the `() => ` from the start:

```
action: index === 5 ? () => setCustomTimedDnd() : () => setDnd(event, index),
```

michelengelen (changes requested):

sass/components/_day-picker.scss
asaadmahmood: @michelengelen The base component for the day picker has changed, so we want this change reflected elsewhere too, which is why we have changes directly to the main component.
You can see the component here:
https://www.figma.com/file/0XY1PaiNrMXP2Z2E4gcsz7/Components---Date-Picker?node-id=86%3A689

asaadmahmood:
@darkLord19 I have fixed some css issues highlighted above. The app doesn't work for me locally, probably because of an old codebase. Do a rebase and get the app in the running state, I can then look at the other css issues.

darkLord19:
@asaadmahmood it should be working now

components/status_dropdown/index.jsx
darkLord19: @michaelgamble something strange is happening but when i do userId: currentUser.id, i am getting error
```
scheduler.development.js:178 Uncaught ReferenceError: userId is not defined
    at Function.mapStateToProps [as mapToProps] (index.jsx:31)
    at mapToPropsProxy (wrapMapToProps.js:41)
    at Function.detectFactoryAndVerify (wrapMapToProps.js:50)
    at mapToPropsProxy (wrapMapToProps.js:41)
    at handleFirstCall (selectorFactory.js:22)
    at pureFinalPropsSelector (selectorFactory.js:63)
    at connectAdvanced.js:319
    at mountMemo (react-dom.development.js:15442)
    at Object.useMemo (react-dom.development.js:15738)
    at useMemo (react.development.js:1521)
```

components/widgets/menu/menu_header.scss
asaadmahmood: @michelengelen These are global changes that need to be represented for all menu headers.
We ideally don't want different menu header sizes in the app. So we should be consistent.

You can see the global figma component here:
https://www.figma.com/file/tbMvXozL5f8Uc8oWYTEzbh/Patterns-Popover-Menus?node-id=532%3A3586

asaadmahmood:
Pushed my changes @darkLord19.
Tried to keep some of the changes in separate component files, but the menu items are actually broken to some extent, the float right on the submenu icon for example would not work well because the spacing from the top varies and its not vertically centered programatically. 

Our menu items are also different in height if they are a submenu compared to if they are not, which should not be the case.
I ideally want to fix them as a whole, so we can open up a separate ticket for that. I'll talk amongst the design team about that.

hmhealey:
Is this still a work in progress? I can add the review labels if not

esethna:
@hmhealey it's waiting server PR approval afaik

michelengelen (approved):
LGTM

darkLord19:
@asaadmahmood there is some gap between menus and if we hover mouse over the gap then the time selection menu is hidden. Can you please have a look at it?
Screenshot 2021-04-16 at 9 27 21 PM

@jespino after clicking on disable notifications button in dnd custom time modal, modal still remains open. Can you point me in the direction on how to fix it?

asaadmahmood:
Sure @darkLord19 I'll have a look at it.

darkLord19:
@esethna it seems done to me except small css issue which Asad is looking into so can you remove Work in Progress label and move this further in dev and test review

michelengelen (approved):
LGTM

esethna:
@jespino please help review when possible

asaadmahmood:
Trying to run the branches, but getting an error locally, will touch base wiht @darkLord19

asaadmahmood:
Fix pushed @darkLord19

michelengelen (approved):

components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.tsx
```diff
@@ -0,0 +1,219 @@
...
+            last_activity_at: toUTCUnix(this.state.currentDate),
+        });
+        this.props.onHide();
+    }
+
+    handleDaySelection = (day: Date) => {
+        const dString = this.formatDate(day);
+        this.setState({
+            selectedDate: dString,
+        }, this.setTimeMenuList);
```
hmhealey, line 96: Why call `setTimeMenuList` after setting the state? That'll cause the component to render twice when the date changes. Similarly, it'll also render twice on mount where `setTimeMenuList` is also called.

Could we maybe change `setTimeMenuList` so that instead of calling `setState` itself, we instead rename it something like `makeTimeMenuList` and have it return an object? Then we could use it here like
```javascript
this.setState({
    selectedDate: dString,
    ...this.makeTimeMenuList(dString),
});
```

Similarly, instead of calling it in `componentDidMount` then, we could just use it in the constructor.

components/generic_modal.scss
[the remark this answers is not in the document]
hmhealey: @asaadmahmood Were we going to change this? It'd be best if we aren't changing unrelated components in this PR.

hmhealey (changes requested):
A few questions about the one component, but everything else looks good.

components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.tsx
darkLord19: Makes sense, thanks. I have made the changes.

asaadmahmood:
@hmhealey Yup, that makes sense, we can remove that @darkLord19

asaadmahmood:
Done @hmhealey

hmhealey (approved):
Thanks for making those changes. Looks good

hmhealey:
@jespino Were you still wanting to give this a review or are we good to go to QA review?

esethna:
Adding @jgilliam17 for QA review (note that test server wont work given the server PR dependency^)

jgilliam17:
Thanks @darkLord19 Can you pull both PRs into a test server for me?

darkLord19:
@jgilliam17 I'm not sure how. I don't have access to any server.

jgilliam17:
No worries @darkLord19, my mistake.  🙂 
I've asked @hmhealey for help and will try on the test server he created here.

hmhealey:
This one didn't pick up the correct version of the server unfortunately. I'll try opening one on the server PR to see if it can pick up the web app correctly.

jgilliam17:
Thanks @hmhealey for giving server creation here a try.
@saturninoabril and @DHaussermann are going to set up a test server for me tomorrow. 
@darkLord19 I'll post test results here, hopefully by Monday eod.

jespino (approved):
Looks good to me.

jgilliam17:
Thank you for you contribution  @darkLord19 
On the first pass I am seeing few issues:
1. Submenu opens separated from the main menu, mostly inaccessible. Workaround used for testing > tab+enter and/or moving the mouse quickly. See video.

https://user-images.githubusercontent.com/52937121/116700044-51596880-a994-11eb-86ce-0450df609e27.mov

2. Submenu is missing a title bar as suggested in ticket functional notes
Screen Shot 2021-04-30 at 11 10 16 AM

3. Change custom modal help text to "Disable notifications until". I confirmed this text change request with technical writer  @cwarnermm Thanks Carrie!

4. Custom setting, initial default today date text color is off, shows inactive color until user opens the calendar and manually selects today's date again
Screen Shot 2021-04-30 at 9 43 05 AM

5. Current date circle identifier is shown on the next month's calendar
Screen Shot 2021-04-30 at 9 42 25 AM

6. Close button X incorrectly styled in light themes ( as expected in dark)

https://user-images.githubusercontent.com/52937121/116704850-df841d80-a999-11eb-90bc-12a2374ecf80.mov

7. Date and Time input fields incorrectly styled in dark themes, including the calendar dropdown. Time dropdown respects the theme.
Screen Shot 2021-04-30 at 9 20 05 AM
Screen Shot 2021-04-30 at 9 21 03 AM

8. Time is reset to 00:00 when changing date, not sure if this is expected. @matthewbirtch can you please advise


https://user-images.githubusercontent.com/52937121/116710353-aa7ac980-a99f-11eb-8de4-976901c87b9c.mov

9. Time input field is missing active border state

https://user-images.githubusercontent.com/52937121/116711448-dba7c980-a9a0-11eb-91ef-6fafbf99ac7f.mov

jgilliam17:
One more thing -  @darkLord19 can you please clarify `today` and `tomorrow` submenu options. Is DND expiring at midnight today and also midnight the next day respectively or at some other preset time e.g. 5pm or similar? Thanks

matthewbirtch:
> Time is reset to 00:00 when changing date, not sure if this is expected. @matthewbirtch can you please advise

@jgilliam17 I agree, it shouldn't reset when changing the date

esethna:
I'm adding @matthewbirtch to do a UX review here as well

darkLord19:
Fixed 1, 2, 3, 8 and 10.
About 6, this issue is impacting generic modal component for which I have raised issue here https://github.com/mattermost/mattermost-server/issues/17467

@asaadmahmood can you help with the 7 and 9?

@jgilliam17 can you elaborate on 5th point? I am not seeing it in my machine and didn't quite understood what you meant above.

darkLord19:
> One more thing - @darkLord19 can you please clarify `today` and `tomorrow` submenu options. Is DND expiring at midnight today and also midnight the next day respectively or at some other preset time e.g. 5pm or similar? Thanks

It's expiring at today midnight and next day midnight respectively

jgilliam17:
Here are steps for 5.  Let me know if this works, otherwise I can record a short video on Monday. 
Steps
1) Change system date to 4/30/2021
2)Open custom dnd modal
3) Open Date selector 
4) Observe that current date 4/30/2021 is highlighted 
5) Switch to next month
6) Observe that one blank spot is highlighted- where April 30th would be shown on the monthly May calendar, top week, right before May 1st

asaadmahmood:
@darkLord19 For some reason I can see the comment that lists down the issues so can't see 7 and 9.

darkLord19:
@asaadmahmood 
- Date and Time input fields incorrectly styled in dark themes, including the calendar dropdown. Time dropdown respects the theme.
- Time input field is missing active border state.

asaadmahmood:
@darkLord19 Thanks, will have a look at it.

matthewbirtch (changes requested):
Looks like @jgilliam17 caught a lot of the issues. @darkLord19 in addition to what's been identified, I found a few other issues. 

1. There is some weirdness happening with the data picker. When I choose a date it seems to visually select the day before my selection for some reason.



2. Once a DND time is set, there is no indication to the user when their current DND timer is set to expire. Can we add something in the status menu that displays this? I'm thinking something like this:
image

3. For the format of the times displayed in the time picker, I think we should make it follow the user's preference for 12-hour or 24-hour clock. Currently, it always seems to display times in 24-hour clock format.

4. Also, for the time picker, if I choose today's date, we should only show the times left in the day. For example, if the time is 10:00 AM and I choose today's date, all the times from 00:00 - 10:00 should be omitted from the dropdown.

5. I'm not 100% sure if it is caused by this PR, but I'm seeing that the hover state background on the custom status menu item isn't spanning the full width
image

darkLord19:
>     4. Also, for the time picker, if I choose today's date, we should only show the times left in the day. For example, if the time is 10:00 AM and I choose today's date, all the times from 00:00 - 10:00 should be omitted from the dropdown.
@matthewbirtch  It was like that only but @jgilliam17 brought up a point that by doing this user has to select date first if they want to select some other time so, I populated menu with all times. I am find either way so which behavior do we want to keep finally?

matthewbirtch:
One other note:

I find it a little confusing that the menu says 'Disable Notifications Until" and "Today" is a menu option. What does this mean? What happens when I choose "Today" from this menu? "Until today" seems confusing.

image

matthewbirtch:
> 1. There is some weirdness happening with the data picker. When I choose a date it seems to visually select the day before my selection for some reason.
> 
> > That's strange. I am not seeing this issue on my local machine or test server created. Did you see this issue with latest changes? Can you reproduce it on this link https://mattermost-webapp-pr-7037.test.mattermost.cloud/pr7037/channels/town-square?

Yes, I just reproduced it with the link provided here

> 1. For the format of the times displayed in the time picker, I think we should make it follow the user's preference for 12-hour or 24-hour clock. Currently, it always seems to display times in 24-hour clock format.
> 
> > I will keep it in backlog currently. Let other functional issues get resolved first and then if time permits then can do this before 5.36. This could be taken up as a separate small PR also after this is merged.

Ok, that's fine. As you know, we have a follow up ticket underway for some UX enhancements, so we can roll it in to that if it can't make it in here.

> 1. Also, for the time picker, if I choose today's date, we should only show the times left in the day. For example, if the time is 10:00 AM and I choose today's date, all the times from 00:00 - 10:00 should be omitted from the dropdown.
> 
> > @matthewbirtch  It was like that only but @jgilliam17 brought up a point that by doing this user has to select date first if they want to select some other time.
> 
> > Time selector should loop, show all available times. Currently only shows times that come after current system time through midnight when setting custom DND Date and Time on first modal open. Related to 8. as it will reset and let user have all time options if you first change the time then date.
> > -- @jgilliam17
> 
> > So, I populated menu with all times. I am fine either way so which behavior do we want to keep finally?

Ya, I was a little confused by the comment, but I see now that these are conflicting directions. Sorry about that. Ultimately, I think it's reasonable to expect users to select the day first and we have a default set for 'today'. I think it causes more confusion if I'm able to select a time for today that has already passed. Can we move forward with that? This will help prevent errors for the user as well. Currently, I can select a time for today that is in the past and there is no feedback that this time is an error - just a button that doesn't work. So, if we remove the times in the past, we prevent the modal from being in an error state at all.

jgilliam17:
Thanks @darkLord19 
Apologies if my comment on 10.  above wasn't clear. What I was seeing (this should be fixed now, based on your convo with @matthewbirtch ) - after opening custom modal and selecting date that is different from today, time options were limited to *times left today*,  instead of all 24h as the date in the future was selected.

Question about DND warning when user starts a new session or reloads browser in current session. 
What is expected when selecting `keep DND`?
Screen Shot 2021-05-03 at 1 44 17 PM
a) Keep current, user custom set time for DND and expire accordingly 

b) Display DND until user changes the setting manually, overrides custom user setting (currently on the test server)

Few other issues 
1. Time dropdown is sometimes missing for Today. Noticed with my newly added user, few plugin 401 errors and one if you click on the button to set with missing time 
Screen Shot 2021-05-03 at 4 21 08 PM

2.  Help text wording is not updated on the modal, still shows `till`
3. Unable to custom set any time in the next 24 h

https://user-images.githubusercontent.com/52937121/116933673-de6d1d80-ac31-11eb-855b-ef454bb46d33.mov

darkLord19:
@jgilliam17 @matthewbirtch so we show time in multiple of 30 minutes so what to do if such slot is not available for today? and if current time is less than Xhours:20 minutes then we don't show Xhours:30 minutes option in dropdown and directly show for X+1 hours:00. Any suggestions to improve this?

asaadmahmood:
@darkLord19 I've looked at the styling on dark themes, and increased the width based on the 5th point.
Regarding the time input active state, its a button, so we may need to add a class on it when the menu is open and remove it when the menu is closed to get the desired effect.

Probably do it with onToggle in the menuWrapper.

darkLord19:
@asaadmahmood  thanks, i will have a look at it.
@jgilliam17 there seems to be some issue with test server created. I have updated dnd modal header text but on server it is still showing old text and also @matthewbirtch was facing some date picker related issue which is not seen in local environment a I've already fixed it but Matthew was able to reproduce on server. Now that most of the issues pointed out are fixed, can we like get a fresh webapp running from the tip of this branch and then list issues still happening.

So, current remaining issues known to me are,
1. time dropdown active state is missing
2. day outside of current month is highlighted in day picker
3. empty time dropdown after current time is > 23:20(need @matthewbirtch 's suggestion on what should be done from ux perspective)

Please add any issues I might have missed and also give another review to current latest webapp changes and list any other issues found so I will work on them and ping you once all are taken care of.

Few UX enhancements also(less priority),
1. have 12hr or 24hr format in time dropdown based on user's preference
2. if dnd is set by user then show expiration tim under DND item in status dropdown

asaadmahmood:
@darkLord19 I will create another PR for the menu wrapper thing to have a class on the wrapper, which we can then reuse here, so ignore that change for now.

darkLord19:
>Question about DND warning when user starts a new session or reloads browser in current session.
What is expected when selecting keep DND?

To answer your question, if you select "No, Keep status as DND" then nothing will happen as dnd status is already set. I mean to say that there won't be any call to server to change/upate status. Webapp will call server only if you select "Yes set my status to online" to update status to online

darkLord19:
@jgilliam17 now only one issue is remaining in my knowledge that is,  "empty time dropdown after current time is > 23:20"

- time dropdown active state is missing >> @asaadmahmood will be handling it via separate pr
- day outside of current month is highlighted in day picker >> fixed

matthewbirtch:
> @jgilliam17 @matthewbirtch so we show time in multiple of 30 minutes so what to do if such slot is not available for today? and if current time is less than Xhours:20 minutes then we don't show Xhours:30 minutes option in dropdown and directly show for X+1 hours:00. Any suggestions to improve this?

@darkLord19 I'm not sure if I follow the question, but here's how I'm interpreting. The time dropdown is displaying empty if the current day is selected in the date input and there is < 30 minutes left in the current day. Is that the problem?

If my interpretation above is correct, we could the following:
- if there are <30 minutes left in the day, set the default date when the modal opens to tomorrow's date then all times will show for the next day
- don't allow users to choose the current day

Thoughts?

matthewbirtch:
> I find it a little confusing that the menu says 'Disable Notifications Until" and "Today" is a menu option. What does this mean? What happens when I choose "Today" from this menu? "Until today" seems confusing.

@darkLord19 any thoughts on the above?

asaadmahmood:
@darkLord19 Added the open class for the menu wrapper.
https://github.com/mattermost/mattermost-webapp/pull/8058

darkLord19:
> @darkLord19 I'm not sure if I follow the question, but here's how I'm interpreting. The time dropdown is displaying empty if the current day is selected in the date input and there is < 30 minutes left in the current day. Is that the problem?
> 
> If my interpretation above is correct, we could the following:
> 
>     * if there are <30 minutes left in the day, set the default date when the modal opens to tomorrow's date then all times will show for the next day
> 
>     * don't allow users to choose the current day
> 
> 
> Thoughts?
Yes I was thinking something like that only. That makes sense. I have implemented it.

darkLord19:
> > I find it a little confusing that the menu says 'Disable Notifications Until" and "Today" is a menu option. What does this mean? What happens when I choose "Today" from this menu? "Until today" seems confusing.
> 
> @darkLord19 any thoughts on the above?

When user selects `Today` option, dnd is set till midnight today. It may be a bit confusing but I am not sure what wording could be used to make it more clear.

darkLord19:
> @darkLord19 Added the open class for the menu wrapper.
> #8058

Awesome. I will merge master to this branch and checkout the time menu.

asaadmahmood:
@matthewbirtch We can update the text, I guess that's in reference to the spec.
We can change it to:
"Disable notifications for"

Or something.

darkLord19:
> @matthewbirtch We can update the text, I guess that's in reference to the spec.
> We can change it to:
> "Disable notifications for"
> 
> Or something.

"Disable notifications for" will only work for today. for other options it will be confusing. Like "Disable notifications for"  Tomorrow so then does it mean disable notification only for tomorrow?

Another option is that Today option itself could be removed. Menu will have `30 mins, 1 hour, 2 hours, Tomorrow, Custom`

matthewbirtch:
> > @matthewbirtch We can update the text, I guess that's in reference to the spec.
> > We can change it to:
> > "Disable notifications for"
> > Or something.
> 
> "Disable notifications for" will only work for today. for other options it will be confusing. Like "Disable notifications for" Tomorrow so then does it mean disable notification only for tomorrow?
> 
> Another option is that Today option itself could be removed. Menu will have `30 mins, 1 hour, 2 hours, Tomorrow, Custom`

Ya, I'm thinking we can just remove 'Today'. I think there are enough options for users if we exclude 'Today'.

darkLord19:
Removed today option from options

matthewbirtch (approved):
I think this is good now...looks like the issues identified have been resolved

esethna:
@jgilliam17 friendly ping here to help prioritize this feature in your testing queue. There is another feature blocked on this one that we have a contributor team eager to start on. So the faster we can merge this and the related server PR the less conflicts may arise: https://community-daily.mattermost.com/core/pl/k71686aiwjfc9jf54tojuqbngr

i18n/en.json
```diff
@@ -2842,6 +2842,10 @@
   "discard_changes_modal.leave": "Yes, Discard",
   "discard_changes_modal.message": "You have unsaved changes, are you sure you want to discard them?",
   "discard_changes_modal.title": "Discard Changes?",
+  "dnd_custom_time_picker_modal.date": "Date",
+  "dnd_custom_time_picker_modal.defaultMsg": "Disable notifications untill",
```
jgilliam17, line 2846:
```suggestion
  "dnd_custom_time_picker_modal.defaultMsg": "Disable notifications until",
```

jgilliam17 (commented):
Thank you @darkLord19 
Looks great! Thank you for fixing all the outstanding issues.
Seeing a small typo, see suggestion below 
Also, @matthewbirtch asked for indication to the user when their current DND timer is set to expire. Are we adding here or in a follow up PR?

darkLord19:
>asked for indication to the user when their current DND timer is set to expire. Are we adding here or in a follow up PR?
That is done via issue Eric mentioned

michelengelen (approved):

jgilliam17 (approved):
Thank you @darkLord19 
Tested, looks good to merge.
- Verified specific and custom submenu options for setting DND status.

jgilliam17:
@michelengelen  Can you please help merge? Thanks 🙂

michelengelen:
@jgilliam17 Done! Thanks @darkLord19!

PR #7049, opened 2020-11-10
MM-28153: Cypress/E2E: Automate backlogs - System Console > Environment (8 test cases)

catalintomai:
#### Summary
MM-28153: Cypress/E2E: Automate backlogs - System Console > Environment (8 test cases)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28153


DHaussermann:
@catalintomai these tests all run green and seem to line up with the TM4J script 

The one concern I have with this is that we might need some cleanup to make sure the server remain usable for other tests.
Changes to `"WebserverMode` and the S3 config make the server unusable if they get pulled in. These cause issues when I ran the test locally and I wonder if they could prevent Cypress from running followup tests

@saturninoabril any thoughts on best practices here?

saturninoabril:
Thanks for checking @DHaussermann! It's fine. We're resetting [some config values to known default](https://github.com/mattermost/mattermost-webapp/blob/master/e2e/cypress/fixtures/partial_default_config.json#L53) including `"WebserverMode": "gzip"` and S3 config in the global `before` hook which executes before each spec file.

e2e/cypress/integration/system_console/environment_spec.js
```diff
@@ -0,0 +1,331 @@
...
+        // Get config again
+        cy.apiGetConfig().then(({config}) => {
+            // * Verify the database setting value is reset to original value
+            expect(config.ServiceSettings.MinimumHashtagLength).to.eq(minimumHashtagOrig);
+        });
+    });
+
+    it('MM-T994 - Fields editable when enabled, but not saveable until validated', () => {
+        // * Check if server has license for Elasticsearch
+        cy.apiRequireLicenseForFeature('Elasticsearch');
```
saturninoabril, line 231: We're separating EE only from other tests. With that, please move this test into its own spec file at `e2e/cypress/integration/enterprise/system_console` folder. Can be of the same filename.

saturninoabril (changes requested):
Thanks @catalintomai! Looks great! Just one request and I'm all good.

saturninoabril (approved):
Thanks @catalintomai! LGTM.

DHaussermann (approved):
Tested and passed
- Tests work as expected and return green
- Test steps consistent with TM4J

LGTM!

PR #7056, opened 2020-11-11
MM-30087 Remove direct dependency between Client4 and Rudder

hmhealey:
We had a report of a user not being able to use Client4 in a Node environment because of some weirdness from importing Rudder, so this separates Rudder from Client4 by having the web app tell Client4 to explicitly use Rudder.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30087

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1293


reflog (approved):

sudheerDev (approved):

jgilliam17 (approved):
Thank you @hmhealey 
Tested, but I am not seeing any matterlytics tracking in the network tab.
 QA will test after merge, as I don't think that we have rudder key set on test servers.

hmhealey:
I had to remove the new unit test that checked if Rudder was actually called to get this to pass. There's something weird with the imports here that were causing it to not be mocked correctly in CI even though it worked fine outside of CI.

PR #7063, opened 2020-11-12
[MM-30104][MM-30105] Fix up banners in payment_info and billing_subscription pages

nickmisasi:
#### Summary
There were a few issues reported around the credit card expired, and almost expired banners shown in a few places in the admin console. While I couldn't reproduce the reported problems, I was able to discover a few other issues in the process. Weirdly, these issues all would contribute to it _not being possible_ for that banner to display.

` const expiryYear = customer.payment_method.exp_year + 2000;` It was assumed that stripe would return the year "y2k" style as just the last 2 digits, but this looks to have been an error in the test data. Stripe returns years in the format `2020`. 

The `useState` hooks were not synced with redux (since it wasn't a connected component, but used redux hooks instead is my assumption), so there would be weird behaviour, since the customer object wouldn't have been loaded at the time the page was, it would never show the banner (and wouldn't update, or would, but would be sporadic). I've chosen to use selectors for this logic now, so it will display when expected to.

Finally, the billing_subscription page never made the request to get customer information, so the banner would never appear on that page until you navigated to another page that _did_ make that call to populate it in redux, and then went back to the billing subscription page.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30105
https://mattermost.atlassian.net/browse/MM-30104

#### Related Pull Requests
N/A
#### Screenshots
N/A

### DIFF
--- a/components/admin_console/billing/billing_subscriptions.tsx
+++ b/components/admin_console/billing/billing_subscriptions.tsx
@@ -6,7 +6,7 @@ import {useDispatch, useStore, useSelector} from 'react-redux';
 import {FormattedMessage, useIntl} from 'react-intl';
 
 import {getStandardAnalytics} from 'mattermost-redux/actions/admin';
-import {getCloudSubscription, getCloudProducts} from 'mattermost-redux/actions/cloud';
+import {getCloudSubscription, getCloudProducts, getCloudCustomer} from 'mattermost-redux/actions/cloud';
 import {savePreferences} from 'mattermost-redux/actions/preferences';
 import {getConfig, getLicense} from 'mattermost-redux/selectors/entities/general';
 import {makeGetCategory} from 'mattermost-redux/selectors/entities/preferences';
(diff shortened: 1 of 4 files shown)


components/admin_console/billing/payment_info.tsx
```diff
@@ -23,25 +23,24 @@ type Props = {
 
 const PaymentInfo: React.FC<Props> = () => {
     const dispatch = useDispatch<DispatchFunc>();
-    const customer = useSelector((state: GlobalState) => state.entities.cloud.customer);
 
-    const isCreditCardAboutToExpire = () => {
+    const isCardAboutToExpire = useSelector((state: GlobalState) => {
+        const {customer} = state.entities.cloud;
```
marianunez, line 28: Same here, why can't the customer be fetched from the state in the previous `useSlelector`?

nickmisasi: The problem didn't come from the selector to get the customer information, it was the local component state not being updated when the redux state was changed. I'm not 100% on why that was happening, I'm still fairly new to react hooks. 

I could separate back out the customer into its own selector, but figured it was unnecessary since we only need it in this scope.

components/admin_console/billing/billing_subscriptions.tsx
```diff
@@ -53,14 +53,15 @@ const BillingSubscriptions: React.FC<Props> = () => {
...
-    const customer = useSelector((state: GlobalState) => state.entities.cloud.customer);
+
     const products = useSelector((state: GlobalState) => state.entities.cloud.products);
+    const isCardExpired = useSelector((state: GlobalState) => isCustomerCardExpired(state.entities.cloud.customer));
     const getCategory = makeGetCategory();
     const preferences = useSelector<GlobalState, PreferenceType[]>((state) => getCategory(state, Preferences.ADMIN_CLOUD_UPGRADE_PANEL));
 
     const contactSalesLink = useSelector((state: GlobalState) => getCloudContactUsLink(state, InquiryType.Sales));
 
-    const [showCreditCardBanner, setShowCreditCardBanner] = useState(isCustomerCardExpired(customer));
```
nickmisasi, line 63: Commenting here to explain the PR @marianunez (did a little more research into when hooks are called). This line is the problem: `const [showCreditCardBanner, setShowCreditCardBanner] = useState(isCustomerCardExpired(customer));` is equivalent to (more or less)
```
constructor(props) {
  this.state = {showCreditCardBanner: isCustomerCardExpired(customer)}
}
```
Meaning that once the state variable is initialized, subsequent re-renders (like the re-render that happens when the customer entity is updated) wouldn't actually change the state value, because the setting of the initial state ( the call to `isCustomerCardExpired(customer)`) is only run once.

devinbinnie (approved):

components/admin_console/billing/billing_subscriptions.tsx
marianunez: Thank you for the detailed explanation, makes sense to me now!

marianunez (approved):

marianunez:
@stevemudie Given that we can't test this out with CWS on this PR yet, let's defer `QA Review` post-merge after the fast-foward on the 16th and it will be part of the `cloud-ga` branch which can be tested from the test portal.

ethervoid (approved):
LGTM!

PR #7064, opened 2020-11-12
MM-20465 Revert usage of getCurrentChannel to getCurrentChannelId where possible

hmhealey:
A recent change involved switching some usage of `getCurrentChannelId` to `getCurrentChannel`, but that added some more cases where null pointer errors might occur when the current channel is not loaded yet. Since we only use the ID in these places, it's safer to just pass the ID around instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30465


nickmisasi (approved):

Willyfrog (approved):

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified DMs load as expected.
@josephbaylon thanks for running the E2Es.

PR #7066, opened 2020-11-13
MM-30373 - Cypress/E2E: Automate backlogs - Messaging > @ autocomplete (9 test cases)

catalintomai:
#### Summary
Cypress/E2E: Automate backlogs - Messaging > @ autocomplete (9 test cases)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30373


saturninoabril (changes requested):
Thanks @catalintomai! Overall looks good, except for couple of requests.

e2e/cypress/integration/messaging/autocomplete_spec.js
```diff
@@ -0,0 +1,333 @@
...
+        cy.getLastPostId().then((postId) => {
+            // * Verify that the correct system message is displayed or not
+            cy.get(`#postMessageText_${postId}`).should('include.text', message);
+
+            // * Click on the link to add the user to the channel
+            cy.get('a.PostBody_addChannelMemberLink').should('be.visible').click().then(() => {
+                // * Verify that the correct system message is displayed or not
+                cy.getLastPostId().then((npostId) => {
+                    cy.get(`#postMessageText_${npostId}`).should('include.text', `@${notInChannelUser.username} added to the channel by you`);
+                });
```
saturninoabril, line 159: Thanks for updating. However, I'm still getting intermittent failing test here, depending on how fast/slow the system message gets posted. I thought `cy.uiWaitUntilMessagePostedIncludes` before `cy.getLastPostId` will help make it more deterministic. But it might only be my setup so this is non-blocking for me.

saturninoabril (approved):
Thanks @catalintomai! LGTM.

DHaussermann:
@saturninoabril `MM-T2206` is failing for me where there is an assert timing out. 

Screen Shot 2020-11-28 at 2 40 51 PM

Note that it seems the target text is visible in the UI so, I'm unsure why the assert is failing. It does not seem like an issue that would be solved by increasing the timeout. 

`MM-T2207`, `MM-T2209` and `MM-T2212` are also failing for me in a similar way. Maybe it's checking the post above instead of the desired post?
Is this something you're able to reproduce?

catalintomai:
@DHaussermann , I updated a bit the code to harden it and ran the test suite several times without errors. Can you give it a try again?

DHaussermann:
`MM-T2207` and  `MM-T2209` are failing. 

After reviewing this again, it seems this is not the same issue as the 1st test pass. 0/5 but what I'm seeing `MM-T2207` clears the cookies and the authentication is deliberately lost. `MM-T2207` Expects less cookies to exist in my browser. @catalintomai  Please see if you car repro.
Screen Shot 2020-12-08 at 5 12 25 PM




`MM-T2209` There are 2 instances where `cy.visit(/${testTeam.name}...` is followed by `cy.get('#post_textbox')...` On my local, it takes about 1 extra seconds to load in the webapp. I can make the test pass by adding a wait on line  `231` and `260`

`MM-T2212` and `MM-T2214` are also failing sometimes possibly for similar reasons. 

@saturninoabril any thoughts on this `MM-T2209`? I'm curious if maybe you see this. but also there is no support function being used where I can adjust the time out. it's simply `cy.visit` followed by `cy.get` and I know it's not ideal to just add in a wait like this.

catalintomai:
@DHaussermann , though i wasn't able to repro, I've added the wait (I was already doing it in some other places) and added some more hardening. Can you give it one more try? Thanks = ).

e2e/cypress/integration/messaging/autocomplete_spec.js
```diff
@@ -0,0 +1,334 @@
...
+                            // * Verify that the correct system message is displayed or not
+                            cy.getLastPostId().then((npostId) => {
+                                cy.get(`#postMessageText_${npostId}`).should('include.text', `@${notInChannelUser.username} added to the channel by you`);
+                            });
+                        });
+                    });
+                });
+            });
+        });
+    });
```
saturninoabril, line 162: @DHaussermann Thanks for running the tests.
@catalintomai I ran it again and got only 1 out of several runs with all tests passed; the rests randomly failed like what Dylan have seen. My recommendations would be:
1. To use `cy.uiWaitUntilMessagePostedIncludes` before doing `cy.getLastPostId` since other ephemeral post is not being posted before `cy.getLastPostId` gets executed.
2. To avoid `.then` as much as possible especially for commands that's not expecting return value/s, since the asynchronous nature of commands are being handled by [Cypress itself](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Commands-Are-Asynchronous). (nit, non-blocking, make more Cypress-y ;) and less Javascript pattern ) 

Something like,

```javascript
    it('MM-T2206 @ autocomplete - not in channel (center), have permission to add (public channel)', () => {
        const message = `@${notInChannelUser.username} did not get notified by this mention because they are not in the channel. Would you like to add them to the channel? They will have access to all message history.`;

        cy.visit(`/${testTeam.name}/channels/${testChannel.name}`).wait(TIMEOUTS.FIVE_SEC);

        // # Clear then type @
        cy.get('#post_textbox').should('be.visible').clear().type('@');

        // * Verify that the suggestion list is visible
        cy.get('#suggestionList').should('be.visible');

        // # Type user name
        cy.get('#post_textbox').type(`${notInChannelUser.username}`);

        // # Post user mention
        cy.get('#post_textbox').type('{enter}{enter}');
        cy.uiWaitUntilMessagePostedIncludes('They will have access to all message history.');

            // * Verify that the correct system message is displayed or not
            cy.get(`#postMessageText_${postId}`).should('include.text', message);

            // * Click on the link to add the user to the channel
            cy.get('a.PostBody_addChannelMemberLink').should('be.visible').click();
            cy.uiWaitUntilMessagePostedIncludes('added to the channel by you');

            // * Verify that the correct system message is displayed or not
            cy.getLastPostId().then((npostId) => {
                cy.get(`#postMessageText_${npostId}`).should('include.text', `@${notInChannelUser.username} added to the channel by you`);
            });
        });
    });
```

catalintomai: Thanks @saturninoabril , I updated the code following your suggestions. The tests pass in my env (thought they were passing before as well), hopefully they'll pass on your envs. as well = )
cc: @DHaussermann

saturninoabril (approved):
@catalintomai Thank you so much! Ran multiple times and consistently passing on me.
Screen Shot 2020-12-10 at 2 44 50 PM
Screen Shot 2020-12-10 at 2 50 14 PM
Screen Shot 2020-12-10 at 3 00 32 PM

DHaussermann (approved):
Updated Cy Status on corresponding TM4J tickets.

PR #7068, opened 2020-11-13
[MM-27927] - Add limit for visible DMs and DM sorting menu

nevyangelova:
#### Summary
This adds the category sorting menu that needs to be implemented in the category menu as well. https://mattermost.atlassian.net/browse/MM-30163 also needs to be implemented. I will add the action as a constant after QA is done testing it. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27927
https://mattermost.atlassian.net/browse/MM-30163

#### Related PRs
https://github.com/mattermost/mattermost-redux/pull/1300 
https://github.com/mattermost/mattermost-redux/pull/1310


nevyangelova:
@matthewbirtch please advice on the text for all direct messages, shown in the parent menu.

matthewbirtch:
> @matthewbirtch please advice on the text for all direct messages, shown in the parent menu.

I think we can just make it say "All"

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,213 @@
...
+    isLeaving: boolean;
+
+    constructor(props: Props) {
+        super(props);
+
+        this.state = {
+            openUp: false,
+            width: 0,
+            selectedDmNumber: 20,
+            selectedCategorySort: CategorySorting.Alphabetical,
```
hmhealey, line 55: These should be coming from redux, but I assume it'll be connected once the mattermost-redux stuff is done.

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,213 @@
...
+    selectedDmNumber: number;
+    selectedCategorySort: string;
+};
+
+enum Settings {
+    limitVisibleDmsGms = 'limit_visible_dms_gms',
+}
+
+export class SidebarCategorySortingMenu extends React.PureComponent<Props, State> {
+    isLeaving: boolean;
```
hmhealey, line 46: This can be removed since it's not used.

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,213 @@
...
+        };
+
+        this.isLeaving = false;
+    }
+
+    handleSortDirectMessages = (sorting: CategorySorting) => {
+        const {category} = this.props;
+
+        this.props.actions.setCategorySorting(category.id, sorting);
+        trackEvent('ui', `ui_sidebar_sort_dm_${sorting}`);
```
hmhealey, line 65: @esethna This makes me think, do we want telemetry events for this?

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,213 @@
...
+        this.setState({selectedCategorySort: sorting});
+    }
+
+    handlelimitVisibleDMsGMs = (number: number) => {
+        const {currentUserId} = this.props;
+        this.props.actions.savePreferences(currentUserId, [{
+            user_id: currentUserId,
+            category: Constants.Preferences.CATEGORY_SIDEBAR_SETTINGS,
+            name: Settings.limitVisibleDmsGms,
+            value: `${number}`,
```
hmhealey, line 75: This might be a slightly more idiomatic way to do this
```suggestion
            value: number.toString(),
```

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,213 @@
...
+            } as SubmenuItemProps;
+        });
+
+        const categoryMenuItems: SubmenuItemProps[] = [];
+        categoryMenuItems.push(
+            {
+                id: 'showAllDms',
+                direction: 'right' as any,
+                text: intl.formatMessage({id: 'sidebar.allDirectMessages', defaultMessage: 'All direct messages'}),
+                action: () => this.handlelimitVisibleDMsGMs(Infinity),
```
hmhealey, line 120: I think we might want to use 0 or -1 for this instead of infinity. Infinity is technically correct for this and it makes the selector slightly simpler since you can just pass it right into slice, but I could see there being difficulties since storing the string "infinity" feels weird.

components/sidebar/sidebar_menu/sidebar_menu.tsx
```diff
@@ -159,7 +161,7 @@ export default class SidebarMenu extends React.PureComponent<Props, State> {
                 >
                     <button
                         ref={this.menuButtonRef}
-                        className='SidebarMenu_menuButton'
+                        className={classNames([`SidebarMenu_menuButton ${additionalClass}`])}
```
hmhealey, line 164: You can just pass `additionalClass` in as a separate argument, and `classNames` will sort out including it if it's actually set. Calling it this way ends up sometimes setting `undefined` as a CSS class like in the snapshot for this file.

components/user_settings/sidebar/user_settings_sidebar.tsx
```diff
@@ -859,7 +859,7 @@ export default class UserSettingsSidebar extends React.PureComponent<UserSetting
 
         const channelOrganizationSection = (showChannelOrganization && channelSidebarOrganizationDisabled) ? this.renderChannelOrganizationSection() : null;
         const channelSidebarOrganizationSection = showChannelSidebarOrganization ? this.renderChannelSidebarOrganizationSection() : null;
-        const autoCloseDMSection = showUnusedOption ? this.renderAutoCloseDMSection() : null;
+        const autoCloseDMSection = showUnusedOption && !channelSidebarOrganizationDisabled ? this.renderAutoCloseDMSection() : null;
```
hmhealey, line 862: Should this maybe be added wherever `showUnusedOption` comes from?

components/widgets/menu/menu_items/submenu_item.test.tsx
```diff
@@ -20,7 +22,15 @@ describe('components/widgets/menu/menu_items/submenu_item', () => {
                 root={true}
             />,
         );
-        expect(wrapper.html()).toEqual('<li class="SubMenuItem MenuItem" role="menuitem" id="1_menuitem"><div id="1" tabindex="0"><span id="channelHeaderDropdownIconLeft_1" class="fa fa-angle-left SubMenu__icon-left-empty" aria-label="submenu icon"></span>test<span id="channelHeaderDropdownIconRight_1" class="fa fa-angle-right SubMenu__icon-right-empty" aria-label="submenu icon"></span><ul class="a11y__popup Menu dropdown-menu SubMenu" style="visibility: hidden;"></ul></div></li>');
+
+        expect(wrapper.html()).toEqual(format(`<li class="SubMenuItem MenuItem" role="menuitem" id="1_menuitem">
```
hmhealey, line 26: Not sure why we use `wrapper.html()` here, but I definitely appreciate the helper to make it easier to read. Are there any changes here beyond being easier to read?

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -44,6 +46,7 @@ type Props = {
     show?: boolean;
     direction?: 'left' | 'right';
     openUp?: boolean;
+    categorySortingMenu?: boolean;
```
hmhealey, line 49: We shouldn't be adding anything sidebar-specific to this since it's a common component.

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -179,25 +182,34 @@ export default class SubMenuItem extends React.PureComponent<Props, State> {
         if (!isMobile) {
             subMenuContent = (
                 <ul
-                    className={'a11y__popup Menu dropdown-menu SubMenu'}
+                    className={classNames(['a11y__popup Menu dropdown-menu SubMenu', {categorySortingMenu}])}
                     style={subMenuStyle}
                 >
                     {hasSubmenu ? subMenu!.map((s) => {
+                        const hasDivider = s.id === 'SidebarChannelMenu-moveToDivider';
```
hmhealey, line 189: For this, maybe we could export a constant that causes a divider to be rendered instead of the normal `SubMenuItem`?

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -222,13 +235,14 @@ export default class SubMenuItem extends React.PureComponent<Props, State> {
                 >
                     <span
                         id={'channelHeaderDropdownIconLeft_' + id}
-                        className={'fa fa-angle-left SubMenu__icon-left' + (hasSubmenu && !isMobile && (direction === 'left') ? '' : '-empty' + (isMobile ? ' mobile' : ''))}
+                        className={classNames([`fa fa-angle-left SubMenu__icon-left${hasSubmenu && !isMobile && (direction === 'left') ? '' : '-empty'}`, {mobile: isMobile, 'sorting-menu-dropdown-icon': categorySortingMenu}])}
```
hmhealey, line 238: Same comment here about classNames. Instead of a template string, you can just pass the `SubMenu__icon-left`/`SubMenu__icon-left--empty` as another argument

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -222,13 +235,14 @@ export default class SubMenuItem extends React.PureComponent<Props, State> {
...
                         aria-label={Utils.localizeMessage('post_info.submenu.icon', 'submenu icon').toLowerCase()}
                     />
                     {textProp}
+                    {selectedValueText && <span className='selected'>{selectedValueText}</span>}
                     <span
                         id={'channelHeaderDropdownIconRight_' + id}
-                        className={'fa fa-angle-right SubMenu__icon-right' + (hasSubmenu && (isMobile || direction === 'right') ? '' : '-empty')}
+                        className={classNames([`fa fa-angle-right SubMenu__icon-right${hasSubmenu && !isMobile && (direction === 'right') ? '' : '-empty'}`, {mobile: isMobile, 'sorting-menu-dropdown-icon': categorySortingMenu}])}
```
hmhealey, line 245: And here

hmhealey (changes requested):

components/sidebar/sidebar_category/sidebar_category_sorting_menu/index.ts
```diff
@@ -0,0 +1,52 @@
...
+        const currentTeam = getCurrentTeam(state);
+
+        if (currentTeam) {
+            categories = getCategoriesForTeam(state, currentTeam.id);
+        }
+
+        return {
+            categories,
+            category: ownProps.category,
+            handleOpenMoreDirectChannelsModal: ownProps.handleOpenMoreDirectChannelsModal,
```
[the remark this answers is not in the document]
nevyangelova: sorry leftover code from a previous approach

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+type State = {
+    openUp: boolean;
+    width: number;
+    selectedDmNumber: number;
+    selectedCategorySort: string;
+};
+
+enum Settings {
+    limitVisibleDmsGms = 'limit_visible_dms_gms',
```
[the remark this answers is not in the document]
nevyangelova: correct

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
nevyangelova: If we pass 0 or -1 or even empty string or null it will remove all items from the array and that adds an extra check for the selector if the value === 0, 3/5 this is more simplified IMO

components/user_settings/sidebar/user_settings_sidebar.tsx
nevyangelova: its coming from 'config.CloseUnusedDirectMessages === 'true'' and the naming is not explicit that the sidebar organisation is disabled unless we change the naming or create a function for it. I am 1/5 on this because it seems clearer to me and we already have the 'channelSidebarOrganizationDisabled' explicitly stored in the file, but I can create a function if you like.

components/widgets/menu/menu_items/submenu_item.test.tsx
nevyangelova: There aren't. Previously I decided to hide the icons completely by setting 'display: block' but that ended up confusing with the styles even more.. so now I just left the helper :P

nevyangelova: I was thinking to open a ticket to refactor this into a proper snapshot file?

components/widgets/menu/menu_items/submenu_item.tsx
nevyangelova: There are many new style changes it's impossible to override without breaking anything unless there is an additional prop. We can take this into consideration when we do the Menu common component, which I advocate to be soon :P

components/widgets/menu/menu_items/submenu_item.tsx
nevyangelova: It actually is more work than one would think because of the submenu and the types it takes, I will need to add more HTML etc. I say we keep it in mind for the new Menu.

components/widgets/menu/menu_items/submenu_item.tsx
nevyangelova: in this case the class is part of the initial string so it wouldn't work passing it as an argument

nevyangelova: I tried but it doesn't work as an additional class, the CSS is only applied as an addition to the class, only one or the other can be set, we can make both conditional but IMO it will be more confusing?

matthewbirtch (changes requested):
Looking great @nevyangelova. A few questions and tweaks to make:

1. Is this PR just for sorting for DMs? I don't see the sorting options here for custom categories - maybe that's coming in another PR?
image

2. For a menu item that has a checkmark, the hover state doesn't go the full width of the menu
image

3. Need to line up the icons and text like they are in the other menus. See the comparison of the channel context menu (top) and DM menu (bottom)
context-menu-red-line

4. Looks like the chevrons on the right are also aligned differently from other menus


5. Also, "20" and "Recent activity" aren't quite right-aligned with each other.
image

6. Chevrons are missing on mobile webview. Was that intentional?


7. Not necessarily part of this PR, but can we change the opacity of the submenu chevrons on all context menus to be 56% `center-channel-color`?

When we refactor the component, I'd like to make adjustments to a number of other things, but I'll leave those out of this PR.

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,204 @@
...
+        );
+
+        const browseDirectMessages = (
+            <Menu.Group>
+                <Menu.ItemAction
+                    additionalClass=' SidebarCategorySortingMenu__ActionItem'
+                    id={'browseDirectMessages'}
+                    onClick={this.props.handleOpenDirectMessagesModal}
+                    icon={<i className='icon-account-plus-outline'/>}
+                    text={intl.formatMessage({id: 'sidebar.openDirectMessage', defaultMessage: 'Open a direct message'})}
```
devinbinnie, line 126: 0/5 but should we be using `FormattedMessage` here instead?

sass/layout/_sidebar-left.scss
```diff
@@ -358,6 +358,70 @@
...
+            }
+
+            .sorting-menu-checkbox i {
+                color: var(--button-bg);
+            }
+
+            .selected {
+                text-transform: none;
+                font-size: 12px;
+                color: var(--center-channel-color-56);
```
devinbinnie, line 420: Can we move towards using the RGB variables?
```suggestion
                color: rgba(var(--center-channel-color-rgb), 0.56);
```

devinbinnie (approved):
Couple non-blocking comments, but otherwise looks good!

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
hmhealey: I mean, do we want telemetry events when a user changes their number of visible DMs?

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
hmhealey: My concern is that Infinity isn't a valid integer, so we're somewhat mixing types with the value stored for this. I just checked and storing `Infinity.toString()` in a preference works fine, but then when we pull it back out of preferences `parseInt('Infinity')` returns NaN instead of infinty, so we might end up with problems because we're working outside of well-defined numbers.

components/user_settings/sidebar/user_settings_sidebar.tsx
hmhealey: Fair enough. I think `channelSidebarOrganizationDisabled` could be used to generate `showUnusedOption` since it's important for whether or not to show that section, but I see your point about `channelSidebarOrganizationDisabled` being used elsewhere as well.

components/widgets/menu/menu_items/submenu_item.test.tsx
hmhealey: That would be a good idea unless there's some reason this wasn't done as a snapshot in the first place. I know the menu components do some stuff that's not React-friendly, so perhaps that's why snapshots weren't used here before.

components/widgets/menu/menu_items/submenu_item.tsx
hmhealey: What's different with the sorting menu? If it's the checkboxes and selected item, I think we could have a more general name for these styles and props.

components/widgets/menu/menu_items/submenu_item.tsx
hmhealey: I don't know. I think a complicated ternary inside a template string inside a rather long line is pretty complicated as-is. Passing `(hasSubmenu && !isMobile && direction === 'left') ? 'SubMenu__icon-left' : 'SubMenu__icon-left-empty'` as an argument into classNames saves some of that nesting at least

nevyangelova:
@matthewbirtch thanks! I thought they had different padding from the other menus based on the design. But unifying it made styling easier! Please check again :) Also yes, the menu chevrons are not visible on mobile. I added the colour you requested to all chevrons.

matthewbirtch:
Thanks @nevyangelova. Apologies for the inconsistency in the design files - a consequence of moving too fast and transferring the project between designers.  

It's looking better now, but just a few minor tweaks:

On mobile webview:
- I think the chevron is still needed to indicate the submenu. What were your thoughts on the rationale behind removing?
- The sorting text is not right aligned
Screen Shot 2020-11-26 at 9 24 16 AM

jgilliam17:
@nevyangelova 
I am seeing an issue with the left dropdown chevron on DM category - it's missing when sorting by show all
Screen Shot 2020-12-01 at 4 20 45 PM

nevyangelova:
@jgilliam17 @matthewbirtch resolved :)

matthewbirtch:
@nevyangelova thanks for resolving the styling issues. Those look good now. However, I'm also seeing what @jgilliam17 is seeing. When I first log in, the DM category is not showing anything and is not showing the expand/collapse chevron. 

image

As soon as I change the number of DMs shown from 20 to anything else, the DMs show up and everything starts working as expected.

Having said that, I can't reliably reproduce this issue. I just know I saw it when I first logged in as user-1

nevyangelova:
@matthewbirtch the state needs re-setting I fixed the issue but the initial load needs to be reset in the PR server.

matthewbirtch (approved):
Thanks @nevyangelova. I think we're good to go!

jgilliam17:
Thanks @nevyangelova 
Category sorting menu looks good.
There is an issue with currently open DM/GM not being visible on the sidebar when changing the `show` number.
For example:
Add 20+ DMs/GMs
Select Alphabetical and show all
Scroll to the bottom of the list on the LHS and open the last DM/GM
Select Show 10

Observed: DM/GM that is currently open is not visible on the sidebar

nevyangelova:
> Thanks @nevyangelova
> Category sorting menu looks good.
> There is an issue with currently open DM/GM not being visible on the sidebar when changing the `show` number.
> For example:
> Add 20+ DMs/GMs
> Select Alphabetically and show all
> Scroll to the bottom of the list on the LHS and open the last DM/GM
> Select Show 10
> 
> Observed: DM/GM that is currently open is not visible on the sidebar

thanks for spotting this, that would be a change in the mattermost-redux repo, so I guess this one is good to go, apart from updating the redux version when the bug is fixed.

nevyangelova:
@jgilliam17 fixed!

jgilliam17:
Thanks @nevyangelova !
- I am still able to reproduce open DM/GM not being on the LHS when showing less. 
- Default `show 20`: I'm seeing 20 preselected at each login, despite user having selected a different number e.g.10 previously, log out, log back in and user sees 10 on the sidebar again, but 20 checked on the menu.
- Refreshing also resets the menu to alpha+show 20, but keeps what the user previous set on the sidebar
e.g. 
select show 40 by recent
refresh
sidebar still shows 40 by recent 
Observed: menu has alpha + show 20 checked, not what's shown on the sidebar
I'm not sure I understand when/how default setting should work. Can you please help me with some examples. Thanks 🙂

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
hmhealey: This should also be coming from redux

hmhealey (changes requested):
I think these components still need to be connected to redux. Some of them are still using component state

nevyangelova:
> I think these components still need to be connected to redux. Some of them are still using component state

@hmhealey I changed the preference type to be coming from redux but the other one is already coming from CategorySorting in mm-redux. Is there anything else?

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,200 @@
...
+        setCategorySorting: (categoryId: string, sorting: CategorySorting) => void;
+        savePreferences: (userId: string, preferences: PreferenceType[]) => void;
+    };
+};
+
+type State = {
+    openUp: boolean;
+    width: number;
+    selectedDmNumber: number;
+    selectedCategorySort: string;
```
hmhealey, line 39: The DM count needs to come from preferences, and the sorting method needs to come from the category itself still.

components/sidebar/sidebar_category/sidebar_category_sorting_menu/sidebar_category_sorting_menu.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+import SidebarMenu from 'components/sidebar/sidebar_menu';
+import SidebarMenuType from 'components/sidebar/sidebar_menu/sidebar_menu';
+import Menu from 'components/widgets/menu/menu';
+import {Props as SubmenuItemProps} from 'components/widgets/menu/menu_items/submenu_item';
+
+type Props = {
+    category: ChannelCategory;
+    handleOpenDirectMessagesModal: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+    categories?: ChannelCategory[];
```
hmhealey, line 23: This prop can be removed since it's not being used anywhere.

hmhealey (changes requested):

hmhealey:
@nevyangelova You'll need to re-add the changes to package-lock.json for the build to pass. I'm surprised CI passes when package.json and package-lock.json don't match because I thought it uses a version of `npm install` that throws an error if they don't.

nevyangelova:
> @nevyangelova You'll need to re-add the changes to package-lock.json for the build to pass. I'm surprised CI passes when package.json and package-lock.json don't match because I thought it uses a version of `npm install` that throws an error if they don't.
> 
> Otherwise, it looks like everything has been fixed, but I'd like to double check on the test server.

I got totally confused because we tell contributors to not commit package-lock 🤦

hmhealey:
Ah, I see. We need it when something actually changes, but not when it's just changing randomly because npm likes to do that sometimes.

jgilliam17:
@nevyangelova looks good, I think we are almost there
I am seeing one issue
Example:
1. User has a long list of DMs
2. Selects to display 10 a-z
3. Selects to display the same number by recency 

Question: Should we go back to check all available before displaying 10 by recency? or 
Are we just rearranging what we currently display, in this case the 10 DMs already showing? <this is observed on test server, it wasn't what I expected

UX question for @matthewbirtch: 
Now that we have a DM menu option to Open a Direct Message, I was wondering if we need to keep `+` on DM category. 
Screen Shot 2020-12-16 at 3 38 05 PM

matthewbirtch:
> UX question for @matthewbirtch:
> Now that we have a DM menu option to Open a Direct Message, I was wondering if we need to keep `+` on DM category.

@jgilliam17 I think we will eventually remove this, but for now we are going to leave it for users who are used to this interaction. Sometimes we leave redundancy to help the transition to new UI changes

hmhealey:
> Question: Should we go back to check all available before displaying 10 by recency? or
Are we just rearranging what we currently display, in this case the 10 DMs already showing? <this is observed on test server, it wasn't what I expected

I think it's correct to have it rearrange what's already visible. The limit of 10 is based on the ones that you've most recently viewed, and that's independent of whether they're sorted alphabetically or by most recent activity. It's also important to clarify the difference between recently viewed and recent activity there since we've definitely confused ourselves with that before 😛

jgilliam17:
Thanks Harrison for the explanation. Yes, I was  definitely confused on viewed vs. activity.  

1. DM that has not been recently viewed or is brand new, while we are filtering e.g. 10 alpha, is not shown on the sidebar when new message is received.

https://user-images.githubusercontent.com/52937121/102803286-04320e00-4386-11eb-84c3-c8ed8f564d79.mov

2. Bottom `More unreads` sidebar toast does not dismiss when user filters by less DMs
Steps: 
Have a long list of DMs with some unreads on the bottom of the list (not in first 10 dms)
Scroll the sidebar, so that bottom more unreads toast is shown
Filter down to display only 10 DMs
Expected: More Unreads toast is dismissed
Observed: More Unreads toast is still shown
Screen Shot 2020-12-21 at 11 55 47 AM

nevyangelova:
@jgilliam17 fixed

jgilliam17:
Thanks @nevyangelova 
I am still seeing both issues 
- new DM not showing up on the sidebar as soon as it was received 
- ↓ More Unreads sidebar toast is slow to show up, when expanding to show all, and persists when shrinking the list to 10 or 15. I varied my browser window size and can repro consistently
 
@hmhealey - thanks again for explaining the rules on how we sort alphabetically (last viewed, then a-z) 
I was concerned about our label `Alphabetically` not being descriptive enough as our a-z sort is more of a a-z hybrid, where the `last viewed` is not readily evident  e.g when viewing `all/alpha` 50+ DMs/GMs > all are sorted in true a-z order, when setting the limit down to 10, I was surprised by the way a-z 10 count looked until Nevy and you described the sort rules to me. How easy will this be for our users to pick up on?

nevyangelova:
> Thanks @nevyangelova
> I am still seeing both issues
> 
> * new DM not showing up on the sidebar as soon as it was received
> * ↓ More Unreads sidebar toast is slow to show up, when expanding to show all, and persists when shrinking the list to 10 or 15. I varied my browser window size and can repro consistently
> 
> @hmhealey - thanks again for explaining the rules on how we sort alphabetically (last viewed, then a-z)
> I was concerned about our label `Alphabetically` not being descriptive enough as our a-z sort is more of a a-z hybrid, where the `last viewed` is not readily evident e.g when viewing `all/alpha` 50+ DMs/GMs > all are sorted in true a-z order, when setting the limit down to 10, I was surprised by the way a-z 10 count looked until Nevy and you described the sort rules to me. How easy will this be for our users to pick up on?

@jgilliam17 I can't repro neither of those. To your last question, after thinking about it it makes sense that when we do the cutoff it takes the most recent DM's and not sort alphabetically for example and DM's that aren't opened in a long time are prioritised.

hmhealey:
If we get feedback that people are confused, we could work on maybe clarifying what sorting by "recency" is (most recent post vs most recently viewed). I've had to be very clear which of those I'm talking about when we're discussing this, so that might need to carry over a bit more to the UI text.

nevyangelova:
@hmhealey @jgilliam17 I cannot repro the DM not appearing in Sidebar bug. Could it be related to the Jtest account? It might not be considering the account as a teammate and therefore skipping the check. Is it a bot account?

hmhealey (changes requested):
I haven't done a code review yet, but I noticed a couple minor styling bugs in the menu. There's a margin on the left of the divider in the submenu, and the checkmark is missing when showing all
Screen Shot 2021-01-11 at 5 43 52 PM

Also, this isn't the cause of the missing checkmark, but, @matthewbirtch, what if we change the first option to just be "All" instead of "All Direct Messges"? I think that's a bit more concise that way

jgilliam17:
Thanks @nevyangelova 
I am seeing new DMs appear on the sidebar, so that issue is fixed and no longer reproducing for me.
The More Unreads toast issue reproduces for me consistently. I don't mind if we need to address this separately if you believe it needs more investigation due to a difference in the the way this is calculated and may depend/be based on user screen size.
There are 2 new issue I am seeing on the test server
1. DM is removed from sidebar as soon as you open it. I am seeing this for sysadmin user, DMs sorted by alpha/displaying 10 See video.

https://user-images.githubusercontent.com/52937121/104363967-f0d91500-54e3-11eb-87e7-38c4bd20cb0d.mov

2. When using sort by Recent Activity/count 10 the list rearranges in real time as soon as you open a DM. I wouldn't expect the recency to be in real time, only when selecting/changing sort order

https://user-images.githubusercontent.com/52937121/104366177-f97f1a80-54e6-11eb-8761-54c8f9c22c0c.mov

hmhealey (approved):

jgilliam17 (approved):
Thank you @nevyangelova and @hmhealey 
Tested, looks good to merge.
- Verified sorting menu and submenu styling.
- Verified default setting for displaying DMs is set to sorted by Recency showing 20 count.
- Verified various limits around sorting and displaying DMs on the sidebar, taking into consideration last viewed time is used as main criteria in determining which DMs are displayed.

PR #7071, opened 2020-11-13
Fix alignment when no icon is provided in post actions

larkox:
#### Summary
Fix alignment when no icon is provided in post actions

#### Ticket Link
None

#### Related Pull Requests
None

### DIFF
--- a/components/dot_menu/dot_menu.tsx
+++ b/components/dot_menu/dot_menu.tsx
@@ -336,12 +336,17 @@ export default class DotMenu extends React.PureComponent<Props, State> {
             }) || [];
 
         const appsBindings = this.props.appsBindings.map((item) => {
+            let icon: JSX.Element | undefined;
+            if (item.icon) {
+                icon = (<img src={item.icon}/>);
+            }
+
             return (
                 <Menu.ItemAction
                     text={item.label}
                     key={item.app_id + item.location_id}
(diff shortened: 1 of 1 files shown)


components/dot_menu/dot_menu.tsx
```diff
@@ -336,12 +336,17 @@ export default class DotMenu extends React.PureComponent<Props, State> {
             }) || [];
 
         const appsBindings = this.props.appsBindings.map((item) => {
+            let icon: JSX.Element | undefined;
+            if (item.icon) {
```
levb, line 340: if this is the case, I am 1/5 inclined to rename the field iconurl. We need to decide on restrictions we apply to static assets - should probably be served out of the rooturl sub-URL, or the marketplace.

levb (approved):

mickmister (approved):

components/dot_menu/dot_menu.tsx
larkox: This is the case now, and I think it is where we should move forward (easy, allow both internal and external files, and since it is the img src, it should be "safe"), but I think we haven't discussed yet what we want here. Also, we should document what format do we allow here. At the moment would be "anything that can be interpreted by the `` html tag.

levb: Filed https://mattermost.atlassian.net/browse/MM-30595

PR #7081, opened 2020-11-16
Cypress/E2E: Fix toast appears unread spec

josephbaylon:
#### Summary
- Per this [comment](https://mattermost.atlassian.net/browse/MM-30534?focusedCommentId=101466), we just need to increase the content size to trigger the div toast
- Specifically testing for 1000 pixels might be finicky when they change later so I opted to just increase the content

@saturninoabril I didn't add v5.29 milestone since today is release day but feel free to add if you think we should

#### Ticket Link
NA

Screen Shot 2020-11-16 at 10 27 40 AM

### DIFF
--- a/e2e/cypress/integration/mark_as_unread/toast_appears_unread_spec.js
+++ b/e2e/cypress/integration/mark_as_unread/toast_appears_unread_spec.js
@@ -52,7 +52,7 @@ describe('Verify unread toast appears after repeated manual marking post as unre
                             Cypress._.times(30, (index) => {
                                 cy.postMessageAs({
                                     sender: otherUser,
-                                    message: index.toString(),
+                                    message: `${index.toString()}\nsecond line\nthird line\nfourth line`,
                                     channelId: townSquareChannel.id,
                                 });
                             });


saturninoabril (approved):
Thanks @josephbaylon! Tested and passed. I thought this is good to cherry-pick into release-5.29, being no effect to product but greatly improve testing. However, I'll defer to @amyblais.

amyblais:
I think v5.29 is fine, though it was already released on Friday.

furqanmlk (approved):
working as expected 👍 
I am just wondering there will always be one div.toast 
cy.findAllByText('Jump to recents').eq(0).should('be.visible'); // works as well

josephbaylon:
> working as expected 👍
> I am just wondering there will always be one div.toast
> cy.findAllByText('Jump to recents').eq(0).should('be.visible'); // works as well

@furqanmlk Thanks for the feedback! We can discuss that strategy in one of our meetings as that would require a whole bunch of other tests to be modified so that they are all consistent. For now, I'll opt to just fix the issue and keep as is.

furqanmlk:
> > working as expected 👍
> > I am just wondering there will always be one div.toast
> > cy.findAllByText('Jump to recents').eq(0).should('be.visible'); // works as well
> 
> @furqanmlk Thanks for the feedback! We can discuss that strategy in one of our meetings as that would require a whole bunch of other tests to be modified so that they are all consistent. For now, I'll opt to just fix the issue and keep as is.

Absolutely 👍

PR #7082, opened 2020-11-16
MM-24072-Position the autocomplete popover relative to the @,~,/ trigger in the post draft

pablovelezvidal:
#### Summary
This pull request adds new functionality to the autocomplete popover list. With this, the popover will get positioned in relation to the @, or / trigger that was written within the post draft.

To get this functionality, we get the caret-x position within the textArea. The solution creates a div>span copy of the textArea with all the styles and the inner text, and then gets the x position of the caret relative to the rect position of the textArea. Then, we also calculate the offset left of the textArea including the paddings, margins and avatar icon sizes. Once we have all these values we make simple calculations to determine:

1) If the x position alignment is less than the post draft rect top-left position, we place the popover at the beginning of the post draft.
2)if the x position is greater that the top-left position, we use the returned value by the calculation to place the popover in that position.
3)if the popover rect top-right position will overflow the most right side part of the viewport, the popover is placed at the end of the post-draft.

This works for edit post, the post thread list and normal post tagging.
Tested in Chrome, Firefox and Edge.

#### Ticket Link
  Fixes:
   https://github.com/mattermost/mattermost-server/issues/15647
   https://mattermost.atlassian.net/browse/MM-24072

#### Screenshots
Screen Shot 2020-11-16 at 8 33 23 PM
Screen Shot 2020-11-16 at 8 33 49 PM
Screen Shot 2020-11-16 at 8 34 10 PM
Screen Shot 2020-11-16 at 8 34 35 PM
Screen Shot 2020-11-16 at 8 34 52 PM
Screen Shot 2020-11-16 at 8 35 24 PM


utils/utils.jsx
```diff
@@ -1066,6 +1066,158 @@ export function getCaretPosition(el) {
...
+
+    // how many pixels to the right should be moved the suggestion box
+    const pxToTheRight = (caretXCoordinateInTxtArea + txtAreaPaddingLft) - (pxToSubstract);
+
+    // the x coordinate in the viewport of the suggestion box border-right
+    const xBoxRightCoordinate = caretXCoordinateInTxtArea + txtAreaOffsetLft + suggestionBoxWidth;
+
+    const willSuggestionBoxOverflowRight = isBoxOutOfRightSideViewport(viewportWidth, xBoxRightCoordinate);
+    return {
+        IsOutOfRightSideViewport: willSuggestionBoxOverflowRight,
```
hmhealey, line 1199: Could you rename this to start with a lower case i to keep it consistent with other fields and reduce the potential for typos?

utils/utils.jsx
```diff
@@ -1066,6 +1066,158 @@ export function getCaretPosition(el) {
...
+        pixelsToMove: Math.max(0, Math.floor(pxToTheRight)),
+    };
+}
+
+export function getPixelsToSubstract() {
```
hmhealey, line 1204: Perhaps this could be passed into `getSuggestionBoxAlgn` instead of being calculated as part of it? I could see this sort of code being used elsewhere where we might just want to find the position of the caret for other reasons.

hmhealey: Also, this might not be what we want when we're displaying other content in the autocomplete. I'll follow up with UX about this outside of the review.

utils/utils.jsx
```diff
@@ -1066,6 +1066,158 @@ export function getCaretPosition(el) {
...
+    const viewportWidth = getViewportSize().w;
+
+    // value in pixels used in suggestion-list__content class line 72 file _suggestion-list.scss
+    const suggestionBoxWidth = Constants.SUGGESTION_LIST_MODAL_WIDTH;
+
+    // value in pixels for the offsetLeft for the textArea
+    const txtAreaOffsetLft = offsetTopLeft(textArea).left;
+
+    // textArea padding left of 15px defined in the _post.scss line 392 + 1px from the border
+    const txtAreaPaddingLft = Constants.TEXTAREA_PADDING_LEFT + Constants.TEXTAREA_BORDER_WIDTH;
```
hmhealey, line 1187: Why do we need to account for the padding here? If `caretXCoordinateInTxtArea` is the screen coordinates of the caret, shouldn't it already include this padding?

hmhealey (commented):
Thanks for the PR, @pablovelezvidal. I just did an initial pass over everything, and it looks good, although I am a bit worried with the number of constants that other people might break it accidentally. That's something we could always improve going forward though.

I have a couple questions about the way the suggestions are positioned as well, but that's  a UX concern since this seems to match the design as far as I can tell.

hmhealey:
@matthewbirtch I have a couple questions about the design from what I'm seeing when testing this out:
1. How should the suggestion box be offset when using different types of autocompletes? The design lines the SuggestionList up with the avatar for at-mention autocompletes, but that doesn't look great with some of the other ones like the emoji autocomplete since it's spaced differently.
    Screen Shot 2020-11-23 at 1 02 39 PM

2. This is also by design, but the hovering autocomplete looks weird when the textbox is longer. I guess this isn't really a change from before though, but I thought I'd mention it
Screen Shot 2020-11-23 at 12 52 36 PM

esethna:
@hmhealey good point on (2), note that this is actually specificed in the spec: 

> If there are multiple lines of text, the autocomplete should float above the line the cursor is on.

So what you're observing is not expected behaviour ^ cc// @pablovelezvidal

matthewbirtch:
> 1\. How should the suggestion box be offset when using different types of autocompletes? The design lines the SuggestionList up with the avatar for at-mention autocompletes, but that doesn't look great with some of the other ones like the emoji autocomplete since it's spaced differently.

@hmhealey The emoji autocomplete should be updated to match the style of the other autocomplete menus (not sure why it didn't get updated). I've submitted a [jira ticket](https://mattermost.atlassian.net/browse/MM-30898) for this.

For the time being, I think it's fine to leave the position as-is until this gets the same spacing/UI update

> 2\. This is also by design, but the hovering autocomplete looks weird when the textbox is longer. I guess this isn't really a change from before though, but I thought I'd mention it

As per the note above from @esethna we should update the positioning for multi-line text scenarios so that the autocomplete does not look disconnected

matthewbirtch:
I can review in more detail once the above is addressed, but I did notice a couple other things:

1. We should align the icons in the autocomplete with the trigger character:
**Designed:**
image

**Implemented:**




2. I was able to get in a scenario where the popover is misaligned with the trigger character (when it can't find a match with what you're typing):
image

3. In the edit post modal:
  a. Rather than pushing the popover under the input box, we should position it directly under and proximal to the trigger character.
  b. Can we limit the width on the popover to 380px so it doesn't go wider than the input? 
image

pablovelezvidal:
ok, thank you guys for the feedback. @hmhealey @matthewbirtch @esethna I have already checked your comments, I am busy right now but I will start addressing/answering them asap.

utils/utils.jsx
```diff
@@ -1066,6 +1066,158 @@ export function getCaretPosition(el) {
     return 0;
 }
 
+export function createHtmlElement(el) {
```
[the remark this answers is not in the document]
pablovelezvidal: That is a practice I have found useful. For this case, in the hypothetical scenario where you want to manipulate in some way the stuff that will be created, if you have these kind of helper functions it saves you time while the code also becomes more self-documented. Again, is a personal practice and if it does not follow the MM standards, I can happily remove it.

utils/utils.jsx
```diff
@@ -1066,6 +1066,158 @@ export function getCaretPosition(el) {
...
+    textArea.selectionEnd = end;
+    textArea.focus();
+    return {
+        x: Math.floor(rect[0].left - textArea.scrollLeft),
+        y: Math.floor(rect[0].top - textArea.scrollTop),
+    };
+}
+
+export function getViewportSize(win) {
+    const w = win || window;
```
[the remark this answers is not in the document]
pablovelezvidal: ok, I will do a deeper analysis of the potential scenarios where we could need the compatMode that do not imply just older browsers compatibilities. I will comment again once I come to a conclusion.

utils/utils.jsx
```diff
@@ -1066,6 +1066,158 @@ export function getCaretPosition(el) {
...
+    return rem * parseFloat(styles.fontSize);
+}
+
+export function getCaretXYCoordinate(textArea) {
+    if (!textArea) {
+        return {x: 0, y: 0};
+    }
+    const start = textArea.selectionStart;
+    const end = textArea.selectionEnd;
+    const copy = copyTextAreaToDiv(textArea);
```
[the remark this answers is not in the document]
pablovelezvidal: while researching for the best solution to get the X,Y coordinates of the caret, I found several posts in different places that pointed at this solution of creating a div copy with the inner text and some styles from the textArea, as the best available solution. The other potential solution was to use Canvas + 2d and get the width, but that one was expensive in terms of performance. I can take a look at how measuring the selection range behaves for this particular situation.

utils/utils.jsx
pablovelezvidal: sure, good catch! thanks!

utils/utils.jsx
pablovelezvidal: ok, that makes perfect sense. I will add a second parameter to getSuggestionBoxAlgn with a default value of 0, and in this scenario, we grab the resulted value from calling getPixelsToSubstract and pass it to getSuggestionBoxAlgn as a second argument.

utils/utils.jsx
pablovelezvidal: good point. I will review why it was not initially grabbing the padding and passing it to the dynamically created div element.

pablovelezvidal:
@esethna @hmhealey about : "If there are multiple lines of text, the autocomplete should float above the line the cursor is on.", sorry, I misunderstood the actual requirement, now with the screenshot example, I got the idea of also align the suggestion box to the y caret coordinate. I will implement that change asap.

utils/utils.jsx
pablovelezvidal: I already reviewed this one, it was a problem in my logic to make the calculations. Awesome catch, thanks!

utils/utils.jsx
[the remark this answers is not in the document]
pablovelezvidal: ok, that validation can safely be removed.

pablovelezvidal:
hey guys, I fixed what was requested in the comments, but I mess with the git history (that usually happens after a long long busy day). I guess the right procedure would be close this PR and open a new one  and migrate all this conversation. The other _dirty_ option is to force push to my repository branch. What would be the right thing to do in this situation?

matthewbirtch:
> I guess the right procedure would be close this PR and open a new one and migrate all this conversation. The other _dirty_ option is to force push to my repository branch. What would be the right thing to do in this situation?

I'll let @hmhealey comment on the procedure here. Once we have a decision on that, tag me for a review again as I think there are a few remaining tweaks

pablovelezvidal:
ok, perfect @matthewbirtch . I already have cleaned the branch and I can do a force push to my branch. I guess that @hmhealey  would like to review it again after I do that.

matthewbirtch:
> ok, perfect @matthewbirtch . I already have cleaned the branch and I can do a force push to my branch. 

Ok, let me know when I can review again

pablovelezvidal:
@matthewbirtch @hmhealey I think that everything that was initially scoped in the JIRA ticket is already implemented. Also, The triggering characters (@,~,:) are aligned with the avatar/icon in the suggestion box. 
Just one thing to notice, the current behavior is that with every stroke the suggestion box moves in the X-axis and keeps aligned with the caret. If that is not the expected behavior from the UX standpoint, I can make it to keep fixed and aligned  at the position where the triggering character started the suggestion-box, it is an small change and the code for that is ready too if needed.

matthewbirtch:
> Just one thing to notice, the current behavior is that with every stroke the suggestion box moves in the X-axis and keeps aligned with the caret. If that is not the expected behavior from the UX standpoint, I can make it to keep fixed and aligned at the position where the triggering character started the suggestion-box, it is an small change and the code for that is ready too if needed.

I would expect the box to stay aligned with the trigger character, but it's aligned with the cursor. This is what I get:
image

Can we update that?

pablovelezvidal:
ok @matthewbirtch, I just pushed the code changes to keep the suggestion-box sticked/aligned to the triggering character.

matthewbirtch:
> I just pushed the code changes to keep the suggestion-box sticked/aligned to the triggering character.

@pablovelezvidal I'm still seeing the below if I enter an `@` character and then add text before it and go back to the `@` character and keep typing the rest of the username. Seems to throw of the alignment in this case.
image

pablovelezvidal:
@matthewbirtch ok, I will take a look? but, is that a valid use case from the users standpoint?

matthewbirtch:
> @matthewbirtch ok, I will take a look? but, is that a valid use case from the users standpoint?

Definitely more of an edge case, but I came across it while testing and I could see it happening. Not a big deal, but wanted to flag it since I found it

pablovelezvidal:
> > @matthewbirtch ok, I will take a look? but, is that a valid use case from the users standpoint?
> 
> Definitely more of an edge case, but I came across it while testing and I could see it happening. Not a big deal, but wanted to flag it since I found it

ok @matthewbirtch , I will continue checking what needs to be done to address that edge case, in the meantime, could you please continue reviewing the other normal use cases for this change?

matthewbirtch (changes requested):
Sure thing A few more minor tweaks I recommend and this should be good to go:

1. In the Right Sidebar, when the suggestion list shows below the trigger character, can we move the autocomplete popover down a touch? It feels a bit tight. Moving it down by 4px more should do the trick.
image

2. In the edit view, can we move the popover down a bit - it feels a touch tight to the line of text it's aligned with. Moving it down by 4px should do the trick.
image

3. In the edit view, when the autocomplete list displays below the last line it is showing below the input entirely and further away from the trigger character. Can we move it up in this case so that it's positioned similarly to what I outlined above?
image

3. In the edit view, on smaller viewports, the autocomplete list displays cut off on the right side. Can we we limit the width of the suggestion list popover so that it doesn't exceed the width the viewport bounds?
image

pablovelezvidal:
ok @matthewbirtch I will let you know as soon as I get those minor changes done. Thanks for the revision.

utils/utils.jsx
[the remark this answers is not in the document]
hmhealey: That's fine to leave them. I was mostly curious since it's not my personal style, but everyone has their own ways of doing stuff 🙂

hmhealey:
> hey guys, I fixed what was requested in the comments, but I mess with the git history (that usually happens after a long long busy day). I guess the right procedure would be close this PR and open a new one and migrate all this conversation. The other dirty option is to force push to my repository branch. What would be the right thing to do in this situation?

Sorry about the delay. I was out for the past week, so I wasn't checking up on this. Force pushing is fine when it's needed. We generally try to avoid it when possible, but something it does make everything easier.

utils/utils.jsx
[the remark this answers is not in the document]
hmhealey: Yeah, I'd imagine that would be more expensive since it's having to set up a different rendering context for that. This solution makes sense to me though.

I wonder if we should look into it for the autosizing textarea in the future since the current method of having an invisible textarea with the same classes causes issues due to CSS getting out of sync occasionally. Something like this would help prevent that.

hmhealey:
I just did another pass through the code, and it looks good to me. I'll do a final review once the UX is settled on this

pablovelezvidal:
ok @matthewbirtch I have made the necessary changes to complete the four points you mentioned.

> I just did another pass through the code, and it looks good to me. I'll do a final review once the UX is settled on this
@hmhealey I just push a fixed snapshot for a test that failed.

esethna (approved):

matthewbirtch (approved):
Thanks @pablovelezvidal! Things are working great. Appreciate the contribution!

pablovelezvidal:
> Thanks @pablovelezvidal! Things are working great. Appreciate the contribution!

My pleasure..., looking forward to send my next contribution! Thank you for the feedback!

hmhealey (approved):
Looks good to me then!

pablovelezvidal:
hey @hmhealey what is left to get this change merged? BTW, I noticed this ticket https://github.com/mattermost/mattermost-server/issues/16514 . won't it be fixed too via this PR?

calebroseland (approved):
This is great! And I agree that this seems to cover https://github.com/mattermost/mattermost-server/issues/16514 as well.

One nonblocking comment is that it might be more appropriate to use/create a specific utility file (e.g. `utils/position_utils`, `utils/box_alignment`, or something), rather than adding to `utils/utils.jsx`, which is a getting to be a kitchen sink of sorts at this point in time 😛.

hmhealey:
@pablovelezvidal We normally require 2 dev reviews and a QA review for each PR, so this is just waiting on the QA review. I'll go make sure that's prioritized.

That other ticket is somewhat of a followup to this. It's mostly to change the padding in the emoji autocomplete to match the ones for at mentions and such.

jgilliam17 (approved):
Thank you @pablovelezvidal 
Tested, looks good to merge.
- Verified autocomplete popover for `@`, `~` and `/` in the post draft area, RHS and edit post modal; popover adjusts and displays based on the location of the tigger characters and sides of the viewport; when multiple lines of text are present, autocomplete popover is shown right above or below the line the cursor is in. Tested on macOS, Windows and Ubuntu.

PR #6774, opened 2020-10-14
[MM-20465] migrate single_image_view to typescript

Tak-Iwamoto:
#### Summary
migrate components/single_image_view to typescript
#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15475
JIRA: https://mattermost.atlassian.net/browse/MM-20465

PR #7085, opened 2020-11-17
[MM-20465][MM-30590] Fix for collapse/expand of image preview

sudheerDev:
#### Summary
  * The toggle function requires id and not post
  * This was most likely changed to fix a ts error

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30590

### DIFF
--- a/components/single_image_view/single_image_view.tsx
+++ b/components/single_image_view/single_image_view.tsx
@@ -5,7 +5,6 @@ import React from 'react';
 
 import {getFilePreviewUrl, getFileUrl} from 'mattermost-redux/utils/file_utils';
 import {FileInfo} from 'mattermost-redux/types/files';
-import {Post} from 'mattermost-redux/types/posts';
 
 import SizeAwareImage from 'components/size_aware_image';
 import {FileTypes} from 'utils/constants';
@@ -18,7 +17,7 @@ import ViewImageModal from 'components/view_image';
 const PREVIEW_IMAGE_MIN_DIMENSION = 50;
 
 type Props = {
-    post: Post;
+    postId: string;
     fileInfo?: FileInfo;
     isRhsOpen: boolean;
     compactDisplay?: boolean;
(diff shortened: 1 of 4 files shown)


components/single_image_view/single_image_view.tsx
```diff
@@ -185,7 +184,7 @@ export default class SingleImageView extends React.PureComponent<Props, State> {
                     show={this.state.showPreviewModal}
                     onModalDismissed={this.showPreviewModal}
                     fileInfos={[fileInfo]}
-                    post={this.props.post}
+                    postId={this.props.postId}
```
sudheerDev, line 187: This was changed here in the PR https://github.com/mattermost/mattermost-webapp/pull/6774 most likely because the `ViewImageModal` expects post and we are passing postId here. 

The connected component takes care of getting the post from postId but ts complains hence adding the prop below

nevyangelova (approved):

components/view_image/view_image.jsx
```diff
@@ -26,7 +26,12 @@ export default class ViewImageModal extends React.PureComponent {
         /**
          * The post the files are attached to
          */
-        post: PropTypes.object.isRequired,
+        postId: PropTypes.string,
```
Willyfrog, line 29: we should probably have custom types for this kind of thing, to prevent people comparing a post.id with a channel.name (or a channel.id) and notice that on compile time

Willyfrog (approved):
Looks good!

jgilliam17 (approved):
Thank you @sudheerDev 
Tested, looks good to merge.
- Verified collapsing/expanding single image preview.

amyblais:
Should be cherry-picked to cloud-branch.

sudheerDev:
cherry-pick cloud

components/view_image/view_image.jsx
sudheerDev: I agree!

PR #7095, opened 2020-11-18
[MM-27260] Added transform to channel collapsing

devinbinnie:
#### Summary
When a category is collapsed, all the channels are rendered in the same location to start since the parent `div` of the channel has a `height: 0`. But since the inner divs can overflow, they all end up overlapping which can show up in some cases.

This PR adds a `transform` to the collapsing style such that the channel truly has no visible height and the channels will unfold gracefully.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27260

### DIFF
--- a/sass/layout/_sidebar-left.scss
+++ b/sass/layout/_sidebar-left.scss
@@ -759,6 +759,8 @@
         list-style-type: none;
         /* border: solid 2px transparent; */
         color: rgba(var(--sidebar-text-rgb), 0.6);
+        transform: scaleY(100%);
+        transform-origin: top;
 
         &:hover {
             .btn-close {
@@ -817,7 +819,7 @@
 
 
     .SidebarChannel.animating {
-        transition-property: height, padding, color;
+        transition-property: height, padding, color, transform;
         transition-duration: 0.15s; /* collapse animation speed */
         transition-timing-function: ease-out;
     }
@@ -827,6 +829,7 @@
     }
 
     .SidebarChannel.collapsed {
+        transform: scaleY(0);
         height: 0 !important;
         padding-top: 0;
         padding-bottom: 0;


hmhealey (approved):

lindalumitchell (approved):

lindalumitchell:
Checked on the test server using desktop app, and it looks great.

devinbinnie:
@lindalumitchell Can you have a quick second look at this PR? We changed the approach a little and I'd like to make sure you're okay with it.

sass/layout/_sidebar-left.scss
```diff
@@ -854,22 +858,27 @@
 
 
     .SidebarChannel.animating {
-        transition-property: height, padding, color;
-        transition-duration: 0.15s; /* collapse animation speed */
-        transition-timing-function: ease-out;
+        transition-property: height, padding, color, opacity, visibility;
+        transition-duration: 0.15s, 0.15s, 0.15s, 0.15s, 0.15s; /* collapse animation speed */
+        transition-timing-function: ease-out, ease-out, ease-out, ease-in, step-start;
```
deanwhillier, line 863: Sorry, I wasn't clear here, this is what I meant.

```suggestion
        transition-timing-function: ease, ease, ease, ease-in, step-start;
```

sass/layout/_sidebar-left.scss
```diff
@@ -854,22 +858,27 @@
...
+
+        &.collapsed {
+            transition-timing-function: ease-out, ease-out, ease-out, ease-out, step-end;
```
deanwhillier, line 866: And here ...

```suggestion
            transition-timing-function: ease, ease, ease, ease-out, step-end;
```

deanwhillier (approved):
2 suggestions because I wasn't clear in my last thought, but otherwise approved. Thanks!

lindalumitchell (approved):
Thanks, yes, I think that's fine.

PR #7101, opened 2020-11-19
If card year is 0 return false for isExpired

nickmisasi:
#### Summary
Not sure how I missed this, but stripe sends the exp_year as 0 when it's not set (rather than null or something) so it would always be a date in the past. If it's 0, just return false.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30104
#### Related Pull Requests
N/A
#### Screenshots
N/A


marianunez (approved):

devinbinnie (approved):

marianunez:
@stevemudie we can test this once merged on the cloud test servers.

PR #7107, opened 2020-11-20
[GH-16162] Migrate 'components/quick_switch_modal' module and associated tests to TypeScript

tweichart:
#### Summary
This PR migrates the components/quick_switch_modal to TypeScript.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16162


components/quick_switch_modal/quick_switch_modal.tsx
```diff
@@ -156,23 +172,23 @@ export default class QuickSwitchModal extends React.PureComponent {
...
         const mode = e.currentTarget.getAttribute('data-mode');
         this.enableChannelProvider();
         this.setState({mode});
         this.focusTextbox();
     }
 
-    handleSuggestionsReceived = (suggestions) => {
-        const loadingPropPresent = suggestions.items.some((item) => item.loading);
+    private handleSuggestionsReceived = (suggestions: ProviderSuggestions): void => {
+        const loadingPropPresent = suggestions.items.some((item: any) => item.loading);
```
nevyangelova, line 184: Can we add a type for the suggestion items?

tweichart: I'm afraid no. items is coming from either SwitchChannelProvider or SwitchTeamProvider, both of their item arrays contain specific data set to that provider. Also both providers aren't migrated to ts, so there's no types yet.

hmhealey: Theoretically, we could use an interface or a parameterized type here to wrap both types of results, but I don't think that's necessary since the providers haven't been migrated like you said. I'm fine to leave as any for now since I imagine we'll have a bunch of followup tickets to clean up anys after this whole migration is eventually done.

components/quick_switch_modal/quick_switch_modal.test.tsx
```diff
@@ -35,7 +35,7 @@ describe('components/QuickSwitchModal', () => {
         it('should do nothing if nothing selected', () => {
             const props = {...baseProps};
 
-            const wrapper = shallow(
+            const wrapper: ShallowWrapper<any, any, QuickSwitchModal> = shallow(
```
hmhealey, line 38: You can actually save some typing here by adding the parameter to `shallow` instead of the `ShallowWrapper`
```suggestion
            const wrapper = shallow<QuickSwitchModal>(
```
The same thing applies for all of these

components/quick_switch_modal/quick_switch_modal.tsx
```diff
@@ -17,33 +17,55 @@ import SuggestionBox from 'components/suggestion/suggestion_box.jsx';
...
-            switchToChannel: PropTypes.func.isRequired,
-        }).isRequired,
-    };
+    /**
+     * Set to show team switcher
+     */
+    showTeamSwitcher: boolean;
+    actions: {
+        joinChannelById: (channelId: string) => Promise<any>;
+        switchToChannel: (channel: Channel) => Promise<SwitchToChannelResult>;
```
hmhealey, line 51: We actually have a type for this already: ActionResult from `mattermost-redux/types/actions`

components/quick_switch_modal/quick_switch_modal.tsx
```diff
@@ -17,33 +17,55 @@ import SuggestionBox from 'components/suggestion/suggestion_box.jsx';
...
+    }
+}
+
+type State = {
+    text: string;
+    mode: string|null;
```
hmhealey, line 57: When would mode ever be null? It seems to only ever be a string here. If we need a default value somewhere though, I'd prefer to use the empty string over null.

components/quick_switch_modal/quick_switch_modal.tsx
```diff
@@ -17,33 +17,55 @@ import SuggestionBox from 'components/suggestion/suggestion_box.jsx';
...
-    static propTypes = {
+type SwitchToChannelResult = {
+    data?: true;
+    error?: true;
+}
+type ProviderSuggestions = {
+    matchedPretext: any,
+    terms: string[],
+    items: any[],
+    component: SwitchChannelProvider|SwitchTeamProvider,
```
hmhealey, line 35: The provider isn't a component itself, but it returns its own component. I'm pretty sure this can just be 
```suggestion
    component: React.ReactNode,
```
so that it accepts and component.

hmhealey (changes requested):
Thanks for the PR. I added some comments below to slightly tighten up the type definitions, but this looks pretty good overall.

components/quick_switch_modal/quick_switch_modal.tsx
tweichart: as far as i can see, the handleOnClick method uses javascript's getAttribute method to get an attribute of an HTML element and sets this state afterwards to that value. In case of the element not being present though (for whatever reasons) this function would return null, hence this state could possibly be null.
source: https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute
so I'd leave it like that if that's ok? or am I seeing this wrong here?

nevyangelova (approved):

components/quick_switch_modal/quick_switch_modal.tsx
hmhealey: Yeah, that's fine. I didn't realize that we were getting something from the DOM there. That looks like part of some unfinished functionality around being able to switch to teams with this modal, so I'm not surprised it's a little weird then.

hmhealey (approved):
Thanks for those changes!

jgilliam17 (approved):
Thank you @tweichart 
Tested, looks good to merge.
- Verified channel switcher works as before.

hmhealey:
Thanks again for this @tweichart! This'll be included in our 5.31 release coming mid-January.

PR #7108, opened 2020-11-20
[MM-30358] - Add account position spec

ghost:
#### Summary
This PR adds the account settings > position spec

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30358


e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
saturninoabril, line 1: Typo error on filename -- should be `account_settings_position_spec.js`.

e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Stage: @prod
+// Group: @account_setting
+import * as TIMEOUTS from '../../../fixtures/timeouts';
```
saturninoabril, line 12: Please remove `@prod`. We'll add it back once we found the tests to be stable for a few runs.

```suggestion
// Group: @account_setting

import * as TIMEOUTS from '../../../fixtures/timeouts';
```

e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
```diff
@@ -0,0 +1,99 @@
...
+            testTeam = team;
+            testUser = user;
+        });
+    });
+
+    it('MM-T2063 Position / Can enter position', () => {
+        cy.apiLogin(testUser);
+
+        // # Visit town square
+        cy.visit(`/${testTeam.name}/channels/town-square`);
```
saturninoabril, line 30: nit: Can be moved to `beforeEach` if all `it` blocks have the same starting point.

e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+        // # Open the hamburger menu
+        cy.findByLabelText('main menu').should('be.visible').click();
+
+        // # Click on Account settings menu item
+        cy.findByText('Account Settings').should('be.visible').click();
+
+        // # Fill-in the position field
+        cy.findByText('Position').should('be.visible').click();
+        cy.get('#position').type('Master hacker');
```
saturninoabril, line 40: nit: Please assign 'Master hacker' to a variable and then reuse it.

e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+        // # Verify that the popover is visible
+        cy.get('#user-profile-popover').should('be.visible').within(() => {
+            // * and contains the entered position
+            cy.findByText('Master hacker').should('be.visible');
+        });
+    });
+
+    it('MM-T2064 Position / 128 characters', () => {
+        const position = 'masterhackermasterhackermasterhackermasterhackermasterhackermasterhackermasterhackermasterhackermasterhackermasterhackermasterha';
```
saturninoabril, line 60: nit: Seems like the actual text is not important. Can this be shortened, like `const position = 'Master Hacker II'.repeat(8)`?

e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+        // # Fill-in the position field with a value of 128 characters
+        cy.findByText('Position').should('be.visible').click();
+        cy.get('#position').type(position);
+        cy.get('#saveSetting').click();
+
+        cy.get('#position').invoke('val').then((val) => {
+            // # Verify that the input value is 128 characters
+            expect(val.length).to.equal(128);
+        });
```
saturninoabril, line 81: I'm getting error here. Should this move before saving?
Screen Shot 2020-11-24 at 7 22 40 PM

e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    beforeEach(() => {
+        cy.apiAdminLogin();
+        cy.apiInitSetup().then(({team, user}) => {
+            testTeam = team;
+            testUser = user;
+        });
+    });
+
+    it('MM-T2063 Position / Can enter position', () => {
```
saturninoabril, line 26:
```suggestion
    it('MM-T2063 Position', () => {
```

e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
```diff
@@ -0,0 +1,99 @@
...
+        cy.get('#position').invoke('val').then((val) => {
+            // # Verify that the position hasn't changed
+            expect(val).to.equal(position);
+        });
+
+        // # Save position
+        cy.get('#saveSetting').click();
+
+        // # Exit the modal
+        cy.get('body').type('{esc}', {force: true});
```
saturninoabril, line 97: I din't see where the ff was verified.
> The text you entered should show up on the profile popover.

saturninoabril (changes requested):
Thanks @nevyangelova! Please see comments. Also, please confirm that both tests are passing on you.

e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
```diff
@@ -0,0 +1,94 @@
...
+        cy.get('#saveSetting').click();
+
+        cy.get('#position').invoke('val').then((val) => {
+            // # Verify that the input value is 128 characters
+            expect(val.length).to.equal(128);
+        });
+
+        // # Try to edit the field with maximum characters
+        cy.findByText('Position').click();
+        cy.wait(TIMEOUTS.ONE_SEC);
```
Willyfrog, line 80: shouldn't we wait for the editing field to be in view instead?

e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+    it('MM-T2063 Position', () => {
+        // # Open the hamburger menu
+        cy.findByLabelText('main menu').should('be.visible').click();
+
+        // # Click on Account settings menu item
+        cy.findByText('Account Settings').should('be.visible').click();
+
+        // # Fill-in the position field
+        cy.findByText('Position').should('be.visible').click();
```
Willyfrog, line 39: both the edit button and the header have an id which might be a better option to get the element.

this text-vs-id also happens in other tests.

e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
```diff
@@ -0,0 +1,94 @@
...
+        cy.get('#position').type(position);
+        cy.get('#saveSetting').click();
```
Willyfrog, line 41: we can check that `#position` is no longer visible to ensure it is saved

e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
```diff
@@ -0,0 +1,94 @@
...
+        cy.get('body').type('{esc}', {force: true});
+
+        // # Post message in the main channel
+        cy.postMessage('hello from master hacker');
+
+        // # Click on the profile image
+        cy.get('.profile-icon > img').as('profileIconForPopover').click();
+
+        // # Verify that the popover is visible
+        cy.get('#user-profile-popover').should('be.visible').within(() => {
```
Willyfrog, line 53: it might be easier to use contains
```suggestion
        cy.contains('#user-profile-popover', 'Master hacker').should('be.visible');
```
although I'm suggesting this from memory, and it might fail

Willyfrog (changes requested):
minor changes, but looking good

e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
nevyangelova: I think we can definitely verify that the field should be visible but the timeout was added because on some runs the opening of the section in the modal was delayed and it was causing flaky fails

e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
nevyangelova: Yes I was confused about how to choose text over id? I guess in this case I used it because of readability, as well as in the context of the modal there is no conflicts with another element with a similar text

Willyfrog (approved):
LGTM

e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
saturninoabril: `id` is more on implementation details but we prefer the use of text whenever possible since it's resembles on how the "real" user "finds" it.

e2e/cypress/integration/account_settings/display/account_settings_positiion_spec.js
saturninoabril: Still failed on me. Could you please check if it's passing on you? Thanks
Screen Shot 2020-12-07 at 7 09 09 PM

e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
```diff
@@ -0,0 +1,93 @@
...
+        cy.get('#position').invoke('val').then((val) => {
+            // # Verify that the position hasn't changed
+            expect(val).to.equal(longPosition);
+        });
+
+        // # Save position
+        cy.get('#saveSetting').click();
+
+        // # Exit the modal
+        cy.get('body').type('{esc}', {force: true});
```
saturninoabril, line 91: Can't see where "After 4. The full position saves". Could you please take a look again? Thanks

nevyangelova: it's in line 88, where I am able to save the setting and before that even checked that it equals the position I specified.

saturninoabril: Ah I see. I thought the verification is to be done at `#positionDesc` but it's actually not clear in the test case so the approach here makes sense too. Thanks for checking again.
Screen Shot 2020-12-09 at 8 37 32 PM

saturninoabril (approved):
Thanks @nevyangelova! All tests passed and LGTM.

PR #7110, opened 2020-11-21
[MM-16139] Migrate 'components/channel_header_dropdown/menu_items/toggle_favorite_channel' module and associated tests to TypeScript

Tzunhei:
#### Summary

Migrate toggle_favorite_channel module and associated tests to Typescript

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/16139


nevyangelova (approved):

components/channel_header_dropdown/menu_items/toggle_favorite_channel/toggle_favorite_channel.test.tsx
```diff
@@ -12,6 +13,21 @@ describe('components/ChannelHeaderDropdown/MenuItem.ToggleFavoriteChannel', () =
     const baseProps = {
         channel: {
             id: 'channel_id',
+            display_name: 'channel_display_name',
```
hmhealey, line 16: For future reference, we also accept if you just force TypeScript to treat this whole object as a Channel like
```typescript
const baseProps = {
    channel: {
        id: 'channel_id',
    } as Channel,
};
```
which could help save some typing if you're wanting to help with this further.

hmhealey (approved):
Thanks for the help with this migration, @Tzunhei! I noted a spot that could be done slightly differently, but both ways of doing that end up working the same.

jgilliam17 (approved):
Thank you @Tzunhei 
Tested, looks good to merge.
- Verified adding to favorites and removing from favorites using channel header toggle - works as expected.
@hmhealey can you please help merge? Thanks 🙂

PR #7120, opened 2020-11-23
[MM-26593] Disable pointer events on LHS scrollbar when scrolling isn't required

devinbinnie:
#### Summary
When hovering over the right-hand-side of the LHS, if you were hovering over a channel you would expect to see a hand-pointer and be able to click and drag the channel. When in a window that was large enough such that vertical scrolling wasn't necessary, your pointer would jump back to a regular cursor briefly.

This issue was caused by `react-scrollbars` taking the pointer-events from the channel and category links, even when it wasn't shown. This PR forces that the scrollbar track does not pick up pointer-events, while the actual scrollbar does, which works with `react-scrollbars`. This fix stops the scrollbar from interfering with the channels or categories in the LHS.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26593


nevyangelova (approved):

jgilliam17:
Thank you @devinbinnie 
There is just one edge case where the scrollbar interferes with the hover, specifically right after collapsing all the categories and removing the need to scroll, the "phantom" scrollbar is visible until the next click, so if you hover right before clicking anywhere else you will get that same jump in hand/pointer cursor.
Considering this is an edge case, not sure if additional effort is justified.
Dec-03-2020 18-07-46

devinbinnie:
> Thank you @devinbinnie
> There is just one edge case where the scrollbar interferes with the hover, specifically right after collapsing all the categories and removing the need to scroll, the "phantom" scrollbar is visible until the next click, so if you hover right before clicking anywhere else you will get that same jump in hand/pointer cursor.
> Considering this is an edge case, not sure if additional effort is justified.
> Dec-03-2020 18-07-46

@jgilliam17 I can't actually reproduce this myself on the test server. Either way, I think it's fine. We could address this in a follow-up ticket if need be.

jgilliam17 (approved):
Thanks for checking @devinbinnie 
Looks good to merge.
- Verified mouse pointer when hovering over the full length of category and channel names when scrolling is not necessary.

I agree - edge case mentioned in previous comment is fine for now.

calebroseland (approved):

PR #7123, opened 2020-11-23
[MM-30159] Move multiselected channel to other categories via the 3-dot menu

devinbinnie:
#### Summary
Enables the 3-dot menu on channels in the new sidebar to move multiselected groups of channels

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30159


matthewbirtch:
@devinbinnie for the context menus when multiple channels are selected, we'll need to do the following:

1. Pluralize any instance of the word "Channel" > "Channels". (e.g. "Mute Channels", "Leave Channels")
2. Remove 'Copy Link', 'Add members'
3. If any of the selected items is a DM, don't show "Channels" in the `Move to` menu
4. If any of the selected items is a channel, don't show "Direct Messages" in the 'Move to' menu

Also, I noticed that if I select items across the 'Channels' and 'Direct Messages' categories and then try to move them all to 'Favorites', it doesn't move them all. It seems to exclude the DMs.

It also appears that if I select both a private and public channel and then choose 'Leave Channel', nothing happens. No modal and items are not removed from the sidebar

devinbinnie:
@matthewbirtch

> 1. Pluralize any instance of the word "Channel" > "Channels". (e.g. "Mute Channels", "Leave Channels")
> 2. Remove 'Copy Link', 'Add members'

The two items above are being handled in a separate ticket where we add support for other operations on multiple channels:
https://mattermost.atlassian.net/browse/MM-30160

> 3. If any of the selected items is a DM, don't show "Channels" in the `Move to` menu
> 4. If any of the selected items is a channel, don't show "Direct Messages" in the 'Move to' menu

Ah, this is different than the dragging behaviour. Will make the change.

> Also, I noticed that if I select items across the 'Channels' and 'Direct Messages' categories and then try to move them all to 'Favorites', it doesn't move them all. It seems to exclude the DMs.

This also came from the dragging behaviour, i'll have to change that for the menus as well.

> It also appears that if I select both a private and public channel and then choose 'Leave Channel', nothing happens. No modal and items are not removed from the sidebar

Moving multiple channels isn't supported, see above ticket. I'll check and see if moving a single channel works.

matthewbirtch:
> Ah, this is different than the dragging behaviour. Do we want these to behave differently?
> 
> >

@devinbinnie I see your point. In the spec for dragging we have:



But in the spec for the menu behaviour, we have:

Screen Shot 2020-11-24 at 8 11 20 PM

So, this definitely presents an inconsistency. I agree, we should be consistent between the two, but the question to me is which is the better of the two? As I play with this more, it feels odd to only move part of the selection. I'm wondering if we should not allow movement to the channels or DM categories at all if it is a mixed selection of DMs & channels. @esethna what's your opinion here? It might be good to get a few other team member's thoughts.

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -209,16 +211,15 @@ export class SidebarChannelMenu extends React.PureComponent<Props, State> {
...
-            case Constants.GM_CHANNEL:
+            const selectedChannels = multiSelectedChannelIds.indexOf(channel.id) === -1 ? [channel] : displayedChannels.filter((c) => multiSelectedChannelIds.indexOf(c.id) !== -1);
+
+            if (selectedChannels.every((selectedChannel) => selectedChannel.type === Constants.DM_CHANNEL || selectedChannel.type === Constants.GM_CHANNEL)) {
                 return category.type !== CategoryTypes.CHANNELS;
-            default:
-                return true;
+            } else if (selectedChannels.every((selectedChannel) => selectedChannel.type !== Constants.DM_CHANNEL && selectedChannel.type !== Constants.GM_CHANNEL)) {
+                return category.type !== CategoryTypes.DIRECT_MESSAGES;
             }
```
Willyfrog, line 220: Sorry, i'm trying to grasp this, but seems I'm probably missing some context.

`multiSelectedChannelIds.indexOf(channel.id) === -1` is this trying to figure out if it is a multiselection or not?

I'm also not sure about this section of code, with many categories and channels selected this might slowdown. Is it going through every category and from there all the selected channels and then every channel selected in the category?

why do we check if it is a dm/gm selection only or a regular channel selection only after filtering by categories? wouldn't that give some errors? also if we can get that info before going throught the filter, we don't have to recalculate this for the whole selection twice. fixing this would probable not cause the problem i mentioned above, but as I said I might be missing something in this context.

Also it might be easier to read if we change the comparison from negatives to positives and maybe assigning that to a named constant, e.g.:

```js
selectedChannels.every((selectedChannel) => selectedChannel.type !== Constants.DM_CHANNEL && selectedChannel.type !== Constants.GM_CHANNEL)
```

vs

```js
const regularChannelsOnly = selectedChannels.every((selectedChannel) => selectedChannel.type === Constants.OPEN_CHANNEL || selectedChannel.type === Constants.PRIVATE_CHANNEL)
```
but this last part might be personal preference :)

Willyfrog (changes requested):
some concerns on one of the sections, but that might be me not knowing that section properly.

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
devinbinnie: So the filtering here is done to only show categories in the menu that you can move the selected channels to.
The algorithm should be O(n^2) as it goes through all selectedChannels for each category to determine whether those selected channels can be moved to the category or not.

The only way I can think to speed this up would be to store this information in redux, but I'm not sure how much it would really speed it up. Might reduce the number of times it's recalculated if anything.

Thoughts?

esethna:
@matthewbirtch 2/5, but my thoughts come down to fast, obvious, forgiving. If a user has a lot of channels selected or various types, then not allowing them to move any of them isn't forgiving or obvious. The user has to A) understand why they can't move them B) find the channels that causing the issue C) unselect them, before trying the action again. If you allow them to move the channels and keep the ones that didn't get moved selected (as done now), it seems faster, more obvious and more forgiving. I agree though that we should be consistent between the menus and the drag actions, so my vote would be to show all categories in the move to menu.

matthewbirtch:
> not allowing them to move any of them isn't forgiving

Excellent point. Thanks for the sound rationale @esethna. Thank you for the input. @devinbinnie I agree with Eric that we should proceed with the consistent approach of showing all the categories in the 'move to' menu. I will update the spec doc to reflect this.

So, I guess that means it's good as-is.

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
Willyfrog: as discussed in community, probably moving the "isGM/DM" and "isRegularChannel" checks outside of the loop should improve everything

devinbinnie: @Willyfrog Ready for re-review!

matthewbirtch (approved):

Willyfrog (approved):
LGTM!

nevyangelova (approved):
LGTM!

jgilliam17:
Thank you @devinbinnie 
Looks good 
few questions:
- After selecting few channels, should a click on any other channel, without holding shift or cmd/ctrl, act as an ESC and dismiss the selection? It's not right now, not sure if it's intended. 
- When selecting Leave channel on the multi-select menu should user leave all selected channels? Right now, user only leaves the one channel that is open on the main channel view, not the other selected channels.

devinbinnie:
> * After selecting few channels, should a click on any other channel, without holding shift or cmd/ctrl, act as an ESC and dismiss the selection? It's not right now, not sure if it's intended.

Yes it should, good catch. I'll make a fix.

devinbinnie:
> > * After selecting few channels, should a click on any other channel, without holding shift or cmd/ctrl, act as an ESC and dismiss the selection? It's not right now, not sure if it's intended.
> 
> Yes it should, good catch. I'll make a fix.

@jgilliam17 Fixed!

jgilliam17:
@devin thanks for the quick fix, clicking on the channel without holding ctrl/cmd/shift deselects the channels now
Trying to understand the line from the spec file:

`3.  When using multi-select (ctrl or shift clicking on channels), the first selected channel/DM/GM is what we display in the center channel.`

a or b?
a) First channel that user clicks on while holding shift or ctrl/cmd, is the first channel selected for the move, regardless of which channel was open in the center channel prior to the cmd/ctrl click. Channel should open in the center channel as it's selected
b) The channel that is currently open in the center channel automatically becomes first selected channel for the move 

What I'm seeing on the test server is b) and it feels a bit off

devinbinnie:
> @devin thanks for the quick fix, clicking on the channel without holding ctrl/cmd/shift deselects the channels now
> Trying to understand the line from the spec file:
> 
> `3. When using multi-select (ctrl or shift clicking on channels), the first selected channel/DM/GM is what we display in the center channel.`
> 
> a or b?
> a) First channel that user clicks on while holding shift or ctrl/cmd, is the first channel selected for the move, regardless of which channel was open in the center channel prior to the cmd/ctrl click. Channel should open in the center channel as it's selected
> b) The channel that is currently open in the center channel automatically becomes first selected channel for the move
> 
> What I'm seeing on the test server is b) and it feels a bit off

It's b), and yes we went back and forth a lot around whether this should be the case.
@esethna Do you have any other thoughts about this?

jgilliam17:
> feels a bit off

@esethna 
I feel I need to clarify my comment a bit.
- Using shift+click to select channels to move - Makes sense to include the current channel as it is one of the end points of the selected range
- Using cmd+click - this is the part that felt odd,  I expected only channels that I clicked on while holding cmd/ctrl to be selected, not current channel (no cmd/ctrl+click)

Workaround exists: user can click on the current channel while holding cmd/ctrl and deselect it, it's fairly easy to discover after moving few groups.

esethna:
I'll let @matthewbirtch comment as well, but I think this is as expected. By taking the action to cmd/ctrl+click you are implying a multiselect (file browsers work the same way), rather than a normal click that implies changing channels

jgilliam17:
Thanks for fixing Move to New Category @devinbinnie 
One small issue came up, channels moved to newly created category remain selected. Differs from drag and drop and when selecting an existing category to move the group.

devinbinnie:
> Thanks for fixing Move to New Category @devinbinnie
> One small issue came up, channels moved to newly created category remain selected. Differs from drag and drop and when selecting an existing category to move the group.

@jgilliam17 Done.

jgilliam17 (approved):
Thank you @devinbinnie 
Tested, looks good to merge.
- Verified multi-selected channels/DMs/GMs can be moved to an existing or new category via 3-dot menu. DMs/GMs remain selected(not moved) if user attempts to move to Channels category, and channels remain selected(not moved) when user tries to move to DM category.  Selection clears after move. Selection can be cleared by clicking esc or anywhere on the sidebar while not holding shift or cmd/ctrl. Checked on various browsers and desktop app on macOS, Windows and Ubuntu.

PR #7127, opened 2020-11-24
Add Billing Permissions

hahmadia:
#### Summary

Adds Billing Permissions for Billing section of admin definition

#### Ticket Link

mattermost.atlassian.net/browse/MM-30950

#### Related Pull Requests

- Has server changes https://github.com/mattermost/mattermost-server/pull/16364


components/admin_console/system_roles/system_role/system_role_permissions.tsx
```diff
@@ -34,6 +34,11 @@ const sectionsList: SystemSection[] = [
         hasDescription: true,
         subsections: [],
     },
+    {
+        name: 'billing',
+        hasDescription: true,
+        subsections: [],
+    },
```
marianunez, line 41: We should only add this section if they are on a cloud installation which can be done by checking the `Cloud` feature in the license

marianunez (changes requested):
Thanks again for helping out on this @hahmadia!

nickmisasi (approved):
+1 on what Maria mentioned - otherwise, LGTM

components/admin_console/system_roles/system_role/system_role_permissions.tsx
hahmadia: Done!

hahmadia:
@marianunez I changed the billing text and also don't include billing in the system roles if not licensed for cloud. You can try it on the server repo since the web app repo doesn't pick up changes properly.

sbishel (approved):

components/admin_console/system_roles/system_role/system_role_permissions.tsx
```diff
@@ -154,6 +161,11 @@ export default class SystemRolePermissions extends React.PureComponent<Props, St
             };
         }
 
+        if (!isLicensedForCloud) {
+            // Remove the billing section if it's not licensed for cloud
+            sectionsList.splice(1, 1);
```
marianunez, line 166: Could we filter here based on section name? It would be more readable and robust than relying on position in the case in the future someone adds another section before `billing` and alters the array positions.

hahmadia: G000000D call on that :) Totally forgot about how in the future, the section may not have the same index. I added a findIndex for it.

hahmadia: THANKYOU

marianunez (approved):
Thanks @hahmadia!

furqanmlk (approved):
Tested and verified with Read-Only / User Manager / System Manager 
Working as expected

PR #7133, opened 2020-11-25
[MM-30969] Drop aria-label from Marketplace item description

hanzei:
#### Summary
A plugin may not have a description. This was incorrectly stated in the propTypes.

Given that a `aria-label` for a `a` and `span` element that just contains text is not needed, I did remove it instead of creating an `aria-label` with different content.

The PR also adds a test for this case.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30969


larkox (approved):
LGTM

Willyfrog (commented):
the aria label is expected to help people that uses the screen reader, so I'm not sure removing it is a good idea, maybe we should think about some different value for it.

I agree we shouldn't expect the description to exist, what I'm not sure is if we should use it for the screen reader to have something to read or instead provide information about the link?

hanzei:
@Willyfrog Does a screen reader pick up `a` and `span` tags automatically? Or are `aria-label`'s need for all of these tags? If that is the case, was is usually provided there besides the plain text inside of the element?

Willyfrog:
> @Willyfrog Does a screen reader pick up `a` and `span` tags automatically? Or are `aria-label`'s need for all of these tags? If that is the case, was is usually provided there besides the plain text inside of the element?

Talking especifically for this scenario, if there isn't a description, I don't understand the need of an empty `span` tag, we should probably remove this completely.

For the `a` tag, it is also using the description to link to the home page. but I'd say it would be more useful to have some sort of "Plugin's website" description at least.

Since both contain text, it is not needed to have the aria label, but if that text is missing as well, because it is no longer required, we might want to look into what we are presenting to users.

hanzei:
@Willyfrog `description` _might_ be empty for some plugin. This PR adjusts the code to take into account that it might be missing. 

Should be change the `aria-label` of the link to just "Plugin's website"?

Willyfrog:
> @Willyfrog `description` _might_ be empty for some plugin. This PR adjusts the code to take into account that it might be missing.
> 
> Should be change the `aria-label` of the link to just "Plugin's website"?

not necesarily the aria-label, changing the text inside would be fine as well, as the screen reader will pick it up. I think it would only be an issue if we leave it empty as it might read the link, which feels a lot less useful for someone who has to rely on screen readers.

hanzei:
> not necesarily the aria-label, changing the text inside would be fine as well, as the screen reader will pick it up. 

Sorry, I might not get your suggestion. `description` and `pluginDetailsInner` already contain meaningful text. Are you suggesting to add the text "Plugin's website" to the UI? Given that a major redesign of the Marketplace is in the works now, that seems like a too big change if we only want to fix the linked bug.

Willyfrog:
> > not necesarily the aria-label, changing the text inside would be fine as well, as the screen reader will pick it up.
> 
> Sorry, I might not get your suggestion. `description` and `pluginDetailsInner` already contain meaningful text. Are you suggesting to add the text "Plugin's website" to the UI? Given that a major redesign of the Marketplace is in the works now, that seems like a too big change if we only want to fix the linked bug.

yes, if there is no description text I was suggesting to use that as default, if it's too much hassle, just add it to the aria-label so it doesn't get in the way of the redesign.

hanzei:
@Willyfrog Added your suggestion in 1683b01cf0e3b7d81b08a812f2835710285d962c. PTAL.

Willyfrog (approved):
LGTM

hanzei:
@prapti Could you please review this PR? You can find the QA steps [in the ticket](https://mattermost.atlassian.net/browse/MM-30969).

prapti:
@hanzei Yeah, I can definitely review this PR. The bigbluebutton plugin is not compiling (neither master nor the commit mentioned in the ticket). Is there a different plugin I can try with? I can review this first thing on Monday.

hanzei:
@prapti Could you please just pick any plugin then and remove the `description` from `plugin.json` and deploy it?

prapti (approved):
Tested on local server with a couple of different local plugins by removing their description (removing the description element and also using just an empty string). Launching marketplace does not crash the webapp.
Looks good to me.

amyblais:
I forgot that this was already in the last cloud release. No need to cherry-pick.

PR #7140, opened 2020-12-01
MM-30992 Add the capability to run part by part for Cypress parallel testing (part 1)

saturninoabril:
#### Summary
Add the capability to run part by part for Cypress parallel testing.

Each part can execute by:
- `node run_tests.js --stage='@prod' --part=1 --of=2` 
- `node run_tests.js --stage='@prod' --part=2 --of=2`

See https://community-release.mattermost.com/core/pl/u96tyu4e9i8xir8gy17ckxgqge for generated test report which will include field for "Test Execution" -- link to Zephyr recorded test execution.

#### Ticket Link
Jira ticket: https://mattermost.atlassian.net/browse/MM-30992

#### Screenshots
Typical job, like in CircleCI:
- Each E2E test part or job will run until complete. Then, each test artifacts will persist to workspace for next job processing which is consolidating and publishing all reports. The last job determines whether E2E failed or succeed.
Screen Shot 2020-12-01 at 8 42 56 PM

### DIFF
--- a/e2e/run_tests.js
+++ b/e2e/run_tests.js
@@ -171,6 +171,10 @@ function getTestFilesIdentifier(numberOfTestFiles) {
     const {part, of} = argv;
     const PART = parseInt(part, 10) || 1;
     const OF = parseInt(of, 10) || 1;
+    if (PART > OF) {
+        throw new Error(`"--part=${PART}" should not be greater than "--of=${OF}"`);
+    }
+
     const multiplier = Math.ceil(numberOfTestFiles / OF);
     const start = (PART - 1) * multiplier;
     let end = start + multiplier;
--- a/e2e/run_tests.js
+++ b/e2e/run_tests.js
@@ -169,8 +169,10 @@ function printMessage(testFiles, overallIndex, currentIndex, lastIndex) {
 
 function getTestFilesIdentifier(numberOfTestFiles) {
     const {part, of} = argv;
-    const multiplier = Math.round(numberOfTestFiles / of);
-    const start = (part - 1) * multiplier;
+    const PART = parseInt(part, 10) || 1;
+    const OF = parseInt(of, 10) || 1;
+    const multiplier = Math.ceil(numberOfTestFiles / OF);
+    const start = (PART - 1) * multiplier;
     let end = start + multiplier;
     end = end < numberOfTestFiles ? end : numberOfTestFiles;
 
--- a/e2e/run_tests.js
+++ b/e2e/run_tests.js
@@ -23,11 +23,12 @@
  *      Selected files are those not matching any of the specified stage or group.
  *
  * Environment:
- *   BROWSER=[browser]      : Chrome by default. Set to run test on other browser such as chrome, edge, electron and firefox.
- *                            The environment should have the specified browser to successfully run.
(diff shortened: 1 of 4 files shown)


e2e/run_tests.js
```diff
@@ -161,7 +161,22 @@ function printMessage(testFiles, index) {
 
     // Log which files were being tested
     console.log(chalk.magenta.bold(`${invert ? 'All Except --> ' : ''}${testStage}${stage && withGroup ? '| ' : ''}${testGroup}`));
-    console.log(chalk.magenta(`(Testing ${index + 1} of ${testFiles.length})  - `, testFile));
+    console.log(chalk.magenta(`(Testing ${overallIndex + 1} of ${testFiles.length})  - `, testFile));
+    if (process.env.CI_BASE_URL) {
+        console.log(chalk.magenta(`Testing ${currentIndex}/${lastIndex} in "${process.env.CI_BASE_URL}" server`));
+    }
```
saturninoabril, line 167: This will generate log, like
Screen Shot 2020-12-01 at 8 55 05 PM

e2e/run_tests.js
```diff
@@ -161,7 +161,22 @@ function printMessage(testFiles, index) {
...
+}
+
+function getTestFilesIdentifier(numberOfTestFiles) {
+    const {part, of} = argv;
+    const multiplier = Math.round(numberOfTestFiles / of);
```
josephbaylon, line 172: I think this needs to be a ceiling,
numberOfTestFiles=10
part=3
of=3
multiplier = Math.ceil(10/3) = 4
start = (3 - 1) * 4 = 8
end = 8 + 4 = 12
end = 12 < 10 ? 12 : 10 = 10

Otherwise if it's round,
numberOfTestFiles=10
part=3
of=3
multiplier = Math.round(10/3) = 3
start = (3 - 1) * 3 = 6
end = 6 + 3 = 9
end = 9 < 10 ? 9 : 10 = 9

will be missing 1 file?

```suggestion
    const multiplier = Math.ceil(numberOfTestFiles / of);
```

josephbaylon (approved):
LGTM except for a possible issue I mentioned regarding round and ceiling.

e2e/run_tests.js
saturninoabril: Nice catch Joseph! You're correct, will update this. 
Also, I missed to add the default value of `1` to `part` and `of` so that running the script without passing such args (ex. `node run_tests.js`) will still work as expected.

saturninoabril: Done, PTAL. Thanks!

prapti (approved):
Looking great!

prapti:
Just a thought: there is no absolute need that checks for the valid number of `part` and `of` but maybe for future improvement we could consider checking whether the files exists, for example, if I run `node run_tests.js --stage='@prod' --group='@enterprise' --part=10 --of=5` but `--group='@enterprise'` only has 45 tests, the current test run won't throw any error.

saturninoabril:
Thanks @prapti! I agree. I'll improve it next time and also add some validation and CLI options/help. For the meantime, I added conditional check that will throw an error whenever value of `--part` is greater than `--of`.

PR #7151, opened 2020-12-02
[MM-30449] - Hide create category option when unread filter is enabled

nevyangelova:
#### Summary
Hide the option to create new category when unread filter is enabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30449


devinbinnie (approved):

jgilliam17:
Thanks @nevyangelova - I see the option to create categories when unread filter is enabled on the test server. Can you please check? Thanks

nevyangelova:
I am a little confused with this test case then. As far as I know the unread filter is clicked and then you only display the unread channels. At that point you shouldn't create categories. If the filter is present which seems to be always the case, shouldn't the user have the option at this point? Or is there a way to not display it at all? From what I am seeing in the code it's not the case.

components/sidebar/sidebar_category/sidebar_category_menu/index.ts
```diff
@@ -9,16 +9,22 @@ import {getCurrentTeam} from 'mattermost-redux/selectors/entities/teams';
 import {ActionFunc} from 'mattermost-redux/types/actions';
 import {GlobalState} from 'mattermost-redux/types/store';
 
+import {GlobalState as WebappGlobalState} from 'types/store';
```
hmhealey, line 12: Instead of using `as`, you can just replace `GlobalState` with this everywhere in this file

jgilliam17 (approved):
Thank you @nevyangelova 
Tested, looks good to merge.
- Verified option to create new category is not present on the main `+` menu when unread filter is enabled.

hmhealey (approved):

PR #7170, opened 2020-12-04
[MM-31045] Fix Error when adding group on channel configuration screen

hahmadia:
#### Summary

Fix a small bug in typescript migration that was causing lots of issues

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31045

**Note to QA**
Ensure that when you click edit on a channel, the channel loads fine and that all the functions on that page work (do not throw a error or white screen of death)


components/admin_console/team_channel_settings/channel/details/channel_groups.tsx
```diff
@@ -26,7 +26,7 @@ interface ChannelGroupsProps {
     isDisabled?: boolean;
 }
 
-export const ChannelGroups: React.SFC<ChannelGroupsProps> = (props: ChannelGroupsProps): JSX.Element => {
+export const ChannelGroups: React.FunctionComponent<ChannelGroupsProps> = (props: ChannelGroupsProps): JSX.Element => {
```
hahmadia, line 29: This was a un-related deprecated function fix. (SFC is deprecated and need to use FunctionComponent)

hahmadia:
Couldn't really think of a way to add a test as there was already a big console error saying that `Channel ID is marked as required but received undefined`

mkraft (approved):

sbishel (approved):

furqanmlk (changes requested):
I am getting `NaN Members` instead of `0 Member`

image

furqanmlk (approved):
Could not reproduce this issue mentioned above.
During Cypress test case, group was deleted using API while UI is visible, so it is not normal behaviour. 
All test cases passed

PR #7172, opened 2020-12-06
15642 - Updating Add to channel modal

asaadmahmood:
#### Summary
15642 - Updating Add to channel modal

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/15642

#### Screenshots
Screenshot 2020-12-06 at 6 40 19 PM
Screenshot 2020-12-06 at 6 40 26 PM


components/channel_invite_modal/channel_invite_modal.tsx
```diff
@@ -250,14 +281,14 @@ export default class ChannelInviteModal<T extends Value> extends React.PureCompo
...
-            <FormattedMessage
-                id='multiselect.numPeopleRemaining'
-                defaultMessage='Use ↑↓ to browse, ↵ to select. You can add {num, number} more {num, plural, one {person} other {people}}. '
-                values={{
-                    num: MAX_SELECTABLE_VALUES - this.state.values.length,
-                }}
-            />
+        const header = (
+            <h1>
+                <FormattedMessage
```
jespino, line 286: This is not a good change, this is going to break how you can "translate the string", in english, clearly you said "Add people to Whatever", but maybe in another language you would say "XXXX ChannelName XXXX" so... you can't assume the order of the words in an translations.

asaadmahmood: @jespino That's the way it was.
I just changed New Members to people.

Previously it was:
"channel_invite.addNewMembers": "Add New Members to ",

I made it:
 "channel_invite.addNewMembers": "Add people to ",

Nothing else.

nevyangelova: Instead of using the <span> you can use the 'values' functionality of React Intl, and that would solve potential issues.

components/multiselect/multiselect.tsx
```diff
@@ -240,6 +247,33 @@ export default class MultiSelect<T extends Value> extends React.PureComponent<Pr
         this.props.handleDelete(values);
     }
 
+    MultiValueRemove = ({children, innerProps}: any) => (
+        <div {...innerProps}>
+            {children || <CloseCircleSolidIcon className='h-4 w-4 d-flex mr-1'/>}
```
nevyangelova, line 252: what is the reasoning for these classNames? We wanted to migrate webapp styles to using BEM. More semantic classes would help with clarity.

components/multiselect/multiselect.tsx
```diff
@@ -240,6 +247,33 @@ export default class MultiSelect<T extends Value> extends React.PureComponent<Pr
...
+        const profileImg = imageURLForUser(user.id, user.last_picture_update);
+
+        return (
+            <React.Fragment>
+                <Avatar
+                    size='sm'
+                    username={user.username}
+                    url={profileImg}
+                />
+                <div style={{margin: '0 2px 0 8px'}}>
```
nevyangelova, line 266: Let's not use inline styles as we are getting a lot of conflicts because of the existing ones.

components/multiselect/multiselect.tsx
```diff
@@ -297,6 +331,68 @@ export default class MultiSelect<T extends Value> extends React.PureComponent<Pr
             );
         }
 
+        let reactSelectDiv;
+        if (this.props.valueWithImage) {
```
nevyangelova, line 335: Can we build a props object and then use ternary statements for specific properties? I know it's more if's in practice but it will be more readable.

components/multiselect/multiselect.tsx
```diff
@@ -345,6 +441,43 @@ export default class MultiSelect<T extends Value> extends React.PureComponent<Pr
             optionsToDisplay = options;
         }
 
+        let multiSelectList;
+        if (this.props.saveButtonPosition === 'bottom') {
```
nevyangelova, line 445: same as above

components/multiselect/multiselect.tsx
```diff
@@ -360,76 +493,55 @@ export default class MultiSelect<T extends Value> extends React.PureComponent<Pr
...
-                        {memberCount}
+            <React.Fragment>
+                <div className='filtered-user-list'>
+                    <div className='filter-row filter-row--full'>
+                        <div className='multi-select__container react-select'>
+                            {reactSelectDiv}
+                            {this.props.saveButtonPosition === 'top' &&
+                            <SaveButton
+                                id='saveItems'
+                                extraClasses='ml-4'
```
nevyangelova, line 504: is it possible to use classNames for the extra classes instead of having this prop?

components/multiselect/multiselect.tsx
```diff
@@ -360,76 +493,55 @@ export default class MultiSelect<T extends Value> extends React.PureComponent<Pr
...
-                    loading={this.props.loading}
-                    selectedItemRef={this.props.selectedItemRef}
-                />
-                <div className='filter-controls'>
-                    {previousButton}
-                    {nextButton}
-                </div>
-            </div>
+                {this.props.saveButtonPosition === 'bottom' &&
+                <div className='text-right pr-6 pb-6'>
```
nevyangelova, line 534: nit: more semantic classes

components/multiselect/multiselect.tsx
```diff
@@ -446,7 +558,10 @@ const nullComponent = () => null;
 const paddedComponent = (WrappedComponent: any) => {
     return (props: {data: any}) => {
         return (
-            <div style={{paddingLeft: '10px'}}>
+            <div
+                className='d-flex align-items-center'
+                style={{padding: '4px 2px 4px 4px'}}
```
nevyangelova, line 563: nit: no inline styles

components/multiselect/multiselect.tsx
```diff
@@ -446,7 +558,10 @@ const nullComponent = () => null;
 const paddedComponent = (WrappedComponent: any) => {
     return (props: {data: any}) => {
         return (
-            <div style={{paddingLeft: '10px'}}>
+            <div
+                className='d-flex align-items-center'
```
nevyangelova, line 562: nit: more semantic classes

components/widgets/inputs/users_invite_input.jsx
```diff
@@ -0,0 +1,331 @@
...
+import LoadingSpinner from 'components/widgets/loading/loading_spinner';
+import Avatar from 'components/widgets/users/avatar';
+import {imageURLForUser, getDisplayName, getLongDisplayNameParts} from 'utils/utils.jsx';
+
+import {t} from 'utils/i18n.jsx';
+import {isGuest} from 'utils/utils';
+
+import './users_emails_input.scss';
+
+export default class UsersEmailsInput extends React.PureComponent {
```
nevyangelova, line 28: Since we are migrating to typescript, all new files in webapp should be tsx.

components/widgets/inputs/users_invite_input.jsx
```diff
@@ -0,0 +1,331 @@
...
+        const profileImg = imageURLForUser(user.id, user.last_picture_update);
+        let guestBadge = null;
+        let botBadge = null;
+
+        if (user.is_bot) {
+            botBadge = <BotBadge/>;
+        }
+
+        if (!isEmail(user.value) && isGuest(user)) {
+            guestBadge = <GuestBadge/>;
```
nevyangelova, line 113: can we do this inline? I know this is an existent practice in mattermost but we have discussed this to be only for complex component changes.

components/widgets/inputs/users_invite_input.jsx
```diff
@@ -0,0 +1,331 @@
...
+                {getDisplayName(user)}
+                {botBadge}
+                {guestBadge}
+            </React.Fragment>
+        );
+    }
+
+    onChange = (value) => {
+        if (this.props.onChange) {
+            this.props.onChange(value.map((v) => {
```
nevyangelova, line 159: nit: semantic value naming.

components/widgets/inputs/users_invite_input.jsx
```diff
@@ -0,0 +1,331 @@
...
+
+    components = {
+        NoOptionsMessage: this.NoOptionsMessage,
+        MultiValueRemove: this.MultiValueRemove,
+        IndicatorsContainer: () => null,
+    };
+
+    handleInputChange = (inputValue, action) => {
+        if (action.action === 'input-blur' && inputValue !== '') {
+            const values = this.props.value.map((v) => {
```
nevyangelova, line 219: nit: more semantic variable name

sass/components/_channel-switcher.scss
```diff
@@ -56,6 +56,120 @@
...
+        }
+
+        .multi-select__wrapper {
+            padding: 12px 0;
+            height: auto;
+            overflow: auto;
+            position: absolute;
+            width: calc(100% - 48px);
+            margin: -28px 0 0 24px;
+            background: v(center-channel-bg);
```
nevyangelova, line 88: can we use the rgba variable here too?

sass/components/_channel-switcher.scss
```diff
@@ -56,6 +56,120 @@
...
+
+            &.react-select__value-container--has-value {
+                padding: 0 8px;
+            }
+        }
+
+        .more-modal__list {
+            height: auto;
+
+            > div {
```
nevyangelova, line 127: let's keep this style semantic as it's a problem within webapp we have constantly crashing styles because of direct element styling.

sass/components/_react-select.scss
```diff
@@ -62,14 +54,17 @@
...
-
             .react-select__indicator {
                 opacity: .8;
             }
         }
     }
 
+    .react-select__control--is-focused {
+        box-shadow: inset 0 0 0 1px v(button-bg);
+        border-color: v(button-bg);
```
nevyangelova, line 65: can we use the rgba variables too? it's part of the migration to compass design system.

sass/utils/_modifiers.scss
```diff
@@ -57,3 +57,7 @@ $positions: 't'top,
     padding: 0;
     border: 0;
 }
+
+.overflow--visible {
+    overflow: visible;
+}
```
nevyangelova, line 63: nit: new line

nevyangelova (changes requested):

esethna:
Thanks @asaadmahmood, minor nit, I think we decided on the design pattern that the autocomplete would overlap the input fields slightly? See this ticket for example: https://mattermost.atlassian.net/browse/MM-30898

Looks like this may have also broken for the regular autocomplete as I believe it used to work that way? Would you be open to submitted a fix?
image

asaadmahmood:
@esethna The design that we have in the Figma file does not have it overlapping the input.
I think the overlapping input was only for the post area input, and I think it would be better if we tackle that in a separate ticket.

Also can you point out what is broken because of this? I don't think we have the overlapping thing implemented yet.

image

esethna:
@asaadmahmood it was implemented in the past as part of: https://mattermost.atlassian.net/browse/MM-22956 (but it is no longer overlapping on community or daily it seems)

I think we should be consistent here too, but I'll let @michaelgamble comment on that.

components/multiselect/multiselect.tsx
asaadmahmood: @nevyangelova These are just to avoid me creating another rule or add extra css class for the icon.
I can just add css to it, but when we just want to give basic css rules, we can just reuse the css we already have instead of adding classes for every simple thing right?

components/channel_invite_modal/channel_invite_modal.tsx
asaadmahmood: Did that.

nevyangelova (approved):

matthewbirtch:
> @asaadmahmood it was implemented in the past as part of: https://mattermost.atlassian.net/browse/MM-22956 (but it is no longer overlapping on community or daily it seems)
> 
> I think we should be consistent here too, but I'll let @michaelgamble comment on that.

This PR contribution implements the autocomplete positioning update (not merged yet): https://github.com/mattermost/mattermost-webapp/pull/7082#issuecomment-740828480

asaadmahmood:
@matthewbirtch So from what I see, we're not overlapping it vertically right?

matthewbirtch:
To be honest, I kind of forgot about the input overlapping. That PR is more about the horizontal positioning.

I'd love to see it overlap the message input field, but we can do that as a later enhancement

asaadmahmood:
Then do we overlap the suggestion list over the input in this ticket too, or keep it at the edge like the way it is?

matthewbirtch:
> Then do we overlap the suggestion list over the input in this ticket too, or keep it at the edge like the way it is?

I don't think we can be consistent everywhere. For example, the search input uses an autocomplete popover, but it won't look right if the popover overlaps the input since it's a smaller input.

So, I say we leave it for now as-is (not overlapping) since it's not overlapping anywhere yet. We can look at whether we actually want to overlap the post draft input later in a separate PR

asaadmahmood:
I agree. I don't think we can have them overlap everywhere.
@esethna So we will keep it as is for now, and the input overlapping can be done as a separate ticket as well as Matt pointed out.

esethna:
@matthewbirtch @asaadmahmood works for me, can we open a follow-up ticket to look into that. Agree it may not be a global pattern in places like search, but with the larger input fields it could certainly make sense (ie post input, edit post, this modal etc)

nevyangelova (commented):
Still some comments are unaddressed

asaadmahmood:
@nevyangelova Which ones? Can you comment them again?

components/multiselect/multiselect.tsx
nevyangelova: @devinbinnie can you advice here?

components/multiselect/multiselect.tsx
nevyangelova: this comment ^

components/multiselect/multiselect.tsx
nevyangelova: ^ this one too

components/multiselect/multiselect.tsx
nevyangelova: here

components/multiselect/multiselect.tsx
nevyangelova: here

components/multiselect/multiselect.tsx
devinbinnie: Yeah I'd agree that we do this here, it will definitely be more readable.

asaadmahmood: @devinbinnie Is there a way to do this?
Not to give a prop at all conditionally?

It's not giving the `react-select` prop a different value based on a condition, rather its to either give a few props to it, or not at all.

components/multiselect/multiselect.tsx
asaadmahmood: Are we sure we want to do that, we would then have to create css rules for every simple thing we want to do if we want to be consistent I guess. So if I have to just give it a margin-left of 4px, I would have to create a rule for that in every separate instance instead of reusing a helper class right?

components/multiselect/multiselect.tsx
asaadmahmood: Not sure how we do this.
Do I add a className prop to the component itself and then use classNames in it?

Furthermore, we are also moving the saveButton itself in the DOM based on the prop value, rather than just changing the class.

asaadmahmood: Though I'm just removing the class altogether, and giving it through css.

components/multiselect/multiselect.tsx
devinbinnie: Yes, so instead of having 2 components that give different sets of props, you would specify each prop as such:
```
prop={booleanValue ? thisProp: thatProp}
```

asaadmahmood:
@devinbinnie @nevyangelova Can we approve this PR, or talk about how we do the ternary properties for React Select?
This PR has been waiting for that quite a while.

esethna (approved):

esethna:
@michaelgamble for a UX review

esethna:
@asaadmahmood are we adding in the statuses to this PR as per conversation in https://github.com/mattermost/mattermost-webapp/pull/7194

We should match the style of the other autocomplete statuses I would think

asaadmahmood:
@esethna These are done from my end, I'm waiting for @michaelgamble to review it.
I plan on adding the statuses in another PR, to avoid multiple features in a single PR. It allows us to every back then if there are issues without reverting the whole thing.

michaelgamble (approved):
Looks good to me :)

components/multiselect/multiselect.tsx
```diff
@@ -297,6 +331,10 @@ export default class MultiSelect<T extends Value> extends React.PureComponent<Pr
             );
         }
 
+        // const reactSelectProps = {
```
devinbinnie, line 334: Can we remove this comment?

devinbinnie (changes requested):

devinbinnie (approved):

ogi-m (commented):
Hey @asaadmahmood, a few issues/questions:
1. The autocomplete is always active instead of only when there's text input, so the "Add'' button is always overlapped and I can't add users to a channel.
2. The text color for selected user's names is hard to read on dark themes:
add_people

3. What's the expected text when there are no results? 
It says _No items found_, while the Invite Members To Team modal shows _No one found matching *entered text*_

asaadmahmood:
1, 2, will fix that.
3. @michaelgamble Any thoughts on the text?

asaadmahmood:
@ogi-m Fixed 1 and 2.
If those are fine, we can approve the PR.
I can update the text for 3 later when I get a confirmation from some of the designers.

josephbaylon (approved):
LGTM. Affected e2e tests were updated and passing locally. There's a minor text bug,

```
"multiselect.list.notFound": "No results found matching **{searchQuery}**"
```

displays as 
```
No results found matching ****
```

ogi-m:
Retested issues 1 and 2 and they're fixed 👍

ogi-m (approved):

metanerd:
Hi, this PR failed in master with a failing snapshot. Do we just have to update the snapshot or could there be something wrong?

asaadmahmood:
@saturninoabril Thoughts?

saturninoabril:
I think it just need to update the snapshot. It's a merge conflict from the other PR - https://github.com/mattermost/mattermost-webapp/pull/7338 - which added the new snapshot on the said failing component. PR submitted - https://github.com/mattermost/mattermost-webapp/pull/7387

PR #7173, opened 2020-12-07
[Cypress/E2E, MM-27222] Add test cases, fix makeClient, add/correct types

calebroseland:
#### Summary
- **`cy.makeClient`**:
  - Fix (errors when executed in tests via `cypress open`)
  - Add typings for intellisense/autocompletion
  - Add caching to reduce login/request thrashing
- Correct assorted types in some `support/api/*.d.ts` files
- **[MM-27222](https://mattermost.atlassian.net/browse/MM-27222)**, E2E test cases:
  - `integration/bot_accounts/post_message_spec.js`:
    - MM-T1812: add
  - `integration/bot_accounts/display_name_spec.js`:
    - MM-T1813: add
  - `integration/bot_accounts/in_teams_and_channels_spec.js`:
    - MM-T1814: propose dropping;  is the same as and handled (with more scrutiny) by MM-T1815
    - MM-T1815: add
    - MM-T1816: add
    - MM-T1817: add
    - MM-T1818: add
  - `integration/bot_accounts/promote_demote_spec.js`
    - MM-T1819: add

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27222

#### Screenshots
CleanShot 2020-12-07 at 02 13 36@2x

CleanShot 2020-12-07 at 01 44 40@2x


hmhealey (commented):
Interesting. I didn't know we could use async/await with Cypress. I thought the pseudo-promises they used wouldn't work with that.

Do the existing tests I added using `makeClient` need to be updated because the return type for it changed?

calebroseland:
> Interesting. I didn't know we could use async/await with Cypress. I thought the pseudo-promises they used wouldn't work with that.
> 

Indeed, awaiting a `cy` sequence/chain would not work and should not be done. Nor should the test body be an async function. But the other way around is true to an extent, as cypress will yield to/resolve promises in the same way it handles `cy.wrap(promise).then()`, and because of the general interoperability between async/await and promises, it should work when used appropriately.

> Do the existing tests I added using `makeClient` need to be updated because the return type for it changed?

The return type shouldn't have changed and I tested it out in `category_muting_spec`, so it should be ok.

josephbaylon (changes requested):
Thanks @calebroseland !! I had some intermittent failures locally. So just minor suggestions to stabilize the tests. Also, please fix conflicting files. Thanks!

Note: I ran other tests that use `makeClient` and most pass except for `channel/archived_channels_spec` > `MM-T1687` (fails on some sidebar issue in this PR), which when I tried to run against master already fails due to a `promise` issue. We don't need to fix that in this PR as that's outside the scope but if you want to take a look, that would be awesome.

calebroseland:
@josephbaylon review changes committed 👍

Re: `makeClient` It's currently broken on `master` when I try to run either of the two tests that use it. I ended up using it a lot in these new tests, so I thought it would be appropriate to fix it here.

e2e/cypress/integration/bot_accounts/promote_demote_spec.js
```diff
@@ -0,0 +1,60 @@
...
+            // # Add bot to team
+            const bot = await client.createBot(createBotPatch());
+            await client.addToTeam(team.id, bot.user_id);
+
+            // # Open member invite screen
+            cy.get('#sidebarHeaderDropdownButton').should('be.visible').click();
+            cy.get('#manageMembers > button').should('contain.text', 'Manage Members').click();
+
+            // # Find bot
+            cy.get('#searchUsersInput').type(`${bot.username}`);
```
josephbaylon, line 45: Sometimes the filter input displays before the modal list and typing gets cut-off. Just needed to make sure modal list is loaded before search is performed.

```suggestion
            cy.get('.more-modal__list').find('.more-modal__row').its('length').should('be.gt', 0);
            cy.get('#searchUsersInput').type(`${bot.username}`);
```

josephbaylon (changes requested):
Thanks @calebroseland for the update. One last change request below

josephbaylon (approved):
LGTM. Tests passed locally. Thanks @calebroseland !!

hmhealey (approved):

PR #7175, opened 2020-12-07
MM-18107: Collapsed Reply Threads, Global Threads

calebroseland:
See also: #7744, as a predecessor to this. (Diff here: https://github.com/mattermost/mattermost-webapp/compare/crt-precursor-supporting...feature/collapsed-reply-threads)

#### Summary

- [MM-30302](https://mattermost.atlassian.net/browse/MM-30302): Global Threads screen
~~- [MM-32019](https://mattermost.atlassian.net/browse/MM-32019): Feature flag detection~~ now in [webapp's redux](https://github.com/mattermost/mattermost-webapp/blob/feature/collapsed-reply-threads/packages/mattermost-redux/src/selectors/entities/preferences.ts#L244-L277)
- [MM-30303](https://mattermost.atlassian.net/browse/MM-30303): System Console setting
- [MM-30305](https://mattermost.atlassian.net/browse/MM-30305): Account Settings toggle
- IDs removed from `RhsThread`/`ThreadViewer`: 
    | Original | E2E Equivalent | Used in styling selectors? |
    | - | - | - | 
    | `#rhsContent` | `#rhsContainer .post-right__content` | No 
    |`#rhsPostList` | `#rhsContainer .post-right-comments-container` | No
    | `#rhsFooter` | `#rhsContainer .post-create` | No

Most of this feature is safely behind the flag `MM_FEATUREFLAGS_COLLAPSEDTHREADS=true`.
 
However, some necessary component abstractions/refactoring affect codepaths not behind the flag. Most notably:
- `<ThreadViewer/>` (the guts of `<RhsThread/>`) needs to be tested extensively (non-feat. UX flows)
- ~~`<RHSSearchNav/>` (search bar, at-mentions, saved posts, help menu)~~ (this was extracted to #7744)

The bulk of the UI is in [components/threading](https://github.com/mattermost/mattermost-webapp/tree/feature/collapsed-reply-threads/components/threading)
```
components/threading
├── channel_threads/thread_footer
├── common
│   ├── button
│   ├── follow_button
├── global_threads
│   ├── thread_item
│   ├── thread_list
│   ├── thread_menu
│   ├── thread_pane
│   ├── global_threads (main screen where above comes together)
├── global_threads_link
├── thread_viewer
```

Prior versions of some of these components have been already been reviewed, but there have been some changes since. (https://github.com/mattermost/mattermost-webapp/pull/6585)

#### Screenshots
CleanShot 2021-02-09 at 13 07 24@2x
CleanShot 2021-02-09 at 13 06 55@2x
CleanShot 2021-02-09 at 13 07 06@2x


matthewbirtch (changes requested):
Awesome work on this @calebroseland a few things I found in my first round of testing. I'm unsure what pieces of the puzzle aren't implemented yet, so maybe these issues are non-issues when it's all merged together.

1. The mention badge on the 'Threads' sidebar item does not go away when I go to the global threads view

3. Mentions in threads still appear on the channel item in the sidebar. In fact, there seems to be some odd behavior with mention badges in general. Sometimes the badges don't show on the Threads sidebar item until I've clicked it. 

4. The mention badge in the global threads inbox view also does not go away once I view the thread.

5. The number of replies shown on a thread in the global threads inbox view is not updating in real-time

6. The time stamp on the global threads inbox items don't seem to match the most recent reply timestamp

7. Unfollowing a thread doesn't seem to remove it from global threads

Overall, it seems like there are a number of things that are getting delayed and not updating in real-time. 

**A few other small UI issues:**

8. Can we update the alignment on the time stamp on consecutive posts so that it's right-aligned instead of left-aligned?
image

9. I got to a state where 2 unread threads display, but the 'No unread threads' empty state also showed. Seems wrong
image

10. There appears to be an issue with timestamps displaying in compact mode in the global threads view
image

11. The illustrations for the empty states aren't using theme colors. I believe there is a way that @nickmisasi used with his work on cloud to ensure the svgs are using theme colors.
image

calebroseland:
Note on demo server: it's not currently reflecting lates mattermost-server changes for some reason. You may need to pull locally. Latest version of server and redux have necessary infra.

calebroseland:
Thanks for checking things out, @matthewbirtch! (cc @reflog Matt's comments above)

I believe 2–5 probably fall under websocket events, and these should have improved a lot recently, though I think Eli's still working on a few things.

On 1, is it expected that the mentions count to go away before viewing any threads? (Cleared just by going to global threads?)

6: fixed (disappears after deselection—we should probably review those selection/unread filter UX states)

On the UI items:
9: fixed
11: fixed
8, 10: still working on

matthewbirtch:
> On 1, is it expected that the mentions count to go away before viewing any threads? (Cleared just by going to global threads?)

Sorry, no. I think we should clear a mention once you've actually read the thread

lieut-data (changes requested):
@calebroseland, I'm slightly worried about merging a feature branch of this size directly into `master`. If anything goes wrong, we'll be in the awkward position of having to revert the whole thing or play whack-a-mole around our release process.

Is it feasible to break this up into changes to existing components and the core, new functionality? (I realize the size of the ask here...)

stylianosrigas:
Removing cloud test server for maintenance reasons. Please feel free to add the label again if you need it.

zefhemel:
To confirm here also what I agreed with @ogi-m: before merging this code (in whatever shape, probably split into smaller ones — but same will apply) the QA focus is on testing this doesn't regress anything with the feature flag **disabled**. Ideally there's also testing for this with the flag enabled, but it's essential there's no regression of the system with it disabled.

matthewbirtch:
@calebroseland I just happened to be playing with the threads PR and there were a few things I just noticed that I thought I should call out. Not sure if these are known issues:

1. the 'No followed threads yet' block is not horizontally centered when the RHS is open.
image

2. If I have one thread in my Global Threads view, and then unfollow it, I get an infinite 'Loading' state
image

3. If I mark an unfollowed thread as unread from the channel view, it doesn't refollow that thread. It should as per the [spec](https://mattermost.atlassian.net/wiki/spaces/Threads/pages/1000669195/Desktop+Webapp+User+Experience#Following%2FUnfollowing-Threads)

components/channel_header_mobile/channel_header_mobile.js
```diff
@@ -90,29 +90,36 @@ export default class ChannelHeaderMobile extends React.PureComponent {
...
+                            <div className='navbar-brand'>
+                                <MobileChannelHeaderDropdown/>
+                                {isMuted && (
+                                    <UnmuteChannelButton
+                                        user={user}
+                                        channel={channel}
+                                    />
+                                )}
+                            </div>
+                        ) : (
```
calebroseland, line 103: test

calebroseland, line 103: test

calebroseland:
> 1. the 'No followed threads yet' block is not horizontally centered when the RHS is open.
I think this should be deferred under future responsiveness improvements (collapsing list view, etc.)

> 2.  If I have one thread in my Global Threads view, and then unfollow it, I get an infinite 'Loading' state
Fixed

> 3. If I mark an unfollowed thread as unread from the channel view, it doesn't refollow that thread. It should as per the spec
I think this should be done as a smaller piece of the in-channel experience story.

lieut-data (changes requested):
Blocking until smaller PR pulling out some of these changes is merged.

angeloskyratzakos:
@calebroseland There were some panics with that cloud workspace from `plugin/hclog_adapter.go:54`

lieut-data:
@calebroseland, if you wanted to merge `crt-precursor-supporting` into this branch (maybe you have already?) we could get started on the review while QA Review completes on the other?

Willyfrog:
I'm still doing the review, one thing I noticed is that if the mention is in a reply it doesn't appear in the channel, is this expected?

Screen Shot 2021-03-25 at 19 35 28

components/sidebar_right/sidebar_right.jsx
```diff
@@ -151,10 +151,11 @@ export default class SidebarRight extends React.PureComponent {
     render() {
         const {
             rhsChannel,
-            currentUserId,
             isFlaggedPosts,
             isPinnedPosts,
             postRightVisible,
+
+            // postRightSameAsSelectedThread,
```
lieut-data, line 158: Unused?

components/sidebar_right/sidebar_right.jsx
```diff
@@ -163,6 +164,10 @@ export default class SidebarRight extends React.PureComponent {
             isExpanded,
         } = this.props;
 
+        /* if (postRightSameAsSelectedThread) {
```
lieut-data, line 167: Unused?

components/threading/channel_threads/thread_footer/thread_footer.tsx
```diff
@@ -90,18 +91,19 @@ function ThreadFooter({
 
             <FollowButton
                 isFollowing={isFollowing}
-                follow={follow}
-                unFollow={unFollow}
+                className='separated'
+                onClick={useCallback(() => {
```
lieut-data, line 95: I haven't seen this pattern in hooks before -- declaring one inline. Wondering if it makes more sense to lift it to the top of the method instead for clarity, especially if the rendering logic one days unexpectedly changes to `return;` early?

components/threading/common/button/button.tsx
```diff
@@ -11,6 +11,7 @@ type Props = {
     append?: ReactNode;
     isActive?: boolean;
     hasDot?: boolean;
+    clamp?: boolean;
```
lieut-data, line 14: It's not super obvious what `clamp` means in the context of a button -- is there a better name we could use? Or do we need a comment here?

components/threading/common/options.ts
```diff
@@ -1,12 +1,18 @@
...
 
-export const THREADING_TIME: TimestampProps['units'] = [
-    'now',
-    'minute',
-    'hour',
-    'day',
-    'week',
-];
+export const THREADING_TIME: Partial<ComponentProps<typeof Timestamp>> = {
+
```
lieut-data, line 8: nit: extra line?

components/threading/global_threads/index.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export {default} from './global_threads';
```
lieut-data, line 4: Same comment as earlier: do we need this `index.ts` at all?

components/timestamp/timestamp.tsx
```diff
@@ -278,7 +284,8 @@ class Timestamp extends PureComponent<Props, State> {
             useRelative = (): ResolvedFormats['relative'] => {
                 const {
                     display,
-                    updateIntervalInSeconds,
+                    updateIntervalInSeconds = this.props.updateIntervalInSeconds,
```
lieut-data, line 287: I'm not familiar with the `autoRange` pattern here, but can you elaborate on why this line is changing?

components/timestamp/timestamp.tsx
```diff
@@ -343,6 +351,14 @@ class Timestamp extends PureComponent<Props, State> {
         }
     }
 
+    static getDerivedStateFromProps(props: Props, state: State) {
+        if (props.value !== state.prevValue) {
+            return ({now: new Date(), prevValue: props.value});
```
lieut-data, line 356: Can you elaborate on the need to touch `now` here? Does this have side effect on other uses of this component?

components/user_profile/index.ts
```diff
@@ -13,14 +13,11 @@ type OwnProps = {
...
-function makeMapStateToProps() {
+function makeMapStateToProps(state: GlobalState, ownProps: OwnProps) {
     const getDisplayName = makeGetDisplayName();
-
-    return (state: GlobalState, ownProps: OwnProps) => {
-        return {
-            displayName: getDisplayName(state, ownProps.userId, true),
-            user: getUser(state, ownProps.userId),
-        };
+    return {
```
lieut-data, line 18: Removing the callback return means `makeGetDisplayName` is created from scratch every time, breaking the intended memoization here, and we should just use the underlying, unmemoized selector instead. Or was this change intentional?

lieut-data (changes requested):
I've gotten through about 20% of the review, but figured I'd push up my comments now to help surface feedback early.

Also, I realized that by only merging the broken out PR back into this one, all the "deltas" are still present and hard to ignore using GitHub's review interface. Propose we continue with developer review here once the other PR is merged and master gets pulled into this branch, narrowing the scope of the files in question considerably.

calebroseland:
@Willyfrog, there's a successor PR that @reflog is working on to make changes to the mention badging, so that is okay for now. There's a spec I'll send you a link to that describes it in more detail.

components/threading/common/options.ts
calebroseland: slip of the keyboard

components/user_settings/display/index.ts
```diff
@@ -54,6 +54,8 @@ function mapStateToProps(state: GlobalState) {
         channelDisplayMode: get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.CHANNEL_DISPLAY_MODE, Preferences.CHANNEL_DISPLAY_MODE_DEFAULT),
         messageDisplay: get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.MESSAGE_DISPLAY, Preferences.MESSAGE_DISPLAY_DEFAULT),
         collapseDisplay: get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.COLLAPSE_DISPLAY, Preferences.COLLAPSE_DISPLAY_DEFAULT),
+        collapsedReplyThreadsAllowUserPreference: isCollapsedThreadsAllowed(state) && getConfig(state).CollapsedThreads as string !== 'always_on',
```
Willyfrog, line 57: nit: 'always_on' as a constant

components/widgets/header/header.tsx
```diff
@@ -0,0 +1,41 @@
...
+// See LICENSE.txt for license information.
+
+import React, {memo, ReactNode, Fragment, HTMLAttributes} from 'react';
+import classNames from 'classnames';
+
+import './header.scss';
+
+type Props = {
+    heading: ReactNode;
+    level?: 0 | 1 | 2 | 3 | 4 | 5 | 6;
```
Willyfrog, line 11: this should be a  named type, I'm guessing this could be reused in future components

stores/hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param name name of stored value, prepended to suffix
+ * @param suffix to provide scope; defaults to user and team
+ */
+export function useGlobalState<TVal>(
+    initialValue: TVal,
+    name: string,
+    suffix: string = useSelector(currentUserAndTeamSuffix),
+): [TVal, (value: TVal) => ReturnType<typeof setGlobalItem>] {
+    const dispatch = useDispatch();
+    const storedKey = name + suffix;
```
Willyfrog, line 43: nit
```suggestion
    const storedKey = `${name}${suffix}`;
```

Willyfrog (changes requested):
the new feature looks awesome! I'm really looking forward to have that in daily :D

I've added a few comments, but it looks great in general.

You might also want to look into views/channel_sidebar.ts (I think that's the place, but might be mistaken) to change the mouse pointer to a prohibited sign when dragging a channel on top of the globalThreadsLink component

another thing i noticed, on a long thread (not fitting the screen) if I press enter I get a different scroll than if I start writing (the add comment button doesn't show in the first case)

also no drag'n drop file, but I guess this might be for the future?

actions/websocket_actions.jsx
```diff
@@ -1359,3 +1375,46 @@ export function handleUserActivationStatusChange() {
 function handleCloudPaymentStatusUpdated() {
     return (doDispatch) => doDispatch(getCloudSubscription());
 }
+
+function handleThreadReadChanged(msg) {
+    return (doDispatch, doGetState) => {
+        if (msg.data.thread_id) {
+            const thread = doGetState().entities.threads.threads?.[msg.data.thread_id];
```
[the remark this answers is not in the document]
calebroseland: yes, it's feature specific. I could see us using  `userThreads.threads`, but it already went into the redux code in prior PRs. The feature flag will still be off and ongoing changes will be made, so I think this should be deferred for another time/follow up PR.

actions/websocket_actions.jsx
```diff
@@ -1359,3 +1375,46 @@ export function handleUserActivationStatusChange() {
...
+    };
+}
+
+function handleThreadUpdated(msg) {
+    return (doDispatch, doGetState) => {
+        try {
+            const threadData = JSON.parse(msg.data.thread);
+            handleThreadArrived(doDispatch, doGetState, threadData, msg.broadcast.team_id);
+        } catch {
+            // invalid JSON
```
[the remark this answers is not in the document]
calebroseland: Probably. This was a part of initial unreads/mentions. @reflog thoughts? is this catch still needed or can we remove it?

components/admin_console/admin_definition.jsx
```diff
@@ -5609,6 +5609,46 @@ const AdminDefinition = {
...
+                                value: 'disabled',
+                                display_name: t('admin.experimental.collapsedThreads.off'),
+                                display_name_default: 'Disabled',
+                            },
+
+                            /* {
+                                value: 'off',
+                                display_name: t('admin.experimental.collapsedThreads.off'),
+                                display_name_default: 'Disabled',
+                            }, */
```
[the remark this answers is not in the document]
calebroseland: `off` was originally in the spec, but ended up as `disabled`, so I think we can remove. The `always_on` is for future use.

components/widgets/header/header.tsx
calebroseland: I'm a huge fan of reusability, but I think `level` is fairly specific to `Headings` below, though it's still accessible via `React.ComponentProps<typeof Header>['level']` if need be, or I could export the props type?

components/threading/global_threads/global_threads.tsx
```diff
@@ -0,0 +1,189 @@
...
+                    )}
+                </div>
+            ) : (
+                <>
+                    <ThreadList
+                        currentFilter={filter}
+                        setFilter={setFilter}
+                        someUnread={Boolean(numUnread)}
+                    >
+                        {(filter === 'unread' ? unreadThreadIds : threadIds).map((id) => (
```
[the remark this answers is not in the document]
calebroseland: `filter` is backed by `ThreadFilter`, so it's typed and all. I feel it's simpler to use string literal unions for smaller/local things like this.

components/threading/global_threads/global_threads.tsx
```diff
@@ -0,0 +1,189 @@
...
+                    id: 'globalThreads.subtitle',
+                    defaultMessage: 'Threads you’re participating in will automatically show here',
+                })}
+                right={<RHSSearchNav/>}
+            />
+            {isEmpty(threadIds) ? (
+                <div className='no-results__holder'>
+                    {isLoading ? (
+                        <LoadingScreen/>
+                    ) : (
```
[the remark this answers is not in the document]
calebroseland: I do agree that large render trees can be a bit difficult to read, and I think this is around the limit of what should be done in a single component—and quite possibly it should be broken down into smaller chunks.

But even with trees like this one, I like the consistency in knowing that _this is the tree_—its internal forks are all here and I know that there aren't any separate fragments that I would need to go hunting for. If it should be a separate chunk/fragment, perhaps it should be its own component.

I did want to maintain a relatively flat hierarchy (e.g. header is generic, with the page-specific content here), but I'm completely open to changes—especially for readability—@Willyfrog @lieut-data what do you guys think?

components/user_settings/display/index.ts
calebroseland: CollapsedThreads is also backed by explicit string literals, but we could possibly change it to an enum. I'll go ahead and do that, but maybe I'll queue up the topic for the next guild meeting? (Constants, enums, string literal unions, etc.)

components/sidebar_right/sidebar_right.jsx
```diff
@@ -163,6 +164,10 @@ export default class SidebarRight extends React.PureComponent {
...
+            return null;
+        } */
```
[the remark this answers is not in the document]
calebroseland: Past workarounds, will remove

components/markdown/index.ts
```diff
@@ -44,6 +47,11 @@ function makeMapStateToProps() {
     return function mapStateToProps(state: GlobalState, ownProps: Props) {
         const config = getConfig(state);
 
+        let channelId;
+        if (ownProps.postId) {
+            channelId = getPost(state, ownProps.postId)?.channel_id;
+        }
```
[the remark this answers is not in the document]
calebroseland: That was my first thought, but I don't think we'd want `getPost` firing for falsy postId values.

components/threading/common/follow_button/follow_button.tsx
```diff
@@ -1,40 +1,37 @@
...
+import classNames from 'classnames';
 
 import Button from '../button';
+import {t} from 'utils/i18n';
 
 type Props = {
-    isFollowing: boolean;
-    follow: MouseEventHandler<HTMLButtonElement>;
-    unFollow: MouseEventHandler<HTMLButtonElement>;
+    isFollowing: boolean | null | undefined;
```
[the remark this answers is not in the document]
calebroseland: You're right that we don't technically need `undefined`, I'll go ahead and remove that. But when null (while thread data is loading), the follow button is disabled to prevent errant changes/toggles.

components/threading/global_threads/thread_menu/thread_menu.tsx
```diff
@@ -1,145 +1,138 @@
...
+                <Menu.ItemAction
+                    text={formatMessage(hasUnreads ? {
+                        id: t('threading.threadMenu.markRead'),
+                        defaultMessage: 'Mark as read',
+                    } : {
+                        id: t('threading.threadMenu.markUnread'),
+                        defaultMessage: 'Mark as unread',
+                    })}
+                    onClick={useCallback(() => {
+                        dispatch(updateThreadRead(currentUserId, currentTeamId, threadId, hasUnreads ? Date.now() : unreadTimestamp));
```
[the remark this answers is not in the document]
calebroseland: `updateThreadRead` is a single action to handle both marking "as read" and "as unread." `now` corresponds to "as read" and `unreadTimestamp` (og post time) for "as unread."

components/channel_header_mobile/channel_header_mobile.js
```diff
@@ -90,29 +90,36 @@ export default class ChannelHeaderMobile extends React.PureComponent {
...
-                                <ShowSearchButton/>
-                                <MobileChannelHeaderPlug
-                                    channel={channel}
-                                    isDropdown={false}
-                                />
-                            </React.Fragment>
+                            <MobileChannelHeaderPlug
+                                channel={channel}
+                                isDropdown={false}
+                            />
```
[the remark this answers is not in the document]
calebroseland: I'm using inline conditionals here to avoid two very similar trees with would-be duplicated content. It would be more straight forward if the items weren't interwoven, and we should probably improve this, but I was trying to reduce changes to core layout when possible.

components/favicon_title_handler/favicon_title_handler.tsx
```diff
@@ -83,21 +84,32 @@ class FaviconTitleHandler extends React.PureComponent<Props> {
...
-            document.title = mentionTitle + unreadTitle + currentChannelName + ' - ' + currentTeam.display_name + ' ' + currentSiteName;
+            document.title = `${mentionTitle + unreadTitle + currentChannelName} - ${currentTeam.display_name} ${currentSiteName}`;
+        } else if (currentTeam && isGlobalThreadsView) {
+            document.title = formatMessage({
+                id: 'globalThreads.title',
+                defaultMessage: '{prefix} Threads - {displayName} {siteName}',
+            }, {
+                prefix: mentionTitle + unreadTitle,
+                displayName: currentTeam.display_name,
+                siteName: currentSiteName,
```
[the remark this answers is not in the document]
calebroseland: Up on :91 okay?

components/favicon_title_handler/index.ts
```diff
@@ -24,6 +28,7 @@ function mapStateToProps(state: GlobalState) {
         currentTeammate,
         siteName: config.SiteName,
         unreads: getUnreads(state),
+        isGlobalThreadsView: match?.path === '/:team' && location?.pathname.includes(`${match?.url}/threads`),
```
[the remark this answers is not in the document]
calebroseland: We're outside of that route path, so match.path will only ever get to /:team here, but yes, there's a better way to check, I'll switch it

components/user_profile/index.ts
calebroseland: Unintentional. I think I forgot to undo some experimentation I was doing with `useSelector`+`createSelector` and the make-selector pattern re: `makeGetDisplayName`. I'll revert

components/timestamp/timestamp.tsx
calebroseland: Most of the time `value` is older than `now` e.g. a fixed timestamp with an ongoing `now` that updates on an interval.

But when a new `value` comes down the pipe, this mechanism ensures that `now` isn't stale, preventing incorrect diffs.

I think I could and probably should move this to a separate PR.

components/threading/common/button/button.tsx
calebroseland: Maybe `clampText` and a comment?

components/admin_console/admin_definition.jsx
```diff
@@ -5609,6 +5609,46 @@ const AdminDefinition = {
...
+                        help_text_default: 'When enabled (default on), this setting enables collapsed reply threads for all users on this server. Users can disable the feature in Account Settings > Display > Collapsed Reply Threads (Beta). When enabled (default off), users must enable collapsed reply threads in Account Settings. When disabled, users cannot access collapsed reply threads. [Learn more about collapsed reply threads](!https://about.mattermost.com/default-threads).',
+                        help_text_markdown: true,
+                        options: [
+                            {
+                                value: 'disabled',
+                                display_name: t('admin.experimental.collapsedThreads.off'),
+                                display_name_default: 'Disabled',
+                            },
+
+                            /* {
```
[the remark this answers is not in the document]
calebroseland: Was, but we can remove `off`, `always_on` is for future use.

actions/websocket_actions.jsx
```diff
@@ -1359,3 +1375,46 @@ export function handleUserActivationStatusChange() {
...
+        }
+    };
+}
+
+function handleThreadFollowChanged(msg) {
+    return (doDispatch) => {
+        handleFollowChanged(doDispatch, msg.data.thread_id, msg.broadcast.team_id, msg.data.state);
```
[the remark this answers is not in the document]
calebroseland: The `handle*` methods are more or less just action creators with an occasional thunk signature—there's no reason we can't remove the thunk and dispatch-passing logic. `actions/threads.ts` developed in part based on proximity and similar/copied actions.

actions/websocket_actions.jsx
```diff
@@ -1359,3 +1375,46 @@ export function handleUserActivationStatusChange() {
...
+                        lastViewedAt: msg.data.timestamp,
+                        prevUnreadMentions: thread.unread_mentions,
+                        newUnreadMentions: msg.data.unread_mentions,
+                        prevUnreadReplies: thread.unread_replies,
+                        newUnreadReplies: msg.data.unread_replies,
+                    },
+                );
+            }
+        } else {
+            handleAllMarkedRead(doDispatch, msg.broadcast.team_id);
```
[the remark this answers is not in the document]
calebroseland: @reflog would that be easily doable?

actions/websocket_actions.jsx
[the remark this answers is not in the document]
calebroseland: I agree that should be possible, but I'll double check to make sure there isn't a specific reason it had to be outside of/before the action.

actions/team_actions.jsx
```diff
@@ -105,3 +110,12 @@ export function updateTeamsOrderForUser(teamIds) {
         dispatch(savePreferences(currentUserId, teamOrderPreferences));
     };
 }
+
+export function fetchThreadMentionCountsByChannel() {
+    return async (dispatch, getState) => {
+        const state = getState();
+        if (isCollapsedThreadsEnabled(state)) {
+            dispatch(getThreadMentionCountsByChannel(getCurrentTeamId(state)));
```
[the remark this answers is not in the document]
calebroseland: (cc @reflog in case I'm off on something or missed anything)

- `getThreadMentionCountsByChannel` is done on a per-team basis
- Looks like this is fired on WS `reconnect` and `focusPost`
- it could get it's own state and short circuit internally, but currently all feature-on-off gate logic is outside of `actions/threads.ts`
  
In the end, looks like this team action is being removed in the next PR as further changes to mentions/unreads are underway.
https://github.com/mattermost/mattermost-webapp/pull/7733/files#diff-42c7d37cdea1c44e8773b4f6a75c9c03291f9bffebb8d268186c0a41ff929a3f

components/legacy_sidebar/sidebar_channel/index.js
```diff
@@ -62,7 +63,11 @@ function makeMapStateToProps() {
         let unreadMsgs = 0;
         let showUnreadForMsgs = true;
         if (member) {
-            unreadMentions = member.mention_count;
+            const threadCounts = getThreadCountsInCurrentTeam(
+                state,
+            );
```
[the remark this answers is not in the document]
calebroseland: Changing in next PR https://github.com/mattermost/mattermost-webapp/pull/7733/files#diff-b1ad3be55e1e68238de30c163be916663fb761d4493ed423aab7cc59c32b9aee

components/legacy_sidebar/sidebar_channel/index.js
```diff
@@ -62,7 +63,11 @@ function makeMapStateToProps() {
...
+            const threadMentionCountInChannel = (threadCounts && threadCounts.unread_mentions_per_channel && threadCounts.unread_mentions_per_channel[channel.id]) || 0;
```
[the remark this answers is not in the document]
calebroseland: @reflog ?

components/legacy_sidebar/sidebar_channel/index.js
```diff
@@ -62,7 +63,11 @@ function makeMapStateToProps() {
...
+            unreadMentions = member.mention_count - threadMentionCountInChannel;
```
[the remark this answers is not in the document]
calebroseland: @reflog ?

components/threading/common/button/button.tsx
lieut-data: `allowTextOverflow`, at least, now tells me what the intent is here. (Wasn't sure until now ;p)

matthewbirtch (approved):
Approving this now to not be a blocker and will do more robust UX testing when merged and on community-daily.

components/threading/global_threads_link/threads_icon.tsx
```diff
@@ -0,0 +1,23 @@
...
+            {...attrs}
+        >
+            <path
+                d='M11.7952 0.00524884C12.1312 0.00524884 12.4144 0.125249 12.6448 0.365248C12.8848 0.595648 13.0048 0.878848 13.0048 1.21485V8.41485C13.0048 8.75085 12.8848 9.03405 12.6448 9.26445C12.4144 9.49485 12.1312 9.61005 11.7952 9.61005H3.4L0.9952 12.0148V1.21485C0.9952 0.878848 1.1104 0.595648 1.3408 0.365248C1.5808 0.125249 1.8688 0.00524884 2.2048 0.00524884H11.7952ZM2.2048 1.21485V9.10605L2.896 8.41485H11.7952V1.21485H2.2048ZM3.4 3.01485H10.6V4.21005H3.4V3.01485ZM3.4 5.40525H8.8V6.61485H3.4V5.40525Z'
+            />
+        </svg>
+    );
+};
+
+export default memo(ThreadsIcon);
```
lieut-data, line 23: Do we require `memo` here? It seems like this component is simple enough that the cost of memoization may be greater than just letting React handle things.

components/widgets/users/avatar/avatar.tsx
```diff
@@ -23,11 +27,16 @@ const Avatar = ({
...
 }: Props & Attrs) => {
     const classes = classNames(`Avatar Avatar-${size}`, attrs.className);
+    const {centerChannelBg, centerChannelColor} = useSelector(getTheme);
 
     if (text) {
         return (
             <div
                 {...attrs}
+                style={{
+                    background: tinycolor.mix(centerChannelBg, centerChannelColor, 8).toRgbString(),
```
lieut-data, line 37: nit: it seems unusual to compute a colour based on our constants instead of using a constant defined by the theme directly. Are we sure that these colours will always "mix" correctly?

reducers/views/threads.ts
```diff
@@ -0,0 +1,24 @@
...
+import {combineReducers} from 'redux';
+
+import {GenericAction} from 'mattermost-redux/types/actions';
+
+import {ViewsState} from 'types/store/views';
+
+import {Threads} from 'utils/constants';
+
+export const selectedThreadIdInTeam = (state: ViewsState['threads']['selectedThreadIdInTeam'] | null = null, action: GenericAction) => {
+    if (action.type === Threads.CHANGED_SELECTED_THREAD) {
```
lieut-data, line 13: nit: can we use the `switch` block despite needing to only handle one action type?

reducers/views/threads.ts
```diff
@@ -0,0 +1,24 @@
...
+        return {
+            ...state,
+            [action.data.team_id]: action.data.id,
```
lieut-data, line 16: nit: should we be using `team_id` and `post_id` or some such to clearly explain what the `id` is?

selectors/views/threads.ts
```diff
@@ -0,0 +1,45 @@
...
+import {UserThread} from 'mattermost-redux/types/threads';
+
+import {GlobalState} from 'types/store';
+import {ViewsState} from 'types/store/views';
+
+export function getSelectedThreadIdInTeam(state: GlobalState) {
+    return state.views.threads.selectedThreadIdInTeam;
+}
+
+export const getSelectedThreadIdInCurrentTeam: (state: GlobalState) => ViewsState['threads']['selectedThreadIdInTeam'][$ID<Team>] = createSelector(
```
lieut-data, line 19: Do we need to memoize this with `createSelector`? (As in, is there evidence that this selector is both called often and slow to compute?) If not, can we just make this a simple function?

selectors/views/threads.ts
```diff
@@ -0,0 +1,45 @@
...
+import {$ID} from 'mattermost-redux/types/utilities';
+import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
+import {getThreads} from 'mattermost-redux/selectors/entities/threads';
+import {Team} from 'mattermost-redux/types/teams';
+import {UserThread} from 'mattermost-redux/types/threads';
+
+import {GlobalState} from 'types/store';
+import {ViewsState} from 'types/store/views';
+
+export function getSelectedThreadIdInTeam(state: GlobalState) {
```
lieut-data, line 15: Using the force, this one doesn't /feel/ like it should be exported, but that clients of the "API" here should be more specific.

selectors/views/threads.ts
```diff
@@ -0,0 +1,45 @@
...
+    getCurrentTeamId,
+    getSelectedThreadIdInTeam,
+    getThreads,
+    (
+        currentTeamId,
+        selectedThreadIdInTeam,
+        threads,
+    ) => {
+        const threadId = selectedThreadIdInTeam?.[currentTeamId];
+        if (!threadId) {
```
lieut-data, line 40: Redundant check given below?

selectors/views/threads.ts
```diff
@@ -0,0 +1,45 @@
...
+    getSelectedThreadIdInTeam,
+    (
+        currentTeamId,
+        selectedThreadIdInTeam,
+    ) => {
+        return selectedThreadIdInTeam?.[currentTeamId] ?? null;
+    },
+);
+
+export const getSelectedThreadInCurrentTeam: (state: GlobalState) => UserThread | null = createSelector(
```
lieut-data, line 30: Same question as above re: `createSelector`.

stores/hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+import {useSelector, useDispatch} from 'react-redux';
+import {createSelector} from 'reselect';
+
+import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
+import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
+
+import {makeGetGlobalItem} from 'selectors/storage';
+import {setGlobalItem} from 'actions/storage';
+
+export const currentUserAndTeamSuffix = createSelector([
```
lieut-data, line 13: Does this need to be memoized?

stores/hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+    getCurrentUserId,
+    getCurrentTeamId,
+], (
+    userId,
+    teamId,
+) => {
+    return `:${userId}:${teamId}`;
+});
+
+export const currentUserSuffix = createSelector([
```
lieut-data, line 23: Does this need to be memoized?

components/widgets/users/avatars/avatars.tsx
```diff
@@ -11,34 +20,88 @@ import SimpleTooltip, {useSynchronizedImmediate} from 'components/widgets/simple
...
+    participants: Props['participants'],
+): null | {users: AvatarProps[]; missingProfiles?: string[]} {
+    if (users) {
+        return {users};
+    }
+    if (!participants) {
+        return null;
+    }
+
+    const getDisplayName = makeGetDisplayName();
```
lieut-data, line 56: This creates a memoized selector on each invocation of this method.

components/widgets/users/avatars/avatars.tsx
```diff
@@ -11,34 +20,88 @@ import SimpleTooltip, {useSynchronizedImmediate} from 'components/widgets/simple
...
+    }
 
 const OTHERS_DISPLAY_LIMIT = 99;
 
+function selectUsers(
+    state: GlobalState,
+    users: Props['users'],
+    participants: Props['participants'],
+): null | {users: AvatarProps[]; missingProfiles?: string[]} {
+    if (users) {
```
lieut-data, line 49: This /seems/ to be a form of memoization? I'm having trouble following this method.

components/widgets/users/avatars/avatars.tsx
```diff
@@ -11,34 +20,88 @@ import SimpleTooltip, {useSynchronizedImmediate} from 'components/widgets/simple
...
+        return {users};
+    }
+    if (!participants) {
+        return null;
+    }
+
+    const getDisplayName = makeGetDisplayName();
+
+    return participants.reduce<NonNullable<ReturnType<typeof selectUsers>>>((result, {id}) => {
```
lieut-data, line 58: This returns a new object every time, which means this selector forces a React reconciliation every time, which may be a performance issue (i.e. "slow typing issue" recently worked around). But maybe that's what we're trying to avoid above with the `users` check?

Should we move all this into the render method instead, and only select the minimum necessary data from the store and do everything consistently at render time?

components/widgets/header/header.tsx
```diff
@@ -0,0 +1,41 @@
...
+                <H>{heading}</H>
+                {subtitle ? <p>{subtitle}</p> : null}
+            </div>
+            <div className='spacer'/>
+            {right}
+        </header>
+    );
+};
+
+export default memo(Header);
```
lieut-data, line 41: Do we need to memoize this?

components/threading/thread_viewer/thread_viewer.tsx
```diff
@@ -0,0 +1,564 @@
...
+        'minute',
+        'hour',
+        'day',
+        'week',
+        'month',
+        'year',
+    ],
+};
+export function renderView(props: Record<string, any>) {
+    return (
```
lieut-data, line 44: nit: for simple methods like these, I like the reduced-line-count pattern that eliminates the `return`:
```js
export function renderView = (props: Record<string, any>) => ({
    <div ...
});
```

components/threading/thread_viewer/thread_viewer.tsx
```diff
@@ -0,0 +1,564 @@
...
+
+    public constructor(props: Props) {
+        super(props);
+
+        this.scrollStopAction = new DelayedAction(this.handleScrollStop);
+
+        const openTime = (new Date()).getTime();
+
+        this.state = {
+            windowWidth: Utils.windowWidth(),
```
lieut-data, line 127: Haven't read the rest of the code, but this "smells" wrong -- how can we avoid needing to compute things based on the height in this way?

components/threading/thread_viewer/thread_viewer.tsx
```diff
@@ -0,0 +1,564 @@
...
+            windowHeight: Utils.windowHeight(),
+            isScrolling: false,
+            topRhsPostId: '',
+            openTime,
+            postsContainerHeight: 0,
+            userScrolledToBottom: false,
+        };
+
+        this.rhspostlistRef = React.createRef();
```
lieut-data, line 136: nit: `rhsPostListRef`?

components/threading/thread_viewer/thread_viewer.tsx
```diff
@@ -0,0 +1,564 @@
...
+        this.containerRef = React.createRef();
+        this.postCreateContainerRef = React.createRef();
+    }
+
+    private getLastPost() {
+        const childPosts = this.rhspostlistRef.current?.children;
+        return childPosts && childPosts[childPosts.length - 1];
```
lieut-data, line 143: Any chance this can be empty in weird cases (like posts being deleted)?

components/threading/thread_viewer/thread_viewer.tsx
```diff
@@ -0,0 +1,564 @@
...
+        }
+
+        const curLastPost = curPostsArray[0];
+
+        if (curLastPost.user_id === this.props.currentUserId || this.state.userScrolledToBottom) {
+            this.scrollToBottom();
+        }
+    }
+
+    public shouldComponentUpdate(nextProps: Props, nextState: State) {
```
lieut-data, line 246: Are we certain `shouldComponentUpdate` is required here? Usually this is a performance improvement, but only needed if we've measured one. And if we have to compute the equality of lots of things anyway, it may be a wash. 

On the flip side, if we're using this to work around a problem of updating when we shouldn't, let's identify and tackle the root problem there.

components/threading/thread_viewer/thread_viewer.tsx
```diff
@@ -0,0 +1,564 @@
...
+            if (cpost.root_id === selected.id) {
+                postsArray.unshift(cpost);
+            }
+        });
+
+        return postsArray;
+    }
+
+    public scrollToBottom = (): void => {
+        if ($('.post-right__scroll')[0]) {
```
lieut-data, line 324: Can we avoid introducing a new JQuery dependency? We're aiming to drop JQuery in Q2 2021.

components/favicon_title_handler/favicon_title_handler.tsx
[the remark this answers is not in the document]
calebroseland: @Willyfrog after a second look at this, I think we should leave it as-is. https://github.com/mattermost/mattermost-webapp/blob/master/components/favicon_title_handler/favicon_title_handler.test.jsx#L53-L56 tests specifically for a null sitename. If we do need to change it, I think this should be in a separate PR as it would change non-threading titles.

components/threading/global_threads_link/global_threads_link.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    const counts = useSelector(getThreadCountsInCurrentTeam);
+    const unreadsOnly = useSelector(isUnreadFilterEnabled);
+    const someUnreadThreads = counts?.total_unread_threads;
+
+    useEffect(() => {
+        dispatch(getThreads(currentUserId, currentTeamId, {perPage: 5}));
+    }, [currentTeamId]);
+
+    if (!isFeatureEnabled || (unreadsOnly && !threadsMatch && !someUnreadThreads)) {
```
lieut-data, line 43: Looks like the feature gate doesn't stop the `dispatch` above? Should it?

components/threading/global_threads_link/global_threads_link.tsx
```diff
@@ -0,0 +1,82 @@
...
+                    {counts?.total_unread_mentions > 0 && (
+                        <ChannelMentionBadge unreadMentions={counts.total_unread_mentions}/>
+                    )}
+                </Link>
+            </li>
+        </ul>
+    );
+};
+
+export default memo(GlobalThreadsLink);
```
lieut-data, line 82: Do we need to memoize this?

components/threading/global_threads/thread_pane/thread_pane.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+    return (
+        <div className='ThreadPane'>
+            <Header
+                className='ThreadPane___header'
+                heading={(
+                    <>
+                        <Button
+                            className='Button___icon Button___large back'
+                            onClick={useCallback(() => select(), [])}
```
lieut-data, line 65: Can we avoid inlining hook definitions?

components/threading/global_threads/thread_list/thread_list.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    return (
+        <div className={'ThreadList'}>
+            <Header
+                heading={(
+                    <>
+                        <Button
+                            className={'Button___large Margined'}
+                            isActive={currentFilter === ''}
+                            onClick={useCallback(() => setFilter(''), [])}
```
lieut-data, line 44: Let's avoid inlining hooks.

components/threading/global_threads/thread_item/thread_item.tsx
```diff
@@ -1,69 +1,128 @@
...
+        }
+    }, [channel, directTeammate]);
+
+    useEffect(() => {
+        if (!channel && thread?.post.channel_id) {
+            dispatch(fetchChannel(thread.post.channel_id));
+        }
+    }, [channel, thread?.post.channel_id]);
+
+    useEffect(() => {
```
lieut-data, line 70: This is duplicated from line 58?

lieut-data (changes requested):
Thanks, @calebroseland! You've tackled a huge project really well -- a few comments below from my final review.

(Also, I probably commented on a few things that amount to old code just being relocated and that I may not have picked up on during the incremental review. Just a heads up!)

components/rhs_comment/rhs_comment.jsx
```diff
@@ -198,6 +199,7 @@ class RhsComment extends React.PureComponent {
                 eventTime={post.create_at}
                 postId={post.id}
                 location={Locations.RHS_COMMENT}
+                timestampProps={{...this.props.timestampProps, style: this.props.isConsecutivePost && !this.props.compactDisplay ? 'narrow' : undefined}}
```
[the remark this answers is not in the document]
calebroseland: Not in this case. The underlying `intl` apis make a distinction between `undefined` and defined(/falsy) values

components/create_comment/create_comment.jsx
```diff
@@ -1233,7 +1233,6 @@ class CreateComment extends React.PureComponent {
             <form onSubmit={this.handleSubmit}>
                 <div
                     role='form'
-                    id='rhsFooter'
```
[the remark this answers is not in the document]
calebroseland: Yes, all E2E selectors relying on this were updated, but on second pass there was one remaining. Should be good now. 

Though I think I'll need to do a deeper follow up PR that addresses another set of Ids `#fileUploadButton`, `#fileUploadInput`, `#addCommentButton`, `#reply_textbox` etc.

components/threading/global_threads_link/threads_icon.tsx
calebroseland: Memo all the things
image

/s

I agree—not worth it! (will remove it)

components/widgets/users/avatar/avatar.tsx
calebroseland: This was a tricky one. It's a transparent bg, but there's an overlapping object so for this one element it needed to be opaque... testing in the 4 default themes we have yielded the expected colors, but there may be edge cases.

I'll mark it down for a more thorough UI check during UX review with custom themes.

reducers/views/threads.ts
calebroseland: Even though I think it's ok when the subject is small and clear, I do think in many cases it's good to avoid generic names.

selectors/views/threads.ts
calebroseland: Looks like a bad merge, will clean it up

selectors/views/threads.ts
calebroseland: Generally speaking, I see almost all these kinds of "simple selectors" being exported around here, so it was more/less aiming to follow the pattern. Though I do agree, since it isn't used externally, there's no need to export.

selectors/views/threads.ts
calebroseland: Generally, I haven't done extensive performance testing on the lower end here with these smaller selectors, so I'm not sure if they need memoization, but I generally default to them akin to preferring `PureComponent` over `Component`.

At this late stage, I'm tempted to say that we should leave them as-is for now, and address any no-need-to-memoize issues in a follow up PR shortly afterwards that would address any performance learnings (on both the too slow/not enough and no-need ends).

stores/hooks.ts
calebroseland: (Follow up PR?)

stores/hooks.ts
calebroseland: (Follow up PR?)

components/widgets/header/header.tsx
calebroseland: Yes, I believe it should, given the number of props and amount of content that may be in them.

It's also my understanding that a `memo`-wrapped functional component is generally equivalent to a `PureComponent` (less the state diffing), and we do usually recommend using that over `Component`.

components/threading/thread_viewer/thread_viewer.tsx
calebroseland: (was moved/existing; didn't want to touch unnecessary)

I like that pattern, too. But I'm up on the air on this one. At least while developing, I may add and remove `const`s/`let`s as the function evolves, so switching between these two becomes somewhat burdensome IMHO..  


```tsx
const thing = () => (
    // jsx
);
```
and
```tsx
const thing = () => {
    return (
        // jsx
    );
};
```

components/threading/thread_viewer/thread_viewer.tsx
calebroseland: (was moved/existing; didn't want to touch unnecessary)

components/threading/thread_viewer/thread_viewer.tsx
calebroseland: (was moved/existing; didn't want to touch unnecessary)

components/threading/thread_viewer/thread_viewer.tsx
calebroseland: (was moved/existing; didn't want to touch unnecessarily)

components/threading/thread_viewer/thread_viewer.tsx
calebroseland: This was already here, and I'm not 100% sure of the history, but the only change I added was was for if the `userThread` was different.

I agree though that it may not be needed or the root problem should be addressed in a more appropriate way, though I do think any significant refactoring should be done separately.

`ThreadViewer` is really just the guts of `RHSThread` extracted to be reusable.

components/threading/global_threads_link/global_threads_link.tsx
calebroseland: Good catch! Yes, it should

components/threading/global_threads_link/global_threads_link.tsx
calebroseland: There's enough logic in it that I believe it is worth it, but I don't have any benchmarks to prove the need.

selectors/views/threads.ts
calebroseland: (Follow up PR?)

components/threading/global_threads_link/global_threads_link.tsx
lieut-data: Propose we drop memoization in call cases until we can identify a specific problem and show that it fixes it -- I'm worried that in some cases, the memoization costs us more due to the component having to re-render too often anyway.

components/widgets/users/avatars/avatars.tsx
calebroseland: The first implementation of this actually did just map in the render method, but there it isn't able to take username display preference into account. 

I knew there was a perf issue with returning an obj/arr here, but doing the selection/mapping at this level _seemed_ best at the time.

Last night I was brainstorming on another redo of this—I think I'll:
- hoist the missing profiles logic and use `getMissingProfilesByIds` higher up (API layer)
- lower/defer `userId`→`Avatar` props selection into `Avatar` and add new selector for `overflowUsers`→namesOverflow tooltip

components/widgets/users/avatars/avatars.tsx
calebroseland: It's an API variance short circuit if the more-direct `AvatarProps[]` were provided instead of `Array<{id: UserId}> which requires selection`

components/widgets/header/header.tsx
calebroseland: After another look—you're right, we don't need this. The render props make it pointless.

components/threading/thread_viewer/thread_viewer.tsx
calebroseland: Even if there's 0 children, it appears to be `null`-safe upstream

```tsx
        const lastPost = this.getLastPost()?.getBoundingClientRect();
        //...
        if (lastPost) {
            const userScrolledToBottom = scrollTop >= maxScroll - lastPost.height;

            this.setState({userScrolledToBottom});
        }
```

components/no_results_indicator/index.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export {default} from './no_results_indicator';
```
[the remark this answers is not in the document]
calebroseland: We had discussed a while back about an indexed-folder pattern to contain like-files together (scss, tests, storybook stories, etc.).

components/timestamp/timestamp.tsx
calebroseland: (`autoRange` just determines which `RangeDescriptor` to use.)

This line specifically provides a default `updateIntervalInSeconds` from props if the range descriptor doesn't provide one itself.

(The fallback to both of these are statically defined intervals based on the display unit.)

components/widgets/users/avatars/avatars.tsx
calebroseland: I reworked this into a new selector that will only memoize the display preference, and return a memoized selector function.

components/widgets/users/avatars/avatars.tsx
calebroseland: I opted to remove the API variance and simplify it to just user ids.

calebroseland:
> another thing i noticed, on a long thread (not fitting the screen) if I press enter I get a different scroll than if I start writing (the add comment button doesn't show in the first case)

@Willyfrog could you elaborate a bit more on this? Is this in Global Threads? RHS? What was focused when you stated typing and/or pressed enter?

Willyfrog:
> > another thing i noticed, on a long thread (not fitting the screen) if I press enter I get a different scroll than if I start writing (the add comment button doesn't show in the first case)
> 
> @Willyfrog could you elaborate a bit more on this? Is this in Global Threads? RHS? What was focused when you stated typing and/or pressed enter?

sorry, I'll try to do a more detailed repro steps, anyways it is a minor thing.
1. open global threads, select a long thread (larger than the screen)
1. click on the text box
1. scroll up the thread
1. hit enter

you'll see that it is scrolled but the add comment button is not visible

repeat steps but type a character instead of enter. now the button is visible.

calebroseland:
> > > another thing i noticed, on a long thread (not fitting the screen) if I press enter I get a different scroll than if I start writing (the add comment button doesn't show in the first case)
> > 
> > 
> > @Willyfrog could you elaborate a bit more on this? Is this in Global Threads? RHS? What was focused when you stated typing and/or pressed enter?
> 
> sorry, I'll try to do a more detailed repro steps, anyways it is a minor thing.
> 
> 1. open global threads, select a long thread (larger than the screen)
> 2. click on the text box
> 3. scroll up the thread
> 4. hit enter
> 
> you'll see that it is scrolled but the add comment button is not visible
> 
> repeat steps but type a character instead of enter. now the button is visible.

Hmm that's strange. This is what I'm seeing: Input change always scrolls to bottom. Enter (for empty input) only scrolls to bottom if the _whole_ input is out of view. (Chrome/Safari/Desktop in Global Threads and RHS—here and on community-daily.)
CleanShot 2021-04-06 at 16 10 50

components/favicon_title_handler/favicon_title_handler.tsx
```diff
@@ -83,21 +84,32 @@ class FaviconTitleHandler extends React.PureComponent<Props> {
...
                 }
             }
-            const mentionTitle = unreads.mentionCount > 0 ? '(' + unreads.mentionCount + ') ' : '';
-            const unreadTitle = !this.isDynamicFaviconSupported && unreads.messageCount > 0 ? '* ' : '';
-            document.title = mentionTitle + unreadTitle + currentChannelName + ' - ' + currentTeam.display_name + ' ' + currentSiteName;
+            document.title = `${mentionTitle + unreadTitle + currentChannelName} - ${currentTeam.display_name} ${currentSiteName}`;
+        } else if (currentTeam && inGlobalThreads) {
+            document.title = formatMessage({
+                id: 'globalThreads.title',
+                defaultMessage: '{prefix} Threads - {displayName} {siteName}',
```
Willyfrog, line 107:
```suggestion
                defaultMessage: '{prefix}Threads - {displayName} {siteName}',
```
also, would it be possible to just `formatMessage` the word threads and use the same formula for both? it will prevent subtle errors like this to happen.

components/threading/channel_threads/thread_footer/thread_footer.tsx
```diff
@@ -17,15 +17,14 @@ import SimpleTooltip from 'components/widgets/simple_tooltip';
 import {THREADING_TIME} from '../../common/options';
 
 type Props = {
-    participants: ComponentProps<typeof Avatars>['users'];
+    participants: Array<{id: string}>; // Post['participants']
     totalParticipants?: number;
```
Willyfrog, line 21: this now seems redundant, right? we can probably just get it from the length of `participants`

components/threading/global_threads/global_threads.tsx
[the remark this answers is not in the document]
Willyfrog: if we are nesting `if`s for me that's a bit too much logic into what should only be rendering.

components/threading/global_threads_link/global_threads_link.tsx
```diff
@@ -0,0 +1,85 @@
...
+        // load counts if necessary
+        if (isFeatureEnabled) {
+            dispatch(getThreads(currentUserId, currentTeamId, {perPage: 5}));
+        }
+    }, [currentTeamId]);
+
+    if (!isFeatureEnabled || (unreadsOnly && !inGlobalThreads && !someUnreadThreads)) {
+        // hide link if feature disabled or filtering unreads and there are no unread threads
+        return null;
+    }
```
Willyfrog, line 49: nit: shouldn't this be before the `useEffect` call?

Willyfrog (changes requested):
I used `/test threaded_post` and it seems it takes a while to load, but then I can scroll with no delay. I'm not really sure how that would work with really long threads in customer setups.

Also, while waiting for the thread to load I didn't have any sort of loading animation, but I guess this could be in a future PR.

Also, by creating a thread with that command, it didn't collapse into a single message

some more weird stuff. I created a second account and joined the `guille` team. this new user didn't see any replies, but just the initial message (screenshot #1). After that I clicked the reply button and closed the RHS, and now it showed that there were 7 replies (screenshot #2). Finally I replied and my global threads lighted as if there were messages unread, but the original user didn't get anything (screenshot #3)
Screen Shot 2021-04-06 at 22 21 55
Screen Shot 2021-04-06 at 22 22 16
Screen Shot 2021-04-06 at 22 22 51

components/threading/global_threads/thread_pane/thread_pane.tsx
```diff
@@ -0,0 +1,123 @@
...
+                                {channel.display_name}
+                            </Button>
+                        </h3>
+                    </>
+                )}
+                right={(
+                    <>
+                        <FollowButton
+                            isFollowing={isFollowing}
+                            disabled={isFollowing == null}
```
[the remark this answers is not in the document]
calebroseland: The main reason it's in so many places is that isFollowing can come from either the `post` or `userThread`, and it is not guaranteed to be present in `userThread`. The disabled part is an earlier piece, but may not be needed. It was used as a momentary block for while the thread was loading (don't want to allow conflicting toggle requests being sent). I'll have to double check, but we may be able to remove in a follow up PR.

components/channel_layout/center_channel/center_channel.tsx
```diff
@@ -76,9 +87,15 @@ export default class CenterChannel extends React.PureComponent<Props, State> {
...
+                            path='/:team/:path(channels|messages)/:identifier/:postid?'
                             component={ChannelIdentifierRouter}
                         />
+                        {isCollapsedThreadsEnabled ? (
+                            <Route
+                                path='/:team/threads/:threadIdentifier?'
+                                component={LazyGlobalThreads}
+                            />
+                        ) : null}
                         <Redirect to={lastChannelPath}/>
```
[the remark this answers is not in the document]
calebroseland: I dug a bit into those errors and it's related to a failed DB migration, it probably caused a number of issues. Re-creating the test server should hopefully resolve those (doing now)

components/threading/channel_threads/thread_footer/thread_footer.tsx
calebroseland: For the purposes of the Avatars tooltip, more than (some threshold) may be too many to display—like with reactions tooltips, I think the first XX are named with the rest as numbered overflow ("and 23 others..."). In the GA/polish phase `participants` will be optimized around this, but for now just shows all.

components/channel_layout/center_channel/center_channel.tsx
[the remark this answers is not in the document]
calebroseland: Also, there have been significant changes to mentions/unread with the feature on. These server changes have already been merged, and so this PR will throw and catch some API errors around this. 

https://github.com/mattermost/mattermost-webapp/pull/7733 is the webapp-side to those changes and will be merged (after this one) before we turn anything on. 

Re: this PR, it's understood that there are issues with mentions/unreads with the feature on—they will be fixed/developed incrementally ASAP. What's crucial is that there are no feature-off regressions.

components/favicon_title_handler/favicon_title_handler.tsx
```diff
@@ -83,21 +84,32 @@ class FaviconTitleHandler extends React.PureComponent<Props> {
             currentTeam,
             currentTeammate,
             unreads,
+            inGlobalThreads,
         } = this.props;
         const {formatMessage} = this.props.intl;
 
         const currentSiteName = siteName || '';
 
+        const mentionTitle = unreads.mentionCount > 0 ? '(' + unreads.mentionCount + ') ' : '';
```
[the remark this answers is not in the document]
calebroseland: I was tempted to make aside improvements like this, but this was how it was—it may have shifted, but I didn't want to change anything unnecessarily.

components/threading/global_threads_link/global_threads_link.tsx
calebroseland: No, hooks need to be called in the same order every time (and before early returns)

components/favicon_title_handler/favicon_title_handler.tsx
calebroseland: I did consider that (to match above), but I felt it would be better to move towards what `sidebar.team_select` is like below.

components/favicon_title_handler/favicon_title_handler.tsx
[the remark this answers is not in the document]
calebroseland: > I used `/test threaded_post` and it seems it takes a while to load, but then I can scroll with no delay. I'm not really sure how that would work with really long threads in customer setups.
> Also, while waiting for the thread to load I didn't have any sort of loading animation, but I guess this could be in a future PR.

@Willyfrog Unfortunately this is about what we have now—opening long threads in RHS will do a blocking render until all posts have loaded+rendered.
CleanShot 2021-04-06 at 22 03 49@2x
But I agree, we should make some improvements (ThreadViewer infinite scrolling with non-blocking rendering, loading indicators, etc.)
 
 
> Also, by creating a thread with that command, it didn't collapse into a single message

Full support for the in-channel experience will come in subsequent PRs. Some server/mm-redux work that Eli did got us part way there with Init channel-loading only including root posts, but a follow up PR is needed to explicitly disregard replies (right now they come in via WS events), add ThreadFooter, and a number of other items.

> some more weird stuff. I created a second account and joined the `guille` team. this new user didn't see any replies, but just the initial message (screenshot #1). After that I clicked the reply button and closed the RHS, and now it showed that there were 7 replies (screenshot #2). Finally I replied and my global threads lighted as if there were messages unread, but the original user didn't get anything (screenshot #3)

I'm hoping all of this is related to a DB migration issue. I re-created this test server, so replies in Global Threads should be working as-expected, though as previously mentioned: sidebar mentions/unread indicators are still in-progress until #7733 is merged.

Willyfrog:
> > > > another thing i noticed, on a long thread (not fitting the screen) if I press enter I get a different scroll than if I start writing (the add comment button doesn't show in the first case)
> > > 
> > > 
> > > @Willyfrog could you elaborate a bit more on this? Is this in Global Threads? RHS? What was focused when you stated typing and/or pressed enter?
> > 
> > 
> > sorry, I'll try to do a more detailed repro steps, anyways it is a minor thing.
> > 
> > 1. open global threads, select a long thread (larger than the screen)
> > 2. click on the text box
> > 3. scroll up the thread
> > 4. hit enter
> > 
> > you'll see that it is scrolled but the add comment button is not visible
> > repeat steps but type a character instead of enter. now the button is visible.
> 
> Hmm that's strange. This is what I'm seeing: Input change always scrolls to bottom. Enter (for empty input) only scrolls to bottom if the _whole_ input is out of view. (Chrome/Safari/Desktop in Global Threads and RHS—here and on community-daily.)
> CleanShot 2021-04-06 at 16 10 50

i was testing that on firefox, maybe it is a firefox-specific thing

lieut-data (approved):
Caleb, I'm content with my first review pass and Guillermo's eyes in wrapping this up, so I'm lifting my changes requested. Looking forward to seeing this live soon!

components/threading/global_threads/global_threads.tsx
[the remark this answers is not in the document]
calebroseland: I can refactor this into smaller pieces, and will, but I believe we should leverage the feature flag and make changes in a follow up PR given this late-stage in the process. (Beta and QA urgency with the next cloud branch cut on Monday.)

calebroseland:
@Willyfrog I think I've covered all the main points, with a few feature-on (flag-protected code) points aimed for follow up improvements. Were there any other feature-off concerns or points before we move on to QA?

Willyfrog (approved):
> @Willyfrog I think I've covered all the main points, with a few feature-on (flag-protected code) points aimed for follow up improvements. Were there any other feature-off concerns or points before we move on to QA?

Yeah, we can keep those for future PRs. lets see if we can get this merged soon 🚀

zefhemel:
Alright, looks like this is ready to merge. Let's do this as soon as the new cloud branch is cut (hopefully later today), to give it us plenty of time to ensure this doesn't regress anything before the next release.

ogi-m (approved):
Thanks Caleb! Approving from QA side 👍🏻 🎉

PR #7186, opened 2020-12-08
Cypress/E2E: Update MM-T131 to include additional step/verification from archived MM-T2154

saturninoabril:
#### Summary
Updated automated test for [MM-T131](https://automation-test-cases.vercel.app/test/MM-T131) to include additional step/verification from archived MM-T2154.

Follow up to: MM-30376


e2e/cypress/integration/messaging/message_draft_then_switch_channel_spec.js
```diff
@@ -24,46 +22,57 @@ describe('Message Draft and Switch Channels', () => {
...
 
-        // * Click on hint in modal to get out of overlapping suggestion list
-        cy.get('#quickSwitchHint').click();
+        // * Verify that the switch model is shown
+        cy.findAllByRole('dialog').first().findByText('Switch Channels').should('be.visible');
 
         // # Type the first few letters of the channel name you typed the message draft in
-        cy.get('#quickSwitchInput').type('tow');
-        cy.wait(TIMEOUTS.HALF_SEC);
+        cy.get('#quickSwitchInput').type(testChannel.substring(0, 3));
```
furqanmlk, line 50:
```suggestion
        cy.findByRole('textbox').type(testChannel.substring(0, 3));
```

furqanmlk: @saturninoabril 
All look good to me, except one suggestion :)

saturninoabril: Thanks for the review. Yeah, I should've done that :). I've added an aria-label and so this could be `cy.findByRole('textbox', {name: 'quick switch input'})`.  Please take a look again.

components/quick_switch_modal/quick_switch_modal.tsx
```diff
@@ -317,6 +317,7 @@ export default class QuickSwitchModal extends React.PureComponent<Props, State>
                         <i className='icon icon-magnify icon-16'/>
                         <SuggestionBox
                             id='quickSwitchInput'
+                            aria-label='quick switch input'
```
saturninoabril, line 320: Hello @devinbinnie, I added you as a reviewer especially for this added `aria-label`.

devinbinnie: Might want this to be a translatable string.

e2e/cypress/integration/account_settings/sidebar/channel_switcher_spec.js
```diff
@@ -76,7 +76,7 @@ describe('Account Settings > Sidebar > Channel Switcher', () => {
         cy.get('#quickSwitchHint').should('be.visible').should('contain', 'Type to find a channel. Use UP/DOWN to browse, ENTER to select, ESC to dismiss.');
 
         // # Type CTRL/CMD+shift+L
-        cy.get('#quickSwitchInput').cmdOrCtrlShortcut('{shift}L');
+        cy.findByRole('textbox', {name: 'quick switch input'}).cmdOrCtrlShortcut('{shift}L');
```
devinbinnie, line 79: Could we not use a `data-testid` tag to find this instead of searching by an `aria-label`?

devinbinnie (changes requested):

e2e/cypress/integration/account_settings/sidebar/channel_switcher_spec.js
saturninoabril: We wanted to test it like on how the user would find the element (either visually or thru assistive technology) and less under the hood. So, it's better to have an aria-label (if possible/applicable). If not, then we can use other approach like `data-testid`.

devinbinnie: I think it would make sense to maybe have one test to make sure the user can find the element that way, but not as a general practice. I think for the other tests it would make more sense to use a `data-testid` attribute so we can reliably find the element. Since we're using an automated testing tool anyways I don't see this being too much of a sacrifice wrt. emulating the UX. 3/5

saturninoabril: If it's going to test other [attribute of the element](https://github.com/mattermost/mattermost-developer-documentation/pull/745/files#diff-b28dacdaa9e85745c4e44fbafc016a2f3a7b6077b31fedce8dab7b94d34847c6R35-R45) like `cy.findByTestId('testId`).should('have.attr', ...)` or accessing a section within, then I agree. But in most cases, we'd like to use queries that are "accessible to everyone". We're documenting this approach. Feel free to comment as well at https://github.com/mattermost/mattermost-developer-documentation/pull/745/files#diff-b28dacdaa9e85745c4e44fbafc016a2f3a7b6077b31fedce8dab7b94d34847c6R10. It's based from [Testing library's "which query to use"](https://testing-library.com/docs/guide-which-query/).

devinbinnie: Didn't know that was a thing, go with that then.

components/quick_switch_modal/quick_switch_modal.tsx
saturninoabril: Done, please take a look again. Thanks!

devinbinnie (approved):

furqanmlk (approved):

e2e/cypress/integration/messaging/message_draft_then_switch_channel_spec.js
```diff
@@ -24,46 +22,57 @@ describe('Message Draft and Switch Channels', () => {
...
 });
 
+function verifyDraftIcon(channelName, isVisible) {
+    cy.get('#sidebar-left').findByLabelText(`${channelName.toLowerCase()} public channel`).
+        should('be.visible').
+        find('#draftIcon').
+        should(isVisible ? 'be.visible' : 'not.exist');
+}
+
+function openChannelFromLhs(teamName, displayName, name) {
```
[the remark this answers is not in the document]
furqanmlk: Looks good to me :)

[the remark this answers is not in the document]
saturninoabril: I agree. Let's have it on a separate PR so we can plan better on how to approach the LHS sidebar and document as well similar to [Account Settings Modal](https://github.com/mattermost/mattermost-developer-documentation/pull/745/files#diff-229615768ae134f8f1d52447b0fe7a87345f9c3cea539c41878b1c795cf8ab82R2).

saturninoabril:
I'm merging this now.

PR #7195, opened 2020-12-09
[MM-30851/MM-31405] Replace ExperimentalChannelSidebarOrganization with EnableLegacySidebar

devinbinnie:
#### Summary
Migrating to a new setting called EnableLegacySidebar to enable the old sidebar.
The new sidebar will be enabled by default system-wide from now on.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30851
https://mattermost.atlassian.net/browse/MM-31405

#### Related Pull Requests
- Has server changes (https://github.com/mattermost/mattermost-server/pull/16529)
- Has redux changes (https://github.com/mattermost/mattermost-redux/pull/1316)


components/admin_console/admin_definition.jsx
```diff
@@ -5626,18 +5584,31 @@ const AdminDefinition = {
...
+                        isHidden: it.configIsFalse('ServiceSettings', 'EnableLegacySidebar'),
+                        isDisabled: it.not(it.userHasWritePermissionOnResource('experimental')),
+                    },
+                    {
+                        type: Constants.SettingsTypes.TYPE_BOOL,
+                        key: 'ServiceSettings.CloseUnusedDirectMessages',
+                        label: t('admin.experimental.closeUnusedDirectMessages.title'),
+                        label_default: 'Autoclose Direct Messages in Sidebar:',
+                        help_text: t('admin.experimental.closeUnusedDirectMessages.desc'),
+                        help_text_default: 'When true, direct message conversations with no activity for 7 days will be hidden from the sidebar. When false, conversations remain in the sidebar until they are manually closed.',
```
nevyangelova, line 5609: this should probably be removed when we implement the limit visible dms by the user because we will get rid of the autoclose logic. Just adding a comment to keep in mind for myself :)

nevyangelova (approved):

components/admin_console/admin_definition.jsx
```diff
@@ -5589,35 +5569,13 @@ const AdminDefinition = {
...
+                        type: Constants.SettingsTypes.TYPE_BOOL,
+                        key: 'ServiceSettings.EnableLegacySidebar',
+                        label: t('admin.experimental.enableLegacySidebar.title'),
+                        label_default: 'Enable Legacy Sidebar',
+                        help_text: t('admin.experimental.enableLegacySidebar.desc'),
+                        help_text_default: 'When enabled, users cannot access new sidebar features including custom, collapsible categories and unread channel filtering. We recommend only enabling the legacy sidebar if users are experiencing breaking changes or bugs.',
```
hmhealey, line 5577: Not to be totally pedantic, but this change is technically a breaking change, so they're guaranteed to experience at least one. We might want to remove that part of the help text.

components/admin_console/admin_definition.jsx
```diff
@@ -5626,18 +5584,31 @@ const AdminDefinition = {
                         label: t('admin.experimental.experimentalChannelOrganization.title'),
                         label_default: 'Channel Grouping and Sorting',
                         help_text: t('admin.experimental.experimentalChannelOrganization.desc'),
-                        help_text_default: 'Enables channel sidebar organization options in **Account Settings > Sidebar > Channel grouping and sorting** including options for grouping unread channels, sorting channels by most recent post and combining all channel types into a single list. These settings are not available if **Account Settings > Sidebar > Experimental Sidebar Features** are enabled.',
+                        help_text_default: 'Enables channel sidebar organization options in **Account Settings > Sidebar > Channel grouping and sorting** including options for grouping unread channels, sorting channels by most recent post and combining all channel types into a single list.',
```
hmhealey, line 5587: I think we might usually disable these with some help text instead of completely hiding the setting. I don't think we usually hide settings unless they're not available for licensing reasons. I could be wrong on that though.

components/admin_console/admin_definition.jsx
hmhealey: We still need it for the old sidebar, so it can't be removed until that goes.

components/admin_console/admin_definition.jsx
devinbinnie: I can disable it instead of hide it.

components/admin_console/admin_definition.jsx
devinbinnie: I think the point is that this should only be turned on if the new sidebar is broken. It's not necessarily a breaking change for all users, only some.

hmhealey: I guess there's just some ambiguity about what constitutes a breaking change. I'm looking at it that something major is changing and some old functionality isn't available because of that, so that's my thinking there. That said, that doesn't particularly matter.

hmhealey (approved):

jgilliam17 (approved):
Thank you @devinbinnie 
Tested, looks good to merge.
- Verified new Enable Legacy Sidebar feature toggle in System Console; old setting has been removed. 
- There are more 20+ changes to E2Es; Approving for now without running the tests and will  follow-up with Saturn after next run on master for any issues that arise from these changes

PR #7200, opened 2020-12-10
[MM-31149] don't allow change of roles if user can't manage members

Willyfrog:
#### Summary
Since there is no specific control from the system console to change `MANAGE_CHANNEL_ROLES` it feels weird that a user can't manage users but can change its role. This PR adds a check for user being able to manage users in addition to channel roles.

#### Ticket Link

[MM-31149](https://mattermost.atlassian.net/browse/MM-31149)


hmhealey:
Doesn't something need to change on the server for this to be enforced there as well?

Willyfrog:
yes, you are right! I'll submit another PR for that

Willyfrog:
@hmhealey added https://github.com/mattermost/mattermost-server/pull/16566

sbishel (approved):

hmhealey (approved):

jgilliam17 (approved):
Thank you @Willyfrog 
Tested, looks good to merge.
- Verified when user lacks permission to manage member roles, View Members modal only displays read-only role labels.

PR #7207, opened 2020-12-11
MM-31275 New sidebar performance improvements

hmhealey:
Here are some performance improvements around the new sidebar when the user is a member of many channels. I focused on two areas:
1. The new sidebar and some other components re-rendering when the user clicks anywhere. This is new to 5.30.
2. The sidebar lags heavily when scrolling. This existed in 5.29, and it may have gotten worse, but my master test server was in dev mode, so that may have made the lag worse.

When investigating the re-rendering:
1. The new sidebar would entirely re-render when clicking anywhere in the app because that would trigger a MULTISELECT_CHANNEL_CLEAR action which caused `state.views.channel_sidebar.multiselectedChannelIds` to change even when it was already empty. That's fixed now.
2. Similarly, some other random components would also re-render when clicking anywhere because of the same action. That happened because:
    1. Some selectors are not memoized properly, so they return new objects on every call.
    2. redux-offline makes every dispatched action trigger the store state to change, so `mapStateToProps` is called for all components on every action.

    Because of both of those, dispatching the MULTISELECT_CHANNEL_CLEAR on every click was causing some older components to re-render. Those should be fixed eventually (I have a list of tickets to write), but in the mean time, we just won't dispatch that action when it's not necessary.

When investigating the scrolling lag:
1. I added a debounce to the `onScroll` handlers used by both the new and old sidebars to show the More Unreads Above/Below indicators. This didn't have a huge effect, but it might be helpful anyways.
2. After a bunch of commenting out stuff, I found that the main source of the sidebar lag was the `SidebarMenu` components because they attached their own scroll handlers to the sidebar so that they could close the menu when the user scrolls. As you get more channels in the sidebar, more handlers were added, and the lag got worse. I remove those, although that does mean that **those menus will stay open and won't scroll with the sidebar**, but I think that's an acceptable defect since they caused a ton of lag.

I also spent a bit of time investigating the sidebar re-rendering when the user was typing because the sidebar listens to local storage where drafts are saved, but that didn't seem to change anything.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31275


components/sidebar/sidebar_menu/sidebar_menu.tsx
```diff
@@ -59,22 +59,6 @@ export default class SidebarMenu extends React.PureComponent<Props, State> {
         }
     }
 
-    // TODO: Temporary code to keep the menu in place while scrolling
-    // This shouldn't be necessary once the menus are fixed up
-    componentDidMount() {
```
devinbinnie, line 64: Are these just not needed anymore?

devinbinnie (approved):

components/sidebar/sidebar_menu/sidebar_menu.tsx
hmhealey: This is breaking the functionality that closes the menus when you scroll the sidebar, but I couldn't come up with a quick, low-risk way to do that with how the menus are written. I think the performance impacts of this change heavily outweigh the slight UX degredation.

enahum: Can you file a ticket so this can be revisited?

enahum (approved):

jgilliam17 (approved):
Thank @hmhealey 
Tested, looks good to merge.
- Verified sidebar More Unreads indicator works as before. Verified multi-selected drag and drop group of channels can be cleared.

jespino (approved):
Lgtm

amyblais:
@hmhealey Seems to need a manual cherry-pick.

hmhealey:
@amyblais Yeah, I realized when talking to Jelena that some of the stuff I fixed wasn't in 5.30. The major performance improvement around scrolling will still have an effect though

PR #7209, opened 2020-12-12
[MM-27209] - Add spec for archived channels part 03

nevyangelova:
#### Summary
This PR adds 7 cases (for web) from the archived channels part 03 epic

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27209


e2e/cypress/integration/channel/archived_channels_spec.js
```diff
@@ -156,4 +158,319 @@ describe('Leave an archived channel', () => {
...
+                cy.get('#moreChannelsList').should('not.contain', channel.name);
+            });
+        });
+    });
+
+    it('MM-T1697 - More... for Public channels shows archived channels option', () => {
+        cy.apiAdminLogin();
+        cy.apiUpdateConfig({
+            ServiceSettings: {
+                ExperimentalChannelSidebarOrganization: 'default_off',
```
devinbinnie, line 193: Can you add `EnableLegacySidebar: 'true'` here as well or will that break things?
We're moving to that new setting so I want to make sure my changes won't break this test.

e2e/cypress/integration/channel/archived_channels_spec.js
```diff
@@ -156,4 +158,319 @@ describe('Leave an archived channel', () => {
...
+                cy.get('#moreChannelsList').should('contain', channel.display_name);
+            });
+        });
+    });
+
+    it('MM-T1698 - More... for Private channels shows archived channels option', () => {
+        cy.apiAdminLogin();
+        cy.apiUpdateConfig({
+            ServiceSettings: {
+                ExperimentalChannelSidebarOrganization: 'default_off',
```
devinbinnie, line 229: In hindsight, the more I see us doing this the more I wonder if we should just change the test procedure to use the new sidebar instead. What do you think/how hard is that to do?

e2e/cypress/integration/channel/archived_channels_spec.js
nevyangelova: Oh yeah sure but maybe we should wait for those changes to go in first? In order to keep the tests passing?

e2e/cypress/integration/channel/archived_channels_spec.js
nevyangelova: The ticket has tests for both new and old sidebar, and for example wants to test old sidebar specific features like channel grouping 🤷‍♀️  If we are getting rid of the old sidebar soon then these test cases will be redundant then too but for now I guess we are keeping them?

devinbinnie: That's true, though I'm wondering what the value is in writing these tests for the old sidebar. Especially considering that the main purpose for the push in writing these tests is for cloud, which will never be using the old sidebar. Let's chat about this in triage.

e2e/cypress/integration/channel/archived_channels_spec.js
devinbinnie: Sure, we'll just need to remember to fix these later. My PRs won't be going in for a while since we don't want to force the new sidebar on until the auto-closing and new unread category are complete. I'll try to remember when I merge my stuff to do that :P

e2e/cypress/integration/channel/archived_channels_spec.js
nevyangelova: Agreed.

saturninoabril (changes requested):
@nevyangelova Could you please run the tests and confirm if all are passing on you. Most are failing on me.
Screen Shot 2020-12-17 at 10 32 46 PM
Also after system config change, I suggests to use a regular user for testing (instead of sysadmin) as much as possible so we can ensure that it starts with the same expected user setting, and will not be affected by other test's change.

nevyangelova:
@saturninoabril @devinbinnie updated all tests to use the new sidebar. Removed Admin login and left it only for creating private channels. All tests are passing for me. Updated the test cases :)

devinbinnie (approved):

e2e/cypress/integration/channel/archived_channels_spec.js
```diff
@@ -156,4 +158,249 @@ describe('Leave an archived channel', () => {
...
+        cy.get('#moreChannelsModal').should('be.visible').within(() => {
+            cy.findByText('Show: Public Channels').should('be.visible').click();
+
+            // # Go to archived channels
+            cy.findByText('Archived Channels').click();
+
+            // # More channels list should contain the archived channel
+            cy.get('#moreChannelsList').should('contain', archivedChannel.display_name);
+        });
+    });
```
saturninoabril, line 405: I'm getting failing tests here, probably due to sysadmin's user preference. If possible, after config or other requests, please use regular user instead of sysadmin. Also, does the config change at MM-T1696 applicable to the succeeding tests. If yes, then I suggest to move all new tests to new spec file and have the config change in the `before` hook.
Screen Shot 2020-12-19 at 3 56 02 AM
Screen Shot 2020-12-19 at 3 56 39 AM

saturninoabril (changes requested):
Thanks @nevyangelova! Please see comments.

e2e/cypress/integration/channel/archived_channels_spec.js
nevyangelova: @saturninoabril  
> Also, does the config change at MM-T1696 applicable to the succeeding tests. If yes, then I suggest to move all new tests to new spec file and have the config change in the before hook.

No, we only need it for that one test case to confirm that when archived channels feature is disabled users cannot viewthem.

nevyangelova:
@saturninoabril fixed, removed all admin user apart from MM-T1696 since that needs to change the config for the test, and I moved it below the other test cases because of the different config settings.

saturninoabril (approved):
Thanks Nevy! LGTM, tested and all passed.

PR #7210, opened 2020-12-12
MM-20440 edit oauth app typescript migration

compiledsound:
#### Summary
edit oauth app Typescript migration

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20440

Fixes 

https://github.com/mattermost/mattermost-server/issues/16134


nevyangelova (approved):
LGTM!

components/integrations/edit_oauth_app/edit_oauth_app.test.tsx
```diff
@@ -21,9 +26,22 @@ describe('components/integrations/EditOAuthApp', () => {
         update_at: 1501365458934,
         callback_urls: ['https://test.com/callback', 'https://test.com/callback2'],
     };
-    const team = {
+    const team: Team = {
```
hmhealey, line 29: Instead of specifying all these extra fields, you could either use `TestHelper.getTeamMock` from `utils/test_helper` or if those fields aren't needed, you can use `const team = {...} as Team;` to force the type to be treated as `Team` despite the missing fields. Either way, you could save some typing that way.

components/integrations/edit_oauth_app/edit_oauth_app.test.tsx
```diff
@@ -76,7 +95,7 @@ describe('components/integrations/EditOAuthApp', () => {
         );
 
         wrapper.setState({showConfirmModal: false});
-        wrapper.instance().handleConfirmModal();
+        (wrapper.instance() as EditOAuthApp).handleConfirmModal();
```
hmhealey, line 98: Instead of using `as` here, you can declare the type of the `wrapper` when calling shallow. The difference here is pretty minor, but it does help if you're ever calling `wrapper.instance()` multiple times.
```javascript
const wrapper = shallow<EditOAuthApp>(
    <EditOAuthApp {...props}/>,
);
```

components/integrations/edit_oauth_app/edit_oauth_app.tsx
```diff
@@ -66,15 +54,15 @@ export default class EditOAuthApp extends React.PureComponent {
...
+    editOAuthApp = async (app: OAuthApp) => {
         this.newApp = app;
 
         if (this.props.oauthApp.id) {
             app.id = this.props.oauthApp.id;
         }
 
-        if (this.props.oauthApp.token) {
-            app.token = this.props.oauthApp.token;
+        if (this.props.oauthApp.client_secret) {
```
hmhealey, line 64: Since there's no `OAuthApp.token` field, I think we can just remove this block entirely instead of pointing this code at a different field.

hmhealey (changes requested):

hmhealey (approved):
Thanks for making those changes. Looks good to me

jgilliam17 (approved):
Thank you for your contribution @compiledsound 
Tested, looks good to merge.
- Verified edit OAuth app screen - as before.

hmhealey:
Thanks for keeping this up to date, @compiledsound! Sorry that it took us a bit to actually merge it. Hopefully there were no conflicts or anything to deal with

PR #7212, opened 2020-12-13
MM-20422 Migrate 'components/user_profile' module and associated tests to TypeScript

lmammino:
#### Summary

Migrate 'components/user_profile' module and associated tests to TypeScript

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/16152

Jira ticket: https://mattermost.atlassian.net/browse/MM-20422


lmammino:
There a couple of failures in the rendering (relative to default props) that I am not sure I understand. I'd love some guidance, please :)

components/overlay_trigger.tsx
```diff
@@ -9,7 +9,11 @@ import {Provider} from 'react-redux';
 
 import store from 'stores/redux_store.jsx';
 
-export {BaseOverlayTrigger};
+interface HideableOverlayTrigger extends BaseOverlayTrigger {
+    hide: () => void
```
hmhealey, line 13: Is there a way to just add the `hide` method to the `BaseOverlayTrigger` interface instead of defining something new? It would be nice to avoid the new name since people using this code don't really care about using the React Bootstrap interface or our custom one.

The reason we're hitting that though is because the type definitions are probably for a newer version of React Bootstrap than the outdated one we're stuck on currently.

components/post_view/post_header/post_header.tsx
```diff
@@ -151,6 +151,7 @@ export default class PostHeader extends React.PureComponent<Props> {
         } else if (isSystemMessage) {
             userProfile = (
                 <UserProfile
+                    userId=''
```
hmhealey, line 154: I think it makes sense to be explicit here by making this required. I'm surprised this was optional to begin with. :+1:

components/user_profile/user_profile.test.tsx
```diff
@@ -4,13 +4,15 @@
...
 import {shallow} from 'enzyme';
 
-import UserProfile from './user_profile';
+import {UserProfile as UserProfileType} from 'mattermost-redux/types/users';
+
+import UserProfile, {UserProfileProps} from './user_profile';
 
 describe('components/UserProfile', () => {
-    const baseProps = {
+    const baseProps: UserProfileProps = {
```
hmhealey, line 12: 0/5, but I don't think this should be necessary since we're specifying the type of `baseProps.user` below

components/user_profile/user_profile.tsx
```diff
@@ -1,52 +1,60 @@
...
-
-    static defaultProps = {
+export type UserProfileProps = {
+    displayName?: string,
+    isBusy?: boolean,
+    overwriteName?: React.ReactNode,
+    overwriteIcon?: React.ReactNode,
+    user?:UserProfileType,
+    userId: string,
+} & Partial<UserProfileDefaultProps>
```
hmhealey, line 22: Why extend a partial type here? I think we generally do this elsewhere by including all the fields in the `Props` type and then defining `defaultProps` as a `Partial<Props>` instead of giving it its own type.

components/user_profile/user_profile.tsx
```diff
@@ -61,13 +69,15 @@ export default class UserProfile extends PureComponent {
             userId,
         } = this.props;
 
-        let name;
-        if (displayUsername) {
+        let name: React.ReactNode;
+        if (user && displayUsername) {
```
hmhealey, line 73: To double check, is this a change in behaviour from before? I guess we probably only passed a user ID before when `displayUsername` was true.

components/user_profile/user_profile.tsx
```diff
@@ -61,13 +69,15 @@ export default class UserProfile extends PureComponent {
...
             name = `@${(user.username)}`;
         } else {
             name = overwriteName || displayName || '...';
         }
 
+        const ariaName: string = typeof name === 'string' ? name.toLowerCase() : '';
```
hmhealey, line 79: Same question here. I want to make sure we're not removing the `aria-label` in a case where we had one before.

hmhealey (changes requested):

components/overlay_trigger.tsx
lmammino: I have tried to work-around this one a little and, eventually, this was the only approach that sounded decent.
The issue is that the current typing definition for `react-bootstrap` does not expose the `hide()` method. I can try to see if there is any Typescript trip to _patch_ that definition and add the method there.

components/user_profile/user_profile.test.tsx
lmammino: I'll double-check this. I think you are right, but I want to make sure the type checker will be happy with removing this hint.

components/user_profile/user_profile.tsx
lmammino: Yeah, I wasn't 100% sure on how to do this. There seem to be many different practices around default props with TypeScript. I'll try to compare with other components in the same code base and make the necessary adjustments to follow the same approach. Thanks for catching this :)

components/user_profile/user_profile.tsx
lmammino: The problem is on the line below. We can't do `user.username` if `user` is `null` or `undefined`.  I think this might have been a potential source of runtime crash and the type checker highlights it.

I am not 100% sure this is actually changing the behaviour in any meaningful way. It might be good to have a more in-depth look from someone that has a bit more expertise on actually "using" the component in different contexts.

I did check all the available snapshots and this didn't seem to break any tested behavior.

components/user_profile/user_profile.tsx
lmammino: Yes, this is also an interesting one. I suspect this might have been another subtle bug that the type checker is helping us to identify.

The issue here is that `overwriteName` might be a `React.node` or a `string`. When it's a `React.node` also `name` will be a `React.node` and we can't really do `toLowerCase()` on it.

Here we might consider to explicitly expose an `ariaName` prop at the component level and enforce that to be a string. This way, even when a node is passed to `overwriteName` we can still control the aria label.

What do you think?

components/overlay_trigger.tsx
devinbinnie: @lmammino Have you tried loading the `react-bootstrap` types for an older version? I'm not sure if they exist and I seem to recall looking for them at one point, but if they're not officially provided it's possible that DefinitelyTyped has something provided.

components/user_profile/index.ts
```diff
@@ -5,14 +5,16 @@ import {connect} from 'react-redux';
...
-import UserProfile from './user_profile';
+import {GlobalState} from 'mattermost-redux/types/store';
+
+import UserProfile, {UserProfileProps} from './user_profile';
 
 function makeMapStateToProps() {
     const getDisplayName = makeGetDisplayName();
 
-    return (state, ownProps) => {
+    return (state: GlobalState, ownProps: UserProfileProps) => {
```
devinbinnie, line 15: Technically the `UserProfile` component shouldn't accept all of the fields as props that the `UserProfile` object provides. I think we should just provide our own `ownProps` type specific to this file in the connector. 2/5.

components/user_profile/index.ts
```diff
@@ -5,14 +5,16 @@ import {connect} from 'react-redux';
...
         return {
-            displayName: getDisplayName(state, ownProps.userId),
+            displayName: getDisplayName(state, ownProps.userId, true),
```
devinbinnie, line 17: If this value wasn't provided, a `boolean` here would default to `false` I believe, unless this is some special case.

components/user_profile/user_profile.tsx
devinbinnie: I don't think `user` should be undefined here, it should be a required prop. We can define that above and avoid this check.

devinbinnie (changes requested):
Thanks for the contribution @lmammino!

components/overlay_trigger.tsx
hmhealey: I'd assume they don't exist that far back. We're very far behind on them. What if you just import the provided type from them with a different name and call our version `BaseOverlayTrigger`?
```typescript
import {OverlayTrigger as OverlayTriggerInterface} from 'react-bootstrap';

interface BaseOverlayTrigger extends OverlayTriggerInterface {
    // same as before
}
```
I can't actually remember the reason that we export BaseOverlayTrigger separately from OverlayTrigger, but I'm just trying to avoid adding a third thing that's exported with all of them being the same as each other.

components/user_profile/user_profile.tsx
hmhealey: @devinbinnie Unfortunately, I think `user` can be undefined when this is used for system messages, so it can't be required.

@lmammino I think it's probably not a functional change looking more into this. If I had to guess, `props.userId` and `props.displayUsername` are always used together currently even though there's no way to represent that in the type definition. This is good to leave as-is then.

components/user_profile/user_profile.tsx
hmhealey: I'm looking into this a bit further, and it looks like we don't run into a case where `displayPopover` is true and `name` is not a string. Both this and the above case only get weird when we're using this component for System posts. Technically, the `overwriteName` isn't even used in that place, but we're passing a `FormattedMessage` for it anyway. In any other case, `overwriteName` is a string if it's provided.

I think we can just blame this on the component being old. We'll be re-writing it at some point as part of a larger effort to make better reusable components, so I think we can just address this then. As it is, this'll behave the same as before, although it won't cause a runtime error if it's used incorrectly later on.

components/user_profile/index.ts
hmhealey: Default values strike again!

https://github.com/mattermost/mattermost-redux/blob/master/src/selectors/entities/users.ts#L618

components/user_profile/index.ts
hmhealey: I think this is fine if it accepts all the props of `UserProfile` since `connect` returns a new component that accepts the props of `UserProfile` plus whatever we use here

components/user_profile/index.ts
devinbinnie: I see.

components/user_profile/index.ts
devinbinnie: The component doesn't though, it only accepts a subset of properties from `UserProfile`. There are a couple that line up, but not across the board.

components/overlay_trigger.tsx
lmammino: I ended up doing something like this:

```typescript
import {OverlayTrigger as OriginalOverlayTrigger, OverlayTriggerProps} from 'react-bootstrap';
// ...
export type BaseOverlayTrigger = OriginalOverlayTrigger & {
    hide: () => void
}
```

What do you think?

components/user_profile/index.ts
lmammino: Glad to see that my intuition was eventually correct. I did a bit of digging and ended up exactly on that line mentioned by @hmhealey and I assumed the default was `true`. Great to have confirmation. Thanks

components/user_profile/index.ts
lmammino: What about doing the following:

```typescript
type OwnProps = {
    userId: string,
}
```

and then replace `UserProfileProps` with `OwnProps`?

Any better name idea for `OwnProps`?

components/user_profile/user_profile.test.tsx
lmammino: I can confirm this was totally unnecessary. Removing it!

components/user_profile/user_profile.tsx
lmammino: Applied the approach you suggested and it actually makes a lot of sense. Much simpler and readable. Thanks :)

components/user_profile/index.ts
lmammino: Actually ended up doing the above. Let me know what you think. I am happy to change approach and names based on your feedback!

components/overlay_trigger.tsx
hmhealey: Yep, that works great!

hmhealey (approved):

components/user_profile/user_profile.tsx
```diff
@@ -1,52 +1,56 @@
...
-        isRHS: PropTypes.bool,
-        overwriteName: PropTypes.node,
-        overwriteIcon: PropTypes.node,
-        user: PropTypes.object,
-        userId: PropTypes.string,
-    };
-
-    static defaultProps = {
+export type UserProfileProps = {
+    displayName?: string,
```
devinbinnie, line 16: I think the linter will complain about using commas to separate fields instead of semicolons, as I'm pretty sure we just added that eslint rule recently, so you might want to change these to semicolons.

components/user_profile/index.ts
devinbinnie: Yep, that's perfect! Thanks @lmammino!

components/user_profile/user_profile.tsx
lmammino: I'll try to get those changes in by tomorrow

lmammino: Addressed :)

components/user_profile/user_profile.tsx
```diff
@@ -1,52 +1,56 @@
...
-    };
-
-    static defaultProps = {
+export type UserProfileProps = {
+    userId: string;
+    displayName?: string;
+    isBusy?: boolean;
+    overwriteName?: React.ReactNode;
+    overwriteIcon?: React.ReactNode;
+    user?:UserProfileType;
```
devinbinnie, line 21: Nit: spacing between `user?:` and `UserProfileType`

devinbinnie (approved):
LGTM! Thanks @lmammino!

components/user_profile/user_profile.tsx
hmhealey: I think ESLint might catch that now once the PR is updated

lmammino:
what's the next step on this one? Is there any action required on my side?

jgilliam17:
Thank you @lmammino 
Apologize for the delay in testing
Tested, looks good to merge.
- Verified profile popover still works as expected.
Seeing one small issue, where incoming webhook profile popover fails to open when viewed on the right hand side thread. Since it also reproduces on master and release, I will file that separately for follow up. 
Thanks again for your contribution! 🙂

jgilliam17 (approved):

devinbinnie:
@lmammino This is almost ready to merge, can you fix the lint errors? I'll merge when that's done.

```
/root/mattermost/mattermost-webapp/components/overlay_trigger.tsx
  13:21  error  Expected a semicolon  @typescript-eslint/member-delimiter-style

/root/mattermost/mattermost-webapp/components/user_profile/index.ts
  13:20  error  Expected a semicolon  @typescript-eslint/member-delimiter-style

/root/mattermost/mattermost-webapp/components/user_profile/user_profile.tsx
  21:10  error  Expected a space after the '?:'  @typescript-eslint/type-annotation-spacing
```

lmammino:
hey @devinbinnie, it should be updated now...

devinbinnie:
@lmammino Looks like there's still errors. We've made some changes to our eslint config recently, so you might need to sync your branch with master to get all of those errors fixed.

lmammino:
🤞

PR #7213, opened 2020-12-13
Custom Statuses Support

darkLord19:
Specs: https://mattermost.atlassian.net/wiki/spaces/CS/pages/983302547/User+Experience+and+Functional+Notes

Developed as a part of MM2020 Hackathon.

[Demo Video](https://streamable.com/59zzkh)


darkLord19:
Video Demo

esethna:
Hey @darkLord19 sweet!! This is an awesome start, wondering your thoughts on taking this all the way to meet the design spec? 

A few high level things that come to mind:

1) Does this work with the "clear after" time? Were there any server changes needed?
2) Implementing the emoji in various places across the UI (as defined [here](https://mattermost.atlassian.net/wiki/spaces/CS/pages/983302547/User+Experience+and+Functional+Notes#Where-the-custom-status-appears-after-it-is-set))
3) Refining the UI a bit to match the figma designs (@asaadmahmood may be able to help here)
4) Adding the status suggestion list and recent statuses to pick from
5) The mobile react native side of this

esethna:
@jespino take a look at this!

darkLord19:
Hey @esethna, yes I'm planning to get it to as close ss design specs.

> 1. Does this work with the "clear after" time? Were there any server changes needed?
Not yet. Server side changes are needed. I'll be using similar approach as used for dnd status clear time i.e having a go routine which checks for expired statuses at a specified interval, clears them and brodcasts new objects to everyone.
> 2. Implementing the emoji in various places across the UI (as defined [here](https://mattermost.atlassian.net/wiki/spaces/CS/pages/983302547/User+Experience+and+Functional+Notes#Where-the-custom-status-appears-after-it-is-set))
Planning on adding thisas well.
> 3. Refining the UI a bit to match the figma designs (@asaadmahmood may be able to help here)
I think that could be done afterwards on top of this PR because I saw a lot of UI changes in figma design in profile popover that would make sense as a seperate PR.
> 4. Adding the status suggestion list and recent statuses to pick from
WIP
> 5. The mobile react native side of this
I'll try this but can't say for sure because I'm not good at react. I can maybe get functionality working but UI will most probably need tweaks from Assad.

esethna:
@jespino any thoughts on the approach outlined above for clear time? ^

components/generic_modal.scss
```diff
@@ -9,13 +11,6 @@
         max-height: 100%;
         padding: 0;
 
-        .form-control {
```
jespino, line 12: I don't know if this is going to break any other modal out there.

components/generic_modal.scss
```diff
@@ -118,3 +114,98 @@
         }
     }
 }
+
+.StatusModal {
```
jespino, line 118: Can we put this css aside the new component as `components/custom_status_input_modal/custom_status_input_modal.scss` and import it directly from the modal typescript file?

components/legacy_sidebar/header/sidebar_header_dropdown_button.jsx
```diff
@@ -66,6 +76,22 @@ export default class SidebarHeaderDropdownButton extends React.PureComponent {
                         className='user__name'
                     >
                         {'@' + this.props.currentUser.username}
+                        {statusEmoji &&
```
jespino, line 79: We shouldn't need the whole Markdown component here to render the emoji. I think we should have something simpler, and if it is not there, we should create it.

components/profile_popover/index.js
```diff
@@ -24,6 +24,7 @@ import {areTimezonesEnabledAndSupported} from 'selectors/general';
 import {getSelectedPost, getRhsState} from 'selectors/rhs';
 
 import ProfilePopover from './profile_popover.jsx';
+import { getCustomStatus } from 'mattermost-redux/actions/users';
```
jespino, line 27: This is imported but not used, I think

components/profile_popover/profile_popover.jsx
```diff
@@ -383,6 +385,40 @@ class ProfilePopover extends React.PureComponent {
...
+            }
+            dataContent.push(
+                <div
+                    key='user-custom-status'
+                >
+                    <FormattedMessage
+                        id='user_profile.custom_status'
+                        defaultMessage='Status: '
+                    />
+                    {statusEmoji &&
```
jespino, line 402: Same problem here, the whole Markdown component only for rendering an emoji :(

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.scss
```diff
@@ -1,12 +1,6 @@
 @charset 'UTF-8';
 
 .GenericModal {
```
jespino, line 3: All this .GenericModal shouldn't be here unless it is inside something like `.SidebarWhatsNewModal` class or something like that (not related to your changes, but something that surprise me).

jespino: @devinbinnie you added this changes. This are changing the `.GenericModal` and the `.modal` classes globally from this place, I think this shouldn't happen. I would expect to have a parent class `.SidebarWhatsNewModal` or a `.SidebarWhatsNewModal.GenericModal` and `.SidebarWhatsNewModal.modal`.

jespino:
The clearing of the custom status shouldn't happen that way, imagine that you have 2000 users (that is not even a big quantity of users), every loop there, is going to full scan the users table because we don't have indexes on that information.

My proposal for this was to set the "expire date" in the information, and that way you only need to do it at client side, if the custom status is expired is becuase you don't have custom status. This relays on the correct client configuration about the time, but like other features in mattermost. We have to make this information aware of the timezone and all that stuff (probably using UTC time always, or something like that).

components/legacy_sidebar/header/sidebar_header_dropdown_button.jsx
darkLord19: I checked codebase but didn't find any reference of rendering emoji without markdown component. I found some old commits where emoji rendering was added for some components and they used markdown so I did the same.

darkLord19:
I read through your approach in specs but I am not sure how to handle status expiring case for already loaded user states?

components/sidebar/sidebar_whats_new_modal/sidebar_whats_new_modal.scss
devinbinnie: You're correct. I think it only works because we never have both components loaded at the same time. But yes the modal should be using something like `.GenericModal.SidebarWhatsNewModal`

chetanyakan:
@darkLord19 You can read more about this in point 3 of the [Confluence doc](https://mattermost.atlassian.net/wiki/spaces/CS/pages/1096810497/Development+Spec?focusedCommentId=1153204226#comment-1153204226).

But the gist is to -
1. Use an internal state - `expired: true/false` to render `null` if required.
2. Use a `setTimeout` in the react component to set this internal state when the expiry time is reached.
3. Update this timeout (unset and set) when someone's status expiry changes.

***

@jespino On second thought, If there are 2000 users, then there can be about 2000 `setTimeout` functions running on each client. How would this impact the `mattermost-webapp` performance?

Would a better method be to run a single `setInterval` per client?
The interval duration can be 1-minute and with each interval, you can clear the custom statuses for all the users, as needed.
This way, the custom status expiry can only be off by 1 minute at most.

We can also tweak this duration if required. Thoughts?

jespino:
@chetanyakan that is not necesarilly true, the idea is to have a setTimeout for each current representation of the custom status in the screen in any time. That means, probably 1 per visible post and no much more than that. And actually, that is not going to be created for already expired custom status and not-set custom statuses, because that is handled through websocket events.

Other option to mitigate data is share the "setTimeout" between the whole system using redux to handled that setTimeout, and allow any component to set add its own timeouts and have a global setTimeout that is based on the list of timeouts registered.

Said that, still I think is not a huge deal the initial approach, and the second proposed approach, can be added later if we see problems with the first one.

What is really important here is to be sure the setTimeouts are properly handled on creation/destroy of the components to avoid any posibility of setTimeout leaks.

chetanyakan:
@jespino I see your point. We can unset the `setTimeout` in `componentWillUnmount` or the `useEffect` hook return method. This way we only maintain a few of these timeout fns at a time.

But, I still think that the interval approach could be simpler to implement and easier to maintain (as we don't need to be concerned with the potential memory leak from so many `setTimeouts`).

- Keep all users' custom statuses, `expiresAt` and `expired: true/false` in the Redux store.
- After each interval duration, update all users' `expired` state, as required.
- Use the value from the redux state in the custom status component.
As the redux state updates, the component will re-render.

What do you think?

components/legacy_sidebar/header/sidebar_header_dropdown_button.jsx
chetanyakan: @darkLord19 Maybe these utils would be useful?
- https://github.com/mattermost/mattermost-webapp/blob/371c302e143727ce6e2c93383e48815ea7d7ec16/utils/text_formatting.tsx#L839
- https://github.com/mattermost/mattermost-webapp/blob/371c302e143727ce6e2c93383e48815ea7d7ec16/utils/emoticons.tsx#L90

darkLord19:
If we do want interval approach then I think keeping it in the backend would be more performant even if we don't have index on props field on db. We can make use of jsonb type of postgresql and returning clause and can get and remove expired statuses in a single database query.

esethna:
Closing this PR for now as we have a partner who will be submitting PRs for custom statuses soon based off this initial work. Thanks so much for getting us a head start @darkLord19!

PR #7217, opened 2020-12-14
New cross team quick switch modal.

crspeller:
#### Summary
Redesigns the quick switch modal to be cross team. Makes many other improvements.

- Switching channels across teams is now possible
- More responsive
- All local channel switching
- Fuzzy searching (eg. tonsq will find Town Square)
- New clean design!
- WIP: Switching to channels you are not in

Thanks to @esethna for helping strengthen out functionality! Thanks to the design team for the new design.

Sorting Logic:
1. Get the top 20 results from the fuzzy finder. The algorithm it uses is https://en.wikipedia.org/wiki/Bitap_algorithm Results that do not score high enough to meet a threshold are removed, so there may be less than 20. They come sorted by the fuzzy finder.
2. Filter out Archived channels and place them separate.
3. If a result contains exact matches to terms entered, sort it first.
4. Then for "similar" scores given by the fizzy finder:
    1. Sort by starts with
    2. Rank GMs lower than any other channel type
    3. Sort by last viewed. 

#### Ticket Link
No ticket, this was created as part of the December 2020 hackathon. 

#### Related Pull Requests
Feature flag: https://github.com/mattermost/mattermost-server/pull/16545

#### Screenshots
Change channels across teams:
image

Fuzzy Searching:
image


esethna:
Thanks @crspeller! Stoked about cross team switching! Can you help write out the logic you ended up with for sorting results?

crspeller:
@esethna Updated the description with the sorting logic.

esethna:
@crspeller curious what are next steps to get this PR in from your end? Is the code fully functional or is it still a WIP?

I'd like to get some more eyes on this from UX and QA:

1) I think we should have @matthewbirtch review the UX/sorting logic compared to the original spec we had (current behaviour) https://mattermost.atlassian.net/browse/MM-22854. We will likely need to merge this and test on daily to get a good feel for the sorting.
2) @matthewbirtch we will also want to review the UI. I think this is using our old channel switcher designs so not sure if we want to fully revert the UI changes or modify some parts. 
3) @jgilliam17 can we do a QA pass on this (similar to how we did a brief pass on the logic with Sudheers channel switcher PRs), then we can do a more thorough pass on it once it's on daily with real data to test against.

esethna:
@matthewbirtch FYI this was from a the hackathon last month

matthewbirtch:
@crspeller this doesn't have all the info and styling from the current Channel Switcher. It's a very short, wide modal and doesn't have a title or display any channels in its default state. For clarity, it should look more like this:
image

I see this is in 'Draft' state, so what part of this should I review? Just the sorting?

crspeller:
@esethna From my perspective, I have a few more things to fix up before I get it dev reviewed. (hence the draft state). 
I have honestly forgotten exactly what I have to do. It would be helpful to get a review on the current behavior and I can make adjustments. 
The one thing that I remember is missing is switching to channels you are not in. We might just consider removing that as a feature, since it's quite laggy given it's need to go to the server and not that useful.

esethna:
@crspeller to verify, does this use strictly local results? What if you don't have a user in your sidebar yet? 

I don't think we should remove the functionality that returns server results - it's important for channel discoverability and opening DMs. We should just append the server results on the end in my mind.

The intention of this ticket https://mattermost.atlassian.net/browse/MM-22854 was to get local results first and immediately (sorted appropriately) then append server returned results

angeloskyratzakos:
Hey all, I will remove the tag `Setup Cloud Test Server` for maintenance reasons. Feel free to re-add it when you need it. Thank you

PR #7231, opened 2020-12-15
MM-23267 Fixed autocomplete bug

shFrank:
fixed the bug that autocomplete would disappear if there are spaces between the channel name

The problem is caused by the regex because the previous version did not consider the situation when there are spaces between the channel name.
ex: 
before: when typing "IN: Town Squ" at the search box, the autocomplete would disappear
Now:  when typing "IN: Town Squ" at the search box, the autocomplete would still be there

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23267
Fixes https://github.com/mattermost/mattermost-server/issues/16242

#### Related Pull Requests

### DIFF
--- a/components/suggestion/search_channel_provider.jsx
+++ b/components/suggestion/search_channel_provider.jsx
@@ -34,7 +34,7 @@ export default class SearchChannelProvider extends Provider {
     }
 
     handlePretextChanged(pretext, resultsCallback) {
-        const captured = (/\b(?:in|channel):\s*(\S*)$/i).exec(pretext.toLowerCase());
+        const captured = (/\b(?:in|channel):([\s\S]*)$/i).exec(pretext.toLowerCase());
         if (captured) {
             let channelPrefix = captured[1];
             const isAtSearch = channelPrefix.startsWith('@');


hmhealey (changes requested):
Thanks for the PR, @shFrank. 

I'm seeing an issue here that we used to hit with the `~channel` autocomplete in the post textbox. Once you use `in:` and fill in a channel name, it will keep searching for autocomplete results as you continue to type more into the textbox.

https://user-images.githubusercontent.com/3277310/102640454-7b6a6680-4128-11eb-9e6d-04ee3ba63b82.mov

You'll want to look at the `ChannelMentionProvider` to see how we solved it there. I think it was by tracking the previous captured text and whether or not that returned results so we could avoid sending a request if we're just adding to something that doesn't return results.

Also, if you watch closely in the video I posted above, it seems to stop returning autocomplete results once I've typed `in: display ` with a space at the end until I get to `in: display na`. I'm not sure if that is a server or clientside issue though.

jfrerich:
@shFrank, thanks for the PR!  Could you address the feedback from @hmhealey? Feel free to ping him with any questions.

jfrerich:
Hi @shFrank!  We would love to get this merged or open the issue back up for another contributor.  Please let us know :)

hmhealey:
Closing this to let others have a shot at it.

Thanks again for your work on this, @shFrank! If you'd like to work on this again, let us know, and we can open this back up. 🙂

PR #7232, opened 2020-12-16
MM-20449 Migrate 'components/integrations/installed_incoming_webhooks' module and associated tests to TypeScript

nikkinagar:
#### Summary
This pull request migrates 'components/integrations/installed_incoming_webhooks' module and associated tests to TypeScript

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16125
JIRA Ticket: https://mattermost.atlassian.net/browse/MM-20449


components/integrations/installed_incoming_webhooks/index.ts
```diff
@@ -0,0 +1,59 @@
...
+        channels: getAllChannels(state),
+        users: getUsers(state),
+        canManageOthersWebhooks,
+        enableIncomingWebhooks,
+    };
+}
+
+function mapDispatchToProps(dispatch: Dispatch<GenericAction>) {
+    return {
+        actions: bindActionCreators<ActionCreatorsMapObject<any>, Actions>({
```
nikkinagar, line 52: I needed to keep the type here as `any` since `loadIncomingHooksAndProfilesForTeam` has not been migrated to TypeScript

components/integrations/installed_incoming_webhooks/index.ts
```diff
@@ -27,20 +35,20 @@ function mapStateToProps(state) {
     const enableIncomingWebhooks = config.EnableIncomingWebhooks === 'true';
 
     return {
+        teamId,
```
devinbinnie, line 38: Should we just remove this prop altogether?

devinbinnie (approved):
Thanks @nikkinagar, this looks great!
Just one non-blocking comment.

components/integrations/installed_incoming_webhooks/index.ts
nikkinagar: Thanks @devinbinnie. I did remove this in the first push but I wasn't sure if that was correct. I think I forgot to change it back to `teamId` in my last commit. If it's confirmed that it's okay to use `team.id` instead of `teamId` in `installed_incoming_webhooks.tsx` I can remove it from here.

nevyangelova (approved):
Thanks! LGTM :)

components/integrations/installed_incoming_webhooks/index.ts
devinbinnie: Yeah I think it's fine, we'll have QA do a pass over to make sure nothing broke.

nikkinagar: Sounds good, addressed.

DHaussermann (approved):
Tested and passed
- Create, Read, Update and Delete test for all fields on incoming webhooks
- No functional issues or JS errors in console found
- Tested in browser and desktop

LGTM!
Thanks @nikkinagar  for this enhancement!

nikkinagar:
./update-branch

PR #7233, opened 2020-12-16
[MM-31403] Ensure other permissions except billing getting deleted

hahmadia:
#### Summary

If the billing section is not found a -1 is returned but we still delete that which deletes the last permission section in the list. We need to check it's greater than -1 to make sure we aren't deleting other permissions

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31403


sbishel (approved):

marianunez (approved):

furqanmlk (approved):
It is working as expected and all test case are passed.

PR #7235, opened 2020-12-16
[MM-31358] Clicking 'Jump to recent' in an archived channel shouldn't redirect to town square

nickmisasi:
#### Summary
Previously when jumping into an archived channel, if you clicked the "Jump to recents" toast, you'd be redirected out of the channel (provided you were significantly in the message history). 

If you jump into a channel and are significantly into the message history, instead of scrolling down, the webapp will simply reload the page based on the localStorage last visited page value, which will result with the user at the most recent messages.

Since we were not storing archived channels in the last visited page value, when you clicked "Jump to recents" in an archived channel, you wouldn't be redirected to the same archived channel, but to some other one (usually town square).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31358

#### Related Pull Requests
N/A

#### Screenshots
N/A


devinbinnie (approved):

hmhealey (approved):

hmhealey:
That makes sense to me. Hopefully we'll eventually get proper channel history instead of tracking the penultimate channel specifically which should make these sorts of things easier to prevent in the future.

jgilliam17 (approved):
Thank you @nickmisasi 
Tested, looks good to merge.
- Verified user is taken to the bottom of the archived channel when clicking on history toast 
 after using `jump` from search results.

PR #4692, opened 2020-01-17
MM-20349 Improve typing performance

sudheerDev:
#### Summary
 * Fix couple of setState calls being called all the time when typing
    This is similar to https://github.com/mattermost/mattermost-webapp/pull/4664.
 * Delay setting draft to localstorage to improve performance
      * Setting of draft is triggering store update invalidating the cache for memoization of selectors
      * Delaying this by a little bit to to distribute the code execution so few more frames can be squeezed in. This mainly helps users with busy JS thread because of sidebar profiles

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20349

#### Screenshots
Before:
Jan-18-2020 00-30-36
After:
Jan-18-2020 00-45-38

For 5.19v branch there is a regression with memoization which needs to be fixed by cherry-picking https://github.com/mattermost/mattermost-redux/pull/1011

PR #7242, opened 2020-12-17
MM-31466 Performance investigation part two

hmhealey:
Round two of the performance improvements including in order from most important to least:
1. We throttled updates to the comment draft when typing in the RHS a la https://github.com/mattermost/mattermost-webapp/pull/4692 which provides a huge improvement when typing rapidly in the RHS. (https://github.com/mattermost/mattermost-webapp/commit/b5b0a3ffa728506b9c021083e14167757c356214)
2. We reduced lag with the new sidebar by adding some slight tweaks when we read the collapsed state of categories from `state.storage.storage`. That part of state changes whenever the draft is updated, and it made some of the heavier sidebar selectors recalculate very often. It's much cheaper for us to do a quick shallow check on a map of categories to their collapsed state than it is to recalculate `getChannelsInCategoryOrder` every time a post draft changes (https://github.com/mattermost/mattermost-webapp/commit/f3ba0a3edf4e1df2cc6d3fd7de5e36f76b4b3f0f).
3. We added a `getCategoriesForCurrentTeam` selector that can be reused for every `SidebarChannel` so that each one isn't calculating and memoizing that list of categories separately. (https://github.com/mattermost/mattermost-webapp/commit/602d697549fcdd0382af99595d163e5b56ad61fc and https://github.com/mattermost/mattermost-webapp/commit/76a8d0ee76eed3d525b4737582b9e1a4691bfa43)
4. We now pass a constant options object into the Textbox's `addLastViewAtToProfiles` to avoid it recalculating unnecessarily (https://github.com/mattermost/mattermost-webapp/commit/d758665696feb27fb39ffceda3e7def170db0ac9)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31466

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/4692
https://github.com/mattermost/mattermost-webapp/pull/7207

### DIFF
--- a/components/create_comment/create_comment.jsx
+++ b/components/create_comment/create_comment.jsx
@@ -297,6 +297,12 @@ class CreateComment extends React.PureComponent {
         this.props.resetCreatePostRequest();
         document.removeEventListener('paste', this.pasteHandler);
         document.removeEventListener('keydown', this.focusTextboxIfNecessary);
+
+        if (this.saveDraftFrame) {
+            cancelAnimationFrame(this.saveDraftFrame);
+
+            this.props.onUpdateCommentDraft(this.state.draft);
+        }
     }
 
     componentDidUpdate(prevProps, prevState) {
@@ -697,7 +703,12 @@ class CreateComment extends React.PureComponent {
 
         const {draft} = this.state;
         const updatedDraft = {...draft, message};
-        this.props.onUpdateCommentDraft(updatedDraft);
+
+        cancelAnimationFrame(this.saveDraftFrame);
+        this.saveDraftFrame = requestAnimationFrame(() => {
+            this.props.onUpdateCommentDraft(updatedDraft);
+        });
+
         this.setState({draft: updatedDraft, serverError}, () => {
             this.scrollToBottom();
         });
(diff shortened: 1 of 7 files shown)


devinbinnie (approved):

components/create_comment/create_comment.jsx
```diff
@@ -697,7 +703,12 @@ class CreateComment extends React.PureComponent {
 
         const {draft} = this.state;
         const updatedDraft = {...draft, message};
-        this.props.onUpdateCommentDraft(updatedDraft);
+
+        cancelAnimationFrame(this.saveDraftFrame);
+        this.saveDraftFrame = requestAnimationFrame(() => {
```
crspeller, line 708: Why `requestAnimationFrame` and not `setTimeout`?

crspeller (approved):
Awesome work! 
Also verified that these fixes are effective in eliminating the lag on my local instance. 

Just one question probably for my information.

lindalumitchell (approved):
Not seeing lag in any input box or functionality, particularly in reply RHS. Smoke tested sidebar with no issues found. Tested new sidebar off and on, custom categories, favorites, DMs, at-mentions, unread filter, etc.

amyblais:
@crspeller Do you think I can auto-merge this PR or should I wait for this question to be addressed https://github.com/mattermost/mattermost-webapp/pull/7242#pullrequestreview-555029474? It seems Harrison is offline now.

amyblais:
Please also cherry-pick this to ``cloud``.

components/create_comment/create_comment.jsx
hmhealey: We discussed different ways of throttling this, but we settled on requestAnimationFrame since it only throttles when the app is lagging and it matches what we're doing in the post textbox already. If we weren't using that, we'd probably go for something like lodash's throttle/debounce methods that we use elsewhere

PR #7246, opened 2020-12-18
MM-22818 - Removing status from post profile picture

asaadmahmood:
#### Summary
MM-22818 - Removing status from post profile picture

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22818

#### Screenshots
Screenshot 2020-12-18 at 5 51 33 PM


hmhealey:
What's the motivation for changing this? In particular, I'm curious why we'd remove this here while we've been adding it in more places recently such as the at-mention autocomplete and channel switcher. I believe it's also used or going to be used in some plugins. The profile picture is something we'll want to extract out as a common component, so we'd generally want it to behave the same everywhere.

asaadmahmood:
@esethna @andrewbrown00 Might know the answer to that better.

andrewbrown00:
> What's the motivation for changing this? In particular, I'm curious why we'd remove this here while we've been adding it in more places recently such as the at-mention autocomplete and channel switcher. I believe it's also used or going to be used in some plugins. The profile picture is something we'll want to extract out as a common component, so we'd generally want it to behave the same everywhere.
> 
> And if we're doing this, I'd also want to look at what other effects this might have on the code. Not showing statuses on posts means we could remove some backing infrastructure that syncs user statuses that are visible in the centre channel, so it's not as simple as just removing the UI elements that show the statuses.

Hey @hmhealey, the main reason is the status doesn't provide much value in the context of consuming content in channels. There is a more detailed breakdown in the the [Jira ticket](https://mattermost.atlassian.net/browse/MM-22818). Also, we are going to let this soak on **Daily** for a bit and test and gather feedback before it goes live. You are correct, we have been adding it in other places that will add value and the proper context to inform users of another users status.

andrewbrown00 (approved):
Looks good to me, thanks @asaadmahmood

image

image

esethna (changes requested):
@asaadmahmood @andrewbrown00 I don't think we can remove this unless we add the status to the profile pop-over, otherwise users have to way to get the status of users in a channel.

andrewbrown00:
> @asaadmahmood @andrewbrown00 I don't think we can remove this unless we add the status to the profile pop-over, otherwise users have to way to get the status of users in a channel.

I agree @esethna, good catch. @asaadmahmood can we add the status in the **profile popover** as part of this PR?

asaadmahmood:
@andrewbrown00 I can have a look at that.

asaadmahmood:
Added the status.
Screenshot 2021-01-20 at 3 26 17 PM

asaadmahmood:
@andrewbrown00 @esethna

andrewbrown00:
Thanks @asaadmahmood, something broke in the PR, this is what I'm seeing

image

esethna (approved):
Thanks Asaad.

@andrewbrown00 I think we should get this in at the beginning of a cloud release cycle so it has time to soak on community-daily and see if anyone has negative feedback. Also can we write up a few brief bullet points on justification for why we are doing this in case we get feedback from community

ogi-m:
@asaadmahmood @andrewbrown00  Should the status outline on the user popover match the theme color?
user_status1 user_status2

asaadmahmood:
@ogi-m It should.

asaadmahmood:
@ogi-m So since that's the away color it is probably matching the theme color.
Compare the colors with what you see in the icon in the LHS header, they should be the same.

asaadmahmood:
The white background is incorrect though, which I will fix.

ogi-m:
@asaadmahmood Yeah I was referring to the white background. Thanks 👍

asaadmahmood:
Should be fixed @ogi-m

andrewbrown00:
> Thanks Asaad.
> 
> @andrewbrown00 I think we should get this in at the beginning of a cloud release cycle so it has time to soak on community-daily and see if anyone has negative feedback. Also can we write up a few brief bullet points on justification for why we are doing this in case we get feedback from community (some would see this as us removing a feature, while others may see this as clean up and simplification)

Agreed @esethna we should let this soak. Here's a list justification for deciding to remove the status from the center channel:
- A users status doesn't provide value in the context of 'consuming' content in channels
- We've made status more visible in the UI and introduced status in more contextual places, such as the channel switcher modal
- Removing the status in the center channels reduces the visual noise / additional distraction in the UI. Statuses can quickly be discovered by clicking on a users avatar which opens their profile popover. From the popover, they can also quickly start a new or open an existing conversation.
- Additionally, we can introduce a 'display setting' to allow users to choose whether they want to see users status in channels

asaadmahmood:
@ogi-m Can you test this again?

esethna:
@asaadmahmood are we adding the display setting for this as per Andrew's comment above?

ogi-m:
Thanks Asaad, verified the background issue on the user popover is fixed 👍🏻 

I'll approve after dev review if there are no further changes.

devinbinnie (approved):

devinbinnie:
Code looks good, but I'm also wondering why we plan to just remove this in general. Personally I like being able to view someone's status on the chat icons. If it's becoming clutter for some people, couldn't we just have an option to turn it on/off as the user sees fit? I'm not a fan of removing functionality that could be useful to some people.

asaadmahmood:
@devinbinnie Agree, we can consider this a test, maybe if a lot of people miss it, we can make it an setting.

devinbinnie:
> @devinbinnie Agreed, we can consider this a test, maybe if a lot of people miss it, we can make it an setting.

Any reason we can't add a setting now? I'd personally miss this feature. 3/5

hmhealey:
Sorry about slowing this down by not reviewing this earlier. I should've queued up an actual meeting to discuss this since I'm still concerned with the effects this has by removing the status from so many places. It's a big change to the infrastructure as well as a big change to UX, so I'd like to have had some more dev input on removing statuses in the first place.

I can queue up a meeting for this if I can find time where we're all available.

hmhealey:
Okay, so I've read back through all this to make sure I'm caught up on the whole discussion.

My main concern was that we'd now be loading a bunch of status information that we no longer need because the posts don't show the status icon, but we still need that now that the status is on the profile popover. Regardless, that sort of thing does need to be considered when making changes like that, but the ticket needs earlier dev input to have that happen.

Regarding the removing of status icon as a feature, I agree with Devin that it's useful to have, so I'm 2/5 that I wouldn't want to see it go myself. Adding it to the profile popover does mitigate that, but it does seem to be something that could be configurable. I am confused a bit about where we're at on adding that setting though since @esethna and @andrewbrown00 sounded like they both wanted it, but they previously approved the PR, so I'm not sure where their approval stands on this.

lieut-data (changes requested):
I think we need to make this configurable: removing the status icon will absolutely break my workflow from day 1.

hmhealey:
I tried to find a time for us to meet, but I wasn't able to find a slot that worked for everyone involved. That said, I'm not really concerned about the effects on the status loading any more like I mentioned in my last comment, so I don't think we need to meet any more.

I do agree that having a setting for this is a really good idea though.

esethna (changes requested):
Thanks Harrison, dismissing my stale review, we should have an account setting for this in Account Setting > Display below teammate name display.

"Show online availability on profile images"

[ ] on
[ x ] off

When enabled, online availability is displayed on profile images in the message list. 

@asaadmahmood

hmhealey:
Should that setting default to true or false? I'd lean towards leaving it on as-is

nevyangelova:
Agreed on all points, and I think it says in the ticket to have the setting, but they want it default off? I am more leaning towards leaving it on and maybe have a tooltip stating that you can remove the status if you want, otherwise how would the users know about this possibility?

asaadmahmood:
Cool @esethna I can work on this.

asaadmahmood:
Though do we want to improve the statement:
"Show online availability on profile images"

We show all statuses there, not just online availability right?

And then this is not hiding all statuses on the profile picture, just in the center channel, we should make that clear as well. (I know its mentioned in the description, but do we want to mention it in the section title as well somewhere)?

hmhealey:
I think something like "Show user status on posts" or "Show online status on posts" might work better. "On posts" clarifies that this only affects the post, and I think "status" is clearer than "availability".

asaadmahmood:
@esethna @andrewbrown00  Please test this out.

esethna:
@asaadmahmood please use the text I suggested above for the setting title. We are trying to narrow in on our terminology and the descriptive text clarifies the part about it applying to the message list.

In terms of the default I think we need to make this as low impact as possible to the current experience. If possible, I'd propose we leave it on by default for existing users (so we don't change their experience), but have it off for new users 

This is a polarizing change, so needs to be considered from a change management perspective.  cc// @andrewbrown00

asaadmahmood:
@esethna Updated the text.
I don't know how to keep it preserved for current users, will ask someone about it.

asaadmahmood:
@esethna Spoke to @hmhealey There's no way for us to keep it enabled for current users and disabled for new users from the preferences panel unfortunately. 
**What do we want to do then?**

esethna:
@andrewbrown00 thoughts here? Options I see, but perhaps this is a good change to do user testing and validation as per UX teams initiatives:

1. Leave enabled but the feature will likely not get used as it lacks discoverability
2. Disable it and soak on cloud and test servers for feedback
3. Kill this PR and the feature as per the FY22 initiative to reduce bloat and narrow our scope
3. Other ways UX plans to validate and iterate that we can utilize here?

andrewbrown00:
> @andrewbrown00 thoughts here? Options I see, but perhaps this is a good change to do user testing and validation as per UX teams initiatives:
> 
> 1. Leave enabled but the feature will likely not get used as it lacks discoverability
> 2. Disable it and soak on cloud and test servers for feedback
> 3. Kill this PR and the feature as per the FY22 initiative to reduce bloat and narrow our scope
> 4. Other ways UX plans to validate and iterate that we can utilize here?

Hey @esethna, agree this would be good to validate, but based on priorities we don't have the cycles to test it. Could disable it for all users and provide an option in the preferences to enable it? Wasn't sure if that was the intention with `option 2`? Then let it soak, and see what type of feedback we get and how many users think it's a bug.

cc @matthewbirtch

esethna:
@asaadmahmood I'd propose we leave it enabled by default for now to minimize the change from today's behaviour. Might be something to consider changing the default if we release accompanying UI changes like the UX overhaul planned later.

asaadmahmood:
Done @esethna

esethna (approved):
Looks good to me as long as the default for the setting is set to enabled (no changes to existing behaviour)

@asaadmahmood can you open a ticket for doing the accompanying mobile changes to respect this setting?

asaadmahmood:
@nevyangelova @devinbinnie Please test.

devinbinnie (approved):

asaadmahmood:
@esethna Waiting for another dev review.

nevyangelova (approved):

asaadmahmood:
@ogi-m Can you please give this a look?

esethna:
@asaadmahmood can you help update the JIRA ticket with what was actually done in the PR (ie where the status icon is removed from or added, account setting, etc) to help QA in testing? cc// @ogi-m

asaadmahmood:
Done @esethna

ogi-m:
Tested and passed
- Setting added to user display settings
- On by default
- When disabled, availability status is hidden on avatars in centre channel and RHS
- Status still shown in other places as expected
- Status added to the profile popover (regardless of setting)

ogi-m (approved):

esethna:
Thanks all, can someone help merge?

esethna:
@amyblais I think merging is blocked on the release note label?

amyblais:
@asaadmahmood Can you help add release notes with these instructions https://github.com/mattermost/chewbacca#applying-a-release-note?

asaadmahmood:
```release-note
NONE
```

asaadmahmood:
```release-note
Added an option to hide the status icon on the user profile in the center channel
```

PR #7252, opened 2020-12-20
[GH-16128] - Migrate 'components/integrations/bots' module and associated tests to TypeScript

lucievr:
#### Summary
- Migrated `components/integrations/bots` module and test files to TypeScript
- Renamed files to have TypeScript extensions and updated incorrect imports
- Created a test helper method for mocking a user access token in `utils/test_helper.ts`
- Check-types check returned no errors
- Fixed a few other things (See comments)

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/16128


components/integrations/bots/bot.test.tsx
```diff
@@ -3,30 +3,36 @@
...
-
 import TestHelper from 'tests/helpers/client-test-helper';
+import {TestHelper as UtilsTestHelper} from 'utils/test_helper';
 
-import Bot from './bot.jsx';
+import Bot from './bot';
 
 describe('components/integrations/bots/Bot', () => {
-    const team = TestHelper.fakeTeam();
+    const team = UtilsTestHelper.getTeamMock();
```
lucievr, line 15: This fakeTeam function doesn't generate an object that matches all the required properties of the Team type so I used this other method instead. The same applies to some other mocked objects here.

components/integrations/bots/bot.tsx
```diff
@@ -194,7 +209,7 @@ export default class Bot extends React.PureComponent {
             if (token.is_active) {
                 activeLink = (
                     <a
-                        name={token.id + '_deactivate'}
+                        id={token.id + '_deactivate'}
```
lucievr, line 212: Name attributes on anchor elements are obsolete. Ids can be used though.

components/integrations/bots/bots.test.tsx
```diff
@@ -87,21 +87,17 @@ describe('components/integrations/bots/Bots', () => {
...
+        const owner = TestHelper.getUserMock({id: bot1.owner_id});
 
-        const user = {
-            id: bot1.user_id,
-        };
+        const user = TestHelper.getUserMock({id: bot1.user_id});
 
         const passedTokens = {
-            id: 'token',
+            id: TestHelper.getUserAccessTokenMock(),
```
lucievr, line 100: This token needs to be a UserAccessToken type object. Created a helper function for mocking this.

components/integrations/bots/bots.tsx
```diff
@@ -1,98 +1,108 @@
...
             loading: true,
         };
     }
 
-    componentDidMount() {
+    public componentDidMount(): void {
         this.props.actions.loadBots(
             Constants.Integrations.START_PAGE_NUM,
-            Constants.Integrations.PAGE_SIZE,
+            Constants.Integrations.BOTS_PER_PAGE_DEFAULT,
```
lucievr, line 105: loadBots function expects a number as its second arg (PAGE_SIZE is a string of '1000'). I copied over the constant BOTS_PER_PAGE_DEFAULT from mattermost-redux to utils/constants.

components/integrations/bots/bot.tsx
```diff
@@ -31,62 +34,73 @@ export function matchesFilter(bot, filter, owner) {
...
+    *  Only used for routing since backstage is team based.
+    */
+    team: Team;
+}
+
+type State = {
+    confirmingId: string;
+    creatingTokenState: string;
+    token: UserAccessToken | Record<string, any>;
+    error: string | JSX.Element;
```
devinbinnie, line 99: You can use `React.ReactNode` here to represent anything that can be rendered to the DOM.

components/integrations/bots/bots.tsx
```diff
@@ -1,98 +1,108 @@
...
 
-import Bot, {matchesFilter} from './bot.jsx';
+import Bot, {matchesFilter} from './bot';
+
+type Props = {
+
+    /**
+    *  Map from botUserId to bot.
+    */
+    bots: Dictionary<BotType>;
```
devinbinnie, line 25: I believe we're trying to move away from these helper types and just use standard TypeScript helper types. So this can just be `Record<string, BotType>`

components/integrations/bots/bots.tsx
devinbinnie: Are we able to just do a cast or a `parseInt` here instead? I don't think we should have 2 copies of the same constant in case we ever need to change it.

devinbinnie: The other option would be to modify the constant in `mattermost-redux` to be a `number` instead.

devinbinnie (commented):
Thanks for the contribution @lucievr!

nevyangelova (approved):

lucievr:
Thanks for your review, @devinbinnie! I've commited the requested changes now.

Regarding the page size constant, I think what initially threw me off was that the BOTS_PER_PAGE_DEFAULT in mattermost-redux was defined as a number of 20 and here the PAGE_SIZE was a string of 1000. But I've changed it back now and I'm simply parsing the value as you suggested.

components/integrations/bots/bot.tsx
```diff
@@ -31,62 +33,73 @@ export function matchesFilter(bot, filter, owner) {
...
+    *  Only used for routing since backstage is team based.
+    */
+    team: Team;
+}
+
+type State = {
+    confirmingId: string;
+    creatingTokenState: string;
+    token: UserAccessToken | Record<string, any>;
+    error: string | ReactNode;
```
devinbinnie, line 98: Should just be able to use `ReactNode` here without the `string`.

devinbinnie (approved):
LGTM! Just one non-blocking comment.
Thanks @lucievr!

lucievr:
Thanks for the comment, @devinbinnie – fixed it now in my latest commit.

DHaussermann (approved):
Tested and passed
- Regression tested Bot account back stage UI
- No functional issues found. All UI elements work as expected
- No new JS issues found in console. 
- I do see some 404s occur when loading the list of bots but they this issue repros without this PR. Created https://mattermost.atlassian.net/browse/MM-31938 to address this
No new release tests needed to cover this.
LGTM!

Thanks @lucievr for this enhancement. Much appreciated!

PR #7254, opened 2020-12-22
[MM 30432] Custom desktop notification sound for channel

marcelo-cardozo:
#### Summary
This PR adds the functionality to specify custom desktop notification settings for each group, if the settings are not set for the group, it uses the account settings. 
A button to reset the default global settings is also added.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16515
JIRA: https://mattermost.atlassian.net/browse/MM-30432


components/channel_notifications_modal/components/desktop_notification_section.test.jsx
```diff
@@ -0,0 +1,124 @@
...
+import {shallow} from 'enzyme';
+
+import {NotificationLevels} from 'utils/constants';
+
+import DesktopNotificationSection from 'components/channel_notifications_modal/components/desktop_notification_section.jsx';
+import * as Utils from '../../../utils/utils';
+
+describe('components/channel_notifications_modal/DesktopNotificationSection', () => {
+    const baseProps = {
+        onChange: () => {}, //eslint-disable-line no-empty-function
```
hmhealey, line 14: These can use `jest.fn()` instead of an arrow function which will get around ESLint

components/channel_notifications_modal/components/desktop_notification_section.test.jsx
```diff
@@ -0,0 +1,124 @@
...
+        memberDesktopSound: undefined,
+        globalDesktopSound: 'true',
+        memberDesktopNotificationSound: undefined,
+        globalDesktopNotificationSound: 'Bing',
+        handleUpdateDesktopSound: () => {}, //eslint-disable-line no-empty-function
+        handleUpdateDesktopNotificationSound: () => {}, //eslint-disable-line no-empty-function
+    };
+
+    beforeEach(() => {
+        jest.clearAllMocks();
```
hmhealey, line 29: I think we do this automatically, so this might not be necessary

components/channel_notifications_modal/components/desktop_notification_section.test.jsx
```diff
@@ -0,0 +1,124 @@
...
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should show notification options', () => {
+        const userAgentMock = jest.requireMock('../../../utils/user_agent');
+        userAgentMock.isDesktopApp.mockImplementationOnce(() => false);
+
+        jest.spyOn(Utils, 'hasSoundOptions').mockImplementationOnce(() => true);
```
hmhealey, line 44: Instead of adding a mock here, it might be a bit easier to `jest.mock` these modules at the top level and then import them which lets you directly change the mocks without spying on Utils in every test.

```javascript
import * as UserAgent from 'utils/user_agent';
import * as Utils from 'utils/utils';

jest.mock('utils/user_agent'); // Mocks everything
jest.mock('utils/utils', () => ({
    ...jest.requireActual('utils/utils'),
    hasSoundOptions: jest.fn(() => true),
});

test('without sound options', () => {
    Utils.hasSoundOptions.mockReturnValue(false);
    // TODO test me
});
```

hmhealey: Also, you should be able to simplify these import statements by leaving off the `.jsx` extension and by using absolute imports. Not strictly necessary, but it saves some typing and will make it easier when the extensions change as we migrate more code to TypeScript in the future

components/channel_notifications_modal/components/__snapshots__/extra_info.test.jsx.snap
```diff
@@ -1,13 +1,6 @@
...
 
-exports[`components/channel_notifications_modal/ExtraInfo should match snapshot, on DESKTOP 1`] = `
-<span>
-  <FormattedMessage
-    defaultMessage="Selecting an option other than \\"Default\\" will override the global notification settings. Desktop notifications are available on Firefox, Safari, and Chrome."
-    id="channel_notifications.override"
-  />
-</span>
-`;
+exports[`components/channel_notifications_modal/ExtraInfo should match snapshot, on DESKTOP 1`] = `""`;
```
hmhealey, line 3: Should this test case be removed entirely since we're no longer using the `ExtraInfo` component for the desktop notifications?

components/channel_notifications_modal/channel_notifications_modal.jsx
```diff
@@ -224,6 +237,16 @@ export default class ChannelNotificationsModal extends React.PureComponent {
         if (serverError) {
             serverErrorTag = <div className='form-group has-error'><label className='control-label'>{serverError}</label></div>;
         }
+        const desktopNotification = {
```
hmhealey, line 240: This passes a new object into `NotificationSection` every time the modal re-renders which causes that to also re-render, even when it doesn't need to. I'd suggest we pass them in as separate props instead, but I think it might be worth not using `NotificationSection` for this at all. See below for what I mean.

components/channel_notifications_modal/channel_notifications_modal.jsx
```diff
@@ -283,8 +306,9 @@ export default class ChannelNotificationsModal extends React.PureComponent {
                                         expand={activeSection === NotificationSections.DESKTOP}
                                         memberNotificationLevel={desktopNotifyLevel}
                                         globalNotificationLevel={currentUser.notify_props ? currentUser.notify_props.desktop : NotificationLevels.ALL}
+                                        desktopNotification={desktopNotification}
```
hmhealey, line 309: Since `DesktopNotificationSection` works so differently from `NotificationSection`, maybe it would be a better idea to use `DesktopNotificationSection` here directly instead of having it be a child of `NotificationSection` so that we're not adding an extra case to `NotificationSection` that behaves entirely differently and requires passing around extra props that are only used in that one case

components/setting_item_max.jsx
```diff
@@ -271,7 +276,10 @@ export default class SettingItemMax extends React.PureComponent {
             <section
                 className={`section-max form-horizontal ${this.props.containerStyle}`}
             >
-                {title}
+                <div className='d-flex'>
```
hmhealey, line 279: Does adding this extra div visually change anything for existing settings that don't pass `actionButton`? I noticed some slight spacing changes between the test server and my local instance

https://user-images.githubusercontent.com/3277310/103822326-e6c79a00-503d-11eb-9b3e-7c8099496c2a.mov

sass/routes/_settings.scss
```diff
@@ -235,13 +235,26 @@
                     @include legacy-pie-clearfix;
                     @include alpha-property('background', $black, .05);
                     margin-bottom: 0;
-                    padding: 1em 0 1.3em;
+                    padding: 1em 1em 1.3em;
```
hmhealey, line 238: Actually, this is probably the change that's causing the differences I showed above. Did you mean to change this padding?

hmhealey (changes requested):
Thanks for the PR! Excited to have this feature in since I know a few people that really want to use this.

I ran into a few issues when testing this out on the test server:
1. The notification sound played doesn't actually change when I change the setting. I don't see any changes to the notification code here, so I don't know if something was missed there.
2. Changes to the notification sound (both when it plays and what sound it uses) don't last through a refresh, so they don't seem to be saved to the server.
3. When I first opened the modal, the Notification Sound selection appeared to be empty. I'd expect it to show whatever the account-wide default is to begin with.
    Screen Shot 2021-01-06 at 4 43 54 PM
4. When notifications are turned off and the notification sound setting isn't visible, there's an extra divider at the bottom of the setting. That should be removed
    Screen Shot 2021-01-06 at 4 43 54 PM

jfrerich:
Lots of great work, @marcelo-cardozo!  Could you address the feedback given by Harrison and feel free to ask for clarification if needed!

esethna:
@marcelo-cardozo this feature is awesome, let us know how we can help you get this merged. Really looking forward to it

esethna:
@marcelo-cardozo anything we can help with to move this along?

esethna:
@marcelo-cardozo let us know if you're still working on this, would love to see this in the app!

lindy65:
Hi @marcelo-cardozo - I'm helping to review stale PRs and, since we haven't heard from you in a few months, I'll close this PR for now. Please feel free to re-open if you're able to prioritize. Thanks!

I've re-added the "Up for Grabs" label on https://github.com/mattermost/mattermost-server/issues/16515

//cc @hmhealey @esethna

biva:
@lindy65 @jfrerich Apparently  @marcelo-cardozo is not able to work on this feature anymore. Would it be possible that the Mattermost team finalizes it, as a lot of work has already been done? It would be great!

jfrerich:
@esethna, what are your thoughts on Mattermost completing this PR?

esethna:
@hmhealey what's your estimate on the amount of work needed to finish this?

esethna:
@hmhealey @jfrerich following up on this ^ ;)

esethna:
Friendly ping @hmhealey, just looking for an estimate on the amount of work left here or if it's easier to start from scratch?

PR #7261, opened 2020-12-26
MM-31185 Added localization to date picker used when searching for posts around a given date with momentjs 

shuang2411:
#### Summary
Enabled localization of react-day-picker in /components/suggestion/search_date_suggestion with momentjs packages.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16520


components/suggestion/search_date_suggestion/index.js
```diff
@@ -8,12 +8,14 @@ import {getUserTimezone} from 'mattermost-redux/selectors/entities/timezone';
...
 import {getCurrentDateForTimezone} from 'utils/timezone';
 import {areTimezonesEnabledAndSupported} from 'selectors/general';
+import {getCurrentUserLocale} from 'selectors/i18n';
 
 import SearchDateSuggestion from './search_date_suggestion.jsx';
 
 function mapStateToProps(state) {
     const currentUserId = getCurrentUserId(state);
     const userTimezone = getUserTimezone(state, currentUserId);
+    const currentLocale = getCurrentUserLocale(state);
```
hmhealey, line 18: I think this should be `getCurrentLocale`, not `getCurrentUserLocale`.

components/suggestion/search_date_suggestion/search_date_suggestion.jsx
```diff
@@ -5,10 +5,31 @@ import React from 'react';
...
 
+import MomentLocaleUtils from 'react-day-picker/moment';
+
 import Suggestion from '../suggestion.jsx';
 
 import 'react-day-picker/lib/style.css';
 
+import 'moment';
+
+import 'moment/locale/de';
```
hmhealey, line 16: Would it be possible to import only the packages here that are actually needed for the user's locale to save a bit of bandwidth? We might be able to do something like
```javascript
const loadedLocales = {};

export default class SearchDateSuggestion extends Suggestion {
    componentDidMount(props) {
        if (props.locale && !loadedLocales[props.locale]) {
            // There might be something special needed here as well since we use `en` instead of
            // `en-ca` for English.
            loadedLocales[props.locale] = require(`moment/locale/${props.locale}`);
        }
    }
    
    componentDidUpdate(prevProps) {
        // do the same thing here if locale changes
    }
}
```

components/suggestion/search_date_suggestion/search_date_suggestion.jsx
```diff
@@ -22,11 +43,19 @@ export default class SearchDateSuggestion extends Suggestion {
                 today: this.props.currentDate,
             };
         }
+
+        let currentLocale;
+        if (this.props.currentLocale) {
```
hmhealey, line 48: Did you ever run into a case where this was empty? I think `getCurrentLocale` always returns a default value

components/suggestion/search_date_suggestion/index.js
hmhealey: Also, we could call the prop `locale` instead of `currentLocale` to save some typing

hmhealey (changes requested):
Thanks for the PR, @shuang2411! I have a few suggestions of some things we might be able to improve.

components/suggestion/search_date_suggestion/search_date_suggestion.jsx
shuang2411: I do not think `currentLocale` would be empty with the current code. I added the if statement as a sanity check to make the code less coupled.  Do you suggest me to remove it?

components/suggestion/search_date_suggestion/search_date_suggestion.jsx
shuang2411: Very clever! Will do!

components/suggestion/search_date_suggestion/index.js
shuang2411: Fixed!

components/suggestion/search_date_suggestion/search_date_suggestion.jsx
shuang2411: I have implemented this change, but it seems lint does not like it. Currently I asked the lint to ignore it, please let me know what to do.

hmhealey: That's totally fine. We've done similar things for other libraries that we only want loaded sometimes.

components/suggestion/search_date_suggestion/search_date_suggestion.jsx
hmhealey: It's fine either way. I was mostly curious if you had run into something like that since it wouldn't be the first time something was null when I expected it to always be set.

hmhealey (approved):
Thanks for making those changes! I believe the other reviewer is out until next week, but he should be back to review it sometime then

shuang2411:
> Thanks for making those changes! I believe the other reviewer is out until next week, but he should be back to review it sometime then

Thank you so much!

jespino (approved):
Looks good to me. Thanks @shuang2411! :tada:

ogi-m (approved):
LGTM 👍

hmhealey:
Thanks again for your PR, @shuang2411! This'll be available on Mattermost Cloud in mid-February and in Mattermost 5.33 in March!

PR #7263, opened 2020-12-30
MM-29089: Disable Admin Filter

sbishel:
#### Summary
A System User with Authentication Permissions is allowed to modify the Admin Attribute/Filter for LDAP/SAML. However it is not allowed to save. (Handled correctly on the server.)

This PR will disable both then `Enable` fields and the Attribute/Filter field, if the user is not a System Admin. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29089

#### Screenshots
Screen Shot 2021-01-04 at 3 25 13 PM
Screen Shot 2021-01-04 at 3 24 44 PM


hahmadia (approved):
Looks good and makes sense to me!

mkraft (approved):

furqanmlk (approved):
System Manger / User Manager are able to view Admin Filter option but it is disabled in both LDAP and SAML
Test case has been created 
[MM-T3720](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/MM-T3720)

PR #7270, opened 2021-01-04
Cypress/E2E: Ignore uncaught exception by cypress

saturninoabril:
#### Summary
Ignore/turn off all uncaught exception by Cypress.
There are several varieties of error and none could easily determined whether it's from the application itself or Cypress. Note that those uncaught exceptions were not observed while checking manually; only during test automation with Cypress.

#### Ticket Link
Failing on daily Cypress test.

#### Screenshots
On local:
Screen Shot 2021-01-04 at 6 36 21 PM

On CI:
Screen Shot 2021-01-04 at 7 09 39 PM


josephbaylon (approved):
LGTM. Thanks @saturninoabril !!

PR #7274, opened 2021-01-05
Merge latest master into cloud

jwilander:


jwilander:
It looks like we've been bitten by the squash and merge commits. I think we need to do merge commits in the future (which aren't enabled right now)

PR #7276, opened 2021-01-05
MM-30053/GH-16204 Clear selection in channel switcher on backspace

rolwin100:
#### Summary
Channel switcher should focus first list result after a backspace

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/16204
https://mattermost.atlassian.net/browse/MM-30053


nevyangelova:
Jira: https://mattermost.atlassian.net/browse/MM-30053

nevyangelova:
Thanks @rolwin100 for your contribution 🎉  please see the failing tests

nevyangelova (changes requested):
test need to be updated :)

rolwin100:
@nevyangelova when I made some changes in the component the test is failing. So do I need to make changes to the test case or I need to modify my component to match the test case.

hmhealey:
You'd want to update the unit test to match the new behaviour if it's breaking because of something you intentionally changed. Based on the failed test, I'm not sure if it's correct or if it's a mistake.

hmhealey:
You'd want to update the unit test to match the new behaviour if it's breaking because of something you intentionally changed. Based on the output of the failed test, I'm not sure if the unit test is correct or if it's the code.

components/suggestion/suggestion_box.jsx
```diff
@@ -617,10 +617,7 @@ export default class SuggestionBox extends React.PureComponent {
         const items = suggestions.items;
         let selection = this.state.selection;
         if (terms.length > 0) {
-            // if the current selection is no longer in the map, select the first term in the list
-            if (!this.state.selection || terms.indexOf(this.state.selection) === -1) {
-                selection = terms[0];
-            }
+            selection = terms[0];
```
hmhealey, line 620: Something we're losing by changing this in this way is that the autocomplete would previously keep your cursor where it was if you select something and then continue typing. For example, if you type "t" and scroll to "Town Square" and then continue typing the rest of the channel name, it would keep Town Square selected even if it's not the first result that appears for "to".

We'll want to keep that, but we want to select the first result when the user hits backspace, so are you able to find a way to treat those differently when it comes to handling the selection?

hmhealey:
Thanks for the contribution by the way! I was too focused on answering your question that I forgot that earlier 😄

components/suggestion/suggestion_box.jsx
rolwin100: got it will work on that change

rolwin100: @hmhealey don't you think the functionality should work while typing as well. Does not make sense to restrict this for only the backspace. Because when a new list comes obviously the 1st should be selected. Please share your thoughts on this. @nevyangelova also told me to just update the test case

hmhealey: Yeah, I'm starting to agree with you there. Back when we first added the `SuggestionBox`, we thought it was important to keep the selection if the user continued typing, but I don't know if it makes sense any more. I've asked our UX team what their thoughts are on changing that in all cases.

rolwin100: @hmhealey any updates on this. We can proceed with the new changes and get the code merged if you agree with me.

rolwin100: It's been some time since this issue is around. We need to get it merged.

esethna (approved):

jgilliam17 (approved):
Thank you @rolwin100 
Tested, looks good to merge.
- Verified first item in the results list on channel switcher is focused when using backspace.
Mouse hover and keyboard arrows selection of the result does not keep the focus on backspace, focus jumps back to top result - we're ok with this as discussed in triage today.

hmhealey (approved):
Thanks for the PR on this. We discussed whether or not we wanted to keep the old behaviour when typing more, and we came to the conclusion that it's probably not needed. Comparing with some other services, most of them didn't worry about keeping your selected term when you continue typing. Hopefully that'll simplify some things when we do some bigger cleanup of the SuggestionBox in the future.

rolwin100:
> test need to be updated :)

I have updated the tests as well can you please check @nevyangelova

nevyangelova (approved):
LGTM!

nevyangelova:
@amyblais can this go into the next release?

nevyangelova (approved):

PR #7280, opened 2021-01-06
Cypress/E2E: Add explicit wait with cy.postMessage

saturninoabril:
#### Summary
On local dev, specifically when running with `npm run cypress:open`, `cy.postMessage` randomly failed waiting for `#post_textbox` to be visible. Not sure if you're seeing it too. As I observed, the channel page took so long to load on Cypress than normal (manual). It seemed that some webapp operation is blocked by Cypress while trying to wait for an element with `cy.get(textboxSelector, {timeout: TIMEOUTS.HALF_MIN})`.
I'm not a fan of explicit wait but I find it reasonable in this case. Please try on your local setup and let me know if something improves.  TIA.

#### Ticket Link
none

### DIFF
--- a/e2e/cypress/support/ui_commands.js
+++ b/e2e/cypress/support/ui_commands.js
@@ -107,6 +107,10 @@ Cypress.Commands.add('uiPostMessageQuickly', (message) => {
 });
 
 function postMessageAndWait(textboxSelector, message) {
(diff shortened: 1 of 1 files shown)


josephbaylon (approved):
LGTM.

prapti (approved):
I have not seen a recent failure with `#post_textbox` in my local setup but I did in the past. Sounds good to me too 👍

PR #7281, opened 2021-01-06
[MM-24635] adds an Input component with max length validation and visual feedback

koox00:
#### Summary
Adds a new component `MaxLengthInput` which provides feedback about the number of times the
length of the value of the input exceeds allowed max length, in case that happens that is.

This is done to add a visual feedback upon creating/editing custom category names,
but can be re-used wherever we use `QuickInput`.

The allowed max characters for custom category names is set to be 22.

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-server/issues/15904

#### Screenshots

Screenshot_2021-01-22 Town Square - test Mattermost


esethna:
Thanks @koox00! Adding @matthewbirtch to do a UX review

matthewbirtch (changes requested):
Nice work on this @koox00. A few very minor things to update:

1. Can we move the count to the left by 4px so it's not quite so close to the clear icon?
2. Can we bump the clear icon and the count down by 1px so the text baseline aligns with the text on the left?
Screen Shot 2021-01-12 at 4 25 50 PM
3. The text and the count collide with really long names or when the browser width is smaller. Maybe we should just hide the overflow before it collides.

image

Other than that, I think it's great!

calebroseland (changes requested):
Overall this is working great! Nice work.

There are a few reusability items that could be done, but this is very subjective given the current state and usage of the existing inputs. 

- `maxlength_input.tsx` feels coupled to QuickInput, so one option might be to move those components into a folder together.

- Big-picture, it would be nice to consolidate the styling and feature variations of these inputs into a central place, but I believe this is out-of-scope, so please see this as an aside comment.


Other than a couple minor ones in maxlength_input.tsx, LGTM 💯

components/maxlength_input.tsx
```diff
@@ -0,0 +1,44 @@
...
+    maxLength: number;
+}
+
+export type Ref = HTMLInputElement;
+
+const MaxLengthInput: FC<InputProps> = forwardRef<Ref, InputProps>(
+    ({className, defaultValue, maxLength, ...props}: InputProps, ref) => {
+        const excess: number = defaultValue ? defaultValue.length - maxLength : 0;
+
+        const styles: string = classNames({
```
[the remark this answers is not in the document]
hmhealey: Yeah, I'd lean in the direction of inlining or using `classes`. We don't use the name `styles` much in the web app, but it's used constantly for StyleSheets in the mobile app

components/generic_modal.scss
```diff
@@ -14,6 +14,21 @@
             box-sizing: border-box;
             border-radius: 4px;
             height: 40px;
+
+            &.has-error {
```
hmhealey, line 18: Can you move these CSS styles out of the `.GenericModal` class? We're going to want to be able to use this component elsewhere, so it should be self-contained. At the same time, can you rename the max-length class to `MaxLengthInput` to match the component? That's a naming convention we're trying to start using for CSS at the root of a component

hmhealey (changes requested):
Looks pretty good, but I have some organizational changes to suggest. In addition to the one I mentioned below, could you move everything for this component into `components/max_length_input` so that it's all in one place? You could even put the CSS in there in its own file as well, and it'll be included in the build as long as it's imported by the TS code

components/generic_modal.scss
koox00: @hmhealey I had 2 concerns and didn't take that route,

1. inputs may vary on height, or padding, etc, so styling the validation message outside the input it's used may also 
break re-usability (it's an absolute positioned element), so my thought was to just pass classes like `.has-error` which other component may override so it would be easier re-usable.

2. `QuickInput` is being styled inside `.GenericModal`, so it seemed like the reasonable place for `MaxLengthInput` to be styled as well.

Please let me know your thoughts on 1.

koox00:
>  3. The text and the count collide with really long names or when the browser width is smaller. Maybe we should just hide the overflow before it collides. 

@matthewbirtch would something like be ok?

image

matthewbirtch:
> @matthewbirtch would something like this be ok?
> 
> image

@koox00 yep, I'm good with that approach for the overflow. Looks to me like the 'x' icon does need to shift down by 1px still though

components/generic_modal.scss
hmhealey: >1. inputs may vary on height, or padding, etc, so styling the validation message outside the input it's used may also
>break re-usability (it's an absolute positioned element), so my thought was to just pass classes like `.has-error` which other component may override so it would be easier re-usable.

That's fair. I think we'll eventually end up making those inputs more consistent since there's a big design system overhaul coming eventually, but we can do that once that's ready.

> 2. `QuickInput` is being styled inside `.GenericModal`, so it seemed like the reasonable place for `MaxLengthInput` to be styled as well.

I didn't realize we were also doing that in there. I guess trying to move `MaxLengthInput` out of there but not `QuickInput` would probably mean we'd run into issues with CSS specificity causing styles to override each other. That's another reason to wait for that until we do the bigger overhaul.

I'd appreciate the renaming still though

hmhealey (approved):
Thanks for changing that and answering my other questions. Looks good to me.

calebroseland (approved):

matthewbirtch (approved):
Looks good!

koox00:
@hmhealey do you mind if I squash and rebase with master?

jgilliam17:
Thank you @koox00, looks great
UX question - @matthewbirtch I stumbled upon some emoji worth 11 characters, should we be more lenient when using emoji in the category name?
This is not new to this PR, the existing hard stop we use on master results in the same problem e.g. only 2 family emoji could be entered and user would be stopped from using additional characters or emoji in the name.
Screen Shot 2021-01-27 at 10 10 57 PM

hmhealey:
@koox00 Feel free to keep it the branch up to date. As long as there's no major changes for that, we don't need to re-review. We prefer merging master in over rebasing though since it makes it much easier to see what changed as reviewers.

@jgilliam17 That can be handled elsewhere. I am though curious if that's an issue on the frontend or a requirement on the backend. If it's dependent on the database, that's probably a limit on the number of bytes that the text takes up, and some emojis like the family ones take up more bytes compared to characters. If it's just the frontend rejecting that, we're just counting those wrong on the frontend

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified text input validation on creating and editing custom category modals; Red border and negative character count feedback present when exceeding the character limit.
Thanks Harrison - I'll follow up with Matthew about emoji issue.

matthewbirtch:
> Thank you @koox00, looks great
> UX question - @matthewbirtch I stumbled upon some emoji worth 11 characters, should we be more lenient when using emoji in the category name?
> This is not new to this PR, the existing hard stop we use on master results in the same problem e.g. only 2 family emoji could be entered and user would be stopped from using additional characters or emoji in the name.

> I am though curious if that's an issue on the frontend or a requirement on the backend. If it's dependent on the database, that's probably a limit on the number of bytes that the text takes up, and some emojis like the family ones take up more bytes compared to characters. If it's just the frontend rejecting that, we're just counting those wrong on the frontend.

Seems like we should queue up another ticket to look in to this. @jgilliam17 would you mind creating on that we can triage/groom?

jgilliam17:
@matthewbirtch  Follow up issue added https://mattermost.atlassian.net/browse/MM-32410

hmhealey:
Thanks again for the PR, @koox00!

PR #7283, opened 2021-01-06
support auto rtl detection in browsers

phntom:
#### Summary
adds rtl support to messages with rtl (hebrew/arabic) unicode as the first non digit/punctuation letter
this is already what the android client does with these messages

#### Screenshots

image


devinbinnie (approved):
LGTM! Thanks @phntom!

phntom:
Hi, so i'm running it on my server for testing, and i've got some feedback.
I'll make a slightly bigger change, mainly:
* it looks better with .post-message__text{text-align:left}
* to be consistent with android - it should be dir: auto on a line level and not on a post level
* tables don't format well
* the editor and the inline editor need this as well

but other than that it's a pretty good change.
you can see how it looks in https://nix.co.il (or kix.co.il if you register)

phntom:
Here's how the new version with post-message__text {
    text-align: left;
} looks like:

image

hmhealey:
Thanks for the PR, @phntom! This sounds really cool since we've discussed how adding full RTL support to Mattermost is difficult, but if we can add something smaller like this to improve support in some small way, that'd be great.

Unfortunately, I don't know who on the team is actually fluent in any RTL language to test this, but I've reached out to our translation community to see if anyone can help from there.

Regarding your feedback:

> - it looks better with .post-message__text{text-align:left}

Is this normal for RTL languages like in other messaging services? I would've thought that it would be right-aligned. I do agree that it might look odd when everything else is left-aligned though.

> - to be consistent with android - it should be dir: auto on a line level and not on a post level
> - tables don't format well

Perhaps we could do this by applying the `dir=auto` more narrowly by adding it per-element in the Markdown renderer. For example, we could add it to individual [headings](https://github.com/mattermost/mattermost-webapp/blob/master/utils/markdown/renderer.tsx#L170) and [paragraphs](https://github.com/mattermost/mattermost-webapp/blob/master/utils/markdown/renderer.tsx#L254)? That would also let us target tables separately.

> - the editor and the inline editor need this as well

I'm also not sure what that would require. We're planning to make some large changes to the post textbox in the next few months, so we could also add that if it's simple, but we might need a bigger effort with more heavy testing if it requires a large amount of work.

hmhealey:
Also, not to undermine your efforts here, but have you taken a look at [the Mattermost RTL plugin](https://github.com/QueraTeam/mattermost-rtl)? That might do some more of the things you're looking for

hmhealey:
Apologies for not getting back to this sooner. We got some feedback from another community member, and they actually preferred the initial version where RTL text was right-aligned. That is consistent with Slack Mobile (but not web) and LinkedIn, and it also lines up with what the RTL plugin does. If you want to read the full conversation, you can sign up for [our community server](https://community.mattermost.com) and view the conversation [here](https://community.mattermost.com/core/pl/rpz3kg4ay3d83ptjmwbb6pznyr).

phntom:
hi, thanks for bringing this up:
so - exactly like you said, slack web opts to left aligned and mobile uses android defaults similar to how mattermost android uses android defaults.
image

it is my opinion the right aligned text will only look good if the rest of the elements including the user icon and nickname would also right align, and we had users experiment with this look.

in general my experience with older RTL user interfaces is:
a full RTL swap for everything *was* the prefered way for older apps like Microsoft Word and even Gmail when everything is translated to an RTL language.

newer clients i.e. slack are more focused towards the millennial crowd - and it's quite common to use English UI apps with RTL languages being left aligned. this is arguably a more modern look for and English UI
and this doesn't preclude a full RTL swap on an RTL language UI

i would like to make a fix for the english UI since hebrew and arabic seem to be quite a long way from being ready.

it is quite jarring having text right aligned with the amount of dead space the web ui has, and having the nickname on the opposite side is uncomfortable - we tested this for 3 days and tested the left-aligned version for 2 weeks and the left aligned version is a clear winner for everyone.

I did look over the plugin, and I will use it's list of CSS elements for a better result.
It's my understanding the plugin uses setTimeout to periodically fix elements on the screen which creates an uncomfortable delay for text to jump on screen.

and apologies for not being more available, i want to get this out but i'm having the worst computer malfunction on my personal rig and it's quite hard to get everything sorted in a pandemic.

esethna:
@phntom this is a really interesting change, and a clever way to approach this issue. It seems to be common standard to automatically flip the behaviour once an RTL character is used. A few thoughts:

1. We need to make sure tables still behave as expected. I'd imagine table rows should not be reordered horizontally, but the text within a cell should be rendered RTL? I'd lean on your experience with other products here
2. I think it would be ideal if the text input area flipped the text horizontally once an RTL character is entered first (similar to Slack desktop) @nevyangelova would this be something to consider in the advanced text editor design?

phntom:
i i'll try to get to this in a day or two and yes, agreed with the feedback, my current goal is an iterative improvement, nothing ground breaking. I'd like the words to be in proper order while mostly still remaining in the same spot and occupying the same space as before.

nevyangelova:
> @phntom this is a really interesting change, and a clever way to approach this issue. It seems to be common standard to automatically flip the behaviour once an RTL character is used. A few thoughts:
> 
> 1. We need to make sure tables still behave as expected. I'd imagine table columns should not be reordered horizontally, but the text within a cell should be rendered RTL? I'd lean on your experience with other products here
> 2. I think it would be ideal if the text input area flipped the text horizontally once an RTL character is entered first (similar to composing a message on Slack desktop) @nevyangelova would this be something to consider in the advanced text editor design?

For sure, you mean auto detecting a keyboard input change and then adjust the caret and direction based on that right?

phntom:
guys, this pr keeps getting out of sync, can someone review there are no regressions and merge this?
i'm running this rebased on 5.29.2 at nix.co.il and it's working fine.
the latest change has auto rtl in messages, text editing, channel topic.
this change does not modify the markdown parser so it doesn't apply to code and tables.

hahmadia:
Gentle ping to @esethna @hmhealey

esethna:
@abdusabri would you mind doing a review of the latest state of this PR? 

@hmhealey concerns about taking this PR as is? To me it seems to satisfy the RTL requirements quite well until we can actually build an RTL UI.

@phntom does the MM RTL plugin still work unaffected after this change? https://github.com/QueraTeam/mattermost-rtl

phntom:
> @abdusabri would you mind doing a review of the latest state of this PR?
> 
> @hmhealey concerns about taking this PR as is? To me it seems to satisfy the RTL requirements quite well until we can actually build an RTL UI.
> 
> @phntom does the MM RTL plugin still work unaffected after this change? https://github.com/QueraTeam/mattermost-rtl

it will work but i did left-align elements like the messages. i feel this is better visually (to keep the text closer to the icon) and for parity/compatibility reasons but... the plugin up to this point was displaying these messages right aligned.

it should be a quick two liner to update the plugin with CSS to force right align if this was the intent.

phntom:
here's how the test server looks like:
image

abdusabri:
@esethna @phntom 

I've tested the latest updates and text display in posted messages LGTM! Definitely better than the current experience 🙂 

On the left is text from daily, on the right is from the test server of this PR
image

However, I've noticed an issue in the text editor (while composing a message), on daily, I could flip the direction and make the caret RTL while composing a message (pressing Right CTL + Right Shift keys on Windows), but I can no longer do the same on the test server of this PR

image

-----------------------------------------------

There is another improvement that could be made, @esethna in case you wanna file a ticket for it, is that for markdown tables, the order of the columns is reversed, at least based on my own testing. Not sure if @phntom has experienced the same.

Anyway, it is not related to the changes in this PR, the behavior here matches the behavior on daily

image

phntom:
@abdusabri as you pointed out the text in the edit box is being force left-aligned, this is intentional, our users seem to prefer the text editor to be more similar to the end result since messages are also left-aligned. This is up for debate of course, mostly this is my preference for text boxes that are wide.
And the tables are force LTR'd because it looks like the markdown library will needs to be modded. This does create a weird experience before the formatting where text is RTL'd but I feel this is an edge case most people won't experience.
I'm open to suggestions, I've tried some other combinations (without changing markdown) for tables and feel like this one looks best under the markdown restrictions.

hmhealey:
> guys, this pr keeps getting out of sync, can someone review there are no regressions and merge this?

Sorry, we have a pretty annoying CI check to try to prevent merge issues that come up without actual conflicts. Feel free to let that check out of date since we can update the branch ourselves before merging.

> @hmhealey concerns about taking this PR as is? To me it seems to satisfy the RTL requirements quite well until we can actually build an RTL UI.

@esethna As far as I can tell, it should be fine. I'm not fully confident judging that since I don't use any RTL languages myself, but I'm good with it based on everyone's feedback.

Regarding the Markdown, I'm curious how RTL works with it elsewhere (like here on GitHub). Could you post the raw markdown for that table? I don't think it needs to be solved for this PR, but it would be useful to have around for future use.

hmhealey (approved):

esethna (approved):

esethna:
Agree, we should create a follow-up ticket for the markdown table formatting cc// @jgilliam17

jgilliam17 (approved):
Thank you @phntom 
Tested, looks good to merge.
- Verified right-to-left was properly detected in the main text box, RHS, edit post modal and channel description.
Follow up to add RTL support to markdown tables created https://mattermost.atlassian.net/browse/MM-33241
@hmhealey Can you please help merge. Thanks 🙂 
Not sure if Jira ticket was created for this PR. If not, let me know and I can add one to keep track and re-test after merge.

lieut-data:
@hmhealey, can you help merge this once successfully built?

hmhealey:
Merged! Thanks again for the PR, @phntom

erezo9:
@phntom I have upgraded to 5.34.2, and it seems that hebrew is not working on the rtl side
all of the posts have been switched to the left
also, it made the plugin of the querateam stopped working

phntom:
> @phntom I have upgraded to 5.34.2, and it seems that hebrew is not working on the rtl side
> all of the posts have been switched to the left
> also, it made the plugin of the querateam stopped working

Hi Erez,

TL;DR
The posts should have switched to be left aligned, But with the correct order of mixed Hebrew/Arabic and English words. And the plugin should also become left aligned until it is updated with the new CSS changes.
Now the English Web UI will be a similar experience to the mobile implementation and the mixed language text will be readable. 

---

I'm pretty sure I noted this somewhere in this discussion, I did take a look at this plugin and how this will affect it. 
There are two issues with it:

First, I've added the dir=auto in more places, some of them require a deep code change to the markdown library (for example the code block), so I've disabled them on these specific cases (i.e. dir=ltr) which will also disable them in the plugin.

Second, after a couple of iterations with our end users and discussing with members here and in the mattermost community server, we've decided to emulate the OG LTR English experience and just first the order of the words the RTL languages.

Once an Arabic and/or Hebrew UI is available we could fully support an RTL text orientation and alignment.

This breaks the plugin because I've forced a left alignment in places the plugin did not "fix" so as I'm pretty sure I've already mentioned in this thread, the plugin will need to be updated for this CSS change.
It's a small one and I can probably do a PR for them, but I still don't think the plugin's approach to right aligning text is the best approach.

Also, It's been a while since I've worked on this back in January, and I've been on a patched version and now an RC for a while, so if I'm missing something feel free to point out.

Anton

erezo9:
@phntom okay, for now in 5.34.2 it is broken and mattermost has published that it does work
https://docs.mattermost.com/help/messaging/sending-messages.html
but it dosent, not by my side at least, i have testes on the production/test and even on community site it does not work

phntom:
@erezo9  can you elaborate of the "it is broken" part? i'm using this version on our servers, and it does what i expected it to do.

if you expect it to be right aligned, it's not doing that by choice, but the word order should be correct for sentences that start with an RTL character, similar to how your message will be displayed in slack, facebook and mattermost mobile.

erezo9:
@phntom 
The rtl part really does work (word order), its what the plugin also did, the only thing now, its not right aligned, thats what broken for now, and it seems that installing the plugin will not make this work, as you said they need to update it.

erezo9:
@phntom Is there a way to make the alignment work somehow?

PR #7284, opened 2021-01-06
[MM-31580] Fix call to getChannel to use new parameter format

devinbinnie:
#### Summary
`makeGetChannel` was changed to take an object including an `id`, but this call to it wasn't updated, causing a bug in notifications. This PR fixes that call.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31580

### DIFF
--- a/actions/notification_actions.jsx
+++ b/actions/notification_actions.jsx
@@ -48,7 +48,7 @@ export function sendDesktopNotification(post, msgProps) {
         }
         const teamId = msgProps.team_id;
 
-        let channel = makeGetChannel()(state, post.channel_id);
+        let channel = makeGetChannel()(state, {id: post.channel_id});
         const user = getCurrentUser(state);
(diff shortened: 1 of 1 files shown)


gigawhitlocks (approved):
Thank you!

jgilliam17 (approved):
Thank you @devinbinnie 
Tested on Ubuntu, looks good to merge.
- Verified notifications are not triggered when user has the desktop app in focus. Also verified notifications are received as expected when app is not in focus/minimized.

amyblais:
@hmhealey Kind reminder to review this PR for today's v5.31 code freeze deadline.

hmhealey (approved):
I can't even remember it not taking an object. I did want to remove the object at some point, but I never got around to it. This must've been broken for quite a while, but I assume no one noticed because the default is to send notifications for all DMs

PR #7285, opened 2021-01-06
MM-31616 - fix js errors shown for switch channel&incident plugin suggestion m…

pablovelezvidal:
#### Summary
Fixes the issue reported when opening the switch channels modal and when adding a new incident via the incident plugin.

This commit makes sure that the logic applied to the suggestion_box to align the caret with the avatar/icon added in original [PR7082](https://github.com/mattermost/mattermost-webapp/pull/7082) will only and only apply to the Mention, Channel or Emoji provider. Also, checks for the two alignment values to not be undefined.

#### Ticket Link:
[MM-31616](https://mattermost.atlassian.net/browse/MM-31616)

### DIFF
--- a/components/suggestion/suggestion_box.jsx
+++ b/components/suggestion/suggestion_box.jsx
@@ -658,8 +658,9 @@ export default class SuggestionBox extends React.PureComponent {
 
             if (handled) {
                 if (this.state.suggestionBoxAlgn.pixelsToMoveX === undefined &&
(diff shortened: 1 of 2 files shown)


hmhealey (approved):

calebroseland (approved):

jgilliam17 (approved):
Thank you @pablovelezvidal 
Tested, looks good to merge.
- No console errors when opening the switch channels modal and when adding a new incident via the incident plugin. Drop downs work as expected.

PR #7287, opened 2021-01-07
MM-31716: Remove the word experimental from gossip setting

agnivade:
This PR just makes a small UI adjustment which removes the word experimental
from the system console. Any other things like variable names and id names
haven't been touched because the entire thing is going to be removed
after 3 months when the feature becomes default.

Test snapshots have been updated, and verified that there are no Cypress tests
matching this string.

https://mattermost.atlassian.net/browse/MM-31716

```release-notes
Removed the word "experimental" from gossip setting in the admin console
```


nevyangelova (approved):

calebroseland (approved):

agnivade:
This is just a UI change. So does not need any QA testing.

PR #7305, opened 2021-01-11
Fix limit banner show modal for cloud in-app purchases

nickmisasi:
#### Summary
Since the cloud banner was initially implemented the underlying logic has been migrated to typescript. Somewhere in this migration the button to open the banner broke. My current hunch is that the initial implementation had some loose typing that allowed it to dispatch the openModal action. In the migration the `showModal` type was changed to function | boolean (admittedly, that prop wasn't well named). I've changed it to `onButtonClick` and adjusted the types. I've also added it explicitly to the onClick handler of the button itself. 

#### Ticket Link
N/A, this one is off the cuff

#### Related Pull Requests
N/A

#### Screenshots
N/A


nickmisasi:
I've opened this PR directly against the cloud branch. We can cherry-pick to master after review.

marianunez:
> I've opened this PR directly against the cloud branch. We can cherry-pick to master after review.

Why directly to `cloud` first?

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
```diff
@@ -27,8 +27,9 @@ type Props = {
     type: string;
     message: React.ReactNode;
     handleClose?: (e?: any) => void;
+    showModal?: boolean;
```
marianunez, line 30: Should we rename this to `showButton` as well for consistency?

marianunez (approved):
Thanks @nickmisasi for the quick find!

nickmisasi:
> Why directly to `cloud` first?

Because I had the cloud branch checked out when I was investigating and making the change :)

nickmisasi:
@stevemudie we're not going to be able to quickly test that this works with the cloud banner, but this could maybe use a smoke test of normal in product banners to make sure they're not broken

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
nickmisasi: showModal is used by `configuration_bar.tsx` as a boolean on whether or not to show the _link_ in the banner (these will be for older banners). I think this is where the typescript conversion tripped up - javascript was a little more lenient on the type of that prop and would use it as a boolean or a function depending on the context. So I've left this in to support that case and then added the new prop for our banners

jwilander:
@nickmisasi this should probably go into `master` first and then get cherry-picked to `cloud`. We should only commit directly to `cloud` when it's a fix that doesn't make sense in `master`

nickmisasi:
> @nickmisasi this should probably go into `master` first and then get cherry-picked to `cloud`. We should only commit directly to `cloud` when it's a fix that doesn't make sense in `master`

I was worried about conflicts and figured I'd get up for the sake of time. Changed now to master so should be all good.

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
devinbinnie: Have we looked at refactoring what those other components are passing in? Probably OOS for this PR but i'm just curious what the effort would be, if it's even possible.

devinbinnie (approved):

e2e/cypress/integration/bot_accounts/bot_api_spec.js
```diff
@@ -7,6 +7,7 @@
 // - Use element ID when selecting an element. Create one if none.
 // ***************************************************************
 
+// Stage: @prod
```
marianunez, line 10: Is this intended?

nickmisasi: No, and not sure where it would have come from. I reverted these 2 files to be the same as master. All good.

marianunez: Thanks!

stevemudie (approved):
Looks good. Will test post-merge.

marianunez:
@nickmisasi all tests have passed, let's merge this so we can cherry-pick for cloud for tomorrow's release.

PR #7310, opened 2021-01-12
MM-20437 - migrate close channels to typescript

pablovelezvidal:
#### Summary
This converts the close channel component  to TypeScript.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16137


components/channel_header_dropdown/menu_items/close_channel/close_channel.tsx
```diff
@@ -2,21 +2,20 @@
...
-    static propTypes = {
-        isArchived: PropTypes.bool.isRequired,
-        actions: PropTypes.shape({
-            goToLastViewedChannel: PropTypes.func.isRequired,
-        }).isRequired,
-    }
+type Props = {
+    isArchived: boolean;
+    actions: {
+        goToLastViewedChannel: () => () => void;
```
devinbinnie, line 13: Does this need to be a function that returns a function? Is just `() => void` not enough here?

devinbinnie (approved):
LGTM! Just one non-blocking comment

components/channel_header_dropdown/menu_items/close_channel/close_channel.tsx
pablovelezvidal: well, yes, I think that in practical terms it is enough; although the goToLastViewChannel is a function that returns a function. I am not sure if there is a convention for this. I am happy to modify it to follow the MM standard.

devinbinnie: If I understand correctly, once the function is bound using a redux connector, the return value of `goToLastViewedChannel` would be the return value of the function it returns, since we aren't passing in `dispatch` or `getState` when calling it here. @hmhealey would know for sure.

hmhealey: Yeah, that's correct. `bindActionCreators` in the connnector takes the `(arguments) => (dispatch, getState) => returns` that makes up a redux-thunk action like `goToLastViewedChannel`, and it turns it into `(arguments) => returns`. In this case, the type just ends up as `() => void`

components/channel_header_dropdown/menu_items/close_channel/close_channel.test.tsx
```diff
@@ -17,7 +17,7 @@ describe('components/ChannelHeaderDropdown/MenuItem.CloseChannel', () => {
     };
 
     it('should match snapshot', () => {
-        const wrapper = shallow(<CloseChannel {...baseProps}/>);
+        const wrapper: ShallowWrapper<any, any, CloseChannel> = shallow(<CloseChannel {...baseProps}/>);
```
hmhealey, line 20: You can simplify these a bit by moving the type parameter to `shallow` instead of putting it on the wrapper itself
```suggestion
        const wrapper = shallow<CloseChannel>(<CloseChannel {...baseProps}/>);
```

hmhealey (changes requested):

components/channel_header_dropdown/menu_items/close_channel/close_channel.test.tsx
pablovelezvidal: Hi @hmhealey , all those suggestions are ready!

hmhealey (approved):
Thanks for changing those. Looks good to me!

PR #7320, opened 2021-01-13
MM-29935 - update pricing visibility and legends

pablovelezvidal:
#### Summary
Updates the subscription billing information page
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29935

```release-note
NONE
```

#### Screenshots
Captura de pantalla 2021-01-14 a las 13 54 50
Captura de pantalla 2021-01-14 a las 13 59 02
Captura de pantalla 2021-01-14 a las 14 02 38


components/admin_console/billing/plan_details.tsx
```diff
@@ -136,9 +142,11 @@ const PlanDetails: React.FC = () => {
     }
 
     let planPricing;
+    let planDetailsDescription;
+    let userCountDisplay;
 
     const showSeatsAndSubscriptionDates = false;
-    let userCountDisplay;
+    const goneBackUnderThreshold = userCount <= userLimit && (userCount + userInactiveCount) > userLimit;
```
marianunez, line 149: We don't take inacive users into account for user limits. If we want to see if they have gone back down from the threshold, we can just check `userCount <= userLimit` and that the subscription type is paid tier `susbscription.is_paid_tier == `true``. This would be the case that they have upgraded and but won't get charged because they are below the limit

components/admin_console/billing/plan_details.tsx
```diff
@@ -196,6 +219,18 @@ const PlanDetails: React.FC = () => {
                 />
             </div>
         );
+        planDetailsDescription = goneBackUnderThreshold ? (
+            <div className='PlanDetails__description'>
+                <div className='PlanDetails__planDetailsName'>
+                    <FormattedMessage
+                        id='admin.billing.subscription.planDetails.planDetailsName.freeUpTo'
+                        defaultMessage='Free for up to {aboveUserLimit} users.'
```
marianunez, line 227: I believe the design indicates here should say "Free for up to 10 users" so we should use here `userLimit`
```suggestion
                        defaultMessage='Free for up to {userLimit} users.'
```

marianunez (changes requested):
@pablovelezvidal looks good! Some minor comments on how to determine when they are down from the user limit threshold

marianunez:
@pablovelezvidal could you include some screenshots of both cases for ease `UX Review`. 

Also, I've updated the test server and now the Subscription screen in the System Console should be available for `QA Review`.

matthewbirtch:
@pablovelezvidal everything looks good to me, but for the screenshot below, can you describe the case? If it's below or _at_ the user limit, shouldn't it still say 'Free' in large type instead of $10.00?
image

pablovelezvidal:
> @pablovelezvidal everything looks good to me, but for the screenshot below, can you describe the case? If it's below or _at_ the user limit, shouldn't it still say 'Free' in large type instead of $10.00?
> image

I think that happens in the case  when paid customers have gone back below the 11 user threshold

marianunez (approved):
Looks great, thanks @pablovelezvidal!

matthewbirtch:
> I think that happens in the case when paid customers have gone back below the 11 user threshold

Ahhh, ok, that makes sense now. I think it should be good to go then! Thanks.

matthewbirtch (approved):

pablovelezvidal:
> > I think that happens in the case when paid customers have gone back below the 11 user threshold
> 
> Ahhh, ok, that makes sense now. I think it should be good to go then! Thanks.

Cool! thank you Matt!

stevemudie (approved):
LGTM

nickmisasi (approved):
LGTM - thanks @pablovelezvidal

PR #7327, opened 2021-01-15
MM-31654: Update steps for google openid

sbishel:
#### Summary
Updates the instructions for setting up OAuth application in google.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31654

#### Screenshots
Screen Shot 2021-02-01 at 8 11 55 AM


hahmadia (approved):

BenCookie95 (approved):

furqanmlk:
Steps are not matching with screenshot posted above

image

components/admin_console/admin_definition.jsx
```diff
@@ -4476,7 +4476,7 @@ const AdminDefinition = {
...
+                                help_text_default: '1. [Log in](!https://accounts.google.com/login) to your Google account.\n2. Go to [https://console.developers.google.com](!https://console.developers.google.com), click **Credentials** in the left hand side.\n 3. Under the **Credentials** header, click **Create credentials**, choose **OAuth client ID** and select **Web Application**.\n 4. Enter "Mattermost - your-company-name" as the **Name**.\n 5. Under **Authorized redirect URIs** enter **your-mattermost-url/signup/google/complete** (example: http://localhost:8065/signup/google/complete). Click **Create**.\n 6. Paste the **Client ID** and **Client Secret** to the fields below, then click **Save**.\n 7. Go to the [Google People API](!https://console.developers.google.com/apis/library/people.googleapis.com) and click *Enable*.',
```
furqanmlk, line 4479: Does not match with expected instructions posted in PR.

components/admin_console/admin_definition.jsx
```diff
@@ -4476,7 +4476,7 @@ const AdminDefinition = {
                                 display_name: t('admin.openid.google'),
                                 display_name_default: 'Google Apps',
                                 help_text: t('admin.google.EnableMarkdownDesc'),
```
hahmadia, line 4478: Hey Scott!

You may want to rename these ID's for all the OpenID since they will conflict with the OAUTH. For example, `admin.google.EnableMarkdownDesc` is also used for OAuth. It may be better to rename this to `admin.google.openid.EnableMarkdownDesc` to avoid the conflict. This is also true for all the providers which I just noticed. It could be a contributing factor to the above problem mentioned by Malik.

hahmadia (commented):
Small suggestion made in regards to above mentioned error.

components/admin_console/admin_definition.jsx
sbishel: Oops. I posted the wrong screenshot. Screenshot is correct now.

components/admin_console/admin_definition.jsx
sbishel: This was actually intentional since the steps for both should be the same.

furqanmlk (approved):
All looks good ✅ 

**Instructions on OAuth**
image

**Instructions on OpenID**
image

PR #7345, opened 2021-01-19
Removing the preview limit for PDF files

jespino:
#### Summary
Removing the preview limit for PDF files. I also made it more lazy about loading and rendering pages to allow very long pdfs to be rendered.

#### Ticket Link
[MM-30856](https://mattermost.atlassian.net/browse/MM-30856)


components/pdf_preview.jsx
```diff
@@ -40,14 +38,20 @@ export default class PDFPreview extends React.PureComponent {
...
         };
-
-        for (let i = 0; i < MAX_PDF_PAGES; i++) {
-            this[`pdfCanvasRef-${i}`] = React.createRef();
-        }
     }
 
     componentDidMount() {
         this.getPdfDocument();
+        this.parentNode = document.querySelector('.modal .modal-image .modal-image__content');
```
jespino, line 45: This is far from be optimal, a workaround for this is to pass the scroll viewport information (top and bottom) from the parent component.

components/pdf_preview.jsx
```diff
@@ -111,19 +133,35 @@ export default class PDFPreview extends React.PureComponent {
...
     }
 
-    onPageLoad = (page) => {
+    loadPages = async (pdf) => {
+        const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+        for (let i = 0; i < pdf.numPages; i++) {
+            await this.loadPage(pdf, i); // eslint-disable-line no-await-in-loop
+
+            // Give time to render the interface
+            await sleep(10); // eslint-disable-line no-await-in-loop
```
jespino, line 154: If i don't do this, this takes a long time to render for big documents because it consumes all the cpu time loading the pdf files and is not able to render.

esethna (approved):
This is awesome, thanks Jesus!

components/pdf_preview.jsx
hmhealey: This is needed because the parent div has `this.container` scroll within it, right? I think ideally we'd make this component scroll, but I'm fine doing it this way since the rest of those modals needs work anyway.

Could we possibly get the parent by calling `this.container.parentNode` instead of using the query selector? That might be a bit less fragile

components/pdf_preview.jsx
```diff
@@ -111,19 +133,35 @@ export default class PDFPreview extends React.PureComponent {
...
     }
 
     onDocumentLoadError = (reason) => {
         console.log('Unable to load PDF preview: ' + reason); //eslint-disable-line no-console
         this.setState({loading: false, success: false});
     }
 
-    onPageLoad = (page) => {
+    loadPages = async (pdf) => {
+        const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
```
hmhealey, line 149: Maybe we could use `requestAnimationFrame` in here instead so that it's throttled by how fast the browser re-renders? That might be a bit better since it'll trigger slower if the user's machine is slower.

hmhealey (commented):
Looks great! I had a few suggestions for things that might improve stuff, but feel free to ignore those entirely

components/pdf_preview.jsx
jespino: Very good point.

jespino:
@hmhealey the parent problem is the `this.container.current` is null on mount, because the loading state. But I can try to wrap everything inside that "container element" or maybe reuse the reference in multiple places.

components/pdf_preview.jsx
jespino: I tried, and doesn't work well, I going to keep my approach because is working correctly and is not hard to understand.

components/pdf_preview.jsx
jespino: Improved it.

components/pdf_preview.jsx
```diff
@@ -40,14 +38,19 @@ export default class PDFPreview extends React.PureComponent {
...
-
-        for (let i = 0; i < MAX_PDF_PAGES; i++) {
-            this[`pdfCanvasRef-${i}`] = React.createRef();
-        }
     }
 
     componentDidMount() {
         this.getPdfDocument();
+        if (this.container.current) {
+            this.parentNode = this.container.current.parentElement.parentElement;
```
hmhealey, line 46: I think you accidentally removed the code to add the scroll handler to the parent

hmhealey (changes requested):

calebroseland (approved):
LGTM! Maybe one opportunity for reusability, but not a blocker by any means.

components/pdf_preview.jsx
```diff
@@ -84,15 +91,33 @@ export default class PDFPreview extends React.PureComponent {
...
+    isInViewport = (page) => {
+        const bounding = page.getBoundingClientRect();
+        const viewportTop = this.container.current.scrollTop;
+        const viewportBottom = viewportTop + this.container.current.parentElement.clientHeight;
+        return (
+            (bounding.top >= viewportTop && bounding.top <= viewportBottom) ||
+            (bounding.bottom >= viewportTop && bounding.bottom <= viewportBottom) ||
+            (bounding.top <= viewportTop && bounding.bottom >= viewportBottom)
+        );
+    };
```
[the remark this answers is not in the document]
jespino: Good point, lets move it there :)

hmhealey (approved):

ogi-m:
Tested with a 75MB, 77 page PDF with lots of images. The doc takes a few seconds to load but after that the scrolling is smooth unless you scroll quickly, in which case it can get choppy. In my view this is acceptable as it is an extreme case.

With a 2.5MB, 35 page PDF it was smooth without any issues at all.

@jespino There is one issue only - attempting to drag the scrollbar on the 77 page doc doesn't work at all in Chrome & Desktop app (but works with the smaller doc).

components/pdf_preview.jsx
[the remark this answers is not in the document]
jespino: I'm thinking again about it, and I think moving that to `utils` without a reason can generate more problems than solutions. If we see this behavior appearing again in other places we can move it later.

jespino:
@ogi-m I did some performance improvements, now the document should work better (Is going to have an extra small delay on rendering because is loading the pages on the fly, but is going to work way better with big documents).

calebroseland:
The issue with scrolling seems to be when there are multiple attachments, so the arrows appear, which covers the scrollbar entirely.
CleanShot 2021-02-03 at 17 28 13@2x

ogi-m (approved):
Good catch, thanks for logging that Caleb!

Performance improvement is definitely noticeable, way smoother! LGTM 👍🏻

calebroseland (approved):

PR #7348, opened 2021-01-19
[MM-32167] Alphabetical and recency sorted categories will now have the same drop behaviour as DMs

devinbinnie:
#### Summary
When dragging and dropping a channel from one category into a category that has a specific sort order set (ie. Alphabetical, Recency), the default drag and drop behaviour is to move the current channels in the category around the new channel when dragging around, giving the user a perception that the new channel would be placed in that exact spot, which wouldn't make sense if the channel is automatically sorted in some way.

This PR takes the same behaviour for dragging and dropping used for the DM category, which is to just outline the category and fade the channel in as it is dropped, and applies that behaviour to other categories that are automatically sorted in the same way.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32167


esethna:
@amyblais this should be for cloud Jan 26 + v5.32

jgilliam17:
Thanks @devinbinnie 
Seeing one issue with the timing/location the outline is triggered, when dragging into a category that has a specific sort order.  For Example, select specific sort order on 2 custom categories, drag from top category into a lower one over category name. Observe the outline does not trigger until you pass the category name. 
This also impacts the ability to drop channels into collapsed category, when dragging channels from top to bottom on the sidebar.
note: Not seeing on Favorites category though.

https://user-images.githubusercontent.com/52937121/105191984-ab30d380-5b05-11eb-8c18-71b94d32b02b.mov

devinbinnie:
> Thanks @devinbinnie
> Seeing one issue with the timing/location the outline is triggered, when dragging into a category that has a specific sort order. For Example, select specific sort order on 2 custom categories, drag from top category into a lower one over category name. Observe the outline does not trigger until you pass the category name.
> This also impacts the ability to drop channels into collapsed category, when dragging channels from top to bottom on the sidebar.
> note: Not seeing on Favorites category though.
> 
>  Screen.Recording.2021-01-20.at.9.49.55.AM.mov

Yeah, this is going to be a problem. The issue is the same one we've encountered with the DM category where dragging into it causes the hitbox not to register immediately. Now unfortunately we can have any arbitrary number of categories like that which is going to mess up dragging even more.

I'll try and see if I can find a solution to this once and for all.

devinbinnie:
> > Thanks @devinbinnie
> > Seeing one issue with the timing/location the outline is triggered, when dragging into a category that has a specific sort order. For Example, select specific sort order on 2 custom categories, drag from top category into a lower one over category name. Observe the outline does not trigger until you pass the category name.
> > This also impacts the ability to drop channels into collapsed category, when dragging channels from top to bottom on the sidebar.
> > note: Not seeing on Favorites category though.
> > Screen.Recording.2021-01-20.at.9.49.55.AM.mov
> 
> Yeah, this is going to be a problem. The issue is the same one we've encountered with the DM category where dragging into it causes the hitbox not to register immediately. Now unfortunately we can have any arbitrary number of categories like that which is going to mess up dragging even more.
> 
> I'll try and see if I can find a solution to this once and for all.
> 
> Link to original bug for context: https://mattermost.atlassian.net/browse/MM-26399

@jgilliam17 I think I have a fix for it, can you have a look and see if that works better for you?

jgilliam17:
@devinbinnie 
Sticky headers are interfering with the category outline on Chrome, Edge and desktop app (not on Safari or Firefox, no sticky headers) 
I probably missed this earlier

https://user-images.githubusercontent.com/52937121/105229970-b5b59200-5b32-11eb-9a4e-8b9e88a14ddc.mov

devinbinnie:
@jgilliam17 Should be fixed now

jgilliam17:
Thanks @devinbinnie 
The category name/sticky header issue is fixed, looks good. 
One small UX issue
I am not seeing a complete outline on collapsed category, bottom is missing
Screen Shot 2021-01-21 at 2 54 11 PM

components/sidebar/sidebar_category/sidebar_category.tsx
```diff
@@ -218,6 +218,23 @@ export default class SidebarCategory extends React.PureComponent<Props, State> {
...
     }
 
+    showPlaceholder = () => {
+        const {channels, draggingState, category, isNewCategory} = this.props;
+
+        if (category.sorting === CategorySorting.Alphabetical ||
+            category.sorting === CategorySorting.Recency ||
+            isNewCategory) {
+            // Always show the placeholder if the channel being dragged is from the current category
+            if (channels.find((channel) => channel.id === draggingState.id)) {
```
hmhealey, line 228: Oh, so this is why the drop target hitbox for the DMs category (and presumably new categories) was incorrect before?

selectors/views/channel_sidebar.ts
```diff
@@ -47,6 +47,18 @@ export const getCategoriesForCurrentTeam: (state: GlobalState) => ChannelCategor
     });
 })();
 
+export const getAutoSortedCategoryIds: (state: GlobalState) => string[] = (() => {
+    const getCategoriesForTeam = makeGetCategoriesForTeam();
+
+    return memoizeResult((state: GlobalState) => {
```
hmhealey, line 53: `createSelector` would be better here since it will prevent recalculating when state changes but `categories` does not. The reason we didn't do that in some of the other selectors is because they didn't have any equivalent like to the filter or because they required chaining results of more.

```javascript
createSelector(
    (state) => getCategoriesForTeam(state, getCurrentTeamId(state)),
    (categories) => categories.filter(...),
);
```

selectors/views/channel_sidebar.ts
```diff
@@ -47,6 +47,18 @@ export const getCategoriesForCurrentTeam: (state: GlobalState) => ChannelCategor
...
+        const currentTeamId = getCurrentTeamId(state);
+        const categories = getCategoriesForTeam(state, currentTeamId);
+        return categories.filter((category) =>
```
hmhealey, line 56: Also, perhaps this should be wrapped in a set to make checking membership a bit quicker?

hmhealey (changes requested):

components/sidebar/sidebar_category/sidebar_category.tsx
devinbinnie: I'm guessing so, I didn't realize there was a still a placeholder left on the original category when the channel left. It wasn't noticeable since it wasn't taking up space, but maybe it needed to be there for `rbd` to work correctly.

deanwhillier (approved):

devinbinnie:
@deanwhillier @jgilliam17 Okay so I've fixed the issue by adding a hack, forcing a `min-height` of 2px on the category content.

In order to make sure the box-shadow shows behind the sidebar header button, we need to force it to have its own box-shadow. This also applies when the header is sticky and needs to continue to show an outline when its parent div is hidden during scrolling. However, we only want it to show the top, left and right sides of the box-shadow on the header, because if there are unreads in the channel, the border will show on the bottom of the header, splitting the header and the content. When the category is collapsed however, the entire parent div is taken up by the header button, blocking off the bottom part of the box-shadow that would normally show. Thus, we need the extra 2px minimum at the bottom of the category content to ensure the bottom of the box-shadow always shows.

If we were able to make the box-shadow show overtop of the button or the header div, that would solve the issue. The reason it does is due to sticky headers requiring an explicit background to be set on the header so that channels do not show underneath it. Removing that would allow the box-shadow to work, but would break the sticky header.

Asking for re-review on CSS in case there are any better solutions, but I think this one impacts UX as minimally as possible.

jgilliam17:
Devin - category drop border looks good; also the issues with collapsed categories border is fixed.
- There's a one new issue with moving channels around in the manually sorted category that I haven't seen on master or daily and not consistently reproducing on the PR. Video is from Firefox, but I also saw this on Chrome.
I can file separately for investigation if it is not related to your changes. Let me know.


https://user-images.githubusercontent.com/52937121/105447619-89893680-5c42-11eb-8239-d1e61b338be4.mov

devinbinnie:
> Devin - category drop border looks good; also the issues with collapsed categories border is fixed.
> 
> * There's a one new issue with moving channels around in the manually sorted category that I haven't seen on master or daily and not consistently reproducing on the PR. Video is from Firefox, but I also saw this on Chrome.
>   I can file separately for investigation if it is not related to your changes. Let me know.
> 
>  Screen.Recording.2021-01-21.at.11.12.03.PM.mov

Hmm, looks like there's something odd going on with respect to actually re-ordering the channels. @hmhealey @nevyangelova Any thoughts?

nevyangelova:
> > Devin - category drop border looks good; also the issues with collapsed categories border is fixed.
> > 
> > * There's a one new issue with moving channels around in the manually sorted category that I haven't seen on master or daily and not consistently reproducing on the PR. Video is from Firefox, but I also saw this on Chrome.
> >   I can file separately for investigation if it is not related to your changes. Let me know.
> > 
> > Screen.Recording.2021-01-21.at.11.12.03.PM.mov
> 
> Hmm, looks like there's something odd going on with respect to actually re-ordering the channels. @hmhealey @nevyangelova Any thoughts?

I can't personally reproduce this issue on the test server but I also don't see the manual sorting logic in 'selectors/entities/channel_categories.ts' in MM-redux. What would classify as "Default"? 

```
  if (category.sorting === CategorySorting.Recency) {
            channels = sortChannelsByRecency(state, channels);
        } else if (category.sorting === CategorySorting.Alphabetical || category.sorting === CategorySorting.Default) {
            if (channels.some((channel) => channel.type === General.DM_CHANNEL || channel.type === General.GM_CHANNEL)) {
                channels = sortChannelsByNameWithDMs(state, channels);
            } else {
                channels = sortChannelsByName(state, channels);
            }
        }
```

hmhealey:
@jgilliam17 There's two things I could see causing that: either the UI and redux are out of sync for what the sorting order of that category actually is or there's an error somewhere that's causing the server to prevent you from reordering those channels similar to what was happening in the past when the server prevented you from reordering categories with deleted channels in them. Are there any errors in the logs when that happens?

That said, that wouldn't be affected at all by this PR, so could we file a separate ticket for that?

@nevyangelova "Default" doesn't have any option in the UI. It's treated the same as alphabetical, except when the user starts to rearrange the sidebar, it updates the `Category.sorting` to manual. I had to add that way back so that the sidebar would behave the same as before after migration (ie it would sort alphabetically), but we wanted it to start manually sorting as soon as the user tried reordering their channels.

hmhealey (approved):

jgilliam17 (approved):
Thanks @devinbinnie 
Tested, looks good to merge 
- Verified categories sorted by recency and alphabetically have the same behavior as DM category; the category outlines when channels are dropped in and dropped channels are automatically sorted per current category sort order.
Harrison, Nevy - thanks for checking on the issue I mentioned. I am no longer able to reproduce and unfortunately didn't check the console when the issue first occurred.
I will re-test on daily/cloud after merge and also file a follow up for investigation.

deanwhillier (approved):

PR #7351, opened 2021-01-20
Promoting tests to prod from unstable - part I

prapti:
Promoting some tests that have been consistently passing in the last five runs to prod from unstable.
There are more tests that need to be looked into more in detail, thus doing it in parts for this week.
Also, note that I have added a comment in `token_spec` for MM-T1880 since there was a duplicate implementation of that test case; we'd want to keep the test from `bot_api_spec` but that file is not stable enough for prod, thus promoting this one for now so that it eliminates the need to run the test manually and has been passing consistently.

### DIFF
--- a/e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
+++ b/e2e/cypress/integration/account_settings/display/account_settings_position_spec.js
@@ -7,7 +7,9 @@
 // - Use element ID when selecting an element. Create one if none.
 // ***************************************************************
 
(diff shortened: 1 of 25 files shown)


saturninoabril (approved):
Thanks Prapti!

josephbaylon (approved):
LGTM. Thanks @prapti !!

furqanmlk (approved):
looks good to me

PR #7352, opened 2021-01-20
[MM-12290] inline image collapse

TheDarkestDay:
#### Summary
The purpose of this PR is to introduce capability for inline Markdown image to be expanded/collapsed. Moreover, they should take into account user preferences in terms of default embedded content display mode - either collapsed or expanded.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/16636

#### Related Pull Requests
N/A

#### Screenshots
Hovering large inline image now displays a special "Collapse" button in the top-left corner of the image:
image

You can collapse it by clicking on the button and then expand it again by clicking on a special collapsed view of the inline image:
image


components/markdown_image_expand/markdown_image_expand.scss
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  .markdown-image-expand--expanded:hover & {
+    display: block;
+  }
+
+  &:hover {
+    border: 1px solid var(--center-channel-color-24);
+
+    @include box-shadow($elevation-2);
```
matthewbirtch, line 43: For some reason `$elevation-2` is rendering the shadow opacity at `0.012` and it should be rendering as `0.12`

matthewbirtch (changes requested):
Thanks @TheDarkestDay! This is great. Just a few tweaks from my standpoint:

1. The image has a border on it when you hover on it. Can we remove that?
2. Is it possible to animate the collapsing transition? This would help orient users. If not, we could look at that in a separate PR
3. Can we reduce the margin on the top of `.markdown-inline-img` to 2px instead of 5px? That should even out the positioning of the arrow button

components/markdown_image_expand/markdown_image_expand.scss
TheDarkestDay: Alright. I'll fix this. By the way, all of the elevations starting from 2 to 6 have an opacity like this. Probably we have here some sort of wide spread issue.

TheDarkestDay:
> Thanks @TheDarkestDay! This is great. Just a few tweaks from my standpoint:
> 
> 1. The image has a border on it when you hover on it. Can we remove that?
> 2. Is it possible to animate the collapsing transition? This would help orient users. If not, we could look at that in a separate PR
> 3. Can we reduce the margin on the top of `.markdown-inline-img` to 2px instead of 5px? That should even out the positioning of the arrow button

Thanks @matthewbirtch! Okay, let me try to tackle this step by step.

I've been digging around point 3 and the only question I am currently having - should we apply this change universally to all of the Markdown inline images or should just create some special edge case for this one? 

I tried applying this thing globally and I didn't manage to find any specific issues caused by this change. On the other hand, for sure I might be missing some use case which would really explain the purpose of this margin setting.

What do you think, does it make sense to apply this change universally, push it here and let you and the team check it?

TheDarkestDay:
> 1. The image has a border on it when you hover on it. Can we remove that?

Oh, that's the same story, actually.

The thing is that each inline Markdown image is supposed to be clickable - and on click it should be shown in full size in a modal window. And I guess the reason for this shadow-on-hover effect is exactly to give a hint that this image is clickable.

So, the question is - should we treat these "images with expand" somehow different? Should we remove this thing for them or rework somehow?

TheDarkestDay:
> 2. Is it possible to animate the collapsing transition? This would help orient users. If not, we could look at that in a separate PR

I think at least I could try. The only thing I am concerned about here is performance of height change animation - but we will only find it out if we try it in the wild. 

May I kindly ask you to share some assets for this animation? Things like duration, easing function, maybe even some gif would be helpful.

components/markdown_image_expand/markdown_image_expand.scss
matthewbirtch: Seems like a mistake to me. Not sure when this was added

matthewbirtch:
> I've been digging around point 3 and the only question I am currently having - should we apply this change universally to all of the Markdown inline images or should just create some special edge case for this one?
> 
> I tried applying this thing globally and I didn't manage to find any specific issues caused by this change. On the other hand, for sure I might be missing some use case which would really explain the purpose of this margin setting.
> 
> What do you think, does it make sense to apply this change universally, push it here and let you and the team check it?

Ok, let's leave this as-is for now and we can look in to it later if we feel strongly that it needs to change.

> The thing is that each inline Markdown image is supposed to be clickable - and on click it should be shown in full size in a modal window. And I guess the reason for this shadow-on-hover effect is exactly to give a hint that this image is clickable.
> 
> So, the question is - should we treat these "images with expand" somehow different? Should we remove this thing for them or rework somehow?

Makes sense. Let's leave the border and shadow effect

> I think at least I can try. The only thing I am concerned about here is performance of height change animation - but we will only find it out if we try it in the wild.
> 
> May I kindly ask you to share some assets for this animation? Things like duration, easing function, maybe even some gif would be helpful.

Give these a go:
```
transition: height 250s cubic-bezier(0.5, 0, 0.5, 1);
```

We may need to tweak the above, but we'll see when it's in the wild.

TheDarkestDay:
@matthewbirtch feel free to have a look at the animation. I am not exactly sure that this is the behavior we would like to achieve but let's at least use this as some sort of starting point.

components/markdown_image_expand/markdown_image_expand.scss
TheDarkestDay: Fixed only `$elevation-2` for now. 

@asaadmahmood I am sorry to bother you but according to Git blame you were the author of these `elevation` settings. May I request some assistance from you - are these opacity values incorrect or there were some reasoning behind this?

matthewbirtch:
Thanks @TheDarkestDay.

- It looks like the animation may be too fast. Maybe we need to make it 300ms in duration.
- I think we can constrain the animation to the height and have it crop the image as the height of the container shrinks.
- It feels a little jittery due to the way the transition happens between the title and collapsing image
  - It looks like the title animates in from the right which seems odd. This element doesn't need to move horizontally at all

If this is adding too much complexity to this PR, we can consider leaving the animation out of this and we could look to enhance with animation in another ticket. Thoughts?

esethna:
@TheDarkestDay one other thing I noticed is that using the /collapse /expand slash commands should cause the inline markdown images to collapse or expand in the view, similar to image link previews (ie posting https://www.mattermost.org/wp-content/uploads/2016/03/logoHorizontal.png)

image

hmhealey:
For expanding/collapsing with those commands, we'd need to make it so that we store the expanded/collapsed state in local storage like the collapsed state for link previews. If it's storing the expanded/collapsed state in the same place, then those commands as well as the "expand/collapse by default" user setting should just work with these images too.

TheDarkestDay:
@esethna thanks for bringing this up. I was completely unaware of this feature in Mattermost. 

I've fixed that - as far as I got for know expand/collapse settings are common for both link previews as well as these Markdown inline images. I assume that this should be fine, right? @hmhealey

TheDarkestDay:
> If this is adding too much complexity to this PR, we can consider leaving the animation out of this and we could look to enhance with animation in another ticket. Thoughts?

I will try my best to play around this during weekend - just in order to be sure that this thing truly deserves separate ticket.

esethna (approved):
Approved from PM standpoint, I'll leave the UX tweaks to you and Matt. Thanks @TheDarkestDay!

components/markdown_image_expand/markdown_image_expand.scss
```diff
@@ -0,0 +1,95 @@
...
+
+.markdown-image-expand__collapse-button {
+    position: absolute;
+    top: 0;
+    left: 0;
+
+    display: none;
+    width: 24px;
+    height: 24px;
+    border: 1px solid var(--center-channel-color-16);
```
hmhealey, line 20: For these `-##` variables that include an opacity, we're trying to get away from using those since it adds a lot of extra CSS variables. Could you instead switch this to use `rgba` to specify that opacity instead? Like this one would look like
```suggestion
    border: 1px solid rgba(var(--center-channel-color-rgb), 0.16);
```

components/markdown_image_expand/markdown_image_expand.tsx
```diff
@@ -0,0 +1,82 @@
...
+    postId: string;
+};
+
+type State = {
+    isExpanded: boolean;
+};
+
+const MarkdownImageExpand: React.FC<Props> = ({children, alt, isEmbedVisible}: Props) => {
+    const [state, setState] = useState<State>({
+        isExpanded: isEmbedVisible,
```
hmhealey, line 20: Could we remove `state.isExpanded` and just have everything depend on `props.isEmbedVisible'?

components/markdown_image_expand/markdown_image_expand.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const {isExpanded} = state;
+
+    useEffect(() => {
+        setState({
+            isExpanded: isEmbedVisible,
+        });
+    }, [setState, isEmbedVisible]);
+
+    const handleToggleButtonClick = () => {
+        setState((oldState) => ({...oldState, isExpanded: !oldState.isExpanded}));
```
hmhealey, line 31: Instead of just writing back to the state, this might need to write back to the redux state and local storage so that the expanded/collapsed state is persisted through a refresh

components/markdown_image_expand/index.ts
```diff
@@ -0,0 +1,16 @@
...
+import {GlobalState} from 'mattermost-redux/types/store';
+import {connect} from 'react-redux';
+
+import {isEmbedVisible} from 'selectors/posts';
+
+import MarkdownImageExpand, {Props} from './markdown_image_expand';
+
+const mapStateToProps = (state: GlobalState, ownProps: Props) => {
+    return {
+        isEmbedVisible: isEmbedVisible(state, ownProps.postId),
```
hmhealey, line 12: I found a slight problem with doing this in this way. Since the collpased state is tied to a single post ID, this won't work if you have two Markdown images in a single post since they'll both be looking at the same state for `isEmbedVisible`. Maybe instead of just passing in a `postId`, we also pass in the image URL to differentiate the two images?

The reason this hasn't been a problem before is that we only ever have one embed in the post. Markdown images are different since you can have any number of them in a post.

components/markdown_image_expand/markdown_image_expand.tsx
```diff
@@ -0,0 +1,82 @@
...
+    return (
+        <div className={wrapperClassName}>
+            <CSSTransition
+                in={isExpanded}
+                timeout={200}
+                mountOnEnter={true}
+                unmountOnExit={true}
+                classNames='markdown-image-expand__expanded-view'
+            >
+                <div>
```
hmhealey, line 45: Is this div necessary or would it be possible to use a Fragment here to avoid the extra DOM node?

hmhealey (changes requested):

components/markdown_image_expand/index.ts
hmhealey: Hmm, actually, that might not work since a user could theoretically post the same Markdown image multiple times in a single post for whatever reason. Perhaps we could do something with the index of the image in the post? I'm not sure if we have any way for the `MarkdownImage` to get that though.

If you could find something like that, that'd be great, but if we're not able to find anything better, we'd probably be fine just sticking with the post ID and image URL together since that would work in most cases.

esethna:
@TheDarkestDay how's this going? Let us know when you're ready for another review

TheDarkestDay:
@esethna sure, thanks for getting in touch. 

I am currently working on fixes proposed by @hmhealey (thank you for review, by the way) and yet another (last, I suppose) take for collapse/expand animation. 

I'll let you know as soon as it is ready.

components/markdown_image_expand/markdown_image_expand.tsx
TheDarkestDay: Fixed.

components/markdown_image_expand/markdown_image_expand.tsx
TheDarkestDay: Fixed.

components/markdown_image_expand/markdown_image_expand.tsx
TheDarkestDay: Fixed.

components/markdown_image_expand/index.ts
TheDarkestDay: Fixed.

components/markdown_image_expand/markdown_image_expand.scss
TheDarkestDay: Fixed.

TheDarkestDay:
Hey, @hmhealey. 

I think I've covered all of the point from your last review. So, feel free to have a look.

The only difference from your proposal here is that inline images visibility data is stored inside storage fields prefixed with dedicated `INLINE_IMAGE_VISIBLE` prefix. I think it might be a good idea to setup this way of distinguishing between regular embeds preview and inline images. Semantically these things are different so that's why I have been tentative to use the same prefix for both of them. 

Let me know your opinion - if general approach is OK, I will finish this work up with some missing unit tests.

hmhealey (changes requested):
Nice! Looks good with multiple images in the post now.

package-lock.json
```diff
@@ -7106,8 +7106,7 @@
           "version": "3.0.0",
           "resolved": "https://registry.npmjs.org/get-stream/-/get-stream-3.0.0.tgz",
           "integrity": "sha1-jpQ9E1jcN1VQVOy+LtsFqhdO3hQ=",
-          "dev": true,
```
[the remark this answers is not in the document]
TheDarkestDay: Done, thanks you.

hmhealey (approved):

esethna:
@amyblais can we add this to the v5.34 milestone?

amyblais:
> @amyblais can we add this to the v5.34 milestone?

@metanerd I don't have permissions to add new milestones for webapp PRs, can you help share the permission with me?

matthewbirtch (approved):
Seems good to me. We can look at animation in another PR later

utils/message_html_to_component.jsx
```diff
@@ -109,6 +111,7 @@ export function messageHtmlToComponent(html, isRHS, options = {}) {
         processingInstructions.push({
             shouldProcessNode: (node) => node.type === 'tag' && node.name === 'img',
             processNode: (node) => {
```
lieut-data, line 113: I'm slightly worried about binding the underlying storage bit to the index being computed here. If someone goes and edits their post, the index will change, but not the storage, so users might see their configurations flip.

Would it make sense to instead use the `src` attribute as the key here? This has the downside that duplicates of /same/ src posted in the same post will get collapsed together, but that seems unlikely. Plus if the post is edited to show a /new/ inline image, the behaviour will correctly reset to the defaults for the new image.

sass/utils/_variables.scss
```diff
@@ -31,7 +31,7 @@ $transition-quick: 0.12s;
 
 // Elevations
 $elevation-1: 0 2px 3px 0 alpha-color($black, .08);
-$elevation-2: 0 4px 6px 0 alpha-color($black, .012);
+$elevation-2: 0 4px 6px 0 alpha-color($black, .12);
```
lieut-data, line 34: I followed the discussion above on this, but I'm worried about seeing this change in isolation. @matthewbirtch, can you help provide guidance here?

components/markdown_image_expand/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import {connect} from 'react-redux';
+
+import {GenericAction} from 'mattermost-redux/types/actions';
+
+import {toggleInlineImageVisibility} from 'actions/post_actions';
+import {isInlineImageVisible} from 'selectors/posts';
+
+import MarkdownImageExpand, {Props} from './markdown_image_expand';
+
+const mapStateToProps = (state: GlobalState, ownProps: Props) => {
```
lieut-data, line 18: Could we call `ownProps` just `props`? I /think/ the pattern for `ownProps` only comes into effect when defining a callback and needing to distinguish the two closures.

components/markdown_image.jsx
```diff
@@ -172,6 +175,21 @@ export default class MarkdownImage extends React.PureComponent {
                             }
                         </>
                     );
+
+                    const availableHeight = height ?? imageMetadata.height;
+                    if (availableHeight >= Constants.MAX_INLINE_IMAGE_HEIGHT) {
```
lieut-data, line 180: Can you expand on the reason to make this conditional? My impression is that embedded images have no such restriction, so it seems weird to be able to collapse some but not others?

If we keep this, I wonder if we can improve on the name `MAX_INLINE_IMAGE_HEIGHT`, which sounds like a bound on how high the inline image can be.

lieut-data (changes requested):
This is really great, @TheDarkestDay! I don't have as much context coming into this review, so I've asked a few clarifying questions below to get up to speed.

sass/utils/_variables.scss
matthewbirtch: @lieut-data `.12` was incorrectly coded as `.012`  value previously and had just never been caught.

asaadmahmood: @lieut-data I agree with @matthewbirtch, it was incorrectly coded, and is not used anywhere.

components/markdown_image.jsx
TheDarkestDay: I think this is more of a product question. It has been an initial requirement of this ticket to apply this expand/collapse behavior only for images taller than 100px. I assume, there should be some specific use case revolving around using small un-collapsable inline images (maybe some emojis or gifs?). Pinging @esethna for confirmation.

Regarding the name - yes, I got your point. I will work through it, thanks.

utils/message_html_to_component.jsx
TheDarkestDay: I have always been tentative about this. I was simply trying to assess which behavior seems more "unacceptable" or confusing from a user perspective - either two images sharing the same collapse/expand state (which will happen in case of using src as a key) or shuffling this state completely in case of image editing (when we use indexes instead).

I think I agree with you that index-based approach causes more "harm" in terms of usability. Because post editing will happen for sure, whereas multiple collapseable inline images with the same URL don't seem to appear that likely. 

I will fix this shortly.

TheDarkestDay:
> Seems good to me. We can look at animation in another PR later

I am terribly sorry that I didn't reach out to you regarding animations. Okay, let's move it to separate ticket, since I anyway haven't managed to find anything feasible.

sass/utils/_variables.scss
lieut-data: Thanks, @matthewbirtch & @asaadmahmood -- should we go ahead and ask @TheDarkestDay to change all of these `elevation-*` values to be correct? Otherwise I fear regressing here by some future "fix" to what /now/ looks like a bug.

asaadmahmood: I don't mind changing them in another PR, or actually having it done in this one. I'm fine either way. @lieut-data

lieut-data: Thanks @asaadmahmood -- to confirm, should `$elevation-1` have an alpha component of `.08` and all of `$elevation-2 .. $elevation-6` have an alpha component of `.12`?

components/markdown_image.jsx
TheDarkestDay: Fixed.

components/markdown_image_expand/index.ts
TheDarkestDay: Kinda fixed.

utils/message_html_to_component.jsx
TheDarkestDay: Fixed.

TheDarkestDay:
Hey, guys. I've fixed most of the latest issues related to key used for keeping expand/collapse state in local storage. On the other hand, I think I've found something concerning from a QA perspective.

At first, I added couple of inline images into one post:
image

Then I decided to expand both of them and I ended up looking at this post with content being cut, just like this:
image

Meanwhile I can observe a special behavior for regular posts which simply hold a lot of text content. They can be expanded/collapsed via "Show more / Show less" buttons. 

So, I guess ideally we would like to have the same "Show more / Show less" feature enabled for posts whose height can be dynamically changed due to inline images expansion or collapse, right?

matthewbirtch:
> So, I guess ideally we would like to have the same "Show more / Show less" feature enabled for posts whose height can be dynamically changed due to inline images expansion or collapse, right?

That's a good catch @TheDarkestDay. I think you're right. We should account for this

utils/message_html_to_component.jsx
hmhealey: That's a good point about editing posts making it weird. The problem with just using the URL is that having the same image in a post twice will cause both of them to expand/collapse together. That said, that's probably a less common case than editing a post with a URL, so I'd think it's better to have work this way even if there is a weird case.

Ideally, I guess out best bet would be to add an index for each time an image URL occurs in a post, so that the first and second instance of each image would be tracked separately. That may still be weird if the user heavily edits a post and, for example, removes one instance of the image, but I think that's probably the best we could do since you're otherwise having to spend a lot more time trying to figure out whether or not the user meant to remove the first or second image which is obviously very hard.

That all said, I think the per-URL method probably makes the most sense for now since it'll cover 99% of cases. I think the case I mentioned is enough of an edge case that we can accept it as a known issue to fix if someone comes up with a more practical use case when that might occur.

hmhealey:
Oh yeah, I hadn't thought of that. We're not triggering the code that checks to show the Show More button in that case ([link](https://github.com/mattermost/mattermost-webapp/blob/master/components/post_view/post_message_view/post_message_view.tsx#L61)) since the image has already been loaded at that point. The naming is a bit unfortunate, but I suppose we could just have the `onImageLoaded` prop called when the image expands or collapsed. That should work since the `MarkdownImage` already has access to that callback anyway.

TheDarkestDay:
> Oh yeah, I hadn't thought of that. We're not triggering the code that checks to show the Show More button in that case ([link](https://github.com/mattermost/mattermost-webapp/blob/master/components/post_view/post_message_view/post_message_view.tsx#L61)) since the image has already been loaded at that point. The naming is a bit unfortunate, but I suppose we could just have the `onImageLoaded` prop called when the image expands or collapsed. That should work since the `MarkdownImage` already has access to that callback anyway.

Got it. Yep, I've tried doing it this way. So far seems to be working, but there is only thing that I would like to consult about. `MarkdownImage` component passes an object containing image dimensions via `onImageLoaded` prop:
https://github.com/mattermost/mattermost-webapp/blob/master/components/markdown_image.jsx#L79

However, there is one party which does not really expect receiving an object like this. This component is called `PostMessageView` 
https://github.com/mattermost/mattermost-webapp/blob/master/components/post_view/post_message_view/post_message_view.tsx#L61

and it actually expected to get only `height` as a dimension, without wrapping object (`handleHeightReceived` gets passed to `MarkdownImage` props via `this.imageProps`). 

As far as I got, this thing has not been working at all so far. Is it safe to simply migrate to obtaining dimensions object from `onImageLoaded`? 

On top of that - what was the actual point of recalculating height change only when `height` is greater than zero? Is it possible to get it negative for some reason?

hmhealey:
I don't think we can completely change the type of `handleHeightReceived` unfortunately since it's also passed into plugins via the Pluggable component in there. Perhaps we could implement a separate method that takes a dimensions object and calls `handleHeightReceived` from that? Then we could pass that in `imageProps` instead.

I don't actually know why we only recalculate when height is greater than 0. It dates back to https://github.com/mattermost/mattermost-webapp/pull/1606, but that was added quite a while ago, and I'm not sure if @saturninoabril would remember why he added that.

TheDarkestDay:
Thanks for guidance, @hmhealey. 

At this moment it looks as if we don't need any information regarding image dimensions in order to figure out, whether we should check post overflow or not. 

That's why I've decided to extend `MarkdownImage` API to expose another prop called `onImageHeightChanged` which is intended to be used when image changes its height during runtime for some reason. Like in our case, this change might occur because of somer special user interaction like expanding/collapsing. So, now `PostMessageView` utilizes this callback as a simple trigger, doing overflow checks when this callbacks gets called.

saturninoabril:
> I don't actually know why we only recalculate when height is greater than 0. It dates back to #1606, but that was added quite a while ago, and I'm not sure if @saturninoabril would remember why he added that.

I've added that for plugin that changes post message view like in the case of [autotranslate plugin](https://github.com/mattermost/mattermost-plugin-autotranslate/blob/master/webapp/src/components/translated_message/translated_message.jsx#L21-L27) where the height changes depending on returned text translation. `handleHeightReceived` might not be named properly specially the `height` parameter since it's not really an actual height change of a text but just a number for the webapp to trigger recompute and eventually render post message with translation correctly (not being cut-off) , and render "Show more" / "Show less" as necessary. This was done so that height change is handled by the webapp and not by the plugin.

Before translation:
Screen Shot 2021-03-08 at 7 22 38 PM

After translation:
Screen Shot 2021-03-08 at 7 06 41 PM

esethna:
@hmhealey @lieut-data gentle ping on this, I'm not clear on next steps here or if it's ready to move to QA review?

lieut-data:
Thanks, @esethna -- taking a look now :)

sass/utils/_variables.scss
lieut-data: @asaadmahmood (or @matthewbirtch), can you confirm my interpretation of this thread and the final `$elevation-*` values above?

components/markdown_image_expand/markdown_image_expand.tsx
```diff
@@ -0,0 +1,65 @@
...
+    onToggle: (isExpanded: boolean) => void;
+    actions: {
+        toggleInlineImageVisibility: (postId: string, imageKey: string) => void;
+    };
+};
+
+const MarkdownImageExpand: React.FC<Props> = ({children, alt, isExpanded, postId, actions, onToggle, imageKey}: Props) => {
+    const {toggleInlineImageVisibility} = actions;
+
+    useEffect(() => {
```
lieut-data, line 21: This `useEffect` block /feels/ wrong -- listening a prop so we can trigger a prop callback wouldn't, ideally, be the responsibility of any component.

Does it suffice to invoke `onToggle` only within `handleToggleButtonClick`? Presumably the "first render" will already have the necessary state for `isExpanded`, allowing the post height computations to occur organically. Thereafter, I'd expect the prop to change only because we toggle the visibility via the click event.

Thoughts?

lieut-data (changes requested):
Thanks for your patience on this one, @TheDarkestDay! Really appreciate the simplifications you've helped drive here.

A few more comments below.

components/markdown_image_expand/markdown_image_expand.tsx
TheDarkestDay: Oh yeah, actually this is a pretty big item. Let me try to explain myself.

So, the thing is that we execute `onToggle` callback we need to provide some data inside. Which is - current toggle state (whether it is expanded or collapsed). 

And a big question here - where should this "computational" logic reside? I mean, logic which should compute next state for our toggle out of the old one? 

Correctly our logic for this state toggling resides within action creators (`toggleInlineImageVisibility` specifically). If I get your approach correctly, in this case we will have to either:

1. Duplicate this logic inside this component as well (do something like `onToggle(!isExpanded)`)
2. Or maybe delegate this logic exclusively to the component and use store just as a storage

Point 2 has come to mind just now to be honest. On the other hand, this kind of "store action isolation inside one component" might be bad - as soon as we wanted to toggle inline image visibility from some other component.

TheDarkestDay: The main thing I don't like about this implementation is data flow which looks kinda bi-directional in this case.
Let's assume that user click "Expand" button - here how the flow goes:
1. Action gets dispatched to store
2. Store propagates new `isExpanded` state to a `MarkdownImageExpand`
3. `MarkdownImageExpand` tracks update (via this `useEffect`) and propagates these changes back to `PostMessageView` component

I have been thinking about flattening this flow to something unidirectional. Can we select something from a store that would for sure change when visibility of some inline image changed and pass it as a prop for `PostMessageView`?

I guess we can somehow select all of the storage elements prefixed with `INLINE_IMAGE_VISIBILITY` and... compare previous visibility array with new one? 

I am sorry if I am overthinking this.

hmhealey: @lieut-data I think the thing that you're missing with those two props is that one comes from the parent component (`onToggle`) and the other comes from redux (`isExpanded`), so it's not really possible for the parent to be responsible for both.

I'm not a huge fan of how it works right now, but I can't see any way around it unless we can give the parent component some concept of "this is the expanded/collapsed state of my children" like @TheDarkestDay is suggesting, but that does seem harder to represent given how we store the data in local storage. I personally think the current way is our best bet since it's almost equivalent to treating `isExpanded` as if it was part of the state of the `MarkdownImageExpand` except for it being stored in redux.

lieut-data: @hmhealey, yeah the source of the props threw me for a loop the first time around, since I didn't see it being passed by the parent and only later realized it was tacked on via `connect`. But my comment about it feeling "wrong" was framed with that perspective already in mind: triggering an event from a prop change to a prop callback seems like an anti-pattern regardless of the source of those props.

**But** I suppose this is really just the hook equivalent of a `componentDidUpdate` lifecycle event that detects the props change. And, for better or worse, we do have this pattern all over the code. So I'm fine leaving as-is :)

components/markdown_image_expand/index.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {bindActionCreators, Dispatch} from 'redux';
+
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
```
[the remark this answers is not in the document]
TheDarkestDay: Oops. Fixed.

TheDarkestDay:
I pulled latest changes from `master` and it looks as though these elevation values had already been updated to `0.12`. Therefore, I assume that this discussion point is resolved.

components/markdown_image_expand/markdown_image_expand.scss
asaadmahmood: @TheDarkestDay The opacity values are updated on master.
I did them on this PR:
https://github.com/mattermost/mattermost-webapp/pull/7696/files

asaadmahmood:
@TheDarkestDay Yup.

lieut-data (approved):
Nice!

jgilliam17:
Thank you @TheDarkestDay 
Looks great, few questions 
- When a post has multiple images, is it expected when clicking on one button to collapse/expand all images at the same time? Each image has a separate collapse button, it made me believe I could action each separately.
- When inline image is linked, button is part of the link and breaks the underline a bit (see image); 
Screen Shot 2021-03-26 at 11 07 00 AM

TheDarkestDay:
Thanks, @jgilliam17 
> * When a post has multiple images, is it expected when clicking on one button to collapse/expand all images at the same time? Each image has a separate collapse button, it made me believe I could action each separately.

This is some sort of drawback we have been aware of. 

The thing is that we have to store each image state (is it collapsed or not) by some key. The only viable option we had here is to use image URL as a key. This is the reason why multiple inline images with the same URL will expand and collapse simultaneously. 

Other approaches might result in more harmful UX (like collapsing/expanding images suddenly after editing) or require a way more efforts.

Our assumption was that it would not be a widespread use case to upload several inline images with the same URL. However, if you see some issues with this assumption - feel free to let us know, maybe we should rethink it.

jgilliam17:
Thank you @TheDarkestDay
I agree, post with multiple inline images may not be used often, I don't have any objections to proceeding with the current implementation.
For the second point
> * When inline image is linked...

I discussed this with @matthewbirtch, his suggestion is to remove underline on hover

TheDarkestDay:
> I discussed this with @matthewbirtch, his suggestion is to remove underline on hover

Oh, many thanks for taking care of it. I will work on this shortly.

TheDarkestDay:
Hey, guys. 

I've finished working on the getting rid of underlining, but it looks as if I am no longer able to run ESLint checks. Here is the output of my `make check-style` command:

```
Checking for style guide compliance
npm run check

> mattermost-webapp@0.0.1 check /Users/abrenchev/Documents/dev/mattermost/mattermost-webapp
> eslint --ext .js,.jsx,.tsx,.ts . --quiet --cache


Oops! Something went wrong! :(

ESLint: 7.19.0

ESLint couldn't find the plugin "eslint-plugin-mattermost".

(The package "eslint-plugin-mattermost" was not found when loaded as a Node module from the directory "/Users/abrenchev/Documents/dev/mattermost/mattermost-webapp".)

It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm install eslint-plugin-mattermost@latest --save-dev

The plugin "eslint-plugin-mattermost" was referenced from the config file in ".eslintrc.json".

If you still can't figure out the problem, please stop by https://eslint.org/chat/help to chat with the team.

npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! mattermost-webapp@0.0.1 check: `eslint --ext .js,.jsx,.tsx,.ts . --quiet --cache`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the mattermost-webapp@0.0.1 check script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/abrenchev/.npm/_logs/2021-04-02T09_00_06_732Z-debug.log
make: *** [check-style] Error 2
```

Node.js version is 10.15.3. 

Have you recently changed something in a repo workflow? Should I add some additional dependencies or maybe post this issues in some other channel?

hmhealey:
There hasn't been any changes to the dependencies in the last week or two. Our ESLint plugin doesn't exist on npm, so you won't be able to install it that easily, but it can be installed by pointing `npm install` to [the git repo directly](https://github.com/mattermost/eslint-plugin-mattermost). That said, it should be installed automatically if you remove `node_modules` and re-run `npm install`.

The linting issues from CircleCI are as follows though:
```
/root/mattermost/mattermost-webapp/components/markdown_image_expand/index.ts
  3:1  error  There should be at least one empty line between import groups                      import/order
  4:1  error  There should be at least one empty line between import groups                      import/order
  5:1  error  `react-redux` import should occur before import of `mattermost-redux/types/store`  import/order
```

TheDarkestDay:
Got latest updates from master branch. Now everything seems to be working fine.

jgilliam17 (approved):
Thank you @TheDarkestDay 
Tested, looks good to merge.
- Verified inline image collapse for images 100px or more in height; heights change animations when using slash commands for expand/collapse are happening in view; underline is removed when inline image is linked.

jgilliam17:
@hmhealey Can you please help merge. Thanks 🙂

hmhealey:
Merged! Thanks again for the PR, @TheDarkestDay. This will be on Mattermost Cloud some time in the next couple weeks and then in (I think) Mattermost 5.35 after that.

PR #7364, opened 2021-01-21
Cypress/E2E: Fix MM-T924 by creating enough users to page through

saturninoabril:
#### Summary
Fix MM-T924 by creating enough users to page through

#### Ticket Link
failed on daily Cypress unstable test run

### DIFF
--- a/e2e/cypress/support/api/helpers.js
+++ b/e2e/cypress/support/api/helpers.js
@@ -1,0 +1,15 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export function buildQueryString(queryParams = {}) {
+    let queryString = '';
+    Object.entries(queryParams).forEach(([k, v], index) => {
+        if (index > 0) {
+            queryString += '&';
+        }
+
+        queryString += `${k}=${encodeURIComponent(v)}`;
+    });
+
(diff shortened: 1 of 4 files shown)


e2e/cypress/support/api/helpers.js
```diff
@@ -0,0 +1,15 @@
...
+// See LICENSE.txt for license information.
+
+export function buildQueryString(queryParams = {}) {
+    let queryString = '';
+    Object.entries(queryParams).forEach(([k, v], index) => {
+        if (index > 0) {
+            queryString += '&';
+        }
+
+        queryString += `${k}=${encodeURIComponent(v)}`;
```
josephbaylon, line 11: Shouldn't key also be encoded?

josephbaylon (approved):
LGTM. Just a minor question.

e2e/cypress/support/api/helpers.js
saturninoabril: I don't think so and honestly don't know if we're doing it in other place. In mattermost-redux, we're only encoding the value as well.  Key is something we have control/validation in server and I don't think we allow a key that is unescaped or will require to be encoded. But happy to change if necessary.

furqanmlk (approved):
All related test cases passed ✅ 
Looks good !

PR #7365, opened 2021-01-21
Cypress/E2E: Add explicit wait to cy.visit

saturninoabril:
#### Summary
Add explicit wait to `cy.visit` as `cy.visitAndWait`. It significantly improve page load faster when visiting a page directly via `cy.visitAndWait`. For some reason, Cypress is taking unreasonably longer than normal to show up a page when using `cy.visit`. It seemed to block some request/render in the process, though I couldn't figure out exactly what's causing it.
This approach is a quick hack but I recommend to use `cy.visitAndWait` by default and avoid `cy.visit` as much as possible.

Test report to follow.


e2e/cypress/support/extended_commands.js
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import * as TIMEOUTS from '../fixtures/timeouts';
+
+Cypress.Commands.add('visitAndWait', (url, options, duration = TIMEOUTS.THREE_SEC) => {
+    cy.visit(url, options).wait(duration);
```
josephbaylon, line 7: Maybe instead of explicit wait, we can try this to see if page is loaded.
```
            cy.waitUntil(() => cy.get('html').then((el) => {
                return el.find('#root > div > div').length > 0;
            }));
```

josephbaylon (approved):
LGTM. I'm ok with this for now. Just a different suggestion which works on both main app and admin console

prapti (approved):
LGTM! This could be a huge improvement. Thanks Saturn!

e2e/cypress/support/extended_commands.js
saturninoabril: I've tried this and I still see the `Loading` indicators in channel then post list view remained longer than usual. Let's have as is but this is well noted. We can change anytime in the future.

saturninoabril:
@furqanmlk I hope you don't mind but I'll go ahead and merge this as I'm currently block with the other fix I'm working. Please post any comment and we'll address it, separately.

PR #7367, opened 2021-01-21
MM-32197 Use other timestamps when autohiding DMs/GMs

hmhealey:
The last_viewed_at timestamp for channel members is only ever updated on the server to the time of the most recent post in the channel, so when the user refreshes, the priority for showing DM channels might change.

The old autohiding solved that by using some other timestamps which I'd like to remove eventually if we can make last_viewed_at work better, but in the mean time, we can still make use of them to get a better idea of when the user actually last viewed the channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32197

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1339

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18394,8 +18394,8 @@
       "integrity": "sha512-6qE4B9deFBIa9YSpOc9O0Sgc43zTeVYbgDT5veRKSlB2+ZuHNoVVxA1L/ckMUayV9Ay9y7Z/SZCLcGteW9i7bg=="
     },
     "mattermost-redux": {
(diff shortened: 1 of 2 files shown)


nevyangelova (approved):

devinbinnie (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified older and never used DMs/GMs, the ones previously cut off by auto-close setting, remain on the sidebar after refresh or team switch.

PR #7380, opened 2021-01-22
MM-32156 Load GMs for new sidebar correctly

hmhealey:
So there's two things that needed to change to get GMs loading correctly with the new sidebar:
1. `filterGMsDMs` was being called with an array of all channels, not just GMs and DMs, so `filterAutoclosedDMs` wasn't being called correctly. It was only returning the most recently viewed channels of any type based on the user's autoclose setting, and if those were all non-GMs, the GMs weren't being loaded. This is fixed by making `loadProfilesForGM` load based on what channels are visible in the sidebar.
    This also fixes a bug that would've existed with the previous version of the sidebar where old DMs/GMs which were in custom categories wouldn't have been loaded correctly either.
2. Because `loadProfilesForSidebar` now depends on the sidebar, we had to change when `DataPrefetch` actually calls that so that it waits for the sidebar to be loaded now. Hopefully this doesn't noticeably increase the amount of time it takes for posts to load.

@jgilliam17 For this, please do some testing similar to what you might've done for https://github.com/mattermost/mattermost-webapp/pull/5547. Unfortunately, loading users for the sidebar is coupled to preloading posts, so hopefully nothing has broken here. At the very least, the unit tests from before work correctly at least. I don't know if there's any E2E tests or anything like that though.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32156


devinbinnie (approved):

jgilliam17:
Thanks @hmhealey 
On my first pass, I can still see few `0`s for the GM member count on the LHS, but as soon as you click on any dm or switch team - number updates. 
I remembered to check the console while testing on the PR server, and there are 2 errors, but no apparent impact on the app and also unrelated as they repro on master. Noting here for visibility.

Screen Shot 2021-01-22 at 7 11 20 PM
Screen Shot 2021-01-22 at 7 12 22 PM
Screen Shot 2021-01-22 at 7 50 02 PM

nevyangelova (approved):
Thanks for handling this 💯

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified loading GMs and DMs on the sideabar and also preloading posts, thank you for linking the older PR.
- I am not able to reproduce `0` count on GMs consistently, only saw it once today on fresh login. I repeated this multiple times, not reproducing for me. I can create a follow up for further investigation if needed. Let me know.
Saturn and Joseph helped running E2Es on this server; I checked the results against the latest production test run - no unusual failures related to this change.

PR #7386, opened 2021-01-24
[MM-29108] Simplifies emoji reaction animation

koox00:
#### Summary

Reaction component's animation had complex local state and logic
which introduced a bug:

>  lack of synchronisation of reaction counts between RHS and the
>  center channel

This commit fixes that by simplifying Reaction's local state,
and local state updating.

**Details:**

We have no need of `userReacted` state, `sortedUsers.currentUserReacted`
prop holds that info.  Also `reactedNumber` and `unreactedNumber` local
state have been removed and now exist as simple values relative to the
prop `reactionCount`.

`displayNumber`, and `reactedClass` state has been kept so we can show
an animation when the user reacts upon a post.  Synchronisation of those
two state variables happens now on `componentDidUpdate` method and
ONLY if the prop `reactionCount` has been changed, that may be a bit
dangerous so, that `if` clause should NEVER be removed.

Furthermore by keeping the redux action's after animation has ended
we have insurance that when the component will update by `reactionCount`
we have already finished our animation, also the state for the specific
instance of Reaction won't update since the state variables are the same
(have been updated already on `handleClick`, and `handleAnimationEnded`).

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15853


nevyangelova (approved):
thanks @koox00 tested and LGTM! 🎉

calebroseland (approved):
@koox00 this is a non-blocker for me, and may not be worth fixing, but is there an easy way to prevent this quick-click glitch? (Where it goes to either 0 or 3, instead of being constrained to 1 or 2)

https://user-images.githubusercontent.com/11724372/105909614-61585980-5fed-11eb-9183-d693afb834ea.mp4

Also, it might be good to convert the index to TS as well, but I realize that TS conversion isn't the focus of this issue, so this too isn't a blocker.

Overall looks good to me!

koox00:
> @koox00 this is a non-blocker for me, and may not be worth fixing, but is there an easy way to prevent this quick-click glitch? (Where it goes to either 0 or 3, instead of being constrained to 1 or 2)
> CleanShot.2021-01-26.at.15.29.53.mp4

Thank you @calebroseland nIce catch, yes I think there is an easy fix, we just need to prevent multiple clicks while animating.

> Also, it might be good to convert the index files to TS as well, but I realize that TS conversion isn't the focus of this issue, so this too isn't a blocker.

I'll get on too that as well, thanks!

nevyangelova (approved):

calebroseland (approved):
Thanks @koox00! LGTM

jgilliam17 (approved):
Thank you @koox00 
Tested, looks good to merge.
- Verified emoji reaction animation and synchronization of reaction counts between RHS and the center channel.
@calebroseland Can you please help merge? Thanks 🙂

PR #7389, opened 2021-01-25
[MM-32087] - Add theme variable for sidebar team bar

ghost:
#### Summary
This PR adds a new theme variable for team sidebar text which would replace the currently constructed colour of the team bar, inclusive of channelHeaderColour + opacity.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32087

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1342


devinbinnie (commented):
Since this is a new theme variable and many current custom themes that users (like myself) use won't have this new variable. What's the plan for those themes wrt. migration to using this new variable?

Are we able to fall back on a different value if need be?

nevyangelova:
> Since this is a new theme variable and many current custom themes that users (like myself) use won't have this new variable. What's the plan for those themes wrt. migration to using this new variable?
> 
> Are we able to fall back on a different value if need be?

I was thinking a migration to use the new variable. Would it be too tedious?

devinbinnie:
> > Since this is a new theme variable and many current custom themes that users (like myself) use won't have this new variable. What's the plan for those themes wrt. migration to using this new variable?
> > Are we able to fall back on a different value if need be?
> 
> I was thinking a migration to use the new variable. Would it be too tedious?

So like a database migration script? That would be alright with me.
Another possible solution would be to null check and fall back on an existing variable.

components/widgets/team_icon/team_icon.scss
```diff
@@ -47,41 +47,19 @@
 
     &:hover {
         box-shadow: 0 0 0 3px rgba(255, 255, 255, 0.32);
-        border-color: var(--sidebar-header-bg);
+        border-color: var(--sidebar-teambar-color);
```
matthewbirtch, line 50: should we change to `--sidebar-teambar-bg` for clarity and consistency?

matthewbirtch:
@nevyangelova the new variable will need to be added in to the customer them UI in the Account Settings as well. Suggest we call it "Team Sidebar BG" in this context:

image

nevyangelova:
> > > Since this is a new theme variable and many current custom themes that users (like myself) use won't have this new variable. What's the plan for those themes wrt. migration to using this new variable?
> > > Are we able to fall back on a different value if need be?
> > 
> > 
> > I was thinking a migration to use the new variable. Would it be too tedious?
> 
> So like a database migration script? That would be alright with me.
> Another possible solution would be to null check and fall back on an existing variable.

Seems like a good discussion to have in triage today.

nevyangelova:
@matthewbirtch nice catch! Fixed.

matthewbirtch (approved):
Looks good to me!

jgilliam17:
Thanks @nevyangelova 
Seeing one small issue, when user selects a custom BG color for the team sidebar, it is not respected on drag. See video.

https://user-images.githubusercontent.com/52937121/106661399-29ae5c00-656f-11eb-8fb0-88da3993224c.mov

nevyangelova:
@jgilliam17 this is expected behaviour because the border colour on active is the same as the sidebar-text-active-border, so if you haven't changed that colour either with uploading a whole team or manually it will show the default, as you have the rest of the sidebar default too.

matthewbirtch:
> @jgilliam17 this is expected behaviour because the border colour on active is the same as the sidebar-text-active-border, so if you haven't changed that colour either with uploading a whole team or manually it will show the default, as you have the rest of the sidebar default too.

I think @jgilliam17 is talking about the internal border color though. That internal border color could be the color of the team sidebar background when dragging
image

nevyangelova:
Oh thank you for clarifying I will fix asap

jgilliam17 (approved):
Thanks @nevyangelova 
Tested, looks good to merge.
- Verified new “Team Sidebar BG” color setting in Account Settings>Display>Custom Theme 
- Verified team icon internal border color matches team sidebar background color when static and on drag.

devinbinnie (approved):

hmhealey (changes requested):
@nevyangelova I think we're getting a partial regression of https://mattermost.atlassian.net/browse/MM-23456 here. That's one of the cases that `applyThemeDefaults` was added for. 

When the user pastes in a custom theme that's missing fields, there's a JS error and some parts of the UI don't show the theme preview correctly any more. This might be a good case for Devin's suggestion about splitting out the logic for backfilling default values in cases when the page's theme isn't coming from the stored preference.

Here's a video of the test server vs community-daily

https://user-images.githubusercontent.com/3277310/107060323-d0c80900-67a4-11eb-9462-66c1dc17a76d.mov

hmhealey (approved):

PR #7400, opened 2021-01-26
Cypress/E2E: Separate TM4J keys for team and enterprise of site statistics

saturninoabril:
#### Summary
Separate TM4J keys for team and enterprise of site statistics
- remain MM-T904 for enterprise, and
- add MM-T3804 for team edition


furqanmlk (approved):
Good changes 👍 
All looks good

prapti (approved):
LGTM!
Should T904 also be moved in TM4J under the "Enterprise" folder?

saturninoabril:
@prapti I think we have many cases like this where some test require a license for some TM4J folder. The current convention we have is to remain as is in TM4J and then only organize in Cypress by having the same subfolder structure inside  `/enterprise` folder. Very much open to discussion/changes on how we'd like to organize moving forward and that might be good topic in QA meeting.

PR #7391, opened 2021-01-25
MM-30540 - display renewal panel in license screen

pablovelezvidal:
#### Summary
MM-30540 - Display renewal panel in License screen
System Console > License screen: show a message indicating that license is up for renewal with button to renew in CWS.

#### Ticket Link
[MM-30540](https://mattermost.atlassian.net/browse/MM-30540)

#### Related Pull Requests

[Depends on](https://github.com/mattermost/mattermost-webapp/pull/7355#issuecomment-767042747)

#### Screenshots
Captura de pantalla 2021-01-23 a las 14 00 34
Captura de pantalla 2021-01-23 a las 14 01 16

PR #7413, opened 2021-01-28
MM 31192  - license renewal telemetry

pablovelezvidal:
#### Summary
Adds the license renewal telemetry to the renewal banner and system console card

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31192

#### Related Pull Requests

https://github.com/mattermost/mattermost-webapp/pull/7391


marianunez (approved):
Looks great! Thanks @pablovelezvidal!

AGMETEOR (commented):
Hey @pablovelezvidal . Everythings looks great. I just have a small question. Why do we have to make ```telemetryInfo``` a prop? Since that `RenewLink` button is dedicated to renewing licenses, I was of the view that the events will always be `renew_license_admin_console_success` and `renew_license_admin_console_fail` and that we could just specify those as contants in the component itself?

AGMETEOR (approved):

pablovelezvidal:
> Hey @pablovelezvidal . Everything looks great. I just have a small question. Why do we have to make `telemetryInfo` a prop? Since that `RenewLink` button is dedicated to renewing licenses, I was of the view that the events will always be `renew_license_admin_console_success` and `renew_license_admin_console_fail` and that we could just specify those as contants in the component itself?

Thanks for the review @AGMETEOR . So, the user story define to send telemetry from two different places, one from the system console and the other from the banner which can be anywhere in the app. So, we have telemetry for `renew_license_admin_console_success` and for `renew_license_banner_success` and this also let the possibility to call the renewal link cmp from any other place and handle different telemetry.

AGMETEOR:
> > Hey @pablovelezvidal . Everything looks great. I just have a small question. Why do we have to make `telemetryInfo` a prop? Since that `RenewLink` button is dedicated to renewing licenses, I was of the view that the events will always be `renew_license_admin_console_success` and `renew_license_admin_console_fail` and that we could just specify those as contants in the component itself?
> 
> Thanks for the review @AGMETEOR . So, the user story define to send telemetry from two different places, one from the system console and the other from the banner which can be anywhere in the app. So, we have telemetry for `renew_license_admin_console_success` and for `renew_license_banner_success` and this also let the possibility to call the renewal link cmp from any other place and handle different telemetry.

Oh yeah. I have now noticed the difference. Thanks a lot, Pablo.

PR #7421, opened 2021-01-29
Cypress/E2E: Fix tests for bots in list and in sidebar

saturninoabril:
#### Summary
- `in_lists_spec.js`
  - use regular user as possible
  - retain MM-T1834 and MM-T1835 which is consistently passing (to easily move into production)
- `sidebar_display_spec.js`
  - extract MM-T1836 into this new file
  - fixed test by using regular using since the list/number of other users and bots are consistent

#### Ticket Link
none, failing daily Cypress run


furqanmlk (approved):
Looks good !
Test cases passed ✅ 
```cy.visitAndWait()```  needs to be replaced

image

prapti (approved):
Looked through it thoroughly and approving. Local server has some issue I need to fix.
Thank you Saturn!

PR #7423, opened 2021-01-29
Rework embedded forms

larkox:
#### Summary
Reworked embedded forms. Failures are due to redux versions.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32334

#### Related Pull Requests
- Redux: Soon
- Apps Proxy: https://github.com/mattermost/mattermost-plugin-apps/pull/56

#### Screenshots
None


components/post_view/embedded_forms/action_menu/action_menu.tsx
```diff
@@ -0,0 +1,104 @@
...
+        const binding = props.binding;
+        this.providers = [];
+        if (binding.bindings) {
+            const options = binding.bindings.map((b) => {
+                return {text: b.label, value: b.location};
+            });
+            this.providers = [new MenuActionProvider(options)];
+        }
+
+        this.state = {};
```
catalintomai, line 51: any default selected value?

utils/apps.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {AppBinding} from 'mattermost-redux/types/apps';
+
+export function fillBindingsInformation(binding: AppBinding) {
```
catalintomai, line 6: Is this something that we need to rely on the client to set/assert, or is this something that needs to generated/validate on the PP/app side? (like setting the parent binding appId on all the contained bindings, right location etc.)

components/post_view/post_body_additional_content/post_body_additional_content.tsx
```diff
@@ -4,16 +4,17 @@
 import {Post, PostEmbed} from 'mattermost-redux/types/posts';
 
 import {getEmbedFromMetadata} from 'mattermost-redux/utils/post_utils';
+import {PostEmbed as AppEmbed} from 'mattermost-redux/types/apps';
```
catalintomai, line 7: Is `PostEmbed` being used?

components/post_view/embedded_forms/embedded_form/embedded_form.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Options specific to text formatting
+     */
+    options?: Partial<TextFormattingOptions>;
+
+    currentRelativeTeamUrl: string;
+}
+
+export default class EmbeddedForm extends React.PureComponent<Props> {
```
catalintomai, line 45: One idea would be to use `EmbeddedFormComponent` on the client instead of `EmbeddedForm` on the client (or whatever PP struct name we come up with instead). and use `embeddedform(s)` instead of `embed(s)` as the supporting data model.

components/post_view/embedded_forms/action_menu/action_menu.tsx
larkox: Not for P1. Right now bindings don't have that functionality, and we must consider where to put it.

utils/apps.ts
larkox: I will let @levb to give take the final call here, but the idea is that all this is taken care by the client, so no redundant information is passed "through the wire" to the user agents.

components/post_view/post_body_additional_content/post_body_additional_content.tsx
larkox: In this component is used, yes. PostEmbed is a completely different structure. I would be more than willing to change the name for the new "PostEmbed", but I am lacking of imagination for the names :sweat_smile:

utils/apps.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        // Compose location
+        b.location = binding.location + '/' + b.location;
+
+        // Propagate call down if not defined
+        if (!b.call) {
+            b.call = binding.call;
+        }
+
+        fillBindingsInformation(b);
```
catalintomai, line 28: In what circumstances do we have more than two levels of bindings?

components/post_view/embedded_bindings/select_binding/select_binding.tsx
```diff
@@ -0,0 +1,104 @@
...
+        const binding = this.props.binding.bindings?.find((b) => b.location === selected.value);
+        if (!binding) {
+            // TODO Error?
+            return;
+        }
+
+        const {userId, post} = this.props;
+
+        const call: AppCall = {
+            url: binding.call?.url || '',
```
catalintomai, line 69: I wonder if it is worth adding an AppCall constructor function somewhere to be reused.

components/post_view/embedded_bindings/select_binding/select_binding.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    handleSelected = (selected: Option) => {
+        if (!selected) {
+            return;
+        }
+
+        this.setState({selected});
+        const binding = this.props.binding.bindings?.find((b) => b.location === selected.value);
+        if (!binding) {
+            // TODO Error?
```
catalintomai, line 62: at the minimum log to the console as opposed to swallowing it?

utils/apps.ts
```diff
@@ -0,0 +1,50 @@
...
+import {AppBinding} from 'mattermost-redux/types/apps';
+import {GlobalState} from 'mattermost-redux/types/store';
+
+export function appsEnabled(state: GlobalState) {// eslint-disable-line @typescript-eslint/no-unused-vars
+    // TODO uncomment when featur flag is in place
+    //return getConfig(state)?.['FeatureFlagApps' as keyof Partial<ClientConfig>];
+    return true;
+}
+
+export function fillBindingsInformation(binding: AppBinding) {
```
catalintomai, line 13: should we validate the `binding` param?

catalintomai (approved):
A few comments, LGTM. Also, any CR-related changes/improvs prob. to be ported to mobile PR as well for completeness.

utils/apps.ts
larkox: With commands we have an undefined amount of level of bindings.
In message attachments, at the moment we have 3 levels: Attachment -> Menu Binding -> Option.
Potentially (depending on UX) we may have an undefined amount of levels too: Attachment -> Other -> Menu item -> Submenu item -> Subsubmenu item...
Similar situation with channel header and post menu items:
ChannelHeader -> Element -> SubMenu -> Subsubmenu...

components/post_view/embedded_bindings/select_binding/select_binding.tsx
larkox: Added a call constructor for the calls and used here. If we move forward with this, I will create a ticket to use the constructor all around the code.

larkox: And on mobile.

utils/apps.ts
larkox: That should never happen, but better safe than sorry.

catalintomai (approved):

PR #7426, opened 2021-01-30
Refactored apps command parser

levb:
Still todo:
- simplify `incomplete`/`incompleteStart` manipulations
- add more tests
  - `BOOL`

@mickmister @larkox I would really appreciate TS/JS style critique, make it more idiomatic.

This 1-char at a time, C-style parsing may not be the fastest in interpreted JS, but I figured it simplifies the logic. The performance of parsing a single command in the user-agent is not really an issue.


components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -84,31 +209,414 @@ export class AppCommandParser {
...
+                    // return the state from the last matching binding
+                    done = true;
+                    break;
+                }
+                parsed.binding = binding;
+                parsed.location += '/' + binding.label;
+                if (binding.bindings) {
+                    bindings = binding.bindings;
+                } else {
+                    bindings = [];
```
larkox, line 285: If we get here, we get the next token before deducing that these are all the bindings we can fetch. Should we take a shortcut here (in other words, end here the loop)?

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -205,19 +713,19 @@ export class AppCommandParser {
...
             }
             callResponse = res.data;
         } catch (e) {
             this.displayError(e);
-            return null;
+            return undefined;
         }
 
-        if (callResponse && callResponse.form) {
+        if (callResponse?.form) {
```
larkox, line 724: We can return directly here `callResponse?.form`, since in case it is not defined, it will return undefined.

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -43,38 +41,165 @@ export type Store = {
...
+
+    // composeCallFromParsed creates the form submission call
+    composeCallFromParsed = (parsed: ParsedCommand): AppCall | null => {
+        if (!parsed.binding) {
             return null;
         }
-        const call = form.call || binding.call;
+
+        let call = parsed.binding.call;
+        if (parsed.form && parsed.form.call) {
```
larkox, line 200: Can we use here `parsed.form?.call`?

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -231,229 +739,155 @@ export class AppCommandParser {
...
-
-        const bindings = this.getCommandBindings();
-        for (const binding of bindings) {
-            let base = binding.app_id;
-            if (!base) {
-                continue;
-            }
+    // getSuggestionsForSubCommands returns suggestions for a subcommand's name
+    getCommandSuggestions = (parsed: ParsedCommand): AutocompleteSuggestion[] => {
+        if (!parsed.binding || !parsed.binding.bindings || !parsed.binding.bindings.length) {
```
larkox, line 744: Can this simplified to `if (!parsed.binding?.bindings?.length)`?

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -231,229 +739,155 @@ export class AppCommandParser {
...
+    getParameterSuggestions = async (parsed: ParsedCommand): Promise<AutocompleteSuggestion[]> => {
+        switch (Number(parsed.state)) {
+        case ParseState.StartParameter: {
+            // see if there's a matching positional field
+            const positional = parsed.form?.fields?.find((f: AppField) => f.position === parsed.position + 1);
+            if (positional) {
+                parsed.field = positional;
+                return this.getValueSuggestions(parsed);
             }
+            return this.getFlagNameSuggestions(parsed);
```
larkox, line 774: Are positional and flag values exclusives? I mean, if I can I do something like: `/ci start-pipeline mypipeline --force=true` and `/ci start-pipeline --force=true mypipeline`, when I finish writing `/ci start-pipeline ` I should receive the suggestion for the positional argument "pipeline name" and the flag argument "force", right?

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -231,229 +739,155 @@ export class AppCommandParser {
...
+                    complete: prefix + (f.label || f.name) + suffix,
+                    suggestion: '--' + (f.label || f.name),
+                    description: f.description,
+                    hint: f.hint,
+                };
             });
-            return this.getSuggestionsFromFlags(available);
         }
 
         return [{suggestion: 'Could not find any suggestions'}];
```
larkox, line 856: Wonder if there is any case where this suggestion would appear along a "execute" suggestion or other kind of suggestions.

larkox (changes requested):
Minor styling issues, and some question here and there.

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -84,31 +209,414 @@ export class AppCommandParser {
...
+                    if (autocompleteMode) {
+                        // Finish in the Command state, 'incomplete' will have the query string
+                        done = true;
+                    } else {
+                        parsed.state = ParseState.EndCommand;
+                    }
+                    break;
+                }
+                case ' ':
+                case '\t': {
```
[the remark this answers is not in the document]
levb: I've assumed (wrongly?) that the new lines are not passed in. Easy to add (as a separate ticket?)

components/suggestion/command_provider/app_command_parser.ts
levb: this whole `if` is fishy, need to revisit, thanks for pointing it out

components/suggestion/command_provider/app_command_parser.ts
levb: they are supposed to be mixable in this parser, will add a couple more tests for it. It does create interesting problems, esp. with boolean flags (optional value).

components/suggestion/command_provider/app_command_parser.ts
levb: I'd like to do another round and simplify, but this should be functional, the iteration should stop at the next subcommand anyway.

larkox (approved):

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -43,72 +41,582 @@ export type Store = {
...
+            if (i < command.length) {
+                c = command[i];
+            }
+
+            switch (Number(parsed.state)) {
+            case ParseState.Start: {
+                if (c !== '/') {
+                    return this.parseError('command must start with a /: ' + i);
+                }
+                if (bindings.length === 0) {
```
catalintomai, line 234: should this check be placed here or at the top?

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -43,72 +41,582 @@ export type Store = {
...
+                    parsed.incompleteStart = i;
+                    parsed.state = ParseState.Command;
+                    break;
+                }
+                }
+                break;
+            }
+
+            default: {
+                return this.parseError('unreachable: unexpected state in matchBinding: ' + parsed.state);
```
catalintomai, line 312: might help to include the command string and the current index)

components/suggestion/command_provider/app_command_parser.test.ts
```diff
@@ -228,400 +240,569 @@ describe('AppCommandParser', () => {
...
+                    expect(parsed.incomplete).toBe('something');
+                    expect(parsed.incompleteStart).toBe(20);
+                }},
+            },
+        ];
+
+        table.forEach((tc) => {
+            test(tc.title, async () => {
+                const a = await parser.matchBinding(tc.command, definitions, true);
+                checkResult(a, tc.autocomplete || tc.submit);
```
catalintomai, line 398: why `|| tc.submit` ?

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -43,72 +41,582 @@ export type Store = {
...
+                case ' ':
+                case '\t':
+                case '=': {
+                    const field = fields.find((f) => f.label === parsed.incomplete);
+                    if (!field) {
+                        return this.parseError('command does not accept flag ' + parsed.incomplete + ': ' + i);
+                    }
+                    parsed.state = ParseState.FlagValueSeparator;
+                    parsed.field = field;
+                    parsed.incomplete = '';
```
catalintomai, line 431: maybe set `parsed.incomplete = i` and remove line 444.

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -43,72 +41,582 @@ export type Store = {
...
+                break;
+            }
+
+            case ParseState.CommandSeparator: {
+                switch (c) {
+                case '': {
+                    // Reached the end of input.
+                    parsed.incomplete = '';
+                    parsed.incompleteStart = i;
+                    done = true;
```
catalintomai, line 293: Could we end with a `parsed.state = ParseState.Command;` here (since we are `done` anyway) ? and remove the check for `CommandSeparator` outside this method?

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -43,72 +41,582 @@ export type Store = {
...
+                    break;
+                }
+                case '`': {
+                    parsed.state = ParseState.TickValue;
+                    i++;
+                    break;
+                }
+                case ' ':
+                case '\t':
+                    return this.parseError('unreachable: unexpected whitespace: ' + i);
```
catalintomai, line 489: not sure if it is a goal for test cases to exercise all `parseErrors` or not.

catalintomai (approved):
Some small comments, LGTM

components/suggestion/command_provider/app_command_parser.ts
levb: It was not the goal yet, but the unreachables are not testable anyway. They are just there as precautions.

levb:
@catalintomai I moved `matchBinding`, `parseForm`, and `parseError=>asError` to improve the code structure and to handle errors like you suggested. 

More consistent handling of the `incomplete` buffer and state is a "nice to have", but perhaps can be done later, as needed? Ditto more tests as we go.

components/suggestion/command_provider/app_command_parser.ts
```diff
@@ -43,72 +41,579 @@ export type Store = {
...
+                case '\t':
+                case '=': {
+                    const field = fields.find((f) => f.label === this.incomplete);
+                    if (!field) {
+                        return this.asError('command does not accept flag ' + this.incomplete);
+                    }
+                    this.state = ParseState.FlagValueSeparator;
+                    this.field = field;
+                    this.incomplete = '';
+                    break;
```
catalintomai, line 298: do we need to set the `this.incompleteStart` as well?

catalintomai (approved):

PR #7427, opened 2021-01-30
[MM-18952/GH-16740] - Migrate 'components/authorize' module and associated tests to TypeScript

lucievr:
#### Summary

- Migrated all files inside `components/authorize` to TypeScript
- Renamed files to have TypeScript extensions and updated incorrect imports
- `Check-types` resulted in no errors inside this component (there were other type errors elsewhere in the repo which you might already be aware of)

#### Ticket Link

Fixes [#16740](https://github.com/mattermost/mattermost-server/issues/16740)
https://mattermost.atlassian.net/browse/MM-18952


components/authorize/authorize.tsx
```diff
@@ -1,41 +1,56 @@
...
-        return {
-            location: PropTypes.object.isRequired,
-            actions: PropTypes.shape({
-                getOAuthAppInfo: PropTypes.func.isRequired,
-                allowOAuth2: PropTypes.func.isRequired,
-            }).isRequired,
-        };
-    }
+export type Params = {
+    responseType: string | null;
```
devinbinnie, line 14: Can these just be optional parameters instead of specifically expecting `null`?
```suggestion
    responseType?: string;
```

components/authorize/authorize.tsx
```diff
@@ -1,41 +1,56 @@
...
+        search: string;
+    };
+    actions: {
+        getOAuthAppInfo: (clientId: string | null) => Promise<{data: OAuthApp; error?: Error}>;
+        allowOAuth2: (params: Params) => Promise<{data?: any; error?: Error}>;
+    };
+}
+
+type State = {
+    app: OAuthApp | null;
```
devinbinnie, line 32: Same here with respect to optional parameters.

components/authorize/authorize.tsx
```diff
@@ -1,41 +1,56 @@
...
+    error: string | null;
+}
+
+export default class Authorize extends React.PureComponent<Props, State> {
+    public constructor(props: Props) {
         super(props);
 
-        this.state = {};
+        this.state = {
+            app: null,
```
devinbinnie, line 41: If we do the above, we shouldn't need to explicitly set `null` here.

devinbinnie (changes requested):
Thanks for the PR @lucievr!
Looks good overall, I just had a few things that I think we should resolve.

components/authorize/authorize.tsx
lucievr: I tried that initially but the issue here is that the `searchParams.get()` function on lines 68 - 72 returns `string | null`. This then throws a type error on line 75 as it's not the same type as `string | undefined`

params

devinbinnie: Ahh, that's a problem. Okay we can leave this one alone.

devinbinnie (approved):
LGTM! Thanks @lucievr!

nevyangelova (approved):
Thanks @lucievr LGTM 🎉

furqanmlk (approved):
All related test cases passed ✅ 

image

image


image

image

image


image

image

PR #7441, opened 2021-02-01
Custom status feature

chetanyakan:
#### Summary
- Create custom status components.
    - Custom status option in status modal
    - Custom status modal

#### Ticket Link
- Specs: https://mattermost.atlassian.net/wiki/spaces/CS/pages/983302547/User+Experience+and+Functional+Notes

#### Related Pull Requests
- https://github.com/mattermost/mattermost-redux/pull/1349
- https://github.com/mattermost/mattermost-server/pull/16835

#### Screenshots

image
image

image
image
image

image
image


esethna:
@matthewbirtch Here's the test server from brightscout that includes webapp, redux and server changes so we can review the feature in it's entirety: http://ec2-3-137-148-232.us-east-2.compute.amazonaws.com:8065/signup_user_complete/?id=559fit8sj7bm8n7mcpqcz9cy5y

esethna (changes requested):
Thanks @chetanyakan! A few minor things from my end, but overall this looks amazing:

1. Can we please move the "Enable Custom Statuses" admin setting to the bottom of the Users and Teams page?
image

2. Slash commands for custom statuses should not be functional if the config setting is disabled. Similarly it should not show up in the autocomplete
image

3. Please work with @matthewbirtch (and @asaadmahmood may be able to help as well) to ensure UI exactly matches designs, looks like it needs some minor pixel pushing and hover states etc on some buttons

4. If recent statuses have text that matches the options in the suggestion list, it should not appear in both places. We should remove it from the suggestion list while it is in the recent list

image

5. @matthewbirtch do we want to consider compressing the items in this list (perhaps to what we planned for channel sidebar? ie 32 px). It feels really spacey right now 0/5

image

6. I think there may be some bugginess with the detection for last post made by the user. What's expected as the user makes new posts?

image

esethna:
Adding @deanwhillier as he was interested in reviewing this PR

matthewbirtch (commented):
This is great @chetanyakan. I have a number of tweaks to make to polish things up a bit. I hope the direction below is clear enough, but let me know if you have any questions.

### Status Menu
- Change the color of the emoji icon to `fill: var(--center-channel-color-56);` And on hover, the icon color changes to `fill: var(--center-channel-color-72);`
- For the active state of the status menu, we don't need to change the color of the text, just the background. Right now it looks like below.

image


### Status Modal
- Can we center the modal vertically in the viewport? Similar to what we're doing with the Channel Switcher
- Padding on the modal need to be updated seens off. I think if you set `modal-content` to have a padding of 24px and remove padding from the inner containers, it should end up better,
- Remove the inner shadow on the Set Status input field
- This input needs to span 100% of the width of its container as well. There appears to be a width `width: calc(100% - 15px);` set on this and I'm not sure what for.
- The position of the clear `x` icon on the Set Status input field seems slightly off. 
  - The clear icon might also be a bit smaller than designed
  - And the clear icon should be using `fill: var(--center-channel-color-56);`
image
- When an emoji is selected, the emoji displayed within the input feels too small. Can we change this to 20x20 instead of 16x16? This would also match with the suggested options below.
- Can we tweak the suggestions below so that the icon horizontally aligns with the icon in the input field and so the text label also aligns with the text in the input field above?
  - I think we might need to increase `margin-right` on the suggestion icon to 16px
  - It seems like the emoji may not be vertically aligned with the text either. It might need to bump up 1px, right now it seems to be sittle a touch too low.
image
- For the hover state on the suggested options, can we add `border-radius: 4px;`
- Also, for each suggestion row, can we make the height 40px instead of 44px? Just to tighten things up a bit.
- Also, can we make sure the row stays within the padding on the modal container as noted above? Right now, the hover state extends right to the edge of the Modal
image
- The clear icon on a 'recent' status item should be using `fill: var(--center-channel-color-56);`
  - it also looks like it might need to bump down 1px to center properly in the row
  - image
- It looks like there is extra spacing between the footer and the suggestions list....maybe because the height of the suggestion list is set to 280px. Maybe we can set the height of this list so that it is the height of the title and 5 suggestions and scrolls beyond that. If we change the height of suggestion items to 40px as noted above, that would make this height 232px.
- Related to the note above, if you set the padding on the `modal-footer` container to `padding: 0;` and let the `modal-content` container have all the padding it should clean up the spacing

### Status on Channel Sidebar Header
- The emoji next to the username looks like it might need to move up 1px and can it move to the left by 2px? It's a bit tight to the main menu button
- I wonder if we should make clicking the emoji here launch the status modal. Right now this opens the main menu which doesn't feel right.
- The tooltip has extra space under the status label that isn't needed
- Also, the emoji in the tooltip looks misaligned
image

### Status showing in messages
- Tooltips here also seem misaligned and extra space is showing on the bottom of the tooltip
image
- The emoji is misaligned with the text in the tooltip as well

### Update your status button in posts
- Can we change the icon size in this button to 12x12?
- Can we add 2px more of padding on the left and right of the button?

### Status in the Channel Sidebar DM
- Looks vertically misaligned. Can we bump it up here by 1px?
image
- Also the icon is supposed to display inline with the name. See the comparison below for what I mean

image
- The tooltip here has the same issues as mentioned above

chetanyakan:
> 4\. If recent statuses have text that matches the options in the suggestion list, it should not appear in both places. We should remove it from the suggestion list while it is in the recent list

@esethna Based on how the 'Recent Statuses' feature is designed and implemented, we don't show the same recent status if the status-text matches but the emoji is different.

E.g. `😪 Working from home` will replace `🏠 Working from home`

In this case, should we show `🏠 Working from home` in the Suggestions again, along with `😪 Working from home` in the recents?

***

@matthewbirtch What should be the length limit for the custom status text? 
Currently, I have set it to 250 characters to match the max length of `CHANNEL_PURPOSE`.
Let me know if that is okay.

deanwhillier (changes requested):
Thanks @chetanyakan! 🎉 Had a look at the channel header drop down so far and thought I'd get some comments in for you to consider while we continue the review.

components/status_dropdown/status_dropdown.jsx
```diff
@@ -105,6 +149,112 @@ export default class StatusDropdown extends React.PureComponent {
...
+                                />
+                            </Tooltip>
+                        }
+                    >
+                        <span
+                            className='input-clear-x'
+                            onClick={this.handleClearStatus}
+                        >
+                            <i className='icon icon-close-circle'/>
+                        </span>
```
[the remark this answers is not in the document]
chetanyakan: @deanwhillier If we make this `close` icon a sibling of the previous set status, the hover state does not really work.

image

Do you have any suggestions?

esethna:
> In this case, should we show `🏠 Working from home` in the Suggestions again, along with `😪 Working from home` in the recents?

@chetanyakan I would propose we follow the same rules. If the text matches then it should not appear twice (so it would only appear in the recent list)

components/status_dropdown/status_dropdown.jsx
```diff
@@ -78,6 +103,25 @@ export default class StatusDropdown extends React.PureComponent {
...
+    }
+
+    showCustomStatusTextTooltip = () => {
+        const element = this.customStatusTextRef;
+        if (element && element.offsetWidth < element.scrollWidth) {
+            this.setState({showCustomStatusTooltip: true});
+        } else {
+            this.setState({showCustomStatusTooltip: false});
+        }
+    }
```
michelengelen, line 123: this can be simplified since the condition in the if-clause basically resembles the status that is set in the `setState`, so it could look like this:

```suggestion
    showCustomStatusTextTooltip = () => {
        const element = this.customStatusTextRef;
        this.setState({showCustomStatusTooltip: element && element.offsetWidth < element.scrollWidth});
    }
```

components/status_dropdown/status_dropdown.jsx
```diff
@@ -105,6 +149,112 @@ export default class StatusDropdown extends React.PureComponent {
...
         );
     }
+    handleClearStatus = (e) => {
+        e.stopPropagation();
+        this.props.actions.unsetCustomStatus();
+    };
+
+    onToggle = (open) => {
+        this.props.actions.setStatusDropdown(open);
+    }
```
michelengelen, line 159: can be simplified with immediate return: 

```suggestion
    onToggle = (open) => this.props.actions.setStatusDropdown(open);
```

components/status_dropdown/status_dropdown.jsx
```diff
@@ -105,6 +149,112 @@ export default class StatusDropdown extends React.PureComponent {
...
+
+    renderCustomStatus = () => {
+        if (!this.props.isCustomStatusEnabled) {
+            return null;
+        }
+        let customStatusText = localizeMessage('status_dropdown.set_custom', 'Set a Custom Status');
+        let customStatusEmoji = <EmojiIcon className={'custom-status-emoji'}/>;
```
michelengelen, line 166: this could be written as a `const` with a ternary conditional assignment (replacing the `if`-clause on lines 170-178)

components/status_dropdown/status_dropdown.jsx
```diff
@@ -105,6 +149,112 @@ export default class StatusDropdown extends React.PureComponent {
...
+                                message={customStatusText}
+                                enableFormatting={true}
+                            />
+                        </Tooltip>
+                    }
+                >
+                    {customStatusTextComponent}
+                </OverlayTrigger>
+            );
+        }
```
michelengelen: this whole block seems very redundant over the different files. 

I would suggest extracting it into a helper function or component that can be imported instead.

components/status_dropdown/status_dropdown.scss
```diff
@@ -0,0 +1,98 @@
...
+.custom-status-suggestion-emoji > span:first-child {
+    background-size: 20px;
+}
+
+.custom_status__clear {
+    display: flex;
+    align-items: center;
+    margin-left: auto;
+    color: rgba(var(--center-channel-color-rgb), 0.52);
+}
```
michelengelen: could you please simplify with SCSS `Current Selector` please?

components/status_dropdown/status_dropdown.scss
```diff
@@ -0,0 +1,98 @@
...
+    top: 0;
+}
+
+.pulsating_dot:after {
+    animation: pulse1 2s ease 0s infinite;
+}
+
+.pulsating_dot:before {
+    animation: pulse2 2s ease 0s infinite;
+}
```
michelengelen, line 74: could you please simplify the `.pulsating_dot` class with SCSS `Current Selector` as well?

components/suggestion/at_mention_provider/at_mention_suggestion.jsx
```diff
@@ -124,6 +126,17 @@ export default class AtMentionSuggestion extends Suggestion {
...
             );
+
+            customStatus = (
+                <CustomStatusEmoji
+                    showTooltip={true}
+                    userID={item.id}
+                    emojiSize={15}
+                    emojiStyle={{
+                        marginLeft: 4,
+                    }}
```
michelengelen, line 137: since most of the time this is set to `marginLeft: 4` I would suggest adding this setting to the `defaultProps` in `components/custom_status/custom_status_emoji.tsx`

sass/layout/_headers.scss
```diff
@@ -729,8 +742,30 @@
...
+            display: flex;
+            height: 16px;
+            min-width: 16px;
+        }
+        .custom-status-emoji > span:first-child {
+            height: 16px;
+            width: 16px;
+            min-height: 16px;
+            min-width: 16px;
+        }
```
michelengelen, line 773: could you please simplify with SCSS `Current Selector &`?

michelengelen (changes requested):
there are generally some code styling comments and @deanwhillier already mentioned the HTML accessibility issues with `button`s und invalid elements (like `div`s within `ul`s).

As mentioned in one of the files I would suggest to ectract the whole render-customStatus logic to a separate file (maybe even a separate component), since it is kind of bloating the other files quite a bit.

michelengelen:
@chetanyakan could you please check the instance you are running on AWS? it is currently not reachable. 
I wanted to check the actual implementation today.

chetanyakan:
> @chetanyakan could you please check the instance you are running on AWS? it is currently not reachable.
> I wanted to check the actual implementation today.

@michelengelen The ec2 URL got updated.

You can use the following link. Your existing credentials should work.
http://ec2-13-58-138-215.us-east-2.compute.amazonaws.com:8065/test/channels/town-square

If you need an invite, you can use this link:
http://ec2-13-58-138-215.us-east-2.compute.amazonaws.com:8065/signup_user_complete/?id=559fit8sj7bm8n7mcpqcz9cy5y

michelengelen:
Thanks, that worked! ✨

matthewbirtch:
> @matthewbirtch What should be the length limit for the custom status text?
> Currently, I have set it to 250 characters to match the max length of `CHANNEL_PURPOSE`.
> Let me know if that is okay.

250 characters would be too long IMO. I would say 100 characters should be a more reasonable limit

components/status_dropdown/status_dropdown.jsx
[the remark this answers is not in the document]
deanwhillier: Hi @chetanyakan,

Try something like this: https://jsfiddle.net/deanwhillier/tsLywxnb/. The hover effect for the full menu item is applied to the parent of the two buttons (the menu item container itself) and adjust the styling of the first button while the second button has it's own hover styling.

As an alternative to putting the buttons side-by-side using flexbox, you could set the second button to position absolute, add some padding to the right of the first button and position the second button relative to the top right of the parent menu item container.

Note: the JS fiddle is just for demonstrations purposes. :P

components/post_view/post_header/post_header.tsx
```diff
@@ -6,12 +6,17 @@ import {FormattedMessage} from 'react-intl';
...
+
 import Constants from 'utils/constants';
 import * as PostUtils from 'utils/post_utils.jsx';
 import PostInfo from 'components/post_view/post_info';
 import UserProfile from 'components/user_profile';
 import BotBadge from 'components/widgets/badges/bot_badge';
 import Badge from 'components/widgets/badges/badge';
+import CustomStatusEmoji from 'components/custom_status/custom_status_emoji';
+import EmojiIcon from 'components/widgets/icons/emoji_icon';
+import './post_header.scss';
```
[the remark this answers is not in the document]
chetanyakan: Done.

components/status_dropdown/status_dropdown.jsx
```diff
@@ -5,19 +5,28 @@ import PropTypes from 'prop-types';
...
+import CustomStatusEmoji from 'components/custom_status/custom_status_emoji';
 import Menu from 'components/widgets/menu/menu';
 import MenuWrapper from 'components/widgets/menu/menu_wrapper';
 import StatusAwayIcon from 'components/widgets/icons/status_away_icon';
 import StatusOnlineIcon from 'components/widgets/icons/status_online_icon';
 import StatusDndIcon from 'components/widgets/icons/status_dnd_icon';
 import StatusOfflineIcon from 'components/widgets/icons/status_offline_icon';
 
+import OverlayTrigger from 'components/overlay_trigger';
+import './status_dropdown.scss';
```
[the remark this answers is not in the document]
chetanyakan: Done.

esethna:
@deanwhillier @michelengelen looks like feedback has been addressed, please take a look when you can! Thanks!

components/status_dropdown/status_dropdown.jsx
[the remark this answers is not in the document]
deanwhillier: I apologize @chetanyakan, I think I might have not clearly communicated my thoughts here so I'll provide specific examples.

Existing generated HTML:
```html
<ul class="Menu__content dropdown-menu">
    ...
    <li class="MenuItem" role="menuitem" id="status-menu-custom-status">
        <button class="style--none custom_status__row" data-toggle="modal toggle" aria-label=" dialog">
            <span class="custom_status__icon"><span class="d-flex"><span class="emoticon" alt=":palm_tree:" data-emoticon="palm_tree" style="background-image: url(&quot;http://ec2-13-58-138-215.us-east-2.compute.amazonaws.com:8065/static/emoji/1f334.png&quot;); background-size: 16px; height: 16px; width: 16px; min-height: 16px; min-width: 16px; margin-left: 0px;"></span></span></span>
            <span class="overflow--ellipsis text-nowrap custom_status__text">
                <p>On a vacation</p>
            </span>
        </button>
    </li>
    <li class="MenuItem" role="menuitem" id="status-menu-custom-status-clear">
        <span class="custom_status__clear">
            <button class="style--none input-clear-x">
                <i class="icon icon-close-circle"></i>
            </button>
        </span>
    </li>
    ...
</ul>
```

Suggested HTML:
```html
<ul class="Menu__content dropdown-menu">
    ...
    <li class="MenuItem" role="menuitem" id="status-menu-custom-status">
        <button class="style--none custom_status__row" data-toggle="modal toggle" aria-label=" dialog">
            <span class="custom_status__icon"><span class="d-flex"><span class="emoticon" alt=":palm_tree:" data-emoticon="palm_tree" style="background-image: url(&quot;http://ec2-13-58-138-215.us-east-2.compute.amazonaws.com:8065/static/emoji/1f334.png&quot;); background-size: 16px; height: 16px; width: 16px; min-height: 16px; min-width: 16px; margin-left: 0px;"></span></span></span>
            <span class="overflow--ellipsis text-nowrap custom_status__text">
                <p>On a vacation</p>
            </span>
        </button>
        <button class="style--none input-clear-x">
            <i class="icon icon-close-circle"></i>
        </button>
    </li>
    ...
</ul>
```

Both the button for the menu item and the clear button should be children of the same `<li>` menu item container regardless of whether they are positioned next to each other using `flexbox` or the clear button is absolutely positioned on top.

components/custom_status/custom_status_text.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+const CustomStatusText = (props: ComponentProps) => {
+    const {tooltipDirection, text, className} = props;
+    const customStatusEnabled = useSelector((state: GlobalState) => {
+        return isCustomStatusEnabled(state);
+    });
+    const [show, setShow] = useState<boolean>(false);
+    let spanElement: HTMLSpanElement | null = null;
+    if (!(customStatusEnabled)) {
```
michelengelen, line 26: The brackets around `customStatusEnabled` are obsolete

michelengelen (changes requested):
I do have only 1 remark this time

michelengelen (approved):
LGTM!

esethna (approved):
Looks like there's still a few minor UI polish things to work on with @matthewbirtch, but functionally it looks great. Approving from the PM perspective. Nice work @chetanyakan!   

One functional thing I'm 4/5 on doing here is allowing the user to open the custom status modal by clicking the emoji in the header area (I caught myself doing this a few times with unexpected results - ie opening the menu menu instead)
image

matthewbirtch (approved):

matthewbirtch:
Looks like the majority of UI tweaks have been made, so I've approved. However, I just noticed this one though:
image

Could we align all the status icons with the first`Set a Custom Status` item?

esethna (changes requested):
@chetanyakan if you could address the two items in the last comments from Matt and myself then I think we're good from the UX and PM perspective

matthewbirtch:
> One functional thing I'm 4/5 on doing here is allowing the user to open the custom status modal by clicking the emoji in the header area (I caught myself doing this a few times with unexpected results - ie opening the menu menu instead)

100% agree with this @esethna. I fully expected this as well

components/status_dropdown/status_dropdown.jsx
```diff
@@ -105,6 +120,87 @@ export default class StatusDropdown extends React.PureComponent {
...
+                </span>
+            ) : (
+                <EmojiIcon className={'custom-status-emoji'}/>
+            );
+
+        const clearButton = isStatusSet &&
+            (
+                <span
+                    className='custom_status__clear'
+                >
```
deanwhillier, line 154: I feel like this span isn't needed and would prefer not to have excess markup if not needed. Can we move the class to the child button and remove the span entirely?

deanwhillier (approved):
One additional comment below, but approving to keep things moving.

Also, I'd still really like to see the visual hover effect for the menu item applied to the li instead of the button  element so that the effect remains in place when you hover over the clear status button. Not a functional blocker though.

components/status_dropdown/status_dropdown.jsx
```diff
@@ -105,6 +120,87 @@ export default class StatusDropdown extends React.PureComponent {
...
+            );
+
+        const pulsatingDot = !isStatusSet && this.props.showCustomStatusPulsatingDot && (
+            <span className='pulsating_dot'/>
+        );
+
+        return (
+            <Menu.Group>
+                <Menu.ItemToggleModalRedux
+                    text={''}
```
hmhealey, line 184: I see why we're passing in no text here and then adding the icon and text as children, but this is going to prevent the underlying element from having its `accessibilityLabel` set. That said, I'm honestly not sure if that's a problem since I would've assumed any sort of accessibility software would just read the contents of the menu item. 

I'll see if I can find any more information about this, but there's probably nothing to change here since it might just be a side effect of whoever added that not having a full understanding of how aria labels should be used. I'll let you know if something does actually need to change here.

components/status_dropdown/status_dropdown.jsx
```diff
@@ -105,6 +120,87 @@ export default class StatusDropdown extends React.PureComponent {
...
+                    modalId={ModalIdentifiers.CUSTOM_STATUS}
+                    dialogType={CustomStatusModal}
+                    dialogProps={{userId: this.props.userId}}
+                    className='custom_status__row'
+                    id={'status-menu-custom-status'}
+                    sibling={clearButton}
```
hmhealey, line 190: Is the sibling something we actually need to pass into the menu item as a prop or could we just pass it in as an actual sibling in a fragment here? This seems like an odd pattern to me, although I wouldn't be surprised if it's just another issue with our menu components.

i18n/en.json
```diff
@@ -4044,7 +4048,9 @@
   "upload_overlay.info": "Drop a file to upload it.",
   "user_list.notFound": "No users found",
   "user_profile.account.editSettings": "Edit Account Settings",
-  "user_profile.account.localTime": "Local Time: ",
+  "user_profile.account.localTime": "Local Time",
+  "user_profile.custom-status": "Status",
+  "user_profile.custom-status.set-status": "Set a status",
```
hmhealey, line 4053: Minor nitpicking, but this should either be `user_profile.customStatus` or `user_profile.custom_status` without the hyphens. Not that we're very consistent with these IDs, but I think we don't use hyphens for them at least.

package-lock.json
```diff
@@ -7165,7 +7165,7 @@
         },
         "p-cancelable": {
           "version": "0.4.1",
-          "resolved": "https://registry.npmjs.org/p-cancelable/-/p-cancelable-0.4.1.tgz",
+          "resolved": "http://registry.npmjs.org/p-cancelable/-/p-cancelable-0.4.1.tgz",
```
hmhealey, line 7168: Could you revert these changes since there's no accompanying `package.json` changes?

sass/components/_emoticons.scss
```diff
@@ -79,7 +79,7 @@
     margin-right: 3px;
     position: absolute;
     width: 315px;
-    z-index: 999;
+    z-index: 1100;
```
hmhealey, line 82: Is this having to change to get the picker to appear on top of the status dropdown?

sass/layout/_sidebar-left.scss
```diff
@@ -1075,7 +1075,6 @@
     }
 
     .SidebarChannel span.SidebarChannelLinkLabel {
-        width: 100%;
```
hmhealey, line 1078: Why remove this?

selectors/views/custom_status.ts
```diff
@@ -0,0 +1,42 @@
...
+import {getCurrentUser, getUser} from 'mattermost-redux/selectors/entities/users';
+import {getConfig} from 'mattermost-redux/selectors/entities/general';
+
+import {getMyPreferences} from 'mattermost-redux/selectors/entities/preferences';
+import {getPreferenceKey} from 'mattermost-redux/utils/preference_utils';
+import {Preferences} from 'mattermost-redux/constants';
+
+import {GlobalState} from 'types/store';
+
+export function getCustomStatus(state: GlobalState, userID?: string) {
```
hmhealey, line 12: This selector needs to be made into a selector factory (like `makeGetCustomStatus`) since it takes an argument and returns new objects each time its called. Anywhere that we're passing its result into a component will cause that component to re-render a lot if we don't have some sort of memoization here, and since we'll be calling this with different user IDs, we'll need to memoize it on a per-component basis.

selectors/views/custom_status.ts
```diff
@@ -0,0 +1,42 @@
...
+    const user = userID ? getUser(state, userID) : getCurrentUser(state);
+    const userProps = user.props || {};
+    return userProps.customStatus ? JSON.parse(userProps.customStatus) : {};
```
hmhealey, line 15: Also, returning an empty object literal can cause similar issues where a new object is returned each time its called for a user that doesn't have a custom status set. You used `defaultProps` in a lot of places to set a default object for custom statuses, so perhaps we could just return undefined here?

If you get an ESList error from doing that, feel free to add an eslint-ignore-line for that. The `no-undefined` rule, if it's enabled, is something we've wanted to look into turning off for a while now.

selectors/views/custom_status.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function getRecentCustomStatuses(state: GlobalState) {
+    const preferences = getMyPreferences(state);
+    const key = getPreferenceKey(Preferences.CATEGORY_CUSTOM_STATUS, Preferences.NAME_RECENT_CUSTOM_STATUSES);
+    return preferences[key] && preferences[key].value ? JSON.parse(`${preferences[key].value}`) : [];
```
hmhealey, line 21: Same thing here where we don't want to return a new empty array each time. This doesn't need a selector factory though since it has no arguments, so you just need something like
```javascript
export const getRecentCustomStatuses = createSelector(
    (state) => get(Preferences.CATEGORY_CUSTOM_STATUS, Preferences.NAME_RECENT_CUSTOM_STATUSES),
    (value) => {
        return value ? JSON.parse(value) : [];
    },
);
```

selectors/views/custom_status.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function isCustomStatusEnabled(state: GlobalState) {
+    const config = getConfig(state);
+    return config && config.EnableCustomUserStatuses === 'true';
+}
+
+function showCustomStatusPulsatingDotAndPostHeader(state: GlobalState) {
+    const preferences = getMyPreferences(state);
```
hmhealey, line 30: Similarly, this can be simplified to just call the `get` selector from the preferences selectors
```javascript
const hasViewedCustomStatusModal = get(Preferences.CATEGORY_CUSTOM_STATUS, Preferences.NAME_CUSTOM_STATUS_TUTORIAL_STATE);
return hasViewedCustomStatusModal !== Preferences.CUSTOM_STATUS_MODAL_VIEWED;
```

hmhealey (changes requested):
Apologies for this taking so long and for dumping a ton of suggested changes relatively late in the process, but there's a lot of stuff here, and I wanted to make sure I understood it. I'll do my best to be responsive on this since I know we want to get it in as soon as possible.

components/custom_status/custom_status_modal.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+import CustomStatusSuggestion from './custom_status_suggestion';
+
+type Props = {
+    onHide: () => void;
+};
+
+const EMOJI_PICKER_WIDTH_OFFSET = 308;
+const defaultCustomStatusSuggestions: UserCustomStatus[] = [
+    {emoji: 'calendar', text: 'In a meeting'},
```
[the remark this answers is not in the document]
esethna: Yes agree

components/custom_status/custom_status_emoji.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const store = mockStore({});
+
+    it('should match snapshot', () => {
+        const wrapper = shallow(
+            <Provider store={store}>
+                <CustomStatusEmoji/>
+            </Provider>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
```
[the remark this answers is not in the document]
calebroseland:
```tsx
mount(thing, {wrappingComponent: Provider, wrappingComponentProps: {store}})
```
This maybe?

components/post_view/post_list/index.js
```diff
@@ -64,13 +65,17 @@ function makeMapStateToProps() {
...
+        let currentUserLastPostId;
         if (postIds) {
             formattedPostIds = preparePostIdsForPostList(state, {postIds, lastViewedAt, indicateNewMessages: true, channelId});
             if (postIds.length) {
                 const latestPostId = memoizedGetLatestPostId(postIds);
                 const latestPost = getPost(state, latestPostId);
                 latestPostTimeStamp = latestPost.create_at;
                 latestAriaLabelFunc = createAriaLabelForPost(state, latestPost);
+                if (showPostHeaderUpdateStatusButton(state)) {
+                    currentUserLastPostId = getLastCurrentUserPostWithHeaderId(state, postIds);
```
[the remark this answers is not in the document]
hmhealey: Alright, so I talked with Eric, Matt, and Dean about this, and we came to the conclusion that this will be very difficult to do with the way the post list is currently structured, specifically because determining the most recent post that fits those requirements is very hard. We have some ideas for solving that in the future for a more general tutorial system, but we think this isn't going to be problem for custom statuses.

We decided to compromise and show the Set a Custom Status on every post that meets the requirements of being made by the current user and having a header. It might be a bit more annoying for the user, but I, personally, think that might make it so users are more likely to click on it.

hmhealey (changes requested):

components/custom_status/custom_status_modal.tsx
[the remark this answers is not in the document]
chetanyakan: @hmhealey This makes sense. I will make that change.

A point to note, however, is that we will require someone to test that the suggestions filtering feature to make sure it works with other languages (i.e, not showing the same status as both a suggestion and recent status).

(cc: @esethna)

components/custom_status/custom_status_emoji.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+                    tooltipDirection='bottom'
+                />
+            </Provider>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    it('should not render when EnableCustomStatus in config is false', () => {
+        (CustomStatusSelectors.isCustomStatusEnabled as jest.Mock).mockReturnValue(false);
```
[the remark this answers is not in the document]
chetanyakan: @hmhealey I think this test case is fine right now. Here's why:

1. We should mock the selectors in components so that we don't have to update all components that use a particular selector when the store hierarchy of a reducer changes.
2. We can have separate test cases that cover whether each selector detects changes in the correct reducer/store hierarchy.

Thoughts?

components/channel_header/channel_header.js
```diff
@@ -286,6 +299,72 @@ class ChannelHeader extends React.PureComponent {
...
+        let customStatusTextComponent = (
+            <span
+                className='overflow--ellipsis text-nowrap'
+                ref={(element) => {
+                    this.customStatusTextRef = element;
+                    this.showCustomStatusTextTooltip();
+                }}
+            >
+                <Markdown
+                    message={customStatus.text}
```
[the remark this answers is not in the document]
chetanyakan: @hmhealey I see your point. Allowing markdown here allows custom statuses which are images/Gifs, which was not our original intention.

We used markdown here to allow emojis to be rendered in the custom status messages.

Screenshot from 2021-02-15 19-35-12

This requirement was mentioned somewhere in the spec. If we remove it here, we won't be able to allow support for that feature.

@esethna @matthewbirtch thoughts?

components/custom_status/custom_status_modal.tsx
```diff
@@ -0,0 +1,292 @@
...
+    const isStatusSet = emoji || text;
+    const firstTimeModalOpened = useSelector((state: GlobalState) => showStatusDropdownPulsatingDot(state));
+
+    const handleCustomStatusInitializationState = () => {
+        if (firstTimeModalOpened) {
+            dispatch(setCustomStatusInitialisationState({hasOpenedSetCustomStatusModal: true}));
+        }
+    };
+
+    useEffect(handleCustomStatusInitializationState, []);
```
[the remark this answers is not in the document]
chetanyakan: @hmhealey As I understand, passing an empty array as the dependencies will cause this `useEffect` to act as `componentDidMount`.

As it is, this won't be called again, no matter which props/state change. So, I don't see a point in updating this.
Thoughts?

components/profile_popover/index.js
```diff
@@ -63,6 +65,8 @@ function mapStateToProps(state, ownProps) {
         teamUrl: getCurrentRelativeTeamUrl(state),
         user: getUser(state, userId),
         modals: state.views.modals.modalState,
+        customStatus: getCustomStatus(state, userId),
+        isCustomStatusEnabled: isCustomStatusEnabled(state),
```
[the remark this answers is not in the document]
chetanyakan: @hmhealey Although these are called together, we can't merge them in the way you recommend.

We use custom status to check if the custom status is set or not.
Even if the custom status is unset, we still need to check if the feature is disabled to hide the components that allow setting the status.

As such, I think what we have should be fine.
Thoughts?

components/profile_popover/profile_popover.scss
```diff
@@ -0,0 +1,6 @@
+.user-popover__set-custom-status-btn {
+    color: rgb(35, 137, 215);
```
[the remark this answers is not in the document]
chetanyakan: @matthewbirtch what theme colours would you recommend here?

This is the profile popover button to set a status.
image

components/custom_status/custom_status_emoji.tsx
```diff
@@ -0,0 +1,90 @@
...
+import {getCustomStatus, isCustomStatusEnabled} from 'selectors/views/custom_status';
+import {GlobalState} from 'types/store';
+import Constants from 'utils/constants';
+import Markdown from 'components/markdown';
+
+interface ComponentProps {
+    emojiSize?: number;
+    showTooltip?: boolean;
+    tooltipDirection?: 'top' | 'right' | 'bottom' | 'left';
+    emojiStyle?: React.CSSProperties;
```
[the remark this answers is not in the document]
chetanyakan: @hmhealey I am not sure how much performance we can gain by using CSS classes here.
The styles need to be customizable per usage, so we decided to use inline styles here.
I think as long as the props don't change, the component won't re-render.
As such, what do you propose we change here?

utils/position_utils.tsx
```diff
@@ -6,6 +6,7 @@ export function popOverOverlayPosition(
     innerHeight: number,
     spaceRequiredAbove: number,
     spaceRequiredBelow?: number,
+    defaultHorizontalPosition?: 'left' | 'right',
```
michelengelen, line 9: IMO this should be renamed to `horizontalPosition`, since a default should be done as a variable inside the component, or as a constant imported to it. I know the reason behind it, and it might be only my personal taste/style, but it does not feel right to name it that way.

components/emoji_picker/emoji_picker_tabs.jsx
```diff
@@ -74,9 +76,12 @@ export default class EmojiPickerTabs extends PureComponent {
                 pickerStyle = {...this.props.style};
             }
 
-            if (pickerStyle.top) {
+            if (pickerStyle.top || this.props.topOffset) {
```
michelengelen, line 79: this should be `if (pickerStyle.top && this.props.topOffset) {` since it will break if `pickerStyle.top` is `undefined`

michelengelen (changes requested):

components/status_dropdown/status_dropdown.jsx
chetanyakan: @hmhealey We really do need to pass the sibling here to keep the HTML semantic.

If you refer to earlier comments by @deanwhillier and @michelengelen, you'll see that we had earlier implemented it by using a method similar to what you proposed.

But, in the interest of using correct tags in the generated HTML (buttons for both clickable elements and li for the list item), we had to implement it this way.

Any advice on how we could implement this in a better way?

components/status_dropdown/status_dropdown.jsx
chetanyakan: @deanwhillier Removing the span is not an issue, but the visual hover effect you mentioned is somewhat tricky.

Let me try to explain why.

Here is the status dropdown.
image

The generated HTML looks something like this:

```html
<li>
    <button>status</button>
    <button>clear</button>
</li>
<li>
    <button>online</button>
</li>
<li>
    <button>away</button>
</li>
...
```


In the rest of the statuses (`online`, `away`), the hover effect is controlled by a CSS class on the button element.

Now, for the custom status as well, the hover state is controlled by the `status` button.

Since the `clear` button is added on top of the `status` button by using `position: absolute`, we can't detect the hover state on the `status` button while hovering over the `clear` button.

The alternative is to control the hover state using the `<li>` tag, which is a big change since we have to change and test the other status buttons as well.

sass/components/_emoticons.scss
chetanyakan: Yes. We had to make:

`z-index(EmojiPicker) > z-index(CustomStatusModal / GenericModal)`

to be able to select emojis from the custom status modal.

sass/layout/_sidebar-left.scss
chetanyakan: @hmhealey Thank you for pointing this out. We were able to identify and fix a related issue.

The name width is changed from 100% to a max of 85% to support custom emojis next to the name. 

image

components/custom_status/custom_status_modal.tsx
```diff
@@ -0,0 +1,284 @@
...
+
+    const clearHandle = () => {
+        setText('');
+        setEmoji('');
+    };
+
+    const disableSetStatus = (currentCustomStatus.text === text && currentCustomStatus.emoji === emoji) ||
+        (text === '' && emoji === '');
+
+    const handleSuggestionClick = (status: any) => {
```
michelengelen, line 130: do we really need an `any` here?
IMO we should refrain from using `any` if possible.

Maybe we can create a type `CustomStatusOption`?

```
type CustomStatusOption = {
  emoji: string;
  text: string;
}
```

components/custom_status/custom_status_modal.tsx
```diff
@@ -0,0 +1,284 @@
...
+                <>
+                    <div className='statusSuggestion__title'>
+                        {formatMessage({id: 'custom_status.suggestions.title', defaultMessage: 'SUGGESTIONS'})}
+                    </div>
+                    {customStatusSuggestions}
+                </>
+            );
+        }
+
+        return null;
```
michelengelen, line 196: please switch the if-clause and use immediate null-return instead:

```suggestion
        if (customStatusSuggestions.length <= 0) {
            return null;
        }

        return (
            <>
                <div className='statusSuggestion__title'>
                    {formatMessage({id: 'custom_status.suggestions.title', defaultMessage: 'SUGGESTIONS'})}
                </div>
                {customStatusSuggestions}
            </>
        );
```

components/custom_status/custom_status_suggestion.tsx
```diff
@@ -0,0 +1,95 @@
...
+            className='statusSuggestion__row cursor--pointer'
+            onMouseEnter={showClearButton}
+            onMouseLeave={hideClearButton}
+            onClick={
+                () => handleSuggestionClick(
+                    {
+                        emoji,
+                        text,
+                    })
+            }
```
michelengelen, line 77: just minor styling:

```suggestion
            onClick={() => handleSuggestionClick({ emoji, text })}
```

components/emoji/render_emoji.tsx
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import React from 'react';
+import {useSelector} from 'react-redux';
+
+import {getEmojiImageUrl} from 'mattermost-redux/utils/emoji_utils';
+
+import {getEmojiMap} from 'selectors/emojis';
+import {GlobalState} from 'types/store';
+interface ComponentProps {
```
michelengelen, line 10: please add a line above

components/emoji/render_emoji.tsx
```diff
@@ -0,0 +1,54 @@
...
+    emojiName: string;
+    size?: number;
+    emojiStyle?: React.CSSProperties;
+    onClick?: () => void;
+}
+
+const RenderEmoji = ({emojiName, emojiStyle, size, onClick}: ComponentProps) => {
```
michelengelen, line 17: this is where the component should probably wrapped in `React.memo( ... )` since a shallow compare for the CSS properties would prevent a re-render when they are equal

components/post_view/post_header/post_header_custom_status.tsx
```diff
@@ -0,0 +1,67 @@
...
+                emojiStyle={{
+                    margin: '4px 0 0 4px',
+                }}
+            />
+        );
+    }
+
+    if (!showUpdateStatusButton) {
+        return null;
+    }
```
michelengelen, line 48: this can be added to the if-clause above (lines 29 - 31)

components/post_view/post_header/post_header_custom_status.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    const updateStatus = () => dispatch(setStatusDropdown(true));
+    return (
```
michelengelen, line 51: please add a line here as separation

components/sidebar/sidebar_channel/sidebar_channel_link/sidebar_channel_link.tsx
```diff
@@ -203,6 +206,17 @@ export default class SidebarChannelLink extends React.PureComponent<Props, State
...
+            <CustomStatusEmoji
+                userID={this.props.teammateId}
+                showTooltip={true}
+                emojiStyle={{
+                    marginLeft: 6,
+                    marginBottom: 1,
+                }}
+            />
+        );
+
```
michelengelen, line 219: please use a ternary here and  set the value to `null` on the second expression

sass/layout/_headers.scss
```diff
@@ -446,8 +446,22 @@
     }
 
     .header-status__text {
+        display: flex;
         margin-right: 7px;
         line-height: 16px;
+        max-width: calc(100% - 225px);
+        align-items: center;
+
+        p{
```
michelengelen, line 455: please add a space after the selector

sass/layout/_headers.scss
```diff
@@ -704,12 +718,16 @@
             }
         }
 
+        .header__info__content{
```
michelengelen, line 721: please add a space after the selector

sass/layout/_headers.scss
```diff
@@ -727,8 +745,31 @@
             font-size: 14px;
             font-weight: 400;
             top: -3px;
+            padding-right: 4px;
             position: relative;
         }
+        .custom-status {
```
michelengelen, line 751: please add a line here as a separator

sass/layout/_headers.scss
```diff
@@ -727,8 +745,31 @@
...
+            &-text {
+                margin-left: 4px;
+            }
+    
+            &-emoji {
+                display: flex;
+                height: 16px;
+                min-width: 16px;
+            }
+            &-emoji > span:first-child {
```
michelengelen, line 765: please add a line here as a separator

sass/layout/_headers.scss
```diff
@@ -727,8 +745,31 @@
...
+                height: 16px;
+                width: 16px;
+                min-height: 16px;
+                min-width: 16px;
+            }
+        }
+
```
michelengelen, line 772: please remove this extra line

michelengelen (changes requested):
there are still some changes necessary before I can approve this. Sorry! 🙇🏻

components/post_view/post_header/post_header_custom_status.tsx
chetanyakan: @michelengelen This must be checked after checking for `isCustomStatusSet` and cannot be merged with the previous if condition in the way that you recommend.

I'll add a comment over this, so no one tries to make that change.

components/channel_header/channel_header.js
```diff
@@ -42,6 +42,9 @@ import * as Utils from 'utils/utils';
 
 import ChannelHeaderPlug from 'plugins/channel_header_plug';
 
+import CustomStatusEmoji from 'components/custom_status/custom_status_emoji';
+import CustomStatusText from 'components/custom_status/custom_status_text';
+
 import HeaderIconWrapper from './components/header_icon_wrapper';
 import UserGuideDropdown from './components/user_guide_dropdown';
```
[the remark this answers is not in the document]
chetanyakan: @michelengelen Apparently I can't change the headers as you suggested.

These files are defined in a component directory inside the current directory.

What I would have to do is:

```suggestion
import CustomStatusEmoji from 'components/custom_status/custom_status_emoji';
import CustomStatusText from 'components/custom_status/custom_status_text';

import HeaderIconWrapper from 'components/channel_header/components/header_icon_wrapper';
import UserGuideDropdown from 'components/channel_header/components/user_guide_dropdown';
```

Would you like me to change it like that??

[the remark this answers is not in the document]
michelengelen: no, I did not notice that ... I am sorry ... when referencing files inside the current directory it is fine to use relative paths.
Although I am pretty sure that the naming of the folder should be reconsidered. 🤔

michelengelen (approved):
LGTM now ... thanks @chetanyakan !

esethna:
@ogi-m if you can please prioritize testing on this feature today to hopefully unblock merging all the PRs (web, server, redux). You can open a test account here: http://ec2-13-58-138-215.us-east-2.compute.amazonaws.com:8065/signup_user_complete/?id=559fit8sj7bm8n7mcpqcz9cy5y

ogi-m:
Several concerns/questions:

1. The user is able to type an unlimited amount of characters and click "Set status" but then most of the text will be lost upon saving the status. Should we maybe add a character count to the field when a user goes over the character limit (e.g. "-2" if the user is 2 characters over the limit) and disable the "Set status" button? 

2. According to the spec. the default statuses shouldn't be moved to the "recent" list and should remain at the bottom, but that's not the case currently, they do get bumped to the top.

3. Should the "Clear Status" button be hidden if there's no status set, to match the design?

4. The emoji picker opens above the modal. I was expecting it to open below. Just want to confirm this is ok?

custom_status_emoji


@chetanyakan @esethna @matthewbirtch

components/channel_header/channel_header.js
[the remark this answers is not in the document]
hmhealey: Since the status already has an emoji as part of it, I'd lean towards just using that. We could always go back and add emoji support later on if we decide we want it too since it's a bit easier to add than remove.

That said, I'll defer to Eric and Matt here, and I'll make sure they see this

components/custom_status/custom_status_emoji.test.tsx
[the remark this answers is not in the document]
hmhealey: Yeah, that's very true. I guess I was thinking of the easy case where a selector is just returning something simple, but even in those cases, we still end up needing to make up a large part of the state tree, and that's not something the people implementing a component need to care about.

We haven't really ran into this before since class components are tested without the redux connection, but that also means the connector often doesn't get tested, so this covers that. The more I think about it, this does seem like the right way to test the component, and the selector can be tested on its own.

components/custom_status/custom_status_modal.tsx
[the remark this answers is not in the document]
hmhealey: Oh, interesting. I assumed it worked the same way as not passing a second argument. That's good to know.

components/profile_popover/index.js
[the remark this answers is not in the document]
hmhealey: Yeah, that's fine. I should've realized the unset case would be relevant in cases like for the status dropdown.

components/profile_popover/profile_popover.scss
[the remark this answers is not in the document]
hmhealey: Other links in the profile popover use `var(--link-color)`, so I assume this'll be the same

components/status_dropdown/status_dropdown.jsx
hmhealey: Nothing easy comes to mind. Perhaps we could do a custom menu item component so we could have more control over the split menu item, but I'm not sure if that's even really possible with the current component like I mentioned. Let's leave this as-is, and we could always revisit it later when we're rewriting the menu components.

sass/components/_emoticons.scss
hmhealey: Fun. Modals and popovers are another thing we need to tackle when it comes to rewriting the menu components like I mentioned before, so hopefully we'll get to a point where we don't need to fight with z-indexes any more.

components/channel_header/channel_header.js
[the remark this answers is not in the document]
esethna: @chetanyakan I'd propose we don't support markdown for now in the status text. This opens up too many corner cases to validate now. We may include this as a future update

esethna:
Thanks for the quick review @ogi-m

1) Can we open a follow-up ticket for this to implement this max length validation component on the status modal? I think this is acceptable as for now and we'll want to implement this component throughout the app wherever max length validation is needed. https://github.com/mattermost/mattermost-webapp/pull/7281

2) This was changed during the implementation process, I'll update the spec accordingly

3) Good point, the clear status button shouldn't exist if no status is set. @ogi-m can you help file a follow-up ticket for this since it's not a blocking issue? 

image

4) I am not able to reproduce this, it probably depends on the screen resolution/aspect being used. However, this behaviour will be ironed out when the webapp team gets to the menu/pop-over common component, so not an issue now.

sass/layout/_sidebar-left.scss
hmhealey: Should that be needed? I thought we size those items with flexbox so that would scale automatically. We size it dynamically with the mention badge, but maybe that's done manually instead of with flexbox. Right now, long channel names are cut off early even if they're not DMs or if they don't have the status icon.

Test server:
Screen Shot 2021-02-16 at 2 40 27 PM
My server:
Screen Shot 2021-02-16 at 2 35 57 PM

hmhealey:
I took another pass through this, and most of the stuff I commented on has been addressed, but there's a few more that still need to be looked into.

ogi-m:
One last issue I'm seeing is the alignment of the status emoji and the text, the emoji is placed too low (confirmed with @matthewbirtch ):
statuses_alignment_1 statuses_alignment

package-lock.json
chetanyakan: Sure. This probably got changed due to different NPM versions.

chetanyakan:
@ogi-m @hmhealey I have ironed out all the aforementioned issues. This PR is good for re-review.

There were 2 issues, however, that are particularly complex and are still unresolved.

1. Custom status emoji icon with channel members dropdown usename overflow

Consider the image below:
image

Note that the first user is the current user and the second one (which is being hovered) is another user, who can be sent messages and thus the comment icon.

Now, if the second user is not hovered, it looks like this:
image

This is because we are using `opacity: 0` instead of `display: none` to hide that comment icon.
I don't know the correct path here, without much investigation and would need guidance on what to change here.

2. Custom status emoji icon with LHS DMs usename overflow

This was already pointed out a bit by @hmhealey as a PR comment above.
On making that change, I was not able to align the emoji with the `+` icon, without causing the problem pointed out by @hmhealey again.

image


Re: problem otherwise (see comments above)
image


@matthewbirtch @esethna  Any guidance on these 2 issues will be appreciated.

package-lock.json
chetanyakan: Will update this while updating the version of `mattermost-redux` commit in `package.json`

components/status_dropdown/status_dropdown.jsx
deanwhillier: Yup, I totally get it! We'll be rebuilding the menu components in the near future so can revisit this later. Thanks for working through all the requested changes!

deanwhillier (approved):

chetanyakan:
Updated test server: http://ec2-52-14-2-53.us-east-2.compute.amazonaws.com:8065/

ogi-m:
Thanks @chetanyakan. The alignment looks better to me now, except for the RHS reply thread. The root post has a different alignment from the rest of the posts below it. The emoji is too high up and there's less space between the emoji and the username:

custom_status_rhs
cc @matthewbirtch

chetanyakan:
@ogi-m Pushed a fix. These should be identical now.

image

package-lock.json
hmhealey: It might also have been something weird like `npm install` vs `npm add` with a new version of mattermost-redux. `npm` is weird like that sometimes

selectors/views/custom_status.ts
```diff
@@ -0,0 +1,46 @@
...
+import {memoizeResult} from 'mattermost-redux/utils/helpers';
+
+import {GlobalState} from 'types/store';
+
+export function makeGetCustomStatus() {
+    return memoizeResult((state: GlobalState, userID?: string) => {
+        const user = userID ? getUser(state, userID) : getCurrentUser(state);
+        const userProps = user.props || {};
+        return userProps.customStatus && JSON.parse(userProps.customStatus);
+    });
```
hmhealey, line 19: This doesn't quite work since `state` will be changing constantly. Instead, you can use `createSelector` and vary the user that's passed into its final function based on whether or not the user ID is provided
```suggestion
    return createSelector(
        (state: GlobalState, userID?: string) => userID ? getUser(state, userID) : getCurrentUser(state),
        (user) => {
            const userProps = user.props || {};
            return userProps.customStatus && JSON.parse(userProps.customStatus);
        },
    );
```
This will make it so that the returned value is only changed whenever `user` changes rather than whenever anything in `state` changes. You could go a bit further and make it only pass the props into that final function, but that's probably not necessary.

hmhealey (changes requested):
Mostly everything looks good, but one more thing here to make the `getCustomStatus` selector work properly.

hmhealey (approved):

ogi-m:
Thanks @chetanyakan, the alignment in the RHS is unified now, but I've noticed it's different from the alignment in the centre channel. It would be great if we could have the emojis in the RHS placed just a bit higher up to match the alignment in the centre channel.

alignment

chetanyakan:
@ogi-m Center-channel and RHS post headers use the same styles now
image

ogi-m:
Thanks @chetanyakan, looks good now. Approving 👍🏻 

I'll open a ticket for the issues you [noted](https://github.com/mattermost/mattermost-webapp/pull/7441#issuecomment-780222868).

ogi-m (approved):

amyblais:
To be cherry-picked to ``cloud`` branch.

PR #7446, opened 2021-02-01
MM-31501 - Add upgrade-now link for admins in invite users

pablovelezvidal:
#### Summary
MM-31501 - Add upgrade-now link for admins in invite users.
For all administrators who are prevented from inviting users due to free tier limitations in either step 3 of the Getting Started onboarding experience, or in the Invite Members modal - include a new CTA to ‘Upgrade Mattermost Cloud’ which opens the subscription modal window and self-service subscription flow. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31501

#### Screenshots
upgradeLink

inviteMerberMultiUser


ethervoid (approved):
LGTM! Great work! Just a minor thingy

components/widgets/links/upgrade_link.tsx
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
```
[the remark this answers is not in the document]
pablovelezvidal: hey @ethervoid, thanks for the review! ok, I will organize the imports then. Thanks for the comment.

components/widgets/links/upgrade_link.tsx
```diff
@@ -0,0 +1,51 @@
...
+import {FormattedMessage} from 'react-intl';
+
+import {openModal} from 'actions/views/modals';
+
+import {trackEvent} from 'actions/telemetry_actions';
+import {ModalIdentifiers} from 'utils/constants';
+
+import PurchaseModal from 'components/purchase_modal';
+
+const UpgradeLink: React.FC = () => {
```
marianunez, line 19: Could we add a test for this new component?

components/widgets/inputs/users_emails_input.scss
```diff
@@ -251,13 +251,26 @@
...
-        text-align: center;
-        color: var(--error-text)
+        span {
+            height: 20px;
+            position: relative;
+            font-size: 16px;
+            line-height: 20px;
+            text-align: center;
+            color: var(--error-text)
+        }
```
marianunez, line 263: Does this css change affect any other usage of this common component?

components/widgets/inputs/users_emails_input.jsx
```diff
@@ -311,18 +312,22 @@ export default class UsersEmailsInput extends React.PureComponent {
...
+                                disableLinks={true}
+                            >
+                                {(message) => (
+                                    <components.NoOptionsMessage>
+                                        {message}
+                                    </components.NoOptionsMessage>
+                                )}
+                            </FormattedMarkdownMessage>
+                            {this.props.errorMessageId === 'invitation_modal.invite_members.hit_cloud_user_limit' &&
+                            <UpgradeLink/>}
```
marianunez, line 329: In order to keep this component generic and not have extra cloud logic here, could we pass in message as a component that the parent will set with the new `UpgradeLink` set as needed?

marianunez (changes requested):

stevemudie (approved):
Looks good!

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
```diff
@@ -32,6 +36,7 @@ type Props = StepComponentProps & {
         sendEmailInvitesToTeamGracefully: (teamId: string, emails: string[]) => Promise<{ data: TeamInviteWithError[]; error: ServerError }>;
         regenerateTeamInviteId: (teamId: string) => void;
         getSubscriptionStats: () => void;
+        openModal?: (modalData: { modalId: string; dialogType: any; dialogProps?: any }) => void;
```
marianunez, line 39: Given that we bind this prop for `openModal` any special reason this is marked as optional?

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
```diff
@@ -313,6 +327,9 @@ class InviteMembersStep extends React.PureComponent<Props, State> {
                                             <span>{this.state.emailError}</span>
                                         </>
                                     }
+                                    {(this.state.emailError && this.state.emails.length >= this.props?.subscriptionStats!.remaining_seats) &&
```
marianunez, line 330: Not sure I follow the usage of the non-null assertion here on `subscriptionStats`, I know we have moved the load to the root but shouldn't we do a null operator here to be safe instead and not assume it will not be null?

I would expect something like:  
```suggestion
                                    {(this.state.emailError && this.state.emails.length >= this.props.subscriptionStats?.remaining_seats) &&
```

components/widgets/inputs/users_emails_input.scss
marianunez: Are we certain that moving this span within the wrapper does not affect other places where the `users_email_input` is used?

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
```diff
@@ -250,6 +255,15 @@ class InviteMembersStep extends React.PureComponent<Props, State> {
...
     }
 
+    handleLinkClick = async (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+        e.preventDefault();
+        trackEvent('upgrade_mm_cloud', 'click_upgrade_invite_members_step');
+        this.props.actions.openModal!({
+            modalId: ModalIdentifiers.CLOUD_PURCHASE,
+            dialogType: PurchaseModal,
+        });
+    };
```
marianunez, line 265: Wondering if it makes sense to move this to the `UpgradeLink` component to avoid duplication when it's used

components/widgets/inputs/users_emails_input.scss
AGMETEOR: @pablovelezvidal wouldn't adding ```justify-content: center;``` to ```.InputErrorBox``` and then removing ```margin: auto;``` from ```.InputErrorBox > span``` do the trick?

pablovelezvidal: thanks @AGMETEOR it worked perfect!

AGMETEOR (approved):

package-lock.json
```diff
@@ -13745,6 +13748,12 @@
         "postcss": "^7.0.14"
       }
     },
+    "icu4c-data": {
+      "version": "0.64.2",
+      "resolved": "https://registry.npmjs.org/icu4c-data/-/icu4c-data-0.64.2.tgz",
+      "integrity": "sha512-BPuTfkRTkplmK1pNrqgyOLJ0qB2UcQ12EotVLwiWh4ErtZR1tEYoRZk/LBLmlDfK5v574/lQYLB4jT9vApBiBQ==",
+      "dev": true
+    },
```
marianunez, line 13756: Is this change intended?

marianunez (approved):
Great work @pablovelezvidal!

package-lock.json
pablovelezvidal: not intended, although it has been being added automatically when running the check-styles task. Not sure if it is a dependency that should now be added to the lock file.

marianunez: We usually avoid these isolated changes only on the package-lock files unless there is an explicit dependency change/upgrade. We can revert this

pablovelezvidal: ok!

pablovelezvidal: done!

stevemudie (commented):
This works but I noticed some things that you may want to address before pushing it out.

Steps to reproduce:
1 - Go to ☰ menu ➜ Invite People and type in 3 email addresses (this will trigger the 'Upgrade now' link)
2 - Click the link to open the modal then dismiss it by tapping the escape key

*Expected*: Returned to the 'Invite Members to Main Team' screen
*Observed*: You're returned but a modal appears telling you have unsent invitations asking you if you'd like to discard them. When you click the cancel button then you're momentarily unable to do anything else on the screen.

image

*Notes*:
- After a few seconds the screen becomes responsive again and you can click to upgrade again or delete an address etc.
- The main concern is when trying to escape the upgrade it thinks you want to leave completely

pablovelezvidal:
> This works but I noticed some things that you may want to address before pushing it out.
> 
> Steps to reproduce:
> 1 - Go to ☰ menu ➜ Invite People and type in 3 email addresses (this will trigger the 'Upgrade now' link)
> 2 - Click the link to open the modal then dismiss it by tapping the escape key
> 
> _Expected_: Returned to the 'Invite Members to Main Team' screen
> _Observed_: You're returned but a modal appears telling you have unsent invitations asking you if you'd like to discard them. When you click the cancel button then you're momentarily unable to do anything else on the screen.
> 
> image
> 
> _Notes_:
> 
> * After a few seconds the screen becomes responsive again and you can click to upgrade again or delete an address etc.
> * The main concern is when trying to escape the upgrade it thinks you want to leave completely

Hey @stevemudie , thanks for the review! Ok, I have investigated what you saw and this is what I found:
- Even when the issue raised in this PR, the real problem happens because of having two modal opened at the same time, and the listener for the escape keypress event responds for both modals without differentiating which one is in top of the other, it means not understanding the user intention of which window to close. It is worth noting that this only happens on the escape keypress event.
- Also, the problem of the screen not responding for a couple of seconds after pressing cancel on the invitation modal exit confirmation window, it's a current misbehavior out of this changes.

Based on this, I already tried to find a solution for the esc-keypress problem, and that would imply modifying the fullScreenModal component behavior along with the store state, so we can know which windows are opened on top of each other and then decide which one to close on the keypess event and stop it to bubble up.
Those changes are not simple/easy changes, so to not polute more the intention of this PR I suggest to open a new defect that covers the two-modals esc-keypress event issue, and merge this PR (5/5). Any objections to this suggestion @stevemudie , @marianunez ?

marianunez:
> > This works but I noticed some things that you may want to address before pushing it out.
> > Steps to reproduce:
> > 1 - Go to ☰ menu ➜ Invite People and type in 3 email addresses (this will trigger the 'Upgrade now' link)
> > 2 - Click the link to open the modal then dismiss it by tapping the escape key
> > _Expected_: Returned to the 'Invite Members to Main Team' screen
> > _Observed_: You're returned but a modal appears telling you have unsent invitations asking you if you'd like to discard them. When you click the cancel button then you're momentarily unable to do anything else on the screen.
> > image
> > _Notes_:
> > 
> > * After a few seconds the screen becomes responsive again and you can click to upgrade again or delete an address etc.
> > * The main concern is when trying to escape the upgrade it thinks you want to leave completely
> 
> Hey @stevemudie , thanks for the review! Ok, I have investigated what you saw and this is what I found:
> 
> * Even when the issue raised in this PR, the real problem happens because of having two modal opened at the same time, and the listener for the escape keypress event responds for both modals without differentiating which one is in top of the other, it means not understanding the user intention of which window to close. It is worth noting that this only happens on the escape keypress event.
> * Also, the problem of the screen not responding for a couple of seconds after pressing cancel on the invitation modal exit confirmation window, it's a current misbehavior out of this changes.
> 
> Based on this, I already tried to find a solution for the esc-keypress problem, and that would imply modifying the fullScreenModal component behavior along with the store state, so we can know which windows are opened on top of each other and then decide which one to close on the keypess event and stop it to bubble up.
> Those changes are not simple/easy changes, so to not polute more the intention of this PR I suggest to open a new defect that covers the two-modals esc-keypress event issue, and merge this PR (5/5). Any objections to this suggestion @stevemudie , @marianunez ?

Agreed, let's track this issue and we can do a fix post merge to not block this PR any further.

pablovelezvidal:
Track defect created https://mattermost.atlassian.net/browse/MM-32901

stevemudie:
Thanks for the clear explanation of what's behind this @pablovelezvidal !

PR #7449, opened 2021-02-02
[MM-32121] Fix neverending load on one-paged search results

mgdelacroix:
#### Summary
This PR adds a second page search right after the first one, to be able to ensure that there are no more results, or enable the load more scrolling if there are. It fixes the infinite loading indicator on one-paged searches that we were experiencing in community.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32121

### DIFF
--- a/components/search_results/search_results.tsx
+++ b/components/search_results/search_results.tsx
@@ -69,6 +69,15 @@ const SearchResults: React.FC<Props> = (props: Props): JSX.Element => {
         scrollbars.current?.scrollToTop();
     }, [props.searchTerms]);
 
+    useEffect(() => {
+        // after the first page of search results, there is no way to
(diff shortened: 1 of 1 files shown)


jespino (approved):
Looks good to me.

lindy65 (approved):
Thanks @mgdelacroix - looks good to me 👍

isacikgoz (approved):
Nice one! 🎉

amyblais:
@mgdelacroix Is this a recent regression and do you think it would be good to cherry-pick this to cloud and v5.32, or is this just a minor bug?

mgdelacroix:
@amyblais the bug is minor in the sense that it's just cosmetic, but I think we should cherry-pick it.

PR #7454, opened 2021-02-02
Cypress/E2E: Split file preview e2e tests

josephbaylon:
#### Summary
This was causing a crash on my cypress locally so I had to split
- Split orig `file_previews_1_spec` to `file_previews_1_spec` through `file_previews_4_spec`
- Renamed orig `file_previews_2_spec` to `file_previews_5_spec`

#### Ticket Link
NA


e2e/cypress/integration/files_and_attachments/file_previews_1_spec.js
```diff
@@ -14,7 +14,7 @@ import * as TIMEOUTS from '../../fixtures/timeouts';
 
 import {downloadAttachmentAndVerifyItsProperties} from './helpers';
 
-describe('Upload Files', () => {
+describe('Upload Files - Image', () => {
```
saturninoabril, line 17: I really like this grouping per file type, would it make sense to separate each file type into its own test case?  Feels like this single test case has a huge weight to be measured as a single unit of test.

e2e/cypress/integration/files_and_attachments/file_previews_5_spec.js
```diff
@@ -0,0 +1,222 @@
...
+    it('MM-T2264_22 - FLAC', () => {
+        const properties = {
+            route: 'mm_file_testing/Audio/FLAC.flac',
+            shouldPreview: true,
+        };
+        testAudioFile(properties);
+    });
+});
+
+function testAudioFile(fileProperties) {
```
saturninoabril, line 77: Some of the helper functions were repeated. Can those be extracted into a separate helper file within the same folder?

saturninoabril (commented):
Thanks Joseph. Left couple of comments.

e2e/cypress/integration/files_and_attachments/file_previews_1_spec.js
josephbaylon: @saturninoabril I'm not quite sure I follow. Each file type is already it's own test case. You can see that each TM4J tag is assigned to each file type

e2e/cypress/integration/files_and_attachments/file_previews_5_spec.js
josephbaylon: Sure

josephbaylon: @saturninoabril done.

e2e/cypress/integration/files_and_attachments/file_previews_1_spec.js
saturninoabril: Oopss, sorry. I meant MM-T2264 for images, then new TM4J test case each for video, audio, generic, etc.

josephbaylon: Ah, gotcha.

josephbaylon: @saturninoabril updated

prapti (approved):
Thank you Joseph! These tests look great and makes all the sense as they are now separated on file type. The ones in file_preview_5 aren't passing because the files are only loading previews. Wondering if that's why that one is not tagged for prod :)

saturninoabril (approved):
Great and thanks @josephbaylon!

Just one observation which we can discuss/agree next meeting. I feel like the test case with multiple steps should collocate in the same spec file. So that when ran in any stage (prod or unstable), test execution result will be saved into Zephyr. Note that Zephyr does not permit recording/saving result if test steps do not match.

PR #7458, opened 2021-02-03
MM-32226: adjustted "more unreads"-indicator positioning

michelengelen:
#### Summary

* doubled top-positioning fo "more unreads"-indicator
* added double-quotes for UTF-8 charset setting
* fixed path-resolving in sass files

#### Ticket Link
Fixes [MM-32226](https://mattermost.atlassian.net/browse/MM-32226)

#### Screenshots
Screenshot 2021-02-03 at 10 50 10

### DIFF
--- a/components/admin_console/__snapshots__/schema_admin_settings.test.jsx.snap
+++ b/components/admin_console/__snapshots__/schema_admin_settings.test.jsx.snap
@@ -341,7 +341,7 @@ exports[`components/admin_console/SchemaAdminSettings should match snapshot with
                 },
                 Object {
                   "order": 11,
-                  "text": "Български",
+                  "text": "Български (Beta)",
                   "value": "bg",
                 },
                 Object {
@@ -457,7 +457,7 @@ exports[`components/admin_console/SchemaAdminSettings should match snapshot with
                 },
                 Object {
                   "order": 11,
-                  "text": "Български",
+                  "text": "Български (Beta)",
                   "value": "bg",
                 },
                 Object {
--- a/components/common/hocs/cloud/with_get_cloud_subcription.test.tsx
+++ b/components/common/hocs/cloud/with_get_cloud_subcription.test.tsx
@@ -26,6 +26,7 @@ describe('/components/common/hocs/with_get_cloud_subcription', () => {
                 isCloud={true}
                 actions={actions}
                 subscription={{}}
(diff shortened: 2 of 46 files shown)


nevyangelova (approved):
LGTM, let's see the Gitlab pipeline issue if it appears again

hmhealey (changes requested):
The extra space is still here when we can't use sticky headers, so we might need to apply another class or something when we identify that that's the case.
Screen Shot 2021-02-04 at 11 46 31 AM

Also, the extra space is used on the old sidebar as well (enabled by an admin going to Experimental > Features > Enable Legacy Sidebar in the System Console). I'll check with Eric and Matt to see if we care about that case though since it might not be worth fixing given the old sidebar is going to be completely removed in a few releases.

michelengelen:
@hmhealey I did add the requested changes.

Feel free to drop me a message, when you clarified the usage in Legacy Sidebar with Eric and/or Matt and I will adjust the code accordingly (if necessary)

matthewbirtch (changes requested):
Positioning looks good @michelengelen. However, I was able to get it display like the below screenshot if I reduced my browser's height:

image

Here's a gif capture as well:



This seems to happen if there are unreads below the visible viewport.

hmhealey:
Looks good to me on Firefox, but I have no idea why Matt's issue is occurring. I'm also seeing it on Chrome. Perhaps the changes to the nesting in that CSS file have caused issues with specificity of different rules, but I don't know why that would only be happening on Chrome then

michelengelen:
I know the cause and will fix it later. It is a bug with nesting styles in SCSS

michelengelen:
@matthewbirtch I did fix the issue with the indicator spanning the full height. Could you please re-check it? Thanks! ✨

matthewbirtch (approved):
Looks good to me!

michelengelen:
@hmhealey all is fixed now ... mind having another look, so we can move forward with it?

jgilliam17 (approved):
Thank you @michelengelen 
Tested, looks good to merge.
- Verified "More unreads' indicator is positioned below channel heading. Positioning of the indicator on the legacy sidebar, where it creates a small gap, appears to be an acceptable minor issue for now.

hmhealey (approved):

PR #7226, opened 2020-12-15
[MM-30337] - Fix mobile menus and remove redundant styles

nevyangelova:
#### Summary
This PR  solves styling issues on mobile menus. It removes some redundant styles for submenu icons that just caused a lot of CSS buildup and overtaking. NOTE: The dividers currently have spaces but this is fixed in the Category sorting PR https://github.com/mattermost/mattermost-webapp/pull/7068

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30337

#### Related PR's
https://github.com/mattermost/mattermost-webapp/pull/7068

PR #7462, opened 2021-02-03
MM-31888: adjusted padding on menu items in sidebar

michelengelen:
#### Summary
Adjustments for a change coming from unifiying styles in [PR #7226](https://github.com/mattermost/mattermost-webapp/pull/7226/files)
Padding was reset to a state before this PR was merged.

#### Ticket Link
Fixes [MM-3188](https://mattermost.atlassian.net/browse/MM-31888)

#### Screenshots
Screenshot 2021-02-03 at 17 39 24

### DIFF
--- a/components/sidebar/sidebar_menu/sidebar_menu.scss
+++ b/components/sidebar/sidebar_menu/sidebar_menu.scss
@@ -1,0 +1,8 @@
+// Manual override of the global menu (and item) styling for the sidebar dropdown menu
+#sidebarDropdownMenu .MenuItem {
+    > button,
+    > div,
+    > a {
+        padding: 0 32px 0 24px;
+    }
+}
--- a/components/sidebar/sidebar_menu/sidebar_menu.scss
+++ b/components/sidebar/sidebar_menu/sidebar_menu.scss
@@ -1,8 +1,0 @@
-// Manual override of the global menu (and item) styling for the sidebar dropdown menu
-#sidebarDropdownMenu .MenuItem {
-    > button,
-    > div,
-    > a {
-        padding: 0 32px 0 24px;
-    }
-}
--- a/components/common/hocs/cloud/with_get_cloud_subcription.test.tsx
+++ b/components/common/hocs/cloud/with_get_cloud_subcription.test.tsx
@@ -26,6 +26,7 @@ describe('/components/common/hocs/with_get_cloud_subcription', () => {
                 isCloud={true}
                 actions={actions}
                 subscription={{}}
+                userIsAdmin={true}
             />,
         );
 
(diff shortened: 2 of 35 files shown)


components/sidebar/sidebar_menu/sidebar_menu.scss
```diff
@@ -0,0 +1,8 @@
+// Manual override of the global menu (and item) styling for the sidebar dropdown menu
+#sidebarDropdownMenu .MenuItem {
+    > button,
+    > div,
+    > a {
```
nevyangelova, line 5: I would go against styling elements directly, since this is the reason why our styles are conflicting all over webapp, since the scope they are in has a lower specificity than the element styles. I know it can be annoying to add classes to every element but until we sort out the structure of our CSS, maybe use help from styled components we will run into issues with other elements in the future. We don't know where in the app a button, div or link might get updated with this :D

nevyangelova (changes requested):

components/sidebar/sidebar_menu/sidebar_menu.scss
michelengelen: I agree, but the HTML structure currently does not provide classes to the elements. 

Should we adjust the component in this case, so that they gain classes to better target them?

BTW: there should never be a override with this kind of selector since the id takes precedence over any class-based selector, but fair enough, you never know what happens in the future. 🕦 😃

nevyangelova: Oh I didn't notice its an ID, then I think we should be good!

nevyangelova (approved):

components/sidebar/sidebar_menu/sidebar_menu.scss
michelengelen: I just double-checked and the way the menu is rendered is pretty over-complicated, so adding a class to each of the items would mean to change at least 6-8 files (if not more). So I think we should be good for now with the id-selector! :D

matthewbirtch (approved):

jgilliam17 (approved):
Thank you @michelengelen 
Tested, looks good to merge.
- Verified left padding on the main menu.

devinbinnie (approved):
LGTM, but do we need an extra file now? I know it's the way going forward, but might be confusing to have the styles in 2 places. 

1/5 I think we should move the styles back into the `_legacy-sidebar-left.scss` file

hmhealey:
Maybe we could leave them in `_legacy-sidebar-left.scss` until we're ready to move the rest of them out? I do like the idea of splitting up CSS more into component-specific files, but I agree that having the CSS split might be problematic.

michelengelen:
Ok, seems like a good idea. I will move it over 📍

michelengelen:
@hmhealy good to merge now?

hmhealey (approved):
Yep, looks good to me

PR #7467, opened 2021-02-03
MM-32425 new data table component for data retention

BenCookie95:
#### Summary
This component already existed in billing history but we needed to re-use it. I pulled it out and made a few adjustments to allow for server side pagination and it is functionally similar to the data grid

See design here: https://www.figma.com/proto/PdZPkf36lNU6ZO41JlEjKd/Team-%26-Channel-Data-Retention?node-id=20%3A181&scaling=min-zoom

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32425

#### Screenshots
Screen Shot 2021-02-03 at 2 21 45 PM


components/admin_console/data_table/data_table.scss
```diff
@@ -0,0 +1,119 @@
...
+    line-height: 16px;
+    font-weight: 600;
+    border-bottom: 1px solid rgba(var(--sys-center-channel-color-rgb), 0.16);
+
+    &:last-child {
+        border-bottom: 1px solid rgba(var(--sys-center-channel-color-rgb), 0.08);
+    }
+}
+
+.Table__table-row {
```
devinbinnie, line 19: Kind of have some smurf naming going on here: https://devcards.io/smurf-naming-convention

components/admin_console/data_table/data_table.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+import DataTableHeader from './data_table_header';
+import DataTableRow from './data_table_row';
+
+import 'components/next_steps_view/next_steps_view.scss';
+
+import './data_table.scss';
+
+export type Column = {
+    name: string | JSX.Element;
```
devinbinnie, line 18: There's a type called `React.ReactNode` you can use here instead of `string | JSX.Element`. It covers a lot more bases of what can be rendered.

devinbinnie (approved):

hahmadia (commented):
I had one question regarding this component.

 We already have a common component called `DataGrid` which looks very similar to the screenshots you’ve posted and has a few extra things but they are optional.

 I think we can replace instances of data tables with data grid (with small modifications such as passing custom css class) to achieve the same thing without needing too create a whole new component. 

This way, we could help to avoid confusion between differentiating between the two components and when each one should be used.

You can take a look at the component here
https://github.com/mattermost/mattermost-webapp/blob/master/components/admin_console/data_grid/data_grid.tsx

And see it being used in the teams and channel page in the system console (System Console > User Management > Teams/Channels). The grid displaying the teams and channels is using this.

Anyhow, let me know what you think. These were just my thoughts after I realized what it was we were creating. Thanks!

BenCookie95:
> I had one question regarding this component.
> 
> We already have a common component called `DataGrid` which looks very similar to the screenshots you’ve posted and has extra things but they are optional (such as a search bar).
> 
> I think we can replace instances of data tables with data grid (with small modifications such as passing custom css class) to achieve the same thing without needing too create a whole new component.
> 
> This way, we could help to avoid confusion between differentiating between the two components and when each one should be used.
> 
> You can take a look at the component here
> https://github.com/mattermost/mattermost-webapp/blob/master/components/admin_console/data_grid/data_grid.tsx
> 
> And see it being used in the teams and channel page in the system console (System Console > User Management > Teams/Channels). The grid displaying the teams and channels is using this.
> 
> Anyhow, let me know what you think. These were just my thoughts after I realized what it was we were creating. Thanks!
> 
> **Edit** I believe you can also see an instance of the DataGrid component being used in the fashion of a Data Table in either the Users section of the Teams/Channels page in the system console. It just has this extra search functionality added.

I agree with everything you said, I've integrated it into the datagrid now. Looks a lot better and the updates were pretty minimal!

devinbinnie (approved):

hahmadia (commented):
Hey! I had a discussion with @devinbinnie and came to realization that the way we are handling the CSS here can be done in a more efficient manner. He will comment about the required changes to make this happen. Thanks!

components/admin_console/data_grid/data_grid.tsx
```diff
@@ -61,6 +63,10 @@ type Props = {
         keys: string[];
         onFilter: (options: FilterOptions) => void;
     };
+
+    customGridClass?: string;
```
devinbinnie, line 67: Thanks to @hahmadia for bringing this to my attention, I missed this in my first pass.

We can do this better by just providing one "custom class" to the component here, and letting CSS manage the styles based on the default class names already included.

So instead of having styles written like:
```
.MyCustomGridClass {
    display: block;
}

.MyCustomHeaderClass {
    padding: 20px;
    // some other styles
}
```

We can do this:
```
.MyCustomGridClass {
    display: block;
   
    .DataGrid_header {
        padding: 20px;
        // some other styles
    }
}
```

devinbinnie (changes requested):

components/admin_console/data_grid/data_grid.tsx
```diff
@@ -278,7 +281,7 @@ class DataGrid extends React.PureComponent<Props, State> {
     public render() {
         return (
             <div
-                className='DataGrid'
+                className={classNames('DataGrid', this.props.customGridClass)}
```
devinbinnie, line 284: Nit: prefer the use of `className` for the name here over `customClass`. Might just be a personal thing, but it lines up with the attribute that other HTML elements use. 0/5 on the change though.

devinbinnie (approved):

hahmadia (approved):
LGTM. Agree with Devin NIT. Thanks!

furqanmlk (approved):
image

PR #7469, opened 2021-02-03
[MM-32429] Content Cards

BenCookie95:
#### Summary
Moving Card component out of billing history because we need to reuse it for data retention. Best way to see it is by running storybook `npm run storybook`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32429

#### Screenshots
Screen Shot 2021-02-03 at 3 28 39 PM


components/admin_console/card/card.scss
```diff
@@ -0,0 +1,45 @@
...
+    border: 1px solid rgba(var(--sys-center-channel-color-rgb), 0.08);
+    box-shadow: 0px 2px 3px rgba(0, 0, 0, 0.08);
+    border-radius: 4px;
+    color: var(--sys-center-channel-color);
+    width: 100%;
+    height: '463px';
+    margin-bottom: 20px;
+}
+
+.Card__cardHeader {
```
devinbinnie, line 12: Your naming style seems to follow smurf naming, which is an anti-pattern. https://devcards.io/smurf-naming-convention

We should be able to avoid this if we just override the base styles as I mentioned above.

devinbinnie (changes requested):
The code looks good, but I'm not seeing many differences between the `components/card` component and this one. I think we should really be using one component here, adding all the functionality we need (unless its conflicting) and then adding extra classes to do any overrides necessary.

Let me know if you want to chat about this.

hahmadia (changes requested):
I’m going to have to agree with @devinbinnie on this.
Although the code looks good, we already have a `components/Card` which I think would be better to modify the needs that this component is trying to satisfy.
I also want to add that since it’s dealing with UX/UI, we should add Michael on it and have him verify that the styling matches expectations.
This requiring UX/UI came from a meeting we had a while back that we all agreed on if there is any sort of UX/UI (new components or css changes or etc), we should include him on the PR.

components/admin_console/card/card.scss
```diff
@@ -0,0 +1,45 @@
+.Card__card {
+    background-color: var(--sys-center-channel-bg);
```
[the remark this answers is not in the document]
BenCookie95: I've moved the changes we need into the existing card component. Added a new class called console that will add the styles I need and the card_header has an override for our title/subtitle/button layout

components/card/card_header.tsx
```diff
@@ -7,13 +7,44 @@ import classNames from 'classnames';
...
+    onClick?: () => void;
 };
 
 const CardHeader: React.FC<Props> = (props: Props) => {
     return (
         <div className={classNames('Card__header', {expanded: props.expanded})}>
-            {props.children}
-            {props.expanded && <hr className='Card__hr'/>}
+            {
+                props.title && props.subtitle ?
```
devinbinnie, line 20: This feels a bit weird to me, architecturally. The aim of the `CardHeader` component is to not be prescriptive about what goes in the header, letting the consumer dictate the direction of what's inside. The way your changes are setup looks to short circuit that by giving the user an alternate method of filling the header that the component provides.

I'm thinking here that we should simply provide either an alternate version of `CardHeader` that only takes the title and subtitle as parameters and generates the HTML you've added, OR we create a component that you can pass into `CardHeader` as a child that does the same.

devinbinnie (changes requested):
A couple more comments on this one. Let me know if you'd like to chat about this further.

components/card/card_header.tsx
```diff
@@ -7,13 +7,44 @@ import classNames from 'classnames';
 type Props = {
     children: React.ReactNode;
```
hahmadia, line 8: I think this should become an optional prop now since it isn't required if we are passing title, subtitle, buttonText.

components/card/card_header.tsx
hahmadia, line 20: I dont know why but this feels odd to me. If this component is already accepting children, meaning we can pass in anything to fill it up with content, why do we provide this alternate.

What about if we make a functional component that accepts these params and can be passed into CardHeader? 
It can look something like
```
<Card.header>
       <CardHeaderContent
             title=fsa
             subtitle=fdsaf
             onClick={void}
       />
</Card.header>
```
I would also say make the title required, subtitle optional, and onClick optional too incase we want a bare minimum of just a title in the card header.

My thoughts though .... Let me know what you guys think.

hahmadia (changes requested):
Looks better but still have some questions

BenCookie95:
@devinbinnie @hahmadia I think the best approach here would be for me to create a new card header component inside here that accepts title, subtitle, buttonText and onClick. So that way you can choose to use the original header or the new one with this predefined stuff. Is this ok with both of you?

devinbinnie:
> @devinbinnie @hahmadia I think the best approach here would be for me to create a new card header component that accepts title, subtitle, buttonText and onClick. So that way you can choose to use the original header or the new one with this predefined stuff. Is this ok with both of you?

@BenCookie95 Either that or the suggestion that @hahmadia works for me :)

hahmadia:
> @devinbinnie @hahmadia I think the best approach here would be for me to create a new card header component that accepts title, subtitle, buttonText and onClick. So that way you can choose to use the original header or the new one with this predefined stuff. Is this ok with both of you?

I like this approach too but I'd lean more towards a content component thats passed into the header rather than another component that has the same name only because it doesn't feel right to overload a component name or have multiple different variations of the same component. Thanks!

components/card/card.test.tsx
```diff
@@ -11,6 +12,23 @@ describe('components/card/card', () => {
         expanded: false,
     };
 
+    const headerProps = {
+        title:
+    <FormattedMessage
```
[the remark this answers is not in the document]
BenCookie95: I tried fixing the indent here but lint fails. This is where lint wants it for some reason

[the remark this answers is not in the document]
devinbinnie: We should look into that then. For now it's fine, thanks for checking.

components/card/card.tsx
```diff
@@ -16,6 +17,7 @@ type Props = {
 
 export default class Card extends React.PureComponent<Props> {
     public static Header = CardHeader;
+    public static HeaderContent = CardHeaderContent;
```
devinbinnie, line 20: 0/5, but not sure if this needs to be a member of the card if it's only used by the admin console. It can be a separate component.

components/card/card_header_content.tsx
```diff
@@ -0,0 +1,39 @@
...
+import React from 'react';
+
+type Props = {
+    title: React.ReactNode;
+    subtitle?: React.ReactNode;
+    buttonText?: React.ReactNode;
+    onClick?: () => void;
+};
+
+const CardHeaderContent: React.FC<Props> = (props: Props) => {
```
devinbinnie, line 13: I would come up with a different name than `CardHeaderContent` here.
Maybe some like `AdminCardHeader` or `TitleAndButtonCardHeader`?

devinbinnie (approved):

components/card/card.tsx
hahmadia: @devinbinnie 
I actually recommended him this only because I got inspiration from how the `Menu` component is done
image
I would still say keep it this way only because if it's a totally separate component, we will probably be passing it as a child of Card.Header almost always anyways so might as well be a part of Card. Although currently we only use it in admin console, it would be easier to extend outside if ever needed.

components/card/card_header_content.tsx
hahmadia: Agreed. I came up with that name off top off my head but I believe Devin's `TitleAndButtonCardHeader` make more sense.

hahmadia (changes requested):
Sorry! Two last changes.

1. Call the component `TitleAndButtonCardHeader`.
2. Make it it's own component and not a child of Card. 

Thanks!

hahmadia (approved):
LGTM!

devinbinnie (approved):

furqanmlk (approved):
image

PR #7470, opened 2021-02-03
Cypress/E2E: Split bot accounts in lists spec

josephbaylon:
#### Summary
- Separated the 2 cases because for some odd reason the first one is adversely affecting the other (but it's not obvious as to why); separated for now so that they pass on cypress run.

#### Ticket Link
NA


prapti (approved):
Looks good. All tests passed.

furqanmlk (approved):
Both Tests passed ✅

PR #7476, opened 2021-02-04
added `channel.name` to the concatenated searchString

michelengelen:
#### Summary
added `channel.name` to the concatenated searchString to make searching for the channel name (as written in the URL) possible as well.

Screenshot shows how searching for `town-square` is correctly showing `Town Square` as an option

#### Ticket Link
Fixes [MM-25609](https://mattermost.atlassian.net/browse/MM-25609)

#### Screenshots
before|after
--|--
Screenshot 2021-02-04 at 16 03 51|Screenshot 2021-02-04 at 14 21 11

### DIFF
--- a/components/suggestion/switch_channel_provider.jsx
+++ b/components/suggestion/switch_channel_provider.jsx
@@ -249,7 +249,7 @@ function makeChannelSearchFilter(channelPrefix) {
     const userSearchStrings = {};
 
     return (channel) => {
-        let searchString = channel.display_name;
+        let searchString = `${channel.display_name}${channel.name}`;
         if (channel.type === Constants.GM_CHANNEL || channel.type === Constants.DM_CHANNEL) {
             const usersInChannel = usersInChannels[channel.id] || new Set([]);
 


devinbinnie (approved):

nevyangelova (approved):

matthewbirtch:
Looks good to me. The only question I have is if we should show the URL here in the same way we do for channel autocomplete:

image

We could likely do this in a different PR though, let me know what you think @michelengelen

michelengelen:
@matthewbirtch I can totally relate to your intention. The problem is, if there is a channel (or more) that have a totally completely different display-name than the name it can be misleading. If you look at your screenshot for example: searching for guidelines would show the channel `Brand Experience` as an option, which does not make sense without knowing that the name is actually `brand-guidelines`.

I suppose we should do this in a separate PR, do you agree?

jgilliam17 (approved):
Thank you @michelengelen
Tested, looks good to merge.
- Verified channel switcher search now uses both, display name and channel URL name, to filter results.

matthewbirtch:
> I suppose we should do this in a separate PR, do you agree?

Yep, totally good with me to do that in another PR.

matthewbirtch (approved):

PR #7479, opened 2021-02-04
MM-29726 Allow disabling link previews from certain domains

berkeka:
#### Summary

Added a UI field for restricted domains under Site Configuration > Posts in the System Console and created english localizations for it.

#### Ticket Link

https://github.com/mattermost/mattermost-server/issues/16846

#### Related Pull Requests

- Has server changes (https://github.com/mattermost/mattermost-server/pull/16869)

#### Screenshots

Screen Shot 2021-02-04 at 20 37 05


esethna:
Awesome @berkeka! Nice work!

i18n/en.json
```diff
@@ -554,6 +554,9 @@
   "admin.customization.gfycatApiSecretDescription": "The API secret generated by Gfycat for your API key. When blank, uses the default API secret provided by Gfycat.",
   "admin.customization.iosAppDownloadLinkDesc": "Add a link to download the iOS app. Users who access the site on a mobile web browser will be prompted with a page giving them the option to download the app. Leave this field blank to prevent the page from appearing.",
   "admin.customization.iosAppDownloadLinkTitle": "iOS App Download Link:",
+  "admin.customization.restrictLinkPreviewsDesc": "Link previews and image link previews will not be shown for the above list of comma-separated domains.",
+  "admin.customization.restrictLinkPreviewsExample": "E.g.: \"corp.mattermost.com, mattermost.org\"",
```
hmhealey, line 558: @esethna Any suggestions for what we might be able to use as example text? We probably don't want to suggest people block us 😛

hmhealey:
I've blocked this on the accompanying server PR since that needs to go in first.

components/admin_console/admin_definition.jsx
```diff
@@ -2368,6 +2368,17 @@ const AdminDefinition = {
...
+                    {
+                        type: Constants.SettingsTypes.TYPE_TEXT,
+                        key: 'ServiceSettings.RestrictLinkPreviews',
+                        label: t('admin.customization.restrictLinkPreviewsTitle'),
+                        label_default: 'Disable link previews from these domains:',
+                        help_text: t('admin.customization.restrictLinkPreviewsDesc'),
+                        help_text_default: 'Link previews and image link previews will not be shown for the above list of comma-separated domains.',
+                        placeholder: t('admin.customization.restrictLinkPreviewsExample'),
+                        placeholder_default: 'E.g.: "corp.mattermost.com, mattermost.org"',
+                        isDisabled: it.not(it.userHasWritePermissionOnResource('site')),
```
Willyfrog, line 2380: @esethna  should this also be disabled if `Enable Link Previews` is disabled? I've looked into the ticket but it is not specified there, but it might make some sense to have that dependency

i18n/en.json
Willyfrog: another possible example: `internal.mycompany.com`

Willyfrog (commented):
agree with @hmhealey that we should use other domains, also added a PM question. other than that it looks good. Thanks @berkeka

i18n/en.json
hmhealey: That might be a good idea, although we prevent link previews from being generated for local addresses by default, so I wouldn't want to necessarily suggest that we don't do that

components/admin_console/admin_definition.jsx
esethna: @Willyfrog @berkeka ya good point. It should be in a disabled state (greyed out), if link previews are disabled

berkeka: @Willyfrog @esethna Implemented it with this 09e91e3 commit.

esethna (approved):

i18n/en.json
esethna: @berkeka can we update the placeholder text to: `internal.mycompany.com, images.example.com`

Willyfrog (approved):

hmhealey (approved):

hmhealey:
@jgilliam17 This requires the server changes to be tested, so let us know if those aren't picked up by the test server. You'll know that they're not picked up since the settings in the system console won't be saved

jgilliam17:
@hmhealey Setting in the system console for "disable link previews" is not being saved. Maybe I misunderstood, and I should probably have to wait for server work to be done before testing. Let me know. Thanks.

hmhealey:
@jgilliam17 I've tried setting up a test server from the server PR (https://github.com/mattermost/mattermost-server/pull/16869). Hopefully that should pull these changes into it so they can be tested together

hmhealey:
Okay, the test server is using a week old version of the code, probably corresponding to when the server branch was last pushed. I'll resolve the merge conflict and push a new version of both branches in a bit to try fixing the issue

jgilliam17 (approved):
Thanks @berkeka 
Tested, looks good to merge.
- Verified setting for disabling previews from certain domains. Setting saves and applies to all existing links on refresh.

hmhealey:
Thanks, Jelena! Nice to finally get this merged now that the test server has decided to behave.

Thanks again for the PR, @berkeka!

PR #7488, opened 2021-02-05
[MM-20378] Convert file_attachment_list and tests to TypeScript

LauSam09:
#### Summary
This pull request converts the `file_attachment_list` component and tests to TypeScript.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16729
JIRA: https://mattermost.atlassian.net/browse/MM-20378


components/file_attachment_list/file_attachment_list.test.tsx
```diff
@@ -4,17 +4,57 @@
...
+import {Post} from 'mattermost-redux/src/types/posts';
+
 import FileAttachment from 'components/file_attachment';
 import SingleImageView from 'components/single_image_view';
 
 import FileAttachmentList from './file_attachment_list';
 
 describe('FileAttachmentList', () => {
-    const post = {id: 'post_id', file_ids: ['file_id_1', 'file_id_2', 'file_id_3']};
+    const post: Post = {
```
devinbinnie, line 15: For future reference, we have a `getPostMock` function in `utils/test_helper` that mocks out all the extra data you need here.

components/file_attachment_list/index.ts
```diff
@@ -6,15 +6,16 @@ import {connect} from 'react-redux';
...
 import {isEmbedVisible} from 'selectors/posts';
 
-import FileAttachmentList from './file_attachment_list.jsx';
+import FileAttachmentList, {Props} from './file_attachment_list';
 
 function makeMapStateToProps() {
     const selectFilesForPost = makeGetFilesForPost();
 
-    return function mapStateToProps(state, ownProps) {
+    return function mapStateToProps(state: GlobalState, ownProps: Props) {
```
devinbinnie, line 18: It's generally not a good practice to use the same `Props` type that the component uses for the redux connector.
What will happen here is that any component that consumes this one will expect all props that are required for the component (eg. `locale`) to be explicitly provided by the consuming component, causing some annoying type errors.

I'd recommend here creating an `OwnProps` type specifically for what the `mapStateToProps` function needs to fetch other data and using that type instead.

devinbinnie (changes requested):
Thanks for the contribution @LauSam09!
Looks good overall, just had a couple small comments to resolve.

components/file_attachment_list/index.ts
LauSam09: Thanks for the explanation - I've made the change you suggested.

components/file_attachment_list/file_attachment_list.test.tsx
LauSam09: Handy - thanks!

components/file_attachment_list/file_attachment_list.test.tsx
```diff
@@ -32,7 +37,21 @@ describe('FileAttachmentList', () => {
             ...baseProps,
             fileCount: 1,
             fileInfos: [
-                {id: 'file_id_1', name: 'file.txt', extension: 'txt'},
+                {
```
devinbinnie, line 40: Could use `getFileInfoMock` here too if you wanted

devinbinnie (approved):
Thanks @LauSam09, this looks great now! Just one non-blocking comment.

components/file_attachment_list/file_attachment_list.test.tsx
LauSam09: Oops - done!

hmhealey (approved):
Thanks, @LauSam09. Looks good to me!

jgilliam17 (approved):
Thank you @LauSam09 
Tested, looks good to merge.
- Verified file attachments and previews - as before.

PR #7489, opened 2021-02-07
MM-27241 Private channel prompt for sysadmin

anurag6713:
#### Summary
Error message or prompt to continue before force-joining private channel. The permission to join is not the same as the intent to join. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27241

#### Checklist
- [x] Prompt on clicking the channel url or permalink
- [x] Prompt on directly accessing the channel url or permalink from address bar
- [x] Should leave the team If user joins the team automatically while trying to open permalink from address bar

#### Screenshots
Screenshot 2021-02-07 at 17 10 35


josephbaylon (changes requested):
Thanks @anurag6713 !! Just minor changes. I also created a new Zephyr ticket to track the e2e: [MM-T3830](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T3830)

josephbaylon (approved):
LGTM. Test passed locally. Thanks @anurag6713 !!

components/join_private_channel_modal.tsx
```diff
@@ -0,0 +1,83 @@
...
+    onCancel: () => void;
+    onHide: () => void;
+    onJoin: () => void;
+}
+
+type State = {
+    show: boolean;
+}
+
+export default class JoinPrivateChannelModal extends React.PureComponent<Props, State> {
```
BenCookie95, line 20: Should we make this into a functional component? It doesn't have any lifecycle functions and only takes a few props

utils/utils.jsx
```diff
@@ -1776,6 +1805,68 @@ export function handleFormattedTextClick(e, currentRelativeTeamUrl) {
         if (!(e.button === MIDDLE_MOUSE_BUTTON || e.altKey || e.ctrlKey || e.metaKey || e.shiftKey)) {
             e.preventDefault();
 
+            const state = store.getState();
+            const user = getCurrentUser(state);
+            if (isSystemAdmin(user.roles)) {
+                const match = isChannelOrPermalink(linkAttribute.value);
+                if (match) {
```
BenCookie95, line 1812: There are a lot of nested if statements inside here and I wonder if we could reduce how many we have. It could lead to complications in the future when someone has to make updates. We can remove the first if statement and do this?:
```suggestion
                if (isSystemAdmin(user.roles) && match) {
```

Anywhere we have something like this:
```
const {data: channelData} = await store.dispatch(getChannelAction(post.channel_id));
if (channelData) {
      channel = channelData;
}
```

We could change to this:
```
const {data: channelData} = await store.dispatch(getChannelAction(post.channel_id));
channel = channelData || channel;
```

These are just suggestions but it might make it more readable. There are probably other places we can clean up in this block too

BenCookie95 (changes requested):
Looks good overall, just a few suggestions that I had!

components/join_private_channel_modal.tsx
anurag6713: Yes .. I wasn't thinking about that. Good one!

utils/utils.jsx
anurag6713: Agreed and I refactored code a bit now.

BenCookie95: Nice, thank you!

BenCookie95 (approved):
Code looks good to me!

pablovelezvidal (approved):
Great stuff @anurag6713! Just one minor non-blocking question.

components/channel_layout/channel_identifier_router/actions.ts
```diff
@@ -155,15 +156,33 @@ export function goToChannelByChannelName(match: Match, history: History): Action
...
             const joinChannelDispatchResult = await dispatch(joinChannel(getCurrentUserId(state), teamObj!.id, '', channelName));
             if ('error' in joinChannelDispatchResult) {
-                const getChannelDispatchResult = await dispatch(getChannelByNameAndTeamName(team, channelName, true));
-                if ('error' in getChannelDispatchResult || getChannelDispatchResult.data.delete_at === 0) {
-                    await dispatch(fetchMyChannelsAndMembers(teamObj!.id));
-                    handleChannelJoinError(match, history, getRedirectChannelNameForTeam(state, teamObj!.id));
-                    return {data: undefined};
+                if (!channel) {
+                    const getChannelDispatchResult = await dispatch(getChannelByNameAndTeamName(team, channelName, true));
+                    if ('error' in getChannelDispatchResult || getChannelDispatchResult.data.delete_at === 0) {
```
pablovelezvidal, line 179: hi! I have a question, is there a reason why you decide to use the -in operator- instead of, in the previous line, destructure the coming values as `const {data, error} = await dispatch(getChannelByName...` ?

anurag6713: Yes, I started like that but was getting a ts error. Turns out ActionResult type was like `{data: any} | {error: any}` and doing `{data, error}` makes one type def present and the other absent. So I looked around and saw other code using `in` and I did the same 😅

nevyangelova (approved):
LGTM!

furqanmlk (approved):
image

image

PR #7493, opened 2021-02-08
MM-31916: channel mention for specific keywords breaks application

michelengelen:
#### Summary
provided a more sophisticated check for certain keywords in the channelNamesMap to check for the type being an object.
The application broke at this point because the value searched for is non-existent and therefore could not extract the `display_name` property, which is why the application was failing at this point. The change that fixes the bug is to be found at line 402

On the way I fixed some warnings and errors in the file

#### Ticket Link
Fixes [MM-31916](https://mattermost.atlassian.net/browse/MM-31916)

### DIFF
--- a/utils/text_formatting.tsx
+++ b/utils/text_formatting.tsx
@@ -200,8 +200,8 @@ export function formatText(
     text: string,
     inputOptions: TextFormattingOptions = DEFAULT_OPTIONS,
     emojiMap: EmojiMap,
-) {
-    if (!text || typeof text !== 'string') {
+): string {
+    if (!text) {
         return '';
     }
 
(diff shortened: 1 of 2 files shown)


nevyangelova (approved):

matthewbirtch:
@michelengelen this seems like more of a technical fix based on the description. Does it need a ux review? If so, what specifically should I look at?

michelengelen:
@matthewbirtch not really ... It is just to check if it renders correctly (how it is supposed to be)

srkgupta (approved):
Thanks @michelengelen. Tested and the fix is working fine now. Looks good from security perspective.

hmhealey (approved):

jgilliam17 (approved):
Thank you @michelengelen 
Tested, looks good to merge.
- Verified specific text, as mentioned, no longer breaks the application.

PR #7468, opened 2021-02-03
MM-31503 - screen to restrict access because workspace capacity"

pablovelezvidal:
#### Summary
Create a new screen for users who are restricted from joining a workspace because the free capacity is exceeded

#### Ticket Link
[MM-31503](https://mattermost.atlassian.net/browse/MM-31503)

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1351

#### Screenshots
maxFreeUsers

PR #7496, opened 2021-02-08
MM-3190 show user limit modal when cloud limit exceed

pablovelezvidal:
#### Summary
Shows the limit user modal to admin when they click the invite users in the welcome text channel and there are no more free seats.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31905

#### Screenshots
upgradeModal4inviteusers


marianunez (changes requested):
Given that we may be adding entry points to the Invite Users screens in other places in the near future, does it make sense to move the logic of checking for the limit and showing the upgrade modal within the actual invite modal? This way we avoid duplicating this logic whenever we need to open the Invite People screen

stevemudie (approved):
Looks good!

nickmisasi:
> Given that we may be adding entry points to the Invite Users screens in other places in the near future, does it make sense to move the logic of checking for the limit and showing the upgrade modal within the actual invite modal? This way we avoid duplicating this logic whenever we need to open the Invite People screen

+1

pablovelezvidal:
> > Given that we may be adding entry points to the Invite Users screens in other places in the near future, does it make sense to move the logic of checking for the limit and showing the upgrade modal within the actual invite modal? This way we avoid duplicating this logic whenever we need to open the Invite People screen
> 
> +1

yes, that was the latest update I did to the PR. @nickmisasi , @marianunez could you please take a look again?

components/invitation_modal/index.js
```diff
@@ -50,16 +51,19 @@ export function mapStateToProps(state) {
     const isLicensed = license && license.IsLicensed === 'true';
     const isGroupConstrained = Boolean(currentTeam.group_constrained);
     const canInviteGuests = !isGroupConstrained && isLicensed && guestAccountsEnabled && haveITeamPermission(state, {team: currentTeam.id, permission: Permissions.INVITE_GUEST});
+    const isCloud = license.Cloud === 'true';
+    const freeTierWithNoFreeSeats = isCloud && subscriptionStats && subscriptionStats.is_paid_tier === 'false' && subscriptionStats.remaining_seats <= 0;
```
nickmisasi, line 55: Is it possible subscriptionStats might be empty here? true && {} (if subscriptionStats is empty) would mean that the next case (subscriptionStats.is_paid_tier) access would throw an error?

components/invitation_modal/invitation_modal.jsx
```diff
@@ -231,90 +234,104 @@ export default class InvitationModal extends React.PureComponent {
...
+                    }
+                </div>
+            </FullScreenModal>
+        </RootPortal>);
+
+        const actionsUserLimitModal = {
+            closeModal: this.props.actions.closeModal,
+            openModal: this.props.actions.openModal,
+        };
+        const userLimitModal = (<>
```
nickmisasi, line 325: Do you need the <> here?

components/invitation_modal/index.js
```diff
@@ -50,16 +51,19 @@ export function mapStateToProps(state) {
...
 
     const canAddUsers = haveITeamPermission(state, {team: currentTeam.id, permission: Permissions.ADD_USER_TO_TEAM});
     return {
         invitableChannels,
         currentTeam,
         canInviteGuests,
         canAddUsers,
+        isFreeTierWithNoFreeSeats: freeTierWithNoFreeSeats,
```
nickmisasi, line 63:
```suggestion
    const isFreeTierWithNoFreeSeats = isCloud && subscriptionStats && subscriptionStats.is_paid_tier === 'false' && subscriptionStats.remaining_seats <= 0;

    const canAddUsers = haveITeamPermission(state, {team: currentTeam.id, permission: Permissions.ADD_USER_TO_TEAM});
    return {
        invitableChannels,
        currentTeam,
        canInviteGuests,
        canAddUsers,
        isFreeTierWithNoFreeSeats,
```

nickmisasi: nit

nickmisasi (changes requested):

components/invitation_modal/index.js
pablovelezvidal: thanks @nickmisasi , good catch. I will validate using the isEmpty lodash function.

components/invitation_modal/invitation_modal.jsx
pablovelezvidal: Not necessary. Removing it to get cleaner code.

marianunez (approved):
Great work refactoring the `UserLimitModal` directly into the `InvitationModal`! This will reduce code duplication moving forward. 

We should remove any current instances where we currently open the `UserLimitModal`. 0/5 if is in this or another PR

stevemudie (approved):
Tested and Passed

The link to invite users at the bottom of the Town Square channel opens the 'You've reached the user limit' modal as expected. The modal also worked and is dismissed in all fashions as you would expect.

I did notice that on first use the share links appear for a flash before going away and the modal showing up. It only happens on the first click after opening the Spinwick but after speaking with @pablovelezvidal about it he's aware and he assured me that it won't be a problem once merged so this one is good to go!

pablovelezvidal:
> Tested and Passed
> 
> The link to invite users at the bottom of the Town Square channel opens the 'You've reached the user limit' modal as expected. The modal also worked and is dismissed in all fashions as you would expect.
> 
> I did notice that on first use the share links appear for a flash before going away and the modal showing up. It only happens on the first click after opening the Spinwick but after speaking with @pablovelezvidal about it he's aware and he assured me that it won't be a problem once merged so this one is good to go!

To clarify what @stevemudie mentions, the PR that contains the necessary code so the flash don't happen is this one https://github.com/mattermost/mattermost-webapp/pull/7468, there is the code where we load the subscriptionStats on app initial boot.

nickmisasi (approved):

PR #7499, opened 2021-02-09
[MM-31502] - Add Notify Admin link to invitation modal

AGMETEOR:
#### Summary
This PR adds a link where users can notify admin about free tier over limit

#### Ticket Link
[MM-31502](https://mattermost.atlassian.net/secure/RapidBoard.jspa?rapidView=79&projectKey=MM&modal=detail&selectedIssue=MM-31502&quickFilter=382)

#### Related Pull Requests
[Server](https://github.com/mattermost/mattermost-server/pull/16882)


nickmisasi (approved):
LGTM! I think you need to run `make i18n-extract` in the mattermost-webapp repo to fix your pipeline issues.

components/widgets/inputs/users_emails_input.jsx
```diff
@@ -323,6 +366,7 @@ export default class UsersEmailsInput extends React.PureComponent {
                                 </components.NoOptionsMessage>
                             )}
                         </FormattedMarkdownMessage>
+                        <NotifyButton/>
```
marianunez, line 369: What is the condition to show this new `NotifyButton`? Given that this is a shared component we don't want to show this for all but only when end-users have reached the limit of invites in cloud. 

I believe you can leverage on @pablovelezvidal work to add the `Upgrade Now` link that was added in a generic way to this component to avoid adding cloud specific logic here: https://github.com/mattermost/mattermost-webapp/pull/7446

marianunez (changes requested):

components/widgets/inputs/users_emails_input.jsx
```diff
@@ -25,6 +26,48 @@ import {isGuest} from 'utils/utils';
 
 import './users_emails_input.scss';
 
+const NotifyButton = () => {
```
pablovelezvidal, line 29: Hi @AGMETEOR! great work! just a simple comment, wouldn't it make sense to extract this button to it's own file?, one reason I see is that users_emails_input is a common used component that in not all contexts the notifyButton makes sense to be just here. Other reason I see for extract notifyButton is that perhaps in the future we could use it from other's sections of the webapp.

components/widgets/inputs/users_emails_input.jsx
```diff
@@ -25,6 +26,48 @@ import {isGuest} from 'utils/utils';
...
+    const [notifyStatus, setStatus] = useState(NOT_STARTED);
+    const {formatMessage} = useIntl();
+
+    const notifyFunc = async () => {
+        try {
+            setStatus(STARTED);
+            await Client4.sendOverLimitAlert();
+            setStatus(SUCCESS);
+        } catch (error) {
+            if (error) {
```
pablovelezvidal, line 44: I think this is a unnecessary if condition. You don't do any special validation, just to make sure there is an error, which I think is implicit if you are in the catch block.

components/widgets/inputs/users_emails_input.jsx
pablovelezvidal: @marianunez is right, I already added the same logic and instead of hardcoding the element here, I send it as a new prop, so it is customizable from the component that calls the user_emails_input.

AGMETEOR: @marianunez and @pablovelezvidal I think it's best if I wait for #7446 to merge and then build on top of it, since Pablo already has a good way to add these links (Upgrade and Notify the Admin) in that PR

marianunez: @AGMETEOR Pablos PR has been merged and should unblock you here

AGMETEOR: @marianunez and @pablovelezvidal closing in favor of #7546

PR #7504, opened 2021-02-09
[MM-32684] Fix RHS Autocomplete options cover the input box for some reply threads

hahmadia:
#### Summary

The slash command autocomplete doesn't hover your text and behaves the same as `@`

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-32684

#### Screenshots

image


components/suggestion/suggestion_box.jsx
```diff
@@ -656,7 +656,7 @@ export default class SuggestionBox extends React.PureComponent {
             if (handled) {
                 if (this.state.suggestionBoxAlgn.pixelsToMoveX === undefined &&
                     this.state.suggestionBoxAlgn.pixelsToMoveY === undefined &&
-                    provider.triggerCharacter && ['@', ':', '~'].indexOf(provider.triggerCharacter) !== -1) {
+                    ['@', ':', '~', '/'].includes(provider.triggerCharacter)) {
```
pablovelezvidal, line 659: hi @hahmadia , the includes function has no support for IE and I think babel doesn't transpile it either. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#browser_compatibility

pablovelezvidal (approved):
Great stuff @hahmadia . Just a minor question about the includes prototype vs indexOf.

components/suggestion/suggestion_box.jsx
hahmadia: Hey @pablovelezvidal !

Thanks for the point out but I think we should be okay. Let me know otherwise. See https://forum.mattermost.org/t/mattermost-is-dropping-support-for-internet-explorer-ie11-in-v5-16/7575

iomodo (approved):
Awesome!

components/suggestion/suggestion_box.jsx
pablovelezvidal: ok, perfect, I was not entirely sure of the MM support to IE, then we are fine by using Array includes prototype.

hmhealey: One of the best things about working on Mattermost is that it doesn't support IE any more 😄

DHaussermann (approved):
Tested and passed
- Slash command autocompleted no longer covers 1st line of input box
- Behavior now seems consistent with previous releases where the options show above the input when there is space. When there is not enough space, the options appear below the cursor which would be on the top line of the input box
- Tested on long and short reply threads
- Tested on browser and desktop with both full and partial resolution   
- Options never cover the text I am typing. 
- Regression tested main input box
- Regression tested user and channel auto-complete
- Regression tested mobile app
- No issues found

Thanks @hahmadia LGTM!

No new release tests needed to cover this.

PR #7512, opened 2021-02-10
MM-28736 Update dependencies

hmhealey:
This is most of the remaining dependency updates minus React and a few things that I don't know enough about to update myself. I'll be submitting a followup PR for Reach and filing some more tickets for the other things.

I'll comment in the changes about anything particularly noteworthy.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28736

#### Related Pull Requests
https://github.com/mattermost/mattermost-redux/pull/1355


Makefile
```diff
@@ -92,10 +92,7 @@ stop: ## Stops webpack
 ifeq ($(OS),Windows_NT)
 	wmic process where "Caption='node.exe' and CommandLine like '%webpack%'" call terminate
 else
-	@for PROCID in $$(ps -ef | grep "[n]ode.*[w]ebpack" | awk '{ print $$2 }'); do \
-		echo stopping webpack watch $$PROCID; \
-		kill $$PROCID; \
-	done
+	@pkill -fl webpack || true
```
hmhealey, line 95: The new version of webpack shows up as `webpack` when you run `ps -ef` instead of `node /some/path/to/webpack`, so the old code didn't work. Luckily, MacOS has come with `pkill` for a couple years now, so we can finally use that everywhere.

The `|| true` is needed since `pkill` returns an error code when no processes are found to kill.

actions/cloud.tsx
```diff
@@ -1,6 +1,6 @@
-
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+
```
hmhealey, line 3: The old version of `eslint-plugin-header` accepted whitespace above the header. The new one does not.

components/admin_console/admin_settings.tsx
```diff
@@ -28,8 +28,6 @@ export type BaseState = {
     errorTooltip: boolean;
 }
 
-type StateKeys = keyof BaseState;
```
hmhealey, line 31: The previous way we had `typescript-eslint` configured was less picky about unused types. You'll also see this in a few places where I removed type parameters like in the Cypress type defintions.

components/admin_console/generated_setting.tsx
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import crypto from 'crypto';
+import crypto from 'crypto-browserify';
```
hmhealey, line 4: Previous versions of `webpack` automatically bundled some Node libraries like `crypto`, but now we have to do it ourselves.

components/admin_console/group_settings/groups_list/groups_list.tsx
```diff
@@ -268,8 +268,6 @@ export default class GroupsList extends React.PureComponent<Props, State> {
         let {searchString} = this.state;
 
         const newState = {...this.state};
-        delete newState.page;
```
hmhealey, line 271: This isn't actually changing anything. When we deleted these items from `newState` before before passing it into `this.setState`, we were just telling React to leave them as-is which is what we're now still doing.

components/more_direct_channels/index.ts
```diff
@@ -99,7 +99,7 @@ const filterGroupChannels = memoizeResult((channels: Array<{profiles: UserProfil
     });
 });
 
-const filterDirectChannels = memoizeResult((channels: Channel[], userId: string) => {
+const filterDirectChannels = memoizeResult((channels: Record<string, Channel>, userId: string) => {
```
hmhealey, line 102: This type was incorrect before, but it didn't matter until I updated the dependencies in mattermost-redux

package.json
```diff
@@ -120,61 +121,57 @@
...
-    "eslint-plugin-header": "3.0.0",
-    "eslint-plugin-import": "2.22.0",
+    "eslint-plugin-cypress": "2.11.2",
+    "eslint-plugin-header": "3.1.0",
+    "eslint-plugin-import": "2.22.1",
     "eslint-plugin-jquery": "1.5.1",
     "eslint-plugin-mattermost": "github:mattermost/eslint-plugin-mattermost#46ad99355644a719bf32082f472048f526605181",
     "eslint-plugin-no-only-tests": "2.4.0",
-    "eslint-plugin-react": "7.20.6",
-    "exports-loader": "1.1.0",
```
hmhealey, line 148: This and `imports-loader` don't appear to be used

package.json
```diff
@@ -242,7 +239,7 @@
   "scripts": {
     "check": "eslint --ext .js,.jsx,.tsx,.ts . --quiet --cache",
     "fix": "eslint --ext .js,.jsx,.tsx,.ts . --quiet --fix --cache",
-    "build": "cross-env NODE_ENV=production webpack --display-error-details --verbose",
```
hmhealey, line 245: These extra attributes aren't supported any more. They were added for Jenkins for some reason, so hopefully they're not needed for CircleCI or GitLab

webpack.config.js
```diff
@@ -80,9 +80,6 @@ var MYSTATS = {
     // Add the hash of the compilation
     hash: true,
 
-    // Set the maximum number of modules to be shown
-    maxModules: 0,
```
hmhealey, line 84: This has been removed

webpack.config.js
```diff
@@ -250,6 +247,9 @@ var config = {
             superagent: 'node_modules/superagent/lib/client',
         },
         extensions: ['.ts', '.tsx', '.js', '.jsx'],
+        fallback: {
+            stream: require.resolve('stream-browserify'),
```
hmhealey, line 251: Similar to `crypto`, we need to import `stream` ourselves as well. Since we don't use it directly (it's used by `crypto-browserify`), we have to tell Webpack to alias it instead of updating import paths

components/admin_console/admin_definition.test.jsx
```diff
@@ -18,7 +18,9 @@ const baseShape = {
 const fieldShape = {
     ...baseShape,
     key: yup.string().required(),
-    help_text: yup.string(),
+
+    // help_text: yup.string(), // Commented out since this doesn't work when help_text is a ReactNode
```
devinbinnie, line 22: Should we just remove this?

devinbinnie (approved):

components/admin_console/admin_definition.test.jsx
hmhealey: I think I left it for documentation purposes. I'm 0/5 since this test will become unnecessary as soon as admin_definitions.jsx is migrated to TS anyway

components/admin_console/generated_setting.tsx
hmhealey: Never mind that. We're now using `crypto-browserify` implicitly since that's required for the type definitions to work.

calebroseland (approved):

hmhealey:
Blocking this because the changes to webpack seem to have broken some sections of the app like the system console

hmhealey:
Okay, I think I've fixed the issue. The reason the system console was broken is because it used the Node `crypto` library which we had to polyfill since Webpack 5 no longer includes built-in polyfills for Node libraries. That was crashing because `crypto` tried to use Node's `process` package which also did not have a polyfill. We now have a separate polyfill for that.

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Short smoke test and complete E2Es cycle run on the PR server compared to last production tests on master; no PR specific failures. Thanks @saturninoabril for running the E2Es for me. 
[PR results](https://community-daily.mattermost.com/core/pl/mqx7osoerbrmzdpyctq9jxn4ow)
[Master prod results](https://community-daily.mattermost.com/core/pl/q4an4wbd4pg35fmzioso67icoy)

PR #7513, opened 2021-02-11
fix group modal tests

sbishel:
#### Summary
Reactivates a couple tests with fixes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-17122


sbishel:
The tests were failing because of line
`expect(instance.handleHide).not.toBeCalled();`
It appears in the code that this function should have been called, but please double check. I was unable to find a reason why this test would have "stopped" working, which worries me I mis-diagnosed.

components/add_groups_to_channel_modal/add_groups_to_channel_modal.test.jsx
```diff
@@ -86,9 +89,9 @@ describe('components/AddGroupsToChannelModal', () => {
         setTimeout(() => {
             expect(instance.handleResponse).toBeCalledTimes(2);
         }, 0);
```
hahmadia, line 91: I think this part can be refactored. Not sure why it's written the way it is but when I tried locally, seemed to pass fine.
```suggestion
          expect(instance.handleResponse).toBeCalledTimes(2);
```

components/add_groups_to_channel_modal/add_groups_to_channel_modal.test.jsx
```diff
@@ -75,6 +77,7 @@ describe('components/AddGroupsToChannelModal', () => {
         wrapper.setState({values: []});
         await wrapper.instance().handleSubmit({preventDefault: jest.fn()});
         expect(actions.linkGroupSyncable).not.toBeCalled();
+        expect(instance.handleHide).not.toBeCalled();
```
hahmadia, line 80: Also  add the following wouldn't hurt
```suggestion
        expect(instance.handleResponse).not.toBeCalled();
        expect(instance.handleHide).not.toBeCalled();
```

components/add_groups_to_channel_modal/add_groups_to_channel_modal.test.jsx
```diff
@@ -86,9 +89,9 @@ describe('components/AddGroupsToChannelModal', () => {
         setTimeout(() => {
             expect(instance.handleResponse).toBeCalledTimes(2);
         }, 0);
-        expect(instance.handleHide).not.toBeCalled();
+        expect(instance.handleHide).toHaveBeenCalledTimes(1);
```
hahmadia, line 92: So the author of the tests initial intention was correct. The `instance.handleHide` should not be called. The reason it's still being called is because they've incorrectly initialized above `linkGroupSyncable`. If we fix that, then the test should fall in line with the authors intentions.

components/add_groups_to_channel_modal/add_groups_to_channel_modal.test.jsx
```diff
@@ -59,7 +61,7 @@ describe('components/AddGroupsToChannelModal', () => {
         expect(wrapper.state('addError')).toEqual(message);
     });
 
-    /*test('should match state when handleSubmit is called', async () => {
+    test('should match state when handleSubmit is called', async () => {
         const linkGroupSyncable = jest.fn().
             mockResolvedValueOnce({error: true}).
             mockResolvedValue({data: true});
```
hahmadia, line 67: Here, the initial author of the test didn't correctly make the mock of `linkGroupSyncable` function which was causing the function to fail below. The following change should make it so that `handleHide` is not called.
```suggestion
        const linkGroupSyncable = jest.fn().mockResolvedValue({error: true, data: true});
```
The way it's currently written, only on the first call of the function will we get a `error = true` and afterwards error will just be `undefined` which is why the `handleHide` was being called once because of the second function (There are 2 ids that are looped through) call happening. I persume the author meant for error to return true every time hence why they said `handleHide.not.called`.

components/add_groups_to_team_modal/add_groups_to_team_modal.test.jsx
```diff
@@ -83,9 +86,9 @@ describe('components/AddGroupsToTeamModal', () => {
         setTimeout(() => {
             expect(instance.handleResponse).toBeCalledTimes(2);
         }, 0);
```
hahmadia, line 88: Can we refactor this to the following? Not sure why it's written the way it is
```suggestion
        expect(instance.handleResponse).toBeCalledTimes(2);
```

components/add_groups_to_team_modal/add_groups_to_team_modal.test.jsx
```diff
@@ -72,6 +74,7 @@ describe('components/AddGroupsToTeamModal', () => {
         wrapper.setState({values: []});
         await wrapper.instance().handleSubmit({preventDefault: jest.fn()});
         expect(actions.linkGroupSyncable).not.toBeCalled();
+        expect(instance.handleHide).not.toBeCalled();
```
hahmadia, line 77: Lets add this here for completeness
```suggestion
        expect(instance.handleResponse).not.toBeCalled();
        expect(instance.handleHide).not.toBeCalled();
```

components/add_groups_to_team_modal/add_groups_to_team_modal.test.jsx
```diff
@@ -56,7 +58,7 @@ describe('components/AddGroupsToTeamModal', () => {
         expect(wrapper.state('addError')).toEqual(message);
     });
 
-    /*test('should match state when handleSubmit is called', async () => {
+    test('should match state when handleSubmit is called', async () => {
         const linkGroupSyncable = jest.fn().
             mockResolvedValueOnce({error: true}).
             mockResolvedValue({data: true});
```
hahmadia, line 64: See above for explanation of this change
```suggestion
        const linkGroupSyncable = jest.fn().mockResolvedValue({error: true, data: true});
```

components/add_groups_to_team_modal/add_groups_to_team_modal.test.jsx
```diff
@@ -83,9 +86,9 @@ describe('components/AddGroupsToTeamModal', () => {
         setTimeout(() => {
             expect(instance.handleResponse).toBeCalledTimes(2);
         }, 0);
-        expect(instance.handleHide).not.toBeCalled();
+        expect(instance.handleHide).toHaveBeenCalledTimes(1);
```
hahmadia, line 89: If we make above change to `linkGroupSyncable` , this can stay as it is without changing it
```suggestion
      expect(instance.handleHide).not.toBeCalled();
```

hahmadia (changes requested):
After triple checking, I was able to reason with what initial authors intentions. I've proposed changes. Let me know if you feel I went wrong anywhere.

components/add_groups_to_channel_modal/add_groups_to_channel_modal.test.jsx
sbishel, line 92:
```suggestion
        expect(instance.handleHide).not.toBeCalled();
```

hahmadia (changes requested):
Some more small changes. Thanks

hahmadia (approved):
LGTM! Thanks for the changes 🤠

mkraft (approved):

PR #7515, opened 2021-02-11
MM-30531: importing slack theme in new format

michelengelen:
#### Summary
updated the importing functionality to accept the new format slack uses for their theming

#### Ticket Link
Fixes [MM-30531](https://mattermost.atlassian.net/browse/MM-30531)

#### Screenshots
**inserting theme-string does not show an error**
---
Screenshot 2021-02-11 at 13 35 08

**theme is updated correctly**
---
Screenshot 2021-02-11 at 13 35 45


michelengelen:
@matthewbirtch I do not know exactly if a UX-review is needed here. I added you just to be on the safe side! 🙇🏻

components/user_settings/import_theme_modal.tsx
```diff
@@ -64,15 +64,36 @@ class ImportThemeModal extends React.PureComponent<WrappedComponentProps, State>
...
         }
 
+        /*
+         * index mapping of slack theme format (index => slack-property name)
+         *
+         * |-------|-------------------------|-------------------------|
+         * | index | Slack theme-property    | MM theme-property       |
+         * |-------|-------------------------|-------------------------|
+         * |   0   | Column BG               | sidebarBg               |
+         * |   1   | ???                     | sidebarHeaderBg         |
```
nevyangelova, line 74: I assume we will have a value for this?

nevyangelova (approved):

components/user_settings/import_theme_modal.tsx
michelengelen: Yes, we do, but there is no value name given in Slack

michelengelen: Screenshot 2021-02-11 at 14 37 18

The value is there, but there is no corresponding field for it above.

nevyangelova: Understand, maybe we add something like Sidebar Header or Column header just for our own clarity. 0/5 on this

michelengelen: well, we could, but since we are mapping it against our own `sidebarHeaderBg` property it feels obsolete.

How about we just remove the `???` and just leave it blank (0/5) or add `---` (1/5)?

components/user_settings/import_theme_modal.tsx
```diff
@@ -15,7 +15,7 @@ const ActionTypes = Constants.ActionTypes;
 
 type State = {
     value: string;
-    inputError: any;
+    inputError: JSX.Element | null;
```
devinbinnie, line 18: Can use the `React.ReactNode` type here instead, it covers all types that can be rendered to the DOM by React.

components/user_settings/import_theme_modal.tsx
```diff
@@ -64,15 +64,36 @@ class ImportThemeModal extends React.PureComponent<WrappedComponentProps, State>
...
         const [
-            sidebarBg, // 0
-            sidebarHeaderBg, // 1
-            sidebarTextActiveBorder, // 2
-            sidebarTextActiveColor, // 3
-            sidebarTextHoverBg, // 4
-            sidebarText, // 5
-            onlineIndicator, // 6
-            mentionBg, // 7
+            sidebarBg,
```
devinbinnie, line 89: Is there ever a case where one of these values is missing?

devinbinnie (approved):

components/user_settings/import_theme_modal.tsx
michelengelen: no, that cannot happen, since we check for the length of the array before ever coming to this step!

matthewbirtch:
The only thing I wondered is if there is a color that maps for the button bg color?

In the Image below, I imported a slack theme called 'Terminal', but it uses the default theme button color in Mattermost.
image

Not sure if there is an obvious choice for what to map. We could use 'Active Item' as the button bg color and then 'active item text' as the button text color.

michelengelen:
@matthewbirtch i am 1/5 that this is out of scope for this PR. Especially with the upcoming changes/features (CC) in mind.

matthewbirtch:
> @matthewbirtch i am 1/5 that this is out of scope for this PR. Especially with the upcoming changes/features (CC) in mind.

Ya, that's fine with me. I think users will expect some things not to perfectly map between slack and MM anyway.

matthewbirtch (approved):

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
-Verified importing themes form Slack - theme applies and displays as expected.

PR #7516, opened 2021-02-11
[MM-31411][MM-32164] - Add profile picture to Direct Channels link

nevyangelova:
#### Summary
This PR adds a profile picture to the sidebar DM links as per design https://www.figma.com/file/RdJqKp46cfH8gzNVehIShd/MM-31441-User-Avatars-in-Channel-Sidebar?node-id=236%3A5219
for Post GA. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31441
https://mattermost.atlassian.net/browse/MM-32164


components/profile_picture/profile_picture.tsx
```diff
@@ -94,14 +97,18 @@ export default class ProfilePicture extends React.PureComponent<Props> {
...
                         size={this.props.size}
                         url={this.props.src}
                     />
                 </span>
-                <StatusIcon status={this.props.status}/>
+                {this.props.newStatusIcon ?
+                    <StatusIconNew
+                        className={this.props.statusClass}
+                        status={this.props.status}
+                    /> : <StatusIcon status={this.props.status}/>}
```
michelengelen, line 111: Could you please wrap the first expression from the ternary in paranthese? Thanks!  🙇🏻

components/sidebar/sidebar_channel/sidebar_direct_channel/sidebar_direct_channel.tsx
```diff
@@ -83,24 +85,6 @@ class SidebarDirectChannel extends React.PureComponent<Props, State> {
             return (
                 <i className='icon icon-archive-outline'/>
             );
```
michelengelen, line 87: I suppose this condition could be reversed.
My suggestion would be something like this:

```
if (teammate?.id && teammate?.delete_at) {
    return <i className='icon icon-archive-outline'/>;
}

return null;
```

michelengelen (changes requested):
just 2 little remarks! ✨

components/sidebar/sidebar_channel/sidebar_direct_channel/index.ts
```diff
@@ -53,6 +54,7 @@ function mapStateToProps(state: GlobalState, ownProps: OwnProps) {
         redirectChannel,
         active,
         botIconUrl: botIconImageUrl(teammate),
+        isMobile: isMobile(),
```
devinbinnie, line 57: Nit: Don't need to pass this through the redux connector, we can just check this directly in the component.

devinbinnie (approved):

components/sidebar/sidebar_channel/sidebar_direct_channel/sidebar_direct_channel.tsx
nevyangelova: we're returning the profile picture after so I can't flip it. Also I believe the early return is by Mattermost convention 🤔

components/sidebar/sidebar_channel/sidebar_direct_channel/index.ts
nevyangelova: If it goes through redux it updates as you resize the window though, otherwise you need a reload 🤷‍♀️

components/sidebar/sidebar_channel/sidebar_direct_channel/sidebar_direct_channel.tsx
michelengelen: Fair enough ... I kinda missed, that the function does not get terminated after that bracket ... my bad 🙇🏻

michelengelen (approved):

matthewbirtch (changes requested):
This is great @nevyangelova! Just a few really minor details:

1. For the outline that appears around the status icon, is it possible for that to match the hover color and active color?

image

image

2. On mobile webview, the avatars get larger and cause the channel names to be misaligned with other channel names. Maybe we just keep it the same size (20x20)?
image

nevyangelova:
@matthewbirtch point 1 is easy to do, but for point 2 we can fix the padding instead of shrinking the icons? Since the design wanted them bigger? But I can do both. LMK!

michelengelen:
> @matthewbirtch point 1 is easy to do, but for point 2 we can fix the padding instead of shrinking the icons? Since the design wanted them bigger? But I can do both. LMK!

@nevyangelova i am 4/5 same-size icons with adjusted padding looks way better! 👍🏻

matthewbirtch:
> @matthewbirtch point 1 is easy to do, but for point 2 we can fix the padding instead of shrinking the icons? Since the design wanted them bigger? But I can do both. LMK!

Yes, fixing the padding so it aligns better makes sense. However, please note that we have another ticket for increasing the size of the public/private channel icons to better balance with the new avatars. https://mattermost.atlassian.net/browse/MM-32164

nevyangelova:
Oh It might make sense to implement togerther!

matthewbirtch:
> Oh It might make sense to implement togerther!

I was wondering that, but I believe it's in the 'common components' epic right now

nevyangelova:
@matthewbirtch updated with the new ticket please take a look

matthewbirtch (changes requested):
Thanks @nevyangelova.

Looks good, except I think we need to adjust the padding on the left side. The icons/avatars look like they might be indented more than they should be. I think if we reduce the space on the left side by 7px, it should do the trick

nevyangelova:
@matthewbirtch I adjusted the icons but also take a look at DM avatars and group messages because I am not sure how much these should be. I did a calculation based on the width of the space and padding around the globe icon to make it the same for the other icons too.

nevyangelova:
@matthewbirtch the test server is finally updated :D

matthewbirtch:
Thanks @nevyangelova indentation looks better now. I just noticed that group message icons and labels aren't quite aligned though.

image

I also noticed in mobile webview, the labels are not aligned and the avatars are 24x24 instead of 20x20?: 
image

Sorry, I didn't catch this in previous reviews.

nevyangelova:
@matthewbirtch apologies, accidentally miscalculated this padding

matthewbirtch:
@nevyangelova what about mobile view?

> I also noticed in mobile webview, the labels are not aligned and the avatars are 24x24 instead of 20x20?:
> image

nevyangelova:
@matthewbirtch  I had some issues with webapp the past few days but now I was able to resolve it. Pushed a fix

michelengelen (approved):

matthewbirtch:
> 24x24 instead of 20x20

Sorry to be nitpicky on this, but is there a reason the the avatar size changes to 24x24 in the mobile webview?

nevyangelova:
> > 24x24 instead of 20x20
> 
> Sorry to be nitpicky on this. The text is now aligned, but is there a reason the avatar size changes to 24x24 in the mobile webview?

As far as I saw the designs specify the icons to be 24x24 in mobile view. Is this not correct?

matthewbirtch:
> > > 24x24 instead of 20x20
> > 
> > 
> > Sorry to be nitpicky on this. The text is now aligned, but is there a reason the avatar size changes to 24x24 in the mobile webview?
> 
> As far as I saw the designs specify the icons to be 24x24 in mobile view. Is this not correct?

Oh, sorry. The Mobile design provided in that file was for the mobile app. Not for the mobile webview

nevyangelova:
@matthewbirtch ok I understand now. I am not sure if it was labelled when I first made the PR but its stated clear now that this is the mobile app view, my mistake. Sorry for the confusion

matthewbirtch:
> @matthewbirtch ok I understand now. I am not sure if it was labelled when I first made the PR but its stated clear now that this is the mobile app view, my mistake. Sorry for the confusion

I think that was my fault. I changed the label in Figma to be clearer now.

nevyangelova:
@matthewbirtch any additional comments on this?

matthewbirtch (approved):
Looks good to go @nevyangelova! Thanks

jgilliam17:
@nevyangelova Has mention badge size been reduced per MM-32164? I'm not seeing any difference on the PR server. Let me know. Thanks.

jgilliam17:
One more thing, channel icons (and maybe draft one too?) are back out of line
Screen Shot 2021-03-03 at 4 57 38 PM

nevyangelova:
@jgilliam17 kind reminder to review

matthewbirtch:
> One more thing, channel icons (and maybe draft one too?) are back out of line

@jgilliam17 is right. Just took a closer look at the latest and the icons are out of alignment again. @nevyangelova can you make sure things are aligned?

image


Also, it looks like the labels for group messages might be off by 1px

jgilliam17:
Thanks @nevyangelova Just took a look and I am not seeing the change on the PR server.  Can you please check?

nevyangelova:
@jgilliam17 should work now :)

jgilliam17 (approved):
Thanks @nevyangelova 
Tested, looks good to merge.
- Verified DM avatars, channel DM/GM name alignment and sidebar icon alignment.

PR #7536, opened 2021-02-17
Remove output from pkill

hmhealey:
`-l` seems to be a non-standard flag that's supported on Mac, not Ubuntu, so we can't use that to have the killed processes output. `-f` is still needed since that causes it to match process arguments as well (since previous versions of webpack ran as `node webpack`).

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/7512

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ stop: ## Stops webpack
 ifeq ($(OS),Windows_NT)
 	wmic process where "Caption='node.exe' and CommandLine like '%webpack%'" call terminate
 else
-	@pkill -fl webpack || true
+	@pkill -f webpack || true
 endif
 
 restart: | stop run ## Restarts the app


Willyfrog (approved):
if you want to have the output of the process that it is going to be killed you could perform a `pgrep -l webpack` prior to that, which should print out the process name.

LGTM anyways

agarciamontoro (approved):

agarciamontoro:
@hmhealey, does this need QA review?

hmhealey:
Nah, it's fine without it since it just affects running from the Makefile. Thanks for checking though

amyblais:
@hmhealey https://github.com/mattermost/mattermost-webapp/pull/7512 is included for the next Cloud release on Feb 24th (which will also become v5.33), do you think it would be good to cherry pick this to the ``cloud`` branch?

hmhealey:
@amyblais That's a good call. It's not strictly necessary for the release since it only has an effect during development, but I'm sure it'll make someone's life easier if they're developing on the cloud/release branch.

PR #7539, opened 2021-02-17
[MM-20386] migrate popover_bar to typescript

kojiGit55:
#### Summary

migrates PopoverBar component to typescript.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-20386


components/view_image/popover_bar/popover_bar.tsx
```diff
@@ -96,16 +99,17 @@ export default class PopoverBar extends React.PureComponent {
             );
         }
 
-        let zoomControls = [];
+        const zoomControls: React.ReactNode[] = [];
+        let wrappedZoomControls: React.ReactNode = null;
```
kojiGit55, line 103: previous code was as follows, and TypeScript complained about assigning a React Node to array:
1. initialize `zoomControls` with empty array
2. push some react nodes to it 
3. wrap`zoomControls` with div and assign it to `zoomControls` itself

so I created new variable `wrappedZoomControls` to make it simpler

michelengelen:
Hi @kojiGit55 and thanks for your contribution. 🎉

components/view_image/popover_bar/popover_bar.tsx
michelengelen: IMO the `wrappedZoomControls` is not needed. It is generally better to avoid such constructs, that is true, but instantiating a new variable just to wrap elements, is a bit much.

Could you please move the wrapping `<div>` to the actual rendering?

components/view_image/popover_bar/popover_bar.tsx
```diff
@@ -235,7 +239,7 @@ export default class PopoverBar extends React.PureComponent {
                         />
                     </span>
                 </div>
-                {zoomControls}
+                {wrappedZoomControls}
```
michelengelen, line 242:
```suggestion
                <div className='modal-column'>
                    {zoomControls}
                </div>
```

michelengelen (approved):
Just a minor remark! Otherwise LGTM!
I restarted the CI, since it just ran out of memory.

components/view_image/popover_bar/popover_bar.tsx
nevyangelova: Agreed, just need to make sure if the rendering of the empty <div> doesn't add any changes to the styling

nevyangelova (approved):
LGTM

components/view_image/popover_bar/popover_bar.tsx
kojiGit55:
```suggestion
{
  this.props.showZoomControls ? zoomControls : (
    <div className='modal-column'>
      {zoomControls}
    </div>
  )
 }```

components/view_image/popover_bar/popover_bar.tsx
kojiGit55: i see, i can move the `<div>` to actual rendering. 
but in your suggestion, we render the div when `this.props.showZoomControls == true` as well, which is different from master's logic. is that OK? or should I do conditional rendering as I wrote in suggestion?

jgilliam17 (approved):
Thank you @kojiGit55 
Tested, looks good to merge.
- Verified popover bar on file attachment modal.
@michelengelen can you please help merge. Thanks 🙂

PR #7543, opened 2021-02-18
Cypress/E2E: Add request on first E2E test

saturninoabril:
#### Summary
Timeout error on first request to update the config consistently happened before on fresh remote test server. However, lately, I'm seeing the same to server pre-generated with sample data (using `sampledata` CLI command). With that, I just moved the dummy call to a common logic and add a `CYPRESS_firstTest` environment variable so it will trigger on first test only upon running `node run_tests.js` (the one we're using in CI). In local dev, it's set to `false` by default.


josephbaylon (approved):
LGTM. Thanks @saturninoabril !!

furqanmlk (approved):
Thanks @saturninoabril

PR #7546, opened 2021-02-18
[MM-31502] - Add Notify Admin link to invitation modal

AGMETEOR:
#### Summary
This PR adds a link where users can notify admin about free tier over limit
Screenshot 2021-02-18 at 16 52 17
Screenshot 2021-02-18 at 16 52 48
Screenshot 2021-02-18 at 16 53 12

#### Ticket Link
[MM-31502](https://mattermost.atlassian.net/secure/RapidBoard.jspa?rapidView=79&projectKey=MM&modal=detail&selectedIssue=MM-31502&quickFilter=382)

#### Related Pull Requests
[Server](https://github.com/mattermost/mattermost-server/pull/16882)


components/widgets/links/notify_link.tsx
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useState} from 'react';
+import {useIntl} from 'react-intl';
+import {Client4} from 'mattermost-redux/client';
+
+const NotifyLink = (): JSX.Element => {
+    const NOT_STARTED = 'Not_Started';
```
pablovelezvidal, line 9: maybe we could leverage in the typescript enum struct for declaring these values?

components/widgets/links/notify_link.tsx
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useState} from 'react';
+import {useIntl} from 'react-intl';
+import {Client4} from 'mattermost-redux/client';
+
+const NotifyLink = (): JSX.Element => {
```
pablovelezvidal, line 8: could we add unit test to cover this new component?

pablovelezvidal:
great work @AGMETEOR ! just a couple of comments, but besides that, it looks good to me.

marianunez (approved):
Nothing more to add to Pablos review, thanks @AGMETEOR!

pablovelezvidal (approved):
LGTM! great stuff @AGMETEOR !

stevemudie (commented):
@AGMETEOR On the spinwick I created 7 users with the `/test` tool and then went about inviting users to trigger the errors but didn't see any on the first pass. All it did was tell me that 2 were sent and 2 were not (only 2 openings left at that point with the admin and the 7 users I created).

I then went in and deactivated some and tried again and did see the errors but overall it wasn't consistent as I signed in with some users and deactivated others etc.

Can you take a look at the Spinwick and see if you get similar results?

marianunez:
> @AGMETEOR On the spinwick I created 7 users with the `/test` tool and then went about inviting users to trigger the errors but didn't see any on the first pass. All it did was tell me that 2 were sent and 2 were not (only 2 openings left at that point with the admin and the 7 users I created).
> 
> I then went in and deactivated some and tried again and did see the errors but overall it wasn't consistent as I signed in with some users and deactivated others etc.
> 
> Can you take a look at the Spinwick and see if you get similar results?

@AGMETEOR I also noticed that it always seemed to be showing the error saying that you can only invite 2 members even though it had 4 seats left. 

In the error message it needs to say how many users are remaining in the free limit in total, so we may need to revisit this check: https://github.com/mattermost/mattermost-webapp/blob/6f25d9e27da685fb817b92a5ba939b80502839c5/components/invitation_modal/invitation_modal_members_step/invitation_modal_members_step.jsx#L152-L156

AGMETEOR:
> > @AGMETEOR On the spinwick I created 7 users with the `/test` tool and then went about inviting users to trigger the errors but didn't see any on the first pass. All it did was tell me that 2 were sent and 2 were not (only 2 openings left at that point with the admin and the 7 users I created).
> > I then went in and deactivated some and tried again and did see the errors but overall it wasn't consistent as I signed in with some users and deactivated others etc.
> > Can you take a look at the Spinwick and see if you get similar results?
> 
> @AGMETEOR I also noticed that it always seemed to be showing the error saying that you can only invite 2 members even though it had 4 seats left.
> 
> In the error message it needs to say how many users are remaining in the free limit in total, so we may need to revisit this check:
> 
> https://github.com/mattermost/mattermost-webapp/blob/6f25d9e27da685fb817b92a5ba939b80502839c5/components/invitation_modal/invitation_modal_members_step/invitation_modal_members_step.jsx#L152-L156

pablovelezvidal:
I validated that scenario, and I am also getting the right number
Captura de pantalla 2021-02-22 a las 12 45 42

stevemudie:
This is working but it took a hard refresh in order to get it to recognize exactly how many users I could invite after I'd deactivated some to test it.

marianunez:
> This is working but it took a hard refresh in order to get it to recognize exactly how many users I could invite after I'd deactivated some to test it.

This is probably what I experienced as well. Given that we are ladino the subscription stats on app load they are not updated when a user is deactivated/activated. We can fix this by refreshing the stats when we get those web sockets event. @stevemudie can you log a bug to be tackled outside of this PR ?

AGMETEOR:
@stevemudie is it ok if we merge this now?

stevemudie:
> @stevemudie is it ok if we merge this now?

Yes, other than the issues we discussed we're all set here from my view.

marianunez:
@amyblais we need to cherry-pick this to cloud for the upcoming release given that without it will show the option to upgrade for end-users and should be limited to Admins only.

PR #7548, opened 2021-02-18
[MM-27663] Don't focus the post textbox if the user presses SPACE while focused on a button or anchor

devinbinnie:
#### Summary
We had some undesirable UX happening when using keyboard navigation to "drag and drop" channels around in the channel sidebar.

When pressing the SPACE key to select a channel, the listener for the post textbox would give focus to the post textbox, and when the drag was finished, if you were to continue tabbing through the app, you would be taken to the post textbox instead of the next channel (despite the post textbox being highlighted).

Normally, `react-beautiful-dnd` would be responsible for returning focus to the original element, but in this case we are focusing the `a` element instead of the `div` element (the `div` element normally contains the drag handle). Since the last focused element, in the eyes of `react-beautiful-dnd` was the `a` and not the `div`, it would not return focus.

Additionally, since the post textbox operates its own listener outside of the `a11y_controller`, it would not visually show that focus had changed.

This PR stops the post textbox from being focused if either a `button` or an `a` is currently focused and the SPACE key is pressed, thus stopping the focus from being jarringly moved away after interacting with either of these elements when using the SPACE key.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27663


hmhealey (approved):
This sounds good to me. I almost wonder if we might want to repurpose the A11yController as a FocusController in the future since we're probably going to want to implement something like focus capture for modals and it might also be beneficial to make it responsible for any "default" focused elements like the post textbox.

deanwhillier (approved):

jgilliam17 (approved):
Thank you @devinbinnie 
Tested, looks good to merge.
- Verified user can continue to scroll through channels using tab after any channel is dropped using space bar. Focus remains on the LHS and no longer jumps to main channel textbox.

PR #7550, opened 2021-02-18
Downgrade @hot-loader/react-dom and fix make dev

hmhealey:
When doing the dependency upgrades, I didn't know how to use `make dev`, so I didn't realize it was broken. This fixes it by:
1. Downgrading `@hot-loader/react-dom` since I accidentally upgraded it when we weren't ready to upgrade to React 17 yet.
2. Switching to the newer way of starting webpack-dev-server using `webpack serve`
3. Fixing the name of the source map type since the old name must've been removed in Webpack 5.

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/7512


package.json
```diff
@@ -242,7 +242,7 @@
     "fix": "eslint --ext .js,.jsx,.tsx,.ts . --quiet --fix --cache",
     "build": "cross-env NODE_ENV=production webpack",
     "run": "webpack --progress --watch",
-    "dev-server": "webpack-dev-server -d",
+    "dev-server": "webpack serve",
```
ethervoid, line 245: Probably we want to pass `--env development --mode development`?

ethervoid (approved):
LGTM! Just a minor comment

package.json
hmhealey: Good call. I'll add that

hahmadia (approved):

AGMETEOR (approved):

PR #7554, opened 2021-02-19
Fix custom status possible race condition

jespino:
#### Summary
Fix custom status possible race condition. Looks like in some cases the user
requests is not yet in the redux-store whenever this is trying to show the
custom status. So it fails. This new approach simply ignore that situation.

#### Ticket Link
[MM-33047](https://mattermost.atlassian.net/browse/MM-33047)

### DIFF
--- a/selectors/views/custom_status.ts
+++ b/selectors/views/custom_status.ts
@@ -15,7 +15,7 @@ export function makeGetCustomStatus(): (state: GlobalState, userID?: string) => 
     return createSelector(
         (state: GlobalState, userID?: string) => (userID ? getUser(state, userID) : getCurrentUser(state)),
         (user) => {
-            const userProps = user.props || {};
+            const userProps = user && user.props ? user.props : {};
             return userProps.customStatus && JSON.parse(userProps.customStatus);
         },
     );
--- a/selectors/views/custom_status.ts
+++ b/selectors/views/custom_status.ts
@@ -15,7 +15,7 @@ export function makeGetCustomStatus(): (state: GlobalState, userID?: string) => 
     return createSelector(
         (state: GlobalState, userID?: string) => (userID ? getUser(state, userID) : getCurrentUser(state)),
         (user) => {
-            const userProps = user && user.props ? user.props : {};
+            const userProps = user?.props || {};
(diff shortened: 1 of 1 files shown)


jespino:
This should be cherry-picked to master whenever is confirmed the fix in community-release.

reflog (approved):

selectors/views/custom_status.ts
```diff
@@ -15,7 +15,7 @@ export function makeGetCustomStatus(): (state: GlobalState, userID?: string) =>
     return createSelector(
         (state: GlobalState, userID?: string) => (userID ? getUser(state, userID) : getCurrentUser(state)),
         (user) => {
-            const userProps = user.props || {};
+            const userProps = user && user.props ? user.props : {};
```
lieut-data, line 18: Would it make sense to simplify to:
```suggestion
            const userProps = user?.props || {};
```

lieut-data (approved):

amyblais:
@jespino Should this be submitted to master too? Or does this only exist in cloud?

jespino:
@amyblais yes, my plan is to cherry-pick it to master once this is confirmed as a working solution. I wasn't able to reproduce it locally.

lindalumitchell:
@jespino and @ogi-m I spun up a server to poke around at a bit, but setting enable custom statuses to true doesn't appear to be "sticking". When I set it to true and save it appears to save as true. But then when I refresh, it reverts to false. I haven't tested this feature before though; am I missing something?

Screen Shot 2021-02-19 at 7 41 41 AM

ogi-m:
No idea why that's happening, it's working fine on the test server for custom statuses. 
@jespino Can you take a look please?

Also, do we know what exactly was causing the crash? I wasn't able to repro the issue at all by switching teams, so I'm not sure how to accurately verify the fix (once we manage to enable it)

jespino:
@ogi-m my guess about the error root cause is that was trying to access a user that wasn't already in the redux-store, basically a race condition.

jespino:
@lindalumitchell it works perfectly fine locally, I'll check it out in the test server.

amyblais:
To be cherry picked to master.

PR #7556, opened 2021-02-19
Handle errors on binding locations

larkox:
#### Summary
Handle errors in the binding locations, instead of the call handler.

#### Ticket Link
None

#### Related Pull Requests
None

#### Screenshots
None


actions/post_actions.jsx
```diff
@@ -185,6 +185,7 @@ export function pinPost(postId) {
         if (rhsState === RHSStates.PIN) {
             addPostToSearchResults(postId, state, dispatch);
         }
+        return {data: true};
```
catalintomai, line 188: are these changes related to apps?

components/dot_menu/dot_menu.tsx
```diff
@@ -290,6 +293,12 @@ export default class DotMenu extends React.PureComponent<Props, State> {
                 post_id: this.props.post.id,
                 root_id: this.props.post.root_id,
             },
+        }).then((res) => {
+            const callResp = (res as {data: AppCallResponse}).data;
```
catalintomai, line 297: can this snippet be factored out and reused throughout somehow?

actions/post_actions.jsx
larkox: In order to be recognized in Typescript as Action func, they cannot return void, they must return either {data: any} or {error: any}.

components/dot_menu/dot_menu.tsx
larkox: I prefer not. Each binding should handle its own errors, and I worry the moment these two bindings start diverging we need to refactor this in one more time.

components/post_view/embedded_bindings/select_binding/select_binding.tsx
```diff
@@ -80,7 +84,31 @@ export default class SelectBinding extends React.PureComponent<Props, State> {
...
+        const callResp = (res as {data: AppCallResponse}).data;
+        const ephemeral = (message: string) => sendEphemeralPost(message, this.props.post.channel_id, this.props.post.root_id);
+        switch (callResp.type) {
+        case AppCallResponseTypes.OK:
+            if (callResp.markdown) {
+                ephemeral(callResp.markdown);
+            }
+            break;
+        case AppCallResponseTypes.ERROR: {
+            const errorMessage = callResp.error || this.props.intl.formatMessage({id: 'apps.error.unknown', defaultMessage: 'Unknown error happenned'});
```
catalintomai, line 97: curious, in what circumstances can we have `AppCallResponseTypes.ERROR` but `callResp.error` not set?

catalintomai (approved):

components/post_view/embedded_bindings/select_binding/select_binding.tsx
larkox: In theory, only at the doAppCall level, in case the app returns a malformed response (error type with no error text).
I still make the check everywhere else to be extra cautious. Potentially, we can assume that error will be always set at this point, and move forward without this logic.

mickmister (approved):
LGTM! Awesome job with the intl :+1:

PR #7563, opened 2021-02-21
MM-32882: Add an unread badge/marker to the Main Menu icon and the ‘Plugin Marketplace’ Menu Item

catalintomai:
#### Summary
Add an unread badge/marker to the Main Menu icon and the ‘Plugin Marketplace’ Menu Item until an admin user visits the marketplace for the first time

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32882

#### Related Pull Requests
server: https://github.com/mattermost/mattermost-server/pull/16992
api: https://github.com/mattermost/mattermost-api-reference/pull/616

#### Screenshots
TODO: Needs further refinement for UX

Screen Shot 2021-02-21 at 10 37 41 PM


components/toggle_modal_button_redux/toggle_modal_button_redux.jsx
```diff
@@ -55,6 +61,16 @@ class ModalToggleButtonRedux extends React.PureComponent {
 
         const ariaLabel = formatMessage({id: 'accessibility.button.dialog', defaultMessage: '{dialogName} dialog'}, {dialogName: props.accessibilityLabel});
 
+        let badge = null;
+        if (this.props.modalId === ModalIdentifiers.PLUGIN_MARKETPLACE) {
+            const firstAdminVisitMarketplaceStatus = getFirstAdminVisitMarketplaceStatus(store.getState());
```
hanzei, line 66: I'm a bit surprised to that logic here. Does it make sense to add a `Prop` e.g. `Unread` to `ModalToggleButtonRedux` and pass the value of `getFirstAdminVisitMarketplaceStatus` as such?

hanzei (changes requested):
I'm not able to test this locally right now, but will do later.

hanzei:
Heads up that I'M getting a JS error when opening the test server:
```
Uncaught TypeError: (0 , i.firstAdminVisitMarketplaceStatus) is not a function
    at Function.mapToProps (index.js:23)
    at n (wrapMapToProps.js:41)
    at Function.n.mapToProps (wrapMapToProps.js:50)
    at n (wrapMapToProps.js:41)
    at selectorFactory.js:22
    at connectAdvanced.js:319
    at Object.useMemo (react-dom.production.min.js:165)
    at t.useMemo (react.production.min.js:25)
    at v (connectAdvanced.js:304)
    at Ko (react-dom.production.min.js:153)
```

larkox (approved):
LGTM

components/root/root.jsx
```diff
@@ -232,6 +233,10 @@ export default class Root extends React.PureComponent {
                 GlobalActions.redirectUserToDefaultTeam();
             }
             this.onConfigLoaded();
+        }).then(() => {
```
larkox, line 236: This is called on component did mount. What happens when the user is still not logged in?

catalintomai:
Looks like the redux piece is missing from the test server? Have you tried building locally?

catalintomai:
Looks like the redux piece is missing from the test server? Have you tried building locally?

components/plugin_marketplace/index.ts
```diff
@@ -7,6 +7,9 @@ import {bindActionCreators, Dispatch, ActionCreatorsMapObject} from 'redux';
 import {GlobalState} from 'mattermost-redux/types/store';
 import {GenericAction, ActionFunc} from 'mattermost-redux/types/actions';
 
+import {setFirstAdminVisitMarketplaceStatus} from 'mattermost-redux/actions/general';
+import {firstAdminVisitMarketplaceStatus as getFirstAdminVisitMarketplaceStatus} from 'mattermost-redux/selectors/entities/general';
```
jfrerich, line 11: Why is `as getFirstAdminVisitMarketPlaceState` necessary?  Can't the `getFirstAdminVisitMarketPlaceState` function be imported by itself and directly in `mapStateToProps`?

jfrerich (approved):
I have one question, but otherwise, LGTM

components/plugin_marketplace/index.ts
catalintomai: It was done since the redux state selector was called `firstAdminVisitMarketPlaceState`, I renamed it to remove the aliasing.

abhijit-singh (commented):
@catalintomai On the test server, I can still see the badge after visiting the Marketplace, not sure if that's intentional for now/covered by another pr or the functionality of hiding it is not working. 

Also, some visual nits (not necessary but it'll be great if you can incorporate these) - 
 
1. For the badge on the hamburger main menu icon, create a border between the red badge and the white menu icon. I tried to do it with the same approach that's used for the user status badge on the avatar(i.e. wrapping the `unread-badge` in a button element. Here's what worked for me:

`<button class="style--none" style="
    position: absolute;
    top: 8px;
    right: 4px;
    height: 12px;
    width: 12px;
    border-radius: 12px;
    background-color: rgb(17, 83, 171);
">
<span class="unread-badge"></span>
</button>`

2. For the highlighted state of the 'Plugin Marketplace' menu item inside the Main Menu, can we make `font-weight: 600;` for the text, and also change the margin for the badge to `margin: 0px 2px 0 40px;`

components/toggle_modal_button_redux/toggle_modal_button_redux.jsx
```diff
@@ -55,12 +58,20 @@ class ModalToggleButtonRedux extends React.PureComponent {
 
         const ariaLabel = formatMessage({id: 'accessibility.button.dialog', defaultMessage: '{dialogName} dialog'}, {dialogName: props.accessibilityLabel});
 
+        let badge = null;
+        if (this.props.modalId === ModalIdentifiers.PLUGIN_MARKETPLACE && !this.props.firstAdminVisitMarketplaceStatus) {
```
hanzei, line 62: I still think we should place this logic further up the stack in https://github.com/mattermost/mattermost-webapp/blob/a5b2f799aba060d88c9458c7b03ecbe127898fb5/components/main_menu/main_menu.jsx#L239 and use a `Prop` like `ShowUnread` in this compontent.

package.json
```diff
@@ -259,4 +259,4 @@
     "build-storybook": "build-storybook -s ./storybook/static",
     "check-types": "tsc"
   }
-}
+}
```
hanzei, line 262: Unrelated change?

components/legacy_sidebar/header/sidebar_header_dropdown_button.jsx
```diff
@@ -62,6 +65,13 @@ export default class SidebarHeaderDropdownButton extends React.PureComponent {
             );
         }
 
+        const isSystemAdmin = isAdmin(this.props.currentUser.roles);
+        if (isSystemAdmin && !this.props.firstAdminVisitMarketplaceStatus) {
+            badge = (
+                <span className={'unread-badge'}/>
+            );
+        }
```
hanzei, line 73: Same comment here. Should we move that logic up the stack?

components/root/root.jsx
catalintomai: removed the statement from root.jsx.

catalintomai:
@abhijit-singh , to get this straight - are we proposing to change the styling of the original menu button? The current style is: 
```
&.style--none {
        background: transparent;
        border: none;
        padding: 0;

        &:focus {
            outline: 0;
            text-decoration: none;
        }

        &.btn {
            padding: 6px 12px;
        }

        &.btn--block {
            text-align: left;
            width: 100%;
        }

        &:hover,
        &:active {
            text-decoration: none;
        }
    }
   ```

File: components/toggle_modal_button_redux/toggle_modal_button_redux.jsx (hunk 4 above)
Comment (catalintomai) [reply components/toggle_modal_button_redux/toggle_modal_button_redux.jsx]: I'm 1/5 on this. On one side I agree that there is some redundancy/ugliness with the `modalId` check, OTOH, the determination logic is at the same place where it is acted upon so it is more readable/clear imo (don't have to fish up-stack for what conditions determined the badge presence).

File: components/root/root.jsx [hunk 7]
```diff
@@ -233,6 +233,7 @@ export default class Root extends React.PureComponent {
             }
             this.onConfigLoaded();
         });
+
```
hanzei, line 236: Nit pick: Unrelated change

components/main_menu/main_menu.jsx
```diff
@@ -343,6 +344,7 @@ class MainMenu extends React.PureComponent {
                             show={!this.props.mobile && this.props.enablePluginMarketplace}
                             dialogType={MarketplaceModal}
                             text={formatMessage({id: 'navbar_dropdown.marketplace', defaultMessage: 'Plugin Marketplace'})}
+                            showUnread={!this.props.firstAdminVisitMarketplaceStatus}
```
hanzei, line 347: Should we move all the logic here and also check if the user is an admin?

hanzei (approved):
There are some nit picks left, but nothing blocking :+1:

components/main_menu/main_menu.jsx
catalintomai: Do we have to though? This option only shows if the user is an admin, no?

abhijit-singh (approved):

DHaussermann (approved):
Tested as part of https://github.com/mattermost/mattermost-server/pull/16992

LGTM!

DHaussermann:
Re tested with latest commits and fixes. 

See comment https://github.com/mattermost/mattermost-server/pull/16992#pullrequestreview-622819564

LGTM!

components/legacy_sidebar/header/dropdown/sidebar_header_dropdown.jsx
```diff
@@ -74,6 +78,8 @@ export default class SidebarHeaderDropdown extends React.PureComponent {
                     teamDisplayName={this.props.teamDisplayName}
                     teamId={this.props.teamId}
                     openModal={this.props.actions.openModal}
+                    getFirstAdminVisitMarketplaceStatus={this.props.actions.getFirstAdminVisitMarketplaceStatus}
+                    showUnread={isAdmin(this.props.currentUser.roles) && !this.props.firstAdminVisitMarketplaceStatus}
```
hmhealey, line 82: Perhaps the SidebarHeaderDropdownButton could get the action and derive showUnread itself since the SidebarHeaderDropdown doesn't seem to care about these other than to just pass them along

components/toggle_modal_button_redux/toggle_modal_button_redux.jsx
```diff
@@ -81,6 +91,7 @@ class ModalToggleButtonRedux extends React.PureComponent {
                 onClick={clickHandler}
             >
                 {children}
+                {badge}
```
hmhealey, line 94: Instead of adding an extra prop to ToggleModalButtonRedux that's specific to this ccase, couldn't the MenuItemToggleModalReduxImpl (now that's a mouthful) pass the badge in as one of this component's children?

packages/mattermost-redux/src/client/client4.ts
```diff
@@ -2309,6 +2314,20 @@ export default class Client4 {
         );
     }
 
+    setFirstAdminVisitMarketplaceStatus = async () => {
+        return this.doFetch<StatusOK>(
+            `${this.getBaseRoute()}/marketplace/first_admin_visit`,
```
hmhealey, line 2319: This obviously isn't something to change at this point for this feature, but this makes me wonder if we could use something like user Preferences but on a system-wide level for this type of random internal state to save us from having to add new endpoints and new data structures whenever we want something like this. We could also use it for some derived state like how we stick a few random things into the client config like the max length of posts and whether or not there are users on the server, neither of which are actual config values.

sass/components/_buttons.scss
```diff
@@ -1,6 +1,15 @@
 @charset 'UTF-8';
 
 button {
+    .unread-badge {
```
hmhealey, line 4: I think this might cause problems in the desktop app since it uses buttons for its sidebar items instead of links. Is there maybe somewhere more specific we could put this?

hmhealey (changes requested):
These are mostly some 0/5 thoughts on things except for the last comment about the `unread-badge` CSS class which I'll double check

sass/components/_buttons.scss
hmhealey: Never mind. We don't use the `unread-badge` name outside of the team sidebar. The CSS class used elsewhere is `badge`. It would be nice to rename this to avoid any potential naming confusion, but I can't think of anything better to call it at the moment.

hmhealey (approved):

components/legacy_sidebar/header/dropdown/sidebar_header_dropdown.jsx
catalintomai: That was the original change - then after addressing this comment: https://github.com/mattermost/mattermost-webapp/pull/7563#discussion_r581923842 - it was moved up the stack - I'm 0/5 on this - I can see both sides.

sass/components/_buttons.scss
DHaussermann: Thanks for raising this as a potential concern @hmhealey This was tested on desktop as well and so far as I can tell it works the same as webapp.

components/toggle_modal_button_redux/toggle_modal_button_redux.jsx
catalintomai: Yes, that works better and cleans-up unneeded parameter passing. I've done the changes but will not check them in right now to not derail the QA'd status. I've opened https://mattermost.atlassian.net/browse/MM-34438 to check it in as a follow-up.

PR #7567, opened 2021-02-22
MM-20371 migrate add_groups_to_team_modal to ts

komik966:
#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/16734


components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
```diff
@@ -129,16 +144,17 @@ export default class AddGroupsToTeamModal extends React.PureComponent {
         this.setState({saving: true});
 
         groupIDs.forEach(async (groupID) => {
-            const {error} = await this.props.actions.linkGroupSyncable(groupID, this.props.currentTeamId, Groups.SYNCABLE_TYPE_TEAM, {auto_add: true});
+            const {error} = await this.props.actions.linkGroupSyncable(groupID, this.props.currentTeamId, Groups.SYNCABLE_TYPE_TEAM, {auto_add: true, scheme_admin: false});
```
komik966, line 147: I'm not sure if it not changes behaviour, but type requires it.
Related issue: https://github.com/mattermost/mattermost-redux/issues/1362

components/add_groups_to_team_modal/add_groups_to_team_modal.test.tsx
```diff
@@ -154,28 +151,26 @@ describe('components/AddGroupsToTeamModal', () => {
...
         function onAdd() {} //eslint-disable-line no-empty-function
+        function onMouseMove() {} //eslint-disable-line no-empty-function
 
-        expect(wrapper.instance().renderOption(option, isSelected, onAdd)).toMatchSnapshot();
+        expect(wrapper.instance().renderOption(option, isSelected, onAdd, onMouseMove)).toMatchSnapshot();
 
         isSelected = true;
-        expect(wrapper.instance().renderOption(option, isSelected, onAdd)).toMatchSnapshot();
-
-        const optionBot = {id: 'id', is_bot: true, last_picture_update: '12345'};
```
komik966, line 170: This test probably was copied from user related modals

components/add_groups_to_team_modal/add_groups_to_team_modal.test.tsx
```diff
@@ -154,28 +151,26 @@ describe('components/AddGroupsToTeamModal', () => {
...
-        const wrapper = shallow(
+        const wrapper = shallow<AddGroupsToTeamModal>(
             <AddGroupsToTeamModal {...baseProps}/>,
         );
 
-        const option = {id: 'id', last_picture_update: '12345', email: 'test@test.com'};
+        const option = {id: 'id_1', label: 'label_1', value: 'value_1'};
         let isSelected = false;
         function onAdd() {} //eslint-disable-line no-empty-function
+        function onMouseMove() {} //eslint-disable-line no-empty-function
```
michelengelen, line 161: Could you refactor all empty-functions to this (or similar):

`const onMouseMove = jest.fn()`

That way we do not need to tell eslint to disable the rule and it is captured by jest (if we need it in a later iteration of the tests).

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
```diff
@@ -162,15 +179,18 @@ export default class AddGroupsToTeamModal extends React.PureComponent {
...
+    // public for tests
+    public handleDelete = (values: GroupValue[]): void => {
         this.setState({values});
     }
 
-    search = (term) => {
+    // public for tests
+    public search = (term: string): void => {
         this.props.actions.setModalSearchTerm(term);
     }
```
michelengelen, line 190: functions like this can be simplified with immediate return:

```suggestion
    public search = (term: string): void => this.props.actions.setModalSearchTerm(term);
```

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
```diff
@@ -109,11 +128,7 @@ export default class AddGroupsToTeamModal extends React.PureComponent {
         });
     }
 
-    handleSubmit = async (e) => {
-        if (e) {
-            e.preventDefault();
```
michelengelen, line 114: `preventDefault` should still be called, but the if condition is not necessary with typescript anymore:

`e?.preventDefault?.();` does the trick

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
michelengelen: that is a good question.

@nevyangelova could you have a look at it, since you are more proficient with redux?

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
```diff
@@ -212,11 +232,12 @@ export default class AddGroupsToTeamModal extends React.PureComponent {
         );
     }
 
-    renderValue(props) {
+    // public for tests
+    public renderValue(props: { data: Value }): string | undefined {
         return props.data.display_name;
     }
```
michelengelen, line 238: this can be simplified as well:

```suggestion
    public renderValue = (props: { data: Value }): string | undefined =>props.data.display_name;
```

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
```diff
@@ -234,21 +255,22 @@ export default class AddGroupsToTeamModal extends React.PureComponent {
 
         let addError = null;
         if (this.state.addError) {
-            addError = (<div className='has-error col-sm-12'><label className='control-label font-weight--normal'>{this.state.addError}</label></div>);
+            addError = (<div className='has-error col-sm-12'><label
+                className='control-label font-weight--normal'>{this.state.addError}</label></div>);
```
michelengelen, line 259: please use proper indentation:

```suggestion
            addError = (
                <div className='has-error col-sm-12'>
                    <label className='control-label font-weight--normal'>
                        {this.state.addError}
                    </label>
                </div>
            );
```

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
```diff
@@ -162,15 +179,18 @@ export default class AddGroupsToTeamModal extends React.PureComponent {
         }
     }
 
-    handleDelete = (values) => {
+    // public for tests
+    public handleDelete = (values: GroupValue[]): void => {
         this.setState({values});
     }
```
michelengelen, line 185:
```suggestion
    // public for tests
    public handleDelete = (values: GroupValue[]): void => this.setState({values});
```

michelengelen (changes requested):
There is nothing breaking here, just some remarks to keep things consistent

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
hanzei: @michelengelen Are there any (Mattermost) code style guidelines that state that is more idiomatic?

michelengelen: No, I had a look around and there are non to be found. It is very much my personal style and IMO it simplifies the structure and increases readability. I don't know if there are any implications on performance (I highly doubt that), so it can be either way.

hanzei: I would love for us to unify the different styles used in the code base. Maybe something like https://developers.mattermost.com/contribute/server/style-guide/ could help with that. Are you aware of any eslint plugins that can enforce this style (or any of the other ones you mentioned in this PR)?

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
komik966: Can you look again? This is probably outdated code.
It seems that `handleSubmit` no longer receives event.  `handleSubmit` is passed to multiselect component, which invokes `handleSubmit` with `Value[]` as argument, not `SyntheticEvent`.
https://github.com/mattermost/mattermost-webapp/blob/3d6ae3cbeff66663b28798e8c04fa22d0e5d6ab4/components/add_groups_to_team_modal/add_groups_to_team_modal.jsx#L289
https://github.com/mattermost/mattermost-webapp/blob/3d6ae3cbeff66663b28798e8c04fa22d0e5d6ab4/components/multiselect/multiselect.tsx#L36
https://github.com/mattermost/mattermost-webapp/blob/3d6ae3cbeff66663b28798e8c04fa22d0e5d6ab4/components/multiselect/multiselect.tsx#L177
https://github.com/mattermost/mattermost-webapp/blob/3d6ae3cbeff66663b28798e8c04fa22d0e5d6ab4/components/multiselect/multiselect.tsx#L220
https://github.com/mattermost/mattermost-webapp/blob/3d6ae3cbeff66663b28798e8c04fa22d0e5d6ab4/components/multiselect/multiselect.tsx#L229-L230

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
komik966: I think that such rules could be automatically checked by eslint and prettier integration. https://prettier.io/docs/en/related-projects.html#eslint-integrations

components/add_groups_to_team_modal/add_groups_to_team_modal.tsx
komik966: this is doable by adding this in eslintrc:
```
"rules": {
    "arrow-body-style": ["error", "as-needed"],
 ```
https://eslint.org/docs/rules/arrow-body-style#as-needed
but in our case this rule will not throw error because we don't have `return` keyword in body

File: components/add_groups_to_team_modal/add_groups_to_team_modal.tsx (hunk 5 above)
Comment (michelengelen) [reply components/add_groups_to_team_modal/add_groups_to_team_modal.tsx:114]: You are absolutely right. It is kind of an anti-pattern though (I think) 🙇🏼‍♂️

File: components/add_groups_to_team_modal/add_groups_to_team_modal.tsx (hunk 7 above)
Comment (michelengelen) [reply components/add_groups_to_team_modal/add_groups_to_team_modal.tsx:259]: They can... we are not using prettier though ... I will take this discussion to the team and see if we can implement something like that

File: components/add_groups_to_team_modal/add_groups_to_team_modal.tsx (hunk 1 above)
Comment (nevyangelova) [reply components/add_groups_to_team_modal/add_groups_to_team_modal.tsx:147]: As far as I can see the scheme_admin prop is required by the server so it can pass validation and it looks like a recent add, do you have any concerns with passing that prop @komik966 ?

**Review by nevyangelova — APPROVED** [review]
LGTM

**Review by michelengelen — APPROVED** [review]
LGTM ... thanks! ✨

File: components/add_groups_to_team_modal/add_groups_to_team_modal.tsx (hunk 1 above)
Comment (komik966) [reply components/add_groups_to_team_modal/add_groups_to_team_modal.tsx:147]: @nevyangelova 
http body before my changes:
```
{ "auto_add": true }
```
http body after my changes:
```
{ "auto_add": true, "scheme_admin": false }
```
When body comes to the server, it goes through this path:

https://github.com/mattermost/mattermost-server/blob/8199e7b5f3f2e8eb8240a09f0fa47ee9d10a206d/api4/group.go#L32-L33

https://github.com/mattermost/mattermost-server/blob/8199e7b5f3f2e8eb8240a09f0fa47ee9d10a206d/api4/group.go#L242

https://github.com/mattermost/mattermost-server/blob/8199e7b5f3f2e8eb8240a09f0fa47ee9d10a206d/model/group_syncable.go#L131

My concern is if server treats `nil` same way as `false`.

jgilliam17 (approved):
Thank you @komik966 
Tested, looks good to merge.
- Verified add group to team modal in Sys. Console > Team configuration.
@michelengelen Can you please help merge. Thanks 🙂

michelengelen:
@jgilliam17 done! 👍🏻

PR #7578, opened 2021-02-23
[MM-33127] Fixing admin console search clear icon

BenCookie95:
#### Summary
Making clear icon visible

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33127

#### Screenshots
Screen Shot 2021-02-22 at 8 07 02 PM

### DIFF
--- a/sass/routes/_admin-console.scss
+++ b/sass/routes/_admin-console.scss
@@ -581,10 +581,12 @@
             }
         }
         .input-clear {
-            color: $white;
-            font-weight: bold;
-            position: absolute;
-            right: 16px
+            .input-clear-x {
+                color: $white;
+                font-weight: bold;
+                position: absolute;
+                right: 16px
+            }
         }
     }
 
--- a/components/custom_status/custom_status.scss
+++ b/components/custom_status/custom_status.scss
@@ -126,7 +126,6 @@
     }
 }
 
-
 .ic-clear-search, .ic-search {
     background-repeat: no-repeat;
     background-position: center;
--- a/components/custom_status/custom_status.scss
+++ b/components/custom_status/custom_status.scss
@@ -115,14 +115,17 @@
         }
     }
 }
-
-.input-clear-x {
-    color: rgba(var(--center-channel-color-rgb), 0.56);
-
-    .icon {
-        font-size: 20px;
+.status-dropdown-menu,
+.StatusModal__input {
+    .input-clear-x {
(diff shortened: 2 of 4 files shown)


sbishel (commented):
Where you able to find a recent change that would have caused this? This would have been a regression. I'm worried this fix would cover up the real issue.

sbishel:
Actually, I think this may be the issue...
Screen Shot 2021-02-22 at 7 33 35 PM

BenCookie95:
> Where you able to find a recent change that would have caused this? This would have been a regression. I'm worried this fix would cover up the real issue.

The issue was due to the new custom status code added here https://github.com/mattermost/mattermost-webapp/pull/7441. They added a style that overrode it inside `components/custom_status/custom_status.scss`. I can also update that stylesheet to be more specific to it's modal so that this doesn't happen again.

The undefined class thing was not the cause of the issue but I can take a look as to why `undefined` it appearing.

BenCookie95:
> Actually, I think this may be the issue...
> Screen Shot 2021-02-22 at 7 33 35 PM

The root of the problem has been fixed and I also fixed that issue with undefined appearing in the class name.

hahmadia (approved):
Looks right to me. Tried it locally and worked.
Due to the recent scoping of the CSS `input-clear-x` class, this looks like the right fix to me.

sass/routes/_admin-console.scss
```diff
@@ -581,10 +581,12 @@
...
-            color: $white;
-            font-weight: bold;
-            position: absolute;
-            right: 16px
+            .input-clear-x {
+                color: $white;
+                font-weight: bold;
+                position: absolute;
+                right: 16px
+            }
```
hahmadia, line 589: I did want to make a quick comment that this is not needed if you wanted to leave it out. Scoping the first CSS like you've done will fix it and you don't need to make this change hence fixing the regression that Scott pointed out without touching this file. I'll leave it up to you as either way works for me.

hahmadia (approved):
Small comment

sass/routes/_admin-console.scss
BenCookie95: I'm ok with leaving this in. It will protect it if the same issue happens again in the future.

sbishel (approved):
Thanks Ben for doing the extra investigation and fix.

lindy65 (approved):
Thanks @BenCookie95 @sbishel - looks good to me - the "X" is fixed/the correct color 👍

PR #7581, opened 2021-02-23
[MM-33107][MM-33312] fix rhs over textbox

Willyfrog:
#### Summary

have the RHS move up in the z-index when expanding

#### Ticket Link

[MM-33107](https://mattermost.atlassian.net/browse/MM-33107)

#### Screenshots
imagen


hahmadia:
Thank you 😍 !

devinbinnie (approved):

sass/layout/_sidebar-right.scss
```diff
@@ -15,10 +15,12 @@
 
     &.move--left {
         @include single-transition(width, 0.25s, ease-in);
+        @include single-transition(z-index, 0s, steps(1, end));
```
deanwhillier, line 18: So unfortunately this new line is overriding the preceding line because, well, CSS. I actually don't love 'convenience functions' like this because while they simplify some of the complexity, the abstraction hides away how CSS works behind the scenes and things don't always work as you might have expected if you aren't aware of how the function is actually applying the CSS. If you inspect the element, you will see that there are multiple `transition` selectors being applied, one for `width` and the other for `z-index` and `z-index` is winning meaning the width animation is not longer being applied.

It totally seems like this is just adding an additional transition to the existing list of defined transitions, while in reality it is actually completely replacing it because of how transitions work. 🤷‍♂️ Unfortunately, the animation is no longer applied and also, while it looks like it's working because there is no currently no animation, if there was the z-index application would actually not be working properly.

What you probably want to do here is replace both of those lines of convenience CSS functions with something like this:

```scss
.sidebar--right {
    @include translateX(100%);
    height: 100%;
    padding: 0;
    position: fixed;
    right: 0;
    width: 400px;
    z-index: 11;

    body.announcement-bar--fixed & {
        height: calc(100% - #{$announcement-bar-height});
    }

    &.move--left {
        transition:
            width 0.25ms 0s ease-in,
            z-index 0.25s 0s step-end;  
    }

    &.sidebar--right--expanded {
        @include box-shadow(0px 8px 24px alpha-color($black, .24));
        z-index: 13;
        
        &.move--left {
            transition:
                width 0.25ms 0s ease-in,
                z-index 0.25s 0s step-start;  
        }

        .sidebar-right__body {
            overflow: hidden;
        }
    }
    ...
}
```

I'm purposely not using the convenience functions here, support for  the `transition` css properly is across the board for us.

This is basically saying that when the animation starts to expand the RHS (after adding the `sidebar--right--expanded` class), the `z-index: 13` style is applied immediately (`step-start`) to put it on top of the centre at the beginning of the transition. When the RHS is collapsed, the default `z-index: 11` is not applied till the end of the animation so that it remains on top of the centre until it is no longer overlapping the centre, at which point the `z-index` is changed back to the default to layer it behind the centre.

Hopefully that make at least some sense.

Disclaimer, I have not tried this locally as my webapp is currently in an unusable state due to some in-progress activity. :P

deanwhillier (changes requested):
See below for some CSS "fun!" 😛

deanwhillier (approved):
LGTM! Hmm, I almost want to refine this a bit more, but that's not part of this ticket, so I'll refrain! :P

devinbinnie (approved):

jgilliam17:
@Willyfrog Looks good when RHS is expanded in any window width.
When RHS is not expanded, and this is probably an edge case that only users with narrow browser window would see, text box remains overlaid on top of RHS hiding any file attachments.

https://user-images.githubusercontent.com/52937121/109212439-86bbbd00-777d-11eb-89ba-02980304ed31.mov

Willyfrog:
> @Willyfrog Looks good when RHS is expanded in any window width.
> When RHS is not expanded, and this is probably an edge case that only users with narrow browser window would see, text box remains overlaid on top of RHS hiding any file attachments.
> Screen.Recording.2021-02-25.at.3.21.36.PM.mov

I'm guessing the input box should resize to match the central channel, right?
also it is present on current daily, so it is a problem that wasn't caused by my fix but the original PR

after further investigation I'm sending this to triage in the team, as I think we need to come up with a solution for the actual problem which is having RHS and center channel intersecting in a weird way in some sizes.

devinbinnie (approved):

deanwhillier (approved):

jgilliam17 (approved):
Thank you @Willyfrog 
Tested, looks good to merge.
- Verified main channel textbox no longer overlaps RHS when expanded/collapsed.

PR #7592, opened 2021-02-24
Fix issues in Custom Status

chetanyakan:
#### Summary
- ~~Fix custom status failing with just the emoji and no message~~
- Fix custom emojis for statuses not loaded on page refresh 
- Fixed the bot showing custom status emoji of integration owner
***
- Fix channel sidebar misalignment
- Fix emoji misaligned with tooltip
- Fix custom status misalign next to username.
- Fix custom status sitting too high in DM channel header
- Fix No hover state on 'Set Status' button in modal

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-33213
- https://mattermost.atlassian.net/browse/MM-33197
- ~~https://mattermost.atlassian.net/browse/MM-33210~~
***
- https://mattermost.atlassian.net/browse/MM-33200
- https://mattermost.atlassian.net/browse/MM-33199
- https://mattermost.atlassian.net/browse/MM-33191
- https://mattermost.atlassian.net/browse/MM-33190
- https://mattermost.atlassian.net/browse/MM-33194
- https://mattermost.atlassian.net/browse/MM-33193

### DIFF
--- a/components/custom_status/custom_status_emoji.tsx
+++ b/components/custom_status/custom_status_emoji.tsx
@@ -1,15 +1,12 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-import React, {useEffect} from 'react';
+import React from 'react';
 import {Tooltip} from 'react-bootstrap';
-import {useDispatch, useSelector} from 'react-redux';
-
-import {getCustomEmojis} from 'mattermost-redux/actions/emojis';
+import {useSelector} from 'react-redux';
 
 import OverlayTrigger from 'components/overlay_trigger';
 import RenderEmoji from 'components/emoji/render_emoji';
 import {makeGetCustomStatus, isCustomStatusEnabled} from 'selectors/views/custom_status';
-import {isCustomEmojiEnabled} from 'selectors/emojis';
 import {GlobalState} from 'types/store';
 import Constants from 'utils/constants';
 
@@ -24,17 +21,8 @@ interface ComponentProps {
 }
 
 const CustomStatusEmoji = (props: ComponentProps) => {
-    const dispatch = useDispatch();
     const getCustomStatus = makeGetCustomStatus();
     const {emojiSize, emojiStyle, spanStyle, showTooltip, tooltipDirection, userID, onClick} = props;
-
-    const customEmojiEnabled = useSelector(isCustomEmojiEnabled);
-    useEffect(() => {
-        if (customEmojiEnabled) {
-            dispatch(getCustomEmojis());
-        }
-    }, [isCustomEmojiEnabled]);
-
     const customStatusEnabled = useSelector(isCustomStatusEnabled);
     const customStatus = useSelector((state: GlobalState) => {
         return getCustomStatus(state, userID);
@@ -92,7 +80,6 @@ CustomStatusEmoji.defaultProps = {
     emojiSize: 16,
     tooltipDirection: 'top',
     showTooltip: false,
-    spanStyle: {},
     emojiStyle: {
         marginLeft: 4,
     },
--- a/components/custom_status/custom_status_emoji.tsx
+++ b/components/custom_status/custom_status_emoji.tsx
@@ -80,6 +80,7 @@ CustomStatusEmoji.defaultProps = {
     emojiSize: 16,
     tooltipDirection: 'top',
     showTooltip: false,
+    spanStyle: {},
     emojiStyle: {
         marginLeft: 4,
     },
--- a/components/custom_status/custom_status_emoji.tsx
+++ b/components/custom_status/custom_status_emoji.tsx
@@ -17,6 +17,7 @@ interface ComponentProps {
     emojiSize?: number;
     showTooltip?: boolean;
     tooltipDirection?: 'top' | 'right' | 'bottom' | 'left';
+    spanStyle?: React.CSSProperties;
     emojiStyle?: React.CSSProperties;
     userID?: string;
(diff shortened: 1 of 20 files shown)


jespino (approved):
Looks good to me.

esethna:
@ogi-m can you help test this?

esethna:
@chetanyakan can you let us know when you're ready for a UX review from @matthewbirtch?

components/custom_status/custom_status_emoji.tsx
```diff
@@ -1,27 +1,40 @@
...
 const CustomStatusEmoji = (props: ComponentProps) => {
+    const dispatch = useDispatch();
     const getCustomStatus = makeGetCustomStatus();
-    const {emojiSize, emojiStyle, showTooltip, tooltipDirection, userID, onClick} = props;
+    const {emojiSize, emojiStyle, spanStyle, showTooltip, tooltipDirection, userID, onClick} = props;
+
+    const customEmojiEnabled = useSelector(isCustomEmojiEnabled);
+    useEffect(() => {
+        if (customEmojiEnabled) {
+            dispatch(getCustomEmojis());
```
hmhealey, line 34: This only gets the first page of custom emojis which doesn't work when there's too many emojis on the server. It also isn't debounced in any way, so we'll be calling it repeatedly for each CustomStatusEmoji mounted at once. There is a getCustomEmojiByName singular action, but I'd still be worried about potentially calling it multiple times if multiple components are mounted at once.

We don't have a great way to handle that sort of loading right now, but I think the best way we can do it is to load the custom emojis when the user profiles are loaded similar to how we call `loadStatusesForProfileLists`(https://github.com/mattermost/mattermost-webapp/blob/master/actions/user_actions.jsx#L44) in a bunch of places. We'll also want to only load the emoji if necessary too.

hmhealey (changes requested):

components/custom_status/custom_status_emoji.tsx
chetanyakan: @hmhealey What if we remove the server version check from `loadMe` in `mattermost-redux`?

Ref - https://github.com/mattermost/mattermost-redux/blob/041dc98731d1317c7d4f8014d79f81d1e9e80663/src/actions/users.ts#L227

chetanyakan:
@esethna Except for the one change suggested by @hmhealey, this should be good for a UX review by @matthewbirtch.

Test server: http://ec2-52-14-2-53.us-east-2.compute.amazonaws.com:8065/signup_user_complete/?id=d9wq9ni8njdk3p6p8t9jwjsahc

matthewbirtch (approved):
Thanks @chetanyakan. I just reviewed and all looks good.

- [MM-33200](https://mattermost.atlassian.net/browse/MM-33200) - fixed
- [MM-33199](https://mattermost.atlassian.net/browse/MM-33199) - fixed
- [MM-33191](https://mattermost.atlassian.net/browse/MM-33191) - fixed
- [MM-33190](https://mattermost.atlassian.net/browse/MM-33190) - fixed
- [MM-33194](https://mattermost.atlassian.net/browse/MM-33194) - fixed

components/custom_status/custom_status_emoji.tsx
hmhealey: That isn't possible. Loading all custom emojis adds a lot of lag for people that have thousands of emojis which is why we switched to loading them on-demand. We went from doing it on page load to loading any unrecognized emojis when we received a batch of posts with them in it. In this case, our best bet is probably to do something similar but doing it when we receive a batch of users with custom statuses enabled.

hmhealey: Since that fix will likely be more involved, we might want to consider moving it out of 5.33 and just merging the other fixes for now. I'm not sure what the full effects of not having the custom emoji for a status might be though.

esethna: I agree. @chetanyakan as suggested here let's pull this specific bug fix from this PR so we can unblock the rest of these fixes. We are aiming to merge this before we cut the RC for v5.33 on wednesday

chetanyakan:
Rebase to HEAD of master

components/custom_status/custom_status_emoji.tsx
chetanyakan: Done @hmhealey @esethna

esethna:
@amyblais lets make sure this gets cherry picked to cloud Mar 10 branch as well included in the v5.33 RC

hmhealey:
@chetanyakan The changes look good, but the snapshot tests in the sidebar are failing now due to the default prop affecting them.

Also, would you mind merging master instead of rebasing? It makes it easier for us to tell what has changed since our last review, and it doesn't matter if it makes the commit history a bit messier since we rebase and squash the branch before merging to master anyway.

chetanyakan:
@hmhealey I've updated the snapshots. I'll keep in mind your point about using merge commits.

esethna (approved):
@ogi-m please help prioritize QA review on this PR

ogi-m (changes requested):
@chetanyakan Bot posts still show the status of the integration owner:
webhook

chetanyakan:
https://mattermost.atlassian.net/browse/MM-33193 is also fixed now.

ogi-m:
Thanks @chetanyakan
- Verified the status is removed from the integration post popover
- "Update your status" alignment looks good and the hover state was added

LGTM 👍🏻

ogi-m (approved):

hmhealey (approved):

PR #7594, opened 2021-02-24
MM-33245 Fix casing on contenthash for CSS

hmhealey:
As Caleb pointed out to me, the casing was incorrect for `contenthash`. I don't know when in the upgrade process that broken, but we went from 0.9.0 to 1.3.5 with that plugin, so it could've been any time.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33245


calebroseland (approved):

gabrieljackson (approved):
Nice catch!

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified CSS file names are listed with an actual hash, no longer listed with [contentHash]

PR #7600, opened 2021-02-26
Fix zoom on pdf previews

jespino:
#### Summary
Fix zoom on pdf previews

#### Ticket Link
[MM-33218](https://mattermost.atlassian.net/browse/MM-33218)

### DIFF
--- a/components/pdf_preview.jsx
+++ b/components/pdf_preview.jsx
@@ -79,6 +79,11 @@ export default class PDFPreview extends React.PureComponent {
         }
         if (this.props.scale !== prevProps.scale) {
             this.pdfPagesRendered = {};
+            if (this.state.success) {
+                for (let i = 0; i < this.state.numPages; i++) {
+                    this.renderPDFPage(i);
+                }
+            }
         }
 
         if (!prevState.success && this.state.success) {


jespino:
I added QA directly because I don't expect much changes from the Dev review :)

calebroseland (approved):

jgilliam17:
@jespino Looks good.
 I have a few questions, possibly out of scope since they also repro on master, may be related in a way to this fix/change
1) When attaching and posting multiple PDF files, open preview, change zoom for one and it applies the same zoom level to all; scroll through all attached PDFs on preview and they will all show that same zoom. I am not sure if this was originally intended.
2) When sufficiently zoomed in to have both vertical and horizontal scroll bars shown, I can only click, hold and drag horizontal scroll bar, but can't do that on the vertical scroll bar.

jespino:
@jgilliam17

about 1. I guess was a pre-existing bug, and is probably easy to fix (I would suggest to create a new ticket for that). Also I think we need to define how it should behave, for example, if I move to second attachment, I expect to reset the zoom, but if I back to the previous one, do I expect to preserve the previous zoom, or to reset it again?

2. The vertical scroll problem was already there before, we have to figure out, how to solve that, it is related to the hover on multi-pdf next/prev arrows. Probably something to evaluate from the UX perspective. I would file a ticket too.

hmhealey (approved):

jespino:
@jgilliam17 are we good to go with this? do you need anything else from this PR or we can defer 1 and 2 to other user stories?

jgilliam17 (approved):
Thanks @jespino and apologize about delay, I wanted to make sure I filed tickets for follow-up  before approving and forgetting 🙂 
Tested, looks good to merge.
- Verified zoom in, out and actual level on PRF preview, single and multi- file attachment.
Follow up issues filed:
https://mattermost.atlassian.net/browse/MM-33370
https://mattermost.atlassian.net/browse/MM-33371

jespino:
Great! thanks @jgilliam17!

PR #7602, opened 2021-02-27
[GH-16739, MM-18954] Converted channel info modal and associated tests to typescript

Crimson-riot:
#### Summary
* Converted channel info modal to typescript and fixed any type check errors as a result of this change. This includes adding additional attributes to the "not found" channel in order to satisfy the type requirements for a channel.
* Updated channel info modal tests to use new types and broke out channel / team mocks into a before each to reduce the amount of repeated code.
* Updated delete channel modal tests to account for type change on the `props` property for a channel. 
* Updated snapshots for channel_info_modal.test.tsx 

#### Ticket Link

Addresses https://github.com/mattermost/mattermost-server/issues/16739

#### Related Pull Requests

- Has redux changes: https://github.com/mattermost/mattermost-redux/pull/1382

#### Screenshots
N/A


components/channel_info_modal/channel_info_modal.test.tsx
```diff
@@ -6,15 +6,56 @@ import {Modal} from 'react-bootstrap';
 import {shallow} from 'enzyme';
 
 import {mountWithIntl} from 'tests/helpers/intl-test-helper';
-import ChannelInfoModal from 'components/channel_info_modal/channel_info_modal.jsx';
+import ChannelInfoModal from 'components/channel_info_modal/channel_info_modal';
+import { Channel } from 'mattermost-redux/src/types/channels';
+import { Team } from 'mattermost-redux/types/teams';
 
 describe('components/ChannelInfoModal', () => {
+    let mockChannel: Channel;
```
hmhealey, line 14: You should be able to use the TestHelper (`utils/test_helper`) to save having to write these fields out. Also, I think it's safe enough to just define the mocks once instead of redefining them for each test.

components/channel_info_modal/channel_info_modal.test.tsx
```diff
@@ -6,15 +6,56 @@ import {Modal} from 'react-bootstrap';
 import {shallow} from 'enzyme';
 
 import {mountWithIntl} from 'tests/helpers/intl-test-helper';
-import ChannelInfoModal from 'components/channel_info_modal/channel_info_modal.jsx';
+import ChannelInfoModal from 'components/channel_info_modal/channel_info_modal';
+import { Channel } from 'mattermost-redux/src/types/channels';
```
hmhealey, line 10: The linting checks for this PR are failing. I think it might be something to do with the spacing of these curly brackets.

components/channel_info_modal/channel_info_modal.tsx
```diff
@@ -52,17 +68,17 @@ export default class ChannelInfoModal extends React.PureComponent {
...
     };
 
-    constructor(props) {
+    constructor(props: Props) {
         super(props);
 
         this.state = {show: true};
 
-        this.getHeaderMarkdownOptions = memoizeResult((channelNamesMap) => (
+        this.getHeaderMarkdownOptions = memoizeResult((channelNamesMap: Function) => (
```
hmhealey, line 76: Instead of declaring this above as private and then defining it in the constructor, it would be better to define it outside of the constructor like  `handleFormattedTextClick`. I'm not entirely sure why it was defined in here to begin with.

components/channel_info_modal/channel_info_modal.tsx
```diff
@@ -90,6 +106,17 @@ export default class ChannelInfoModal extends React.PureComponent {
                 purpose: notFound,
                 header: notFound,
                 id: notFound,
+                team_id: notFound,
```
hmhealey, line 109: Interesting. I didn't know we passed around a fake channel object here. I feel like that could cause other issues, but I'm not too concerned about that right now. Nothing else needs to change here, but I thought I'd mention it.

components/channel_info_modal/index.js
```diff
@@ -7,7 +7,7 @@ import {getCurrentRelativeTeamUrl, getCurrentTeam} from 'mattermost-redux/select
 
 import {getIsRhsOpen} from 'selectors/rhs';
 
-import ChannelInfoModal from './channel_info_modal.jsx';
+import ChannelInfoModal from './channel_info_modal.tsx';
```
hmhealey, line 10: You can actually remove the file extension here entirely and it'll pick up the correct file for you.

hmhealey (changes requested):

components/channel_info_modal/index.js
Crimson-riot: Good call, fixed.

components/channel_info_modal/channel_info_modal.tsx
Crimson-riot: Ah yea that is a good idea, I did that as well as retyping channelNamesMap to use the actual `channelNamesMap` type

components/channel_info_modal/channel_info_modal.test.tsx
Crimson-riot: Thanks for pointing that out! I had not idea TestHelper existed. I had originally put the definitions in a beforeEach because I did not know for sure if any of the tests made manipulations to the mock channel / mock team but agree that the beforeEach can be removed.

components/channel_info_modal/channel_info_modal.test.tsx
Crimson-riot: Thanks for the reminder! I think I fixed all of the lint issues but will circle back to it again if the build fails.

Crimson-riot: Seems like I missed some lint errors again but these should be fixed now.

For the `type-check` stage, the fix for this has been merged. How should I go about upgrading the `mattermost-redux` package in `package-lock.json`? Running `npm upgrade mattermost-redux` did not change the version commit for me.

hmhealey: You can run `npm install mattermost/mattermost-redux#commithash` with the merge commit from your PR into master. NPM really doesn't like working with commit hashes, so it doesn't update automatically unfortunately. We're working on getting away from that since it's kind of annoying like that.

hmhealey:
I'm going to update this branch to fix the `check-deps` CI check. Hopefully that resolves the issue with the `test` as well.

hmhealey:
@Crimson-riot Looks like the tests for that modal are still failing after that update. Are you able to take a look at what's causing that?

Crimson-riot:
> @Crimson-riot Looks like the tests for that modal are still failing after that update. Are you able to take a look at what's causing that?

Sorry about getting to this so late! The tests should be passing now.

hmhealey (approved):
Thanks for fixing that. Looks good to me now!

calebroseland (approved):
LGTM, thanks @Crimson-riot! (and sorry for the delay). BTW, I just updated the branch/conflict because there was a larger migration with mattermost-redux coming into mattermost-webapp.

jgilliam17 (approved):
Thank you @Crimson-riot 
Tested, looks good to merge.
- Verified channel info modal is as before.
@hmhealey Can you please help merge? Thanks. 🙂

hmhealey:
Thanks again for the PR, @Crimson-riot! This'll be in for Mattermost 5.35 releasing mid-May.

PR #7629, opened 2021-03-03
Files search interface

jespino:
#### Summary
Files search interface

#### Ticket Link
[MM-28019](https://mattermost.atlassian.net/browse/MM-28019)


jespino:
@ethervoid looks like this is not getting the feature flag :(

ethervoid:
@jespino updated that Spinwick group. Your server has the FF activated now

matthewbirtch (changes requested):
This is awesome @jespino. I think this will bring a lot of value to our users. Below are the things I found in my first pass that would be good to have tightened up from a UI/UX perspective if we can.
 
### Channel-specific Files list in RHS
1. Did we remove the search from the RHS context from the original design?
2. Looks like there is extra space below `Recent files`. I'm seeing 28px of padding that doesn't need to be there
3. The `Filter` icon button should have the same styles as the other icon buttons in the UI (like mentions, saved, etc). Can we borrow from those components here? The various states (hover, focus, etc) should follow those buttons as well
4. For the filter, if I select `Documents` as a filter, `.txt` files don't seem to be included in that. I think maybe it should be included. What types of files are included in this category?
5. If there are no files in a channel, I think we can remove the 'Recent files' and Filter icon
6. The title for the RHS here should just be "Files | {Channel Name}" rather than 
"Channel Files | {Channel Name}"

### Global Search Input
7. The button style for `Messages` and `Files` isn't quite right:
image
  - The border should be 1px and it seems to be getting a darker right and bottom edge. I think if you apply `border: solid 1px var(--button-bg);` that would resolve the two issues above
  - The button height should be 32px. If you change the padding to `padding: 8px 16px;` and the add `line-height: 1;`
  - The space between the icon and the label within the button need to be increased. If you add `margin-right: 6px;` to the icon within the button that should add enough space.
8. For the initial 'What are you searching for' popover, can we remove 8px of space below the two buttons?
9. Once a selection is made, the tag that gets added to the search input needs a tweak on the `x`
 image
   - can we change the icon's font-size to 12px?
   - can we then add 2px padding on all sides of the icon `<i>` tag?
   - can we also add 2px of padding on the right of the badge itself?
10. For the `Messages` or `Files` tags that get added to the search input, I can't seem to clear them if the RHS is not open. I just seems to expand or collapse the search input when I try to clear the tag.

### Search Results
11. Keywords search doesn't seem to be returning results for file with names that match the keyword. I added a file called 'colors.txt' in a channel and when I search for files using the keyword `colors` it doesn't return any results
image

12. Searching for `pdf` didn't return any file results even through there are PDFs in the Town Square channel.

13. In one case I got `2+` in the `Messages` results count. I'm not sure what `2+` means here. Is it because it appears to be loading? If so, the loading never stopped. For what it's worth, I couldn't reproduce this every time I searched for 'test'.
image

14. The results badge within the Messages/Files tabs are not vertically centered.
  - can we bump the counter badge up 1px within the container?
  - let's change the counter badge to have `padding: 1px 6px;`

15. Are we still returning results based on content within the documents? If so, it didn't seem to be working for me.

jespino:
>     1. Did we remove the search from the RHS context from the original design?
Yes for the first iteration.
 
>     2. Looks like there is extra space below `Recent files`. I'm seeing 28px of padding that doesn't need to be there
Fixed

>     3. The `Filter` icon button should have the same styles as the other icon buttons in the UI (like mentions, saved, etc). Can we borrow from those components here? The various states (hover, focus, etc) should follow those buttons as well
Fixed

>     4. For the filter, if I select `Documents` as a filter, `.txt` files don't seem to be included in that. I think maybe it should be included. What types of files are included in this category?
Added. We have a list of file extensions. 'doc', 'docx', 'odt', 'pdf', 'rtf' and now 'txt'.

>     5. If there are no files in a channel, I think we can remove the 'Recent files' and Filter icon
Done

>     6. The title for the RHS here should just be "Files | {Channel Name}" rather than
>        "Channel Files | {Channel Name}"
Done

> ### Global Search Input
> 
>     1. The button style for `Messages` and `Files` isn't quite right:

Done 

> 
>     1. For the initial 'What are you searching for' popover, can we remove 8px of space below the two buttons?
Done

>     2. Once a selection is made, the tag that gets added to the search input needs a tweak on the `x`
Done

>     3. For the `Messages` or `Files` tags that get added to the search input, I can't seem to clear them if the RHS is not open. I just seems to expand or collapse the search input when I try to clear the tag.
This is the same behavior used for the text. We can not show it when is collapsed, but what is the expected behavior, when it opens again, I show again the badge?

> ### Search Results
> 
>     1. Keywords search doesn't seem to be returning results for file with names that match the keyword. I added a file called 'colors.txt' in a channel and when I search for files using the keyword `colors` it doesn't return any results
>        image
Yes, that is relate to the database search, that doesn't work specially well for the full text search, I'll try to improve that, but is a backend change. If you search the whole filename, the results are going to be there.

>     2. Searching for `pdf` didn't return any file results even through there are PDFs in the Town Square channel.
Same problem here.

>     3. In one case I got `2+` in the `Messages` results count. I'm not sure what `2+` means here. Is it because it appears to be loading? If so, the loading never stopped. For what it's worth, I couldn't reproduce this every time I searched for 'test'.
>        image
I have to review that.

>     4. The results badge within the Messages/Files tabs are not vertically centered.
> 
> 
>     * can we bump the counter badge up 1px within the container?
> 
>     * let's change the counter badge to have `padding: 1px 6px;`
> 
> 
>     1. Are we still returning results based on content within the documents? If so, it didn't seem to be working for me.
Yes, but you need to enable that in the admin console and "re-index" or add new files and the content is only search in the new files (if you are not reindexing that).

jespino:
@matthewbirtch :point_up:

jespino:
The improvements around the search results are added in this PR mattermost/mattermost-server#17112, it is a small one, so I expect to see that merged soon.

matthewbirtch:
> > 3. For the `Messages` or `Files` tags that get added to the search input, I can't seem to clear them if the RHS is not open. It just seems to expand or collapse the search input when I try to clear the tag.
> 
> This is the same behavior used for the text. We can not show it when is collapsed, but what is the expected behavior, when it opens again, I show again the badge?

I just tried this again, and this is what I get (see gif below). I should be able to remove the tag from the input whether the input is expanded, collapsed or whether the RHS is open or closed. To your other point, I think we want to show the tag whether the input is collapsed or not (like it currently does)



> > 1. Are we still returning results based on content within the documents? If so, it didn't seem to be working for me.
> 
> Yes, but you need to enable that in the admin console and "re-index" or add new files and the content is only search in the new files (if you are not reindexing that).

I tried to get this to work in the system console on the cloud test server, but couldn't. Connecting to the Elasticsearch server failed. Not sure what the settings should be though.

> > 1. Keywords search doesn't seem to be returning results for file with names that match the keyword. I added a file called 'colors.txt' in a channel and when I search for files using the keyword `colors` it doesn't return any results
> >    image
> 
> Yes, that is relate to the database search, that doesn't work specially well for the full text search, I'll try to improve that, but is a backend change. If you search the whole filename, the results are going to be there.

Ok, this feels pretty important to be able to search a partial filename rather than the whole filename. I hope we can get this in this feature.

jespino:
> I just tried this again, and this is what I get (see gif below). I should be able to remove the tag from the input whether the input is expanded, collapsed or whether the RHS is open or closed. To your other point, I think we want to show the tag whether the input is collapsed or not (like it currently does)

This is funny, but is hard to fix, I'll fix it tomorrow.

> I tried to get this to work in the system console on the cloud test server, but couldn't. Connecting to the Elasticsearch server failed. Not sure what the settings should be though.
I have to add the setting in the admin console. After that, you don't need elasticsearch, you will enable that, and you will start indexing files by content, but only the new ones.

> Ok, this feels pretty important to be able to search a partial filename rather than the whole filename. I hope we can get this in this feature.

I already created other PR fixing that, actually this works fine in MySQL, in postgres I have to do some small tricks.

@matthewbirtch :point_up:

matthewbirtch:
> > I just tried this again, and this is what I get (see gif below). I should be able to remove the tag from the input whether the input is expanded, collapsed or whether the RHS is open or closed. To your other point, I think we want to show the tag whether the input is collapsed or not (like it currently does)
> 
> This is funny, but is hard to fix, I'll fix it tomorrow.

ok, thanks @jespino. It think this is the only remaining issue. All the other issues are resolved now. I'll approve once the above is resolved

jespino:
@matthewbirtch fixed the badge problem, and the server side problem is fixed in the other PR as soon is merged (probably tomorrow morning) we can start with code review, and then QA review :)

jespino:
@matthewbirtch I also added the channel name badge in the file search results (it is limited in size and show "Directe Message" and "Group Messages" for that cases without being more speicifc).

matthewbirtch:
> @matthewbirtch I also added the channel name badge in the file search results (it is limited in size and show "Directe Message" and "Group Messages" for that cases without being more speicifc).

I don't see this in the test server, but it looks like it failed

matthewbirtch:
@jespino a few other notes I just found:

1. When I switch between the `Messages` and `Files` tabs in the RHS, the height of the tabs area changes and causes the tabs to jump around
2. For the main search input, if I choose either `Messages` or `Files` the badge gets added, and then if I click away from the search input, I think maybe we can just remove the badges from the input.
3. Can we update the spacing above the search results to reduce the padding to 16px instead of 28px?
Screen Shot 2021-03-12 at 10 15 44 AM

jespino:
@matthewbirtch 1 and 3 fixed, the 2 is not consistent with the rest of the experience, if I enter in a search and I close the search, the text is not removed, I don't expect the badge to be remove. Probably we should think how is the whole experience around that badges.

jespino:
@esethna I added the admin console setting for content extraction, would be great to have a review of the texts, and where it is placed.

I called it "Enable Document Content Extraction" or something like that, but probably should be something like "Enable Search Files by Content" or something like that. And the description can be better for sure.

matthewbirtch:
> @matthewbirtch 1 and 3 fixed, the 2 is not consistent with the rest of the experience, if I enter in a search and I close the search, the text is not removed, I don't expect the badge to be remove. Probably we should think how is the whole experience around that badges.

Fair point for #2. We can leave that out for now, but I still think we should be able to remove the badge when the search input is in it's collapsed state

matthewbirtch:
@jespino I'm still not seeing the channel badge in global file search
image

jespino:
@matthewbirtch sorry, yes, it depends on mattermost-server changes that are already reviewed from the code perspective, but are waiting for QA, as soon as that changes are merges that will be there.

matthewbirtch:
> @matthewbirtch sorry, yes, it depends on mattermost-server changes that are already reviewed from the code perspective, but are waiting for QA, as soon as that changes are merges that will be there.

Ok, let me know when that's merged and I can review again.

esethna:
@jespino @matthewbirtch I think the server PR is merged now?

matthewbirtch:
> @jespino @matthewbirtch I think the server PR is merged now?

If so, we should be good to go. How can I see them? Does the test server need to be updated?

calebroseland:
(Current)
CleanShot 2021-03-16 at 16 11 00@2x

@jespino @matthewbirtch what do you guys think about this? 

I feel there should be a dot to separate the channel from the filesize, and possibly a touch smaller spacing (`8px` -> `6px`), and possibly a bump from `10px` -> `11px` on the channel name font size. (just an opinion)

CleanShot 2021-03-16 at 16 57 58@2x



Also, it looks like the size of generated thumbnails (`40x`) doesn't match the icons (`48x`).

CleanShot 2021-03-16 at 16 18 05@2x

calebroseland:
Using channel files, filtering for a type where none exists can cause a flow that kicks you out and makes you start over.

CleanShot 2021-03-16 at 17 20 11

matthewbirtch:
thanks for reviewing @calebroseland! Good eye

> 1a. I feel there should be a dot to separate the channel from the filesize, and possibly a touch smaller spacing (`8px` -> `6px`), and the channel name seems a bit off (font size/color/uppercase).

I don't think we need the dot separator between the badge and the filesize since the badge acts as a separator, but we could do the following to help:
- adjust the margin on the right of the badge to `margin: 0 8px 0 0;`
- as per your suggestion: reduce the margin to the right of the filesize to `margin-right: 6px;`

> 1b. Also, it looks like the size of generated thumbnails (`40x`) doesn't match the icons (`48x`).

Agree this looks a bit odd, but I don't think it's due to thumbnail size. The thumbnail size is correct at 40px. However I'm seeing more descrepencies with the styling of the attachment cards that display in channel. 

<table>
<tr>
<td>file in RHS</td>
<td>file in channel</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

When I compare these two, I notice a few things:
- the left padding on the RHS version needs to be increased
- the font size on the content in the RHS version of the attachment card needs to be changed to 12px
- the margin to the right of the image thumbnail
- The position of the right side actions is much closer to the right side in the RHS version. Can we match the distance from the right edge of the card that we have in the channel attachment card?
- We should add the tooltips for `Download` and `More Actions` for the icon buttons on the RHS file card

Sorry I didn't notice this before @jespino. Comparing the two things now, we should make sure the cards are structured consistently and displayed consistently

> 2. Using channel files, filtering for a type where none exists can cause a flow that kicks you out and makes you start over.

Good catch here too, Caleb. The filter control shouldn't go away here. The only time the filter control should be hidden here is if the channel doesn't have any files at all. Otherwise, we need to keep it visible so you can get out of the case Caleb mentioned here.

jespino:
@calebroseland @matthewbirtch PTAL

matthewbirtch (changes requested):
Thanks @jespino. Looks like most of the tweaks are done. Just a couple other small details:

1. The icons and thumbnails and the text within the cards don't line align for cards with image thumbnails vs. other filetypes (see example below). Looks like there might be extra padding to the left of the file icon that's causing this. 
Screen Shot 2021-03-17 at 9 39 14 AM

2. For the hover action icons on the cards, can we reduce the space between the two icons? It should be 8px of space between the two icons, but it looks more like 16. But keep the last icon the same distance from the card's right edge that you have it now, that looks good.
image

3. For the time display, can we remove the `0` if the hour in the time is not 2 digits? e.g. `04:53` --> `4:53`)

components/file_search_results/file_search_result_item.scss
```diff
@@ -0,0 +1,105 @@
...
+                margin-right: 6px;
+            }
+        }
+    }
+    .post-image {
+        min-width: 4rem;
+    }
+    .action-icon {
+        @include single-transition(opacity, .15s, ease);
+        @include border-radius(4px)
```
calebroseland, line 78:
```suggestion
        @include border-radius(4px);
```

matthewbirtch (approved):
Looks good to me @jespino

esethna:
@matthewbirtch @jespino a few things from my end:

1. When files are selected in the search pop-over, the help text on the search options pop-over should say "Files" instead of "Messages" in all the options:
image
2. Also in the above screen "Options" should be lower case
3. The "Copy link" option on a file search result card has incorrect link formatting (it's missing the server url in the link). Example:
`pr7629/pl/a5fajmbpjjy1xkdpi4ros3d3uy`
4. The menu that launches from the ... on the file search result card should not disappear when you mouse off the card
image
5. I get an infinite loader on various different searches intermittently (doesn't repro consistently though):
image
image

components/file_search_results/file_search_result_item.scss
```diff
@@ -0,0 +1,105 @@
...
+        font-weight: 600;
+        font-size: 12px;
+        .fileDataName {
+            overflow: hidden;
+            text-overflow: ellipsis;
+            margin-bottom: 3px;
+        }
+        .fileMetadata {
+            font-weight: 400;
+            opacity: 0.56;
```
calebroseland, line 65:
```suggestion
```

components/file_search_results/file_search_result_item.scss
```diff
@@ -0,0 +1,105 @@
...
+            display: flex;
+            flex-direction: row;
+            >span {
+                margin-right: 6px;
+            }
```
calebroseland, line 70:
```suggestion
            > :not(.Badge) {
                opacity: 0.56;
            }
```

components/file_search_results/file_search_result_item.scss
```diff
@@ -0,0 +1,108 @@
...
+        background-color: var(--center-channel-bg);
+        border: 1px solid rgba(var(--center-channel-color-rgb), 0.16);
+    }
+    .Badge {
+        margin: 0 8px 0 0;
+        .file-search-channel-name {
+            max-width: 100px;
+            overflow: hidden;
+            text-overflow: ellipsis;
+        }
```
calebroseland, line 28:
```suggestion
            color: rgba(var(--center-channel-color-rgb), 0.72);
        }
```

components/file_search_results/file_search_result_item.tsx
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {FormattedDate, FormattedTime} from 'react-intl';
```
calebroseland, line 5:
```suggestion
```

components/file_search_results/file_search_result_item.tsx
```diff
@@ -0,0 +1,127 @@
...
+import {Tooltip} from 'react-bootstrap';
+
+import {FileInfo} from 'mattermost-redux/types/files';
+
+import {fileSizeToString, copyToClipboard, localizeMessage} from 'utils/utils';
+import {isSameYear} from 'utils/datetime';
```
calebroseland, line 11:
```suggestion
```

jespino:
@esethna 1, 2, 3 and 4 fixed. The spinner problem is not reproducing for me locally, lets wait until the pass on QA.

esethna (approved):
Approving from the PM perspective, will need QA from @ogi-m 

Note the "copy link" button should close the menu (same as copying a permalink to a post in the message list) @jespino

esethna:
@deanwhillier @lieut-data adding you guys for dev review. Feel free to reassign to others from your team if needed.

actions/views/rhs.ts
```diff
@@ -113,13 +128,28 @@ export function performSearch(terms: string, isMentionSearch?: boolean) {
         const teamId = getCurrentTeamId(getState());
         const config = getConfig(getState());
         const viewArchivedChannels = config.ExperimentalViewArchivedChannels === 'true';
+        const extensionsFilters = getFilesSearchExtFilter(getState() as GlobalState);
+
+        const extensions = extensionsFilters.length > 0 ? ' ' + extensionsFilters.map((ext) => `ext:${ext}`).join(' ') : '';
+        const termsWithExtensionsFilteres = terms + extensions;
```
lieut-data, line 134: Would it make sense to write this as:
```suggestion
        const extensions = extensionsFilters?.map((ext) => `ext:${ext}`).join(' ');
        const termsWithExtensionsFilteres = [terms, extensions].filter(x => x?.length > 0).join(' ');
```

Also, I note the typo: `Filteres` should be `Filters`.

actions/views/rhs.ts
```diff
@@ -251,6 +281,46 @@ export function showPinnedPosts(channelId?: string) {
     };
 }
 
+export function showChannelFiles(channelId?: string) {
```
lieut-data, line 284: Is it feasible to make `channelId` required?

actions/views/rhs.ts
```diff
@@ -251,6 +281,46 @@ export function showPinnedPosts(channelId?: string) {
...
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState();
+        const currentChannelId = getCurrentChannelId(state);
+        const teamId = getCurrentTeamId(state);
+
+        dispatch(batchActions([
```
lieut-data, line 290: Can we eliminate `batchActions` here given there's just a single one?

actions/views/rhs.ts
```diff
@@ -251,6 +281,46 @@ export function showPinnedPosts(channelId?: string) {
...
+            data = results[1].data;
+        }
+
+        dispatch(batchActions([
+            {
+                type: SearchTypes.RECEIVED_SEARCH_POSTS,
+                data,
+            },
+            {
+                type: SearchTypes.RECEIVED_SEARCH_TERM,
```
lieut-data, line 311: The use of `RECEIVED_*` between these two cases seems unexpected -- I'm imagining the former contains the results of a query to the server, whereas the latter seems like purely local state. Should the action type be renamed to reflect this?

components/channel_header/channel_header.js
```diff
@@ -205,6 +207,15 @@ class ChannelHeader extends React.PureComponent {
         }
     };
 
+    showChannelFiles = (e) => {
+        e.preventDefault();
```
lieut-data, line 211: Is this necessary? Presumably this is just a pure callback and not a link navigation event we need to cancel or a form we shouldn't submit?

components/search/search.tsx
```diff
@@ -222,12 +228,47 @@ const Search: React.FC<Props> = (props: Props): JSX.Element => {
...
+        actions.updateSearchType('');
     };
 
     const handleShrink = (): void => {
         props.actions.setRhsExpanded(false);
     };
 
+    const handleSetSearchFilter = (filterType: SearchFilterType): void => {
+        switch (filterType) {
+        case 'documents':
```
lieut-data, line 240: Is there, by chance, an existing mime type library that we could use to manage this mapping?

reducers/views/rhs.js
```diff
@@ -40,6 +40,17 @@ function selectedPostFocussedAt(state = 0, action) {
     }
 }
 
+// selectedPostFocussedAt keeps track of the last time a post was selected, whether or not it
```
lieut-data, line 43: Can you update this comment?

reducers/views/rhs.js
```diff
@@ -113,6 +124,26 @@ function searchTerms(state = '', action) {
     }
 }
 
+function searchType(state = '', action) {
```
lieut-data, line 127: Can you elaborate on the difference between the search type and the initial search type? Wondering if we can simplify this paradigm somehow.

components/search_results_item/search_results_item.jsx
```diff
@@ -40,6 +42,11 @@ class SearchResultsItem extends React.PureComponent {
         */
         post: PropTypes.object,
 
+        /**
```
lieut-data, line 45: nit: alignment relative to other comments, but I'm wondering if we can just omit these semi-useless comments.

components/search_results_item/search_results_item.jsx
```diff
@@ -222,8 +230,44 @@ class SearchResultsItem extends React.PureComponent {
...
+            return (
+                <div
+                    data-testid='search-item-container'
+                    className='search-item__container'
+                >
+                    <FileSearchResultItem
+                        fileInfo={this.props.fileInfo}
+                        teamName={this.props.currentTeamName}
+                        channelDisplayName={channelName}
+                        onClick={() => this.setState({showPreview: this.props.fileInfo.postId})}
```
lieut-data, line 259: `showPreview` sounds like it's a boolean, but we're actually storing a `postId` here. Can we pick another name?

components/search_results_item/index.js
```diff
@@ -30,28 +30,33 @@ function mapStateToProps() {
...
 
     return (state, ownProps) => {
-        const config = getConfig(state);
-        const preferences = getMyPreferences(state);
-        const enablePostUsernameOverride = config.EnablePostUsernameOverride === 'true';
         const {post} = ownProps;
-        const user = getUser(state, post.user_id);
-        const channel = getChannel(state, post.channel_id) || {delete_at: 0};
-        const directTeammate = getDirectTeammate(state, channel.id);
+        if (post) {
```
lieut-data, line 34: Having wildly divergent props based on the use case smells wrong here -- can we break up search results item into two connected components, and render them from a parent that does the right thing instead?

components/search_results/index.tsx
```diff
@@ -48,24 +55,65 @@ function makeMapStateToProps() {
                 if (channel && channel.delete_at !== 0 && !viewArchivedChannels) {
                     return;
                 }
+                channels[post.channel_id] = channel;
 
                 posts.push(post);
             });
         }
 
+        const newFilesResults = getSearchFilesResults(state);
```
lieut-data, line 64: I'm worried about React reconciliations here given that we'll be creating a new object each time as a prop, forcing this component to re-render for /every/ state change.

Can we explore managing these more discretely via reducers instead?

lieut-data (dismissed):
Thanks, @jespino! A handful of comments below.

I'm also worried about the size of this PR (#files / #incremental commits) going to straight into master. The chance of regressions with the scope here is high, but the ability to revert will be impacted by the wide range of files being touched.

Is there an opportunity to break this up more discretely?

actions/views/rhs.ts
jespino: I think is harder to undrestand, I going to change it a bit to make it more explicit.

actions/views/rhs.ts
jespino: This was using the same approach already there for other like the `showPinnedPosts`, but yes, it can be changed.

actions/views/rhs.ts
jespino: Nope as this PR, this is already complex enough and we are talking about a lot of places where it should be changed, for example for others like `showPinnedPosts`

components/channel_header/channel_header.js
jespino: :+1:

components/search/search.tsx
jespino: This is a mapping between our set of filter types and file extensions, we could find something more "generic" but I think that is good enough for now.

reducers/views/rhs.js
jespino: The initialSearchType was something that was never used. I removed it from the code. Thanks for catch that.

reducers/views/rhs.js
jespino: Done

components/channel_header/channel_header.js
jespino: Done

actions/views/rhs.ts
jespino: Done

actions/views/rhs.ts
jespino: Done

actions/views/rhs.ts
jespino: Changed with something else

components/search_results_item/search_results_item.jsx
jespino: I would love to do that, but as a separate PR, this PR is already complex enough to add cosmetic changes.

components/search_results_item/search_results_item.jsx
jespino: Actually it can be boolean, I'll change it.

components/search_results/index.tsx
jespino: This is based on the pre-existing code related to posts. I think is not needed to change it as part of this PR.

components/search_results_item/index.js
jespino: Done

components/search_results_item/search_results_item.jsx
jespino: Done

jespino:
@lieut-data PTAL

jespino:
@lieut-data About splitting it, it is going to be complicated, I already splitted this in other PRs that are already merged. Probably we can split that in multiple PRs, but is going to generate more problems than solutions now.

esethna:
@ogi-m when QAing this can we be sure to test:

1) Feature flag off state to verify no regressions
2) Accessibility of the search pop-over options with the keyboard only (and also with a screen reader) (ie buttons in the pop-over for files, messages, and search modifiers in, on, from, etc)

components/admin_console/admin_definition.jsx
```diff
@@ -957,6 +957,31 @@ const AdminDefinition = {
                         onConfigSave: (displayVal) => displayVal * MEBIBYTE,
                         isDisabled: it.not(it.userHasWritePermissionOnResource('environment')),
                     },
+                    {
+                        type: Constants.SettingsTypes.TYPE_BOOL,
+                        key: 'FileSettings.ExtractContent',
+                        label: t('admin.image.extractContentTitle'),
+                        label_default: 'Enable document search by content:',
+                        help_text: t('admin.image.extractContentDescription'),
+                        help_text_default: 'When enabled, supported document types are searchable by their content. Search results for existing documents may be incomplete until a data migration is executed.',
```
esethna, line 966: Please link the words: "until a data migration is executed" to https://www.mattermost.com/file-content-extraction

esethna: @jespino ^

lieut-data:
Thanks, @jespino! On the size of the PR, can I ask you to sync with @amyblais and @jwilander on the potential impact to the cloud release cadence here? My concern with all changes of this size is the difficulty in reverting them given the likelihood of later, unrelated changes piling on top. There have been a few "long running" feature branches being merged to `master` lately that present this risk, and I want to make sure we're aligned on the potential risks there.

I'll take a closer look tomorrow at the new changes to complete my portion of the review.

jespino:
@amyblais @jwilander I can explain the size and the risk of this PR. The amount of different files touched here have 2 reasons, 1 is because redux store makes a lot of changes in a lot of different places, so makes it more "scattered", but at the end of the day are simple changes in a lot of places that are specifically used in a few places only. The second reason is the addition of new tests that is creating new files and new snapshots.

Other important thing is, this feature is focused in very specific places and most of the code is behind a feature flag, my expectation is to merge this in ASAP and use the feature flag to get this tested in community daily during some time. I wouldn't expect any need of a rollback because the feature flag. The QA pass of the PR is going to be focused on not introducing regressions when the flag is disabled, the QA phase when it is in master is going to be focus on finding problems when the feature flag is enabled.

I think the risk here is low specially because the locality of the feature and the feature flag disable it almost entirely.

components/admin_console/admin_definition.jsx
jespino: Done :)

amyblais:
@lieut-data @jespino I don't know when you had planned to merge this PR, but below are my thoughts (if the feature is risky, though it sounds like it's mostly behind a feature flag):

1. If this gets merged to master today, then this will soak in community-daily for a few days before we merge the master branch into cloud branch on Monday (on March 29th, for April 7th cloud release).
2. If this is **not** ready to merge today, then I would recommend waiting merging this PR until **after** we've merged the master branch into the cloud branch on Monday (March 29th) so that there is additional soak time for this feature on community-daily before this goes to a cloud release.

lieut-data:
Thanks, @amyblais! Just to be clear, while this feature is behind a feature flag, the refactoring to support this feature is not (nor can it realistically expect to be). Thanks for sharing the dates in question -- that's really useful context to help drive merging timeframes :)

jespino:
@lieut-data I don't think this is adding any sustancial refactor in any place. It is mainly new code, and small localized changes here and there.

lieut-data:
You're right, @jespino -- refactoring really wasn't the right word to use in this case. I think the "small localized changes" to support the core feature are my focus in this case. Thanks for helping loop in @amyblais so we can align around cloud release, and I'll continue my review (alongside @deanwhillier?).

jespino:
@lieut-data from my understanding of the feature flags, we have to be sure about not breaking the thing without the feature flag, but if something is broken when the feature flag is enable, doesn't matter, doesn't matter for the cloud or for the release, we only care about not breaking things behind the feature flag to remove that feature flag.

jespino:
@esethna be aware of the timing restrictions to make this into the cloud release. If you want to have this merged for the next release you will have to push some people ;)

esethna:
Chatted with @jespino, we are going to merge this AFTER the March 29th cloud release branching. We are aiming to ship this in the April 21st cloud release after it has some soak time on community daily cc// @amyblais

esethna:
Removing Dean and Jesse for dev review as they are both out this week and we want to merge asap

@ogi-m can you help start a QA pass on this asap?

zefhemel (approved):

ogi-m:
Opened [a ticket](https://mattermost.atlassian.net/browse/MM-34652) for a couple of accessibility issues.
@jespino can we address them as part of this PR? 

cc @esethna

jespino:
@ogi-m Because the problem is behind the feature flag I would prefer to fix it after merge.

ogi-m:
That's fine with me. @esethna can we fix this post merge?

esethna:
@ogi-m @jespino that's fine but we need to make sure this feature is accessible before it ships. ie submit the PR to fix the accessibility issues before the feature ships in cloud on April 21

calebroseland:
For some reason `/api/v4/teams/{teamId}/files/search` is returning 501 on https://mattermost-webapp-pr-7629.test.mattermost.cloud/. It seems to be working locally, but using the test server isn't working for me at the moment.

components/file_search_results/file_search_result_item.tsx
```diff
@@ -0,0 +1,168 @@
...
+                        overlay={
+                            <Tooltip id='file-name__tooltip'>
+                                {localizeMessage('file_search_result_item.more_actions', 'More Actions')}
+                            </Tooltip>
+                        }
+                    >
+                        <div onClick={this.stopPropagation}>
+                            <MenuWrapper
+                                onToggle={this.keepOpen}
+                            >
```
calebroseland, line 118: Could be simplified, `MenuWrapper` has a built-in `stopPropagationOnToggle` (minor; maybe for later)

components/admin_console/admin_definition.jsx
```diff
@@ -975,6 +975,32 @@ const AdminDefinition = {
...
+                    {
+                        type: Constants.SettingsTypes.TYPE_BOOL,
+                        key: 'FileSettings.ExtractContent',
+                        label: t('admin.image.extractContentTitle'),
+                        label_default: 'Enable document search by content:',
+                        help_text: t('admin.image.extractContentDescription'),
+                        help_text_markdown: true,
+                        help_text_default: 'When enabled, supported document types are searchable by their content. Search results for existing documents may be incomplete [until a data migration is executed](!https://www.mattermost.com/file-content-extraction).',
+                        isDisabled: it.any(
+                            it.not(it.userHasWritePermissionOnResource('environment')),
```
calebroseland, line 987: Heads up that the keys here were changed and moved to constants. (Happened over on the other feat. branch too 😛)

```suggestion
                            it.not(it.userHasWritePermissionOnResource(RESOURCE_KEYS.ENVIRONMENT.FILE_STORAGE)),
```

CleanShot 2021-04-08 at 20 46 33@2x

components/admin_console/admin_definition.jsx
```diff
@@ -975,6 +975,32 @@ const AdminDefinition = {
...
+                    },
+                    {
+                        type: Constants.SettingsTypes.TYPE_BOOL,
+                        key: 'FileSettings.ArchiveRecursion',
+                        label: t('admin.image.archiveRecursionTitle'),
+                        label_default: 'Enable searching content of documents within ZIP files:',
+                        help_text: t('admin.image.archiveRecursionDescription'),
+                        help_text_default: 'When enabled, content of documents within ZIP files will be returned in search results. This may have an impact on server performance for large files. ',
+                        isDisabled: it.any(
+                            it.not(it.userHasWritePermissionOnResource('environment')),
```
calebroseland, line 999: here too
```suggestion
                            it.not(it.userHasWritePermissionOnResource(RESOURCE_KEYS.ENVIRONMENT.FILE_STORAGE)),
```

calebroseland (changes requested):
LGTM after a 2nd pass, only one blocker in admin definition, the rest should be follow up items but documenting here anyway.

components/file_search_results/file_search_result_item.tsx
```diff
@@ -0,0 +1,168 @@
...
+                            />
+                        </div>
+                    </div>
+                    <OverlayTrigger
+                        delayShow={1000}
+                        placement='top'
+                        overlay={
+                            <Tooltip id='file-name__tooltip'>
+                                {localizeMessage('file_search_result_item.more_actions', 'More Actions')}
+                            </Tooltip>
```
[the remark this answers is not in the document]
jespino: https://github.com/mattermost/mattermost-webapp/pull/7629#pullrequestreview-615807833 <-- I already change it to have the current behavior :)

components/admin_console/admin_definition.jsx
jespino: Done

components/admin_console/admin_definition.jsx
jespino: Done

components/file_search_results/file_search_result_item.tsx
jespino: Done

ogi-m (approved):
LGTM!
- Cypress tests passing with feature flag off
- Manual smoke test with feature flag off found no issues

components/file_search_results/file_search_result_item.tsx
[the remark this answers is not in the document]
calebroseland: Closing the menu part is right now, yes :). This is about the More Actions tooltip being stuck open afterwards.

calebroseland (approved):

components/file_search_results/file_search_result_item.tsx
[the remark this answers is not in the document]
jespino: ohh, ok I think we can file a ticket for that

PR #7632, opened 2021-03-03
Cypress/E2E: Fix timezone display mode spec

josephbaylon:
#### Summary
- Fixed to reflect the changes in dropdown implementation of change timezone
- Also added reset to opposite for each case to make sure setting the timezone is correct

#### Ticket Link
NA

Screen Shot 2021-03-03 at 1 12 02 PM


prapti (approved):
Looks good to me. Ran tests locally and they passed.
Thanks Joseph!

furqanmlk (approved):
Thanks @josephbaylon 
All looks good and passing 👍

PR #7638, opened 2021-03-04
Fix custom emojis for statuses not loaded on page refresh

chetanyakan:
#### Summary
- Fetch custom emoji for custom status only if not loaded.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-33210

#### Related Pull Requests
- https://github.com/mattermost/mattermost-webapp/pull/7592

Refer the conversation here: https://github.com/mattermost/mattermost-webapp/pull/7592#discussion_r583879528

```release-note
Fix custom emojis not loaded on page refresh in custom status.
```


michelengelen (approved):
LGTM

hmhealey (commented):
How does this behave if you have multiple `CustomStatusEmoji` mounted for the same user? Does it send out multiple requests at once for the same emoji? I'd be worried about that with loading them in the component.

michelengelen:
> How does this behave if you have multiple `CustomStatusEmoji` mounted for the same user? Does it send out multiple requests at once for the same emoji? I'd be worried about that with loading them in the component.

Is that even possible?

hmhealey:
Yeah, if you've made multiple posts in a channel, this component appears next to each of them in the channel, so theoretically it could send out multiple requests as long as those aren't deduplicated somehow.

michelengelen:
> Yeah, if you've made multiple posts in a channel, this component appears next to each of them in the channel, so theoretically it could send out multiple requests as long as those aren't deduplicated somehow.

In theory ... yes, you are right!

although it only dispatches an action... should'nt redux handle this in case of duplicate requests?

hmhealey:
> although it only dispatches an action... should'nt redux handle this in case of duplicate requests?

Nope, unfortunately not. A few routes used to have that logic, but they were always one-off things. I think that those might have been removed when switching to redux too. That is something that would be very useful though.

hmhealey:
The way I suggested doing it in the previous PR (loading emojis for statuses when receiving user profiles) should still work though.

chetanyakan:
> The way I suggested doing it in the previous PR (loading emojis for statuses when receiving user profiles) should still work though.

@hmhealey Apart from the users visible in LHS (for DMs and Group chats), the other user profiles are fetched on a need-to-know basis. We keep getting new user profiles on switching channels or even scrolling up in a channel and also, the users' status emojis also keep changing.

I can't find a single place where a `loadCustomEmojisForUsers` common API can be called to effectively ensure all custom emojis are always loaded.

Can you elaborate on how you were suggesting to solve this?

hmhealey:
Unfortunately, it's not as easy as I originally thought since we load profiles/statuses in a few different ways which makes this tougher. I was hoping it was more like posts where every new post goes through one function...

So the first place is loading profiles in bulk for certain components. I think if you look at where `loadStatusesForProfilesList` is used and call `loadCustomEmojisForUsers` at the same time, that should cover most cases. There'll be changes to three components (the More DMs modal and channel/team members modals) and then adding to a few functions in status_actions.jsx. That should cover all cases where a component is mounted and needs some stuff.

Next, we'd need to handle profiles loaded from posts. Thankfully, those are all done in one place like I mentioned, so you just need to add it to [`getProfilesAndStatusesForPosts`](https://github.com/mattermost/mattermost-webapp/blob/cb3c40bb0a44199319d3309d71d1cfcc9e114e09/packages/mattermost-redux/src/actions/posts.ts#L915). That could be done in the same call where custom emojis are loaded for posts, but I don't think that's actually necessary since posts come with their custom emojis included now.

Lastly, we'd need to handle the websocket event for a user being updated. That's another simpler one though since it's just [`handleUserUpdated`](https://github.com/mattermost/mattermost-webapp/blob/master/actions/websocket_actions.jsx#L952) though.

I think that's all of them. We definitely need to look into a better way of doing this sort of incidental loading logic since it's obviously not great how we do it currently. We got away with a lot of that previously be passing down a lot more data with post requests (post metadata), but we don't have this for other objects like users, channels, etc, so it makes it tough to add additional info that's not directly in the object

esethna:
@chetanyakan friendly ping on this PR, are next steps clear or should we sync up on it?

esethna:
@amyblais if this gets completed in time I'd like to cherry pick it to v5.35

esethna:
@hmhealey can you help rereview when you get a chance

michelengelen (approved):

esethna:
Gentle ping @hmhealey @ogi-m on this one

actions/emoji_actions.js
```diff
@@ -0,0 +1,167 @@
...
+        userIds.forEach((userId) => {
+            const customStatus = getCustomStatus(state, userId);
+            if (!customStatus || !customStatus.emoji) {
+                return;
+            }
+
+            emojisToLoad.add(customStatus.emoji);
+        });
+
+        dispatch(loadCustomEmojisIfNeeded(Array.from(emojisToLoad)));
```
hmhealey, line 100: Minor nit, but we could just return the result of this dispatch to satisfy the typing and allow the inner action to return an error if necessary instead of returning `{data: true}` on the line below this.

actions/emoji_actions.js
```diff
@@ -0,0 +1,167 @@
...
+
+            if (customEmojisByName.has(emoji)) {
+                // We have the emoji, no need to fetch
+                return;
+            }
+
+            emojisToLoad.push(emoji);
+        });
+
+        dispatch(EmojiActions.getCustomEmojisByName(emojisToLoad));
```
hmhealey, line 145: Same thing here

hmhealey (approved):
I had some minor nitpicking about return values to save some typing, but there's nothing that actually needs to change here. The new loading logic looks good!

chetanyakan:
@ogi-m Can you review/test this PR?

esethna:
@ogi-m @jgilliam17 gentle ping on this

ogi-m (approved):
LGTM 👍🏻

PR #7640, opened 2021-03-04
Cypress/E2E: Stabilize file previews specs

josephbaylon:
#### Summary
- Stabilize file previews specs

#### Ticket Link
NA


prapti (approved):

furqanmlk (approved):
Thanks @josephbaylon 
question for future, would that be a good idea to put these three lines in local functions ?

josephbaylon:
@furqanmlk I don't think that's necessary. These are more patch work that actual lines functional code. It's only waiting before and after and that can change in the future

PR #7647, opened 2021-03-05
MM-33581 Fix off-by-one error in wrapEmojis

hmhealey:
Jelena found a weird bug this morning where we don't render the name of sidebar categories with one-character names. Turns out that happened because `wrapEmojis` had an off-by-one error that made it sometimes not include the last character in its input, so when that was called on a one-character name, it would return the empty string.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33581

### DIFF
--- a/utils/emoji_utils.test.jsx
+++ b/utils/emoji_utils.test.jsx
@@ -282,4 +282,24 @@ describe('wrapEmojis', () => {
             </span>,
         ]);
     });
+
+    test('should return a one character string if it contains no emojis', () => {
+        const input = 'a';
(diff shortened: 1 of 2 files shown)


michelengelen (approved):
LGTM! 🍬

Willyfrog (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified 'one character' name categories are created, no longer blank. Also, checked emoji in category name - no change.

PR #7648, opened 2021-03-05
MM-24009: remove redux-offline

lieut-data:
#### Summary
See <PR> for details, with this PR containing the necessary changes to support removing redux-offline as a library.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-24009


lieut-data:
I'm not yet sure why this is failing in CI, as the build passes locally. Will investigate the typing issues.

store/index.js
```diff
@@ -195,21 +194,43 @@ export default function configureStore(initialState) {
...
                 return state[key];
             },
             _stateSetter: (state, key, value) => {
+                // eslint-disable-next-line no-process-env
+                if (process.env.NODE_ENV === 'production') {
+                    if (key.indexOf('storage:') === 0) {
+                        state.storage = state.storage || {storage: {}};
+                        state.storage.storage[key.substr(8)] = value;
+                    } else {
+                        state[key] = value;
```
lieut-data, line 208: Note that previously `state[key] = value` occurred /in addition/ to the redirected overwrite. I tested this and couldn't find any reason to do this, so removed the spurious write.

store/index.js
```diff
@@ -195,21 +194,43 @@ export default function configureStore(initialState) {
             },
             _stateGetter: (state, key) => {
                 if (key.indexOf('storage:') === 0) {
-                    state.storage = state.storage || {storage: {}};
-                    return state.storage.storage[key.substr(8)];
+                    return state.storage?.storage?.[key.substr(8)];
```
lieut-data, line 197: Avoid mutating the store at all here.

hmhealey: I'm surprised that didn't cause us any problems before since that should've caused it to error out

hmhealey (approved):

devinbinnie (approved):

devinbinnie:
@lieut-data Does this need QA? Not sure there's much for QA to review here, perhaps on the webapp PR?

lieut-data:
Good call, @devinbinnie -- nothing for QA review on /this/ PR. Let's focus our efforts on the webapp portion.

lieut-data:
(Just realized this /is/ the webapp PR, not the redux one :p)

jgilliam17:
@lieut-data You mentioned in the linked ticket "posting, deleting posts, editing preferences, leaving a channel and retry posting should it fail" Are there any other areas I should concentrate testing on? Let me know. Thanks.

lieut-data:
@jgilliam17, I think that's all the areas that are "touched" by this change, having previously relied on redux-offline for some extra behaviour. Thanks for double checking! :)

jgilliam17:
@lieut-data seeing few errors in console
Screen Shot 2021-03-11 at 9 05 29 AM

lieut-data:
@jgilliam17, yikes that server seems to be spewing gateway errors. Do you know the context in which that happened? It /seems/ like a transient cloud issue, maybe because the server went down overnight for some maintenance?

jgilliam17:
I believe I was mentioning a user (not currently a member of the channel) and then adding them to channel when a whole page of errors loaded

lieut-data:
Got it, @jgilliam17 -- given the myriad of endpoints failing there, I suspect it's something cloud specific. Can you raise it with the cloud team just so we can be sure to trace down the "hiccup" here? Also, I suspect this is not reproducible?

jgilliam17:
Not reproducible. I cleared the console and so far see only one line for incident plugin 
Screen Shot 2021-03-11 at 9 18 03 AM
Also, this is probably a Chrome issue, when throttling network and going `Offline` I am still receiving messages, which is weird, but can't send (as expected). 
Screen Shot 2021-03-11 at 9 19 09 AM 1

lieut-data:
Sounds good -- the incident collaboration is expected, in that 404s are being used semantically vs. indicative of an error.

Yes, chrome is unusual in that their "offline mode" doesn't stop active websockets, so incoming messages still get processed. It's caught me off guard several times :)

jgilliam17:
@lieut-data I've reviewed [E2E resutls](https://mm-cypress-report.s3.amazonaws.com/3743-webapp-pr-7648-mattermost-enterprise-edition:9131b3e/mochawesome.html), compared with latest master prod test run; most failures are timed out tests and not repro-ing when manually tested on the test server (or have a fix pending)
There is an error and perpetually loading screen when leaving an archived channel; reproduces manually on the test server. 
Screen Shot 2021-03-12 at 11 25 44 AM

lieut-data:
Thanks for your careful eye, @jgilliam17! Can you help me with reproducing the above archived channel oddity? I've tested leaving archived channels in a variety of ways both locally and on the test server for this PR without encountering the specific issue seen above.

Perhaps it requires a particular setup before getting stuck in this state?

jgilliam17:
Odd, it's intermittent for me now.
On the PR server - try searching for an archived channel you are not a member off (as sysadmin, `mmm` channel), click on the name to join and then use channel menu option to Leave Channel

lieut-data:
This is a tremendous find, @jgilliam17! So when things seem non-deterministic like this, I often find that throttling the network tends to surface them more easily (i.e. via Chrome/Firefox devtools). In this case, simulating a 2G connection allowed me to recreate the problem quite easily, /and/ verify that it's not occurring on `master`.

Looking at the code again, I observed a subtle semantic difference to my changes: the parent action now waits for the Client4 API call to finish (or fail) before returning to the caller. Previously, by virtue of being embedded within redux-offline magic, it would occur asynchronously.

It turns out some of the webapp code is sensitive to this (probably incorrectly), and ends up racing with the code to determine the next steps, incorrectly sending the user back to the channel they just left since the Client4 API call leaving the channel probably results in a websocket event or some such and a later event.

To avoid this, I've restored the original behaviour by wrapping everything in an immediately-evaluated-function-expression (just to keep the `async` capabilities) but on which the parent action does not wait. I could not reproduce the behaviour after making this change.

I've asked for a second dev review, and will loop you back in once that passes muster. Thanks again :)

lieut-data:
(For clarity, the changes to review are in https://github.com/mattermost/mattermost-redux/pull/1396).

devinbinnie (approved):

hmhealey (approved):

jgilliam17 (approved):
Thanks @lieut-data - Let me know if you think we need to run E2Es again
Quick smoke test and it looks good to merge ( tested with no throttling and slow 3G)
- Verified posting, deleting posts, editing preferences, leaving public, private or archived channels and `retry` posting after connection loss - no issues.

PR #7651, opened 2021-03-06
Remove Enable/Disable and Remove from System Console for cloud apps

larkox:
#### Summary
Since the enable/disable status is controlled by the feature flag, we should hide those buttons from the system console.

Screenshot from 2021-03-06 11-47-20

Screenshot from 2021-03-06 11-48-03

#### Ticket Link
None


components/admin_console/custom_plugin_settings/index.js
```diff
@@ -64,9 +64,11 @@ function makeGetPluginSchema() {
                 });
             }
 
-            const pluginEnableSetting = getEnablePluginSetting(plugin);
-            pluginEnableSetting.isDisabled = it.any(pluginEnableSetting.isDisabled, it.not(it.userHasWritePermissionOnResource('plugins')));
-            settings.unshift(pluginEnableSetting);
+            if (plugin.id !== 'com.mattermost.apps') {
```
catalintomai, line 67: TODO: move `com.mattermost.apps` to some const

catalintomai (approved):

components/admin_console/plugin_management/plugin_management.tsx
```diff
@@ -360,6 +360,23 @@ const PluginItem = ({
         );
     }
 
+    if (pluginStatus.id === 'com.mattermost.apps') {
+        activateButton = (<></>);
+        removeButton = (<></>);
```
mickmister, line 365: 0/5 Can these be set to `null` instead? I think that's usually the convention used in the webapp and mobile repos when we want to render nothing for a given local variable.

mickmister (approved):

larkox:
From my perspective we have the following cases:
- Cloud servers: They should obey the feature flag
- Average on prem server: The apps feature should be desabled by default
- Early adopters / developers on prem: They shouldn't install directly the proxy plugin. They should set up their instance with the Feature flag environment variable set. The split io is a way to set that environment variable for our environment, but in the general case it depends on how the installation is (e.g. for local servers, you can just run `make MM_FEATUREFLAG_AppsEnabled=true` (or something like that)).

larkox:
Update behavior to allow change the status when the feature flag is on.

Here screenshots:
Feature flag off:
Screenshot from 2021-03-17 13-01-31
Screenshot from 2021-03-17 13-01-40

Feature flag on:
Screenshot from 2021-03-17 13-02-20
Screenshot from 2021-03-17 13-02-14

@aaronrothschild Can you confirm the string used here is the correct one?

catalintomai (approved):

mickmister (approved):

PR #7661, opened 2021-03-08
remove legacy open-sans fonts, upgrade to v18

lieut-data:
#### Summary
Adopt the "Practical Level of Browser Support" for our Open Sans font, as outlined on https://css-tricks.com/snippets/css/using-font-face/, and effected via the [google-webfonts-helper](https://google-webfonts-helper.herokuapp.com/fonts/open-sans?subsets=cyrillic,cyrillic-ext,greek,greek-ext,latin,latin-ext,vietnamese), along with documentation on how to do this in the future.

As part of this, upgrade open sans from `v13` to `v18` (not the embedded version number, but the one reported by google-webfonts-helper), and eliminating an issue with the kerning table that generated errors on Firefox.

This /should/ be identical to our current font, down to the supported charsets. I'm starting with dev review, then will loop in UX and QA to help validate things that I'm missing.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-33632


lieut-data:
(For context, this is part of my quest to eliminate the spam that are console error logs.)

nevyangelova (approved):
Thanks @lieut-data for this much needed change. I am not entirely sure how to test this thoroughly but from what I checked everything loads correctly (Cyrillic looks good hehe) so LGTM! But perhaps @deanwhillier can provide a more accurate review.

deanwhillier (approved):
LGTM!

@lieut-data, is the zip file needed? (fonts/open-sans-v18-vietnamese_latin-ext_latin_greek-ext_greek_cyrillic-ext_cyrillic.zip)

lieut-data:
Good eye, @deanwhillier! No, the zip file is unnecessary -- removed.

lieut-data:
Queuing UX & QA review in parallel, since I suspect the overlap of focus is similar and without risk of wasted effort should tweaks be required.

matthewbirtch (approved):
I did a pass through all the key screens I could think of and I'm not seeing any visible differences, which is a good. Seems to be good to go from my perspective.

PR #7672, opened 2021-03-09
MM-33661 Spurious "join private channel" warning fix

anurag6713:
#### Summary
We are getting unwanted prompts on navigating to already joined private channel by switch between teams as the membership data is not present in the redux store. As a fix we will fallback to server check if membership is false from redux store data. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33661


BenCookie95 (approved):
Code LGTM, I think we need to add tests for the ticket use case

furqanmlk (approved):

amyblais:
> Needs cherry-picks to `release-5.33` and `cloud`.

@anurag6713 ^

PR #7697, opened 2021-03-14
[MM_20380/GH_16727] Migrate 'components/permalink_view' module and associated tests to TypeScript

kritiag20:
Signed-off-by: Kriti Agarwal <kritiagarwal2098@gmail.com>

#### Related Pull Requests


aggmoulik:
Please change the title according to the Contribution guide.

components/permalink_view/actions.ts
```diff
@@ -1,27 +1,27 @@
...
+import { browserHistory } from 'utils/browser_history';
+import { joinPrivateChannelPrompt } from 'utils/channel_utils';
+import { ActionTypes, Constants, ErrorPageTypes } from 'utils/constants';
+import { getUserIdFromChannelId, isSystemAdmin } from 'utils/utils';
 
 let privateChannelJoinPromptVisible = false;
 
-export function focusPost(postId, returnTo = '', currentUserId) {
-    return async (dispatch, getState) => {
+export function focusPost(postId: string, returnTo?: string, currentUserId?: string): Promise<void> {
```
lieut-data, line 18: Can you use `ActionFunc` from mattermost-redux as the return type for this action creator?

components/permalink_view/actions.ts
```diff
@@ -1,27 +1,27 @@
...
+    return async (dispatch: any, getState: any) => {
```
lieut-data, line 19: Can you help avoid using `any` here? There are `DispatchFunc` and `GetStateFunc` definitions available too.

components/permalink_view/actions.ts
lieut-data: The way we pass `currentUserId` into this method seems odd -- the only component I can find using it this way just gets it from the state anyway. @nevyangelova, this goes back to last year, but wondering if you can help shed context? I'd love to ask @kritiag20 to make this parameter required and then we can skip all the checks below.

components/permalink_view/index.ts
```diff
@@ -0,0 +1,38 @@
...
+import { connect } from 'react-redux';
+import { bindActionCreators, Dispatch} from 'redux';
+
+import { getCurrentChannel } from 'mattermost-redux/selectors/entities/channels';
+import { getCurrentTeam } from 'mattermost-redux/selectors/entities/teams';
+import { getCurrentUserId } from 'mattermost-redux/selectors/entities/users';
+
+import { focusPost } from './actions';
+import PermalinkView from './permalink_view';
+import {GlobalState} from 'mattermost-redux/types/store'
```
lieut-data, line 13: Can you group these imports with the rest of `mattermost-redux` above?

components/permalink_view/permalink_view.tsx
```diff
@@ -2,41 +2,46 @@
...
-        actions: PropTypes.shape({
-            focusPost: PropTypes.func.isRequired,
-        }).isRequired,
-        currentUserId: PropTypes.string.isRequired,
+
+import Constants from 'utils/constants';
+import * as Utils from 'utils/utils';
+
+
+type Props= {
```
lieut-data, line 10: Can you help fix the linting errors here?

lieut-data (changes requested):
Thanks, @kritiag20! It looks like there a bunch of linting issues to resolve here, as well as some improvements I've highlighted below, especially around `any`. Can you take a look?

components/permalink_view/actions.ts
kritiag20: I used ActionFunc as the return type and it's giving the same error -
Type '(dispatch: DispatchFunc, getState: GetStateFunc) => Promise<void>' is missing the following properties from type 'Promise<ActionFunc>': then, catch, finally, [Symbol.toStringTag]

components/permalink_view/actions.ts
```diff
@@ -53,13 +54,16 @@ export function focusPost(postId, returnTo = '', currentUserId) {
...
+                membership = await dispatch(getChannelMember(channel.id, currentUserId));
+            }
 
-            const membership = await dispatch(getChannelMember(channel.id, currentUserId));
             if ('data' in membership) {
                 myMember = membership.data;
             }
 
-            if (!myMember) {
+            if (!myMember && typeof currentUserId !== 'undefined') {
```
aggmoulik, line 66:
```suggestion
            if (!myMember && currentUserId) {
```
if condition will automatically check for `undefined`

components/permalink_view/permalink_view.tsx
```diff
@@ -57,25 +56,26 @@ export default class PermalinkView extends React.PureComponent {
         this.mounted = false;
     }
 
-    doPermalinkEvent = async (props) => {
-        const postId = props.match.params.postid;
-        await this.props.actions.focusPost(postId, this.props.returnTo, this.props.currentUserId);
+    doPermalinkEvent = async (props: Props) => {
+        const postId = props?.match?.params.postid;
+        if (this.props.actions) {
```
aggmoulik, line 61:
```suggestion
        if (this.props?.actions) {
```
It is because actions are optional in `Props`

lieut-data (changes requested):
Thanks for the efforts here, @kritiag20! Can you take a look at the linting issues and comments raised above, and ping me when ready for another review?

M-ZubairAhmed:
completed in https://github.com/mattermost/mattermost-webapp/pull/10703

PR #7698, opened 2021-03-14
[MM_20387/GH_16719] Migrate 'components/sidebar_right' module and associated tests to TypeScript

aggmoulik:
Signed-off-by: Moulik Aggarwal <qwertymoulik@gmail.com>

#### Summary
It fixes the migration of sidebar_right component in TypeScript. 

#### Related Pull Requests


components/sidebar_right/index.ts
```diff
@@ -59,4 +63,4 @@ function mapDispatchToProps(dispatch) {
     };
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(SidebarRight);
+export default connect(mapStateToProps, mapDispatchToProps)(SidebarRight as unknown as React.ComponentClass);
```
Willyfrog, line 66: why not simply do an `as React.ComponentClass`?

components/sidebar_right/sidebar_right.tsx
```diff
@@ -52,7 +61,7 @@ export default class SidebarRight extends React.PureComponent {
         };
     }
 
-    setPrevious = () => {
+    setPrevious = (): unknown => {
```
Willyfrog, line 64: by changing the if condition, we don't need to define a type for doing nothing
```tsx
    setPrevious = (): void => {
         if (this.props.isOpen) {
           this.previous = {
              searchVisible: this.props.searchVisible,
              isMentionSearch: this.props.isMentionSearch,
              isPinnedPosts: this.props.isPinnedPosts,
              isFlaggedPosts: this.props.isFlaggedPosts,
              selectedPostId: this.props.selectedPostId,
              selectedPostCardId: this.props.selectedPostCardId,
              previousRhsState: this.props.previousRhsState,
           };
       }
     }
```

components/sidebar_right/sidebar_right.tsx
```diff
@@ -129,26 +138,26 @@ export default class SidebarRight extends React.PureComponent {
...
     };
 
-    handleUpdateSearchTerms = (term) => {
+    handleUpdateSearchTerms = (term: string): void => {
         this.props.actions.updateSearchTerms(term);
         this.focusSearchBar();
     }
 
-    getSearchBarFocus = (focusSearchBar) => {
+    getSearchBarFocus = (focusSearchBar: unknown): void => {
```
Willyfrog, line 156: this seems weird, I think you have to do that because at some point it is being called with a function that return void  as an argument, would you be so kind to fix it so we can have a simpler signature?

Willyfrog (changes requested):
Hey! looking good and thanks a lot for the effort.

I found a couple of weird signatures that I think can be improved because I couldn't find a good reason to keep it (but by all means tell me I'm wrong if I misunderstood some part), can you take a look at those?

components/sidebar_right/index.ts
aggmoulik: I did the same before this hack but it was still showing errors by type-check and I did some research and I get to this hack.

components/sidebar_right/sidebar_right.tsx
aggmoulik: This is the simplest signature I can think of right now as TypeScript is growing too fast and it is really hard to keep up.

But, I don't understand how can I simplest this further because I have added a type for the argument and added a return type for the function. If I don't add return type then ESLint is showing errors for same.

aggmoulik: I didn't get what to define type for `focusSearchBar` that's why I defined the `unknown` type.

components/sidebar_right/index.ts
aggmoulik:
```
Conversion of type 'typeof SidebarRight' to type 'ComponentClass<{}, any>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

Conversion of type 'typeof SidebarRight' to type 'ComponentClass<{}, any>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Types of parameters 'props' and 'props' are incompatible.
    Type '{}' is missing the following properties from type 'Props': isExpanded, isOpen, currentUserId, searchVisible, and 2 more.
```

I am still getting this error for `unknown` type.

components/sidebar_right/sidebar_right.tsx
aggmoulik: I can also use `any` but over the time TypeScript introduced `unknown` as `any` skip the type checking which can turn into heavy errors.  So, I used `unknown` instead of `any`.

components/sidebar_right/sidebar_right.tsx
```diff
@@ -1,49 +1,58 @@
...
+};
+
+export type State = {
+    isOpened: boolean;
+}
 
-    constructor(props) {
+export default class SidebarRight extends React.PureComponent<Props, State> {
+    sidebarRight: React.RefObject<HTMLDivElement>;
+    previous: { searchVisible: boolean | undefined; isMentionSearch: boolean | undefined; isPinnedPosts: boolean | undefined; isFlaggedPosts: boolean | undefined; selectedPostId: string | undefined; selectedPostCardId: string | undefined; previousRhsState: string | undefined } | undefined;
```
devinbinnie, line 52: You can use `fieldName?: type` instead of `fieldName: type | undefined` as a shorthand for optional parameters.

devinbinnie: It might be worth adding this type in the `types` folder since it's also used by the action that the `previous` field is called with.

devinbinnie (commented):
Thanks for the PR @aggmoulik, I've added some comments for you to review.

components/sidebar_right/sidebar_right.tsx
Willyfrog: sorry, maybe I should have been more explicit, but I don't really see the point of nesting a call that doesn't generate a result, wouldn't it make more sense to concatenate them?
```js
searchBarFocus();
search();
```
so the signature now would be `() => void`

components/sidebar_right/sidebar_right.tsx
aggmoulik: I have added in the types folder. Let me know if you need any changes or suggest. @devinbinnie

components/sidebar_right/sidebar_right.tsx
aggmoulik: I found that `searchBarFocus` has been in `search_bar` as well as in `search` under the same signature of `getFocus`.

devinbinnie:
@aggmoulik Looks like you have a lint error to resolve:
```
/root/mattermost/mattermost-webapp/components/sidebar_right/index.ts
  5:1  error  There should be at least one empty line between import groups  import/order
  ```

File: components/sidebar_right/sidebar_right.tsx [hunk 5]
```diff
@@ -103,22 +113,24 @@ export default class SidebarRight extends React.PureComponent {
...
     }
 
-    determineTransition = () => {
-        const transitionInfo = window.getComputedStyle(this.sidebarRight.current).getPropertyValue('transition');
+    determineTransition = (): void => {
+        const transitionInfo = window.getComputedStyle(this.sidebarRight.current as Element).getPropertyValue('transition');
         const hasTransition = Boolean(transitionInfo) && transitionInfo !== 'all 0s ease 0s';
 
-        if (this.sidebarRight.current && hasTransition) {
+        if (this.sidebarRight.current && hasTransition && this.props?.isOpen) {
```
devinbinnie, line 133: Is `props` ever `undefined`? I wouldn't think it would be.

components/sidebar_right/sidebar_right.tsx
```diff
@@ -129,29 +141,30 @@ export default class SidebarRight extends React.PureComponent {
...
+    onFinishTransition = (e: TransitionEvent): void => {
+        if (e.propertyName === 'transform' && this.props.isOpen) {
             this.setState({isOpened: this.props.isOpen});
         }
     }
 
-    onShrink = () => {
-        this.props.actions.setRhsExpanded(false);
+    onShrink = (): void => {
+        this.props?.actions?.setRhsExpanded(false);
```
devinbinnie, line 151: Same here wrt. `props` being `undefined`

components/sidebar_right/sidebar_right.tsx
```diff
@@ -129,29 +141,30 @@ export default class SidebarRight extends React.PureComponent {
...
     };
 
-    handleUpdateSearchTerms = (term) => {
-        this.props.actions.updateSearchTerms(term);
-        this.focusSearchBar();
+    handleUpdateSearchTerms = (term: string): void => {
+        this.props?.actions?.updateSearchTerms(term);
```
devinbinnie, line 155: Same here wrt. `props` being `undefined`

components/sidebar_right/sidebar_right.tsx
```diff
@@ -166,6 +179,16 @@ export default class SidebarRight extends React.PureComponent {
         let content = null;
         const isSidebarRightExpanded = (postRightVisible || postCardVisible || isPluginView || searchVisible) && isExpanded;
 
+        let currentUserId = '';
```
devinbinnie, line 182: This shouldn't be needed if we force `currentUserId` as required.

types/actions.ts
```diff
@@ -13,3 +13,13 @@ export function isSuccess(result: ActionResult): result is SuccessResult {
 export function isError(result: ActionResult): result is ErrorResult {
     return Boolean((result as ErrorResult).error);
 }
+
+export type Previous = {
```
devinbinnie, line 17: Thanks for adding the type :)
I don't think each field needs to be optional however.
If we ever needed that case, we could always use `Partial<Previous>`

devinbinnie (commented):
Couple more comments for you @aggmoulik 
Thanks for your hard work :)

devinbinnie:
@aggmoulik Let us know if you have any questions :)

M-ZubairAhmed:
merged in https://github.com/mattermost/mattermost-webapp/pull/10726

PR #7699, opened 2021-03-15
[MM_20388/ GH_16718] Migrate 'components/signup/signup_email' module and associated tests to TypeScript

EricMontague:
#### Summary
Migrated all modules in the above directory to Typescript, including tests

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16718


components/signup/signup_email/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export type PasswordConfig = {
+    minimumLength: number;
+    requireLowercase: boolean;
+    requireNumber: boolean;
+    requireSymbol: boolean;
+    requireUppercase: boolean;
+}
+
+export type Props = {
```
hmhealey, line 31: Can you move these definitions into `signup_email.tsx` and re-export them from there if necessary? That's generally where we define all of the Props and State types unless the `mapStateToProps` needs additional props of its own

components/signup/signup_email/index.ts
```diff
@@ -0,0 +1,104 @@
...
+    loginById: (id: string, password: string, mfaToken?: string) => Promise<{data: boolean} | {error: ServerError}>;
+    setGlobalItem: (name: string, value: string) => {data: boolean};
+    getTeamInviteInfo: (inviteId: string) => Promise<{data: TeamInviteInfo} | {error: ServerError}>;
+};
+
+function mapStateToProps(state: GlobalState) {
+    const config = getConfig(state);
+
+    const enableSignUpWithEmail = config.EnableSignUpWithEmail === 'true';
+    const siteName = config.SiteName as string;
```
hmhealey, line 75: Since `config.SiteName` is nullable, it might be better to define this as `config.SiteName || ''` instead of using a type assertion since that could cause problems elsewhere if someone were to check something like `siteName.length`

components/signup/signup_email/index.ts
```diff
@@ -0,0 +1,104 @@
...
+        actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc | GenericAction>, Actions>({
+            createUser,
+            loginById,
+            setGlobalItem,
+            getTeamInviteInfo,
+        }, dispatch),
+    };
+}
+
+/* This is a workaround to handle the issue of Typescript not being able to correctly infer the types of the component's Props */
```
hmhealey, line 103: Is this working around the types of the actions being incorrect coming from mattermost-redux? I think we generally work around it some other way, but I don't think it's exactly better than this, so I'd be fine leaving that it if that's the case.

components/signup/signup_email/signup_email.test.tsx
```diff
@@ -4,24 +4,24 @@
...
 import {browserHistory} from 'utils/browser_history';
 
 describe('components/SignupEmail', () => {
     const requiredProps = {
         location: {
-            query: {
-                token: '9f392f193973g11ggh398h39hg0ghH',
-                email: 'test@example.com',
-            },
+            search: '?token=9f392f193973g11ggh398h39hg0ghH&email=test@example.com',
```
hmhealey, line 17: Good catch on this. I guess this test didn't work before since it was an invalid object.

components/signup/signup_email/signup_email.tsx
```diff
@@ -122,13 +138,13 @@ export default class SignupEmail extends React.PureComponent {
         }
     }
 
-    handleSignupSuccess = (user, data) => {
+    handleSignupSuccess = (user: UserSignUpInfo, data: UserProfile) => {
```
hmhealey, line 141: I think the type of `User` will also be `UserProfile`, and since we're asserting the type on the object we pass in, I think that should be accepted automatically. It's weird that we're passing basically duplicate data as both parameters though.

components/signup/signup_email/signup_email.tsx
```diff
@@ -41,16 +53,20 @@ export default class SignupEmail extends React.PureComponent {
...
+
+    constructor(props: Props) {
         super(props);
 
         const data = (new URLSearchParams(this.props.location.search)).get('d');
         const token = (new URLSearchParams(this.props.location.search)).get('t');
         const inviteId = (new URLSearchParams(this.props.location.search)).get('id');
 
-        this.state = {};
+        this.state = {} as State;
```
hmhealey, line 67: Since the only non-nullable field in the state is `loading`, it might be better to have `this.state = {loading: false};` here instead of using a type assertion.

components/signup/signup_email/signup_email.tsx
```diff
@@ -290,7 +306,7 @@ export default class SignupEmail extends React.PureComponent {
                     defaultMessage='Valid email required for sign-up'
                 />
             </span>
-        );
+        ) as React.ReactNode; /* Assertion used to allow for setting variable to an empty string below */
```
hmhealey, line 309: Can we avoid the type assertion here if we change the code below to replace `emailHelpText = ''` with `emailHelpText = undefined`? They should be functionally equivalent, and I think the only reason we're using strings here is that it's old code from before we realized passing `null` or `undefined` was better than passing a string for an unrendered element.

package-lock.json
```diff
@@ -9399,8 +9399,7 @@
           "version": "3.0.0",
           "resolved": "https://registry.npmjs.org/get-stream/-/get-stream-3.0.0.tgz",
           "integrity": "sha1-jpQ9E1jcN1VQVOy+LtsFqhdO3hQ=",
-          "dev": true,
-          "optional": true
+          "dev": true
```
hmhealey, line 9402: Do you mind reverting the changes to this file? One of our dependencies causes it to change randomly, and we'd like to avoid potential conflicts by changing it unncessarily.

hmhealey (changes requested):
Thanks for the PR, @EricMontague! This looks really good, but I have a few suggestions to make things a bit more consistent with other code we've migrated over.

Let me know if you have any questions about those!

components/signup/signup_email/index.ts
EricMontague: The reason I did this was because I wasn't sure if exporting it from signup_email.tsx would cause a circular import issue or not due to needing Props and State for the connect workaround. I'll move them back where they belong if they don't cause any issues.

components/signup/signup_email/index.ts
EricMontague: This is the error I was getting from Typescript before doing this workaround. I searched in the mattermost community channels and didn't find anything on it. After a bunch of Googling and reading Stack Overflow posts and Github Issues this is the only thing that seemed to work. I'm definitely open to finding a better solution, as I'm still slightly confused as to why Typescript is giving this error.

`Argument of type 'typeof SignupEmail' is not assignable to parameter of type 'ComponentType<never>'.
  Type 'typeof SignupEmail' is not assignable to type 'ComponentClass<never, any>'.
    Types of property 'propTypes' are incompatible.
      Type '{ location: Requireable<object>; enableSignUpWithEmail: Validator<boolean>; siteName: Requireable<string>; termsOfServiceLink: Requireable<string>; privacyPolicyLink: Requireable<string>; customDescriptionText: Requireable<...>; passwordConfig: Requireable<...>; hasAccounts: Validator<...>; actions: Validator<...>; }' is not assignable to type 'undefined'.ts(2345)`

components/signup/signup_email/signup_email.tsx
EricMontague: Just as a heads up, Typescript ended up wanting me to explicitly type the user parameter as UserProfile, so on my fix, both parameters will have the same twice.

components/signup/signup_email/signup_email.tsx
EricMontague: Typescript says that 'undefined' is not assignable to type JSX.Element. The same thing goes for when I assign emailHelpText to null. I'm going to do some digging and see if I can find a cleaner fix.

components/signup/signup_email/index.ts
nevyangelova: I am 5/5 on this too as to avoid type casting.

components/signup/signup_email/signup_email.tsx
nevyangelova: Second this ☝️

components/signup/signup_email/signup_email.tsx
nevyangelova: You can always pass JSX.Element | undefined for conditionally rendered elements.

nevyangelova (changes requested):

components/signup/signup_email/signup_email.tsx
EricMontague: @nevyangelova Are you saying that I should replace `as React.ReactNode` with `as JSX.Element | undefined` and then set emailHelpText to undefined?

components/signup/signup_email/index.ts
hmhealey: I'd think having them in `signup_email.tsx` should prevent circular references since `index.ts` would import `signup_email.tsx` and that wouldn't happen the other way around. That said, if you do run into that, using `import types` to just import the types should prevent circular references causing any problems as long as its just with type defintions.

components/signup/signup_email/index.ts
hmhealey: Strange. The type being undefined definitely sounds like a circular reference issue, but if I'm understanding it correctly, it might have to do with `Props` being undefined in `SignupEmail` when that's first loaded due to it being defined in the index file. I'm not certain about that though.

components/signup/signup_email/signup_email.tsx
hmhealey: Perhaps you could define `emailHelpText` as `let emailHelpText: React.ReactNode = (...);` instead then? The differences between the two ways of doing it are subtle, but it gives us more type safety by avoiding `as`. More specifically:
1. `let emailHelpText: React.ReactNode = foo;` tells TS that `emailHelpText` should contain a `React.ReactNode` so it can throw an error if `foo` happens to not be something that matches the type.
2. `let emailHelpText = foo as React.ReactNode;` tells TS that `foo` is definitely a `React.ReactNode`, but that prevents it from throwing a type error if `foo` is the wrong type.

components/signup/signup_email/index.ts
EricMontague: Even with exporting Props from SignupEmail, Typescript still gives this error. I thought it might have to do with a mismatch between Props and propTypes. I had made siteName and location required in type Props, despite them being defined as optional in propTypes, since it seemed like they were always passed in. I'm going to set them to optional in Props and see if that changes things

components/signup/signup_email/signup_email.tsx
EricMontague: Okay, then I'll go with the type annotation instead of the type assertion. Thanks for the explanation, I didn't realize how dangerous type assertions could be.

components/signup/signup_email/index.ts
EricMontague: When I change siteName to optional in Props, TS complains about it potentially being undefined in some places where it expects a string. Would it be fine to set siteName to string | '' instead like we did in the index.ts file?

hmhealey: Yep, that would be fine

components/signup/signup_email/signup_email.tsx
```diff
@@ -23,7 +26,43 @@ import SiteNameAndDescription from 'components/common/site_name_and_description'
...
+    requireLowercase: boolean;
+    requireNumber: boolean;
+    requireSymbol: boolean;
+    requireUppercase: boolean;
+}
+
+export type Props = {
+    location?: {search: string};
+    enableSignUpWithEmail: boolean;
+    siteName: string | '';
```
hmhealey, line 42: Sorry, I meant that you could set siteName to `config.SiteName || ''` in mapStateToProps. We don't need to use the union type here since this just says it can be a string or the empty string which is redundant.
```suggestion
    siteName: string;
```

components/signup/signup_email/signup_email.tsx
```diff
@@ -23,7 +26,43 @@ import SiteNameAndDescription from 'components/common/site_name_and_description'
 
 import FormattedMarkdownMessage from 'components/formatted_markdown_message.jsx';
 
-export default class SignupEmail extends React.PureComponent {
+import {Actions} from 'components/signup/signup_email/index';
```
hmhealey, line 29: Does Actions still need to be defined in the index file or could it be defined here and imported from the index?

hmhealey (changes requested):
Thanks for the changes. I just had a couple more things to nitpick over with the types, but otherwise, it looks good.

components/signup/signup_email/signup_email.tsx
EricMontague: No I don't think so. I'll move it to this file to change where it's imported from

nevyangelova (approved):
LGTM

jgilliam17 (approved):
Thank you @EricMontague 
Tested, looks good to merge.
- Verified signup email module - as before.

jgilliam17:

PR #7708, opened 2021-03-15
MM-33540 Move category menu out of button

hmhealey:
There's currently a massive warning in the web app right now that makes the console nearly impossible to use because the sidebar category menu button is nested immediately inside the button that makes up the expand/collapse button for the header. This tries to fix that without breaking anything else.

It seems like just moving it fixes the bug (although I had to fix the alignment of the New badge), although I'm worried there was something else I missed with this. I seem to remember trying this before and breaking something, but I haven't been able to reproduce any issues. As far as I can tell, this also doesn't seem to change the clickable area for the expand/collapse even though it should theoretically have extended past the menu before.

This also seems to fix the issue where the category menu appears behind the plus button on the DMs category ~~and the scroll bar~~ too.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33540


devinbinnie (commented):
How does this interact with new categories and the `NEW` badge that pops up?
What about categories with really long names?

I don't think you broke anything, those are just cases that could potentially break this.

hmhealey:
I tested both of those cases since I knew they'd be potentially broken, but they both seemed stable. The only thing I found was that the NEW badge alignment was off, but I included a fix for that.

devinbinnie (approved):

hmhealey:
@deanwhillier Would you have a chance to look at this in the next few days? If not, I can re-assign.

deanwhillier (approved):
LGTM. Didn't see anything immediately and/or obviously broken.

jgilliam17:
Thanks @hmhealey 
There is an issue with the header when new badge is present, category header background doesn't extend all the way across.

https://user-images.githubusercontent.com/52937121/112658114-77d23400-8e29-11eb-9dd8-b700dea27f1d.mov

- The scroll bar (transparency issue) is visible over category menu. Noting it does not overlap on channel menus.
Screen Shot 2021-03-26 at 11 53 26 AM

hmhealey:
@jgilliam17 I've fixed the missing background on the NEW badge, but I can't fix the scroll bar overlap, so I've moved that ticket back to the backlog.

jgilliam17:
Thanks @hmhealey 
Background for `New` badge looks good; but the fix broke category menu 
Screen Shot 2021-03-29 at 10 11 53 AM

This (below) repros on daily, noting here since I just realized while testing this PR. Let me know if we need to address this and I can file a separate ticket. 
- Category menu click area is larger (spans to the right) than it's button on hover

https://user-images.githubusercontent.com/52937121/112850321-e3114580-9077-11eb-8601-de5cc6280658.mov

hmhealey:
Closing in favour of #8001

PR #7714, opened 2021-03-16
Only use fake timers for tests that need them

hmhealey:
Turning on fake timers for everything breaks a bunch of tests in mattermost-redux since those rely a lot more on async code and timeouts. Now we'll only use them on the specific test files that need them.


devinbinnie (approved):

lieut-data (approved):

hmhealey:
Skipping QA since this change only affects unit tests

PR #7718, opened 2021-03-17
[MM-19327] Add keyboard shortcut to focus search bar and search in current channel

devinbinnie:
#### Summary
Added a keyboard shortcut (Ctrl/Cmd+Shift+F on browser, Ctrl/Cmd+F on desktop) to focus the top search bar, and set the search term to `in:{currentChannelName}` as a shortcut for the user to quickly search the channel they're working in.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-19327


components/search/search.tsx
```diff
@@ -78,6 +79,23 @@ const Search: React.FC<Props> = (props: Props): JSX.Element => {
         new SearchUserProvider(actions.autocompleteUsersInTeam),
     ]);
 
+    const handleKeyDown = (e: KeyboardEvent) => {
+        if (Utils.cmdOrCtrlPressed(e) && e.key === 'f' && !isSideBarRight) {
+            if (isDesktopApp() || (!isDesktopApp() && e.shiftKey)) {
+                handleAddSearchTerm(`In:${currentChannelName} `);
```
michelengelen, line 85: it is a bit misleading, but the variable `currentChannelName` can possibly be undefined if I am not mistaken.
At least that is what can be derived from the props-mapping:

`currentChannelName: getCurrentChannel(state)?.name,`

The nullish coalescent operator is what kind of leaves a question mark here.

michelengelen (approved):
just a smallish question that might not even be important!

esethna:
@devinbinnie what happens on a new server, old desktop app (where we haven't removed the browser search functionality)? Does the browser search override this in channel search, or will there be conflicts?

devinbinnie:
> @devinbinnie what happens on a new server, old desktop app (where we haven't removed the browser search functionality)? Does the desktop find search override this in channel search, or will there be conflicts?

@esethna There will be conflicts. Unfortunately since the old desktop app assumes it's the only application accepting `Ctrl+F`, we won't be able to override that. We could see if there's a way to have the webapp detect the desktop app version and turn off the new behaviour so that there's no conflict.

esethna:
> We could see if there's a way to have the webapp detect the desktop app version and turn off the new behaviour so that there's no conflict.

+1, I think it's worth investigating since this will be a common case (new server, old desktop app)

michelengelen (approved):

components/search/search.tsx
devinbinnie: First of all, I see what you did there :P
Second, I'm pretty sure `getCurrentChannel` only returns null before the app is fully loaded, so it was more of a protection to make sure that the component loaded correctly and didn't throw an error. By the time the app is loaded it should always be defined.

nevyangelova (approved):
LGTM :)

esethna (approved):

esethna:
cc// @jgilliam17 on this test case: https://github.com/mattermost/mattermost-webapp/pull/7718#issuecomment-801447514

jgilliam17:
@devinbinnie 
On macOS Catalina Desktop app Mattermost v4.7.0-develop commit: b5c59fa

cmd+f opens both find search box on the tab bar (focused) and adds `in:ChannelName` to search bar (not focused)
Screen Shot 2021-03-25 at 10 25 23 AM
cmd+shift+f also works on desktop app, opens search bar and pre-fills with `in:`
Screen Shot 2021-03-25 at 10 29 37 AM

devinbinnie:
> @devinbinnie
> On macOS Catalina Desktop app Mattermost v4.7.0-develop commit: b5c59fa
> 
> cmd+f opens both find search box on the tab bar (focused) and adds `in:ChannelName` to search bar (not focused)
> Screen Shot 2021-03-25 at 10 25 23 AM
> cmd+shift+f also works on desktop app, opens search bar and pre-fills with `in:`
> Screen Shot 2021-03-25 at 10 29 37 AM

@jgilliam17 You'll need the latest version of the desktop app, the changes were only merged yesterday. For release, the app checks for version v4.7, so as long as you're using v4.6 or earlier the new behaviour shouldn't happen at all.

As for Ctrl+Shift+F, I don't think there's anything wrong with having that enabled, as it allows people using v4.6 to use the new behaviour should they care to.

jgilliam17:
Thanks @devinbinnie I'll re-test tomorrow, new mac build wasn't available today.
(Oh, I was just adding cmd+shift+f note as a bonus 😄 )

jgilliam17:
Thanks @devinbinnie 
Retested and seeing few issues.
1. On browsers and desktop app on all OSes: 
When only search icon is present (no text input box) in lower browsers widths cmd/ctrl+shift+f and cmd/ctrl+f in desktop fail to focus the search bar. Also shortcuts are not working on mobile view, not sure if we are considering implementing mobile view at this time or not.
Screen Shot 2021-03-29 at 4 11 49 PM

2. On Windows and Ubuntu (browsers only) ctrl+shift+f is not focusing the search bar; ctrl+f on desktop works as expected (as long as search bar is present due to issue 1.)

devinbinnie:
@calebroseland @michelengelen @nevyangelova Needed to refactor some stuff to fix the issues Jelena rose, so would like to have dev re-review.
@jgilliam17 Issues should be fixed.

nevyangelova (approved):

jgilliam17:
@devinbinnie here's what I observed when going back and testing on previous desktop app releases
On 4.6.2 desktop app (macOS):
- 7718 server 
    - cmd+f opens find textbox on tab header
    - cmd+shift+f focuses search, inputs in: text (it does not work if find textbox is open on header) 
- daily (master) 
    - cmd+f opens find textbook on tab header, old functionality - as expected
    - cmd+shift+f  nothing happens
I'm having some issues with Windows, restarting and will re-test later on both win10 and ubuntu.

components/search/search.tsx
```diff
@@ -79,6 +82,30 @@ const Search: React.FC<Props> = (props: Props): JSX.Element => {
...
+            }
+        }, [currentChannelName, hideSearchBar]);
+
+        useEffect(() => {
+            document.addEventListener('keydown', handleKeyDown);
+
+            return () => {
+                document.removeEventListener('keydown', handleKeyDown);
+            };
+        }, [currentChannelName, hideSearchBar]);
```
michelengelen, line 106: I am 3/5 that this needs to only be called once (like with `componentDidMount`), because every time `currentChannelName` or `hideSearchBar` changes it will add an additional event listener to the `document`

```suggestion
        }, []);
```

components/search/search.tsx
```diff
@@ -79,6 +82,30 @@ const Search: React.FC<Props> = (props: Props): JSX.Element => {
...
+                        actions.openRHSSearch();
+                        setKeepInputFocused(true);
+                    }
+                    handleUpdateSearchTerms(`in:${currentChannelName} `);
+                    handleFocus();
+                }
+            }
+        }, [currentChannelName, hideSearchBar]);
+
+        useEffect(() => {
```
michelengelen, line 100: hook should never be inside a conditional. when needed it should handle the condition inside the hook callback

components/search/search.tsx
```diff
@@ -79,6 +82,30 @@ const Search: React.FC<Props> = (props: Props): JSX.Element => {
         new SearchUserProvider(actions.autocompleteUsersInTeam),
     ]);
 
+    const isDesktop = isDesktopApp() && isServerVersionGreaterThanOrEqualTo(getDesktopVersion(), '4.7.0');
+    if (enableFindShortcut) {
+        const handleKeyDown = useCallback((e: KeyboardEvent) => {
```
michelengelen, line 87: the `useCallback` can be removed here, when the function definition is moved inside the useEffect hook

michelengelen (changes requested):
Sorry for the late review ... I somehow missed it in my list 🙇🏻‍♂️

components/search/search.tsx
devinbinnie: The issue is that within the listener, `currentChannelName` and `hideSearchBar` are fixed when the listener is created. So if those values change (which they will depending on current channel and RHS state), the listener needs to be re-created.

@hmhealey recommended this pattern.

components/search/search.tsx
devinbinnie: Wouldn't it make more sense to not create the hook unless it was needed at all? Seems weird to add an action that will just check a conditional every single time and then not do anything else. Or is there a performance optimization I'm missing here?

components/search/search.tsx
devinbinnie: I didn't think it worked correctly when I ran the callback from `useEffect` without the `useCallback` hook.

components/search/search.tsx
hmhealey: Yeah, if I understand those hooks correctly, the props values get bound when the callback is created, so they wouldn't update without recreating the listener

devinbinnie:
> @devinbinnie here's what I observed when going back and testing on previous desktop app releases
> On 4.6.2 desktop app (macOS):
> 
> * 7718 server
>   
>   * cmd+f opens find textbox on tab header
>   * cmd+shift+f focuses search, inputs in: text (it does not work if find textbox is open on header)
> * daily (master)
>   
>   * cmd+f opens find textbook on tab header, old functionality - as expected
>   * cmd+shift+f  nothing happens
>     I'm having some issues with Windows, restarting and will re-test later on both win10 and ubuntu.

@jgilliam17 Are you able to re-test?

jgilliam17:
@devinbinnie I was able to re-test
-  Windowsnew server(7718)/ old app (4.6.2 and later)
ctrl+f - opens Find text box in header
shift+ctrl+f - opens search,  inputs in: text

devinbinnie:
> @devinbinnie I was able to re-test (edited, submitted too soon)
> 
> * Windows 10 and Ubuntu 20.04.2
> 
> _new server(7718)/new app (4.7beta2)_
> shift+ctrl+f and ctrl+f - opens search, inputs in: text as expected - passed
> 
> _old server/new app (4.7beta2)_
> ctrl+f - nothing happens, including using menu option view>find - failed

@jgilliam17 This is expected, because we had to completely remove the Find function from all versions of v4.7 due to the issues we experienced with it on Windows/Linux. But since old servers won't have the code to trigger the in-app find functionality nothing would happen.

@esethna Are we okay with this?

> shift+ctrl+f - nothing happens, as expected
> 
> _new server(7718)/ old app (4.6.2 and earlier)_
> ctrl+f - opens Find text box in the header
> shift+ctrl+f - opens search, inputs in: text - I believe we said this is ok, a bonus, to allow people on earlier versions of the app to still use this functionality
> 
> _old server/old app (4.6.2 and earlier)_
> ctrl+f - opens Find text box in the header, as expected
> shift+ctrl+f - nothing happens, as expected

components/search/search.tsx
calebroseland: I think the risk is with dynamic conditionals (render errors when hook order changes). Right now `enableFindShortcut` is always true, but if it ever became dynamic in the future it might cause errors.

components/search/search.tsx
```diff
@@ -85,6 +88,30 @@ const Search: React.FC<Props> = (props: Props): JSX.Element => {
...
+                if (isDesktop || (!isDesktop && e.shiftKey)) {
+                    if (hideSearchBar) {
+                        actions.openRHSSearch();
+                        setKeepInputFocused(true);
+                    }
+                    handleUpdateSearchTerms(`in:${currentChannel.name} `);
+                    handleFocus();
+                }
+            }
+        }, [currentChannel, hideSearchBar]);
```
calebroseland, line 104: In some routes there's no currentChannel, such as global threads, and maybe others in the future

```suggestion
    useEffect(() => {
        if (!enableFindShortcut) {
            return undefined;
        }

        const handleKeyDown = (e: KeyboardEvent) => {
            if (Utils.cmdOrCtrlPressed(e) && Utils.isKeyPressed(e, Constants.KeyCodes.F)) {
                if (isDesktop || (!isDesktop && e.shiftKey)) {
                    if (hideSearchBar) {
                        actions.openRHSSearch();
                        setKeepInputFocused(true);
                    }
                    if (currentChannel) {
                        handleUpdateSearchTerms(`in:${currentChannel.name} `);
                    }
                    handleFocus();
                }
            }
        };

        document.addEventListener('keydown', handleKeyDown);
        return () => {
            document.removeEventListener('keydown', handleKeyDown);
        };
    }, [currentChannel, hideSearchBar]);
```

components/search/search.tsx
```diff
@@ -1,17 +1,20 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {ChangeEvent, FormEvent, useEffect, useState, useRef} from 'react';
+import React, {ChangeEvent, FormEvent, useEffect, useState, useRef, useCallback} from 'react';
```
calebroseland, line 4:
```suggestion
import React, {ChangeEvent, FormEvent, useEffect, useState, useRef} from 'react';
```

components/search/types.ts
```diff
@@ -31,7 +34,7 @@ export type StateProps = {
     isPinnedPosts: boolean;
     isChannelFiles: boolean;
     filesSearchEnabled: boolean;
-    currentChannelId: string;
+    currentChannel: Channel;
```
calebroseland, line 37:
```suggestion
    currentChannel: Channel | undefined;
```

components/search/search.tsx
```diff
@@ -264,7 +295,7 @@ const Search: React.FC<Props> = (props: Props): JSX.Element => {
         }
         setSearchFilterType(filterType);
         if (props.isChannelFiles) {
```
calebroseland, line 297:
```suggestion
        if (props.isChannelFiles && currentChannel) {
```

calebroseland (changes requested):
This is game-changer, searching in-channel is about to get a lot easier! Request is only to echo Michel's point, and another regarding `currentChannel`, otherwise L(Great)TM

components/search/search.tsx
michelengelen: They are not really "bound" to the callback, but it certainly works a little like they would. 

useCallback creates a memoized instance of the function that's being passed in, so that it does not get re-defined on every small re-render.

components/search/search.tsx
michelengelen: Yeah, that's because functions called inside a hook should always be defined in that hook or wrapped in a useCallback to memoize it.

I did not take a look under the hood, but that is how it works.

components/search/search.tsx
michelengelen: @calebroseland is absolutely right with this one. There is the possibility of errors when the order or number of hooks change throughout (re-)renders.

esethna:
> This is expected, because we had to completely remove the Find function from all versions of v4.7 due to the issues we experienced with it on Windows/Linux. But since old servers won't have the code to trigger the in-app find functionality nothing would happen. @esethna Are we okay with this?

Yes this would be expected

components/search/types.ts
michelengelen: This will be sufficient, so that typescript can add the `| undefined`

```suggestion
    currentChannel?: Channel;
```

components/search/search.tsx
hmhealey: I mean I think that the values from props in the function are from when the callback is declared and that they're not updated as the props might change. This is as opposed to in a class component where `this.props.currentChannelName` always points to the latest version of the props.

That said, I like Caleb's version better where we just use a single `useEffect` and declare the callback inside of it. I wasn't a huge fan of using multiple hooks to begin with even though I might've proposed it

devinbinnie:
@hmhealey @calebroseland @michelengelen Implemented some changes, have a look guys.

michelengelen (approved):
LGTM

calebroseland (approved):

esethna:
Gentle ping @jgilliam17 for re-review

jgilliam17:
Thanks @devinbinnie 
Looks good on re-test.
One small issue on Firefox, the keyboard shortcut is shared; it opens search, inputs in: text as expected, but also simultaneously triggers `Enter/Exit Full Screen`

devinbinnie:
> Thanks @devinbinnie
> Looks good on re-test.
> One small issue on Firefox, the keyboard shortcut is shared; it opens search, inputs in: text as expected, but also simultaneously triggers `Enter/Exit Full Screen`

@jgilliam17 Should be fixed now

jgilliam17 (approved):
Thanks @devinbinnie 
Tested, looks good to merge.
- Verified keyboard shortcuts (Ctrl/Cmd+Shift+F on browser, Ctrl/Cmd+F on desktop).  Using shortcuts focuses the top search bar and sets the search term to in:[channel name] so user to quickly search the channel they are currently in.

PR #7723, opened 2021-03-18
Move mattermost-redux into mattermost-webapp

hmhealey:
I don't know what exactly can be done to review this as a dev, so feel free to look at [the migration scripts](https://gist.github.com/hmhealey/159034370ed63dc1f9e0487f2e98c350) and browse the git branch to make sure everything seems to be in the right place. I put the changes to existing web app files in their own commit in case you wanted to view those separately.

I've also assigned more reviewers than necessary just so that everyone's aware of what's changing and in case anyone's curious how it might affect them. Ideally, little of this should affect anyone negatively minus having to hop between repos when browsing history.

Also, I know this'll inflate my lines added count, but I had to do something to catch up to Saturn and Christopher 🙃.


components/admin_console/admin_sidebar_header/admin_sidebar_header.tsx
```diff
@@ -3,8 +3,9 @@
 
 import React from 'react';
 import {FormattedMessage} from 'react-intl';
+
 import {Client4} from 'mattermost-redux/client';
-import {UserProfile} from 'mattermost-redux/src/types/users';
+import {UserProfile} from 'mattermost-redux/types/users';
```
hmhealey, line 8: A few of these files incorrectly imported things from mattermost-redux's src directory since importing from `mattermost-redux` takes it from `packages/mattermost-redux/src` directly now, that needs to be changed to prevent it from trying to look at `src/src`

package.json
```diff
@@ -27,6 +27,8 @@
     "flexsearch": "0.6.32",
     "flux": "4.0.1",
     "font-awesome": "4.7.0",
+    "form-data": "3.0.0",
```
hmhealey, line 30: These are all dependencies of mattermost-redux that we need to include ourselves now. Once we get the actual monorepo infrastructure sorted out, these will move back into the subpackages.

package.json
```diff
@@ -198,6 +207,8 @@
       "text-summary"
     ],
     "moduleNameMapper": {
+      "^mattermost-redux/test/(.*)$": "<rootDir>/packages/mattermost-redux/test/$1",
+      "^mattermost-redux/(.*)$": "<rootDir>/packages/mattermost-redux/src/$1",
```
hmhealey, line 211: This is used to rewrite import statements when running in Jest. The test files need their own alias since they don't live in the src folder.

tests/setup.js
```diff
@@ -9,7 +9,7 @@ import '@testing-library/jest-dom';
 global.$ = $;
 global.jQuery = $;
 global.performance = {};
-global.fetch = jest.fn().mockResolvedValue({status: 200});
+require('isomorphic-fetch');
```
hmhealey, line 12: Previously, we couldn't use fetch in webapp tests. Now we can!

Also, it might make sense to swap this for `node-fetch` in the future since that's what `isomorphic-fetch` uses under the hood for non-browser environments.

tests/setup.js
```diff
@@ -60,14 +53,13 @@ beforeAll(() => {
...
 beforeEach(() => {
-    logs = [];
     warns = [];
     errors = [];
 });
 
 afterEach(() => {
-    if (logs.length > 0 || warns.length > 0 || errors.length > 0) {
-        const message = 'Unexpected console logs' + logs + warns + errors;
+    if (warns.length > 0 || errors.length > 0) {
```
hmhealey, line 61: Minor change to match how redux handled log statements in its tests. I personally think this is nicer since I tend to use `console.log` while debugging tests.

tsconfig.json
```diff
@@ -20,6 +20,9 @@
     "experimentalDecorators": true,
     "jsx": "react",
     "baseUrl": ".",
+    "paths": {
+        "mattermost-redux/*": ["packages/mattermost-redux/src/*"],
```
hmhealey, line 24: This rewrites the import paths for the type checker and for VS Code's TS plugin.

webpack.config.js
```diff
@@ -245,6 +245,8 @@ var config = {
             path.resolve(__dirname),
         ],
         alias: {
+            'mattermost-redux/test': 'packages/mattermost-redux/test',
+            'mattermost-redux': 'packages/mattermost-redux/src',
```
hmhealey, line 249: Finally, this rewrites the import paths for Webpack, Babel, and ESLint.

hmhealey (commented):
Here are the important changes done in the webapp code. 99% of them are just rearranging imports now that mattermost-redux is considered as internal code.

packages/mattermost-redux/src/action_types/admin.ts
```diff
@@ -0,0 +1,210 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import keyMirror from 'mattermost-redux/utils/key_mirror';
```
hmhealey, line 3: Note that the import paths within mattermost-redux all had to be rewritten to start with the package name. The alias let us avoid that with all the webapp code, but it's needed in here. This can hopefully be reverted once the monorepo is properly set up, but it's necessary for now.

packages/mattermost-redux/src/actions/helpers.ts
```diff
@@ -0,0 +1,151 @@
...
+
+// Debounce function based on underscores modified to use es6 and a cb
+
+export function debounce(func: (...args: any) => unknown, wait: number, immediate: boolean, cb: () => unknown) {
+    let timeout: NodeJS.Timeout|null;
+    return function fx(...args: any[]) {
+        const runLater = () => {
+            timeout = null;
+            if (!immediate) {
+                Reflect.apply(func, null, args);
```
hmhealey, line 115: Minor change which shows up in a few places. We used to pass `this` instead of `null` when binding. For some reason, TypeScript complains now while TTypeScript that used to be used does not.

packages/mattermost-redux/src/actions/admin.test.js
```diff
@@ -0,0 +1,1272 @@
...
+        assert.ok(analytics[Stats.USERS_WITH_POSTS_PER_DAY]);
+
+        const teamAnalytics = state.entities.admin.teamAnalytics;
+        assert.ok(teamAnalytics);
+        assert.ok(teamAnalytics[TestHelper.basicTeam.id]);
+        assert.ok(teamAnalytics[TestHelper.basicTeam.id][Stats.USERS_WITH_POSTS_PER_DAY]);
+    });
+
+    it('overwritePlugin', async () => {
+        const data1 = fs.createReadStream('tests/setup.js');
```
hmhealey, line 830: Another small change. These tests used to upload a copy of mattermost-redux's `test/setup.js` for this test, but that was removed since it now uses the webapp's setup script. I don't think it's necessary to upload something other than just a string, but in case it is, I just pointed it at the webapp's setup script instead.

packages/mattermost-redux/src/actions/admin.test.js
```diff
@@ -0,0 +1,1272 @@
...
+
+        const state = store.getState();
+        const request = state.requests.admin.uploadPlugin;
+        if (request.status === RequestStatus.FAILURE) {
+            throw new Error('uploadPlugin request failed err=' + request.error);
+        }
+    });
+
+    it('uploadPlugin', async () => {
+        const testFileData = fs.createReadStream('packages/mattermost-redux/test/assets/images/test.png');
```
hmhealey, line 856: Another path that needed to change since Node doesn't look at any of the import aliasing that's now set up.

packages/mattermost-redux/src/reducers/entities/emojis.ts
```diff
@@ -0,0 +1,115 @@
...
+// See LICENSE.txt for license information.
+import {combineReducers} from 'redux';
+
+import {EmojiTypes, PostTypes, UserTypes} from 'mattermost-redux/action_types';
+import {EmojisState, CustomEmoji} from 'mattermost-redux/types/emojis';
+import {GenericAction} from 'mattermost-redux/types/actions';
+import {Post} from 'mattermost-redux/types/posts';
+import {IDMappedObjects} from 'mattermost-redux/types/utilities';
+
+export function customEmoji(state: IDMappedObjects<CustomEmoji> = {}, action: GenericAction): IDMappedObjects<CustomEmoji> {
```
hmhealey, line 11: There were some weird typings here before. Most of these were imported as `import * as types from 'types'` which meant all of them were referred to as things like `types.entities.CustomEmoji`. It wasn't strictly necessary to fix if I updated the import path, but I couldn't help myself.

packages/mattermost-redux/src/reducers/entities/emojis.ts
```diff
@@ -0,0 +1,115 @@
...
+    switch (action.type) {
+    case EmojiTypes.RECEIVED_CUSTOM_EMOJI: {
+        const nextState = {...state};
+        nextState[action.data.id] = action.data;
+        return nextState;
+    }
+    case EmojiTypes.RECEIVED_CUSTOM_EMOJIS: {
+        const nextState = {...state};
```
hmhealey, line 19: A bunch of files wrote this sort of thing as `const nextState: any = {...state};` for some reason. TypeScript didn't like that, so I removed the unnecessary type change everywhere.

packages/mattermost-redux/src/reducers/entities/users.ts
```diff
@@ -0,0 +1,638 @@
...
+        const {data} = action;
+
+        if (data && data.length) {
+            const nextState = {...state};
+            data.forEach((d: any) => {
+                const {channelId, users} = d.data;
+                const nextSet = new Set(state[channelId]);
+
+                users.forEach((u: UserProfile) => nextSet.add(u.id));
+                nextState[channelId] = nextSet as any; // TODO the type of state should Record<string, Set<string>>
```
hmhealey, line 363: This one was a weird one. The type of `nextSet` seems to be incorrect since everywhere else expects each entry in `state` to be an array instead of a Set. Assuming the above comment isn't lying, this was only used by the mobile app, so I didn't bother fixing it now.

packages/mattermost-redux/src/selectors/entities/general.ts
```diff
@@ -0,0 +1,106 @@
...
+import {createSelector} from 'reselect';
+
+import {General} from 'mattermost-redux/constants';
+
+import {GlobalState} from 'mattermost-redux/types/store';
+import {ClientConfig, FeatureFlags} from 'mattermost-redux/types/config';
+
+import {isMinimumServerVersion} from 'mattermost-redux/utils/helpers';
+
+export function getConfig(state: GlobalState): Partial<ClientConfig> {
```
hmhealey, line 13: There seems to be some argument around types for `state.entities.general.config` and `getConfig` since this function previously was defined as returning a full `ClientConfig` but the redux state held a partial one. The partial seems correct because the client config might not be loaded or might be missing fields when not logged in, but I wasn't entirely sure about that.

It might be something to consider changing in the future when we properly type the values coming from the server instead of just sending down a load of strings.

packages/mattermost-redux/src/selectors/entities/preferences.ts
```diff
@@ -0,0 +1,277 @@
...
+);
+
+export const getVisibleGroupIds: (state: GlobalState) => string[] = createSelector(
+    getGroupShowPreferences,
+    (groups) => {
+        return groups.filter((dm) => dm.value === 'true' && dm.name).map((dm) => dm.name);
+    },
+);
+
+export const getTeammateNameDisplaySetting: (state: GlobalState) => string = createSelector(
```
hmhealey, line 97: The return value of this used to be nullable which the web app used to be fine with before. Now that we've merged the repos, it started complaining about type issues, so I made it not nullable and added a default value below.

packages/mattermost-redux/src/store/configureStore.dev.ts
```diff
@@ -0,0 +1,80 @@
...
+ * getAppReducer - A function that returns the appReducer as defined above. Only used in development to enable hot reloading.
+ * clientOptions - An object containing additional options used when configuring the redux store. The following options are available:
+ *     additionalMiddleware - func | array - Allows for single or multiple additional middleware functions to be passed in from the client side.
+ *     enableThunk - bool - default = true - If true, include the thunk middleware automatically. If false, thunk must be provided as part of additionalMiddleware.
+ */
+export default function configureServiceStore(preloadedState: any, appReducer: any, persistConfig: any, getAppReducer: any, clientOptions: any): Store {
+    const baseState = Object.assign({}, initialState, preloadedState);
+
+    const store = createStore(
+        createDevReducer(baseState, serviceReducer, appReducer) as any,
```
hmhealey, line 35: I had to change these invocations of `createDevReducer` to return `any` as well. There's definitely a better fix for this, but something is really messed up already with all these `any`s thrown around in this file.

hmhealey (commented):
And here is anything even vaguely noteworthy in the redux changes. I skipped over a few that were even less consequential, but most of this is pretty much just squashing minor type issues and such.

crspeller (approved):
Good work!
I checked it it builds locally for me. All my editor stuff like jump to definition works. Seems painless. 
Unfortunate that we couldn't preserve the history. I guess we will just have to blame you for everything 🙃

saturninoabril:
@hmhealey You've probably considered it already but just like to mention here. We're importing [mattermost-redux](https://github.com/mattermost/mattermost-webapp/blob/master/e2e/package.json#L28) in Cypress, primarily for `Client` and types. I wonder how/when to update it. Thanks!

hmhealey:
@saturninoabril How often do you need to update that for new changes in mattermost-redux? Since we can't easily update the npm package until we get the monorepo fully up and running, I can look at having it bypass npm for it for the time being if that's necessary.

hahmadia (approved):
Looked through the changes and they all made sense. 
Thank you and everyone who worked on this for the hard work!

nickmisasi (approved):
Built and poked around the webapp, didn't see any obvious issues. My mac really chugged in loading the diff but what I could see looked fine! Awesome work!

saturninoabril:
> How often do you need to update that for new changes in mattermost-redux?

Honestly, it's not often and in most cases, it can wait for the published version. However, it would be good as well to bypass the npm so that any local dev change in `mattermost-redux` (sub)package can be used side-by-side for Cypress scripting.

jgilliam17:
@hmhealey Here are the [Cypress test results ](https://mm-cypress-report.s3.amazonaws.com/3840-webapp-pr-7723-mattermost-enterprise-edition:80ee7ab/mochawesome.html) and [master prod results](https://mm-cypress-report.s3.amazonaws.com/3832-master-e20-prod-mattermost-enterprise-edition:master/mochawesome.html) for comparison. Failures and skipped tests are known issues. 
Also thank you @josephbaylon and @saturninoabril for starting and checking the test.

hmhealey:
Sounds like the E2E tests will fit perfectly in a monorepo environment :)

deanwhillier (approved):

i18n/en.json
```diff
@@ -3914,6 +3914,7 @@
   "signup.office365": "Office 365",
   "signup.saml.icon": "SAML Icon",
   "signup.title": "Create an account with:",
+  "someting.string": "defaultString",
```
lieut-data, line 3917: Looks like a historical typo? "something"?

lieut-data (approved):
Looks excellent, @hmhealey! Looking forward to breaking this up slowly in the future :)

i18n/en.json
hmhealey: Yeah, it's a historical typo. This string is actually just in a mattermost-redux unit test, not in the actual code, so I'm not too concerned about fixing it right now. We should probably update `i18n-extract` at some point to not look at test files.

calebroseland (approved):
This is going to improve so many workflows! Thanks for all your (and others') work on this. 🔥

nevyangelova:
🎉 🎉 🎉

grundleborg (approved):
Giving a :+1: to the general approach. I haven't even attempted a line-by-line review of the diff.

hmhealey:
> I haven't even attempted a line-by-line review of the diff.

Yeah, I wasn't expecting that at all. Even reviewing the notes I took above was probably unnecessary.

PR #7724, opened 2021-03-18
Fixed edge case: 26 char channel name in URL

levb:
#### Summary
When the channel name in the URL is exactly 26 characters, and the state is fresh - navigation fails, it gets falsely recognized as a channel_id.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32912

### DIFF
--- a/components/channel_layout/channel_identifier_router/actions.ts
+++ b/components/channel_layout/channel_identifier_router/actions.ts
@@ -38,7 +38,18 @@ export function onChannelByIdentifierEnter({match, history}: MatchAndHistory): A
             return {data: undefined};
         }
 
-        const channelPath = getPathFromIdentifier(state, path, identifier);
+        let channelPath = getPathFromIdentifier(state, path, identifier);
+
+        // A channel name that is exactly 26 characters long is recognized by
+        // getPathFromIdentifier as a 'channel_id'. So, if the channel_id does
+        // not exist, try it as a name. See
+        // https://mattermost.atlassian.net/browse/MM-32912.
+        if (channelPath === 'channel_id') {
+            const ch = getChannel(state, identifier.toLowerCase());
+            if (!ch) {
+                channelPath = 'channel_name';
+            }
+        }
 
         switch (channelPath) {
         case 'channel_name':


mickmister (commented):
I think this PR assumes "If the channel with an id that matches the URL isn't currently loaded, then it must be a channel name".

What are the gotcha's here? Why always fall back on `channel_name`, when it could be any of the other options in the switch statement?

levb:
> What are the gotcha's here? Why always fall back on `channel_name`, when it could be any of the other options in the switch statement?

@mickmister See https://github.com/mattermost/mattermost-webapp/blob/eb12f73a071bd0a4b49c4fa5b1cf673ac251521a/components/channel_layout/channel_identifier_router/actions.ts#L81. For a 26-letter identifier the only other possible choice is a `channel_name`, so this shouldn't have any side effects.

It is likely possible to move the logic inside the `getPathFromIdentifier` function itself, but it seemed less "surgical".

mickmister (approved):

larkox (approved):

components/channel_layout/channel_identifier_router/actions.ts
```diff
@@ -38,7 +38,18 @@ export function onChannelByIdentifierEnter({match, history}: MatchAndHistory): A
...
         }
 
-        const channelPath = getPathFromIdentifier(state, path, identifier);
+        let channelPath = getPathFromIdentifier(state, path, identifier);
+
+        // A channel name that is exactly 26 characters long is recognized by
+        // getPathFromIdentifier as a 'channel_id'. So, if the channel_id does
+        // not exist, try it as a name. See
+        // https://mattermost.atlassian.net/browse/MM-32912.
+        if (channelPath === 'channel_id') {
```
hmhealey, line 47: Could this be done in `getPathFromIdentifier` so that we could add a unit test for it?

hmhealey: Actually, looking at `getPathFromIdentifier`, we already have a check there to see if a 26 character identifier is an ID or name, but it's looking for a channel with the given name and giving up if it can't find one. Perhaps it's not looking at the correct team since the repro steps for this bug involve multiple teams?

hmhealey (changes requested):
The only thing that I'd be worried about changing here is that we're losing the ability to join a public channel by navigating to the URL with its ID. That said, I feel like nobody is actually using that functionality even though that's entirely using the force. I don't remember the exact reason that we added the ability to navigate to a channel by its ID, but it seems like it would be rarely used even

wiersgallak:
@levb Can you give an update on what the status of this PR is?

hmhealey:
@wiersgallak The ticket has been moved to Web Platform since the scope of it is larger than originally believed. More specifically, we're worried that fixing this specific case will cause regressions elsewhere. I'd be glad to discuss the issue more, but for the time being, we're not working on this issue.

PR #7726, opened 2021-03-19
[MM-32430] New hybrid input

BenCookie95:
#### Summary
New Dropdown-Input Hybrid for data retention.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32430
Design: https://www.figma.com/proto/oaUgpozKua2RuTnJRqkCeD/Hybrid-Input%2FDropdown?node-id=236%3A111&scaling=min-zoom

#### Screenshots
dropdown-hybrid


components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,262 @@
...
+            showTextInput(inputValue, false);
+        }
+    }, [inputValue]);
+
+    useEffect(() => {
+        if (!inputValue && !focused && !inputFocused) {
+            onDropdownChange(props.defaultValue);
+            showTextInput('');
+        }
+    }, [focused, inputFocused]);
```
hahmadia, line 127: Why are these in seperate use effects? Can't they just be in one?

hahmadia (commented):
Although the code looks good to me, I have some questions.
I will also be adding Michel from the UI platform team to get their UI expertise on this common component creation as I don't think it would be a bad idea.

components/widgets/inputs/dropdown_input_hybrid.tsx
michelengelen: @hahmadia you could do it in one, but it is an anti-pattern in react. It is a separation of concerns. 

Take for example focused and inputFocused. You don't want to run this callback on every change to the inputValue. It will create side-effects that are not easy to handle when combined.

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,262 @@
...
+        display: 'none',
+    }),
+    menuPortal: (provided: CSSProperties) => ({
+        ...provided,
+        zIndex: 99999999,
+    }),
+};
+
+
+const IndicatorsContainer = (props: any) => {
```
michelengelen, line 55: using `any` is strongly discouraged.
Please use as strictly defined types as possible!

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,262 @@
...
+    return (
```
michelengelen, line 56: all of these inline Component definitions can be written with immediate returns

e.g.
```typescript
export const IndicatorsContainer = (props: any) => (
    <div className="DropdownInput__indicatorsContainer">
        <components.IndicatorsContainer {...props}>
            <i className="icon icon-chevron-down" />
        </components.IndicatorsContainer>
    </div>
);
```

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,262 @@
...
+        inputType,
+        defaultValue,
+        ...otherProps
+    } = props;
+
+    const containerRef = useRef<HTMLInputElement>(null);
+    const inputRef = useRef<HTMLInputElement>(null);
+    const [inputFocused, setInputFocused] = useState(false);
+    const [focused, setFocused] = useState(false);
+    const [showInput, setShowInput] = useState(inputValue ? true : false);
```
michelengelen, line 108: you can replace this with either `!!inputValue` (if the linting rules allow that) or `Boolean(inputValue)`

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,262 @@
...
+        }
+        return {};
+    }
+
+    const onInputBlur = () => {
+        setInputFocused(false);
+    };
+    const onInputFocus = () => {
+        setInputFocused(true);
+    };
```
michelengelen, line 147: these can also be simplified, since a function body is not necessary here:

e.g.
```typescript
const onInputBlur = () => setInputFocused(false);
```

michelengelen (changes requested):

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+    const onInputFocus = () => setInputFocused(true);;
+
+    const onDropdownInputFocus = (event: React.FocusEvent<HTMLElement>) => {
+        const {onFocus} = props;
+        
+        setFocused(true);
+
+        if (onFocus) {
+            onFocus(event);
+        }
```
calebroseland, line 147: nit: this (and those below) could be simpler.
```suggestion   
        setFocused(true);
        props.onFocus?.(event);
```

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+    useEffect(() => {
+        if (!inputFocused) {
+            showTextInput(inputValue, false);
+        }
+    }, [inputValue]);
+
+    useEffect(() => {
+        if (!inputValue && !focused && !inputFocused) {
+            onDropdownChange(props.defaultValue);
```
calebroseland, line 118: Below there's an if check for onDropdownChange, should the same apply here?
(and `defaultValue` is destructured above)
```suggestion
            onDropdownChange?.(defaultValue);
```

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+    
+    // We want to show the text input when we have a dropdown value selected and 
+    const showTextInput = (val: string, focus = true) => {
+        if (!val || exceptionToInput.includes(val)) {
+            setShowInput(false);
+        } else {
+            setShowInput(true);
+            if (inputRef.current && focus) {
+                inputRef.current.focus();
+            }
```
calebroseland, line 178: nit: could be simpler
```suggestion
            if (focus) {
                inputRef.current?.focus();
            }
```

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+        className={classNames('DropdownInput__option', {
+            selected: props.isSelected,
+            focused: props.isFocused,
+        })}
+    >
+        <components.Option {...props}/>
+    </div>
+);
+
+const DropdownInputHybrid = <T extends ValueType>(props: Props<T>) => {
```
calebroseland, line 80: I think `T` downstream of `Props` here is undefined (defaults to its built-in/standard `OptionType`) unless T is explicitly provided, so it might be worth tying it in as a default type param either here as:
```suggestion
const DropdownInputHybrid = <T extends ValueType = ValueType>(props: Props<T>) => {
```

Or above in props:
```tsx
type Props<T = ValueType> = Omit<SelectProps<T>, 'onChange'> & {
```

calebroseland (commented):
I tried to fire Storybook up, but I'm getting a few errors with that, so most of my comments are TS-related. Looks great from the screenshots, though!

components/widgets/inputs/dropdown_input_hybrid.tsx
michelengelen: nice catch!

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+    const containerRef = useRef<HTMLInputElement>(null);
+    const inputRef = useRef<HTMLInputElement>(null);
+    const [inputFocused, setInputFocused] = useState(false);
+    const [focused, setFocused] = useState(false);
+    const [showInput, setShowInput] = useState(!!inputValue);
+
+    useEffect(() => {
+        if (showInput && !inputValue && inputRef.current) {
+            inputRef.current.focus();
+        }
```
michelengelen, line 107: nit: this condition could be simplified

```suggestion
        if (showInput && !inputValue) {
            inputRef.current?.focus();
        }
```

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+            return {
+                menu: (provided: CSSProperties) => ({
+                    ...provided,
+                    width: containerRef.current ? `${containerRef.current.offsetWidth}px` : '0px',
+                    left: inputRef.current ? `-${inputRef.current.offsetWidth}px` : '0px',
+                }),
+            };
+        }
+        return {};
+    }
```
michelengelen, line 134: I personally don't like conditional returns (I would rather do it before the `getMenuStyles` invocation to prevent invocation altogether when `showInput` is not set), but in this case it would probably simplify it by adding a ternary instead of a complete if-block

```suggestion
    const getMenuStyles = () =>
        showInput
            ? {
                  menu: (provided: CSSProperties) => ({
                      ...provided,
                      width: containerRef.current ? `${containerRef.current.offsetWidth}px` : '0px',
                      left: inputRef.current ? `-${inputRef.current.offsetWidth}px` : '0px',
                  }),
              }
            : {};
```

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+        const {onBlur} = props;
+
+        setFocused(false);
+
+        if (onBlur) {
+            onBlur(event);
+        }
+    };
+
+    const onValueChange = async (event: T) => {
```
michelengelen, line 160: I think this does not need to be asynchronous

michelengelen (approved):
LGTM ... apart from some codestyle related things

I could not run storybook as well, so all comments are solely based on the code.

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -115,74 +115,60 @@ const DropdownInputHybrid = <T extends ValueType>(props: Props<T>) => {
...
-    const onValueChange = async (event: T) => {
-        const {onDropdownChange} = props;
-
+    const onValueChange = (event: T) => {
         showTextInput(event.value);
 
-        if (onDropdownChange) {
-            onDropdownChange(event);
-        }
+        props.onDropdownChange(event)
```
michelengelen, line 153: this is already destructured from `props` above, so you can just call it directly

```suggestion
        onDropdownChange(event)
```

michelengelen (approved):

components/widgets/inputs/dropdown_input_hybrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+                        id={`DropdownInput_${name}`}
+                        placeholder={focused ? '' : placeholder}
+                        components={{
+                            IndicatorsContainer,
+                            Option,
+                            Control,
+                        }}
+                        className={classNames('Input', className, {Input__focus: showLegend})}
+                        classNamePrefix={'DropDown'}
+                        onChange={onValueChange as any}
```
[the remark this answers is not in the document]
BenCookie95: @calebroseland I've been playing around with this a lot and I haven't been able to type this `onChange` correctly. I removed the async part. Do you have any suggestions?

michelengelen (approved):

hahmadia (approved):

calebroseland (approved):

components/widgets/inputs/dropdown_input_hybrid.tsx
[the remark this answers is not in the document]
calebroseland: @BenCookie95 sorry for the delay in getting back to this. I had done some digging last week and I think there's a version discrepancy that's messing the types up. You probably saw `components/multi_input.tsx` and `components/multiselect/multiselect.tsx` have comments and `any` casting where the types don't line up, so I think it's fine to follow that here and say "we know better than the compiler." 🐵🙈

Perhaps there should be a separate PR to re-sync `@types/react-select` versions, but we're a few versions behind at the moment..

furqanmlk:
Small GIF (400x232)

furqanmlk (approved):

PR #7729, opened 2021-03-20
[MM-32933] Custom Statuses: Implement max length validation on the status modal

Hampusholmstrom:
#### Summary

Includes a component called MaxLengthInput for the QuickModal component. MaxLengthInput provides feedback as a number of exceeded characters allowed and makes the border red. The componted was reused from the following pr to provide visual feedback for setting custom statuses: https://github.com/mattermost/mattermost-webapp/pull/7281

The PR also includes a condition to disable the ability to set a status when exceeding the character count.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/17140

#### Screenshots

set-status


esethna:
Thanks @Hampusholmstrom, sorry for the delayed review here. 

I think the alignment and padding looks off for the number, adding @matthewbirtch for UX review:  
image

esethna (approved):

matthewbirtch (changes requested):
Thanks @esethna. You're right. The alignment is off and there should be some breathing room between the character count and the input text. This is probably due to the difference in size of the input field here vs. the one used in the new category modal.

Need to adjust the following for this specific modal:
```
. MaxLengthInput__validation {
  right: 44px;
  top: 17px;
}
```

and

```
.MaxLengthInput.form-control.has-error {
    padding-right: 76px;
}

michelengelen (approved):
LGTM

michelengelen (approved):

Hampusholmstrom:
> Thanks @esethna. You're right. The alignment is off and there should be some breathing room between the character count and the input text. This is probably due to the difference in size of the input field here vs. the one used in the new category modal.
> 
> Need to adjust the following for this specific modal:
> 
> ```
> . MaxLengthInput__validation {
>   right: 44px;
>   top: 17px;
> }
> ```
> 
> and
> 
> ```
> .MaxLengthInput.form-control.has-error {
>     padding-right: 76px;
> }
> ```

Thank you for the input. I used "!important" to override the properties of the component in the specific modal. Not sure if it was the right way to go but as far as I can see, it works as expected.

matthewbirtch (changes requested):
Thanks @Hampusholmstrom. It looks like the positioning is better now for the character count. However, I noticed the following: 
- When the input field is focused, it looks like the border width gets increased by 1px on the inside of the box.
- This is much more noticeable when the field is in an error state (see gif below).
- We didn't do this on the 'Create New Category' Modal, so I'm wondering where this might have come from. Can we remove the inset border on focus?

michelengelen:
> Thanks @Hampusholmstrom. It looks like the positioning is better now for the character count. However, I noticed the following:
> 
> * When the input field is focused, it looks like the border width gets increased by 1px on the inside of the box.
> * This is much more noticeable when the field is in an error state (see gif below).
> * We didn't do this on the 'Create New Category' Modal, so I'm wondering where this might have come from. Can we remove the inset border on focus?
>   

The part in question is line 45 in this file: `components/custom_status/custom_status.scss`

Hampusholmstrom:
> * We didn't do this on the 'Create New Category' Modal, so I'm wondering where this might have come from. Can we remove the inset border on focus?

I removed the "on focus" border so the text box looks better now.

Willyfrog (approved):
lgtm thanks!

michelengelen (approved):

ogi-m (approved):
LGTM 👍🏻

PR #7733, opened 2021-03-22
MM-33748 Support for new mention_count_root and msg_count_root

reflog:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-33748


components/sidebar/sidebar_channel/index.ts
```diff
@@ -26,16 +27,17 @@ function makeMapStateToProps() {
...
         let unreadMentions = 0;
         let unreadMsgs = 0;
         let showUnreadForMsgs = true;
         if (member) {
-            unreadMentions = member.mention_count;
+            unreadMentions = collapsed ? member.mention_count_root : member.mention_count;
 
             if (channel) {
-                unreadMsgs = Math.max(channel.total_msg_count - member.msg_count, 0);
+                unreadMsgs = collapsed ? Math.max(channel.total_msg_count_root - member.msg_count_root, 0) : Math.max(channel.total_msg_count - member.msg_count, 0);
```
calebroseland, line 40: I think there were some requests on the other PR to make this logic into a selector. Would that be possible?

components/sidebar/sidebar_channel/sidebar_channel_link/index.ts
```diff
@@ -30,16 +29,12 @@ function mapStateToProps(state: GlobalState, ownProps: OwnProps) {
...
-
-        if (isCollapsedThreadsEnabled(state)) {
-            const threadMentionCountInChannel = getThreadCountsInCurrentTeam(state)?.unread_mentions_per_channel?.[ownProps.channel.id] || 0;
-            unreadMentions -= threadMentionCountInChannel;
-        }
+        unreadMentions = collapsed ? member.mention_count_root : member.mention_count;
 
         if (ownProps.channel) {
-            unreadMsgs = Math.max(ownProps.channel.total_msg_count - member.msg_count, 0);
+            unreadMsgs = collapsed ? Math.max(ownProps.channel.total_msg_count_root - member.msg_count_root, 0) : Math.max(ownProps.channel.total_msg_count - member.msg_count, 0);
```
calebroseland, line 37: same here

calebroseland (commented):
There seems to be a discrepancy between channels and the summarized team and favicon/title mention counts. 

Here, the 2 mentions seem correct in sidebar_channel (they are root-post mentions), but there are an extra 35 thread-reply mentions that don't seem to affect the team and title level counts (followed threads that are read/unread don't seem to affect these counts.)

CleanShot 2021-04-07 at 16 14 17

components/legacy_team_sidebar/legacy_team_sidebar_controller.tsx
```diff
@@ -227,10 +228,10 @@ export default class LegacyTeamSidebar extends React.PureComponent<Props, State>
                     tip={team.display_name}
                     active={team.id === this.props.currentTeamId}
                     displayName={team.display_name}
-                    unread={member.msg_count > 0}
+                    unread={this.props.collapsedThreads ? member.msg_count_root > 0 : member.msg_count > 0}
                     order={index + 1}
                     showOrder={this.state.showOrder}
-                    mentions={member.mention_count}
+                    mentions={this.props.collapsedThreads ? member.mention_count_root : member.mention_count}
```
calebroseland, line 234:
```suggestion
                    mentions={this.props.collapsedThreads ? (member.mention_count_root + this.props.threadCounts?.[team.id]?.total_unread_mentions) : member.mention_count}
```

esethna:
@ogi-m please help prioritize this PR by smoke testing with the feature flag for collapsed reply threads off.

esethna:
Ping @hmhealey on reviewing this asap please, thank you!

calebroseland:
@reflog @hmhealey @ogi-m After firing this up again, I believe there's one UX issue remaining for multi-team users—only the current team loads initial thread counts (unreads/mentions), and that can produce inaccurate mention counts in the Team Icons and Title (`favicon_title_handler`).

hmhealey:
@reflog @esethna I'm not able to review this right now since it looks like it needs to be rebased onto master. I assume it's not actually 7000 lines added at this point

reflog:
@hmhealey sorry about that. A large PR that this was based on got merged yesterday.
This PR is ready for review now. Thanks in advance

esethna:
Ping @calebroseland @hmhealey @ogi-m please help prioritize reviews here to unblock us starting to test the feature on daily

calebroseland:
After a couple passes and some fixes, there're 3 items left on my radar:

1. Initial Load of multi-team Thread Counts (`total_unread_replies` and `total_unread_mentions` in `ThreadsState['counts']`).
 These values pipe into favicon/title indicators and team icon indicators. With only current-team counts loading via GlobalThreadsLink, other-team values are not yet initialized and the indicators will be inaccurate. 
 WS events will increment other-team values after init load, but we have yet to get them initialized.
--
I tried to do a quick "80%"/beta-quality/hacked version using the existing `getThreads` api/action, but repeating this on a per-team basis in user actions `loadMe` and `loginComplete` is heavy. After a quick try, it seems it would be probably a clunky implementation.
--
I think we should add it to the existing `Client4.getMyTeamUnreads` and its mattermost-server counterpart, and then watch for `RECEIVED_MY_TEAM_UNREADS` in the [`ThreadsState['counts']` reducer](https://github.com/mattermost/mattermost-webapp/blob/MM-33748/packages/mattermost-redux/src/reducers/entities/threads.ts#L121) to initialize other-team thread counts.

2. Favicon/Title unread indicator (via [`getUnreads`](https://github.com/mattermost/mattermost-webapp/blob/MM-33748/packages/mattermost-redux/src/selectors/entities/channels.ts#L478)) needs to factor in `total_unread_replies` in the same way it already factors in `total_unread_mentions`. (Currently, I believe the unreads dot behaves as though the feature is off.)
CleanShot 2021-04-14 at 23 44 53@2x


3. The team icon indicators need to subtract _**DM/GM-channel** thread-reply unread reply/mention counts_, otherwise this happens as you click through other teams.
CleanShot 2021-04-14 at 23 13 49@2x

calebroseland (commented):
After a couple passes and some fixes, there're 3 (feature flag on) items left on my radar as of now.

1. (Separate ticket/PR?) Initial Load of other-team Thread Counts (`total_unread_replies` and `total_unread_mentions` in `ThreadsState['counts']`).
 These values pipe into favicon/title indicators and team icon indicators. With only current-team counts loading via GlobalThreadsLink, other-team values are not yet initialized and the indicators may be inaccurate. 
 (WS events will increment other-team values after init load—but the initial values need to be loaded)
--
I tried to do a quick "80%"/beta-quality/hacked version using the existing `getThreads` api/action, but repeating this on a per-team basis in user actions `loadMe` and `loginComplete` is heavy. After a quick try, it seems it would be probably a clunky implementation.
--
I think we should add them to the existing `Client4.getMyTeamUnreads` and its mattermost-server counterpart, and then watch for `RECEIVED_MY_TEAM_UNREADS` in the [`ThreadsState['counts']` reducer](https://github.com/mattermost/mattermost-webapp/blob/MM-33748/packages/mattermost-redux/src/reducers/entities/threads.ts#L121) to initialize other-team thread counts.

2. Favicon/Title unread-dot indicator (via [`getUnreads`](https://github.com/mattermost/mattermost-webapp/blob/MM-33748/packages/mattermost-redux/src/selectors/entities/channels.ts#L478)) needs to factor in `total_unread_replies` in the same way it already factors in `total_unread_mentions`. (Currently, I believe the unreads dot behaves as though the feature is off.)
CleanShot 2021-04-14 at 23 44 53@2x


3. The team icon indicators need to subtract _**DM/GM-channel** thread-reply unread reply/mention counts_ (phew 😮), otherwise the following happens as you click through other teams. I'm not sure yet where to source these values. We may be able to write a new selector and do some filtering, but I think we need to discuss this part.
CleanShot 2021-04-14 at 23 13 49@2x

components/permalink_view/permalink_view.test.jsx
```diff
@@ -178,6 +178,7 @@ describe('components/PermalinkView', () => {
             });
 
             test('should redirect to DM link with postId for permalink', async () => {
+                Date.now = () => new Date(0).getMilliseconds();
```
hmhealey, line 181: Minor nit, but this will affect other test cases in this file as well. I don't think it'll cause any problems, but it's something to be aware of. Perhaps we should move it up to the start of the file so it's more obvious that it affects multiple cases instead of just affecting the tests after this one.

packages/mattermost-redux/src/selectors/entities/general.ts
```diff
@@ -11,7 +11,7 @@ import {ClientConfig, FeatureFlags} from 'mattermost-redux/types/config';
 import {isMinimumServerVersion} from 'mattermost-redux/utils/helpers';
 
 export function getConfig(state: GlobalState): Partial<ClientConfig> {
-    return state.entities.general.config;
+    return state.entities.general?.config;
```
hmhealey, line 14: Why add the `?` here? When would `state.entities.general` ever be undefined?

packages/mattermost-redux/src/selectors/entities/preferences.ts
```diff
@@ -18,7 +17,7 @@ import {getPreferenceKey} from 'mattermost-redux/utils/preference_utils';
 import {setThemeDefaults} from 'mattermost-redux/utils/theme_utils';
 
 export function getMyPreferences(state: GlobalState): { [x: string]: PreferenceType } {
-    return state.entities.preferences.myPreferences;
+    return state.entities.preferences?.myPreferences ?? {};
```
hmhealey, line 20: Same question here. Also, we shouldn't be returning object literals from non-memoized selectors

packages/mattermost-redux/src/selectors/entities/preferences.ts
```diff
@@ -112,7 +111,7 @@ export const getTeammateNameDisplaySetting: (state: GlobalState) => string = cre
 
 const getThemePreference = createSelector(
     getMyPreferences,
-    getCurrentTeamId,
+    (state) => state.entities.teams.currentTeamId,
```
hmhealey, line 114: Why is this changing?

packages/mattermost-redux/src/actions/threads.ts
```diff
@@ -94,21 +75,21 @@ export function handleThreadArrived(dispatch: DispatchFunc, getState: GetStateFu
         type: ThreadTypes.RECEIVED_THREAD,
         data: {
             thread,
-            team_id: teamId,
+            team_id: teamId || currentTeamId,
```
hmhealey, line 78: When would we ever receive a websocket event for a thread where we didn't also receive a team ID?

packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -506,20 +513,24 @@ export const getUnreads: (state: GlobalState) => {
                 otherUserId = getUserIdFromChannelName(currentUserId, channel.name);
 
                 if (users[otherUserId] && users[otherUserId].delete_at === 0) {
-                    mentionCountForCurrentTeam += m.mention_count;
+                    mentionCountForCurrentTeam += (collapsed ? m.mention_count_root : m.mention_count);
+                }
+            } else if (channel.delete_at === 0) {
+                if (m.mention_count > 0) {
```
hmhealey, line 519: I thought our ESLint rules would complain about this pattern, but I guess it's fine with it as long as the outer one is an else-if. Regardless of ESLint, I definitely prefer this since it makes it easier to understand the thinking here.

hmhealey (changes requested):

packages/mattermost-redux/src/selectors/entities/general.ts
reflog: @hmhealey during testing. A incomplete store is being mocked and tests crash

packages/mattermost-redux/src/selectors/entities/preferences.ts
reflog: @hmhealey to avoid import loop caused by getCurrentTeamId import

components/permalink_view/permalink_view.test.jsx
calebroseland: Would `import {fakeDate} from 'tests/helpers/date';` work?

```ts
    let resetFakeDate: () => void;

    beforeEach(() => {
        resetFakeDate = fakeDate(new Date(...));
    });

    afterEach(() => {
        resetFakeDate();
    });
```

packages/mattermost-redux/src/actions/threads.ts
calebroseland: DM/GM `thread_updated` websocket events don't have a teamId since DM/GM channels travel with the current-team.

components/permalink_view/permalink_view.test.jsx
calebroseland: (Or keep and restore the original reference before/after)

reflog:
@hmhealey I've address your comments. Please re-review.
@ogi-m please approve

ogi-m (approved):

packages/mattermost-redux/src/selectors/entities/general.ts
hmhealey: If it's trying to access the config in that test though, shouldn't the config be defined then? Regardless, I'm only 0/5 on that since I know redux testing is a pain. We really need an equivalent of the web app's `testConfigureStore` to fill in all the blanks for us

packages/mattermost-redux/src/selectors/entities/preferences.ts
hmhealey: Ah, I see. We could also solve that by moving `getCurrentTeamId` into `selectors/common` like we did with `getCurrentUserId` and some others before, but I'm good with it either way.

hmhealey (approved):

PR #7736, opened 2021-03-22
Proposal/Demo: Introduce EPICS to handle side effects from redux actions

AGMETEOR:
#### Summary
[Motivation Ticket 1](https://mattermost.atlassian.net/browse/MM-33132?atlOrigin=eyJpIjoiMDg1Y2Q1NTdkMGE5NDJiYjhmNGNiMGUyZWY3MTU3Y2EiLCJwIjoiaiJ9)
[Motivation Ticket 2](https://mattermost.atlassian.net/browse/MM-33374?atlOrigin=eyJpIjoiYWE2NjFiMTkxYzE3NDQwMjg3N2IyYmM0NTNmZmIzMzgiLCJwIjoiaiJ9)

#### Example 1
In Motivation Ticket 2,  we had a bug caused by this flow. First, at boot time, if we had a cloud licence we would load `subscriptionStats` [here](https://github.com/mattermost/mattermost-webapp/blob/eb12f73a071bd0a4b49c4fa5b1cf673ac251521a/actions/views/root.js#L28). The bug occurred when, at boot time, the license was not cloud, and then later a user witches the license to cloud. This means there'll be components expecting to use ```subscriptionStats``` but they were not loaded. The components were nolonger working properly. To fix, we might have to go into each of the components and in ```compnentDidMount``` or ```useEffect``` hook check whether the subscriptionStats exist and then fetch them.

A more compact and predictable solution would be to listen to redux events and then react to them. In this demo PR, I use ```redux-observable``` library and ```rxjs``` to listen to ```UPLOAD_LISENCE_SUCCESS``` redux action and then as a side effect trigger fetching of ```subscriptionStats``` if the uploaded lisence is a Cloud one.

In Motivation Ticket 1, the solution was to listen to a webhook event and then update the data. This can also easily be solved with an appropriate epic.

#### Ticket Link
None

#### Similar
redux sagas


lieut-data:
@AGMETEOR, I apologize for the delay on reviewing this PR. There have been a handful of feature branches merging to `master` in the past week that have mostly consumed our reviewing bandwidth. I haven't forgotten about this one, but it will be a while before it gets my full attention :)

I haven't looked at the code yet, but I'm excited to see a concrete proposal for side effects. I'm wondering if you can share context into why you picked `redux-observable` and `rxjs`, and if you have any experience contrasting these libraries with other side effect libraries such as `redux-saga`? We'll end up needing to make a decision about which library to consume, as we'll probably not ship more than one library. Although my experience (and leaning) has been with `redux-saga`, I'd love to be able to compare and contrast others' experiences to make the optimal decision for the team.

AGMETEOR:
> Thanks, @AGMETEOR! I think this is a really cool example of the power of using the Redux bus for side effects.
> 
> I'm personally not terribly familiar with the rxjs paradigms, preferring redux-saga myself, but the concepts are pretty much identical. There might even be space for both to co-exist.
> 
> I'm not actively working on the Web Platform right now, so I'll leave it to you to decide next steps with @hmhealey and the team. Requesting changes "for now" as such.

Hey @lieut-data thanks for taking a look at this. I think we should go with something a lot of people are comfortable with. Here, I was just trying to elaborate the concept but I think both redux-saga and epics do literally the same thing. Let's see what Harrison has to say :)

hmhealey:
Unfortunately, no one on Web Platform is working on this right now, so I haven't really been able to invest any time in understanding either of them. I've used things like generators a few times before, so that's why I've been suggesting sagas, but rxjs also sounds interesting if I were to spend some time reading into it.

I don't think we'll be looking into either of them ourselves on Web Platform until late June at the earliest, so I'd be open to the idea of trying epics out for the time being to see if they work or not for what we need them for. I don't imagine either epics or sagas will fully replace thunks, so I think we should be able to try them out and if they don't happen to work out or we decide something else is better, we can always change them later.

lieut-data:

PR #7741, opened 2021-03-22
MM-25140 Fix unit tests which are broken on Node 14

hmhealey:
These tests are broken on Node 14 due to it running some async code differently. I fixed it by having the modals' `handleSubmit` methods actually wait for their results (which shouldn't affect the UI since it doesn't await on `handleSubmit`) and by figuring out how to properly mock the `PQueue` used by the `DataPrefetch` component. That one was a lot trickier since the tests did a lot of bizarre stuff like waiting on functions they'd mocked which didn't do anything other than letting other awaited calls proceed. With the mock, we can now manually pop and run queued tasks instead of having them run asynchronously at poorly-defined times.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25140

#### Related Pull Requests

### DIFF
--- a/components/data_prefetch/data_prefetch.test.tsx
+++ b/components/data_prefetch/data_prefetch.test.tsx
@@ -12,6 +12,13 @@ import {TestHelper} from 'utils/test_helper';
 
 import DataPrefetch from './data_prefetch';
 
+const mockQueue: Array<() => Promise<void>> = [];
+
+jest.mock('p-queue', () => class PQueueMock {
+    add = (o: () => Promise<void>) => mockQueue.push(o);
+    clear = () => mockQueue.splice(0, mockQueue.length);
+});
+
 jest.mock('actions/user_actions.jsx', () => ({
     loadProfilesForSidebar: jest.fn(() => Promise.resolve({})),
 }));
@@ -24,8 +31,7 @@ describe('/components/data_prefetch', () => {
             trackDMGMOpenChannels: jest.fn(() => Promise.resolve()),
         },
         prefetchQueueObj: {
-            1: ['mentionChannel'],
-            2: ['unreadChannel'],
+            1: [],
         },
         prefetchRequestStatus: {},
         sidebarLoaded: true,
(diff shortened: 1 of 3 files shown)


components/data_prefetch/data_prefetch.test.tsx
```diff
@@ -66,124 +72,212 @@ describe('/components/data_prefetch', () => {
...
+            <DataPrefetch {...props}/>,
         );
-        const instance = wrapper.instance();
-        instance.prefetchPosts = jest.fn();
+
+        expect(props.actions.trackDMGMOpenChannels).not.toHaveBeenCalled();
+
+        // Change channels
+        wrapper.setProps({currentChannelId: 'currentChannelId'});
+        await Promise.resolve(true);
```
devinbinnie, line 111: I'm guessing there's no other way of waiting for the component to update and ensure that the function has been called?

devinbinnie (approved):

components/data_prefetch/data_prefetch.test.tsx
hmhealey: The component is updating fine, but its `componentDidUpdate` only runs up until it awaits a call ([link](https://github.com/mattermost/mattermost-webapp/blob/master/components/data_prefetch/data_prefetch.tsx#L58)). After that, it waits for the test to pause again for its chance to run. When the test awaits on anything at all, it lets `componentDidUpdate` continue running. The previous version awaited random mocked functions, so I replaced those with dummy promises to be clearer that it's just to let other code proceed.

mkraft (approved):

hmhealey:
@jgilliam17 For testing this, can you confirm that the Add Groups to Channel and Add Groups to Team modals in the system console still close and submit properly? The other changes are part of actual unit tests.

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified Add Groups to Channel and Add Groups to Team modals in the system console close and submit properly.
Note: Saw an error when removing a group from team, reproduces on master, filing separately.

PR #7744, opened 2021-03-23
MM-18107: Precursory refactoring and supporting code

calebroseland:
#### Summary
- `makeAsyncComponent` can now be provided with a loader or other fallback content (no testing req'd)
- `<NoResultsIndicator/>` becomes more flexible and can be supplied with content without hardcoding a new `NoResultsVariant`, and an icon color is fixed
- Change some implicit `currentChannelId` to explicit `channelId` for components used in `<ThreadViewer/>` (Testing detail: Profile Popover in center channel and RHS)
- Extract `<RHSSearchNav`/> for reuse in Global Threads header

#### Screenshots
CleanShot 2021-03-22 at 22 56 04@2x
CleanShot 2021-03-22 at 23 23 19@2x
CleanShot 2021-03-23 at 15 27 25@2x


calebroseland:
@lieut-data not sure who you'd like to review, so please assign as you see fit

actions/views/profile_popover.js
```diff
@@ -8,21 +8,21 @@ import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
 
 import {getSelectedPost} from 'selectors/rhs';
 
-export function getMembershipForCurrentEntities(userId) {
+export function getMembershipForCurrentEntities(userId, channelId) {
```
lieut-data, line 11: It looks like `channelId` is optional here, but it seems to only be called in a single place. Can we just make it required instead and skip the `defaultChannelId` logic altogether?

actions/views/profile_popover.js
```diff
@@ -8,21 +8,21 @@ import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
...
 
-        let channelId;
+        let defaultChannelId;
         if (selectedPost.exists === false) {
-            channelId = currentChannel.id;
+            defaultChannelId = currentChannel.id;
         } else {
-            channelId = selectedPost.channel_id;
+            defaultChannelId = selectedPost.channel_id;
         }
```
Willyfrog, line 24: minor nitpick, mostly to avoid a negative condition and to have a `const` instead of a `let` if there is no need to change it.
```js
const defaultChannelId = selectedPost.exists ? selectedPost.channel_id : currentChannel.id;
```

components/file_upload/file_upload.jsx
```diff
@@ -75,9 +75,9 @@ class FileUpload extends PureComponent {
     static propTypes = {
 
         /**
-         * Current channel's ID
+         * ChannelID
```
lieut-data, line 78: nit: can we just remove this comment altogether?

components/channel_header/channel_header.js
```diff
@@ -1,6 +1,5 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-/* eslint-disable react/no-string-refs */
```
Willyfrog, line 3: thanks for taking care of this, seems to be a leftover we overlooked when reviewing the string refs campaing 👍

components/file_upload/index.js
```diff
@@ -13,12 +13,12 @@ import {canUploadFiles} from 'utils/file_utils';
...
 import FileUpload from './file_upload.jsx';
 
-function mapStateToProps(state) {
+function mapStateToProps(state, ownProps) {
     const config = getConfig(state);
     const maxFileSize = parseInt(config.MaxFileSize, 10);
 
     return {
-        currentChannelId: getCurrentChannelId(state),
+        channelId: ownProps.channelId ?? getCurrentChannelId(state),
```
lieut-data, line 21: Is it feasible to eliminate this fallback and always pass in `channelId`?

components/no_results_indicator/no_results_indicator.tsx
```diff
@@ -16,19 +17,25 @@ import SearchIcon from 'components/widgets/icons/search_icon';
...
     subtitleValues?: Dictionary<ReactNode>;
+    style?: CSSProperties;
 }
 
 const iconMap: {[key in NoResultsVariant]: React.ReactNode } = {
     [NoResultsVariant.ChannelSearch]: <SearchIcon className='no-results__icon'/>,
     [NoResultsVariant.Mentions]: <MentionsIcon className='no-results__icon'/>,
     [NoResultsVariant.FlaggedPosts]: <FlagIcon className='no-results__icon'/>,
     [NoResultsVariant.PinnedPosts]: <PinIcon className='no-results__icon'/>,
+    [NoResultsVariant.PinnedPosts]: <PinIcon className='no-results__icon'/>,
```
lieut-data, line 35: This seems to be duplicated?

components/no_results_indicator/no_results_indicator.tsx
```diff
@@ -58,30 +65,63 @@ const subtitleMap: {[key in NoResultsVariant]: MessageDescriptor } = {
...
 
-const NoResultsIndicator = (props: Props) => {
-    return (
+import './no_results_indicator.scss';
+
+const NoResultsIndicator = ({
+    expanded,
+    style,
+    variant,
+    iconGraphic = variant ? (
```
lieut-data, line 74: Can we inline these instead of writing the defaults in this way?

components/profile_popover/index.js
```diff
@@ -39,15 +39,19 @@ function mapStateToProps(state, ownProps) {
...
-    let channelId;
-    if (selectedPost.exists === false) {
-        const currentChannel = getCurrentChannel(state) || {};
-        channelId = currentChannel.id;
-    } else {
-        channelId = selectedPost.channel_id;
+
+    let channelId = ownProps.channelId;
+
+    if (!channelId) {
```
lieut-data, line 45: Is it feasible to eliminate this and expect `channelId` to always be supplied?

lieut-data (changes requested):
Looks good -- just a handful of comments.

components/profile_popover/index.js
```diff
@@ -39,15 +39,19 @@ function mapStateToProps(state, ownProps) {
...
+        if (selectedPost.exists === false) {
+            const currentChannel = getCurrentChannel(state) || {};
+            channelId = currentChannel.id;
+        } else {
+            channelId = selectedPost.channel_id;
```
Willyfrog, line 50: can we have a positive condition instead? it is usually easier to read for people
```suggestion
        if (selectedPost.exists) {
            channelId = selectedPost.channel_id;
        } else {
            channelId = getCurrentChannel(state)?.id;
```

Willyfrog (changes requested):
looking good, some minor nitpics on how to handle some of the logic around defaults

actions/views/profile_popover.js
calebroseland: I was focusing on least disruptive here, and didn't realize it was only called in one place. We have `currentTeamId` there already, so we don't even need a lot of this—thanks for bringing it up!

components/file_upload/index.js
calebroseland: Yep, looks like we have ready-access to the channel in create_post, so no need for the getCurrentChannelId..

components/no_results_indicator/no_results_indicator.tsx
calebroseland: The multiple props/attrs on the `FormattedMessage`s force multiline per eslint, so I wrapped them for consistency. And I opted for defaults vs intl.formatMessage() to simplify the func/render body, but if there's another way to format them, I don't have much preference either way.

components/profile_popover/index.js
calebroseland: `AtMention` was the hold up. Took a bit, but I think we should be good now without this logic. Thanks for catching—an upstream refactor I just committed should prevent a situation in GlobalThreads ThreadViewer where the channelId to `ProfilePopover` (of an `AtMention`) would've been undefined (resulting in a soft fail and just wouldn't display the 'Channel Admin' badge. Really minor but good to catch now.

components/profile_popover/index.js
calebroseland: Would've used this, but N/A anymore per above

lieut-data (approved):
Thanks, @calebroseland!

Willyfrog (approved):

angeloskyratzakos:
@calebroseland There was a panic for that Cloud workspace. 
caller: `plugin/hclog_adapter.go:54` 
error: `panic: runtime error: invalid memory address or nil pointer dereference`

calebroseland:
@angeloskyratzakos are there any other debug messages/traces/call stacks? Is it specific to just this and the #7175 workspace? Or were any other servers showing this?

ogi-m (approved):
Tested and passed:
- Search, @mentions, saved posts and empty states
- help menu
- channel switcher keyboard shortcut
- profile popover

PR #7755, opened 2021-03-24
[MM-20426] Migrate 'components/post_markdown' module and associated tests to TypeScript

majidsajadi:
#### Summary

migrate post markdown to typescript

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/16148


devinbinnie:
Thanks for the PR @majidsajadi, could I ask that you merge `master` into your branch and fix any conflicts before I perform a review? Thanks again!

majidsajadi:
@devinbinnie done. but some tests still fail. I don't think these tests are relevant to my changes.

devinbinnie:
@majidsajadi Looks like the snapshots for the unit tests are missing, since the files were renamed. You can fix that by running `npm run test -- -u` and committing the results.

majidsajadi:
@devinbinnie I run the script and the old snapshot is removed now. but no new snapshot created. already exists in previous commits.

components/post_markdown/post_markdown.tsx
```diff
@@ -1,56 +1,56 @@
...
+
+    /*
+     * The post text to be rendered
+     */
+    message: string;
+
+    /*
+     * The optional post for which this message is being rendered
+     */
+    post?: Partial<Post>;
```
devinbinnie, line 35: Any particular reason that this is `Partial`?

components/post_markdown/post_markdown.tsx
```diff
@@ -1,56 +1,56 @@
...
+    channel?: Channel;
+    options?: TextFormattingOptions;
+    pluginHooks?: Array<Record<string, any>>;
+
+    /**
+     * Whether or not to place the LinkTooltip component inside links
+     */
+    hasPluginTooltips?: boolean;
+    isUserCanManageMembers?: boolean;
+    mentionKeys?: MentionKey[];
```
devinbinnie, line 50: This is required in the original component.

components/post_markdown/system_message_helpers.tsx
```diff
@@ -376,11 +380,11 @@ const systemMessageRenderers = {
     [Posts.POST_TYPES.ME]: renderMeMessage,
 };
 
-export function renderSystemMessage(post, channel, isUserCanManageMembers) {
+export function renderSystemMessage(post: Post, channel?: Channel, isUserCanManageMembers?: boolean): ReactNode {
```
devinbinnie, line 383: Can this ever actually be `undefined`?

devinbinnie (changes requested):

components/post_markdown/system_message_helpers.tsx
majidsajadi: since the channel is the optional props in the `PostMarkdown` we need to either make this `channel` optional or make the `channel` props in the component required.

components/post_markdown/post_markdown.tsx
majidsajadi: some of the `post` mocks in the test don't satisfy the `Post` type required properties.

components/post_markdown/index.ts
```diff
@@ -31,17 +37,34 @@ export function makeGetMentionKeysForPost() {
     );
 }
 
+type OwnProps = {
+    channelId?: string;
```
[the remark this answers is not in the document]
majidsajadi: This is a little bit complicated. 

in the `Textbox` component, we called the `PostMarkdown` and we don't pass the `post`. so the `getMentionKeyForPost` has a type error since the post is required. so if I make the `post` optional I should change the `makeGetMentionKeysForPost`. instead i change the logic in `mapStateToProps`.


also in the Textbox component, the `channelId` prop is optional so the `channelId` here may be `undefined`. 

about the mentionKeys, some components are providing mentionKeys as props and we also make mention key in mapStateToProps so which one should be passed to the component.

components/post_markdown/post_markdown.tsx
majidsajadi: this is related to the first comment.

majidsajadi:
@devinbinnie Thanks for review, left a few comments.

components/post_markdown/post_markdown.tsx
devinbinnie: We shouldn't be basing the typings of the component on test requirements. Can we use `TestHelper.getPostMock(post: Partial<Post>)` and pass in the `Partial<Post>` we're trying to mock so that we have the full object?

components/post_markdown/system_message_helpers.tsx
devinbinnie: We can make `channel` in the component props required, as it was originally always passed in by redux. I think we might need to make `channelId` required as well for `OwnProps` in the connector.

components/post_markdown/index.ts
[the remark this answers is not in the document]
devinbinnie: In `getMentionKeysForPost`, we are already coalescing on the `post` object in case it is null. If it is, we return just the value of `getCurrentUserMentionKeys`. So I think the `post` parameter in that selector should be optional and we should restore the logic in `mapStateToProps` to what it originally was.

As for `channelId`, every use of `Textbox` currently passes in `channelId`, so it should always be defined. It's possible when that component was converted that we should have made `channelId` required, but in this case it might make sense to simply assert the channelId parameter as passed into `PostMarkdown` from `Textbox` so that it's always defined.

components/post_markdown/post_markdown.tsx
majidsajadi: Yes. but I didn't want to change any test logic. I'm not totally on board with the logic yet. I will update the tests.

components/post_markdown/system_message_helpers.tsx
```diff
@@ -194,8 +198,8 @@ function renderHeaderChangeMessage(post) {
                         username,
                         old: oldHeader,
                         new: newHeader,
-                        strong: (chunks) => (<strong>{chunks}</strong>),
-                        br: (x) => (<><br/>{x}</>),
+                        strong: (chunks: Record<string, any>) => (<strong>{chunks}</strong>),
+                        br: (x: Record<string, any>) => (<><br/>{x}</>),
```
hmhealey, line 202: The types for both of these can be `React.ReactNode` instead of using a `Record`

hmhealey (changes requested):
Thanks for the PR and helping with this migration, @majidsajadi! I just have one suggestion for a better type, but otherwise, this looks good to me.

components/post_markdown/system_message_helpers.tsx
majidsajadi: @devinbinnie I have tried this but all the tests have type errors. since the channel we assume we have is handled in the `index` container and the tests imported the component itself. even `TestHelper.getPostMock` doesn't handle `channelId` properly.

devinbinnie: Since in `OwnProps` we are assuming that `channelId` is defined, I'm not sure why we can't assume it's defined in the component as well. If it's just the tests that are breaking, can we make sure that we always pass some default `channelId` in every test? Doesn't matter what the value is if the test doesn't care.

hmhealey (approved):

devinbinnie (approved):
Thanks @majidsajadi, this is looking good :)

saturninoabril (approved):
Thanks @majidsajadi! Tested and passed (Cypress/manual). LGTM.

PR #7760, opened 2021-03-25
MM-34235 - remove trailing slash causing nav bar to hide

pablovelezvidal:
#### Summary
remove trailing slash causing nav bar to hide

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34235


marianunez (approved):
The easiest bug fix ever! 🤣 Thanks @pablovelezvidal!

marianunez:
@lieut-data Just an FYI - you may find this behavior interesting or know why this was happening

pablovelezvidal:
> The easiest bug fix ever! 🤣 Thanks @pablovelezvidal!

All week investigating this bug 😢

lieut-data:
Ah, yeah, we ran into this before and did the same fix instead of addressing the root cause. (I never imagined another entry point ;))

Should work a treat, but hopefully we get back to the root cause in our plugin later.

nickmisasi (approved):

PR #7766, opened 2021-03-26
MM-34114 - grab users stats on demand to fix invite buttons logic

pablovelezvidal:
#### Summary
A bug was found in the logic for when to show the invite people to workspace or show the invite members to workspace. The problem was the in the initial implementation, the state.entities.admin.analytics.TOTAL_USER value was used, and it worked fine, but just for admins. This pull requests fixes that fetching on demand the total_users using the redux getTotalUsersStats action function.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34114

#### Screenshots

The totalUsers shown for an admin user:
Captura de pantalla 2021-03-26 a las 16 10 25

The totalUsers shown for a NON admin user:
Captura de pantalla 2021-03-26 a las 16 11 01

After Adding more users than the usersLimit, now the button changes:
Captura de pantalla 2021-03-26 a las 16 26 52

Same behavior for a NON admin user:
Captura de pantalla 2021-03-26 a las 16 57 42

### DIFF
--- a/components/post_view/channel_intro_message/add_members_button.tsx
+++ b/components/post_view/channel_intro_message/add_members_button.tsx
@@ -30,13 +30,13 @@ export interface AddMembersButtonProps {
     channel: Channel;
     setHeader: React.ReactNode;
     theme: any;
-    showAddMemberButton: boolean;
 }
 
-const AddMembersButton: React.FC<AddMembersButtonProps> = ({totalUsers, usersLimit, channel, setHeader, theme, showAddMemberButton}: AddMembersButtonProps) => {
+const AddMembersButton: React.FC<AddMembersButtonProps> = ({totalUsers, usersLimit, channel, setHeader, theme}: AddMembersButtonProps) => {
     const isPrivate = channel.type === Constants.PRIVATE_CHANNEL;
     const inviteUsers = totalUsers < usersLimit;
-    if (!showAddMemberButton) {
+
+    if (!totalUsers) {
         return (<LoadingSpinner/>);
     }
     return (
--- a/components/post_view/channel_intro_message/add_members_button.tsx
+++ b/components/post_view/channel_intro_message/add_members_button.tsx
@@ -20,6 +20,8 @@ import * as Utils from 'utils/utils.jsx';
 
 import './add_members_button.scss';
 
+import LoadingSpinner from 'components/widgets/loading/loading_spinner';
+
 import MembersSvg from './members_illustration.svg';
 
 export interface AddMembersButtonProps {
@@ -28,12 +30,15 @@ export interface AddMembersButtonProps {
     channel: Channel;
     setHeader: React.ReactNode;
     theme: any;
+    showAddMemberButton: boolean;
 }
 
-const AddMembersButton: React.FC<AddMembersButtonProps> = ({totalUsers, usersLimit, channel, setHeader, theme}: AddMembersButtonProps) => {
+const AddMembersButton: React.FC<AddMembersButtonProps> = ({totalUsers, usersLimit, channel, setHeader, theme, showAddMemberButton}: AddMembersButtonProps) => {
     const isPrivate = channel.type === Constants.PRIVATE_CHANNEL;
     const inviteUsers = totalUsers < usersLimit;
-
+    if (!showAddMemberButton) {
+        return (<LoadingSpinner/>);
+    }
     return (
         inviteUsers && !isPrivate ? lessThanMaxFreeUsers(setHeader, theme) : moreThanMaxFreeUsers(channel, setHeader)
     );
@@ -133,4 +138,4 @@ const moreThanMaxFreeUsers = (channel: Channel, setHeader: React.ReactNode) => {
     );
 };
 
-export default AddMembersButton;
+export default React.memo(AddMembersButton);
--- a/components/post_view/channel_intro_message/channel_intro_message.tsx
+++ b/components/post_view/channel_intro_message/channel_intro_message.tsx
@@ -50,7 +50,7 @@ type Props = {
 
 export default class ChannelIntroMessage extends React.PureComponent<Props> {
     componentDidMount() {
-        if (this.props.stats && !this.props.stats.total_users_count) {
+        if (!this.props.stats?.total_users_count) {
             this.props.actions.getTotalUsersStats();
         }
     }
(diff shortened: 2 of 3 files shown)


components/post_view/channel_intro_message/channel_intro_message.tsx
```diff
@@ -43,9 +43,21 @@ type Props = {
...
+    actions: {
+        getTotalUsersStats: () => any;
+    };
 }
 
 export default class ChannelIntroMessage extends React.PureComponent<Props> {
+    showAddMemberButton = false;
+
+    componentDidMount() {
+        if (this.props.actions.getTotalUsersStats) {
```
marianunez, line 55: Should this be checking if the data is loaded in the state (`props.stats`) vs checking the existence of the action?

components/post_view/channel_intro_message/channel_intro_message.tsx
```diff
@@ -190,12 +202,12 @@ function createDMIntroMessage(channel: Channel, centeredIntro: string, teammate:
...
 
-function createOffTopicIntroMessage(channel: Channel, centeredIntro: string, stats: any, usersLimit: number, theme: any) {
+function createOffTopicIntroMessage(channel: Channel, centeredIntro: string, stats: any, usersLimit: number, theme: any, showAddMemberButton: boolean) {
     const isPrivate = channel.type === Constants.PRIVATE_CHANNEL;
     const children = createSetHeaderButton(channel);
     let totalUsers = 0;
-    if (stats && (typeof stats.TOTAL_USERS === 'number')) {
-        totalUsers = stats.TOTAL_USERS;
+    if (stats && (typeof stats.total_users_count === 'number')) {
+        totalUsers = stats.total_users_count;
```
marianunez, line 210: Is this type check needed given that the type defined for this is already `number` [here](https://github.com/mattermost/mattermost-redux/blob/3d1028034d7677adfda58e91b9a5dcaf1bc0ff99/src/types/users.ts#L118)?

If not, I believe we could simplify this to: 

```suggestion
   totalUsers = stats?.total_users_count | 0;
```

Also, just noted below that this is being also defaulted in the `index` file.

components/post_view/channel_intro_message/channel_intro_message.tsx
```diff
@@ -254,14 +267,15 @@ export function createDefaultIntroMessage(
...
     isReadOnly?: boolean,
     teamIsGroupConstrained?: boolean,
 ) {
     let teamInviteLink = null;
     let totalUsers = 0;
-    if (stats && (typeof stats.TOTAL_USERS === 'number')) {
-        totalUsers = stats.TOTAL_USERS;
+    if (stats && (typeof stats.total_users_count === 'number')) {
+        totalUsers = stats.total_users_count;
     }
```
marianunez, line 279: Same comment as above

components/post_view/channel_intro_message/channel_intro_message.tsx
```diff
@@ -370,14 +385,14 @@ export function createDefaultIntroMessage(
...
+function createStandardIntroMessage(channel: Channel, centeredIntro: string, theme: any, showAddMemberButton: boolean, stats: any, usersLimit: number, locale: string, creatorName: string) {
     const uiName = channel.display_name;
     let memberMessage;
     const channelIsArchived = channel.delete_at !== 0;
 
     let totalUsers = 0;
-    if (stats && (typeof stats.TOTAL_USERS === 'number')) {
-        totalUsers = stats.TOTAL_USERS;
+    if (stats && (typeof stats.total_users_count === 'number')) {
+        totalUsers = stats.total_users_count;
```
marianunez, line 395: Another instance :)

components/post_view/channel_intro_message/add_members_button.tsx
```diff
@@ -28,12 +30,15 @@ export interface AddMembersButtonProps {
...
 }
 
-const AddMembersButton: React.FC<AddMembersButtonProps> = ({totalUsers, usersLimit, channel, setHeader, theme}: AddMembersButtonProps) => {
+const AddMembersButton: React.FC<AddMembersButtonProps> = ({totalUsers, usersLimit, channel, setHeader, theme, showAddMemberButton}: AddMembersButtonProps) => {
     const isPrivate = channel.type === Constants.PRIVATE_CHANNEL;
     const inviteUsers = totalUsers < usersLimit;
-
+    if (!showAddMemberButton) {
+        return (<LoadingSpinner/>);
+    }
```
marianunez, line 41: Could we infer this maybe by `totalUsers`? If it's `0` then show the `<LoadingSpinner>`? There should never be a server with zero users, at least an admin should be present. 

0/5 just trying to avoid the extra prop all over the other places we are calling this.

marianunez (changes requested):
Overall LGTM, some comments below. 

Also, to note that the `getTotalUsersStats` only returns the amount of users from the teams they have permissions to but that's the closest we will get. It will not hurt to see the prompt bigger if your team is small but the server has more than 10 in general.

components/post_view/channel_intro_message/add_members_button.tsx
pablovelezvidal: agree, that suggestion will work!

components/post_view/channel_intro_message/channel_intro_message.tsx
pablovelezvidal: ok!

components/post_view/channel_intro_message/channel_intro_message.tsx
pablovelezvidal: ok!

components/post_view/channel_intro_message/channel_intro_message.tsx
```diff
@@ -43,9 +43,17 @@ type Props = {
...
     theme: any;
     usersLimit: number;
+    actions: {
+        getTotalUsersStats: () => any;
+    };
 }
 
 export default class ChannelIntroMessage extends React.PureComponent<Props> {
+    componentDidMount() {
+        if (this.props.stats && !this.props.stats.total_users_count) {
```
marianunez, line 53:
```suggestion
        if (!this.props.stats?.total_users_count) {
```

marianunez (approved):
LGTM, thanks @pablovelezvidal! Non-blocking comment below

stevemudie (approved):
Tested and passed

1 - Invited user
2 - Opened user on separate browser
3 - Used /test to add enough users to surpass the 10 limit
4 - Refreshed admin and user
5 - Observed change in invite button as expected

nickmisasi (approved):

PR #7767, opened 2021-03-26
Fix error when root post has been deleted

ThiefMaster:
#### Summary
When posts were deleted in bulk there may be posts whose root post doesn't exist anymore. I believe this usually happens when posts are deleted due to an EE data retention policy.

This caused a JS error when clicking the reply button on the remaining post from the thread because it tried to get the `reply_count` from a `null` (or `undefined`, I don't remember).

#### Ticket Link
[MM-34310](https://mattermost.atlassian.net/browse/MM-34310)

### DIFF
--- a/components/rhs_thread/rhs_thread.test.tsx
+++ b/components/rhs_thread/rhs_thread.test.tsx
@@ -9,17 +9,8 @@ import {UserProfile} from 'mattermost-redux/types/users';
 import {Post} from 'mattermost-redux/types/posts';
 
 import {TestHelper} from 'utils/test_helper';
-import * as Utils from 'utils/utils.jsx';
 
 import RhsThread from './rhs_thread';
-
-jest.mock('utils/utils', () => {
-    const original = jest.requireActual('utils/utils');
-    return {
-        ...original,
-        getRootPost: jest.fn().mockImplementation(original.getRootPost),
-    };
-});
 
 describe('components/RhsThread', () => {
     const post: Post = TestHelper.getPostMock({
@@ -144,10 +135,13 @@ describe('components/RhsThread', () => {
     });
 
     test('should not break if root post is missing', () => {
-        (Utils.getRootPost as any).mockImplementation(() => undefined);
+        const props = {
+            ...baseProps,
+            posts: [{...baseProps.posts[0], root_id: 'something'}],
+        };
 
         shallow(
-            <RhsThread {...baseProps}/>,
+            <RhsThread {...props}/>,
         );
     });
 });
--- a/components/rhs_thread/rhs_thread.test.tsx
+++ b/components/rhs_thread/rhs_thread.test.tsx
@@ -12,6 +12,14 @@ import {TestHelper} from 'utils/test_helper';
 
 import RhsThread from './rhs_thread';
 
+jest.mock('utils/utils', () => {
+    const original = jest.requireActual('utils/utils');
+    return {
+        ...original,
+        getRootPost: jest.fn().mockReturnValue(undefined),
+    };
+});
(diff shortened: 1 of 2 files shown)


nevyangelova (approved):

mkraft (changes requested):
The code change looks good and we just need a test to prove that it fixes the bug.

ThiefMaster:
iirc. there aren't any existing tests for this particular logic so if someone more familiar with the webapp code added them it'd probably be faster

Willyfrog (approved):
LGTM! Thanks!

+1 on tests

mkraft:
> iirc. there aren't any existing tests for this particular logic so if someone more familiar with the webapp code added them it'd probably be faster

Thanks @ThiefMaster, I've sent the request for the development of the tests to the appropriate team.

ThiefMaster:
I had another look, and I think I managed to add a test that covers it. Easier than I expected it to be in the end.

Tye only thing I don't like yet is the fact that the mock is global, but when doing it inside the test function it didn't apply (the test did not fail without my fix).

components/rhs_thread/rhs_thread.test.tsx
```diff
@@ -133,4 +142,12 @@ describe('components/RhsThread', () => {
 
         expect(scrollToBottom).not.toHaveBeenCalled();
     });
+
+    test('should not break if root post is missing', () => {
+        (Utils.getRootPost as any).mockImplementation(() => undefined);
```
ThiefMaster, line 147: I don't like the `(... as any)` here but vscode was showing a type violation. The tests succeed anyway. Let me know if you prefer it without the typecast.

mkraft (approved):

jgilliam17:
Thank you @ThiefMaster I am not sure how to test. Deleting a parent post via api didn't work as it removes replies. Any suggestions? Thanks.

ThiefMaster:
In my case I deleted it directly via the database - I think that's the only quick way to do this. If you have database access just get the post id and run `DELETE FROM posts WHERE id = 'xxx';` and then reload the page.

Using an EE retention policy *may* work as well, but I'm not sure how fine-grained it is, so using that to test may be a bit tedious and take too long...

jgilliam17:
Thanks @ThiefMaster I'm re-assigning to @furqanmlk for testing.

furqanmlk (commented):
image

Unable to reproduce error

Willyfrog:
@ThiefMaster can you help @furqanmlk with the process to reproduce the error

ThiefMaster:
Since https://github.com/mattermost/mattermost-webapp/pull/7175 got merged (where the developer cherrypick/ported the changes from my PR) my guess would be that my PR is no longer necessary...

Willyfrog:
> Since #7175 got merged (where the developer cherrypick/ported the changes from my PR) my guess would be that my PR is no longer necessary...

is it ok to close the PR then? double checking just in case you weren't meaning that :)

ThiefMaster:
Yes, I would say so. Maybe @calebroseland can confirm that his change indeed fully covers what I changed here though.

Willyfrog:
ok, if @calebroseland doesn't confirm, we can always reopen. Still! thanks for the PR!

calebroseland:
> Yes, I would say so. Maybe @calebroseland can confirm that his change indeed fully covers what I changed here though.

Yep, the picked/ported [change](https://github.com/mattermost/mattermost-webapp/commit/e93ecd995fe6643896563f418ab7e78ad43f3277) should fully cover this. (`rootPost && ...` merged into the changes #7175 did.)

PR #7771, opened 2021-03-26
MM-27380 Move collapsed categories to server and make sidebar more optimistic

hmhealey:
Instead of storing the collapsed state of categories in local storage which is per-browser and wiped on logout, we'll now be storing collapsed state on the server where it's synced between all of the user's sessions.

To make it still perform nicely now that collapsing/expanding a category now requires a server trip, I've made the new `patchCategory` action (used for collapsing/expanding, renaming, and changing the sorting method for a category) optimistic as well.

Also, this PR changes the web app and mattermost-redux at the same time! Oooooh!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27380

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/17194


packages/mattermost-redux/src/actions/channel_categories.ts
```diff
@@ -34,28 +34,55 @@ import {$ID} from 'mattermost-redux/types/utilities';
...
+
+        try {
+            await Client4.updateChannelCategory(currentUserId, category.team_id, patchedCategory);
+        } catch (error) {
+            dispatch({
+                type: ChannelCategoryTypes.RECEIVED_CATEGORY,
+                data: category,
+            });
+
+            forceLogoutIfNecessary(error, dispatch, getState);
```
nevyangelova, line 80: would this be necessary if we get a category collapse error?

nevyangelova (commented):
LGTM :) Is the server PR ready to merge as well ?

packages/mattermost-redux/src/actions/channel_categories.ts
hmhealey: All API requests actually have this in the error handler. It logs the app out if we receive a 401 error from the server since that means the user's session expired or was revoked. Although now that I say it's everywhere, it would be nice to have some sort of middleware handling this...

nevyangelova (approved):

matthewbirtch (approved):
This works like a dream @hmhealey! Nice work.

I tested across multiple instances of the test server in Chrome. Didn't see any issues at all. Once mobile custom categories are ready, we'll have to test with the mobile app as well and see if this is desirable behavior between the mobile app and desktop

devinbinnie (approved):

esethna:
@jgilliam17 ping on this

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified collapsed categories state is stored and synced across sessions. Tested on various browsers and OSs (macOS, Win10 and Ubuntu)

PR #7785, opened 2021-03-29
Re-enable snapshotting of node_modules

hmhealey:
I turned off snapshotting of node_modules in https://github.com/mattermost/mattermost-webapp/pull/7598 since Webpack wasn't rebuilding mattermost-redux when it was compiled using `npm run dev:watch`. Since we don't need to use that any more, we can re-enable snapshotting to prevent Webpack from having to watch for changes in node_modules and hopefully save a bit of time and memory there.

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/7598


hmhealey:
No QA review required since this only has an effect during development.

Willyfrog (approved):

jfrerich (approved):

PR #7786, opened 2021-03-29
Mm 20425 migrate youtubevideo to typescript

majidsajadi:
#### Summary

Migrate `youtube_video` to typescript.

I need to point to two specific problems. 

1. there is no `type` attribute on `iframe`. i deleted this attribute but let me know if break something.
2. `metadata` prop was optional in the original component so i didnt change it to required but there are some cases which needed to check if `metadata` exists and should not be `undefined`. i handled these cases but i guess this not a bulletproof solution. let me know if i should handle this problem differently. (also couldnt make it required because `getOpenGraphMetadataForUrl` may returns undefined.)
 
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/16149


devinbinnie:
Looks like a couple of the tests are failing due to out-of-date snapshots.
Can you run `npm run test -- -u` and commit the changes?

majidsajadi:
@devinbinnie done. updated snapshots.

components/youtube_video/youtube_video.tsx
```diff
@@ -68,18 +71,17 @@ export default class YoutubeVideo extends React.PureComponent {
                     src={'https://www.youtube.com/embed/' + videoId + '?autoplay=1&autohide=1&border=0&wmode=opaque&fs=1&enablejsapi=1' + time}
                     width='480px'
                     height='360px'
-                    type='text/html'
```
hmhealey, line 71: Looks like that was a mistake in whatever old code sample we based the YouTube preview off of: https://stackoverflow.com/a/4810144

hmhealey (approved):
Looks good to me!

jgilliam17 (approved):
Thank you @majidsajadi 
Tested, looks good to merge.
- Verified YouTube previews - as before.

PR #7791, opened 2021-03-30
Handle modal double submit

larkox:
#### Summary
Handle double submit for modals

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34410


mickmister (approved):
I find it hard to believe that this fixes the issue because this is single-threaded, but idk what was causing the issue. Code LGTM

catalintomai (approved):

DHaussermann (approved):
Tested and passed
- Confirmed that this issue could repro before with jttp-hello
- When deploying this change I am no longer reproduce this issue no matter when the the submit button is clicked
- Tested on webapp and desktop 

LGTM!

PR #6609, opened 2020-10-02
Feature/MM 25430 show online status in channel switcher

Prescise:
#### Summary: 

This PR show the online status of a user in the channel switcher.

#### JIRA Ticket: 

https://mattermost.atlassian.net/browse/MM-25430
Fixes https://github.com/mattermost/mattermost-server/issues/15651

PR #7797, opened 2021-03-30
MM-34371: handle undefined status in switcher

lieut-data:
#### Summary
https://github.com/mattermost/mattermost-webapp/pull/6609 introduced the display of user statuses in the channel switcher, but without loading the status for users not already known to the client.

This PR works around that limitation by correctly rendering an undefined status. It does not attempt to load the status for these users at this time.

Instead of:
image

We now show:
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34371

### DIFF
--- a/sass/components/_mentions.scss
+++ b/sass/components/_mentions.scss
@@ -63,7 +63,6 @@
 
     .status {
         width: auto;
-        margin: 0 8px;
         display: block;
         top: 1px;
 
@@ -71,6 +70,12 @@
             width: 12px;
             height: 12px;
         }
+    }
+
+    // .mentions__name is way too multi-purpose. This rule targets its use in
+    // SwitchChannelSuggestion, applying margin in a way that handles null elements correctly.
+    .status, .mentions__fullname {
+        margin: 0 0 0 8px;
     }
 
     .fa-user {


sass/components/_mentions.scss
```diff
@@ -73,6 +72,12 @@
         }
     }
 
+    // .mentions__name is way too multi-purpose. This rule targets its use in
+    // SwitchChannelSuggestion, applying margin in a way that handles null elements correctly.
+    .status, .mentions__fullname {
```
lieut-data, line 77: `.status` and `.mentions__fullname` is only used in `SwitchChannelSuggestion`. I'd love to see us holistically tackle the CSS here to avoid "weird" edge cases like this.

lieut-data:
@matthewbirtch, if you can use the channel switcher to find a user (DM channel) that you haven't otherwise engaged with, you should be able to verify the behaviour in place.

nevyangelova (approved):

amyblais:
@hmhealey Please prioritize this PR review today (plan is to cut RC-1 for v5.34 tomorrow).

sass/components/_mentions.scss
hmhealey: Yeah, the autocomplete is another component in desperate need of a rewrite

hmhealey (approved):

jgilliam17 (approved):
Thank you @lieut-data 
Tested, looks good to merge.
- Verified channel switcher user list, including the cases when users have an undefined status  - renders as expected, no longer missing spacing.

PR #7799, opened 2021-03-30
make totalUsers optional

lieut-data:
#### Summary
The `totalUsers` prop expected by `AddMembersButton` isn't actually required, since the component has code to handle when the prop is falsy.  The property itself isn't available until after the background stats have loaded.

This removes a warning during development:
image

#### Ticket Link
None.


lieut-data:
Skipping `3: QA Review` since this change mostly targets the development environment.

michelengelen (approved):
LGTM... although the 'if (!totalUsers)' could potentially break, when it is '0'.

This is nothing the current change introduced, but it could be a potential cause for a bug!

pablovelezvidal (approved):

PR #7800, opened 2021-03-30
make actorId optional

lieut-data:
#### Summary
The combined system message component assembles a message from a set of system posts, but the TypeScript definition assumes that all system messages have actors. Some do not, including the opening system messages of `Town Square`, for which the join message is initiated by the server and not a user.

This eliminates the following console error:
image

#### Ticket Link
None.

### DIFF
--- a/components/post_view/combined_system_message/combined_system_message.tsx
+++ b/components/post_view/combined_system_message/combined_system_message.tsx
@@ -174,7 +174,7 @@ export type Props = {
     currentUsername: string;
     intl: IntlShape;
     messageData: Array<{
-        actorId: string;
+        actorId?: string;
         postType: string;
         userIds: string[];
     }>;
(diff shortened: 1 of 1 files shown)


lieut-data:
Skipping `3: QA Review` since this only affects typing.

components/post_view/combined_system_message/combined_system_message.tsx
```diff
@@ -320,7 +320,7 @@ export class CombinedSystemMessage extends React.PureComponent<Props> {
         );
     }
 
-    renderMessage(postType: string, userIds: string[], actorId: string): JSX.Element {
+    renderMessage(postType: string, userIds: string[], actorId?: string): JSX.Element {
         return (
             <React.Fragment key={postType + actorId}>
```
lieut-data, line 325: The use of `undefined` here in the concatenation remains unresolved by this PR, but technically works since we just concatenate the word `undefined` to get a unique case.

hanzei (approved):
LGTM

devinbinnie (approved):

PR #7812, opened 2021-04-01
fixed positioning (and with that styling) of the react-select menu

michelengelen:
#### Summary
Fixes incorrect theming of select inputs

#### Ticket Link
Fixes: [MM-34217](https://mattermost.atlassian.net/browse/MM-34217)

#### Screenshots

**Before**
Screenshot 2021-04-01 at 09 34 43

**After**
Screenshot 2021-04-01 at 09 33 28

### DIFF
--- a/components/user_settings/display/manage_timezones/manage_timezones.tsx
+++ b/components/user_settings/display/manage_timezones/manage_timezones.tsx
@@ -209,7 +209,6 @@ export default class ManageTimezones extends React.PureComponent<Props, State> {
                     className='react-select react-select-top'
                     classNamePrefix='react-select'
                     id='displayTimezone'
-                    menuPortalTarget={document.body}
                     styles={reactStyles}
                     options={timeOptions}
                     clearable={false}
@@ -217,6 +216,7 @@ export default class ManageTimezones extends React.PureComponent<Props, State> {
                     value={useAutomaticTimezone ? {label: this.state.automaticTimezone, value: this.state.automaticTimezone} : this.state.selectedOption}
                     aria-labelledby='changeInterfaceTimezoneLabel'
                     isDisabled={useAutomaticTimezone}
+                    menuPosition={'absolute'}
                 />
                 {serverError}
             </div>


michelengelen:
@matthewbirtch added you as reviewer to check if the assigned colors are correct.

nevyangelova (approved):

jgilliam17:
Thanks @michelengelen 
Time zone dropdown looks good in all themes, but language and number of direct messages dropdowns are still incorrectly themed
Screen Shot 2021-04-01 at 9 11 07 AM
Screen Shot 2021-04-01 at 9 08 12 AM

matthewbirtch:
> @matthewbirtch added you as reviewer to check if the assigned colors are correct.

@michelengelen I think we should change the hover state to 8% center channel text color rather than the button color. This way we're consistent with the menu component we have

calebroseland (commented):
@michelengelen I think making this one absolute may not be the way to go. It would be somewhat prohibitive to do on the language select given it is near the bottom edge of the modal and would cause scrolling or be cut off rather than being overflow-visible/on top.


The selectors in `_react-select.scss` are all based on the inline container root class `.react-select`, this isn't the case when the normal (body-appended/portal-like) method is used.

All but two of the class/selectors are all very specific to react-select and its elements, I'd propose removing the first level `.react-select` (except on those two that need it).

This does reduce the selector specificity, so it _might_ affect a few other places, but from the surface this patch seems to fix it.

[MM-34217.patch.zip](https://github.com/mattermost/mattermost-webapp/files/6246681/MM-34217.patch.zip)

michelengelen:
> @michelengelen I think making this one absolute may not be the way to go. It would be somewhat prohibitive to do on the language select given it is near the bottom edge of the modal and would cause scrolling or be cut off rather than being overflow-visible/on top.
> 
> 
> 
> 
> 
> The selectors in `_react-select.scss` are all based on the container's root class (`.react-select`), when the normal (body-appended/portal-like) method is used, this selector/path doesn't exist for the options list.
> 
> 
> 
> All but two of the class/selectors are all very specific to react-select and its elements, I'd propose removing the first level `.react-select` (except on those two that need it).
> 
> 
> 
> This does reduce the selector specificity, so it _might_ negatively affect instances eleswhere, but from the surface this patch seems to fix it. (just the selectors, not colors)
> 
> 
> 
> [MM-34217.patch.zip](https://github.com/mattermost/mattermost-webapp/files/6246681/MM-34217.patch.zip)
> 
> 

Hey @calebroseland i know that way of fixing this bug. It is a more "global" approach and that's why I have gone down the other path first. I just did not want to provide a solution, that possibly has side effects in other places.

Since those effects generally seem to be less likely to happen I will change my proposal accordingly.

calebroseland:
This LGTM for the most part. My only hangup is I'm noticing this odd UI state—where the welcome tutorial jumps.
CleanShot 2021-04-06 at 12 12 03

Closer up, depends on scroll position (maybe whether or not the portal/dropdown content overflows browser viewport?)
CleanShot 2021-04-06 at 12 53 01

calebroseland (commented):
This LGTM for the most part. My only hangup is I'm noticing this odd UI state—where the welcome tutorial jumps. Idk exactly what's going on.
CleanShot 2021-04-06 at 12 12 03

Closer up, depends on scroll position (maybe whether or not the portal/dropdown content overflows browser viewport?)
CleanShot 2021-04-06 at 12 53 01

matthewbirtch:
> @michelengelen I think we should change the hover state to 8% center channel text color rather than the button color. This way we're consistent with the menu component we have

@michelengelen did you see this note? 

1. Let's change the hover state to 8% opacity center channel text color.
2. Also, I just noticed that the active state when I press the mouse down makes the option illegible. We should change the active state to 8% opacity button background color

michelengelen:
@matthewbirtch I did change the opacity on `hover` and `active` state.
You wrote that `active` should have 8% as well, but since that opacity value is already on `hover` I doubled the value like we do with the buttons, is that OK?

**Screenshot**
Screenshot 2021-04-08 at 10 55 48

michelengelen:
1 => selected
2 => active
3 => hover

@nevyangelova and @calebroseland I did change the way the styles are structured to cover the usage with `non-portaled` selects as well ... just in case! :D

nevyangelova (approved):

matthewbirtch:
> @matthewbirtch I did change the opacity on `hover` and `active` state.
> You wrote that `active` should have 8% as well, but since that opacity value is already on `hover` I doubled the value like we do with the buttons, is that OK?

Sorry, I think I wasn't clear @michelengelen.

hover: use `var(--center-channel-color-08)` for the background
active (mouse down) state: use `var(--button-bg-16)` for the background
selected: use `var(--button-bg)` for the background and use `var(--button-color)` for the text
hover + selected: use `var(--button-bg-88)` for the background and use `var(--button-color)` for the text (right now this look broken)

amyblais:
@michelengelen Do we know which PR originally caused this bug and if we could consider reverting that? Just want to make sure were not introducing a last-minute regression when this PR gets merged by tomorrow's code freeze deadline.

michelengelen:
@amyblais i done know what bug you are referring to. The one @calebroseland found with the overflowing select menu?

amyblais:
@michelengelen This: https://mattermost.atlassian.net/browse/MM-34217

michelengelen:
No, I don't know which PR was the cause for this.

michelengelen:
ok, @matthewbirtch could you please re-check?

**Screenshot**
Screenshot 2021-04-08 at 18 55 26

1 => selected + hovering - `var(--button-bg-88)` + `var(--button-color)`
2 => selected - `var(--button-bg)` + `var(--button-color)`
3 => active - `var(--button-bg-16)` + `color: inherit`
4 => hover - `var(--center-channel-color-08)` + `color: inherit`

Is this correct?

matthewbirtch (approved):
Looks great now @michelengelen

amyblais:
@jgilliam17 @calebroseland If you're able, please help review by tomorrow's v5.34 code freeze.

jgilliam17 (approved):
Thank you @michelengelen 
Tested, looks good to merge.
- Verified positioning and styling of dropdowns in account settings for language, timezone and number of DMs to show - as expected.

calebroseland (approved):
LGTM; the UI glitch/jumping is unrelated, will file separately.

PR #7815, opened 2021-04-01
GH-17288/MM-34367: TypeScript conversion

CarbonHeartDev:
* Converted jsx tests to tsx

#### Summary
I've found the component files already converted to TypeScript, only the tests were in jsx so I switched them to jsx.
Once I've changed the extension the type check game me 7 errors, 6 came from the instances of the component `Reaction` declared for the tests and were caused by the absence of many mandatory fields in the hardcoded testing props declared at line 22 (`baseProps`) while the other came from the method call `wrapper.instance().loadMissingProfiles()` at line 100, the type inferred from the variable `wrapper` declared at line 99 `ShallowWrapper<any, Readonly<{}>, React.Component<{}, {}, any>>` seems to have an `instance` method which doesn't return an object which exposes the `loadMissingProfiles`.
Putting an `any` at line 22 and 99 seems to fix the type check errors but maybe there are better solutions or I can do something else to fit better with the rest of the codebase. 

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/17288

Once I've get a feedback I can close even the #17287 which is really similar.


components/post_view/reaction/reaction.test.tsx
```diff
@@ -19,7 +19,7 @@ describe('components/post_view/Reaction', () => {
     };
     const currentUserId = 'user_id_1';
 
-    const baseProps = {
+    const baseProps: any = {
```
hmhealey, line 22: Which props are missing from baseProps? We generally want to add dummy values for any unused props since we'll get a warning then whenever something new is added

components/post_view/reaction/reaction.test.tsx
```diff
@@ -96,7 +96,7 @@ describe('components/post_view/Reaction', () => {
         const newActions = {...actions, getMissingProfilesByIds: jest.fn()};
         const props = {...baseProps, actions: newActions};
 
-        const wrapper = shallow(<Reaction {...props}/>);
+        const wrapper: any = shallow(<Reaction {...props}/>);
```
hmhealey, line 99: Instead of defining the wrapper as any, you should be able to get the typing working here by defining the type on the `shallow` call instead

```suggestion
        const wrapper = shallow<Reaction>(<Reaction {...props}/>);
```
```

hmhealey (changes requested):
Thanks for the PR, @EmiSan1998! I guess the script we have generating all these tickets didn't take into account that someone else happened to migrate the files a while back, but we appreciate getting the test migrated over to match.

devinbinnie:
Hey @EmiSan1998, let us know if you have any questions about the above comments :)

M-ZubairAhmed:
completed i https://github.com/mattermost/mattermost-webapp/pull/9903

PR #7825, opened 2021-04-06
Bump plugin channel header button limit

crspeller:
#### Summary
Increases the channel header button limit for plugins. Temporary solution until we can come up with better UX here.


plugins/channel_header_plug/channel_header_plug.tsx
```diff
@@ -210,7 +210,7 @@ export default class ChannelHeaderPlug extends React.PureComponent<ChannelHeader
 
         if (components.length === 0) {
             return null;
-        } else if (components.length <= 5) {
+        } else if (components.length <= 15) {
```
jwilander, line 213: Isn't 15 a bit high? I imagine the header would be pretty squished at this point. What about 8 or 9?

crspeller: I went high because if someone installs that many, they intend to use them. Why hide it from them? I figure just do it on an absolutely ridiculous number of buttons.

jwilander: Fair enough

jwilander (approved):

agarciamontoro (approved):
Looks good!

jgilliam17:
Thanks @crspeller Can you please share a Jira ticket number for this issue. 
Is there a way to test this on the PR server? I'm not sure we have 15 plugins (with header icons) to test the limit. Any suggestions

crspeller:
@jgilliam17 No ticket for this one. It's direct response to this thread here: https://community-daily.mattermost.com/core/pl/e7rb7a1qqir5tnjbbshquw33xc

For testing, if you feed it's necessary, you could create a bunch of variants of https://github.com/mattermost/mattermost-plugin-demo with different plugin IDs. Then you would have the extra icons you need.

jgilliam17 (approved):
Thanks @crspeller 
I'm not familiar with building plugins for testing. QA will test this on daily, after merge.

PR #7830, opened 2021-04-07
[MM-34501] - Bring back margin: auto for the no results indicator

nevyangelova:
#### Summary
A change introduced in https://github.com/mattermost/mattermost-webapp/commit/d3744849c8b4cbfdb29f2a8d3510513fdabf7511 removed the margin: auto from the no results indicator which breaks the alignment for the channel switcher. I have brought it back. Wanted to ask what is the reason behind removing it in the first place, because the way I was intending the no results indicator to work was to be centered independently and then let its parent define the positioning. IMO this is less error prone. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34501


components/no_results_indicator/no_results_indicator.scss
```diff
@@ -15,6 +15,7 @@
     text-align: center;
     word-break: break-word;
     max-width: 500px;
+    margin: auto;
```
michelengelen, line 18: is `margin: auto` correct or should it rather be `margin: 0 auto`? Just asking out of curiosity!

michelengelen (approved):
LGTM!

calebroseland (approved):

calebroseland:
Hey @nevyangelova, I can give some background to https://github.com/mattermost/mattermost-webapp/commit/d3744849c8b4cbfdb29f2a8d3510513fdabf7511. `margin: auto`-centering can work great in flexbox and other scenarios, but it doesn't work/behave the same in css grid, so the prop `expanded` adds a `.no-results__holder` which fills the space it occupies and centers `.no-results__wrapper` inside it.

I guess I didn't test enough of the other usages (RHS has its own/additional `margin: auto`), but `margin: auto`/`0 auto` on `.no-results__wrapper` doesn't conflict with `expanded` at all, so I shouldn't have removed it. Thanks for bringing it back!

components/no_results_indicator/no_results_indicator.scss
nevyangelova: I think auto works fine as we will probably be dependent on the parent height in some scenarios

nevyangelova:
> Hey @nevyangelova, I can give some background to [d374484](https://github.com/mattermost/mattermost-webapp/commit/d3744849c8b4cbfdb29f2a8d3510513fdabf7511). `margin: auto`-centering can work great in flexbox and other scenarios, but it doesn't work/behave the same in css grid, so the prop `expanded` adds a `.no-results__holder` which fills the space it occupies and centers `.no-results__wrapper` inside it.
> 
> I guess I didn't test enough of the other usages (RHS has its own/additional `margin: auto`), but `margin: auto`/`0 auto` on `.no-results__wrapper` doesn't conflict with `expanded` at all, so I shouldn't have removed it. Thanks for bringing it back!

thanks for explaining Caleb <3

jgilliam17 (approved):
Thanks @nevyangelova 
Tested, looks good to merge.
- Verified alignment of the `no result` indicator on channel switcher.

PR #7831, opened 2021-04-07
[MM-34563] - Fix hover in category sorting menu for all theme

nevyangelova:
#### Summary
This PR fixes the hovering issues on the category sorting menu

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34563


michelengelen (approved):
LGTM!

hmhealey:
It seems like there might be some doubling up of opacities here since the hover effect for both channel and category submenus is now brighter than the menus themselves. The category submenu has closer opacities between the two menus while the channel menus have more of a difference between them.
Screen Shot 2021-04-08 at 8 11 55 AM
Screen Shot 2021-04-08 at 8 17 46 AM

Here's how it looks on 5.33:
Screen Shot 2021-04-08 at 8 16 45 AM
Screen Shot 2021-04-08 at 8 14 41 AM

nevyangelova:
@hmhealey fixed.. this is happening again because we are styling all divs directly. Changing the SubmenuItem container to span fixed this partially. I can go into even deeper digging but might not be worth it given that compass is coming :P

hmhealey (approved):

hmhealey:
Thanks! That looks better. I agree that we could probably fix it in a "better" way, but waiting until we rewrite the whole thing is probably a better use of time.

jgilliam17 (approved):
Thank you @nevyangelova 
Tested, looks good to merge.
- Verified hover on category menu items.

PR #7850, opened 2021-04-09
Hide bots and apps UI when managed by Apps Framework

larkox:
#### Summary
Hide bots and OAuth apps UI when managed by the apps framework.

When the apps framework plugin is disabled, they show as normal apps/bots.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33158

#### Related Pull Requests
Proxy Plugin: https://github.com/mattermost/mattermost-plugin-apps/pull/138

```release-note
NONE
```


packages/mattermost-redux/src/reducers/entities/integrations.ts
```diff
@@ -195,6 +195,66 @@ function oauthApps(state: IDMappedObjects<OAuthApp> = {}, action: GenericAction)
     }
 }
 
+function appsOAuthAppsIDs(state: string[] = [], action: GenericAction) {
```
catalintomai, line 198: nit: curious why using `App**s**IDs` (plural) vs `BotIDs` (singular), since they are both apps-related artifacts.

components/integrations/bots/bot.test.tsx
```diff
@@ -206,6 +210,7 @@ describe('components/integrations/bots/Bot', () => {
                 accessTokens={accessTokens}
                 team={team}
                 actions={actions}
+                fromApp={false}
```
mickmister, line 213: Can we add a test for a bot that has `fromApp={true}`?

components/integrations/bots/bots.test.tsx
```diff
@@ -120,6 +126,8 @@ describe('components/integrations/bots/Bots', () => {
                 owners={owners}
                 users={users}
                 actions={actions}
+                appsEnabled={false}
+                appsBotIDs={[]}
```
mickmister, line 130: Same here, can we have some tests that have app bots involved?

components/integrations/bots/bots.tsx
```diff
@@ -169,6 +187,7 @@ export default class Bots extends React.PureComponent<Props, State> {
                 accessTokens={(this.props.accessTokens && this.props.accessTokens[bot.user_id]) || {}}
                 actions={this.props.actions}
                 team={this.props.team}
+                fromApp={this.props.appsBotIDs.includes(bot.user_id)}
```
mickmister, line 190: Is there an async operation that could momentarily have `fromApp={false}` before the `getAppBotIDs()` response comes back?

components/integrations/bots/index.ts
```diff
@@ -59,6 +66,7 @@ type Actions = {
 function mapDispatchToProps(dispatch: Dispatch<GenericAction>) {
     return {
         actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc>, Actions>({
+            getAppsBotIDs: fetchAppsBotIDs,
```
mickmister, line 69: 0/5 Maybe name the prop `fetchAppsBotIDs` so it's more clear what's happening within the component's code

components/integrations/installed_oauth_app/installed_oauth_app.test.tsx
```diff
@@ -36,6 +36,7 @@ describe('components/integrations/InstalledOAuthApp', () => {
         onRegenerateSecret: jest.fn(),
         onDelete: jest.fn(),
         filter: '',
+        fromApp: false,
```
mickmister, line 39: Does it make sense to have an apps-related test here?

components/integrations/installed_oauth_app/installed_oauth_app.tsx
```diff
@@ -242,34 +247,42 @@ export default class InstalledOAuthApp extends React.PureComponent<InstalledOAut
...
+                        id='installed_integrations.fromApp'
+                        defaultMessage='Managed by Apps Framework'
+                    />
+                </span>
+            </div>
+        );
+        if (!this.props.fromApp) {
+            appInfo = (
+                <>
                     <div className='item-details__row'>
```
mickmister, line 286: Do we need this fragment here?

components/integrations/installed_oauth_apps/index.ts
```diff
@@ -27,6 +29,7 @@ function mapStateToProps(state: GlobalState) {
     return {
         canManageOauth: haveISystemPermission(state, {permission: Permissions.MANAGE_OAUTH}),
         oauthApps: getOAuthApps(state),
+        appsOAuthAppsIDs: appsEnabled(state) ? getAppsOAuthAppsIDs(state) : [],
```
mickmister, line 32: Can we put this in a memoized selector to avoid recreation of the `[]` here?

components/integrations/installed_oauth_apps/installed_oauth_apps.test.tsx
```diff
@@ -49,6 +49,7 @@ describe('components/integrations/InstalledOAuthApps', () => {
             deleteOAuthApp: jest.fn(),
         },
         enableOAuthServiceProvider: true,
+        appsOAuthAppsIDs: [],
```
mickmister, line 52: Opportunity for test?

packages/mattermost-redux/src/actions/integrations.ts
```diff
@@ -293,6 +293,20 @@ export function getOAuthApps(page = 0, perPage: number = General.PAGE_SIZE_DEFAU
     });
 }
 
+export function getAppsOAuthAppsIDs(): ActionFunc {
+    return bindClientFunc({
+        clientFunc: Client4.getAppsOAuthAppsIDs,
+        onSuccess: [IntegrationTypes.RECEIVED_APPS_OAUTH_APP_IDS],
```
mickmister, line 299: Do we need the arrays around the `onSuccess` values? It looks inconsistent throughout the file, but it caught me off guard at first glance here.

packages/mattermost-redux/src/reducers/entities/integrations.ts
```diff
@@ -195,6 +195,66 @@ function oauthApps(state: IDMappedObjects<OAuthApp> = {}, action: GenericAction)
...
+        if (state.length === 0 && action.data.length === 0) {
+            return state;
+        }
+
+        if (state.length !== action.data.length) {
+            return action.data;
+        }
+
+        const orderedState = state.concat().sort();
+        const orderedData = state.concat().sort();
```
mickmister, line 210:
```suggestion
        const orderedData = action.data.concat().sort();
```

packages/mattermost-redux/src/reducers/entities/integrations.ts
```diff
@@ -195,6 +195,66 @@ function oauthApps(state: IDMappedObjects<OAuthApp> = {}, action: GenericAction)
...
```
mickmister, line 242:
```suggestion
        const orderedData = action.data.concat().sort();
```

mickmister (changes requested):
Functionality looks good :+1:

I have just a few requests. For my comments related to adding more tests, I'd say to prioritize testing on components that have logic around the props, rather than testing a component that is just passing a prop to a child.

packages/mattermost-redux/src/reducers/entities/integrations.ts
larkox: There are two main reasons for this:
- Me not being native speaker, and doubting which of the two ways sounded better.
- Me not going over the changes and ensure consistency.

That being said, I have no preference over singular or plural, so which one do you think is better? I will make sure it is consistent for the choice we decide.

components/integrations/installed_oauth_app/installed_oauth_app.tsx
larkox: No, we don't.

packages/mattermost-redux/src/reducers/entities/integrations.ts
catalintomai: 0/5 on singular vs. plural (prob. singular makes more sense), I just advocate aligning on one vs. another for both.

components/integrations/bots/bots.tsx
larkox: If getAppBotIDs takes too long, the user will see the whole information, and then disappear when that information finally arrives. We could keep the loading state until we have all the information. What do you think?

components/integrations/installed_oauth_app/installed_oauth_app.tsx
larkox: Sorry, yes we do. We have several elements at the same level, so we need to somehow put them together.

packages/mattermost-redux/src/actions/integrations.ts
larkox: No. It is not needed. Nevertheless, for consistency, I would prefer to keep it as a list.

hmhealey: The array is left over from when we always used to have a _SUCCESS action dispatched for every request in addition to something like a POST_RECEIVED action. I think it might still be used in some cases, but we could definitely consider adding a non-array form in the future.

packages/mattermost-redux/src/client/client4.ts
```diff
@@ -2548,6 +2548,20 @@ export default class Client4 {
         );
     };
 
+    getAppsOAuthAppIDs = () => {
```
hmhealey, line 2551: Is there any chance either of these will ever need paging in the future? That's obviously a bigger question than just something that affects the web app, but I find it never hurts to ask

packages/mattermost-redux/src/selectors/entities/integrations.ts
```diff
@@ -25,6 +27,22 @@ export function getOAuthApps(state: GlobalState) {
     return state.entities.integrations.oauthApps;
 }
 
+export const getAppsOAuthAppIDs: (state: GlobalState) => string[] = createSelector(
+    appsEnabled,
+    (state: GlobalState) => state.entities.integrations.appsOAuthAppIDs,
+    (apps, ids) => {
+        return apps ? ids : [];
```
hmhealey, line 34: 👍

hmhealey (approved):
One question, but I assume it's not going to result in any changes unless we want to add some heavy future proofing.

packages/mattermost-redux/src/client/client4.ts
larkox: Being small amount of information (a list of ids) I hope there is no case where this may become a problem (e.g. having 100 Apps installed?). That being said, I am open to suggestions or comments.

components/integrations/bots/bots.tsx
mickmister: I think we should keep showing the loading animation. But if the bot id fetch fails, then show all of the information.

mickmister (approved):

DHaussermann:
@larkox there may be an issue here where the UI can revert to the old version. I'm still investigating. I think I turned off bot creation in the system console and re-enabled then the UI no longer appers changed. I'll try and pin this down. 

For the time being... a question - is `@appsbot` new to this bersion of the proxy?
image
This should have a Mattermost logo and not Jira. If this bot is unrelated to the changes, I'll make a new issue.

DHaussermann:
We can disregard the comment above. I cannot repro this issue. I took some time to investigate this and it seems as though the bots where left over data from before the corresponding proxy plugin build was deployed. Seem when you remove apps using debug commands, the corresponding bot is left behind which causes some confusion.

DHaussermann (approved):
Tested and passed
- Backstage UI change to hide config are implemented successfully
- Bots and oAuth have the option removed
- No changes to conventional bots or oAth apps unrelated to MM Apps
- No issues found on subsequent installations when using proper uninstall and reinstall commands
- A couple unrelated issues were found. I will create follow up tasks.

LGTM!

DHaussermann:
@larkox  I need to test additional cases. The Jira ticket you referenced may require a broader solution than removing the option. We may actually need the ownership to change. 0/5 The Apps plugin should own the bot. 

This UI clean-up does not guard against the owner being deactivated. Further more - as pointed out by @mickmister, it causes an additional issue if the sysadmin who owns the bot becomes deactivated, as the bot can then no longer be re-enabled.

larkox:
@DHaussermann For the bots, it might be possible to add something like owned by the plugin (actually, we could create these bots with the plugin RPC methods). The potential problem I see is for OAuth Apps, which with the current APIs we would not be able to do something like that.

hanzei:
@catalintomai Gentle reminder to review this PR

DHaussermann:
@hanzei is this the same solution as originally implemented by @larkox?

If so, I'm still unclear on 1 point - If the original installer of the app is disabled would this not hide the UI allowing a different admin and take ownership?

hanzei:
> @hanzei is this the same solution as originally implemented by @larkox?

Exactly. 

>  If the original installer of the app is disabled would this not hide the UI allowing a different admin and take ownership

@DHaussermann Yes it would. My suggest is that we document that behavior at https://developers.mattermost.com/integrate/apps/ as it's an edge case. What do you think?

DHaussermann:
@hanzei deactivating a sysadmin may be a bit uncommon. But I feel like if we still need some type of workaround that will make the App usable again. 

Would there be a workaround we can document?

cc @catalintomai or @mickmister may have some thoughts as well.

mickmister:
@hanzei @DHaussermann We're currently creating the bot using the sysadmin creds, and creating an access token for the bot using the sysadmin creds.

The plugin API supports a `CreateBot` method, that we can use as an alternative to using a specific user to own the bot. However, there is no `CreateAccessToken` plugin API method. Maybe the change to make is add this to the plugin API, so there is no coupling to any specific user?

DHaussermann:
@catalintomai and @aaronrothschild I have done some testing around the unhappy paths with no UI barriers

The good news: 
- I can't see any issues that prevent the app from being usable
- Manipulating the Bot for the App is totally recoverable with uninstall / reinstall

The bad news is without this PR - there are no restrictions for Oauth apps either. Deleting the oAuth app is not a recoverable scenario (without debug commands). The app can't properly be uninstalled and reinstalls don't work to re-create the app

What would be best IF possible - Can we open a PR with a subset of this code to only hide oAuth apps. Since this has nothing to do with the blocking problem.

cc @mickmister since you also have some insight on this.

DHaussermann:
@mickmister note that in the comment above I meant that none of the actions prevent Mattermost from being usable. This also means I have no repro steps for https://mattermost.atlassian.net/browse/MM-35111 as I can't see the problem where `pending` bindings calls prevent channel content from loading.

catalintomai:
Thanks for the validation investigation @DHaussermann. @mickmister , how feasible do you think it is to scope down this change (we can use the same PR) to just the hide oAuth apps in the time frame of today (? - or is it until tomorrow Dylan - when they cut the RC).

mickmister:
>  how feasible do you think it is to scope down this change (we can use the same PR) to just the hide oAuth apps in the time frame of today

@catalintomai That should be no problem

DHaussermann:
@catalintomai I just came from triage and it seems based on when we need RC1 cherry-picking tomorrow could be a possibility.

mickmister:
@catalintomai change has been made to show the admin controls for bot accounts. Be sure to set the "no whitespace" option when reviewing the commit

catalintomai (approved):

DHaussermann (approved):
~~Tested and passed~~
~~- Ran through all Valid, Invalid and Error cases using the test app~~
~~- Tested these cases from command parser, channel header and post menu bindings~~
~~- No issues found~~
 
~~LGTM!~~

DHaussermann:
@amyblais  This PR can be cherry-picked now. IS this something you can do or do you need @mickmister to assist?

DHaussermann:
My Apologies for the ping. That comment was intended for a different PR. 

This is being tested now.

DHaussermann (approved):
Tested and passed
- Confirmed changes to Bots improves labeling and association 
- Confirmed oAuth App for MM Apps now hide UI element to manipulate them
- Tested with multiple apps - no issues found
- Regression tested creating generic Apps and Bots to ensure they are not affected

LGTM!

(This approval correctly intended for the bots PR this time :) )

DHaussermann:
cc
@amyblais this is tested and passed. It now ready for cherry-pick

amyblais:
@hanzei Can you also help cherry-pick this to ``cloud``?

PR #7883, opened 2021-04-12
[MM-33462] Extend manifest type

hanzei:
#### Summary
`AppManifest` was missing a couple of fields are are both useful for the webapp and for app developers.

I intentionally leave the fields out that are purely for the backend like callbacks and the AWS information.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33462


mickmister (approved):
Found one typo, LGTM otherwise :+1:

hanzei:
@jfrerich Gentle reminder to take a look when you have the time.

amyblais:
@jfrerich @catalintomai Kind reminder for PR review.

packages/mattermost-redux/src/types/apps.ts
```diff
@@ -4,11 +4,38 @@
...
 // Apps Framework feature is experimental, and the contents of this file are
 // susceptible to breaking changes without pushing the major version of this package.
+
+export enum AppType {
+    HTTP = 'http',
+    AWSLambda = 'aws_lambda',
+    Builtin = 'builtin',
+}
+
+export enum Permission {
```
jfrerich, line 14: Suggest that we use the plural to match the `Locations` enum
```suggestion
export enum Permissions {
```

jfrerich (approved):
One small suggestion.

packages/mattermost-redux/src/types/apps.ts
hanzei: https://www.typescriptlang.org/docs/handbook/enums.html is using singular names for enums. https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/4x252001(v=vs.71)?redirectedfrom=MSDN also states

> Use a singular name for most Enum types, but use a plural name for Enum types that are bit fields.


Hence, I'm leaving it as singular.

PR #7908, opened 2021-04-13
use the new server value is_free_trial

pablovelezvidal:
#### Summary
Use the new value from server is_free_trial

#### Ticket Link
n/a

### DIFF
--- a/components/admin_console/billing/billing_subscriptions/index.tsx
+++ b/components/admin_console/billing/billing_subscriptions/index.tsx
@@ -64,7 +64,7 @@ const BillingSubscriptions: React.FC<Props> = () => {
 
     let isFreeTrial = false;
     let daysLeftOnTrial = 0;
-    if (isCloud && subscription && subscription.is_free_trial === 'true') {
+    if (subscription?.is_free_trial === 'true') {
         isFreeTrial = true;
         daysLeftOnTrial = getRemainingDaysFromFutureTimestamp(subscription.trial_end_at);
     }
--- a/components/admin_console/billing/billing_subscriptions/index.tsx
+++ b/components/admin_console/billing/billing_subscriptions/index.tsx
@@ -64,7 +64,7 @@ const BillingSubscriptions: React.FC<Props> = () => {
 
     let isFreeTrial = false;
     let daysLeftOnTrial = 0;
-    if (subscription && subscription.trial_end_at > 0 && subscription.status === 'trialing') {
+    if (isCloud && subscription && subscription.is_free_trial === 'true') {
         isFreeTrial = true;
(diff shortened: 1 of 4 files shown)


components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -64,7 +64,7 @@ const BillingSubscriptions: React.FC<Props> = () => {
 
     let isFreeTrial = false;
     let daysLeftOnTrial = 0;
-    if (subscription && subscription.trial_end_at > 0 && subscription.status === 'trialing') {
+    if (isCloud && subscription && subscription.is_free_trial === 'true') {
```
marianunez, line 67: I think `isCloud` is not needed at this point given that the subscription screen does not show in non-cloud and is already checked by the System Console

components/announcement_bar/cloud_trial_announcement_bar/index.ts
```diff
@@ -31,7 +31,7 @@ function mapStateToProps(state: GlobalState) {
     let isFreeTrial = false;
     let daysLeftOnTrial = 0;
 
-    if (isCloud && subscription && subscription.trial_end_at > 0 && subscription?.status === 'trialing') {
+    if (isCloud && subscription && subscription?.is_free_trial === 'true') {
```
marianunez, line 34: Why is `is_free_trial` a string?

marianunez (approved):
Thanks for the quick PR @pablovelezvidal!

components/announcement_bar/cloud_trial_announcement_bar/index.ts
pablovelezvidal: the pr from the server defines it as a string

components/admin_console/billing/billing_subscriptions/index.tsx
pablovelezvidal: ok

components/announcement_bar/cloud_trial_announcement_bar/index.ts
marianunez: Yes, just curious why it can't be a boolean? @nickmisasi  This seems error prone

pablovelezvidal: I just also noticed that some config values from the server like EnableDeveloper are also handled in the same way. It is not ideal and potentially error prone, but since we have only two places where we are validating this now, I am not sure if it is worth modifying the value again.

marianunez:
@nickmisasi can we get a review here to merge to the feature branch?

nickmisasi (approved):

PR #7913, opened 2021-04-14
Cypress/E2E: Fix CI run of Cypress and add flag to sort specs to test first

saturninoabril:
#### Summary
- Changes in `e2e/run_test_cycle.js` fix the issue where the container is being rerun unexpectedly. Probably due to mishandling of recursion thru async/await operation.
- In `e2e/utils/file.js` - added ability to sort specs to first. This is in the case for Elasticsearch where it's better to test first, so that indexing will not take a lot of time instead of doing last where there are lots of test data already. Typical used would be:
```
node generate_test_cycle.js --stage="@prod" --sortFirst="@elasticsearch,@ldap_group,@ldap" --sortLast="@saml,@keycloak,@mfa"
```

Test report - https://mm-cypress-report.s3.amazonaws.com/4454-master-e20-prod-mattermost-enterprise-edition:master/mochawesome.html

### DIFF
--- a/e2e/run_test_cycle.js
+++ b/e2e/run_test_cycle.js
@@ -48,7 +48,7 @@ const {
     REPO,
 } = process.env;
 
-async function runTest(specExecution, currentTestCount) {
+async function runCypressTest(specExecution, testIndex) {
     const browser = 'chrome';
     const headless = true;
 
@@ -85,7 +85,7 @@ async function runTest(specExecution, currentTestCount) {
     });
 
     // Write and update test environment details once
-    if (currentTestCount === 1) {
+    if (testIndex === 0) {
         const environment = {
             cypress_version: result.cypressVersion,
             browser_name: result.browserName,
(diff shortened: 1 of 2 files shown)


e2e/utils/file.js
```diff
@@ -90,42 +90,60 @@ function getTestFiles() {
...
+function getWeightedFiles(metadata, sortFirst = true) {
+    let weightedFiles = [];
+    if (metadata) {
+        metadata.split(',').forEach((word, i, arr) => {
             const sl = grepCommand(word);
             const files = grepFiles(sl).map((file) => {
-                return {file, sortWeight: i + 1};
+                return {
+                    file,
+                    sortWeight: sortFirst ? (i - arr.length) : (i + 1),
```
josephbaylon, line 101: Quick question, sort first weights are negative values right

saturninoabril: Correct

e2e/run_test_cycle.js
```diff
@@ -142,61 +166,69 @@ async function testLoop() {
...
+        tryNext: true,
+        count: newCount,
+        retry: 0,
+        message: 'Continue testing',
+    };
+}
 
-        return acc;
-    }, {});
+async function runSpec(count = 0, retry = 0) {
```
josephbaylon, line 217: Isn't this still recursion thru async/await operation?

josephbaylon (approved):
LGTM. Just a couple of questions

e2e/run_test_cycle.js
saturninoabril: Right, still thru async/await, the main difference which I though where the issue is coming from is the [code here](https://github.com/mattermost/mattermost-webapp/pull/7913/files#diff-8686cf1e4a2ca1851e0b683287c8e22cde5d7c441ea88385df53c80893f16722L175):
```javascript
return testLoop(); // eslint-disable-line consistent-return
```
as compared to the [new approach here](https://github.com/mattermost/mattermost-webapp/pull/7913/files#diff-8686cf1e4a2ca1851e0b683287c8e22cde5d7c441ea88385df53c80893f16722R218):
```javascript
const fragment = await runSpecFragment(count, retry);
```

I maybe totally wrong but I tried couple of prod and unstable runs and CircleCI didn't complain/restart the container.

furqanmlk (approved):
Thanks @saturninoabril

PR #7918, opened 2021-04-14
Prod test fixes

prapti:
Fixing the incoming webhook and gm reply tests:
- Looks like the incoming webhook tests need forced clicks on the channel names.
- The gm reply test looks like is looking for the reply sooner than it's available.

Screen Shot 2021-04-14 at 4 43 21 PM
Screen Shot 2021-04-14 at 4 40 23 PM
Screen Shot 2021-04-14 at 4 39 03 PM
Screen Shot 2021-04-14 at 4 38 06 PM


josephbaylon (approved):
LGTM. All tests passed locally.

furqanmlk (approved):
locally passed
image

image

image

image

PR #7920, opened 2021-04-15
Adding missing Zephyr keys to existing E2E tests

prapti:
Adding missing Zephyr keys to the following E2E tests:
- message_display_mode_spec.js
- long_post_attachments_spec.js
- message_edit_post_with_attachment_spec.js
Screen Shot 2021-04-14 at 5 44 45 PM
Screen Shot 2021-04-14 at 5 39 48 PM
Screen Shot 2021-04-14 at 5 43 24 PM

### DIFF
--- a/e2e/cypress/integration/account_settings/display/message_display_mode_spec.js
+++ b/e2e/cypress/integration/account_settings/display/message_display_mode_spec.js
@@ -12,6 +12,25 @@
 
 import {getRandomId} from '../../../utils';
 
+describe('Account Settings', () => {
+    before(() => {
+        // # Login as new user and visit town-square
+        cy.apiInitSetup({loginAfter: true}).then(({team}) => {
+            cy.visit(`/${team.name}/channels/town-square`);
+        });
+    });
+
+    it('MM-T103_1 Compact view: Line breaks remain intact after editing', () => {
+        // * Verify line breaks do not change and blank line is still there in compact view.
+        verifyLineBreaksRemainIntact('COMPACT');
+    });
+
+    it('MM-T103_2 Standard view: Line breaks remain intact after editing', () => {
+        // * Verify line breaks do not change and blank line is still there in standard view.
+        verifyLineBreaksRemainIntact('STANDARD');
+    });
+});
+
 function verifyLineBreaksRemainIntact(display) {
     cy.uiChangeMessageDisplaySetting(display);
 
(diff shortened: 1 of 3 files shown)


e2e/cypress/integration/account_settings/display/message_display_mode_spec.js
```diff
@@ -12,55 +12,53 @@
 
 import {getRandomId} from '../../../utils';
 
-describe('Account Settings > Display > Message Display', () => {
+describe('Messaging', () => {
```
josephbaylon, line 15: Maybe either move this file under `Messaging` folder or change describe to `Account Settings` and update zephyr folder to point to `Account Settings`

josephbaylon (approved):
LGTM. Just one minor comment below.

e2e/cypress/integration/account_settings/display/message_display_mode_spec.js
```diff
@@ -12,55 +12,53 @@
...
         cy.apiInitSetup({loginAfter: true}).then(({team}) => {
             cy.visit(`/${team.name}/channels/town-square`);
         });
     });
 
-    ['COMPACT', 'STANDARD'].forEach((display) => {
-        it(`M14283 ${display} view: Line breaks remain intact after editing`, () => {
-            cy.uiChangeMessageDisplaySetting(display);
+    it('MM-T103 Compact view: Line breaks remain intact after editing', () => {
+        cy.uiChangeMessageDisplaySetting('COMPACT');
```
saturninoabril, line 24: I wonder if we should still test Standard view as well. Could be on another step or a separate test case. Thoughts?

saturninoabril (commented):
Thanks Prapti! LGTM except for one comment before approving the PR.

e2e/cypress/integration/account_settings/display/message_display_mode_spec.js
prapti: Good catch. Thanks Joseph!

e2e/cypress/integration/account_settings/display/message_display_mode_spec.js
prapti: That's a good point. I've added a second test step for the standard view instead of creating a whole new test case. I think that simplifies it a lot.
Also, I've moved it from "Messaging" to "Account Settings" in Zephyr to address Joseph's concern.
Re-requesting review -- thanks @saturninoabril!

e2e/cypress/integration/account_settings/display/message_display_mode_spec.js
```diff
@@ -12,55 +12,63 @@
...
+            should('contain', '(edited)');
+    });
+}
 
-                // * Post should have (edited)
-                cy.get(`#postEdited_${postId}`).
-                    should('be.visible').
-                    should('contain', '(edited)');
-            });
+describe('Account Settings', () => {
```
saturninoabril, line 57: I don't think we've documented it but we typically placed the `describe` block on top so that it's immediately seen when you open the file.
Typical arrangement would be:
```
import...

describe () {
    // variables that are shared among `it` blocks
    const
    let

    before/after()

    it()
}

// helper functions
function() {}
```

saturninoabril (approved):
Thanks Prapti! Tested and passed. Left non-blocking comment on style guide.

Feel free to merge.

e2e/cypress/integration/account_settings/display/message_display_mode_spec.js
prapti: Thanks for catching it @saturninoabril! That was an oversight in a rush to push the changes through :) Fixed!

PR #7921, opened 2021-04-15
MM-32591: Reliable Websockets: Client side changes

agnivade:
We send connection_id and sequence_number as part of the query params.

On receiving an event, we check if it's a hello packet, and if so,
we inspect the connection_id sent with it. If it's different to what
we have already, then we make the API calls and set sequence to 0.

And if the sequence number is different to what we expect, then
we just disconnect the connection, and expect the server to send
from where we left off.

```release-note
NONE
```

https://mattermost.atlassian.net/browse/MM-32591


client/websocket_client.tsx
```diff
@@ -1,10 +1,14 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {SocketEvents} from 'utils/constants';
+
 const MAX_WEBSOCKET_FAILS = 7;
 const MIN_WEBSOCKET_RETRY_TIME = 3000; // 3 sec
 const MAX_WEBSOCKET_RETRY_TIME = 300000; // 5 mins
 
+const reliableWebSockets = true;
```
agnivade, line 10: This needs to be changed to actually take from the server config `ServiceSettings.EnableReliableWebSockets`

client/websocket_client.tsx
```diff
@@ -132,9 +150,39 @@ export default class WebSocketClient {
                     Reflect.deleteProperty(this.responseCallbacks, msg.seq_reply);
                 }
             } else if (this.eventCallback) {
-                if (msg.seq !== this.serverSequence && this.missedEventCallback) {
-                    console.log('missed websocket event, act_seq=' + msg.seq + ' exp_seq=' + this.serverSequence); //eslint-disable-line no-console
-                    this.missedEventCallback();
+                if (reliableWebSockets) {
+                    // We check the hello packet, which is always the first packet in a stream.
+                    if (msg.event === SocketEvents.HELLO && this.missedEventCallback) {
+                        console.log('got connection id ', msg.data.connection_id); //eslint-disable-line no-console
```
agnivade, line 156: There are 2 new `console.log`s here. They are deliberate and will help in debugging any issues for this feature. Hopefully, it should be fine.

agnivade:
Hey @devinbinnie - I am handing this over to you. All that's needed is to actually take from the server config instead of a hardcoded boolean flag. There are 2 lint failures, but they might actually change when we change it to take from config.

Feel free to polish it however way you feel. But the logic is all there, and does not need to change.

devinbinnie:
@agnivade Alright I've setup the correct server config item to drive the use of reliable websockets. I also needed to make a change on the server branch `wsfinal` to add the item to the client config.

One question: currently if the server changes the config item and the websocket drops, reconnection won't work properly until the webapp is refreshed. Are we okay with that or should we be trying to find something more robust to make sure the switch over is seamless. I don't see the case coming up very often outside of a server admin turning on the feature and rebooting the server, disconnecting everyone.

agnivade:
Great question. Doesn't the webapp listen to config change events? If the admin changes the setting and reboots the server, the connection should try to reconnect and in the initialize function, `getConfig(store.getState());` should return the new state?

devinbinnie:
> Great question. Doesn't the webapp listen to config change events? If the admin changes the setting and reboots the server, the connection should try to reconnect and in the initialize function, `getConfig(store.getState());` should return the new state?

Yes you're right, looks like it will call that again.
Funny enough, our reconnect logic really slows down after 8 attempts, going to every 3 minutes instead of 3 seconds (we multiply by the fail count twice), I wonder if that should change.

agnivade:
Yeah, I am aware of that. I don't want to touch that at this point. It should be a proper exponential backoff. Will look into it once the basic implementation gets in.

devinbinnie (approved):

client/websocket_client.tsx
hmhealey: I'm fine with adding these. I think our "no console logs" rule might be a bit aggressive sometimes when they can be really helpful for debugging.

client/websocket_client.tsx
```diff
@@ -55,19 +65,28 @@ export default class WebSocketClient {
...
+            }
 
             if (token) {
                 this.sendMessage('authentication_challenge', {token});
             }
 
             if (this.connectFailCount > 0) {
                 console.log('websocket re-established connection'); //eslint-disable-line no-console
-                if (this.reconnectCallback) {
+                if (!reliableWebSockets && this.reconnectCallback) {
```
hmhealey, line 89: So with the reliable websocket enabled, the reconnect logic is entirely handled when we receive the HELLO request and not when the connection is opened, right?

It's absolutely not necessary, but perhaps it would make sense to do the first connection handling callback in that same case as well

client/websocket_client.tsx
```diff
@@ -132,7 +154,35 @@ export default class WebSocketClient {
...
+                    }
+
+                    // Now we check for sequence number, and if it does not match,
+                    // we just disconnect and reconnect.
+                    if (msg.seq !== this.serverSequence) {
+                        console.log('missed websocket event, act_seq=' + msg.seq + ' exp_seq=' + this.serverSequence); //eslint-disable-line no-console
+                        // We are not calling this.close() because we need to auto-restart.
+                        this.connectFailCount = 0;
+                        this.responseSequence = 1;
+                        this.conn?.close(); // Will auto-reconnect after MIN_WEBSOCKET_RETRY_TIME.
```
hmhealey, line 182: How will closing and reopening the websocket let the client get back in sync? Does this cause the server to resend any missed events? It also looks like we're not calling the old reconnect logic anywhere either which I thought we needed in the case where there was a longer disconnect.

client/websocket_client.tsx
hmhealey: I mean that we could potentially move the `firstConnectCallback` call into the `onmessage` handler like the reconnect instead of having it done in `onopen`. That's just a thought though.

client/websocket_client.tsx
hmhealey: Oh, I see. I didn't realize we already had handling for missed events from back when the sequence numbers were originally added

hmhealey (approved):

hmhealey:
@agnivade Does this need QA review here or is that being done elsewhere?

agnivade:
>  Does this need QA review here or is that being done elsewhere?

Yeah Saturn is doing it on the server PR. He will take all the branches.

client/websocket_client.tsx
agnivade: Ah I see. I would say that it's slightly out of scope for this PR. We can do it any time later.

saturninoabril (approved):
Tested and passed.

PR #7924, opened 2021-04-15
MM-31717: Remove HTTP clustering

agnivade:
Gossip is now GA, so we remove the option
from admin console.

https://mattermost.atlassian.net/browse/MM-31717


hmhealey (approved):
Since we're just removing the setting from the system console, this doesn't need to wait on the server PR, does it?

agnivade:
Yeah, but there's no rush on this. I'll prefer to merge them together.

hmhealey:
I'll assign you then just so no one else happens to merge the PR once it's done reviews.

agnivade:
/update-branch

Gentle ping on this @calebroseland

calebroseland (approved):
(Thanks for your patience, @agnivade)

LGTM, though it seems these keys could be removed from searchableStrings https://github.com/mattermost/mattermost-webapp/blob/removehttp/components/admin_console/admin_definition.jsx#L1390-L1391?

PR #7925, opened 2021-04-15
Fix accesibility problems in files/message selectors in hint and RHS

jespino:
#### Summary
Fix accesibility problems in files/message selectors in hint and RHS

#### Ticket Link
[MM-34652](https://mattermost.atlassian.net/browse/MM-34652)


components/search_hint/search_hint.tsx
```diff
@@ -46,14 +46,20 @@ const SearchHint = (props: Props): JSX.Element => {
                         />
                     </div>
                     <div className='button-container'>
-                        <button onClick={() => props.onSearchTypeSelected && props.onSearchTypeSelected('messages')}>
+                        <button
+                            className={props.highlightedIndex === 0 ? 'highlighted' : ''}
```
michelengelen, line 50: aren't we using something like `clsx` in the webapp?

In this case it is not really needed, but when combining multiple conditional classname assignments it really helps a lot.

Just a suggestion to look into 👾

components/search_results/messages_or_files_selector.tsx
```diff
@@ -26,7 +31,9 @@ export default function MessagesOrFilesSelector(props: Props): JSX.Element {
             <div className='buttons-container'>
                 <span
```
michelengelen, line 32: nothing related to this PR, but semantically `span` should never have action listeners added to them.
Why are we not using `button`s here?

michelengelen (approved):
LGTM

components/search_results/messages_or_files_selector.tsx
jespino: Good point, I going to change it to button.

components/search_hint/search_hint.tsx
jespino: Done, thanks!

components/search_results/messages_or_files_selector.tsx
jespino: Done

nevyangelova (approved):

ogi-m:
1. Accessing the search dropdown options is now possible with the down arrow, but still not with tab - tabbing will still dismiss the dropdown

2. Tabbing in Search Results now selects "Messages" and "Files" as expected, but it doesn't select file results.
This was not specified in the ticket, but I think we should be able to tab through the file results as well @esethna ?

jespino:
@ogi-m 1. That is exactly the same behavior previously existing with search hint, nothing new there. We can fix that, but that is another bug, I think.
2. I going to fix that, makes a lot of sense to have that.

jespino:
@ogi-m 2. fixed, PTAL.

esethna:
Looks good overall. @ogi-m can you file another issue for the tabbing bug you identified.

I think the expected here is that the search pop-over should not appear unless the user hits ENTER (or space) when the search field is focused (note this shouldn't change the behaviour of using the mouse to click on the search field and see the pop-over). That way we can be sure that the users tabs are directed within the modal (if they hit enter or space on the search field to open the modal) OR if they are trying to tab past the search field to the next icons.

esethna:
@ogi-m can we please test this with a screen reader to ensure there are no issues with read out on the buttons or other elements?

ogi-m:
#2 Is fixed now, thanks @jespino 

Opened a ticket for the tab issue - https://mattermost.atlassian.net/browse/MM-34969

@esethna The current behaviour is that the dropdown opens when focused with the keyboard - should we file a separate ticket for that?

esethna:
@ogi-m 0/5, could be part of the same ticket you filed above, but not blocking this PR

matthewbirtch:
> I think the expected here is that the search pop-over should not appear unless the user hits ENTER (or space) when the search field is focused (note this shouldn't change the behaviour of using the mouse to click on the search field and see the pop-over). That way we can be sure that the users tabs are directed within the modal (if they hit enter or space on the search field to open the modal) OR if they are trying to tab past the search field to the next icons.

@eric I'm not sure this is the best approach. Right now, I think it works well that I can tab to focus on the search input and start typing right away. With the proposal above, we'd be adding another keystroke before you can start typing in the search input. Plus, `ENTER` is used to execute the search. It feels odd to tab to focus on the search field and press ENTER before I type anything.

Can we not just tab through the `messages` and `files` options in popover?

The expected outlined in the ticket makes sense to me:
> Expected: "Messages" is highlighted and selected in the dropdown, tabbing will move the selection to the “Files” option. Hitting Enter should add the “Messages” or “Files” element in the search box.

esethna:
@matthewbirtch good call, that makes sense, I agree we don't want to require another keystroke before typing a search query

ogi-m:
Thanks guys! Approving the PR.

ogi-m (approved):

PR #7932, opened 2021-04-16
Feature/mm 34670

michelengelen:
#### Summary
some changes to the quick switch modal search results:
- moves the status indicator to the avatar (similar to sidebar)
- added the `@` to the username for DM-Channels
- adds channel slugs in relation to the changes made in #7476 

Changes are according to this [FIGMA design](https://www.figma.com/file/J0poAU9hou1ajA9xOPrO58/MM-34670-User-Status-in-Channel-Switcher?node-id=0%3A1)

#### Ticket Link
Fixes [MM-34670](https://mattermost.atlassian.net/browse/MM-34670) and [MM-33837](https://mattermost.atlassian.net/browse/MM-33837)

#### Screenshots
Screenshot 2021-04-16 at 09 23 19


components/suggestion/switch_channel_provider.jsx
```diff
@@ -115,15 +122,31 @@ class SwitchChannelSuggestion extends Suggestion {
         } else {
             icon = (
                 <div className='pull-left'>
-                    <Avatar
+                    <ProfilePicture
```
jespino, line 125: The profile picture itself is a bit heavy and is though for places where the user itself is clickable and should show the profile popover. I'm not strictly against this, but I think we should have something like "AvatarWithStatus" component, used in the `ProfilePicture` and here, (probably in other places too).

michelengelen: there is actually a conditional inside the `ProfilePicture` that renders the whole component without the action (I think it is a span).

I am not entirely sure if we should really introduce a new component. IMO the Avatar could have a conditional, that shows the status, when it is set.

jespino (approved):
LGTM

matthewbirtch (changes requested):
Looks good @michelengelen. Only thing we need to do is remove the status from bots (similar to how we do for the sidebar) since bot status is not important.

image

package-lock.json
```diff
@@ -15354,10 +15354,7 @@
       "version": "1.3.1",
       "resolved": "https://registry.npmjs.org/full-icu/-/full-icu-1.3.1.tgz",
       "integrity": "sha512-VMtK//85QJomhk3cXOCksNwOYaw1KWnYTS37GYGgyf7A3ajdBoPGhaJuJWAH2S2kq8GZeXkdKn+3Mfmgy11cVw==",
-      "dev": true,
```
devinbinnie, line 15357: We can remove these `package-lock.json` changes.

devinbinnie (approved):

michelengelen:
Hi @matthewbirtch ... I added a change to the PR so that the status is not shown anymore for bots

matthewbirtch:
> Hi @matthewbirtch ... I added a change to the PR so that the status is not shown anymore for bots

Thanks! I didn't notice this previously, but now it looks like the avatar is sitting too high/off-center vertically?
Screen Shot 2021-04-19 at 9 27 17 AM

michelengelen:
Hey @matthewbirtch good catch.

Although there might be different approaches to handle this.

IMO it should be like this:

Screenshot 2021-04-19 at 16 55 43

visually centering the Avatar and keep the status with a little offset to bottom-right.
WDYT?

matthewbirtch:
> visually centering the Avatar and keep the status with a little offset to bottom-right.

100% agree. I think we could reduce the offset by a few pixels as well

matthewbirtch (approved):

jgilliam17 (approved):
Thank you @michelengelen 
Tested, looks good to merge.
- Verified DMs on channel switcher show user status overlapping with their avatar, @ is prepended to usernames for DM channels and status is not shown on bot icons.
- Verified channels have URL “slug” information displayed on channel switcher.

PR #7944, opened 2021-04-19
MM-34437 Preventing infinite trial requests

harshilsharma63:
#### Summary
Preventing infinite trial requests from Admin Console.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34437

#### Release Note
* API additions— new endpoint `GET` `/trial-license/prev` for fetching last used trial license.

#### Related Pull Requests
- Has server changes https://github.com/mattermost/mattermost-server/pull/17442

#### Screenshots
Screenshot 2021-04-19 at 3 38 59 PM


components/admin_console/license_settings/license_settings.jsx
```diff
@@ -61,11 +65,12 @@ export default class LicenseSettings extends React.PureComponent {
         this.fileInputRef = React.createRef();
     }
 
-    componentDidMount() {
+    async componentDidMount() {
         if (!this.props.enterpriseReady) {
             this.reloadPercentage();
         }
         this.props.actions.getLicenseConfig();
+        this.props.actions.getPrevTrialLicense();
```
BenCookie95, line 73: We can probably send these requests in a promise all with an `await`?

components/admin_console/license_settings/license_settings.jsx
```diff
@@ -135,6 +140,7 @@ export default class LicenseSettings extends React.PureComponent {
 
         await this.props.actions.getLicenseConfig();
         this.setState({fileSelected: false, fileName: null, serverError: null, removing: false});
+        this.props.actions.getPrevTrialLicense();
```
BenCookie95, line 143: combine `this.props.actions.getLicenseConfig()` & `this.props.actions.getPrevTrialLicense()` into promise all and use an `await`?

package-lock.json
```diff
@@ -12784,8 +12784,7 @@
           "version": "3.0.0",
           "resolved": "https://registry.npmjs.org/get-stream/-/get-stream-3.0.0.tgz",
           "integrity": "sha1-jpQ9E1jcN1VQVOy+LtsFqhdO3hQ=",
-          "dev": true,
-          "optional": true
+          "dev": true
```
BenCookie95, line 12787: I'm not sure if all these lock file changes should be here?

components/admin_console/feature_discovery/feature_discovery.tsx
```diff
@@ -108,41 +158,7 @@ export default class FeatureDiscovery extends React.PureComponent<Props, State>
...
-                        />
-                    </a>
-                    {gettingTrialError}
-                    <p className='trial-legal-terms'>
-                        <FormattedMarkdownMessage
-                            id='admin.license.trial-request.accept-terms'
-                            defaultMessage='By clicking **Start trial**, I agree to the [Mattermost Software Evaluation Agreement](!https://mattermost.com/software-evaluation-agreement/), [Privacy Policy](!https://mattermost.com/privacy-policy/), and receiving product emails.'
-                        />
-                    </p>
+                    {this.props.prevTrialLicense && this.props.prevTrialLicense.IsLicensed === 'true' ? Utils.renderPurchaseLicense() : this.renderStartTrial(learnMoreURL, gettingTrialError)}
```
hahmadia, line 161: I love me some TypeScript optional chaining 😸 !!
```suggestion
                    {this.props.prevTrialLicense?.IsLicensed === 'true' ? Utils.renderPurchaseLicense() : this.renderStartTrial(learnMoreURL, gettingTrialError)}
```

components/admin_console/feature_discovery/feature_discovery.tsx
```diff
@@ -27,10 +28,13 @@ type Props = {
 
     imgPath: string;
 
+    prevTrialLicense: ClientLicense;
+
     stats?: Dictionary<number | AnalyticsRow[]>;
     actions: {
         requestTrialLicense: (users: number, termsAccepted: boolean, receiveEmailsAccepted: boolean, featureName: string) => Promise<{error?: string; data?: null}>;
         getLicenseConfig: () => void;
+        getPrevTrialLicense: any;
```
hahmadia, line 37: Probably want this (trying to avoid the use of any when possible)
```suggestion
        getPrevTrialLicense: () => void;
```

BenCookie95, line 37: This needs to be typed properly

components/admin_console/license_settings/license_settings.jsx
```diff
@@ -357,31 +395,7 @@ export default class LicenseSettings extends React.PureComponent {
...
-                        </button>
-                    </p>
-                    {gettingTrialError}
-                    <p className='trial-legal-terms'>
-                        <FormattedMarkdownMessage
-                            id='admin.license.trial-request.accept-terms'
-                            defaultMessage='By clicking **Start trial**, I agree to the [Mattermost Software Evaluation Agreement](!https://mattermost.com/software-evaluation-agreement/), [Privacy Policy](!https://mattermost.com/privacy-policy/), and receiving product emails.'
-                        />
-                    </p>
+                    {this.props.prevTrialLicense && this.props.prevTrialLicense.IsLicensed === 'true' ? Utils.renderPurchaseLicense() : this.renderStartTrial(isDisabled, gettingTrialError)}
```
hahmadia, line 398:
```suggestion
                    {this.props.prevTrialLicense?.IsLicensed === 'true' ? Utils.renderPurchaseLicense() : this.renderStartTrial(isDisabled, gettingTrialError)}
```

BenCookie95 (changes requested):
Just a few minor things

components/announcement_bar/contact_sales/contact_sales.tsx
```diff
@@ -0,0 +1,32 @@
...
+export interface Props {
+    buttonTextElement: JSX.Element;
+    eventID?: string;
+}
+
+const ContactSales: React.FC<Props> = (props: Props) => {
+    const handleContactSalesLinkClick = async (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+        e.preventDefault();
+        trackEvent('admin', props.eventID || 'in_trial_contact_sales');
+        window.open('https://mattermost.com/contact-us/', '_blank');
```
hahmadia, line 19: Since the link isn't leading to anything specific to sales but rather a general contact us form,
should we consider renaming this component and repurposing it as a `ContactUs` component where you can change the text it displays but ultimately leads to the same form?

hahmadia: I noticed on this website the inquiry types ... maybe if it accepts query parameters to fill those fields it, we can have props that you can pass to this component that fill those query params (I've asked in developers channel)

package-lock.json
hahmadia: No they shouldn't

packages/mattermost-redux/src/actions/admin.ts
```diff
@@ -445,6 +445,24 @@ export function removeLicense(): ActionFunc {
...
+        let data;
+        try {
+            data = await Client4.getPrevTrialLicense();
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            return {error};
+        }
+
+        dispatch(batchActions([
+            {type: AdminTypes.PREV_TRIAL_LICENSE_SUCCESS, data},
```
hahmadia, line 459: no need for batch actions since it's only one action we are dispatching

packages/mattermost-redux/src/client/client4.ts
```diff
@@ -3120,6 +3120,13 @@ export default class Client4 {
         );
     };
 
+    getPrevTrialLicense = () => {
+        return this.doFetch<StatusOK>(
```
hahmadia, line 3124: I think the `StatusOK` should be changed to the returned data type. I believe the backend does not return StatusOK but rather information which should be the proper type here

package-lock.json
hahmadia: This file should not be commited

hahmadia (changes requested):
Some comments

components/admin_console/license_settings/license_settings.jsx
harshilsharma63: We can do that but that won't be the redux way (action -> reducer -> state).  That's why we didn't consume the function response.

components/admin_console/feature_discovery/feature_discovery.tsx
harshilsharma63: Yes, was missing. Fixed now.

components/admin_console/feature_discovery/feature_discovery.tsx
harshilsharma63: Done.

components/admin_console/feature_discovery/feature_discovery.tsx
harshilsharma63: Done.

components/admin_console/license_settings/license_settings.jsx
harshilsharma63: Done.

components/announcement_bar/contact_sales/contact_sales.tsx
harshilsharma63: 1. Done the part of having a default text in the component itself and allowing the consumer to override it.
2. For the pre-filled form, we need some way to collect those inputs. So collecting them in Mattermost then passing them over to the contact us page would be the same for the user as they are effectively entering the information once. So I think we can pass this.

harshilsharma63: Let me know if the second point makes sense if you have some idea.

package-lock.json
harshilsharma63: Reverted it.

packages/mattermost-redux/src/actions/admin.ts
harshilsharma63: Nice catch! FIxed.

packages/mattermost-redux/src/client/client4.ts
harshilsharma63: Fixed.

components/admin_console/license_settings/license_settings.jsx
hahmadia: You don't need to consume it to use Promise.all. You can use the following example to do what ben is saying
This example was pulled from `group_details.jsx`
```js
        await Promise.all([
            this.props.actions.getGroupSyncables(this.props.groupID, Groups.SYNCABLE_TYPE_CHANNEL),
            this.props.actions.getGroupSyncables(this.props.groupID, Groups.SYNCABLE_TYPE_TEAM),
        ]);
```

hahmadia: It'll run all those actions in parallel and it doesn't need to wait one by one to finish and go on to the next

hahmadia: We just do these actions in parallel and don't need to consume it still gets updated in state.

components/announcement_bar/contact_sales/contact_us.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const handleContactSalesLinkClick = async (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+        e.preventDefault();
+        trackEvent('admin', props.eventID || 'in_trial_contact_sales');
+        window.open('https://mattermost.com/contact-us/', '_blank');
+    };
+
+    return (
+        <button
+            className='contact-sales'
+            onClick={(e) => handleContactSalesLinkClick(e)}
```
hahmadia, line 28:
```suggestion
            onClick={(e) => handleContactUsLinkClick(e)}
```

components/announcement_bar/contact_sales/contact_us.tsx
```diff
@@ -0,0 +1,38 @@
...
+        >
+            <FormattedMessage
+                id={props.titleID || t('admin.license.trialCard.contactSales')}
+                defaultMessage={'Contact sales'}
+            />
```
hahmadia, line 33: Make contactUsButton a react component you can pass in. Doing this props.titleID || t('') is not something I like doing for formattedmessage.
```suggestion
            {contactUsButton}
```

Something like 
```
                buttonTextElement={
                    <FormattedMessage
                        id='admin.license.trialCard.purchase_license'
                        defaultMessage='Purchase a license'
                    />
                }
```

components/announcement_bar/contact_sales/contact_us.scss
```diff
@@ -0,0 +1,19 @@
+@import 'utils/variables';
+
+.contact-sales {
```
hahmadia, line 3:
```suggestion
.contact-us {
```

components/announcement_bar/contact_sales/contact_us.tsx
```diff
@@ -0,0 +1,38 @@
...
+const ContactUs: React.FC<Props> = (props: Props) => {
+    const handleContactSalesLinkClick = async (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+        e.preventDefault();
+        trackEvent('admin', props.eventID || 'in_trial_contact_sales');
+        window.open('https://mattermost.com/contact-us/', '_blank');
+    };
+
+    return (
+        <button
+            className='contact-sales'
```
hahmadia, line 27:
```suggestion
            className='contact-us'
```

components/admin_console/license_settings/trial_license_card/trial_license_card.tsx
```diff
@@ -75,15 +69,7 @@ const TrialLicenseCard: React.FC<Props> = ({license}: Props) => {
...
-                <button
-                    className='ContactSales'
-                    onClick={(e) => handleContactLinkClick(e)}
-                >
-                    <FormattedMessage
-                        id='admin.license.trialCard.contactSales'
-                        defaultMessage='Contact sales'
-                    />
-                </button>
+                <ContactUs/>
```
hahmadia, line 72:
```suggestion
                <ContactUsButton/>
```

components/admin_console/license_settings/trial_license_card/trial_license_card.scss
```diff
@@ -17,10 +17,7 @@
         display: inline-block;
         margin-top: 0;
     }
-    button.ContactSales {
-        background-color: var(--sys-center-channel-bg);
-        color: var(--sys-button-bg);
-        border: 1px solid var(--sys-button-bg);
+    button.contact-sales {
```
hahmadia, line 20:
```suggestion
    button.contact-us {
```

hahmadia (changes requested):
Some reviews

components/announcement_bar/contact_sales/contact_us.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import {t} from 'utils/i18n';
+
+export interface Props {
+    titleID?: string;
+    eventID?: string;
+}
+
+const ContactUs: React.FC<Props> = (props: Props) => {
+    const handleContactSalesLinkClick = async (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
```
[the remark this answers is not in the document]
harshilsharma63: Done.

components/announcement_bar/contact_sales/contact_us.tsx
harshilsharma63: Done.

components/announcement_bar/contact_sales/contact_us.tsx
harshilsharma63: But we would still keep a default message in the component, right? It would be like this -

```
<button
            className='contact-us'
            onClick={(e) => handleContactUsLinkClick(e)}
        >
            {props.titleID || (
                <FormattedMessage
                    id={'admin.license.trialCard.contactSales'}
                    defaultMessage={'Contact sales'}
                />
            )}
        </button>
```

components/announcement_bar/contact_sales/contact_us.scss
harshilsharma63: Done.

components/announcement_bar/contact_sales/contact_us.tsx
harshilsharma63: Done.

components/admin_console/license_settings/trial_license_card/trial_license_card.tsx
harshilsharma63: Done. Renamed the component.

components/admin_console/license_settings/trial_license_card/trial_license_card.scss
harshilsharma63: Done.

components/admin_console/license_settings/license_settings.jsx
harshilsharma63: As discussed with @hahmadia, we don't need t use `promise.all` as `getLicenseConfig` and `getPrevTrialLicense ` are already non-blocking.

components/admin_console/license_settings/license_settings.jsx
harshilsharma63: @BenCookie95 we don't need to use promise.all here but I have moved `getPrevTrialLicense` before `getLicenseConfig` so `getLicenseConfig` runs in background without affecting the way we're using `getLicenseConfig` right now.

components/announcement_bar/contact_sales/contact_us.tsx
harshilsharma63: I've done this now.

components/admin_console/license_settings/license_settings.jsx
```diff
@@ -505,7 +520,12 @@ export default class LicenseSettings extends React.PureComponent {
     renderE0Content = () => {
         let serverError = '';
         if (this.state.serverError) {
-            serverError = <div className='col-sm-12'><div className='form-group has-error'><label className='control-label'>{this.state.serverError}</label></div></div>;
+            serverError = (<div className='col-sm-12'><div className='form-group has-error'><label className='control-label'>
+                <Markdown
+                    enableFormatting={true}
+                    message={this.state.serverError}
+                />
+            </label></div></div>);
```
hahmadia, line 528: Possible to format this where the opening and closing tags are on the same line? non-blocking

components/announcement_bar/contact_sales/contact_us.tsx
```diff
@@ -0,0 +1,38 @@
...
+import React from 'react';
+
+import {FormattedMessage} from 'react-intl';
+
+import {trackEvent} from 'actions/telemetry_actions';
+import './contact_us.scss';
+
+export interface Props {
+    buttonTextElement?: JSX.Element;
+    eventID?: string;
```
hahmadia, line 13: My gut feeling says that these should be required and passed down instead of optional.
Do you think it's better to keep as a hybrid like component? I'm not a huge fan of the `|| contact-sales` aspect.

I'd be more in favor of making these required and passing them down from the contact sales so we don't need the `||` operator.

Or you could do it like the purchase link component where only the EventID is optional but buttomTextElement is required. This would be an ideal solution.

components/announcement_bar/contact_sales/contact_us.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+    return (
+        <button
+            className='contact-us'
+            onClick={(e) => handleContactUsLinkClick(e)}
+        >
+            {props.buttonTextElement || (
+                <FormattedMessage
+                    id={'admin.license.trialCard.contactSales'}
+                    defaultMessage={'Contact sales'}
```
hahmadia, line 31: I'd say lets pass this down as a proper and make butttonTextElement required.

hahmadia (commented):
Good fixes! Just some small comments

components/admin_console/feature_discovery/feature_discovery.tsx
```diff
@@ -68,6 +76,48 @@ export default class FeatureDiscovery extends React.PureComponent<Props, State>
         this.props.actions.getLicenseConfig();
     }
 
+    renderStartTrial = (learnMoreURL: string, gettingTrialError: React.ReactNode) => {
+        return (
+            <React.Fragment>
```
BenCookie95, line 81: Copied from Hossein's comment: we can use `<> </>` as a short hand for react fragment.

BenCookie95 (commented):
Only one minor comment! Nice work!

components/admin_console/license_settings/license_settings.jsx
```diff
@@ -61,11 +66,12 @@ export default class LicenseSettings extends React.PureComponent {
         this.fileInputRef = React.createRef();
     }
 
-    componentDidMount() {
+    async componentDidMount() {
```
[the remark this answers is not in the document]
harshilsharma63: Done.

components/admin_console/license_settings/license_settings.jsx
```diff
@@ -186,6 +193,38 @@ export default class LicenseSettings extends React.PureComponent {
         setTimeout(this.checkRestarted, 1000);
     }
 
+    renderStartTrial = (isDisabled, gettingTrialError) => {
+        return (
+            <React.Fragment>
```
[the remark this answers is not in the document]
harshilsharma63: I think's its more clear with `React.Fragment`. Also we have mixed usage of both variants. And `<>` breaks tag matcher in my IDE, so I think `React.Fragment` avoids that issue, without either variant offering any advantage as such.

components/admin_console/license_settings/license_settings.jsx
harshilsharma63: I think it looks cleaner this way.

components/announcement_bar/contact_sales/contact_us.tsx
harshilsharma63: Well, the whole idea of creating a contact us component in one of the comments was to have it with the default texts, and only as a flexibility, to allow overriding the content, if ever needed anywhere in the app. 

Making the params mandator defeats that purpose.

Removing the params vs having optional params doesn't make much difference as I don't think it's worth the effort going in and making these small changes that have no effect.

components/announcement_bar/contact_sales/contact_us.tsx
harshilsharma63: No, it's a standard text and passing the same param from everywhere doesn't make sense. The default value is here because in 99.9% cases this is the text we want. The ability to override is there as a flexibility as it took only 2 lines of code.

components/admin_console/feature_discovery/feature_discovery.tsx
harshilsharma63: I think it's really subjective unless we have a project-wide convention. Either variant has no advantage over the other and like I mentioned in the other review comment, `<>` breaks tag matching in my IDE so that's 0:1 for `<>` vs `React.Fragment`.

hahmadia (approved):

BenCookie95 (approved):

furqanmlk (approved):
Following Scenarios are passed

<google-sheets-html-origin><style type="text/css"></style>
Case | Expected | Actaul | License Type | Applied With API | Result
-- | -- | -- | -- | -- | --
With Flag 29 days | Should Apply | Is Applied | Trial | Is Applied | Pass
Without Flag 29 days | Should Apply | Is Applied | Paid | Is Applied | Pass
With Flag less than 29 days | Should Apply | Is Applied | Trial | Is Applied | Pass
With Flag 30 days | Should Not Apply | Did Not Apply | Trial | Did Not Apply | Pass
Without Flag 30 days | Should Not Apply | Did Not Apply | Trial | Did Not Apply | Pass
With Flag 31 days | Should Apply | Is Applied | Trial | Is Applied | Pass
With Flag more than 31 days | Should Apply | Is Applied | Trial | Is Applied | Pass
Without Flag more than 31 days | Should Apply | Is Applied | Paid | Is Applied | Pass

</google-sheets-html-origin>

harshilsharma63:
@amyblais this can be merged now.

PR #7950, opened 2021-04-19
MM-34975 - hide payment information to free trial customers

pablovelezvidal:
#### Summary
Hide payment information for trialing customers

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34975

### DIFF
--- a/components/admin_console/admin_definition.jsx
+++ b/components/admin_console/admin_definition.jsx
@@ -182,11 +182,11 @@ export const it = {
     enterpriseReady: (config, state, license, enterpriseReady) => enterpriseReady,
     licensed: (config, state, license) => license.IsLicensed === 'true',
     licensedForFeature: (feature) => (config, state, license) => license.IsLicensed && license[feature] === 'true',
-    isPaidTier: (config, state, license, enterpriseReady, consoleAccess, cloud) => {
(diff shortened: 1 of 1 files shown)


nickmisasi (approved):

components/admin_console/admin_definition.jsx
```diff
@@ -182,11 +182,11 @@ export const it = {
     enterpriseReady: (config, state, license, enterpriseReady) => enterpriseReady,
     licensed: (config, state, license) => license.IsLicensed === 'true',
     licensedForFeature: (feature) => (config, state, license) => license.IsLicensed && license[feature] === 'true',
-    isPaidTier: (config, state, license, enterpriseReady, consoleAccess, cloud) => {
+    hidePaymentInfo: (config, state, license, enterpriseReady, consoleAccess, cloud) => {
         if (!cloud?.subscription) {
             return false;
         }
```
marianunez, line 188: If now this is to `hide` then if there is no cloud subscription then we do want to hide so it shouldn't return `false` here anymore.

Wondering if we just remove this check and add the null checks below?

pablovelezvidal: thanks, good catch! fixing it now.

marianunez (approved):

nickmisasi (approved):

PR #7969, opened 2021-04-21
MM-35074: short circuit slash command hooks

lieut-data:
#### Summary
When the Confluence plugin consumes a slash command for client-side processing (i.e. `/confluence subscribe`), it signals this by returning an empty object as per the required interface.

However, the webapp passes this empty object through all other registered slash command hooks. When the Confluence and Incident Collaboration plugins are installed (and Incident Collaboration "registers last"), Incident Collaboration receives this empty object and [mistakenly initializes](https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/98c01d6a3e76bfa48d9dd4860136d4f9a7dbde85/webapp/src/slash_command.ts#L9-L11) the `undefined` args parameter in testing whether it should process the slash command. The final result consumed by the webapp is now partially initialized, leading to a JavaScript exception and a create post component that won't process slash commands until it mounts afresh.

While we can address this in Incident Collaboration, it's unexpected that the webapp would notify plugins at all in this case, given there is literally no metadata on which to try to do anything meaningful. Short-circuit in this case, avoiding the original report altogether.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-35074

#### Release Note
```release-note
Short-circuit slash command hooks when a plugin consumes one for client-side processing.
```

### DIFF
--- a/actions/hooks.js
+++ b/actions/hooks.js
@@ -50,6 +50,12 @@ export function runSlashCommandWillBePostedHooks(originalMessage, originalArgs) 
 
                 message = result.message;
                 args = result.args;
+
+                // The first plugin to consume the slash command by returning an empty object
+                // should terminate the processing by subsequent plugins.
+                if (Object.keys(result).length === 0) {
+                    break;
+                }
             }
         }
 
--- a/actions/hooks.test.js
+++ b/actions/hooks.test.js
@@ -248,11 +248,36 @@ describe('runSlashCommandWillBePostedHooks', () => {
         const result = await store.dispatch(runSlashCommandWillBePostedHooks(message, args));
 
         expect(result.data).toEqual({message: '/testabc', args});
-        expect(hook1).toHaveBeenCalledWith(message, args);
-        expect(hook2).toHaveBeenCalled();
-        expect(hook2).not.toHaveBeenCalledWith(message, args);
-        expect(hook3).toHaveBeenCalled();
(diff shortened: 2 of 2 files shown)


lieut-data:
This was a ... fascinating confluence of the three hats I'm wearing this week as SET Lead, Web Platform Lead, and Incident Collaboration Lead. I imagine it would have been hard to find had I not had the Incident Collaboration plugin installed by default :)

cpoile (approved):
Wow, that's a surprising bug. Nice!

> This was a ... fascinating confluence of the three hats I'm wearing this week

I see what you did there.

cpoile:
Filed a bug for the Incident Collaboration plugin: https://mattermost.atlassian.net/browse/MM-35096

hmhealey (approved):

hanzei:
cc @DHaussermann

DHaussermann:
@lieut-data can you please clarify the scope for this?
1. It's unclear to me what type of webapp regression testing may be needed here.
2. It seems possible to test this by installing both plugins and using them without an actual working confluence instance. This would make testing much easier. Is that correct?

lieut-data:
@DHaussermann,

1. The scope should entail any kind of slash command interaction, with an emphasis on plugins that handle some slash commands on the frontend (vs. the default behaviour of routing to the backend). I think Jira would be an excellent additional test case.
2. Yes, you can reproduce without an actual working Confluence instance: that's exactly how I tested.

PR #7983, opened 2021-04-23
[MM-34178] Using zoom for non-Mozilla browsers on emojis for less blurry images

devinbinnie:
#### Summary
On Chrome, at certain zoom levels or desktop resolutions, our base emojis would look blurry compared to custom emojis.
This had to do with the `transform: scale()` function at certain values smoothing out the scaling by blurring.

Instead, we are using the non-standard `zoom` CSS attribute which seems to do a better job of scaling without blurring. This will only be used on browsers other than Firefox as Firefox doesn't support `zoom` (but also doesn't exhibit this behaviour)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34178

#### Screenshots
Before:
Screen Shot 2021-04-23 at 7 08 21 PM

After:
Screen Shot 2021-04-23 at 7 13 13 PM

#### Release Note
```release-note
NONE
```


hmhealey (approved):

jgilliam17 (approved):
Thanks @devinbinnie 
Tested, looks good to merge.
- Verified base vs. custom emoji, emoji images are no longer blurry.

PR #7984, opened 2021-04-24
MM-32801 : Webapp: In mention autocomplete, remove (Nickname) when (you) is present 

M-ZubairAhmed:
#### Summary
Removed nickname in the mentions suggestion list for the current user.

#### Ticket Link

 Fixes https://github.com/mattermost/mattermost-server/issues/16947

#### Related Pull Requests
N/A

#### Screenshots

image

#### Release Note
None

```release-note
```


M-ZubairAhmed:
❓ 
@hmhealey Let me know if the UI is looking good, then i can write up the tests for it

components/suggestion/at_mention_provider/at_mention_suggestion.jsx
```diff
@@ -112,12 +112,22 @@ export default class AtMentionSuggestion extends Suggestion {
...
             itemname = item.username;
 
-            if ((item.first_name || item.last_name) && item.nickname) {
-                description = `${Utils.getFullName(item)} (${item.nickname})`;
-            } else if (item.nickname) {
-                description = `(${item.nickname})`;
-            } else if (item.first_name || item.last_name) {
-                description = `${Utils.getFullName(item)}`;
+            const fullName =
+                item.first_name || item.last_name ? `${Utils.getFullName(item)}` : '';
```
hmhealey, line 116: The template strings for both of these aren't necessary. The full name one isn't needed at all, and the nickname one can have the brackets moved down to where we use the nickname

components/suggestion/at_mention_provider/at_mention_suggestion.jsx
```diff
@@ -112,12 +112,22 @@ export default class AtMentionSuggestion extends Suggestion {
...
+            const nickName = item.nickname ? `(${item.nickname})` : '';
```
hmhealey, line 117: Some more nitpicking, but nickname shouldn't have the second N capitalized since it's only one word instead of two

hmhealey (changes requested):
UI and everything looks good, but I had a couple things to nitpick about in the code. Tests would be very much appreciated though too.

devinbinnie (approved):
LGTM other than @hmhealey's comments.
Thanks @M-ZubairAhmed!

components/suggestion/at_mention_provider/at_mention_suggestion.jsx
M-ZubairAhmed: ✅  Updated as per request/discussion

components/suggestion/at_mention_provider/at_mention_suggestion.jsx
M-ZubairAhmed: ✅  Updated as per request/discussion

M-ZubairAhmed:
@hmhealey i am thinking how can i write the test, since the test all written are towards the functions not the text inside of jsx component. Should i write e2e instead? or can it be done in jest itself, if yes then can you help me out?

hmhealey:
I think unit tests with [enzyme](https://enzymejs.github.io/enzyme/) would be your best bet here. E2E tests could do the same thing, but they're slower and more complicated to run.

If you render the component using Enzyme's `shallow` function, you can control what props you pass in and use something like
```js
const wrapper = shallow(<AtMentionSuggestion props={...});

// You'll need a jQuery-style selector to find the text. I think `.text()` is the method you want,
// but it might be called something else
expect(wrapper.find('.something').text()).toBe('First Last (You)');
```
or you can just use a snapshot test which will fail in the future if something ever changes
```js
```
There's a ton of examples of snapshot tests around for other components if you want one.

hmhealey:
@M-ZubairAhmed Are you still wanting to add unit tests for this? I don't think it's strictly necessary here since it's fairly simple logic.

M-ZubairAhmed:
@hmhealey can you please retest, i added the tests

hmhealey (approved):
Cool, those look good to me. Thanks for adding them since I wasn't expecting tests with this 😄

jgilliam17 (approved):
Thank you @M-ZubairAhmed 
Tested, looks good to merge.
- Verified (Nickname) is removed from mention autocomplete when (you) is present.

PR #7996, opened 2021-04-26
MM-32786: Root post thread footer

calebroseland:
#### Summary
Adds ThreadFooter to root posts

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32786
~~https://mattermost.atlassian.net/browse/MM-35017 (baby story: exclude root post user from ThreadFooter avatars)~~

#### Screenshots
CleanShot 2021-04-26 at 09 59 19@2x

#### Release Note

```release-note
Added ThreadFooter to root posts in channels as part of Collapsed Reply Threads
```


components/post_view/post/index.js
```diff
@@ -40,8 +44,21 @@ function makeMapStateToProps() {
...
         const channel = getChannel(state, post.channel_id);
+        const replyCount = getReplyCount(state, post) || post.reply_count || 0;
+        const collapsedThreads = isCollapsedThreadsEnabled(state);
+
+        let thread;
+        if (collapsedThreads && !post.root_id && replyCount) {
+            thread = getThreadOrSynthetic(state, post);
+
+            if (thread.is_following) {
+                fetchThread(currentUserId, currentTeamId, post.id);
```
Willyfrog, line 59: If the thread is a long one, would this affect the performance?

It also seems that all the info that is needed is used by the footer, so wouldn't it be better if we just pass the id to the footer and have that get the thread?

This will also prevent aditional thread fetches as the thread id won't change (only the first time that it goes from being a single post to being a thread) so there would only be one fetch regardless of any state change affecting the post.

components/post_view/post/post.test.js
```diff
@@ -183,4 +184,58 @@ describe('Post', () => {
             });
         }
     });
+
+    test('should show the thread footer for root posts', () => {
```
Willyfrog, line 188: we should probably have a test to see that the thread footer doesn't appear in case of it being a comment or a single post with no comments

Willyfrog (changes requested):
left a couple of coments, but looking good in general

components/post_view/post/index.js
calebroseland: The thread here is just the `UserThread`, not the posts in it, so the performance would be consistent. But that's a good point to move it to ThreadFooter.. I'll do that.

esethna:
@jespino please help prioritize a review when you get a chance

calebroseland:
WIP label context: shoring up a few unit test cases, aiming to have done today

esethna:
@Willyfrog @jespino if you guys can please help with a priority dev review as we'd like to merge this asap. Thanks!

esethna:
Minor issue I notices, after clicking the "follow" or "following" button in the footer, the hover state is not removed even if I'm not hovering directly over the post anymore. I have to click away to remove the hover state @calebroseland

calebroseland:
> Minor issue I noticed, after clicking the "follow" or "following" button in the footer, the hover state is not removed even if I'm not hovering directly over the post anymore. I have to click away to remove the hover state @calebroseland
> 
> image

That's because of the focus-state keeping it open, I believe. But it does highlight a discrepancy—it only activates for focus within thread_footer, does not focus of other post content, as the hover does..

@matthewbirtch could we get your take—what should focusing-within-post do?

matthewbirtch (changes requested):
Thanks @calebroseland. Looks great. Just a few things

1. For any thread that only has the original author in it, we don't show any avatar and this feels a bit odd. I think maybe we should consider the root poster a participant if they reply to their own message, but only if they post a reply. Without the avatar, the footer looks misaligned
Screen Shot 2021-05-05 at 5 15 40 PM

2. This one is minor, but maybe we should add 4px of space between the two buttons. It feels a bit off when you hover and the are right up against eachother:
image

3. If I unfollow a thread by clicking the 'Following' button, it stays in the hovered state until I click somewhere else:
image

 

Other than that, it's working great!

jgilliam17:
Thanks @calebroseland, looks great. Just adding few issues
On the test server I need to refresh to hide the replies on the main channel view
Screen Shot 2021-05-05 at 5 41 07 PM

And this is probably out of scope here, profile icon without background color looks a bit off on hover with the outline (maybe it's just me 🙂 )
Screen Shot 2021-05-05 at 5 51 30 PM

calebroseland:
> Thanks @calebroseland, looks great. Just adding few issues
> On the test server I need to refresh to hide the replies on the main channel view
> Screen Shot 2021-05-05 at 5 41 07 PM
> 
> And this is probably out of scope here, profile icon without background color looks a bit off on hover with the outline (maybe it's just me 🙂 )
> Screen Shot 2021-05-05 at 5 51 30 PM

The first one is handled over in #8027

But the transparent-bg-avatar issues is indeed one that may need fixing, I'll take a look at it. There's a similar issue with `focus` causing it to become a squircle.

calebroseland:
@matthewbirtch @jgilliam17 what about this? (Avatar having a fully-opaque BG that matches the '+3' BG)
CleanShot 2021-05-05 at 17 39 00@2x

matthewbirtch:
> @matthewbirtch @jgilliam17 what about this? (Avatar having a fully-opaque BG that matches the '+3' BG)
> 
> CleanShot 2021-05-05 at 17 39 00@2x
> 
> CleanShot 2021-05-05 at 17 42 07@2x
> 
> 

Works for me!

matthewbirtch:
> 1. What should focusing anywhere-within-post do?
I'm not sure I understand the question. 

> 2. Should `focus` on Reply or Follow cause that button to change background (same styling change as hover), or should the only focus indicator be the standard blue outline?
I think the blue outline is fine for now. We will eventually have standardized focus states for all buttons once the compass components are done, but for now I'm fine with leaving it as the standard blue outline.

calebroseland:
> For any thread that only has the original author in it, we don't show any avatar and this feels a bit odd. I think maybe we should consider the root poster a participant if they reply to their own message, but only if they post a reply. Without the avatar, the footer looks misaligned
Screen Shot 2021-05-05 at 5 15 40 PM

@matthewbirtch 
Okay, in that case I'll revert the post-author avatar removal code. Performance wise, we won't be able to conditionally exclude based on the replies—we should reevaluate https://mattermost.atlassian.net/browse/MM-35017.

---

> ...what we should do when any element inside a post has focus? If so, let's give the post the same background as hover

@matthewbirtch @esethna
Okay, so `.post:focus-within` will match the UX of `.post:hover`—it will have the same background and will display the last-reply timestamp, as it does on hover. Is that correct?

When you click follow/unfollow, it stays focused and triggers `.post:focus-within`, showing the timestamp until you click somewhere else—this is what is causing the issue you guys are seeing, but we can't have it both ways on this.

calebroseland:
@jgilliam17 the avatar issue should be resolved now (no transparency and no squircles on `focus`)

calebroseland:
> When I click 'Follow' on a thread, the order of the participants seems to change?

I'm not sure what the current ordering criteria is, but yes it does differ between Post-participants and UserThread-participants—I think we should address this in MM-35017.

> I updated the spec for clarity: "Participants include any user who has replied to the thread excluding the original root post. The original author is only included in the participants list if they reply."

Understood 👍. I think we need to separate this effort, it's not feasible to do this conditionally client side. If we look at it holistically, there's still ordering to address (sort by most recent replies that was brought up by a community member). 

I kicked [MM-35017](https://mattermost.atlassian.net/browse/MM-35017) back to triage for now. I think we should use that to address
1) excluding the root post from scope of participants, and 
2) sort avatars by recency of replies.

jgilliam17:
@calebroseland I pair-tested few actions with Ogi and saw a crash
Steps:
User-1 is viewing channel1
User-2 posts a reply to a thread on channel2 that user-1 is following
User-1 refreshes (while viewing channel1)
Webapp crashes with the following error:
Screen Shot 2021-05-10 at 10 41 43 AM

matthewbirtch (approved):
One last thing I'm thinking we could/should do is remove the 5px of padding on the top of the thread footer to tighten it up with the root post it's related to.
Screen Shot 2021-05-10 at 11 34 13 AM

Knowing the other issues I identified will be addressed in follow-up PRs and I'll let the QA folks approve once their identified issues are fixed. So, I'm going to approve this as it's working well from a UX standpoint.

esethna:
@ogi-m @jgilliam17 can we prioritize reviews and iterations on these threading PRs so we have them in for our testing session later this week?
 
Where is makes sense and for non critical issues we can create follow-up tickets to help get this in faster.

calebroseland:
@matthewbirtch 5px removed ✓
note it will also affect posts with no reactions 
CleanShot 2021-05-10 at 17 07 25@2x

I thought there was a ticket open for it already, but I opened one for all the (combined) in-channel styling changes to avoid odd styling as things are rolled out. https://mattermost.atlassian.net/browse/MM-35602

calebroseland:
> @calebroseland I pair-tested few actions with Ogi and saw a crash
> Steps:
> User-1 is viewing channel1
> User-2 posts a reply to a thread on channel2 that user-1 is following
> User-1 refreshes (while viewing channel1)
> Webapp crashes with the following error:
> Screen Shot 2021-05-10 at 10 41 43 AM

@jgilliam17 I couldn't reproduce this locally (I think depends on which profiles are already loaded) but I added a fix for where this errors occurs and it _should_ be resolved

jgilliam17 (approved):
Thanks @calebroseland 
Tested, looks good to merge.
- Verified root post thread footer and reduced spacing above. Verified buttons to follow/unfollow, reply with count and user avatars, including transparent BG ones - as expected.

matthewbirtch:
> @matthewbirtch 5px removed ✓
> 
> note it will also affect posts with no reactions 

> I thought there was a ticket open for it already, but I opened one for all the (combined) in-channel styling changes to avoid odd styling as things are rolled out. https://mattermost.atlassian.net/browse/MM-35602, perhaps we can do all the finite styling there.

Awesome, thanks for this @calebroseland!

calebroseland:
@Willyfrog @jespino gentle ping; ready for dev review

koox00 (approved):
LGTM! +1 for synthetic threads :)

esethna:
Great, can someone merge this? @Willyfrog @koox00 @calebroseland

PR #8000, opened 2021-04-26
Making search result type selector independent from the search result filter in the search box

jespino:
#### Summary
Making search result type selector independent from the search result filter in the search box

#### Ticket Link
[MM-34834](https://mattermost.atlassian.net/browse/MM-34834)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/search_results/search_results.tsx
+++ b/components/search_results/search_results.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 /* eslint-disable react/no-string-refs */
 
-import React, {useEffect, useRef} from 'react';
+import React, {useEffect, useRef, useState} from 'react';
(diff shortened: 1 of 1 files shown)


koox00 (commented):
Should we also update the `onChange` on `MessageOrFileSelector` to call both the local function and the prop `setSearchType`?
It's not clear to me if there is a valid case for that though.

jespino:
The point is to have a local state set on new searches (when terms changes) but keep it local later on, for that reason we only use the local one on change, and not the global one.

calebroseland (approved):
LGTM!

koox00 (approved):

ogi-m (approved):
LGTM

PR #8001, opened 2021-04-26
MM-33540 Remove sticky headings from sidebar to fix nesting errors

hmhealey:
As in #7708, this fixes the nesting issues in the sidebar that caused some warning spam and likely caused navigation issues for accessibility tools, but to do that, we had to give up sticky category headings on browsers that support it. Once the menu components are rewritten to change how the menus themselves are nested, sticky headings can be re-added, but we can't have both at once due to some weird browser-specific interactions between `position: sticky`, z-indices, and overflows.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33540

#### Release Note
```release-note
Removed sticky sidebar headings in favor of fixing nesting errors
```

### DIFF
--- a/components/sidebar/sidebar_category_header.tsx
+++ b/components/sidebar/sidebar_category_header.tsx
@@ -6,7 +6,6 @@ import React from 'react';
 import {DraggableProvidedDragHandleProps} from 'react-beautiful-dnd';
 
 import {wrapEmojis} from 'utils/emoji_utils';
-import * as UserAgent from 'utils/user_agent';
 
 type StaticProps = {
     children?: React.ReactNode;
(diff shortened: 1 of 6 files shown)


devinbinnie:
There should also be some E2E tests to remove here that test sticky headers.

michelengelen (approved):
LGTM!

hmhealey:
@devinbinnie I've removed the one test I could find involving the sticky headings

devinbinnie (approved):

michelengelen (approved):
LGTM and nice catch @devinbinnie

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified sticky header for category label has been removed; position of the top and bottom More Unread toast on the LHS is as expected without sticky headers. 
- Verified LHS menus and DM + button no longer appearing over category menus.

PR #8011, opened 2021-04-28
[MM-34343] Migrate 'components/suggestion/search_channel_suggestion' and associated tests to TypeScript

arvinDarmawan:
Summary
This code migrate "components/suggestion/search_channel_suggestion" component to TypeScript.

Ticket Link
Jira card: https://mattermost.atlassian.net/browse/MM-34343
Github issue: https://github.com/mattermost/mattermost-server/issues/17280

```release-note
Release Note
- Refactor component to typescript, no changes in UI
```

### DIFF
--- a/components/suggestion/search_channel_suggestion/index.ts
+++ b/components/suggestion/search_channel_suggestion/index.ts
@@ -5,10 +5,17 @@ import {connect} from 'react-redux';
 import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
 
 import {getDirectTeammate} from 'utils/utils';
+import {GlobalState} from 'types/store';
 
 import SearchChannelSuggestion from './search_channel_suggestion';
 
-const mapStateToProps = (state, ownProps) => {
+type OwnProps = {
+    item: {
+        id: string;
+    };
+}
+
+const mapStateToProps = (state: GlobalState, ownProps: OwnProps) => {
     return {
         teammate: getDirectTeammate(state, ownProps.item.id),
         currentUser: getCurrentUserId(state),
--- a/components/suggestion/search_channel_suggestion/search_channel_suggestion.test.tsx
+++ b/components/suggestion/search_channel_suggestion/search_channel_suggestion.test.tsx
@@ -1,0 +1,89 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {shallow} from 'enzyme';
+
+import {TestHelper} from 'utils/test_helper';
+
+import SearchChannelSuggestion from './search_channel_suggestion';
+
+describe('components/suggestion/search_channel_suggestion', () => {
+    const mockChannel = TestHelper.getChannelMock();
+    const mockTeamMate = TestHelper.getUserMock();
+
+    const baseProps = {
+        item: mockChannel,
+        isSelection: false,
+        teammate: mockTeamMate,
+        currentUser: 'userid1',
+        term: '',
+        matchedPretext: '',
+    };
+
+    test('should match snapshot', () => {
+        const wrapper = shallow(
+            <SearchChannelSuggestion {...baseProps}/>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should match snapshot, isSelection is false', () => {
+        const props = {...baseProps, isSelection: false};
+        const wrapper = shallow(
+            <SearchChannelSuggestion {...props}/>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should match snapshot, isSelection is true', () => {
+        const props = {...baseProps, isSelection: true};
+        const wrapper = shallow(
+            <SearchChannelSuggestion {...props}/>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should match snapshot, channel type DM_CHANNEL', () => {
+        mockChannel.type = 'D';
+        const props = {...baseProps, item: mockChannel, isSelection: true};
+        const wrapper = shallow(
+            <SearchChannelSuggestion {...props}/>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should match snapshot, channel type GM_CHANNEL', () => {
+        mockChannel.type = 'G';
+        const props = {...baseProps, item: mockChannel, isSelection: true};
+        const wrapper = shallow(
+            <SearchChannelSuggestion {...props}/>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should match snapshot, channel type OPEN_CHANNEL', () => {
+        mockChannel.type = 'O';
+        const props = {...baseProps, item: mockChannel, isSelection: true};
+        const wrapper = shallow(
+            <SearchChannelSuggestion {...props}/>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should match snapshot, channel type PRIVATE_CHANNEL', () => {
+        mockChannel.type = 'P';
+        const props = {...baseProps, item: mockChannel, isSelection: true};
+        const wrapper = shallow(
+            <SearchChannelSuggestion {...props}/>,
+        );
+
+        expect(wrapper).toMatchSnapshot();
+    });
+});
--- a/components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
+++ b/components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
@@ -3,15 +3,13 @@
 import React from 'react';
 
 import {getUserIdFromChannelName} from 'mattermost-redux/utils/channel_utils';
-
 import {imageURLForUser} from 'utils/utils.jsx';
 import Constants from 'utils/constants';
 import Avatar from 'components/widgets/users/avatar';
 import BotBadge from 'components/widgets/badges/bot_badge';
+import Suggestion from '../suggestion';
 
-import Suggestion from '../suggestion.jsx';
-
-function itemToName(item, currentUser) {
+function itemToName(item: any, currentUser: string) {
     let itemMarkup = (item);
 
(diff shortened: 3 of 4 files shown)


arvinDarmawan:
Hi  @devinbinnie ,  I already create the pull request for this issue  but I set it to "Draft" because I have some questions for this task.

I check the previous code and I see that the class "SearchChannelSuggestion" is a child class from parent class "Suggestion".
Currently I only translate "SearchChannelSuggestion" to typescript but for Suggestion still in javascript language.

Should I change Suggestion class to typescript too ?
Because if I change it to typescript, I think there are several classes (ChannelSuggestion, EmoticonSuggestion, SearchSuggestionList, etc) that need to be checked again since it's related with 'Suggestion'.

Or should I convert it from Inheritance to Composition style for class "SearchChannelSuggestion" ??

components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
```diff
@@ -3,15 +3,13 @@
...
 import {imageURLForUser} from 'utils/utils.jsx';
 import Constants from 'utils/constants';
 import Avatar from 'components/widgets/users/avatar';
 import BotBadge from 'components/widgets/badges/bot_badge';
+import Suggestion from '../suggestion';
 
-import Suggestion from '../suggestion.jsx';
-
-function itemToName(item, currentUser) {
+function itemToName(item: any, currentUser: string) {
```
devinbinnie, line 12: `item` should be of type `Channel` here

components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
```diff
@@ -70,7 +68,9 @@ function itemToName(item, currentUser) {
 }
 
 export default class SearchChannelSuggestion extends Suggestion {
-    render() {
+    private node = React.createRef<HTMLDivElement>();
```
devinbinnie, line 71: This should be done in the constructor, not in the declared field.
Also if you're going to use this pattern for refs (which is the better pattern), anywhere the ref is used needs to be updated as well. Although I don't see any references to `node`, so maybe we don't even need the ref?

components/suggestion/search_channel_suggestion/index.ts
```diff
@@ -5,10 +5,17 @@ import {connect} from 'react-redux';
 import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
 
 import {getDirectTeammate} from 'utils/utils';
+import {GlobalState} from 'types/store';
 
 import SearchChannelSuggestion from './search_channel_suggestion';
 
-const mapStateToProps = (state, ownProps) => {
+type OwnProps = {
+    item: {
```
devinbinnie, line 13: I think `item` is of type `Channel` here as well.

components/suggestion/search_channel_suggestion/search_channel_suggestion.test.tsx
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
devinbinnie, line 1: Thank you for adding tests! <3

devinbinnie (commented):
Thanks for the PR @arvinDarmawan!

As for migrating `Suggestion`, I think we can leave it alone for now since there is another ticket out there for it, and it doesn't look like any type errors are coming up as a result.

components/suggestion/search_channel_suggestion/index.ts
arvinDarmawan: Hi @devinbinnie  I already add this.

components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
arvinDarmawan: Hi @devinbinnie  I already add this.

components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
arvinDarmawan: Yes, I already remove it.

components/suggestion/search_channel_suggestion/index.ts
```diff
@@ -2,16 +2,21 @@
...
-const mapStateToProps = (state, ownProps) => {
+type OwnProps = {
+    item: Channel;
+}
+
+const mapStateToProps = (state: GlobalState, ownProps: OwnProps) => {
     return {
         teammate: getDirectTeammate(state, ownProps.item.id),
-        currentUser: getCurrentUserId(state),
+        currentUser: getCurrentUser(state),
```
arvinDarmawan, line 19: Hi @devinbinnie , I change it from "getCurrentUserId" to "getCurrentUserId"  because previous jsx code using getCurrentUserId

components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
```diff
@@ -3,22 +3,23 @@
...
 import BotBadge from 'components/widgets/badges/bot_badge';
+import Suggestion from '../suggestion';
 
-import Suggestion from '../suggestion.jsx';
+import {UserProfile} from 'mattermost-redux/types/users';
+import {Channel} from 'mattermost-redux/types/channels';
 
-function itemToName(item, currentUser) {
-    let itemMarkup = (item);
+function itemToName(item: Channel, currentUser: UserProfile) {
```
arvinDarmawan, line 15: Hi @devinbinnie , I set the data type for param currentUser (previously string in jsx file) to be  "UserProfile".

components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
```diff
@@ -3,22 +3,23 @@
...
+    let itemMarkup;
 
     if (item.type === Constants.DM_CHANNEL) {
         const profilePicture = (
             <Avatar
-                username={item.username}
-                url={imageURLForUser(getUserIdFromChannelName(currentUser, item.name), item.last_picture_update)}
+                username={currentUser.username}
```
devinbinnie, line 21: I'm not sure why we're changing the logic here, can you elaborate as to why?

components/suggestion/search_channel_suggestion/index.ts
devinbinnie: Is this something we need to change?

components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
arvinDarmawan: Hi @devinbinnie , since object  item has data type "Channel" , it doesn't have property username.
Username in tag "Avatar" should refer o name of user, right ? 

So I change the code from `username={item.username}` to be  `username={currentUser.username}`.

components/suggestion/search_channel_suggestion/index.ts
arvinDarmawan: Hi @devinbinnie , I change it so the currentUser data type will be UserProfile (previously string).

By changing the data type of currentUser to be "UserProfile",  in file "components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx", I can render html tag Avatar to become like this : 
`<Avatar
          username={currentUser.username}
          url={imageURLForUser(getUserIdFromChannelName(currentUser.id, item.name), currentUser.last_picture_update)}
          size='sm'
 />`

Previous code in jsx file like this : 
`<Avatar
          username={item.username}
          url={imageURLForUser(getUserIdFromChannelName(currentUser, item.name), item.last_picture_update)}
          size='sm'
 />`

As you can see in previous code, `item.username` and `item.last_picture_update` will not working if we think item as "Channel" since username and last_picture_update are UserProfile properties.

components/suggestion/search_channel_suggestion/search_channel_suggestion.tsx
devinbinnie: Hmm, well that would create a logic change which is not what we want. I think `item` is actually of type `UserProfile` as well since that would make sense with the fields that are used.

devinbinnie: Did some further investigating, it looks like we only get `Channel` objects there and they don't have the `last_picture_update` field, so it will always be undefined. Maybe we just take out the fields that are missing and leave the rest of the logic the way it is?

Sorry for the confusion.

arvinDarmawan: Hi @devinbinnie ,so I will revert the **currentUser to be string again** and the logic of code will be like this ?

 ```
 function itemToName(item: Channel, currentUser: string) {
    let itemMarkup;


    if (item.type === Constants.DM_CHANNEL) {
        const profilePicture = (
            <Avatar
               // delete username={item.username} in <Avatar> since Channel.username is always undefined
                url={imageURLForUser(getUserIdFromChannelName(currentUser, item.name))} // will not set last_picture_update 
                size='sm'
            />
        );

        itemMarkup = (
            <React.Fragment>
                {profilePicture}
                <span className='ml-3'>{'@'}{item.display_name}</span>
            </React.Fragment>
        );
    }
    ....... other code ......
  }
````
  
and in file components/suggestion/search_channel_suggestion/index.ts, i will revert back 
`currentUser: getCurrentUser(state)` (UserProfile)  to be `currentUser: getCurrentUserId(state)` (string)  ?

devinbinnie: Yes I believe that's correct. Thanks for the hard work :)

arvinDarmawan: Hi @devinbinnie , I already change it. Thank you

devinbinnie (approved):
Thanks @arvinDarmawan, this is looking good now :)

hmhealey (approved):
Looks good to me. Migrating Suggestion and fixing the inheritance problem would definitely be helpful, but I don't think it needs to be done in this PR.

jgilliam17 (approved):
Thank you @arvinDarmawan 
Tested, looks good to merge.
- Verified search channel autocomplete - as before.
@hmhealey Can you please help merge? Thanks.

PR #8019, opened 2021-04-28
MM-22678 Update Edit Channel Header modal error text

hmhealey:
As in the related PR, this updates the error text when a user enters too many characters into the box.

I also updated the code in the Edit Channel Purpose modal, but it wasn't strictly needed because that input has a maxLength set. I would've looked at making the two behave the same as each other, but that seems out of the scope for this ticket. I did however clean up some weird looking login in the Edit Channel Purpose modal that seemed unnecessary.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22678

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/17520

#### Release Note
```release-note
Improve error text in Edit Channel Header modal.
```


components/edit_channel_purpose_modal/edit_channel_purpose_modal.jsx
```diff
@@ -46,30 +48,13 @@ class EditChannelPurposeModal extends React.PureComponent {
...
-            serverError: '',
+            serverError: null,
             show: true,
             submitted: false,
             requestStarted: false,
         };
     }
 
-    setError = (err) => {
-        if (err.id === 'api.context.invalid_param.app_error') {
```
hmhealey, line 57: I didn't go back far enough to find the origin of this part of the code, but this wasn't even the right error ID that would be returned by the server if the user's input was too long, so I just decided to get rid of it entirely.

I also changed the type of `this.state.serverError` to be an object like most other places in the web app.

Willyfrog (approved):
LGTM!

jespino (approved):
LGTM

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified error message on Edit Header modal.
I agree, it might make sense to make both header and purpose modals behave the same. I'll reach out to @matthewbirtch and see if this is something we would consider improving further.

hmhealey:
Ignoring the git-new CI check since I think it's broken for everything right now.

PR #8021, opened 2021-04-29
MM-34460 Added framework for having admin console settings in sections

harshilsharma63:
#### Summary
Added framework for having admin console settings in sections and refactored AD/LDAP config for use this framework.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34460

#### Screenshots
https://i.imgur.com/1oyjcF4.png

#### Release Note
Added visual grouping of related settings under AD/LDAP in Admin Console.


michaelgamble (changes requested):
The only problem I'm seeing is that in the design file only the first section and last section have subtitles, in your uploaded image, the subtitle of the first section is repeated as a placeholder in a few of the other sections...

I'd say for this release, remove all subtitles except the first and last section.

harshilsharma63:
> The only problem I'm seeing is that in the design file only the first section and last section have subtitles, in your uploaded image, the subtitle of the first section is repeated as a placeholder in a few of the other sections...
> 
> I'd say for this release, remove all subtitles except the first and last section.

Yeah, that was a typo. I'll update them.

harshilsharma63:
@michaelgamble I've fixed the incorrect subtitles now. I've also updated the screenshot in PR so you can take a look there.

michaelgamble (approved):
LGTM

sbishel (commented):
I think the section should be handled as any other type of "setting". I don't really like having the schema to contain either  `sections` or `settings` as a child. And actually, if I follow correctly, you can't have both. (ie, some fields in sections, others not.) 
I'm not sure if you ran into a technical issue, but would rather see `settings` as the container for everything in the schema

You may have gone down that path already and there is a reason it won't work.

harshilsharma63:
@sbishel

Yes, we cannot have both `settings` and `sections` together, i.e., a system console page will either have to have everything in sections or nothing. But the sections title and subtitles are purely optional if it helps.

The idea of `settings` and `sections` was to keep the changes isolated and to avoid `sections` from affecting anywhere we don't use it. It also made the different styling of `sections` vs non-sections part much easier and cleaner. 

We need to know what group of settings constitute a section and this was the simplest idea I came up with. I'm open to any other ideas.

mkraft (approved):

sbishel (approved):
Thanks for explanation and walking me through it.

furqanmlk (commented):
thanks @harshilsharma2 

I ran Cypress tests, all passed.
However I found some UI mismatched with Figma 

image

image

image

image

image

image

image

image

harshilsharma63:
@furqanmlk My bad, I didn't mention this in our conversation that we're not implementing those buttons right now. Those would need API changes and much more time, so we simply did a cosmetic facelift of the AD/LDAP section.

For Canceled vs Cancelled, I see that in our [codebase](https://github.com/mattermost/mattermost-webapp/blob/7104c8afac692542064a2aa0f149e511728b1e5a/i18n/en.json#L1102) we actually use `Canceled` instead of `Cancelled`.

furqanmlk:
> @furqanmlk My bad, I didn't mention this in our conversation that we're not implementing those buttons right now. Those would need API changes and much more time, so we simply did a cosmetic facelift of the AD/LDAP section.
> 
> For Canceled vs Cancelled, I see that in our [codebase](https://github.com/mattermost/mattermost-webapp/blob/7104c8afac692542064a2aa0f149e511728b1e5a/i18n/en.json#L1102) we actually use `Canceled` instead of `Cancelled`.

Thanks @harshilsharma63 no problem.
All look good , just last question, can you have a look at the first issue, "the contact card icon is missing in AD/LDAP field"

harshilsharma63:
@furqanmlk It's not a Mattermost thing. I believe the icon is coming from the browser's auto-fill feature or from LastPass.

I get this from LastPass and Safari -

Screenshot 2021-05-19 at 3 19 54 PM

Screenshot 2021-05-19 at 3 27 50 PM

furqanmlk (approved):
Thanks @harshilsharma63 

All look good, tests passed

PR #8041, opened 2021-05-02
[GH_ 17286] Migrate 'components/claim/components' and associated tests to TypeScript

vitowidigdo:
#### Summary

Migrate components/claim/components to TypeScript
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/17286

### DIFF
--- a/components/claim/components/email_to_ldap.tsx
+++ b/components/claim/components/email_to_ldap.tsx
@@ -1,8 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import PropTypes from 'prop-types';
-import React from 'react';
+import React,{ReactNode} from 'react';
 import {FormattedMessage} from 'react-intl';
 
 import {emailToLdap} from 'actions/admin_actions.jsx';
@@ -11,81 +10,95 @@ import {t} from 'utils/i18n.jsx';
 import LoginMfa from 'components/login/login_mfa.jsx';
 import LocalizedInput from 'components/localized_input/localized_input';
 
-export default class EmailToLDAP extends React.PureComponent {
-    static propTypes = {
-        email: PropTypes.string,
-        siteName: PropTypes.string,
-        ldapLoginFieldName: PropTypes.string,
-    };
+type Props = {
+    email?: any,
+    siteName?: any,
+    ldapLoginFieldName?: any,
+};
 
-    constructor(props) {
+
+type State = {
+    passwordError?: string,
+    ldapError?: string,
+    ldapPasswordError?: string,
+    serverError?: string,
+    showMfa?: boolean,
+};
+
+export default class EmailToLDAP extends React.PureComponent <Props, State>{
+    private emailPasswordInput: React.RefObject<HTMLInputElement>;
+    private ldapIdInput: React.RefObject<HTMLInputElement>;
+    private ldapPasswordInput: React.RefObject<HTMLInputElement>;
+
+    public constructor(props: Props) {
         super(props);
-
-        this.state = {
-            passwordError: '',
-            ldapError: '',
-            ldapPasswordError: '',
-            serverError: '',
-            showMfa: false,
-        };
 
         this.emailPasswordInput = React.createRef();
         this.ldapIdInput = React.createRef();
         this.ldapPasswordInput = React.createRef();
+        this.state = this.getDefaultState();
     }
 
-    preSubmit = (e) => {
-        e.preventDefault();
-
-        var state = {
+    getDefaultState() {
+        return {
             passwordError: '',
             ldapError: '',
             ldapPasswordError: '',
             serverError: '',
+            ldapId: '',
+            password: '',
+            ldapPassword: '',
         };
+    }
 
+    public preSubmit = (e: React.SyntheticEvent): void => {
+        e.preventDefault();
+        
+        const defaultState = this.getDefaultState();
+        
         const password = this.emailPasswordInput.current && this.emailPasswordInput.current.value;
         if (!password) {
-            state.passwordError = Utils.localizeMessage('claim.email_to_ldap.pwdError', 'Please enter your password.');
-            this.setState(state);
+            defaultState.passwordError = Utils.localizeMessage('claim.email_to_ldap.pwdError', 'Please enter your password.');
+            this.setState(defaultState);
             return;
         }
 
         const ldapId = this.ldapIdInput.current && this.ldapIdInput.current.value.trim();
         if (!ldapId) {
-            state.ldapError = Utils.localizeMessage('claim.email_to_ldap.ldapIdError', 'Please enter your AD/LDAP ID.');
-            this.setState(state);
+            defaultState.ldapError = Utils.localizeMessage('claim.email_to_ldap.ldapIdError', 'Please enter your AD/LDAP ID.');
+            this.setState(defaultState);
             return;
         }
 
         const ldapPassword = this.ldapPasswordInput.current && this.ldapPasswordInput.current.value;
         if (!ldapPassword) {
-            state.ldapPasswordError = Utils.localizeMessage('claim.email_to_ldap.ldapPasswordError', 'Please enter your AD/LDAP password.');
-            this.setState(state);
+            defaultState.ldapPasswordError = Utils.localizeMessage('claim.email_to_ldap.ldapPasswordError', 'Please enter your AD/LDAP password.');
+            this.setState(defaultState);
             return;
         }
 
-        state.password = password;
-        state.ldapId = ldapId;
-        state.ldapPassword = ldapPassword;
-        this.setState(state);
+        defaultState.password = password;
+        defaultState.ldapId = ldapId;
+        defaultState.ldapPassword = ldapPassword;
+        this.setState(defaultState);
 
         this.submit(this.props.email, password, '', ldapId, ldapPassword);
     }
 
-    submit = (loginId, password, token, ldapId, ldapPassword) => {
+    public submit = (loginId: string, password: string, token: string, ldapId: string, ldapPassword: string): void => {
+        const defaultState = this.getDefaultState();
         emailToLdap(
             loginId,
             password,
             token,
-            ldapId || this.state.ldapId,
-            ldapPassword || this.state.ldapPassword,
-            (data) => {
+            ldapId || defaultState.ldapId,
+            ldapPassword || defaultState.ldapPassword,
+            (data: { follow_link: string; }) => {
                 if (data.follow_link) {
                     window.location.href = data.follow_link;
                 }
             },
-            (err) => {
+            (err: { server_error_id: string; id: any; message: any; }) => {
                 if (!this.state.showMfa && err.server_error_id === 'mfa.validate_token.authenticate.app_error') {
                     this.setState({showMfa: true});
                 } else {
@@ -109,7 +122,7 @@ export default class EmailToLDAP extends React.PureComponent {
         );
     }
 
-    render() {
+    public render(): ReactNode {
         let serverError = null;
         let formClass = 'form-group';
         if (this.state.serverError) {
@@ -146,11 +159,12 @@ export default class EmailToLDAP extends React.PureComponent {
         }
 
         let content;
+        var defaultState = this.getDefaultState()
         if (this.state.showMfa) {
             content = (
                 <LoginMfa
                     loginId={this.props.email}
-                    password={this.state.password}
+                    password={defaultState.password}
                     submit={this.submit}
                 />
             );
--- a/components/claim/components/email_to_ldap.tsx
+++ b/components/claim/components/email_to_ldap.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React,{ReactNode} from 'react';
+import React, {ReactNode} from 'react';
 import {FormattedMessage} from 'react-intl';
 
 import {emailToLdap} from 'actions/admin_actions.jsx';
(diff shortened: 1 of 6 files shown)


components/claim/components/email_to_ldap.tsx
```diff
@@ -11,81 +10,95 @@ import {t} from 'utils/i18n.jsx';
...
-export default class EmailToLDAP extends React.PureComponent {
-    static propTypes = {
-        email: PropTypes.string,
-        siteName: PropTypes.string,
-        ldapLoginFieldName: PropTypes.string,
-    };
+type Props = {
+    email?: any,
+    siteName?: any,
+    ldapLoginFieldName?: any,
```
Willyfrog, line 16: why not use string on these?

components/claim/components/email_to_ldap.tsx
```diff
@@ -11,81 +10,95 @@ import {t} from 'utils/i18n.jsx';
...
         this.ldapIdInput = React.createRef();
         this.ldapPasswordInput = React.createRef();
+        this.state = this.getDefaultState();
     }
 
-    preSubmit = (e) => {
-        e.preventDefault();
-
-        var state = {
+    getDefaultState() {
```
Willyfrog, line 42: this should return type state, doesn't it?

components/claim/components/email_to_ldap.tsx
```diff
@@ -11,81 +10,95 @@ import {t} from 'utils/i18n.jsx';
...
+        return {
             passwordError: '',
             ldapError: '',
             ldapPasswordError: '',
             serverError: '',
+            ldapId: '',
+            password: '',
+            ldapPassword: '',
```
Willyfrog, line 50: missing `showMfa`, I see some other fields not present in type state.

if we are initializing to the empty string, why did we add `?` to the types?

components/claim/components/email_to_ldap.tsx
```diff
@@ -146,11 +159,12 @@ export default class EmailToLDAP extends React.PureComponent {
         }
 
         let content;
+        var defaultState = this.getDefaultState()
         if (this.state.showMfa) {
             content = (
                 <LoginMfa
                     loginId={this.props.email}
-                    password={this.state.password}
+                    password={defaultState.password}
```
Willyfrog, line 167: why storing password in the state (there is even an `ldappassword` field in it if we are always passing `''`?

components/claim/components/email_to_oauth.tsx
```diff
@@ -11,27 +10,41 @@ import * as Utils from 'utils/utils.jsx';
 import {t} from 'utils/i18n.jsx';
 import LoginMfa from 'components/login/login_mfa.jsx';
 import LocalizedInput from 'components/localized_input/localized_input';
+import { boolean } from '@storybook/addon-knobs';
```
Willyfrog, line 13: was this imported automatically by the IDE?

components/claim/components/email_to_oauth.tsx
```diff
@@ -11,27 +10,41 @@ import * as Utils from 'utils/utils.jsx';
...
-export default class EmailToOAuth extends React.PureComponent {
-    static propTypes = {
-        newType: PropTypes.string,
-        email: PropTypes.string,
-        siteName: PropTypes.string,
-    };
+type Props = {
+    newType?: any,
+    email?: any,
+    siteName?: string,
```
Willyfrog, line 18: do we really need any?

this happens across the PR, can you take a look at the others as well?

components/claim/components/email_to_oauth.tsx
```diff
@@ -40,24 +53,24 @@ export default class EmailToOAuth extends React.PureComponent {
 
         this.setState({password});
 
-        state.error = null;
+        // state.error = null;
```
Willyfrog, line 56: there is no need to leave any commented code

Willyfrog (changes requested):
hello and thanks for tackling the issuel,

can you sign the CLA, let me know if you already did.
I already added some comments, but you should also take a look at the errors from CI

Willyfrog:
hey @vitowidigdo , let me know if you need clarification on the feedback or if I can be of any help

components/claim/components/email_to_ldap.tsx
vitowidigdo: already change this to string

components/claim/components/email_to_ldap.tsx
vitowidigdo: I don't quite understand this I'm sorry

components/claim/components/email_to_ldap.tsx
vitowidigdo: I'm correcting this

components/claim/components/email_to_ldap.tsx
vitowidigdo: so just using this.state.password not referring to defaultState?

components/claim/components/email_to_oauth.tsx
vitowidigdo: OK I will delete this also across PR

components/claim/components/email_to_oauth.tsx
vitowidigdo: change it to string too like the above that you commented?

vitowidigdo:
> hey @vitowidigdo , let me know if you need clarification on the feedback or if I can be of any help

Hey thank you for your help, I already reply to your comments. Waiting for your answers thanks!

components/claim/components/email_to_oauth.tsx
Willyfrog: It is either `string` or whatever the right type is. `any` should probably just be reserved to when we really don't know what will be for the variable, otherwise we are not getting any benefits from typing.

components/claim/components/email_to_ldap.tsx
Willyfrog: nevermind this, I think I misread something

components/claim/components/email_to_ldap.tsx
Willyfrog: by not definining the return type, it might be any, so you are returning some object with properties that might or not match what it is expected by the caller. having a type for it will ensure that the compiler can actually help you figuring out if you forgot to add something in the code.

vitowidigdo: So I'm returning type state like this?

`getDefaultState() {
        type State = {
            passwordError: '',
            ldapError: '',
            ldapPasswordError: '',
            serverError: '',
            ldapId: '',
            password: '',
            ldapPassword: '',
            showMfa: false
        };
    }`

components/claim/components/email_to_oauth.tsx
vitowidigdo: Yes and I'm already deleting it

components/claim/components/email_to_ldap.tsx
vitowidigdo: So I need to list all other fields like `showMfa` and also deleting `?` in type state like this?
Screen Shot 2021-06-23 at 11 42 04

components/claim/components/email_to_ldap.tsx
vitowidigdo: If above is correct, how do I access the state fields?
Screen Shot 2021-06-23 at 12 51 09

Willyfrog: it doesn't change how you access them, it just tells the compiler "this is the object I'll return, check that I didn't forget to add any properties or added any unexpected ones"

components/claim/components/email_to_ldap.tsx
Willyfrog: I asked about '?' because I'm not sure it can be null, if it can't we shouldn't allow it, so the compiler knows about it.

components/claim/components/email_to_ldap.tsx
vitowidigdo: Ok so type state as I capture above is correct?

components/claim/components/email_to_ldap.tsx
vitowidigdo: Yes I agree with you the type State cannot be null, but their defaultState is null

components/claim/components/email_to_ldap.tsx
Willyfrog: yes, but you have to use it for that error to go away

vitowidigdo: so how do I use it for the error to go away? I'm confuse about how to get State object inside the function

Willyfrog: You should define it outside of the function and add it as the return type.
```ts
type StateType = {...}

getDefaultState(): StateType {
return ....
}
```

components/claim/components/email_to_ldap.tsx
```diff
@@ -11,81 +10,103 @@ import {t} from 'utils/i18n.jsx';
...
             passwordError: '',
             ldapError: '',
             ldapPasswordError: '',
             serverError: '',
+            password: '',
+            ldapId: '',
+            ldapPassword: '',
         };
 
+        const defaultState = state;
```
hmhealey, line 66: Instead of creating a separate `defaultState`, couldn't you just use `state` below?

components/claim/components/email_to_ldap.tsx
```diff
@@ -11,81 +10,103 @@ import {t} from 'utils/i18n.jsx';
...
+    public submit = (loginId: string, password: string, token: string, ldapId: string, ldapPassword: string): void => {
+        const defaultState = this.state;
         emailToLdap(
             loginId,
             password,
             token,
-            ldapId || this.state.ldapId,
-            ldapPassword || this.state.ldapPassword,
-            (data) => {
+            ldapId || defaultState.ldapId,
```
hmhealey, line 102: Same here

components/claim/components/email_to_ldap.tsx
```diff
@@ -11,81 +10,103 @@ import {t} from 'utils/i18n.jsx';
...
+            ldapPassword || defaultState.ldapPassword,
+            (data: { follow_link: string }) => {
                 if (data.follow_link) {
                     window.location.href = data.follow_link;
                 }
             },
-            (err) => {
+            (err: { server_error_id: string; id: any; message: any }) => {
```
hmhealey, line 109: `id` and `message` are both strings

components/claim/components/email_to_ldap.tsx
```diff
@@ -146,11 +167,12 @@ export default class EmailToLDAP extends React.PureComponent {
         }
 
         let content;
+        const defaultState = this.state;
```
hmhealey, line 170: Same question here about `defaultState`

components/claim/components/email_to_oauth.tsx
```diff
@@ -12,26 +11,34 @@ import {t} from 'utils/i18n.jsx';
...
-export default class EmailToOAuth extends React.PureComponent {
-    static propTypes = {
-        newType: PropTypes.string,
-        email: PropTypes.string,
-        siteName: PropTypes.string,
-    };
+export type State = {
+    showMfa?: boolean;
+    password?: string;
+    error?: string;
```
hmhealey, line 17: Could these values be made non-nullable if we specify values for them in the constructor? Since they all have default zero values like the empty string or false, I think we can keep things simpler by using those instead of null/undefined.

components/claim/components/email_to_oauth.tsx
```diff
@@ -12,26 +11,34 @@ import {t} from 'utils/i18n.jsx';
...
-        super(props);
+type Props = {
+    email: string;
+    newType: string;
+    siteName: string;
+}
+export default class EmailToOAuth extends React.PureComponent <Props, State> {
+    private passwordInput: React.RefObject<HTMLInputElement>;
 
-        this.state = {showMfa: false, password: ''};
```
hmhealey, line 25: Maybe I'm wrong, but don't we need to set `this.state` in the constructor? Maybe something changed with React in a newer version, but I thought we had to set it to at least an empty object if we're using it in a component.

components/claim/components/email_to_oauth.tsx
```diff
@@ -12,26 +11,34 @@ import {t} from 'utils/i18n.jsx';
...
+    preSubmit = (e: React.SyntheticEvent) => {
         e.preventDefault();
-        var state = {};
 
-        var password = this.passwordInput.current.value;
+        const state = {
+            error: '',
+        };
+
+        const password = this.passwordInput.current && this.passwordInput.current.value;
```
hmhealey, line 41: This pattern can be simplified using the `?.` operator to
```javascript
const password = this.passwordInput.current?.value;
```
wherever we're using it in these components

components/claim/components/email_to_oauth.tsx
```diff
@@ -40,24 +47,23 @@ export default class EmailToOAuth extends React.PureComponent {
...
             token,
             this.props.newType,
-            (data) => {
+            (data: { follow_link: string }) => {
                 if (data.follow_link) {
                     window.location.href = data.follow_link;
                 }
             },
-            (err) => {
+            (err: { server_error_id: string; message: any }) => {
```
hmhealey, line 66: `message` can also be a string here

components/claim/components/ldap_to_email.tsx
```diff
@@ -64,36 +69,37 @@ export default class LDAPToEmail extends React.PureComponent {
...
 
         this.submit(this.props.email, password, '', ldapPassword);
     }
 
-    submit = (loginId, password, token, ldapPassword) => {
-        this.props.switchLdapToEmail(ldapPassword || this.state.ldapPassword, this.props.email, password, token).then(({data, error: err}) => {
-            if (data && data.follow_link) {
+    submit = async (loginId: string, password: string, token: string, ldapPassword: string) => {
+        const res = await this.props.switchLdapToEmail(ldapPassword, this.props.email, password, token);
+        if ('data' in res) {
```
hmhealey, line 86: This is a slight change in behaviour from the previous code since it'll now skip the else-if statement below if `res` looks like `{data: {}, err: {server_error_id: 'something'}}`.

That said, I don't think that should actually cause problems here since we almost always just return just one of `res.data` or `res.error`, but it's good to be aware of when we might be changing things accidentally.

hmhealey (changes requested):
I'm not sure if this is ready for review again, but it seems like it should be. Let me know if you have any questions about these.

components/claim/components/email_to_oauth.tsx
vitowidigdo: How do I change the value to empty string or false?

components/claim/components/email_to_oauth.tsx
vitowidigdo: Ok my mistake for this I didn't set state object. But making state object give me this kinds of error. How do I resolve it?
Screen Shot 2021-07-26 at 11 14 12
Screen Shot 2021-07-26 at 11 13 58
Screen Shot 2021-07-26 at 11 13 46
Screen Shot 2021-07-26 at 11 13 36

components/claim/components/ldap_to_email.tsx
vitowidigdo: So for this line I need to delete if statement?

components/claim/components/email_to_oauth.tsx
Willyfrog: you'll need to initialize it with a value like `error=''` so that it at least contains the required type

components/claim/components/email_to_oauth.tsx
Willyfrog: without seeing the new code you are adding, I'd say you probably need to call `super(props)` first thing in the constructor

components/claim/components/ldap_to_email.tsx
hmhealey: It should be the same if you combine the outer and inner if statements so it's like `if ('data' in res && res.data.follow_link)` again

M-ZubairAhmed:
completed in https://github.com/mattermost/mattermost-webapp/pull/10408

PR #8043, opened 2021-05-02
[MM-34353] Migrate 'components/status_dropdown' and associated tests to TypeScript

arvinDarmawan:
Summary
This code migrate "components/status_dropdown" component to TypeScript.

Ticket Link
Jira card: https://mattermost.atlassian.net/browse/MM-34353
Github issue: https://github.com/mattermost/mattermost-server/issues/17283

```release-note
Release Note
- Refactor component to typescript, no changes in UI
```


components/status_dropdown/status_dropdown.tsx
```diff
@@ -24,76 +23,81 @@ import StatusOfflineIcon from 'components/widgets/icons/status_offline_icon';
...
+    isUserOutOfOffice = (): boolean => {
         return this.props.status === UserStatuses.OUT_OF_OFFICE;
     }
 
-    setStatus = (status) => {
+    setStatus = (status: string): void => {
         this.props.actions.setStatus({
             user_id: this.props.userId,
             status,
+            manual: false,
```
arvinDarmawan, line 65: hi @lieut-data is this code line already ok or better I set UserStatus to be like this ?
```
// from mattermost-redux/types/users
export type UserStatus = {
    user_id: string;
    status: string;
    manual?: boolean;
    last_activity_at?: number;
    active_channel?: string;
};

```

components/status_dropdown/index.ts
```diff
@@ -13,18 +13,17 @@ import {get} from 'mattermost-redux/selectors/entities/preferences';
...
+import {GlobalState} from 'types/store';
 
-function mapStateToProps(state) {
+import StatusDropdown from './status_dropdown';
+
+function mapStateToProps(state: GlobalState) {
     const currentUser = getCurrentUser(state);
     const getCustomStatus = makeGetCustomStatus();
 
-    if (!currentUser) {
```
lieut-data, line 24: Can you elaborate on removing this check? If there is no current user -- as unlikely as that may be, it can sometimes trigger during a logout, for example -- will the subsequent checks spuriously fail?

components/status_dropdown/status_dropdown.tsx
lieut-data: I /think/ this is ok, but will defer to @devinbinnie.

components/status_dropdown/status_dropdown.tsx
```diff
@@ -24,76 +23,81 @@ import StatusOfflineIcon from 'components/widgets/icons/status_offline_icon';
...
+    setDnd = (event: Event): void => {
         event.preventDefault();
         this.setStatus(UserStatuses.DND);
     }
 
-    showStatusChangeConfirmation = (status) => {
+    showStatusChangeConfirmation = (status: string): void => {
         const resetStatusModalData = {
-            ModalId: ModalIdentifiers.RESET_STATUS,
+            modalId: ModalIdentifiers.RESET_STATUS,
```
lieut-data, line 92: Is the rename of `ModalId` to `modalId` supported by the code that expects this action?

components/status_dropdown/status_dropdown.tsx
```diff
@@ -105,13 +109,13 @@ export default class StatusDropdown extends React.PureComponent {
...
 
-    renderDropdownIcon = () => {
+    renderDropdownIcon = (): ReactNode => {
         return (
             <FormattedMessage
                 id='generic_icons.dropdown'
                 defaultMessage='Dropdown Icon'
             >
-                { (title) => (
+                { (title: any) => (
```
lieut-data, line 118: Is it feasible to pick a narrower type here?

components/status_dropdown/status_dropdown.tsx
```diff
@@ -212,7 +217,6 @@ export default class StatusDropdown extends React.PureComponent {
         return (
             <MenuWrapper
                 onToggle={this.onToggle}
-                style={this.props.style}
```
lieut-data, line 215: To confirm, does someone else pass this now?

lieut-data (changes requested):
Thanks, @arvinDarmawan! A few comments -- I'll also loop in @devinbinnie as the domain expert here :)

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: Hi @lieut-data , I remove it because I didnt find componenet MenuWrapper  use props "style".
In component "components/sidebar/sidebar_header/sidebar_header.tsx", component SidebarHeader call component StatusDropdown without sending any props.

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: Hi @lieut-data  , I change it because after I check it on "actions/view/modals.js" the code is like :

```
import {ActionTypes} from 'utils/constants';

export function openModal(modalData) {
    return (dispatch) => {
        const action = {
            type: ActionTypes.MODAL_OPEN,
            modalId: modalData.modalId,
            dialogProps: modalData.dialogProps,
            dialogType: modalData.dialogType,
        };

        dispatch(action);
        return {data: true};
    };
}

// another function closeModal
```

As you can see,  in action. modalId, it receive modalData.modalId, not modalData.ModalId

components/status_dropdown/index.ts
arvinDarmawan: Hi @lieut-data , I remove the code to clean the typescript error. 

If I add code checking currentUser:
`if (!currentUser) { 
      return {};
  }
`

the data type of userId, profilePicture, autoResetPref, customStatus, and etc in components/status_dropdown/index.ts will be  **their own data type(string/bool/UserCustomStatus)**  or **undefined**. If the type can be **undefined**, I need to set the  type Props in components/status_dropdown/status_dropdown.tsx to allow undefined.

To resolve this issue, I check another  example of mattermost-webapp typescript code, and there is no logic checking if currentUser exists in file "index.ts". So, I remove the checking in status_dropdown.ts file and test it manually in my local.

But, if I check the javascript code (jsx),  there are several code like this (checking currentUser).
What do you think about this ?  



Screen Shot 2021-05-03 at 19 59 39

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: hi @lieut-data , I already change data type to "string".

lieut-data (approved):
Thanks, @arvinDarmawan!

components/status_dropdown/index.ts
devinbinnie: I think this is okay, but to be safe I think we should add coalescing operators in front of `currentUser` in this connector so that we never end up in a situation where the app will whitescreen if `currentUser` is `undefined`

components/status_dropdown/status_dropdown.tsx
```diff
@@ -24,76 +23,81 @@ import StatusOfflineIcon from 'components/widgets/icons/status_offline_icon';
...
+    userId: string;
+    profilePicture: string;
+    autoResetPref?: string;
+    actions: {
+        openModal: (modalData: {modalId: string; dialogType: any; dialogProps?: any}) => void;
+        setStatus: (status: UserStatus) => ActionFunc;
+        unsetCustomStatus: () => ActionFunc;
+        setStatusDropdown: (open: boolean) => void;
+    };
+    customStatus: UserCustomStatus | Record<string, any>;
```
devinbinnie, line 43: How come we need to union with `Record<string, any>` here?

components/status_dropdown/status_dropdown.tsx
devinbinnie: I think it's better to fix the type if the function doesn't expect all of the parameters to be there.

components/status_dropdown/index.ts
```diff
@@ -13,22 +13,21 @@ import {get} from 'mattermost-redux/selectors/entities/preferences';
...
-        return {};
-    }
-
-    const userId = currentUser.id;
+    const userId = currentUser?.id;
```
arvinDarmawan, line 27: Hi @devinbinnie  I already add coalescing operator here.

components/status_dropdown/index.ts
```diff
@@ -13,22 +13,21 @@ import {get} from 'mattermost-redux/selectors/entities/preferences';
...
     return {
         userId,
-        profilePicture: Client4.getProfilePictureUrl(userId, currentUser.last_picture_update),
+        profilePicture: Client4.getProfilePictureUrl(userId, currentUser?.last_picture_update),
```
arvinDarmawan, line 30: Hi @devinbinnie  I already add coalescing operator here.

components/status_dropdown/index.ts
arvinDarmawan: Hi @devinbinnie  I already add coalescing operator  on several lines

components/status_dropdown/status_dropdown.tsx
devinbinnie: Yeah it doesn't look like we use this anywhere. Definitely removeable.

devinbinnie (commented):
Thanks for the PR @arvinDarmawan!
I've added a couple comments for review.

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: Hi @devinbinnie, it is because the defaultProps of customStatus in components/status_dropdown/status_dropdown.tsx is empty object.

Screen Shot 2021-05-03 at 22 35 49

devinbinnie: Can we try casting it as `UserCustomStatus`?

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: Hi @devinbinnie I already alter the type UserStatus.

components/status_dropdown/status_dropdown.tsx
```diff
@@ -120,32 +125,33 @@ export default class StatusDropdown extends React.PureComponent {
...
+    }
 
-    renderCustomStatus = () => {
+    renderCustomStatus = (): ReactNode => {
         if (!this.props.isCustomStatusEnabled) {
             return null;
         }
         const customStatus = this.props.customStatus;
-        const isStatusSet = customStatus && (customStatus.text || customStatus.emoji);
+        const isStatusSet = customStatus && (customStatus.text.trim().length !== 0 || customStatus.emoji.trim().length !== 0);
```
devinbinnie, line 143: Is it possible that the strings will contain only blank spaces?
I'm just wary of changing logic here unnecessarily.

If it's a type issue you're running into, you can cast strings as a Boolean here.

arvinDarmawan: Done @devinbinnie

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: Done @devinbinnie

devinbinnie (approved):
LGTM! Thanks @arvinDarmawan!

jgilliam17:
Hello @arvinDarmawan can you please check the error and update to latest master? Thanks

devinbinnie:
@arvinDarmawan Let us know if you need any assistance with this PR.

arvinDarmawan:
> @arvinDarmawan Let us know if you need any assistance with this PR.

Hi @devinbinnie , I already update the MR since the old MR (which already reviewed)  have several conflicts

components/status_dropdown/status_dropdown.tsx
```diff
@@ -266,16 +276,16 @@ export default class StatusDropdown extends React.PureComponent {
         );
     }
 
-    refCallback = (ref) => {
+    refCallback = (ref: MenuWrapper): void => {
         if (ref) {
             this.setState({
-                width: ref.state.width,
```
arvinDarmawan, line 272: Hi @devinbinnie , there is no state width in MenuWrapper, should I update state width in this component ?

components/status_dropdown/status_dropdown.tsx
```diff
@@ -266,16 +276,16 @@ export default class StatusDropdown extends React.PureComponent {
...
-                open: ref.state.open,
```
arvinDarmawan, line 273: hi @devinbinnie , there is no state open, so i rename it to state openUp

components/status_dropdown/status_dropdown.tsx
```diff
@@ -290,21 +300,20 @@ export default class StatusDropdown extends React.PureComponent {
...
             direction: 'right',
             text: localizeMessage('status_dropdown.dnd_sub_menu_header', 'Disable notifications until:'),
-        }].concat(
+        } as any].concat(
             this.dndTimes.map((time, index) => {
                 return {
                     id: `dndTime-${time.split(' ').join('')}`,
                     direction: 'right',
                     text: localizeMessage('status_dropdown.dnd_sub_menu_item.time', time),
                     action: index === 4 ? () => setCustomTimedDnd() : () => setDnd(event, index),
```
arvinDarmawan, line 309: Hi @devinbinnie , for event data type, what  is the best data type for this ?
Currently this code will return eslint error "'event' is deprecated".

components/status_dropdown/status_dropdown.tsx
devinbinnie: @arvinDarmawan I think this logic is leftover from a previous version of `MenuWrapper`. And since we don't use either of those variables anywhere, I think we can actually remove the whole ref here since it doesn't do anything.

components/status_dropdown/status_dropdown.tsx
devinbinnie: See above

components/status_dropdown/status_dropdown.tsx
devinbinnie: @arvinDarmawan I don't think that's a typing issue but I think we can ignore that issue for now since it won't trip up the type checker.

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: ok, done

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: ok, done @devinbinnie

components/status_dropdown/status_dropdown.tsx
arvinDarmawan: It will return error on type-check @devinbinnie 
Screen Shot 2021-05-18 at 09 13 49

devinbinnie: I wonder if we can just remove references to `event`? I'm not even sure where the reference to `event` is coming from, so it's probably always going to be undefined.

arvinDarmawan: Hi @devinbinnie , done.

arvinDarmawan:
Hi, maybe CodeQL / Analyze (javascript)  must be retried  since it's cancelled automatically on github

jgilliam17 (approved):
Thank you @arvinDarmawan 
Tested, looks good to merge.
- Verified user status dropdown - as before.
@devinbinnie Can you please help merge? Thanks.

PR #8056, opened 2021-05-03
MM-35407 Update to NPM 7 and prepare to support Node 16

hmhealey:
The biggest changes to support NPM v7 are:
- The lockfile format changed, but it's backwards-compatible, so NPM v6 users will still be able to work with the app until we start using proper workspaces.
- node-sass didn't work for whatever reason, but it's deprecated in favour of dart-sass anyway, so it was an easy swap.
- **This is the big one.** Uncaught promise rejections (aka when an async function errors out in the background) in tests used to just print a warning, but now they cause the test to fail, so those all had to be fixed.

I commented below on some smaller and more context-specific changes.

Note that the ThreadFooter snapshot is still broken with Node 12+ because CI still runs Node 11. That'll be updated when we update CI to use a newer version of Node.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35407

#### Release Note
```release-note
NONE
```


actions/post_actions.test.js
```diff
@@ -92,6 +96,7 @@ describe('Actions.Posts', () => {
                 channels: {
                     current_channel_id: {team_a: 'team_a', id: 'current_channel_id'},
                 },
+                manuallyUnread: {},
```
hmhealey, line 99: Adding this causes an async action to no longer error out which means some extra actions are recorded by the mock store below.

components/add_users_to_team_modal/add_users_to_team_modal.test.tsx
```diff
@@ -53,7 +53,7 @@ describe('components/admin_console/add_users_to_team_modal/AddUsersToTeamModal',
 
         actions: {
             getProfilesNotInTeam: jest.fn(),
-            searchProfiles: jest.fn(),
+            searchProfiles: jest.fn().mockResolvedValue({data: []}),
```
hmhealey, line 56: A lot of these were just actions with missing return values. The code using that return value would then error out asynchronously.

components/admin_console/team_channel_settings/channel/details/channel_details.tsx
```diff
@@ -153,51 +153,49 @@ export default class ChannelDetails extends React.PureComponent<ChannelDetailsPr
         }
     }
 
-    async componentDidMount() {
+    componentDidMount() {
```
hmhealey, line 156: React doesn't wait on an async `componentDidMount` and neither did our tests, so I removed the async/await for simplicity.

components/admin_console/team_channel_settings/channel/details/channel_details.tsx
```diff
@@ -153,51 +153,49 @@ export default class ChannelDetails extends React.PureComponent<ChannelDetailsPr
...
-                if (permission.name === Permissions.CHANNEL_MODERATED_PERMISSIONS.USE_CHANNEL_MENTIONS && (!currentCreatePostRoles[channelRole].value || !currentCreatePostRoles[channelRole].enabled)) {
-                    return {
-                        name: permission.name,
-                        roles: {
-                            ...permission.roles,
-                            [channelRole]: {
-                                value: false,
-                                enabled: false,
+        const currentCreatePostRoles: any = channelPermissions.find((element) => element.name === Permissions.CHANNEL_MODERATED_PERMISSIONS.CREATE_POST)?.roles;
+        if (currentCreatePostRoles) {
```
hmhealey, line 181: Just adding a null check

components/admin_console/team_channel_settings/channel/details/channel_details.tsx
```diff
@@ -43,7 +43,7 @@ export interface ChannelDetailsProps {
     groups: Group[];
     totalGroups: number;
     allGroups: Dictionary<Group>;
-    channelPermissions?: ChannelPermissions[];
+    channelPermissions: ChannelPermissions[];
```
hmhealey, line 46: This was always provided except by unit tests, so I added it to the test props, made it non-optional, and removed a bunch of not-null assertions to make this code easier to read.

store/index.js
```diff
@@ -62,37 +45,6 @@ const whitelist = {
 window.Observable = Observable;
 
 export default function configureStore(initialState) {
-    const setTransformer = createTransform(
```
hmhealey, line 65: This whole transformation logic was used on mobile to serialize/deserialize Set objects being persisted in the store. The web app doesn't do that, so I removed it for simplicity's sake while mocking redux-persist (see below)

tests/redux-persist_mock.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+jest.mock('redux-persist', () => ({
```
hmhealey, line 4: This was previously causing a lot of noise in unit tests since it was causing a lot of null pointer errors. Now that it's mocked out, it won't do that any more

tests/setup.js
```diff
@@ -37,11 +39,36 @@ Object.defineProperty(document, 'execCommand', {
 
 document.documentElement.style.fontSize = '12px';
 
+// isDependencyWarning returns true when the given console.warn message is coming from a dependency using deprecated
+// React lifecycle methods.
+function isDependencyWarning(params) {
```
hmhealey, line 44: This isn't strictly necessary, but the previous version of this code still printed deprecation warnings from those components even if it prevented them from failing the tests. This will completely silence those errors while running tests so we only see useful output.

Willyfrog (approved):
LGTM!

package.json
```diff
@@ -75,6 +75,7 @@
     "regenerator-runtime": "0.13.7",
     "reselect": "4.0.0",
     "rudder-sdk-js": "1.0.16",
+    "sass": "1.32.12",
```
cpoile, line 78: goodbye annoying node-sass! 🎉

cpoile (approved):
Awesome, thanks @hmhealey!

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified [E2E results](https://mm-cypress-report.s3.amazonaws.com/4918-webapp-pr-8056-mattermost-enterprise-edition:dbd6447/mochawesome.html) compared to master e2e [here](https://mm-cypress-report.s3.amazonaws.com/4911-master-e20-prod-mattermost-enterprise-edition:master/mochawesome.html) and short manual smoke test on the PR server - app looks as before.

PR #8057, opened 2021-05-03
Migrated components searchable_user_list and its container to TS.

greenbackboogie97:
**Summary**

-  Migrated components searchable_user_list & searchable_user_list_container to TypeScript.
- Updated imports in relevant components.
- Minor type adjustments in relevant components. 

-------------------------
**Ticket Link**

https://mattermost.atlassian.net/browse/MM-34342

-------------------------

#### Release-Note

```release-note
NONE
```

### DIFF
--- a/components/searchable_user_list/searchable_user_list.tsx
+++ b/components/searchable_user_list/searchable_user_list.tsx
@@ -2,47 +2,47 @@
 // See LICENSE.txt for license information.
 /* eslint-disable react/no-string-refs */
 
-import React, {createRef, ReactNode} from 'react';
+import React, {createRef, ElementType, ReactElement, ReactNode} from 'react';
 
-import {FormattedMessage, injectIntl} from 'react-intl';
+import {FormattedMessage, injectIntl, IntlShape} from 'react-intl';
 
 import QuickInput from 'components/quick_input';
 import UserList from 'components/user_list.jsx';
 import LocalizedInput from 'components/localized_input/localized_input';
 
 import {t} from 'utils/i18n';
-import {IntlProps} from 'components/search_results/types';
+import {UserProfile} from 'mattermost-redux/types/users';
 
 const NEXT_BUTTON_TIMEOUT = 500;
 
  type Props = {
-     users: Array<Record<string, unknown>>;
+     users: UserProfile[];
      usersPerPage: number;
      total: number;
-     extraInfo?: Record<string, unknown>;
-     nextPage: (page?: number) => Promise<void> | void;
-     previousPage: (page?: number) => Promise<void> | void;
+     extraInfo?: Record<string, ReactNode>;
+     nextPage: (page?: number) => Promise<void>;
+     previousPage: (page?: number) => Promise<void>;
      search: (term: string) => void;
-     actions?: ReactNode[];
+     actions?: ReactElement[];
      actionProps?: Record<string, unknown>;
      actionUserProps?: Record<string, unknown>;
      focusOnMount?: boolean;
      renderCount?: (count: number, total: number, startCount: number, endCount: number, isSearch: boolean) => ReactNode;
      filter?: string;
-     renderFilterRow?: (func: {(e: React.FormEvent<HTMLInputElement>): void}) => void;
+     renderFilterRow?: (func: {(e: React.FormEvent<HTMLInputElement>): void}) => ReactNode;
      page: number;
-     term: string | number;
+     term: string;
      onTermChange: (term: string) => void;
-     intl: (msg: Record<string, string>) => string;
+     intl: IntlShape;
      isDisabled?: boolean;
 
      // the type of user list row to render
-     rowComponentType?: ReactNode;
+     rowComponentType?: ElementType;
  }
 
  type State = { nextDisabled: boolean }
 
-class SearchableUserList extends React.PureComponent<Props & IntlProps, State> {
+class SearchableUserList extends React.PureComponent<Props, State> {
     static defaultProps = {
         users: [],
         usersPerPage: 50,
@@ -57,7 +57,7 @@ class SearchableUserList extends React.PureComponent<Props & IntlProps, State> {
     private userList: React.RefObject<UserList>;
     private nextTimeoutId: number;
 
-    constructor(props: Props & IntlProps) {
+    constructor(props: Props) {
         super(props);
 
         this.state = {
@@ -106,7 +106,7 @@ class SearchableUserList extends React.PureComponent<Props & IntlProps, State> {
         this.props.search(e.currentTarget.value);
     }
 
-    renderCount = (users: Array<Record<string, unknown>>) => {
+    renderCount = (users: UserProfile[]) => {
         if (!users) {
             return null;
         }
@@ -170,7 +170,7 @@ class SearchableUserList extends React.PureComponent<Props & IntlProps, State> {
     render() {
         let nextButton;
         let previousButton;
-        let usersToDisplay = this.props.users;
+        let usersToDisplay: UserProfile[];
         const formatMessage = this.props.intl;
 
         if (this.props.term || !this.props.users) {
--- a/components/searchable_user_list/searchable_user_list.tsx
+++ b/components/searchable_user_list/searchable_user_list.tsx
@@ -2,10 +2,8 @@
 // See LICENSE.txt for license information.
 /* eslint-disable react/no-string-refs */
 
-import $ from 'jquery';
-import PropTypes from 'prop-types';
-import React from 'react';
-import ReactDOM from 'react-dom';
+import React, {createRef, ReactNode} from 'react';
+
 import {FormattedMessage, injectIntl} from 'react-intl';
 
 import QuickInput from 'components/quick_input';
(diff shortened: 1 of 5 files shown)


components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
-        // the type of user list row to render
-        rowComponentType: PropTypes.func,
-    };
-
+ type Props = {
+     users: Array<Record<string, unknown>>;
+     usersPerPage: number;
+     total: number;
+     extraInfo?: Record<string, unknown>;
+     nextPage: (page?: number) => Promise<void> | void;
```
hmhealey, line 23: Why can `nextPage` return a promise or nothing? If it's because the functions return nothing in testing, we should update the tests to return a promise instead.

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
+     actionProps?: Record<string, unknown>;
+     actionUserProps?: Record<string, unknown>;
+     focusOnMount?: boolean;
+     renderCount?: (count: number, total: number, startCount: number, endCount: number, isSearch: boolean) => ReactNode;
+     filter?: string;
+     renderFilterRow?: (func: {(e: React.FormEvent<HTMLInputElement>): void}) => void;
+     page: number;
+     term: string | number;
+     onTermChange: (term: string) => void;
+     intl: (msg: Record<string, string>) => string;
```
hmhealey, line 36: This can be typed using the IntlShape type from react-intl

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
+     isDisabled?: boolean;
+
+     // the type of user list row to render
+     rowComponentType?: ReactNode;
+ }
+
+ type State = { nextDisabled: boolean }
+
+class SearchableUserList extends React.PureComponent<Props & IntlProps, State> {
```
hmhealey, line 45: Similarly, we shouldn't need to reference `IntlProps` as defined by that other component if we use the type from react-intl

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -53,59 +54,59 @@ class SearchableUserList extends React.PureComponent {
...
         clearTimeout(this.nextTimeoutId);
     }
 
-    nextPage = (e) => {
-        e.preventDefault();
-
+    nextPage = () => {
         this.setState({nextDisabled: true});
-        this.nextTimeoutId = setTimeout(() => this.setState({nextDisabled: false}), NEXT_BUTTON_TIMEOUT);
+        setTimeout(() => this.setState({nextDisabled: false}), NEXT_BUTTON_TIMEOUT);
```
hmhealey, line 87: Did you intend to remove the use of `nextTimeoutId` here?

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
-        onTermChange: PropTypes.func.isRequired,
-        intl: PropTypes.any,
-        isDisabled: PropTypes.bool,
-
-        // the type of user list row to render
-        rowComponentType: PropTypes.func,
-    };
-
+ type Props = {
+     users: Array<Record<string, unknown>>;
```
hmhealey, line 19: This should be typed `UserProfile[]` where `UserProfile` comes from `mattermost-redux/types/users`

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
+     usersPerPage: number;
+     total: number;
+     extraInfo?: Record<string, unknown>;
+     nextPage: (page?: number) => Promise<void> | void;
+     previousPage: (page?: number) => Promise<void> | void;
+     search: (term: string) => void;
+     actions?: ReactNode[];
```
hmhealey, line 26: I think the correct type for this is `React.ElementType`, not `ReactNode`. I'm not 100% sure since `UserList` isn't migrated, but that seems correct.

The difference between the two is that `ElementType` is used when you're passing around a component class (or functional component or HTML element name), and `ReactNode` is the result of rendering something.
```js
let Element: React.ElementType;
let component: React.ReactNode = <Element/>;
```

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
+     actionProps?: Record<string, unknown>;
+     actionUserProps?: Record<string, unknown>;
+     focusOnMount?: boolean;
+     renderCount?: (count: number, total: number, startCount: number, endCount: number, isSearch: boolean) => ReactNode;
+     filter?: string;
+     renderFilterRow?: (func: {(e: React.FormEvent<HTMLInputElement>): void}) => void;
```
hmhealey, line 32: The return type for this should be `ReactNode` since it's rendering components.

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
+     page: number;
+     term: string | number;
```
hmhealey, line 34: When would `term` be a number?

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
+     onTermChange: (term: string) => void;
+     intl: (msg: Record<string, string>) => string;
+     isDisabled?: boolean;
+
+     // the type of user list row to render
+     rowComponentType?: ReactNode;
```
hmhealey, line 40: This should also be `ElementType`

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -53,59 +54,59 @@ class SearchableUserList extends React.PureComponent {
...
-    handleInput = (e) => {
-        this.props.onTermChange(e.target.value);
-        this.props.search(e.target.value);
+    handleInput = (e: React.FormEvent<HTMLInputElement>) => {
+        this.props.onTermChange(e.currentTarget.value);
+        this.props.search(e.currentTarget.value);
     }
 
-    renderCount = (users) => {
+    renderCount = (users: Array<Record<string, unknown>>) => {
```
hmhealey, line 109: This should also be `UserProfile[]`

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -169,8 +170,8 @@ class SearchableUserList extends React.PureComponent {
     render() {
         let nextButton;
         let previousButton;
-        let usersToDisplay;
-        const {formatMessage} = this.props.intl;
+        let usersToDisplay = this.props.users;
```
hmhealey, line 173: You might be able to change this to `let usersToDisplay: UserProfile[];`. Right now, it looks a bit confusing since we set `usersToDisplay` multiple times in a row

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -169,8 +170,8 @@ class SearchableUserList extends React.PureComponent {
...
+        const formatMessage = this.props.intl;
```
hmhealey, line 174: This is causing errors on the test server since intl is still an object containing a `formatMessage` function

components/searchable_user_list/searchable_user_list_container.tsx
```diff
@@ -1,26 +1,30 @@
...
-        extraInfo: PropTypes.object,
-        nextPage: PropTypes.func.isRequired,
-        search: PropTypes.func.isRequired,
-        actions: PropTypes.arrayOf(PropTypes.func),
-        actionProps: PropTypes.object,
-        actionUserProps: PropTypes.object,
-        focusOnMount: PropTypes.bool,
-    };
+type Props = {
+    users: Array<Record<string, unknown>>;
```
hmhealey, line 9: This is also `UserProfile[]`

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
 // See LICENSE.txt for license information.
 /* eslint-disable react/no-string-refs */
```
hmhealey, line 3: Also thanks for removing the Jquery and ReactDOM usage from this file. We have tickets around to remove those as well, so I can see if they can be linked to this PR too.

Since you've removed the string refs from this file, could you also remove this comment?

hmhealey (changes requested):
Thanks for the PR and for the help getting everything moved over to TypeScript, @greenbackboogie97!

I have some suggestions for how to improve the typing a bit, particularly around the props passed into UserList since it hasn't been migrated yet. Also, the View Team Members modal is now causing the app to crash which needs to be fixed, but I think I might've found the cause (see below).

components/searchable_user_list/searchable_user_list.tsx
```diff
@@ -2,46 +2,47 @@
...
-
-        // the type of user list row to render
-        rowComponentType: PropTypes.func,
-    };
-
+ type Props = {
+     users: Array<Record<string, unknown>>;
+     usersPerPage: number;
+     total: number;
+     extraInfo?: Record<string, unknown>;
```
devinbinnie, line 22: This can be `Record<string, React.ReactNode>`

components/searchable_user_list/searchable_user_list.tsx
devinbinnie: I think it's actually `ComponentType`, not fully sure though.

components/searchable_user_list/searchable_user_list_container.tsx
```diff
@@ -1,26 +1,30 @@
...
+    usersPerPage: number;
+    total: number;
+    extraInfo?: Record<string, unknown>;
```
devinbinnie, line 12: Can also be `Record<string, React.ReactNode>`

devinbinnie (changes requested):
Thanks for the PR @greenbackboogie97!
I've added a few comments for you to review.

greenbackboogie97:
Thanks for the explanations, already learned a lot working on these components. will address these changes as soon as possible! (gotta get that mug).

components/searchable_user_list/searchable_user_list.tsx
greenbackboogie97: Don't know how to use IntlShape type in general and specifically in this case.

components/searchable_user_list/searchable_user_list.tsx
greenbackboogie97: Adding IntlProps resolved the type errors returned from ejectIntl. Again don't really sure how this works.

components/searchable_user_list/searchable_user_list.tsx
greenbackboogie97: nextTimeoutId expects a number so I removed this, now I can see that this is not the solution, an idea here would be great.

components/searchable_user_list/searchable_user_list.tsx
greenbackboogie97: When I change this to UserProfile[], it says that "usersToDisplay" is used before being assigned.

components/searchable_user_list/searchable_user_list.tsx
greenbackboogie97: removed it and got an error that says "Using string literals in ref attributes is deprecated."

devinbinnie: I think we were mistaken about removing the string refs in this file, there are still a few in here.
This ticket isn't about removing string refs, so you don't have to do anything here if you don't want.

components/searchable_user_list/searchable_user_list.tsx
devinbinnie: @greenbackboogie97 Looks like this still needs to be fixed

components/searchable_user_list/searchable_user_list.tsx
devinbinnie: @greenbackboogie97 Can we address this as well?

components/searchable_user_list/searchable_user_list.tsx
devinbinnie: @greenbackboogie97 Can we address this as well?

components/searchable_user_list/searchable_user_list.tsx
devinbinnie: We can import the `IntlShape` type from `react-intl` and you can use that here.

components/searchable_user_list/searchable_user_list.tsx
devinbinnie: You shouldn't need to add `IntlProps` here. What's the error you're getting if you remove it?

components/searchable_user_list/searchable_user_list.tsx
devinbinnie: `setTImeout` returns a `number`, so leaving this alone should be fine.

devinbinnie (commented):
@greenbackboogie97 I've added some comments, and unresolved some issues that still need to be addressed. If you have questions about any of them let us know :)

devinbinnie:
@greenbackboogie97 Looks like there are still some type check issues happening. Make sure to run `make check-types` before you push up any changes. Let us know if we can help with any of the issues :)

hmhealey:
Hi @greenbackboogie97. Is there anything we can help with on this? There's a few type checking issues that need to be resolved, and the one question I asked [here](https://github.com/mattermost/mattermost-webapp/pull/8057#discussion_r625950589) that I don't think was answered.

greenbackboogie97:
Hey sorry I left this open for so long.
Please unassign me from this issue, I got my priorities changed and took on my self more things then I should have.
Sorry for the inconvenience.

jasonblais:
@greenbackboogie97 all good, thanks for your help, I've unassigned the help wanted issue and opened it back up for grabs.

PR #8071, opened 2021-05-05
[MM-34303] Add team ID to response when linking a channel to a group

BenCookie95:
#### Summary
This PR fixes the console error that was occurring. It also fixes an issue where if a team was removed from a group, the channels were not removed from the group in redux state.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34303

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/17576

#### Release Note
```release-note
When a team is removed from a group, we now also remove the associated channels from the group in redux state.
```


packages/mattermost-redux/src/reducers/entities/groups.ts
```diff
@@ -78,21 +78,39 @@ function syncables(state: Dictionary<GroupSyncablesState> = {}, action: GenericA
...
+                nextChannels.splice(index, 1);
+            }
         }
 
         return {
             ...state,
             [action.data.group_id]: {
                 ...state[action.data.group_id],
                 teams: nextTeams,
+                channels: nextChannels,
```
mkraft, line 113: Is this `channels` list tested somewhere?

BenCookie95: Added some tests for the syncable state

mkraft (approved):

shazm (changes requested):
Could we also fix up the broken `components/threading/channel_threads/thread_footer/thread_footer.test.tsx` test in this? I think it might just be updating the snapshot;

```
    -         hour12={true}
    +         hourCycle="h12"
```

Otherwise I've run the tests, and tested manually - seems to be working as intended with the teams / channels 👍

BenCookie95:
> Could we also fix up the broken `components/threading/channel_threads/thread_footer/thread_footer.test.tsx` test in this? I think it might just be updating the snapshot;
> 
> ```
>     -         hour12={true}
>     +         hourCycle="h12"
> ```
> 
> Otherwise I've run the tests, and tested manually - seems to be working as intended with the teams / channels 👍

That test doesn't fail for me locally, might be something to do with your node version? Also, the test passes in the pipeline

shazm (approved):

furqanmlk (approved):
Related Tests Passed ✅

PR #8081, opened 2021-05-07
MM-32788: Permalink'ed replies open thread viewer and jump to in RHS

koox00:
#### Summary

When CRT is enabled clicking to open a permalink to a reply should open
the thread viewer in the RHS, highlight, and jump to the reply linked.

The center view should remain the same AKA the channel user clicked the reply (`returnTo`),
with the exception, if the permalink is not in the same team as the current center channel, 
or there is no `returnTo` (opened in a new tab), then redirect to the channel where the 
reply's root post exists, and open the RHS thread viewer there.

PS: Making the opening of RHS, upon clicking a permalink, seamless (no URL redirect) will be addressed in a follow up [ticket](https://mattermost.atlassian.net/browse/MM-35703) and PR.
 
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-32788

#### Release Note

```release-note
NONE
```


components/permalink_view/permalink_view.jsx
```diff
@@ -47,8 +38,10 @@ export default class PermalinkView extends React.PureComponent {
         document.body.classList.add('app__body');
     }
 
-    componentDidUpdate() {
-        if (!this.state.valid) {
+    componentDidUpdate(prevProps) {
+        const prevPostid = prevProps.match.params.postid;
+        const currPostid = this.props.match.params.postid;
+        if (prevPostid !== currPostid) {
```
koox00, line 44: I fail to see how the post id might change, I kept it only for the sake of discussion

calebroseland (approved):

actions/views/rhs.ts
```diff
@@ -390,6 +391,34 @@ export function selectPost(post: Post) {
...
+        type: ActionTypes.HIGHLIGHT_REPLY,
+        postId: post.id,
+    };
+}
+
+export const clearHighlightReply = {
+    type: ActionTypes.CLEAR_HIGHLIGHT_REPLY,
+};
+
+export const debouncedClearHighlightReply = debounce((dispatch) => {
```
hmhealey, line 405: Interesting. I don't think I've seen debounce used with the trailing option set. This works similar to using setTimeout every time this is called and cancelling the previous timeout if it hadn't occurred yet, right?

components/permalink_view/actions.js
```diff
@@ -3,18 +3,78 @@
...
+        if (channel.type === Constants.DM_CHANNEL) {
+            const userId = getUserIdFromChannelId(channel.name);
+            await dispatch(getMissingProfilesByIds([userId]));
+            const user = getUser(getState(), userId);
+            browserHistory.replace(`/${team.name}/messages/@${user.username}/${postId}`);
+        } else if (channel.type === Constants.GM_CHANNEL) {
+            browserHistory.replace(`/${team.name}/messages/${channel.name}/${postId}`);
+        } else {
+            browserHistory.replace(`/${team.name}/channels/${channel.name}/${postId}`);
+        }
```
hmhealey, line 44: Perhaps we could factor this out into a helper since it looks like we're doing the same thing twice? I wouldn't be surprised if we'd need this logic elsewhere

components/permalink_view/actions.js
```diff
@@ -3,18 +3,78 @@
...
+    return async (dispatch, getState) => {
+        await dispatch(getPostThread(post.parent_id));
+
+        const state = getState();
+
+        const currentChannel = getCurrentChannel(state);
+        const sameTeam = currentChannel && currentChannel.team_id === team.id;
+
+        if (!sameTeam) {
+            dispatch(selectChannel(channel.id));
```
hmhealey, line 58: Would we possibly need to switch teams here as well?

components/permalink_view/actions.js
```diff
@@ -3,18 +3,78 @@
...
+            browserHistory.replace(`/${team.name}/messages/@${user.username}/${postId}`);
+        } else if (channel.type === Constants.GM_CHANNEL) {
+            browserHistory.replace(`/${team.name}/messages/${channel.name}/${postId}`);
+        } else {
+            browserHistory.replace(`/${team.name}/channels/${channel.name}/${postId}`);
+        }
+    };
+}
+
+function focustReplyPost(post, team, channel, returnTo) {
```
hmhealey, line 48: Minor typo here
```suggestion
function focusReplyPost(post, team, channel, returnTo) {
```

components/permalink_view/permalink_view.jsx
hmhealey: I think this would happen if you clicked on a few different permalinks in a short time period because I think it might reuse the PermalinkView between them.

hmhealey (changes requested):

hmhealey:
Looks solid! Just a bit of nitpicking and a couple questions, but it's otherwise good.

actions/views/rhs.ts
koox00: yes it's exactly the same

components/permalink_view/actions.js
koox00: the permalink contains the team, so that happens regardless.

hmhealey (approved):
Changes look good to me, although now there's some conflicts that need to be resolved unfortunately

jgilliam17:
Thanks @koox00 
Looks good, seeing few issues that are not impacting overall function
1. Error when clicking on permalink to GM reply post. User is taken to the correct post but there is an error present. Additionally webapp reloads if you click on the link on global threads (not thread viewer, last bit on the attached video)

https://user-images.githubusercontent.com/52937121/119193861-4607d300-ba50-11eb-994c-316aa5eed281.mov

2. When opening permalink that opens in RHS, bottom of the RHS has an added bar, like an empty scrollbar. Reproduces on Chrome, Safari & Edge, not Firefox
Screen Shot 2021-05-21 at 3 59 51 PM

3. When Switching Thread OFF in account setting, saw few errors, not sure if it's related to this change, no apparent impact in the app, all permalinks navigate to correct location.
Screen Shot 2021-05-21 at 4 15 36 PM

koox00:
> webapp reloads if you click on the link on global threads (not thread viewer, last bit on the attached video)

Thanks @jgilliam17! Good catch! Not introduced in this PR, but I'll push a fix here since it directly affects permalinks!
I'll take a look at console errors as well, thanks again!

jgilliam17:
Thanks @koox00 
Verified various permalink re-directs - looks good. 
Clicking on the link to reply in thread item no longer re-loads the app and verified horizontal bar was removed on the RHS. 
I'll wait to approve after dev reviews just in case I need to retest any requested changes.

matthewbirtch:
@koox00 for permalinks that direct to another channel other than the active one, is there a way to avoid the screen reload and only open the thread in the RHS? I find the reload distracting. Not a big deal, but wanted to ask the question.

koox00:
> @koox00 for permalinks that direct to another channel other than the active one, is there a way to avoid the screen reload and only open the thread in the RHS? I find the reload distracting. Not a big deal, but wanted to ask the question.

Yes I agree @matthewbirtch, I've opened another [ticket](https://mattermost.atlassian.net/browse/MM-35703
) for that

> Also, aren't permalinks supposed to be truncated to "Jump to convo"?

I think "jump to convo" is added by a plugin, maybe this one https://github.com/mattermost/mattermost-plugin-autolink?

matthewbirtch:
Two questions: 
- if I'm in global threads (not RHS) and I click a permalink to a reply that I'm following, I wonder if we should consider opening the thread in the global threads view.  
- Similarly, if I'm in the global threads view and I click a permalink to a post that is from the current thread I'm viewing, we should keep the context in the current thread view and scroll to the highlighted post.

Maybe a later enhancement in a separate PR?

koox00:
> @koox00 Two more questions:
> 
>  * if I'm in global threads (not RHS) and I click a permalink to a reply that I'm following, I wonder if we should consider opening the thread in the global threads view.
> 
>  * Similarly, if I'm in the global threads view and I click a permalink to a post that is from the current thread I'm viewing, we should keep the context in the current thread view and scroll to the highlighted post.
> 
> 
> Maybe a later enhancement in a separate PR?

Does adding those improvements the ["Make clicking on permalinks seamless"](https://mattermost.atlassian.net/browse/MM-35703) makes sense to you?

matthewbirtch:
> Does adding those improvements the ["Make clicking on permalinks seamless"](https://mattermost.atlassian.net/browse/MM-35703) makes sense to you?

I'm fine with that...might want to update the description and title to reflect this, but I'm fine with that approach so we can merge this work and iterate

matthewbirtch (approved):
Approving this with further improvements moving to follow-up PRs

koox00:
@calebroseland, @hmhealey I've added a couple commits to address some issues that came up on QA testing, can you review? tia

calebroseland (approved):

hmhealey (approved):

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.

esethna:
@koox00 feel free to merge when ready given reviews are complete

jwilander:
@koox00 there's some conflicts now that need to be resolved

PR #8082, opened 2021-05-07
MM-35431: fixes save post disabled when opened from threads

koox00:
#### Summary

In the context of threads there is no current channel, so it appeared as
the user never had permissions to edit owned post.
This commit fixes that by using **not** the current channel's id but the
editing post's channel id to get permissions.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35431

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/edit_post_modal/index.js
+++ b/components/edit_post_modal/index.js
@@ -27,14 +27,14 @@ function mapStateToProps(state) {
     const config = getConfig(state);
     const editingPost = getEditingPost(state);
     const currentUserId = getCurrentUserId(state);
-    const channelId = getCurrentChannelId(state);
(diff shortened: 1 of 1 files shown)


jespino (approved):
LGTM

calebroseland (approved):
LGTM! I might suggest is making it based on `editingPost` before currentChannel, as you could edit a post from an RHS-opened thread, yet have a different channel selected, but there's probably other gates before this one.

jgilliam17 (approved):
Thank you @koox00 
Tested, looks good to merge.
- Verified user can open edit modal, edit their own post and save while on thread viewer.

PR #8094, opened 2021-05-11
MM-35591 - multiple sku options in billing subscription pages

pablovelezvidal:
#### Summary
This PR shows the billing information about the cloud subscription depending on the type of tier.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35591

#### Screenshots
Captura de pantalla 2021-05-11 a las 12 46 45
Captura de pantalla 2021-05-11 a las 12 47 21
Captura de pantalla 2021-05-11 a las 12 47 57
Captura de pantalla 2021-05-11 a las 12 48 45

#### Release Note
```release-note
Added a new config setting ServiceSettings.FooBar. Added a new column Foo to the Users table.
```


components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -63,11 +63,36 @@ const BillingSubscriptions: React.FC<Props> = () => {
...
+        subscriptionPlan = 'CLOUD_PROFESSIONAL';
+        break;
+    }
+
     let isFreeTrial = false;
     let daysLeftOnTrial = 0;
     if (subscription?.is_free_trial === 'true') {
         isFreeTrial = true;
         daysLeftOnTrial = getRemainingDaysFromFutureTimestamp(subscription.trial_end_at);
+        subscriptionPlan = 'FREE_TRIAL';
```
marianunez, line 95: We shouldn't change the type of plan if they are in the free trial, in the future, they may be in a trial for enterprise or a trial for professional. We should based any trial decisions in the `isFreeTrial` bool above

components/admin_console/billing/plan_details/plan_details.tsx
```diff
@@ -137,48 +145,50 @@ export const planDetailsTopElements = (userCount: number, isPaidTier: boolean, i
...
-    //         <FormattedMessage
-    //             id='admin.billing.subscription.planDetails.productName.mmCloud'
-    //             defaultMessage='Mattermost Cloud'
-    //         />
-    //     );
-    //     break;
-    // }
+    switch (subscriptionPlan) {
+    case 'CLOUD_PROFESSIONAL':
+    case 'FREE_TRIAL':
```
marianunez, line 159: The free trial condition should not be part of the type of product check, as it could happen that the trial can be started for any plan in the future

components/admin_console/billing/plan_details/plan_details.tsx
```diff
@@ -291,14 +301,14 @@ export const featureList = (isFreeTrial: boolean, isPaidTier: boolean) => {
...
-    //     localizeMessage('admin.billing.subscription.planDetails.features.unlimittedUsersAndMessagingHistory', 'Unlimited users & message history'),
-    //     localizeMessage('admin.billing.subscription.planDetails.features.unlimitedIntegrations', 'Unlimited Integrations'),
-    //     localizeMessage('admin.billing.subscription.planDetails.features.mfa', 'Multi-Factor Authentication (MFA)'),
-    //     localizeMessage('admin.billing.subscription.planDetails.features.multilanguage', 'Multi-language translations'),
-    // ];
+    const featuresCloudStarter = [
+        localizeMessage('admin.billing.subscription.planDetails.features.groupAndOneToOneMessaging', 'Group and one-to-one messaging, file sharing, and search'),
+        localizeMessage('admin.billing.subscription.planDetails.features.incidentCollaboration', 'Incident collaboration'),
+        localizeMessage('admin.billing.subscription.planDetails.features.unlimittedUsersAndMessagingHistory', 'Unlimited users & message history'),
+        localizeMessage('admin.billing.subscription.planDetails.features.unlimitedIntegrations', 'Unlimited Integrations'),
```
marianunez, line 308: Katie mentioned that it is not clear if they will offer "Unlimited Integrations" so we should remove this from the list for now

components/admin_console/billing/plan_details/plan_details.tsx
```diff
@@ -309,42 +319,39 @@ export const featureList = (isFreeTrial: boolean, isPaidTier: boolean) => {
...
 
-    // if (isPaidTier) {
-    //     switch (typeSubscription) {
-    //     case 'FREE_TRIAL':
-    //     // case 'CLOUD_PROFESSIONAL':
-    //         features = featuresCloudProfessional;
-    //         break;
+    if (isPaidTier) {
+        switch (subscriptionPlan) {
+        case 'FREE_TRIAL':
```
marianunez, line 335: Same comment as above about treating free trial as part of the plans

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
```diff
@@ -28,61 +33,52 @@ export const contactSalesCard = (contactSalesLink: any, isFreeTrial: boolean, tr
...
-        );
     }
 
     return (
         <div className='PrivateCloudCard'>
             <div className='PrivateCloudCard__text'>
                 <div className='PrivateCloudCard__text-title'>
-                    {title}
-                    {/* {typeSubscription === 'CLOUD_STARTER' &&
+                    {subscriptionPlan === 'FREE_TRIAL' && title}
```
marianunez, line 42: Same comment as above regarding treating Free Trial as a plan

marianunez (changes requested):
Some minor comments below regarding treating Free Trial as a type of plan but the main change that we need to do is look up the type of plan by price

components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -63,11 +63,36 @@ const BillingSubscriptions: React.FC<Props> = () => {
 
     const [showCreditCardBanner, setShowCreditCardBanner] = useState(true);
 
+    const product = useSelector((state: GlobalState) => {
+        if (state.entities.cloud.products && subscription) {
+            return state.entities.cloud.products[subscription?.product_id];
```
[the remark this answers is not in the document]
marianunez: @nickmisasi Will the subscription have price id or price name so it can be managed by nickname? We need to include this in the Mattermost server models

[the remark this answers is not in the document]
nickmisasi: The subscription holds the product ID, and there are already endpoints/state for holding the different possible products. So you should be able to get the product based on the product ID from state and have all of the information you need there.

nickmisasi:
Here's an example of a multiple products return for the getCloudProducts action:
```
[
    {
        "id": "prod_JLoXfwr9wf0vZn",
        "name": "Mattermost Cloud Starter",
        "description": "",
        "price_per_seat": 4,
        "add_ons": null,
        "sku": "",
        "product_family": "cloud"
    },
    {
        "id": "prod_J62eIfR2GZBQU0",
        "name": "Mattermost Cloud Professional",
        "description": "",
        "price_per_seat": 10,
        "add_ons": null,
        "sku": "",
        "product_family": "cloud"
    }
]
```

Assume there is one more product, but it should work the same for any number (2+) of products. I think you'll need to do some tweaking to this PR to make everything work with this payload. I'll review once those changes have been made

pablovelezvidal:
> Here's an example of a multiple products return for the getCloudProducts action:
> 
> ```
> [
>     {
>         "id": "prod_JLoXfwr9wf0vZn",
>         "name": "Mattermost Cloud Starter",
>         "description": "",
>         "price_per_seat": 4,
>         "add_ons": null,
>         "sku": "",
>         "product_family": "cloud"
>     },
>     {
>         "id": "prod_J62eIfR2GZBQU0",
>         "name": "Mattermost Cloud Professional",
>         "description": "",
>         "price_per_seat": 10,
>         "add_ons": null,
>         "sku": "",
>         "product_family": "cloud"
>     }
> ]
> ```
> 
> Assume there is one more product, but it should work the same for any number (2+) of products. I think you'll need to do some tweaking to this PR to make everything work with this payload. I'll review once those changes have been made

The current implementation handles the list of products as an Array of Products and then transforms it to a Dictionary of keyed-Object Products
https://github.com/mattermost/mattermost-webapp/blob/master/packages/mattermost-redux/src/reducers/entities/cloud.ts#L32

```
function products(state: Dictionary<Product> | null = null, action: GenericAction) {
    switch (action.type) {
    case CloudTypes.RECEIVED_CLOUD_PRODUCTS: {
        const productList: Product[] = action.data;
        const productDict = productList.reduce((map, obj) => {
            map[obj.id] = obj;
            return map;
        }, {} as Dictionary<Product>);
        return {
            ...state,
            ...productDict,
        };
    }
    default:
        return state;
    }
}
```

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
```diff
@@ -91,10 +110,18 @@ export const contactSalesCard = (contactSalesLink: any, isFreeTrial: boolean, tr
...
+                    }
+                    {subscriptionPlan === 'CLOUD_STARTER' &&
+                        <FormattedMessage
+                            id='admin.billing.subscription.privateCloudCard.upgradeNow'
+                            defaultMessage='Upgrade Now'
+                        />
+                    }
                 </a>
             </div>
             <div className='PrivateCloudCard__image'>
```
marianunez, line 127: Super minor detail: There is a new image for the trial version that doesn't have the lock. See https://www.figma.com/file/B3YyrLCLAjNfdPUNq0shQq/Cloud-Professional-Trial-Flow?node-id=103%3A227350

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
```diff
@@ -91,10 +110,18 @@ export const contactSalesCard = (contactSalesLink: any, isFreeTrial: boolean, tr
...
+                            id='admin.billing.subscription.privateCloudCard.contactSales'
+                            defaultMessage='Contact Sales'
+                        />
+                    }
+                    {subscriptionPlan === 'CLOUD_STARTER' &&
+                        <FormattedMessage
+                            id='admin.billing.subscription.privateCloudCard.upgradeNow'
+                            defaultMessage='Upgrade Now'
+                        />
+                    }
```
marianunez, line 124: For this one which is actually "Upgrade Now", this should open the upgrade modal instead of contact sales right?

components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -63,6 +63,31 @@ const BillingSubscriptions: React.FC<Props> = () => {
 
     const [showCreditCardBanner, setShowCreditCardBanner] = useState(true);
 
+    const product = useSelector((state: GlobalState) => {
+        if (state.entities.cloud.products && subscription) {
+            const product = state.entities.cloud.products[subscription?.product_id];
+            return product.product_family === 'cloud' ? product : null;
```
marianunez, line 69: I believe the endpoint already returns all products that are only cloud products so this check here may not be required. @nickmisasi can correct me if I'm wrong here

components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -63,6 +63,31 @@ const BillingSubscriptions: React.FC<Props> = () => {
...
+    switch (product?.name) {
+    case 'Mattermost Cloud Starter':
+        subscriptionPlan = 'CLOUD_STARTER';
+        break;
+    case 'Mattermost Cloud Professional':
+        subscriptionPlan = 'CLOUD_PROFESSIONAL';
+        break;
+    case 'Mattermost Cloud Enterprise':
+        subscriptionPlan = 'CLOUD_ENTERPRISE';
+        break;
```
marianunez, line 85: The product names probably won't include the word "Mattermost" in each, but regardless, should we maybe rely on the nickname that we can keep the same instead of the actual name? @nickmisasi thoughts ?

components/admin_console/billing/plan_details/plan_details.tsx
marianunez: I believe this comment is still outstanding

components/admin_console/billing/plan_details/plan_details.tsx
marianunez: ^

marianunez (changes requested):

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
pablovelezvidal: I thought the same initially, but I was not sure how is the process of upgrade once you have the cloud_starter tier. Is it via the purchase modal?

components/admin_console/billing/billing_subscriptions/index.tsx
pablovelezvidal: ok, @nickmisasi is that validation unnecessary?

components/admin_console/billing/billing_subscriptions/index.tsx
pablovelezvidal: @nickmisasi have you already defined how this nickname will look like?

nickmisasi: Products don't have nicknames. Only prices do. With the multiple products model, this is the only real way we can do this, unless you want to add more metadata

components/admin_console/billing/billing_subscriptions/index.tsx
nickmisasi: Endpoint does indeed only return cloud products so this is probably not needed.

pablovelezvidal: ok

components/admin_console/billing/billing_subscriptions/index.tsx
pablovelezvidal: How hard would it be to add a nickname in the products model @nickmisasi?  @marianunez I think we could use some nickname metadata, something that looks more like an actual key for identifying the tier and not relying in the names, which is a very error prone approach. Any thoughts on this?

marianunez: Will this logic in the CWS also use names for the products? At the end is the same concept but in the backend. Not sure how much we gain there

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
marianunez: [MM-35835](https://mattermost.atlassian.net/browse/MM-35835) will cover the upgrade option and will be addressed outside of this PR - as discussed in today's triage

marianunez:
@pablovelezvidal just noticed on the test server here that we need to remove the reference of the free tier in the upgrade modal:

Screen Shot 2021-05-18 at 4 07 59 PM

Also, can you update the images above for @annelieseklein to do a first UX review pass for the different cloud products?

components/admin_console/billing/billing_subscriptions/index.tsx
marianunez: If we do go with the name approach, we may need to only check if it contains `Starter` or `Professional` but not sure that the actual name will be that verbose like `Mattermost Cloud Starter`

marianunez (approved):
Besides the final decision on how to distinguish the product (by name, here or in CWS?), everything else LGTM!

components/admin_console/billing/billing_subscriptions/index.tsx
pablovelezvidal: ok, I will go with the checking if the name contains a reference for starter, professional or enterprise

pablovelezvidal:
> @pablovelezvidal just noticed on the test server here that we need to remove the reference of the free tier in the upgrade modal:
> 
> Screen Shot 2021-05-18 at 4 07 59 PM
> 
> Also, can you update the images above for @annelieseklein to do a first UX review pass for the different cloud products?

ok!

pablovelezvidal:
> @pablovelezvidal just noticed on the test server here that we need to remove the reference of the free tier in the upgrade modal:
> 
> Screen Shot 2021-05-18 at 4 07 59 PM
> 
> Also, can you update the images above for @annelieseklein to do a first UX review pass for the different cloud products?

In the other PR I have open for the changes in the purchase modal, this is done https://github.com/mattermost/mattermost-webapp/pull/8119/files#diff-acb502910cc86e76a73b356eba7ef5024d0db3aaa31441bbe70ff3c83dc164beR190 . But only when the subscription is freeTrial. But was my understanding that we were still going to keep reference to freeTier to honor the current users that are still in that Tier. Is not that longer the case?

annelieseklein (approved):
looks great!

components/admin_console/billing/billing_subscriptions/index.tsx
nickmisasi: See here: https://github.com/mattermost/mattermost-webapp/pull/8094#discussion_r632726404

nickmisasi: Here's what the products will look like with this latest SKU change:

```
[
    {
        "id": "prod_JLoXfwr9wf0vZn",
        "name": "Mattermost Cloud Standard",
        "description": "This product is used to help testing for multiple cloud tiers",
        "price_per_seat": 4,
        "add_ons": null,
        "sku": "cloud-starter",
        "price_id": "price_1IpfPOC2mR71ZtK2jyMhVIjx",
        "product_family": "cloud"
    },
    {
        "id": "prod_J62eIfR2GZBQU0",
        "name": "Mattermost Cloud Professional",
        "description": "",
        "price_per_seat": 10,
        "add_ons": null,
        "sku": "cloud-professional",
        "price_id": "price_1ITqZPC2mR71ZtK2BP1KVRgs",
        "product_family": "cloud"
    }
]
```

pablovelezvidal: got it!

pablovelezvidal: done @nickmisasi

components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -63,6 +67,46 @@ const BillingSubscriptions: React.FC<Props> = () => {
...
+    case 'cloud-professional':
+        subscriptionPlan = 'CLOUD_PROFESSIONAL';
+        break;
+    case 'cloud-enterprise':
+        subscriptionPlan = 'CLOUD_ENTERPRISE';
+        break;
+    default:
+        subscriptionPlan = 'CLOUD_PROFESSIONAL';
+        break;
+    }
```
nickmisasi, line 108: Now that we have the SKU's in a more code-friendly manner, is this switch statement needed? We could just pass the SKU as the subscriptionPlan, and then use the SKU names `cloud-starter` `cloud-professional` etc here: https://github.com/mattermost/mattermost-webapp/pull/8094/files#diff-24ded2951469e8a6dd5fa613f85043cb6c88fcb5fdd9815b90cb8ad704cf9905R40

nickmisasi, line 108: Let's make the SKU's constants, so you can use them everywhere and only have to change one var.

components/admin_console/billing/plan_details/plan_details.tsx
```diff
@@ -137,48 +145,49 @@ export const planDetailsTopElements = (userCount: number, isPaidTier: boolean, i
...
+        break;
+    default:
+        productName = (
+            <FormattedMessage
+                id='admin.billing.subscription.planDetails.productName.cloudProfessional'
+                defaultMessage='Cloud Professional'
+            />
+        );
+        break;
+    }
```
nickmisasi, line 190: This switch statement can also just compare against the SKU instead of these customer _ cases. And they can use the constants you made earlier :)

nickmisasi (changes requested):
Couple changes I think will simplify the code. Everything else LGTM.

marianunez:
> In the other PR I have open for the changes in the purchase modal, this is done [#8119 (files)](https://github.com/mattermost/mattermost-webapp/pull/8119/files#diff-acb502910cc86e76a73b356eba7ef5024d0db3aaa31441bbe70ff3c83dc164beR190) . But only when the subscription is freeTrial. But was my understanding that we were still going to keep reference to freeTier to honor the current users that are still in that Tier. Is not that longer the case?

@pablovelezvidal apologies, I got my PRs crossed here. I see you have removed it on the other PR 👍 

The customers that are in the current free tier will be honored, however, once they go to the purchase modal to change plans, they will no longer have any free users. Also, the link on how billing works has all the details of this so we are covered there as well.

components/admin_console/billing/billing_subscriptions/index.tsx
marianunez: +1 on this approach 👍

pablovelezvidal: yes, it is changed already!

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
```diff
@@ -29,76 +37,105 @@ export const contactSalesCard = (contactSalesLink: any, isFreeTrial: boolean, tr
...
-            />
-        );
-        description = (
-            <FormattedMessage
-                id='admin.billing.subscription.privateCloudCard.cloudProfessional.description'
-                defaultMessage='Optimize your processes with VPC Peering, a dedicated AWS account and premium support.'
-            />
-        );
+        switch (subscriptionPlan) {
+        case 'cloud-starter':
```
nickmisasi, line 41: Make these `cloud-starter` `cloud-professional` etc constants. If we ever change them, you will have to update multiple switch statements, but with constants only change one variable.

nickmisasi (approved):
one suggestion but otherwise LGTM.

stevemudie (approved):
Looks good so far.

Will test post merge.

amyblais:
@marianunez Looking at the commits for v5.36 release, this will be included in v5.36 and was included in Cloud June 2nd release. Is that ok? I think Mario's latest fast-forward included PRs merged on May 25.

marianunez:
> @marianunez Looking at the commits for v5.36 release, this will be included in v5.36 and was included in Cloud June 2nd release. Is that ok? I think Mario's latest fast-forward included PRs merged on May 25.

Yes, that is fine because the multiple products will not show until we enable it in Stripe.

PR #8002, opened 2021-04-26
MM-32782: RHSThread Follow Button

calebroseland:
#### Summary
Adds follow button in `RHSThread`. 

`actions/websocket_actions.jsx` changes overlap exactly with #7996 as it is required by both.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32782

#### Screenshots
CleanShot 2021-04-26 at 17 16 24@2x

#### Release Note

```release-note
In Collapsed Reply Threads, added ability to follow thread from RHS
```

PR #8107, opened 2021-05-13
CRT: Adds follow/unfollow item in dot menu for RHS

koox00:
#### Summary

This commit enables the user to follow or unfollow a thread or message
from the post's dot menu in the RHS thread viewer.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-32784

#### Screenshots

image

image

#### Release Note

```release-note
NONE
```


components/dot_menu/dot_menu.tsx
```diff
@@ -441,6 +465,24 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
...
+                        show={(
+                            this.props.isCollapsedThreadsEnabled &&
+                                (
+                                    this.props.location === Locations.RHS_ROOT ||
+                                    this.props.location === Locations.RHS_COMMENT
+                                )
+                        )}
+                        {...(this.props.isFollowingThread) ? {
+                            text: this.props.threadReplyCount ? Utils.localizeMessage('threading.threadMenu.unfollow', 'Unfollow thread') : Utils.localizeMessage('threading.threadMenu.unfollowMessage', 'Unfollow message'),
+                            extraText: Utils.localizeMessage('threading.threadMenu.unfollowExtra', 'You won’t be notified about replies'),
```
mickmister, line 479: Cool trick with the spread :+1:

components/dot_menu/dot_menu.tsx
```diff
@@ -441,6 +465,24 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
...
+                        } : {
+                            text: this.props.threadReplyCount ? Utils.localizeMessage('threading.threadMenu.follow', 'Follow thread') : Utils.localizeMessage('threading.threadMenu.followMessage', 'Follow message'),
+                            extraText: Utils.localizeMessage('threading.threadMenu.followExtra', 'You will be notified about replies'),
+                        }}
+                        onClick={this.handleSetThreadFollow}
```
mickmister, line 484: Can this prop be moved up to be more visible?

components/dot_menu/dot_menu.tsx
```diff
@@ -441,6 +465,24 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
...
                     </ChannelPermissionGate>
+                    <Menu.ItemAction
+                        id={`follow_post_thread_${this.props.post.id}`}
+                        show={(
+                            this.props.isCollapsedThreadsEnabled &&
+                                (
+                                    this.props.location === Locations.RHS_ROOT ||
+                                    this.props.location === Locations.RHS_COMMENT
+                                )
+                        )}
```
mickmister, line 476: I would think that a benefit of the post menu option added in this PR is to be able to follow within the center channel. Otherwise you can click the "Follow thread" button that's already at the top of the thread view right?

mickmister (approved):
Nice work @koox00 :+1:

I have a suggestion and a question about when the option should show

components/dot_menu/dot_menu.tsx
koox00: @calebroseland deserves the credit for that on! I just saw it, liked it, and stole it :smile:

components/dot_menu/dot_menu.tsx
koox00: > I would think that a benefit of the post menu option added in this PR is to be able to follow within the center channel

There is a "follow" button in the center channel, since yesterday I think, in the footer of the post :wink:

> Otherwise you can click the "Follow thread" button that's already at the top of the thread view right?

Actually the "Follow thread" is in the global thread viewer, but not in the RHS thread viewer.
Maybe that's confusing? cc @matthewbirtch

koox00: image

calebroseland: https://github.com/mattermost/mattermost-webapp/pull/8002 is almost in for the RHSThread follow button.

calebroseland: There does appear to be a center channel option https://www.figma.com/file/JmNM3WmTKHTKU5Cf7uNwOj/MM-18107-Threads-Overhaul-v3?node-id=1751%3A0

> In the case where there are no replies, the actions menu“Follow/Unfollow thread” menu item reads “Follow/Unfollow Message”

@matthewbirtch

1. Could you confirm if Following menuitem should only be visible when there are no replies? Or should it always be visible?

2. Should Reply menuitem be removed? (when CRT:ON)

components/dot_menu/index.ts
```diff
@@ -65,8 +68,21 @@ function mapStateToProps(state: GlobalState, ownProps: Props) {
     const currentTeam = getCurrentTeam(state) || {};
     const currentTeamUrl = `${getSiteURL()}/${currentTeam.name}`;
 
+    const rootId = post.root_id || post.id;
+    const systemMessage = isSystemMessage(post);
+    let threadReplyCount = 0;
+    let isFollowingThread = false;
+    let threadId = rootId;
+
+    if (rootId && !systemMessage) {
```
calebroseland, line 77: This should be behind feature flag

packages/mattermost-redux/src/reducers/entities/posts.ts
```diff
@@ -241,6 +241,18 @@ export function handlePosts(state: RelationOneToOne<Post, Post> = {}, action: Ge
...
     }
 
+    case ThreadTypes.FOLLOW_CHANGED_THREAD: {
+        const {id, following} = action.data;
+        const post = state[id];
+        return {
+            ...state,
+            [id]: {
+                ...post,
+                is_following: following,
```
calebroseland, line 251: Nice! Thanks for this :)

calebroseland (changes requested):
Overall looks good to me! I'm just seeing one crash-error when you open RHS @mentions.  Here's a patch for a few suggestions regarding that and a couple others: [MM-32784.patch.zip](https://github.com/mattermost/mattermost-webapp/files/6481652/MM-32784.patch.zip) (probably needs a few tweaks—e.g. CenterChannel dot-menu reply)


When I added `getThreadOrSynthetic`, I went back/forth if the signature should be `(state, threadId)` or `(state, post)`. Reason for the latter: I think search-result posts aren't added to `PostsState['posts']`.

components/dot_menu/dot_menu.tsx
matthewbirtch: > In CenterChannel:
> 
> 1. could you confirm if Following menuitem should only be visible when there are no replies? Or should it always be visible?

I think users should always have the option to follow a message from the menu even if there are no replies yet. I just suggested a language change between "Follow/Unfollow Thread" vs" Follow/Unfollow Message"

> 2. Should Reply menuitem be removed? (when CRT:ON)

I don't see a reason to remove the reply menu item—at least for now. Users are still used to having it there, so there may be a transitional period where having it there redundantly is ok.

components/dot_menu/index.ts
koox00: > Overall looks good to me! I'm just seeing one crash-error when you open RHS @mentions. Here's a patch for a few suggestions regarding that and a couple others: [MM-32784.patch.zip](https://github.com/mattermost/mattermost-webapp/files/6481652/MM-32784.patch.zip) (probably needs a few tweaks—e.g. CenterChannel dot-menu reply)

Thanks, I didn't see that!
 
> When I added `getThreadOrSynthetic`, I went back/forth if the signature should be `(state, threadId)` or `(state, post)`. Reason for the latter: I think search-result posts aren't added to `PostsState['posts']`.

You are right, search results throws, I updated the function signature to `(state, post)`,
actually it should be `(state, root)`, else we are going to have incorrect data e.g `reply_count` is 0 for replies.

We could `throw` if there is a `post.root_id` to enforce it, a bit HC though, what do you think?

packages/mattermost-redux/src/selectors/entities/threads.ts
```diff
@@ -57,15 +56,13 @@ export function getThread(state: GlobalState, threadId: $ID<UserThread> | undefi
     return getThreads(state)[threadId];
 }
 
-export function getThreadOrSynthetic(state: GlobalState, threadId: $ID<UserThread>): UserThread | UserThreadSynthetic {
-    const thread = getThreads(state)[threadId];
+export function getThreadOrSynthetic(state: GlobalState, post: Post): UserThread | UserThreadSynthetic {
```
calebroseland, line 59: > actually it should be (state, rootPost), else we are going to have incorrect data e.g reply_count is 0 for replies.

I agree, 100% we should be explicit `(state, rootPost)`

> We could throw if there is a post.root_id to enforce it, a bit HC though, what do you think?

I could see throwing if there's a `root_id` as a development env error, but I think it'd be overkill TBH.

calebroseland (approved):
🎉

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified new dot menu items to Follow/Unfollow Thread and Follow/Unfollow Message

PR #8109, opened 2021-05-13
MM-35509 (part 1) Performance improvements when receiving new posts

jwilander:
#### Summary
This PR strives to fix two different components that render unnecessarily when a new post is received.

The first commit is for the SidebarCategory component. Before my changes, all mounted instances of this component would re-render 1-2 times anytime a new post was received. This was due to the `channels` reducer changing and causing all the selector memoization to fail. I fixed it by adding a deep comparison and some manual memoization into the primary selector. In addition, I also changed the prop type from an array of channels to just an array of channel IDs since only the IDs were used.

The second commit is for the Markdown component. Before my changes, all mounted instances of this component would re-render whenever a new post was received. That means every post (center channel & RHS), the channel header, etc. would re-render. The was due to two props changing for the Markdown component. The first was the `options` prop which was being re-created by the parent component. I fixed that with some memoization. The second was the `channelNamesMap` which came directly from the redux store. It was not memoized for the same reason as the SidebarCategory component had, the root `channels` reducer was changing. I fixed it in a similar way with a deep comparison and manual memoization.

These improvements seemed to give about a 30-40% boost in performance in my local testing.

Here's a profile from before the changes where it took about 320ms for all the processing triggered by a receiving a new post to complete:
Screen Shot 2021-05-13 at 3 44 54 PM

And here's one after the changes, which took about 170ms:
Screen Shot 2021-05-13 at 3 44 29 PM

There's still a lot of room for improvement but hopefully this can have a good impact.

Note: these profiles were done with production builds of the web app.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35509

#### Release Note
```release-note
Performance improvements when receiving new posts
```

### DIFF
--- a/components/post_markdown/post_markdown.tsx
+++ b/components/post_markdown/post_markdown.tsx
@@ -65,8 +65,7 @@ export default class PostMarkdown extends React.PureComponent<Props> {
                 disableGroupHighlight,
                 mentionHighlight,
             };
-        }
-    )
+        })
 
     render() {
         let {message} = this.props;
--- a/components/post_markdown/post_markdown.tsx
+++ b/components/post_markdown/post_markdown.tsx
@@ -2,6 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React from 'react';
+import memoize from 'memoize-one';
 
 import {Post} from 'mattermost-redux/types/posts';
 import {Channel} from 'mattermost-redux/types/channels';
@@ -57,6 +58,16 @@ export default class PostMarkdown extends React.PureComponent<Props> {
         options: {},
     };
 
+    getOptions = memoize(
+        (options: TextFormattingOptions | undefined, disableGroupHighlight: boolean, mentionHighlight: boolean | undefined) => {
+            return {
+                ...options,
+                disableGroupHighlight,
+                mentionHighlight,
+            };
+        }
+    )
+
     render() {
         let {message} = this.props;
         const {post, mentionKeys} = this.props;
@@ -71,10 +82,6 @@ export default class PostMarkdown extends React.PureComponent<Props> {
         // Proxy images if we have an image proxy and the server hasn't already rewritten the post's image URLs.
         const proxyImages = !post || !post.message_source || post.message === post.message_source;
         const channelNamesMap = post && post.props && post.props.channel_mentions;
-        const options = {
-            ...this.props.options,
-            disableGroupHighlight: post?.props?.disable_group_highlight === true, // eslint-disable-line camelcase
-        };
 
         this.props.pluginHooks?.forEach((o) => {
             if (o && o.hook && post) {
@@ -82,9 +89,16 @@ export default class PostMarkdown extends React.PureComponent<Props> {
             }
         });
 
+        let mentionHighlight = this.props.options?.mentionHighlight;
         if (post && post.props) {
-            options.mentionHighlight = !post.props.mentionHighlightDisabled;
+            mentionHighlight = !post.props.mentionHighlightDisabled;
         }
+
+        const options = this.getOptions(
+            this.props.options,
+            post?.props?.disable_group_highlight === true, // eslint-disable-line camelcase
+            mentionHighlight,
+        );
 
         return (
             <Markdown
--- a/components/sidebar/sidebar_channel/sidebar_channel.tsx
+++ b/components/sidebar/sidebar_channel/sidebar_channel.tsx
@@ -143,7 +143,6 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
             multiSelectedChannelIds,
             autoSortedCategoryIds,
         } = this.props;
-        console.log('render sidebar channel ' + channel.id);
 
         let ChannelComponent: React.ComponentType<{channel: Channel; currentTeamName: string; isCollapsed: boolean}> = SidebarBaseChannel;
         if (channel.type === Constants.DM_CHANNEL) {
--- a/components/sidebar/sidebar_channel/sidebar_channel.tsx
+++ b/components/sidebar/sidebar_channel/sidebar_channel.tsx
@@ -143,6 +143,7 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
             multiSelectedChannelIds,
             autoSortedCategoryIds,
         } = this.props;
+        console.log('render sidebar channel ' + channel.id);
 
         let ChannelComponent: React.ComponentType<{channel: Channel; currentTeamName: string; isCollapsed: boolean}> = SidebarBaseChannel;
         if (channel.type === Constants.DM_CHANNEL) {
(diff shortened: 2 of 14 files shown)


components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -143,6 +143,7 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
             multiSelectedChannelIds,
             autoSortedCategoryIds,
         } = this.props;
+        console.log('render sidebar channel ' + channel.id);
```
Willyfrog: seems like a debug statement, right? if so we should remove it

package.json
```diff
@@ -132,6 +132,7 @@
     "@types/tinycolor2": "1.4.2",
     "@typescript-eslint/eslint-plugin": "4.22.0",
     "@typescript-eslint/parser": "4.22.0",
+    "@welldone-software/why-did-you-render": "6.1.1",
```
Willyfrog, line 136: I'm ok with keeping it as a dev dependency, but we should probably document this somewhere so people can use it frecuently and we don't end up having multiple measurements around.

packages/mattermost-redux/src/selectors/entities/channel_categories.ts
```diff
@@ -468,14 +469,32 @@ export function makeSortChannels() {
     };
 }
 
-export function makeGetChannelsForCategory() {
+
+export function makeGetChannelIdsForCategory() {
```
Willyfrog, line 472: can a plugin be using this? if so, we should mark it as a breaking change at least.

Willyfrog (commented):
I'm halfway through the review, I'll try to finish it later but at least I left a couple of comments on it

components/sidebar/sidebar_channel/sidebar_channel.tsx
jwilander: I removed it in a later commit

package.json
jwilander: Actually let me remove it for now. It was helpful but a little heavyweight. I'd prefer to introduce something like that in a separate PR.

packages/mattermost-redux/src/selectors/entities/channel_categories.ts
jwilander: They could be but my guess is it's unlikely. How do we typically handle changing selectors?

I suppose I could leave the old selectors in instead of replacing them

Willyfrog: I don't think I've done it before, maybe @devinbinnie or @hmhealey knows

components/sidebar/sidebar_channel/sidebar_channel.tsx
Willyfrog: my bad, I was going through commits separately to keep apart the changes

Willyfrog (approved):

components/post_markdown/post_markdown.tsx
```diff
@@ -57,6 +58,15 @@ export default class PostMarkdown extends React.PureComponent<Props> {
         options: {},
     };
 
+    getOptions = memoize(
```
lieut-data, line 61: It feels weird to have to pre-emptively memoize the parameters passed to `Markdown`, instead of having that component own that logic itself. Should we implement a custom `shouldComponentUpdate` to replace the default one provided by the `PureComponent` that `Markdown` extends today?

packages/mattermost-redux/src/selectors/entities/channel_categories.ts
lieut-data: Technically, plugins are still using the `mattermost-redux` repository, so this won't break them at present. I doubt anyone is using this anyway.

packages/mattermost-redux/src/selectors/entities/channel_categories.ts
```diff
@@ -468,14 +469,31 @@ export function makeSortChannels() {
...
 
-        return filterAndSortChannelsForCategory(state, channels, category);
+        if (channels.length === 0 && prevChannels.length === 0) {
+            return prevFilteredChannelIds;
+        }
+
+        // getChannels will almost never return a memoized result.
+        // Assumption here is that the added comparison computation is more beneficial
+        // than the cost of returning a non memoized result.
+        if (channels.length === prevChannels.length &&
```
lieut-data, line 488: It looks like `fast-deep-equal` already does a length check: 

https://github.com/epoberezkin/fast-deep-equal/blob/a8e7172b6c411ec320d6045fd4afbd2abc1b4bde/src/index.jst#L56-L59

And it also optimizes for checking the array in reverse (probably assuming that items are changed there first). Does it make sense to simplify this to just the `deepEqual` call?

lieut-data (approved):
Great work, @jwilander! A few non-blocking comments, as I know this is somewhat time-sensitive.

Longer term, it would be great to revisit our Redux data store and evaluate storing the "dirty" bits of channels separately so that things like `getChannels` works "sanely".

components/post_markdown/post_markdown.tsx
jwilander: It is a little odd but I'm hesitant to introduce manual `shouldComponentUpdate` functions unless it's absolutely necessary. They'll run a lot more often than `render` will and can be easy to screw up. I might be overreacting a bit but I remember in the past overusing `shouldComponentUpdate` introduced a lot of headaches. Plus manual `shouldComponentUpdate` functions means we need to check all the props manually ourself.

Maybe this is something we should talk about and come to a guideline or heuristic on for handling these types of things.

packages/mattermost-redux/src/selectors/entities/channel_categories.ts
jwilander: You know what, I should have checked that. I'll remove the manual length check 👍

jwilander:
>Longer term, it would be great to revisit our Redux data store and evaluate storing the "dirty" bits of channels separately so that things like getChannels works "sanely".

@lieut-data 100% agree.

jwilander:
@jgilliam17 ready for your testing now. Can you check the following areas:

* Channels in the sidebar still update to show unreads correctly when someone posts into a channel
* Channel links, eg. `~Off-topic` still update correctly when a channel name gets changed
* Various smoke testing on new posts being received and all the behaviour being normal

jgilliam17:
@jwilander Manual smoke testing listed areas looks good.
 Can you please take a look at the failed check. I've asked Saturn to run E2Es but the PR failed on build. Thanks.

jwilander:
I chatted with @jgilliam17 and we're ok merging this now and running the E2E tests on master.

jgilliam17 (approved):
Thanks @jwilander 
Tested looks good to merge.

PR #8120, opened 2021-05-17
ref: migrate request_button to ts

majidsajadi:
#### Summary

migrate request_button to typescript

#### Ticket Link

 Fixes https://github.com/mattermost/mattermost-server/issues/17249

#### Release Note

```release-note
NONE
```


components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+     * Typically, this will be a <FormattedMessage/>.
+     */
+    helpText?: React.ReactNode;
+
+    /**
+     * A component to be displayed on the button.
+     *
+     * Typically, this will be a <FormattedMessage/>
+     */
+    loadingText?: React.ReactNode;
```
majidsajadi, line 44: this prop was originally defined as a string in `jsx` version but I guess that was a mistake? so I changed that to `ReactNode` according to the description comment.

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+    /**
+     * True if the success message should be show when the request completes successfully,
+     * otherwise false.
+     */
+    showSuccessMessage?: boolean;
+
+    /**
+     * The message to show when the request completes successfully.
+     */
+    successMessage: {
```
majidsajadi, line 88: this one was optional, but I guess since we provide this `successMessage` in default props there is no problem making it required so we don't have to check if `successMessage` is defined to access the inner properties.

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
-        /**
-         * A component that displays help text for the request button.
-         *
-         * Typically, this will be a <FormattedMessage/>.
-         */
-        helpText: PropTypes.element,
+    /**
+     * The message to show when the request returns an error.
+     */
+    errorMessage: {
```
majidsajadi, line 104: same as `successMessage`

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+    busy: boolean;
+    fail: string | null;
+    success: boolean;
+}
 
-    static defaultProps = {
-        id: null,
+export default class RequestButton extends React.PureComponent<Props, State> {
+    static defaultProps: Partial<Props> = {
         disabled: false,
```
majidsajadi, line 141: since the `id` props was optional i remove the default `null` value for `id` because there will be an error if we pass null to `div` as it only accepts `string` or `undefined`

components/admin_console/request_button/request_button.tsx
```diff
@@ -165,7 +171,8 @@ export default class RequestButton extends React.PureComponent {
             success: false,
         });
 
-        const doRequest = () => { //eslint-disable-line func-style
+        const doRequest = () => {
+            //eslint-disable-line func-style
```
hmhealey, line 175: Since moving this onto a new line didn't change anything, I think we might be able to just remove it

hmhealey (changes requested):
Thanks for the PR!

It looks like CI is failing because there are some import paths still looking for the file as a `.jsx` file. You should just be able to remove the extension and then those should find the file correctly.

You'll probably also need to recreate the snapshot for the tests under the new name since that appears to be missing.

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+     * A component that displays help text for the request button.
+     *
+     * Typically, this will be a <FormattedMessage/>.
+     */
+    helpText?: React.ReactNode;
+
+    /**
+     * A component to be displayed on the button.
+     *
+     * Typically, this will be a <FormattedMessage/>
```
michelengelen, line 42: same as above

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+     */
+    requestAction: (
+        success: (data?: any) => void,
+        error: (error: any) => void
+    ) => void;
+
+    /**
+     * A component that displays help text for the request button.
+     *
+     * Typically, this will be a <FormattedMessage/>.
```
michelengelen, line 35: nit: `<FormattedMessage />` (just to be consistent among JSX implementation and documentation)

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+     * ID to assign to the form
+     */
+    id?: string;
+
+    /**
+     * The action to be called to carry out the request.
+     */
+    requestAction: (
+        success: (data?: any) => void,
+        error: (error: any) => void
```
michelengelen, line 29: nit: @hmhealey can't we type this more strict? I personally don't like it when we are using `any`.

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+     * A component to be displayed on the button.
+     *
+     * Typically, this will be a <FormattedMessage/>
+     */
+    loadingText?: React.ReactNode;
+
+    /**
+     * A component to be displayed on the button.
+     *
+     * Typically, this will be a <FormattedMessage/>
```
michelengelen, line 49: same as above

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+     */
+    buttonText: React.ReactNode;
+
+    /**
+     * The element to display as the field label.
+     *
+     * Typically, this will be a <FormattedMessage/>
```
michelengelen, line 56: same as above

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+     */
+    saveNeeded?: boolean;
+
+    /**
+     * Action to be called to save the config, if saveNeeded is set to true.
+     */
+    saveConfigAction?: (callback: () => void) => void;
+
+    /**
+     * True if the success message should be show when the request completes successfully,
```
michelengelen, line 80:
```suggestion
     * True if the success message should be shown when the request completes successfully,
```

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+
+    /**
+     * An element to display adjacent to the request button.
+     */
+    alternativeActionElement?: React.ReactNode;
+};
+
+type State = {
+    busy: boolean;
+    fail: string | null;
```
michelengelen, line 135:
```suggestion
    fail?: string;
```

components/admin_console/request_button/request_button.tsx
```diff
@@ -165,7 +171,8 @@ export default class RequestButton extends React.PureComponent {
             success: false,
         });
 
-        const doRequest = () => { //eslint-disable-line func-style
+        const doRequest = () => {
```
michelengelen, line 174: this can be simplified with an immediate return:

```typescript
        const doRequest = () => this.props.requestAction(
        ...
```

components/admin_console/request_button/request_button.tsx
```diff
@@ -16,122 +14,130 @@ import WarningIcon from 'components/widgets/icons/fa_warning_icon';
...
+
+type State = {
+    busy: boolean;
+    fail: string | null;
+    success: boolean;
+}
 
-    static defaultProps = {
-        id: null,
+export default class RequestButton extends React.PureComponent<Props, State> {
```
michelengelen, line 139: nit: @hmhealey shouldn't we port this to functional components as well? Just being curious here! 👀

components/admin_console/request_button/request_button.tsx
```diff
@@ -255,17 +264,21 @@ export default class RequestButton extends React.PureComponent {
                         >
                             <LoadingWrapper
                                 loading={this.state.busy}
-                                text={this.props.loadingText || Utils.localizeMessage('admin.requestButton.loading', ' Loading...')}
+                                text={
+                                    this.props.loadingText ||
+                                    Utils.localizeMessage(
+                                        'admin.requestButton.loading',
+                                        ' Loading...',
```
michelengelen, line 271:
```suggestion
                                        'Loading...',
```

michelengelen (changes requested):
Some minor changes and `nit`s ... otherwise LGTM!

michelengelen:
@majidsajadi One question about this: Why did the snapshot got removed?

hmhealey:
Hi @majidsajadi. Is there anything we can help with on this? Like is there any feedback you had questions about>

components/admin_console/request_button/request_button.tsx
majidsajadi: i don't like using `any` either but I could find any defined type for `data` or `error` (`error is not a valid javascript `Error` type.

hmhealey (approved):

hmhealey:
Thanks for following up!

michelengelen (approved):
LGTM

jgilliam17 (approved):
Thanks @majidsajadi 
Tested, looks good to merge.
- Verified function of buttons for `Test Connection` in System Console>Environment - as before.

PR #8124, opened 2021-05-18
MM-35689 Fix error caused by trying to render non-string attachment field

hmhealey:
As far as I can tell, field values are the only thing in a message attachment that's allowed to be a string or a number, and we failed to check for that on the client before sending them into the Markdown component. I tried swapping other fields for numbers, but it seems like the server sanitizes those already even though I couldn't find exactly where that happens.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35689

#### Release Note
```release-note
Fixed error caused by a post created with a non-string attachment field.
```


devinbinnie (approved):

BenCookie95 (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified incoming webhook post created with a number in its attachment field renders as expected, no longer crashing webapp.

PR #8129, opened 2021-05-19
Added functionality to archive/unarchive teams from system console

manojmalik20:
#### Summary
Added feature to archive/unarchive teams from the system console

#### Screenshots

image

image

image

#### Release Note

```release-note
Added new feature to archive and unarchive teams through the Teams section in the system console
```


mkraft:
@esethna I've added you for PM review and @matthewbirtch for UX review. Please redirect as appropriate.

matthewbirtch (changes requested):
Thanks for the contribution @manojmalik20! Seems like a great idea to have this feature.

Just a couple notes from me:
1. I can't seem to consistently reproduce this, but when I archive a team and save that change, it kicked me out of the system console and back in to the chat UI. Can we make sure it always routes back to the Teams listing page in the system console?
2. If I archive the currently active team, and I click 'back to {team}' in the top-left, it errors out with 'Team not found'. This makes sense, but we should consider activating the last accessed unarchived team and changing the 'back to {team}' so that it reflects that team.
3. If the server only has one team that is not archived, should we allow archiving of the one team? I wonder if the archive button should be disabled or hidden in this case - maybe with some help text to explain that 'Archiving is disabled when there is only one active team`. I tried archiving the last remaining team and as soon as I archived the last team, it kicked me out of the system console to this screen below. 
image

4. For the teams list view, can we change the `(Archived)` color so that it's not red? Red implies an error. I would suggest going with a 56% or 64% opacity of the standard text color

components/admin_console/team_channel_settings/team/details/team_details.jsx
```diff
@@ -111,6 +118,27 @@ export default class TeamDetails extends React.PureComponent {
         let saveNeeded = false;
 
         const {team, groups: origGroups, teamID, actions} = this.props;
+        if (this.teamToBeArchived()) {
+            const result = await actions.deleteTeam(team.id);
+            if ('error' in result) {
+                serverError = <FormError error={result.error.message}/>;
+                saveNeeded = true;
```
mickmister, line 125: Can we have this instance of `saveNeeded` scoped to this block, since we return at the end of the block? This way there's no confusion on whether "changing" that value here could affect code below this block.
```suggestion
        if (this.teamToBeArchived()) {
            let saveNeeded = false; // add this line here
            const result = await actions.deleteTeam(team.id);
            if ('error' in result) {
                serverError = <FormError error={result.error.message}/>;
                saveNeeded = true;
```

components/admin_console/team_channel_settings/team/details/team_details.jsx
```diff
@@ -111,6 +118,27 @@ export default class TeamDetails extends React.PureComponent {
         let saveNeeded = false;
```
mickmister, line 118: This line would then go below the `if (this.teamToBeArchived())` and `else if (this.teamToBeRestored() && this.state.serverError === null)` blocks, to be a separate variable declaration.

components/admin_console/team_channel_settings/team/details/team_details.jsx
```diff
@@ -361,57 +479,44 @@ export default class TeamDetails extends React.PureComponent {
...
+                            message={
+                                <FormattedMessage
+                                    id='admin.team_settings.team_detail.archive_confirm.message'
+                                    defaultMessage='Saving will archive the team and make its contents inaccessible for all users. Are you sure you wish to save and archive this team?'
+                                />
+                            }
+                            confirmButtonText={
+                                <FormattedMessage
+                                    id='admin.team_settings.team_detail.archive_confirm.button'
+                                    defaultMessage='Save and Archive Team'
```
mickmister, line 505: It seems odd that the save button and modal title both say "Save and Archive Team". Maybe the button should just say "Archive"?

components/admin_console/team_channel_settings/team/details/team_details.jsx
```diff
@@ -330,19 +358,109 @@ export default class TeamDetails extends React.PureComponent {
...
+            // errors, they're no longer relevant.
+            newState.previousServerError = serverError;
+            newState.serverError = null;
+        } else {
+            // if the channel is being unarchived (maybe the user had toggled
+            // and untoggled) the button, so reinstate any server errors that
+            // were present.
+            newState.serverError = previousServerError;
+            newState.previousServerError = null;
+        }
```
mickmister, line 410: I'm not sure that I understand this logic. Why is toggling the archive button showing and hiding a server error? Is it possible that the error has nothing to do with archiving? I'm thinking we should show the error until we have another form submission.

mickmister (changes requested):
Thanks for the PR @manojmalik20! Mostly looks good :+1:  I have a few requests

esethna:
Adding @wiersgallak @aaronrothschild as they might have more context on system console features. If the concept of "archiving a team" didn't exist before this, then I think we need to think through what that actually means, and what happens to existing channels, users, memberships, accounts etc on that team.

components/admin_console/team_channel_settings/team/details/team_details.jsx
manojmalik20: Suppose there is a server error that is because of removing or adding a user in the team, but now the user is archiving the team so those server errors are irrelevant now and if the user has just toggled and untoggled the archive button then those errors should persist in the state. Also, on the channel details screen, the server errors are handled in the same way on archiving/unarchiving the channel. That's why I added the same logic here.

components/admin_console/team_channel_settings/team/list/team_list.tsx
```diff
@@ -228,6 +228,15 @@ export default class TeamList extends React.PureComponent<Props, State> {
                             <div className='TeamList_nameText'>
                                 <b data-testid='team-display-name'>
                                     {team.display_name}
+                                    {team.delete_at !== 0 &&
+                                        <span className='archived-label'>
```
mickmister, line 232: Can we add parens here as well?

```suggestion
                                    {team.delete_at !== 0 && (
                                        <span className='archived-label'>
```

mickmister (commented):
> I was wondering how to stop that event handler to stop redirecting if the event is fired because of the `Archive Team` option in the system console.

Since the system console is separate from the channel view, can we decouple their relationship as far as the "current team"?  We could have conditional logic in the websocket handler for handling the case of being in the system console, that effectively sets the currently active team to null. Then when the user tries to navigate back to the channel view, that's when we send them to the select team page. @manojmalik20 What are your thoughts on this?

aaronrothschild:
> Adding @wiersgallak @aaronrothschild as they might have more context on system console features. If the concept of "archiving a team" didn't exist before this, then I think we need to think through what that actually means, and what happens to existing channels, users, memberships, accounts etc on that team.

I agree with this....what happens to channels?  How can I find them again?  Unarachive the team then unarchive the channels?  

@manojmalik20 Right now, does this essentially "hide" a team without deleting it?

manojmalik20:
> I agree with this....what happens to channels? How can I find them again? Unarchive the team then unarchive the channels?
> 
> @manojmalik20 Right now, does this essentially "hide" a team without deleting it?

Yes @aaronrothschild, this essentially "hides" a team without deleting it. Once you unarchive the team, you don't have to separately unarchive channels unless they were already archived before archiving the team. So basically, on unarchiving a team, it restores back to the state in which it was before archiving.

manojmalik20:
> We could have conditional logic in the websocket handler for handling the case of being in the system console, that effectively sets the currently active team to null. Then when the user tries to navigate back to the channel view, that's when we send them to the select team page. @manojmalik20 What are your thoughts on this?

@mickmister Yes, I agree with you. Sending the user to the select team page instead of finding the previously active unarchived team seems reasonable. So, I'll make that change.

manojmalik20:
@mickmister Yes, I agree with you. But instead of sending the user to select team page directly, what I've done is I'm sending the user to the homepage (eg - `http://{mattermost-url}/`) and if the user has a previously active unarchived team then the user gets redirected to that team and if he doesn't have such a team, then he gets redirected to the select team page. What do you think about this?

manojmalik20:
@matthewbirtch Can you please re-review this?

matthewbirtch:
@manojmalik20 it looks like the issues that I identified are resolved,

>  instead of sending the user to select team page directly, what I've done is I'm sending the user to the homepage (eg - http://{mattermost-url}/) and if the user has a previously active unarchived team then the user gets redirected to that team and if he doesn't have such a team, then he gets redirected to the select team page. What do you think about this?

I think this could work, although it doesn't feel overly smooth with the redirects and flashes of other screens in that process. Not sure if there is any way to avoid that flicker?

manojmalik20:
@matthewbirtch That flash of other screen is because we are directing the user to the "/" path. That flash can be removed by redirecting the user directly to the `select_team` screen, but we cannot do that because if the user is part of an unarchived team, he should be redirected to that team instead of the `select_team` page. So I guess we cannot remove that flicker. If there's any other solution, maybe the developers can help me out. @mickmister

matthewbirtch (approved):

manojmalik20:
This PR is ready for re-review @mickmister @mkraft @aaronrothschild @wiersgallak

mkraft (approved):

esethna:
@manojmalik20 really appreciate the PR. 

I have some thoughts, I'm not sure if some of them were addressed since there's no ticket for this. I'm worried there's a lot to the idea of "Archiving a Team" that we may not be considering (correct me if I'm wrong), off the top of my head here are a few: 

1. What happens to users where this was their only team? If they get dropped at the select team page how are users informed what happened or where their team went?
2. What happens to users where this was one of a few teams they were members of? Does it hide the team in the team sidebar? If so, how is the user informed of why they can't find that team anymore?
3. What about the above cases if the user is logged in via mobile?
4. What if a user hits a permalink or email notification to the archived team?
5. What if the user taps a push notification to the archived team?

I'm sure there's more considerations than the above. I think this is a large feature and probably shouldn't be taken without a ticket and expected behaviour that's agreed on by UX and PM. My proposal would be to close this PR until that exercise is completed. 

cc// @matthewbirtch @aaronrothschild

chetanyakan:
@esethna Before we decide to close this PR, I just wanted to point out that the functionality to archive/unarchive teams already exists in Mattemost CLI commands.
This PR just exposes the functionality to the system console UI.

Maybe it makes sense to brand this feature as `beta`/`experimental` with some warnings instead of closing the PR entirely. (1/5)

esethna:
Hmm thanks @chetanyakan I wasn't aware. The challenge there is that but surfacing the UI button to archive, we are promoting the use of this feature (even if our docs label it as experimental or beta). And without the comprehensive design I'm worried about admins using this feature with the expectation that all of the above considerations are handled gracefully.

esethna:
@chetanyakan Can you help me understand the use cases for needing this feature? Maybe that will help with understanding how important the considerations I listed above might be?

esethna:
@mkraft @sbishel any thoughts on [the above ](https://github.com/mattermost/mattermost-webapp/pull/8129#issuecomment-863495151), maybe you guys are more familiar with the use cases for this feature and if these concerns are warranted given the ability to archive a team already exists in the CLI

mkdbns:
@esethna , the use case for this is pretty simple. There are customers who have MM System Admins that want to be able to archive / unarchive teams but do not have access to the host / cli - or they don't want to go through the hassle of ssh-ing in. This just gives them the ability to do so via the system console just like they can do for channels. Honestly, it seems like a gap in the Teams Admin UI that we're just trying to improve on. Does that make sense?

esethna:
Thanks @mkdbns 

@wiersgallak looping you in here as I think you might be more familiar with this area of the product than I am. Given that archiving a team already exists via the CLI I'm not opposed to accepting this PR. 

With that said, regardless of if an admin is archiving a team from the CLI or these new system console buttons, we should recognize there are likely gaps in the UX of this feature from the end-user perspective as I noted above. However, if the typical use case for archiving a team is one where end users are no longer active in the team, and they wouldn't notice if it simply disappeared, then maybe my concerns are not warranted (and could potentially be solved by documentation).  

@wiersgallak I will leave it your hands to approve this PR or not, given your understanding of typical admin workflows and any data you might have on requests for this feature.

wiersgallak (approved):
Thanks, Eric.  This is approved from my perspective.  There are some edge cases that we will need to build upon, but we can queue some additional tickets for this.  This definitely adds value in that the action is now exposed in the UI.  We will ensure that we add the nuances of the ideal situation for end-users in the documentation of this feature until we are able to iterate on the experience.

components/admin_console/team_channel_settings/team/details/team_profile.jsx
```diff
@@ -56,6 +69,35 @@ export function TeamProfile({team}) {
 
                         </div>
                     </div>
+                    {/* We are using same classes used for Archive Channel button so that we can use the same styling as Arvhive channel button */}
```
mickmister, line 72: I personally think the comments regarding class names are not needed here. We usually want to include comments when there is logic that needs to be explained. I think we are fine without the comment here and the one below.

Though if we do include the comments, they should use the same comment style:
```suggestion
                    // We are using same classes used for Archive Channel button so that we can use the same styling as Archive channel button
```

mickmister (approved):
Thanks for implementing the changes @manojmalik20 :+1: I'm approving with one non-blocking request regarding the code comments added in the last commit.

manojmalik20:
@mickmister I've removed the comments.

mickmister:
@manojmalik20 Please avoid force-pushing when possible. When a PR's commits are force-pushed, the commit history of the PR is not maintained, so there's no way to tell what has changed since last review, and so the PR needs to be reviewed again. Especially with an open-source project, this is very important as we need to be certain about any code that is introduced.

Hopefully this makes sense 🙂

manojmalik20:
@mickmister Yeah this makes sense. I'll keep this in mind from now on.

manojmalik20:
@mkraft @mickmister @wiersgallak Is there anything left on this PR?

jasonblais:
Sorry for the delay, and thanks for the contribution @manojmalik20! 

The last step is a QA review, I've requested @furqanmlk to help. Once completed, we're ready to merge! :)

furqanmlk (commented):
N/A

furqanmlk (approved):
Working as expected. all e2e test cases passed ✅

jasonblais:
Thanks again @manojmalik20! 

cc @amyblais not sure if this should have a Jira ticket for release test coverage.

PR #8132, opened 2021-05-19
Update CI to use Node 14

hmhealey:
We're finally able to update the build image again, so we can finally start getting CI onto the versions of Node we're recommending. Upgrading to Node 14 requires updating one snapshot that changed, and it also lets us remove the library that added internationalization data for testing since that's now built into node.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34363

#### Release Note
```release-note
NONE
```


metanerd (approved):
LGTM!

calebroseland (approved):

PR #8133, opened 2021-05-19
ref: migrate announcement_bar to ts

majidsajadi:
#### Summary

migrate announcement_bar to typescript

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-18948
 Fixes https://github.com/mattermost/mattermost-server/issues/16742

#### Release Note

```release-note
NONE
```


components/announcement_bar/announcement_bar_controller.tsx
```diff
@@ -13,26 +17,25 @@ import CloudAnnouncementBar from './cloud_announcement_bar';
...
-        totalUsers: PropTypes.number,
-        warnMetricsStatus: PropTypes.object,
-        actions: PropTypes.shape({
-            dismissError: PropTypes.func.isRequired,
-        }).isRequired,
-    }
+type Props = {
+    license?: ClientLicense;
+    config?: Partial<ClientConfig>;
+    user?: Partial<UserProfile>;
```
hmhealey, line 23: I'm not sure why the propTypes had this as a shape object. Perhaps we could just use UserProfile for this now instead of the partial? If that doesn't work because that the tests don't provide all the fields to make up a user object, you can use `TestHelper.getUserMock` to add the missing ones.

components/announcement_bar/announcement_bar_controller.tsx
```diff
@@ -13,26 +17,25 @@ import CloudAnnouncementBar from './cloud_announcement_bar';
...
+    canViewSystemErrors: boolean;
+    latestError?: {
+        error: any;
+    };
+    totalUsers?: number | AnalyticsRow[];
```
hmhealey, line 28: I don't know what's going on with the type here, but since it isn't used by ConfigurationAnnouncementBar, can we also remove it from here?

components/announcement_bar/announcement_bar_controller.tsx
hmhealey: Actually, this isn't used, so could you just remove it?

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
```diff
@@ -35,7 +33,7 @@ type Props = {
     modalButtonDefaultText?: string;
     showLinkAsButton: boolean;
     icon?: React.ReactNode;
-    warnMetricStatus?: Dictionary<WarnMetricStatus>;
```
hmhealey, line 38: Is this correct? It looks like we're still using it as a dictionary in `ConfigurationAnnouncementBar`

hmhealey (changes requested):
Thanks for the PR. The tests are failing for this, so would you mind fixing those? It looks like the snapshots are broken, possibly because of the changes to some of the props that aren't used anymore.

components/announcement_bar/announcement_bar_controller.tsx
majidsajadi: there were the same props in `configuration_bar.tsx` I removed these props from that component too. 

also, I removed these props from `announcement_bar/index.ts`

components/announcement_bar/announcement_bar_controller.tsx
majidsajadi: done

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
majidsajadi: in `announcement_bar/index.ts` we get `warnMetricsStatus` with type ` Dictionary<WarnMetricStatus>` from `getWarnMetricsStatus`.
then we pass it to `AnnouncementBarController`and then from there to `ConfigurationAnnouncementBar`. in   `ConfigurationAnnouncementBar` we loop over `warnMetricsStatus` dictionary and pass `status` to `AnnouncementBar`. 

also after your comment, I checked the code and find out in the `WarnMetricAckModal` (the modal type we call in `AnnouncementBar`), the `warnMetricStatus` prop has type `any` which I guess is wrong and should be `WarnMetricStatus`.

components/announcement_bar/announcement_bar_controller.tsx
```diff
@@ -2,37 +2,36 @@
...
+    actions: {
+        dismissError: (index: number) => void;
+    };
+}
 
+export default class AnnouncementBarController extends React.PureComponent<Props> {
     render() {
         let adminConfiguredAnnouncementBar = null;
-        if (this.props.config.EnableBanner === 'true' && this.props.config.BannerText.trim()) {
+        if (this.props.config && this.props.config.EnableBanner === 'true' && this.props.config.BannerText?.trim()) {
```
marianunez, line 34: nit: Optional chaining may also work here to reduce the checks

```suggestion
        if (this.props.config?.EnableBanner === 'true' && this.props.config.BannerText?.trim()) {
```

components/announcement_bar/announcement_bar_controller.tsx
```diff
@@ -57,7 +56,7 @@ export default class AnnouncementBarController extends React.PureComponent {
         let cloudAnnouncementBar = null;
         let paymentAnnouncementBar = null;
         let cloudTrialAnnouncementBar = null;
-        if (this.props.license.Cloud === 'true') {
+        if (this.props.license && this.props.license.Cloud === 'true') {
```
marianunez, line 59:
```suggestion
        if (this.props.license?.Cloud === 'true') {
```

components/announcement_bar/text_dismissable_bar.tsx
```diff
@@ -13,27 +12,31 @@ import AnnouncementBar from './default_announcement_bar';
...
-        allowDismissal: PropTypes.bool.isRequired,
-        text: PropTypes.string.isRequired,
-        onDismissal: PropTypes.func,
+type AnnouncementBarProps = React.ComponentProps<typeof AnnouncementBar>;
 
-        // Any props that AnnouncementBar supports
-    };
+interface Props extends Partial<AnnouncementBarProps> {
+    allowDismissal?: boolean;
+    text?: string;
```
marianunez, line 19: Should `allowDismissal` and `text` be optional? It seems that in the original props they were required

marianunez (approved):
Thanks @majidsajadi! Minor comments below

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
hmhealey: Good catch on the `any`. I guess I just got a bit confused reading the code. This looks good to me now.

hmhealey (approved):

hmhealey:
Updating the branch to fix the tests. There were some issues with CI that were fixed upstream.

jgilliam17 (approved):
Thank you @majidsajadi 
Tested, looks good to merge.
- Verified announcement banner functions as before.

PR #8138, opened 2021-05-21
Update PULL_REQUEST_TEMPLATE.md

chetanyakan:
#### Summary
- Remove `has redux changes` link option as there will be no changes to redux repo.

```release-note
NONE
```


hmhealey (approved):
Good catch. That template is crowded enough as it is with relevant info, so removing irrelevant stuff is an excellent idea.

hanzei:
Thanks @chetanyakan :+1:

PR #8142, opened 2021-05-21
Multiselect for commands

larkox:
#### Summary
Add multiselect for commands

#### Ticket Link
None

#### Screenshots
Screenshot from 2021-05-21 17-46-02
Screenshot from 2021-05-21 17-46-15
Screenshot from 2021-05-21 17-46-07

#### Release Note
```release-note
Add support for multiselect on apps slash commands
```

### DIFF
--- a/components/suggestion/command_provider/app_command_parser/app_command_parser.ts
+++ b/components/suggestion/command_provider/app_command_parser/app_command_parser.ts
@@ -3,6 +3,8 @@
 
 /* eslint-disable max-lines */
 
+import {AppSelectOption} from 'mattermost-redux/types/apps';
+
 import {
     AppCallRequest,
     AppBinding,
@@ -63,6 +65,16 @@ export enum ParseState {
     EndQuotedValue = 'EndQuotedValue',
     EndTickedValue = 'EndTickedValue',
     Error = 'Error',
+    MultiselectStart = 'MultiselectStart',
+    MultiselectStartValue = 'MultiselectStartValue',
+    MultiselectNonspaceValue = 'MultiselectNonspaceValue',
+    MultiselectQuotedValue = 'MultiselectQuotedValue',
+    MultiselectTickValue = 'MultiselectTickValue',
+    MultiselectEndValue = 'MultiselectEndValue',
+    MultiselectEndQuotedValue = 'MultiselectEndQuotedValue',
+    MultiselectEndTickedValue = 'MultiselectEndTickedValue',
+    MultiselectValueSeparator = 'MultiselectValueSeparator',
+    MultiselectNextValue = 'MultiselectNextValue',
 }
 
 interface FormsCache {
@@ -86,7 +98,7 @@ export class ParsedCommand {
     formsCache: FormsCache;
     field: AppField | undefined;
     position = 0;
-    values: {[name: string]: string} = {};
+    values: {[name: string]: string | string[]} = {};
     location = '';
     error = '';
     intl: Intl;
@@ -396,6 +408,16 @@ export class ParsedCommand {
                         id: 'apps.error.parser.unexpected_whitespace',
                         defaultMessage: 'Unreachable: Unexpected whitespace.',
                     }));
+                case '[':
+                    if (!this.field?.multiselect) {
+                        return this.asError(this.intl.formatMessage({
+                            id: 'apps.error.parser.unexpected_squared_bracket',
+                            defaultMessage: 'Unexpected list opening.',
+                        }));
+                    }
+                    this.state = ParseState.MultiselectStart;
+                    this.i++;
+                    break;
                 default: {
                     this.state = ParseState.NonspaceValue;
                     break;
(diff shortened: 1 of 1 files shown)


levb (changes requested):
I question the choice of the syntax. If we already know that it's a multivalue, what is the need for `[` and `]`? Why can it not just be `--name=v1,v2... --other=something`?

larkox:
Good question. Probably it is not needed, but I preferred having things more strictly defined. For example, imagine I have `/command subcommand --multivalue v1 , v2 , v3 `. What should be the suggestions I should be receiving? If we don't have limits that have to be closed for lists, the list of suggestions would be the following:
- `,` to add another element
- any other field in the form

I like it better that the two only options you have at that level are "add one more element or close the list".

larkox:
Started a discussion with UX here: https://community.mattermost.com/core/pl/6ftyowg91tds9yfk8yw8d45ccr

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
```diff
@@ -396,6 +408,16 @@ export class ParsedCommand {
                         id: 'apps.error.parser.unexpected_whitespace',
                         defaultMessage: 'Unreachable: Unexpected whitespace.',
                     }));
+                case '[':
+                    if (!this.field?.multiselect) {
+                        return this.asError(this.intl.formatMessage({
+                            id: 'apps.error.parser.unexpected_squared_bracket',
+                            defaultMessage: 'Unexpected list opening.',
+                        }));
+                    }
```
mickmister, line 417: What if it makes sense in the user's case for a non-multiselect text value to begin with `[`?

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
```diff
@@ -526,6 +548,230 @@ export class ParsedCommand {
...
+                    this.i++;
+                    break;
+                default:
+                    this.state = ParseState.MultiselectStartValue;
+                }
+                break;
+            default:
+                return this.asError(this.intl.formatMessage({
+                    id: 'apps.error.parser.unexpected_state',
+                    defaultMessage: 'Unreachable: Unexpected state in matchBinding: `{state}`.',
```
mickmister, line 771: I think we're trying to limit the number of low-level error messages going forward. Can we instead log the error and display a more generic error?

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
```diff
@@ -731,65 +977,203 @@ export class AppCommandParser {
             }
             switch (f.type) {
             case AppFieldTypes.DYNAMIC_SELECT:
+                if (f.multiselect && Array.isArray(values[f.name])) {
+                    const options: AppSelectOption[] = [];
+                    const commandValues = values[f.name] as string[];
+                    for (let i = 0; i < commandValues.length; i++) {
```
mickmister, line 983: We can have slightly cleaner syntax here

```suggestion
                    for (const value of commandValues) {
```

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
```diff
@@ -731,65 +977,203 @@ export class AppCommandParser {
...
                         fieldName: f.name,
-                        option: values[f.name],
+                        option: value,
                     });
+                };
+
+                if (f.multiselect && Array.isArray(values[f.name])) {
+                    const options: AppSelectOption[] = [];
+                    const commandValues = values[f.name] as string[];
+                    for (let i = 0; i < commandValues.length; i++) {
```
mickmister, line 1019: Same here

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
```diff
@@ -731,65 +977,203 @@ export class AppCommandParser {
...
+                        fieldName: f.name,
+                        option: username,
+                    });
+                };
+
+                if (f.multiselect && Array.isArray(values[f.name])) {
+                    const options: AppSelectOption[] = [];
+                    const commandValues = values[f.name] as string[];
+                    /* eslint-disable no-await-in-loop */
+                    for (let i = 0; i < commandValues.length; i++) {
```
mickmister, line 1075: Same here

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
```diff
@@ -731,65 +977,203 @@ export class AppCommandParser {
...
+                        fieldName: f.name,
+                        option: channelName,
+                    });
+                };
+
+                if (f.multiselect && Array.isArray(values[f.name])) {
+                    const options: AppSelectOption[] = [];
+                    const commandValues = values[f.name] as string[];
+                    /* eslint-disable no-await-in-loop */
+                    for (let i = 0; i < commandValues.length; i++) {
```
mickmister, line 1141: Same here

mickmister (approved):
Great work @larkox 

I have a request to avoid adding more user-facing low-level error messages. No issues found otherwise. I'm curious about the design discussion of using `[` for the lists

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
larkox: They would have to use quotes. Not sure if we can make it obvious.

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
larkox: Being an unreachable error message (this should never happen to a end user) I think it is ok having it as a low-level message. @aaronrothschild thoughts?

mickmister: I'm not sure if the likelihood of the error changes the principle. We are sort of polluting the `i18n.json` file with errors that are not very helpful. @enahum may have some input on this

mickmister (approved):

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
larkox: Should we make this error not translated?

levb (approved):

larkox:
@DHaussermann friendly reminder for review.

DHaussermann:
@larkox I'm seeing a webapp crash here. The webapp breaks when I try to load the channel members  


image
`The above error occurred in the <ChannelMentionSuggestion> component:`
chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/react_devtools_backend.js 

Steps:
- Deploy this Webapp PR (Server branch is on `master`)
- Enable the test-app
- Type `/test valid multiselect_command --static sv1 --user @member1 --channel`
- Type the space character   

Please advise if you can reproduce this.

larkox:
@DHaussermann Thank you for the feedback. There was an error introduced by one of the merges. Should be solved now. Going to re-request @mickmister for the changes in the code, just in case.

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
mickmister: If we haven't added it to the `en.json` in a previous PR, I think we should avoid adding it here. We never arrived at a full plan to handle this, discussed briefly here https://community-daily.mattermost.com/core/pl/jg79zt5otpnuzccdo79wpb75gy

mickmister (approved):
LGTM

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
larkox: These are already in place and translated. We can remove them, but I think it would be better in a different PR.
The strings added in this PR are all user facing.

larkox:
@DHaussermann gentle reminder for your review on this.

PR #8143, opened 2021-05-21
MM-35838/MM-35923 Fix memoization of selectors

hmhealey:
Turns out https://github.com/mattermost/mattermost-webapp/pull/8109 caused some issues around memoization since we weren't checking results at the right time. In particular, `makeGetChannelIdsForCategory` was checking the channels for changes before doing filtering and sorting which caused the results to not be updated when they should've changed because something else (sort order or DM autohiding limits) should've caused their results to change.

`getChannelsNameMapInCurrentTeam` didn't cause any of these issues, but it was only returning a new result when one of the channels' display names changed. While this seems like it was okay for the cases we used it in, I'd be worried about there being some cause we didn't expect here where it would break. This will result in slightly worse performance of the Markdown component (notably, it'll now re-render again when a post is made because one of the channels it contains had its last_post_at change), but we could make a new selector that just maps names to display names if we think that's an issue. I started doing that, but the code around the channel linking in Markdown is a bit sketchy looking, so I didn't want to risk breaking it if we didn't need it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35838
https://mattermost.atlassian.net/browse/MM-35923
https://mattermost.atlassian.net/browse/MM-35981
https://mattermost.atlassian.net/browse/MM-35982
https://mattermost.atlassian.net/browse/MM-35983

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8109

#### Release Note
```release-note
NONE
```


packages/mattermost-redux/src/selectors/entities/channel_categories.ts
```diff
@@ -472,26 +471,20 @@ export function makeSortChannels() {
...
-            return prevFilteredChannelIds;
-        }
+        const filteredChannelIds = filterAndSortChannelsForCategory(state, channels, category).map((channel) => channel.id);
 
-        // getChannels will almost never return a memoized result.
-        // Assumption here is that the added comparison computation is more beneficial
-        // than the cost of returning a non memoized result.
-        if (deepEqual(channels, prevChannels)) {
-            return prevFilteredChannelIds;
+        if (shallowEquals(filteredChannelIds, lastChannelIds)) {
```
hmhealey, line 482: This was the big change for this one since we're checking equality of the filtered IDs instead of the prefiltered channels.

packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -406,28 +405,23 @@ export const getChannelsNameMapInTeam: (state: GlobalState, teamId: string) => N
...
             const channel = channels[id];
             channelMap[channel.name] = channel;
-            channelDisplayNameMap[channel.name] = channel.display_name;
         });
 
         // getAllChannels will almost never return a memoized result.
         // Assumption here is that the added comparison computation is more beneficial
         // than the cost of returning a non memoized result.
-        if (deepEqual(channelDisplayNameMap, prevChannelDisplayNameMap)) {
+        if (shallowEquals(channelMap, prevChannelMap)) {
```
hmhealey, line 421: The deep equality check isn't actually needed for these since channel objects are "immutable" when they come from the redux store

jwilander:
@hmhealey if the changes to `getChannelsNameMapInCurrentTeam` weren't causing any issues can we leave that as is? The Markdown component re-rendering was a fairly large part of the performance issues that my profiling was showing

packages/mattermost-redux/src/selectors/entities/channel_categories.ts
```diff
@@ -472,26 +471,20 @@ export function makeSortChannels() {
...
+            return lastChannelIds;
         }
 
-        prevChannels = channels;
-        prevFilteredChannelIds = filterAndSortChannelsForCategory(state, channels, category).map((channel) => channel.id);
-        return prevFilteredChannelIds;
+        lastChannelIds = filteredChannelIds;
+        return filteredChannelIds;
```
lieut-data, line 487: nit: can we return `lastChannelIds` for "consistency" here? I had to read this a second time before I recognize the pattern.

lieut-data (approved):
Awesome to see forward momentum on memoization issues!

hmhealey:
@jwilander I'll take another look, but I can't remember if it was actually causing issues or not. The issue would be the same as the channel sidebar one if it did occur, but the nature of the problem makes it difficult to find since the code looks right.

I'll check again to see if it's actually causing an issue at the moment, but if it's not, I could revert the changes I made and cover it with comments explaining the bug instead. I can also look at fixing it in a way that behaves how we want it to.

Willyfrog (approved):

jwilander (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
MM-35981 Verified DM displays on the sidebar when DM URL is pasted in the browser bar.
MM-35982 Verified archived channels remains on the sidebar when pasting/opening a link to an archived channel into browser bar
MM-35983 Verified archived channel is displayed on the sidebar after archiving and is removed from sidebar when changing the channel.
MM-35838 Verified DMs/GMs are removed from LHS once closed 
MM-35923 Verified category sort no longer needs a refresh to take effect.

amyblais:
For cherry-picking to ``cloud``.

amyblais:
@hmhealey You can cherry-pick it manually.

PR #8149, opened 2021-05-24
MM-35717: CRT, new messages line

koox00:
#### Summary

Adds a "new messages" line separator in the thread viewer at the event of unread messages.

Adds a `state.views.threads.lastViewedAt` to cache the `last_viewed_at` of
the thread when:
- opening a thread either from the global threads or via
- `selectPost` action, or 
- when marking a reply as unread.

Websocket events also update the `lastViewedAt`, but on the condition that
the window is not active or the thread is not the selected thread.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35717

#### Screenshots

image

#### Release Note

```release-note
NONE
```


matthewbirtch:
@koox00 is this working only in the context of the RHS?  I'm not seeing the new message line showing in the global thread view?

Also, it seems like the message is too tightly pushed up against the new message line. This is especially noticeable with consecutive messages that don't show the post header. I think we need to add some padding to the top of the post that appears after the new message line - the same way we do in-channel. 
image

koox00:
> @koox00 is this working only in the context of the RHS? I'm not seeing the new message line showing in the global thread view?

It should be working in global threads as well, 
it seems that it's clearing the new messages line as quick as you enter, I'll push a fix, thx!
PS: mark as unread works.
image

koox00:
@matthewbirtch both issues should be OK now, can you verify?

koox00:
> Now that marking as unread is working. If I mark the root post as unread from the thread view (either in RHS or Global Threads context), the new message line does not display under the root post as expected.

thanks @matthewbirtch! It should be fixed now.

matthewbirtch (approved):
Thanks @koox00! Looks like it is resolved

Willyfrog (approved):
LGTM

matthewbirtch (approved):
@koox00 I previously approved, but just tested one more time and it seems to be working quite well. Was there something new you wanted me to review again?

calebroseland (approved):
Great stuff!

calebroseland:
cc @jgilliam17 / @ogi-m

ogi-m:
@koox00 LGTM 🎉

New messages line appearing as expected:
- when opening an unread thread
- when new messages are received while viewing the thread
- when marking as unread
- tested in global threads and LHS

ogi-m (approved):

PR #8155, opened 2021-05-25
MM - 35835 show cc info in purchase modal

pablovelezvidal:
#### Summary
NOTE: temporarily base is MM-35435-... branch. since this work is a continuation of that one.

This PR adds the Credit card information on the first load of the purchase modal. It shows also an edit button which when clicked will display the payment form.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35835

#### Screenshots
Captura de pantalla 2021-05-27 a las 22 39 26

When clicking "Edit" button:
Captura de pantalla 2021-05-27 a las 22 40 58

#### Release Note

```
NONE
```


components/admin_console/billing/payment_info_details.tsx
```diff
@@ -0,0 +1,63 @@
...
+                    defaultMessage='Billing Address'
+                />
+            </div>
+            <div className='PaymentInfoDisplay__paymentInfo-address'>
+                <div>{address.line1}</div>
+                {address.line2 && <div>{address.line2}</div>}
+                <div>{`${address.city}, ${address.state}, ${address.postal_code}`}</div>
+                <div>{address.country}</div>
+            </div>
+            {children || null}
```
marianunez, line 58: If `children` is null, wouldn't it be the same as 
```suggestion
            {children}
```

components/admin_console/billing/payment_info_details.tsx
```diff
@@ -0,0 +1,63 @@
...
+import {GlobalState} from 'types/store';
+
+export interface PaymentInfoDetailsProps {
+    children?: React.ReactNode;
+}
+
+const PaymentInfoDetails: React.FC<PaymentInfoDetailsProps> = ({children}: PaymentInfoDetailsProps) => {
+    const paymentInfo = useSelector((state: GlobalState) => state.entities.cloud.customer);
+
+    if (!paymentInfo?.payment_method && !paymentInfo?.billing_address) {
```
marianunez, line 19: It seems confusing seeing in the same statement the non-null assertion `!` but at the same time doing the optional chaining in case it's null. Maybe just: 

```suggestion
    if (paymentInfo?.payment_method && paymentInfo?.billing_address) {
```

components/purchase_modal/purchase_modal.tsx
```diff
@@ -252,11 +264,32 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
-                        onInputChange={this.onPaymentInput}
-                        onCardInputChange={this.handleCardInputChange}
-                    />
+                    {this.state.editPaymentInfo ?
+                        <PaymentForm
+                            className='normal-text'
+                            onInputChange={this.onPaymentInput}
+                            onCardInputChange={this.handleCardInputChange}
+                        /> :
+                        <div className='paymentInfoDetails'>
```
marianunez, line 273: Should we be checking here if they have payment details in the state to show in saved payment mode?

marianunez (changes requested):
@pablovelezvidal as discussed today, we need to add which is the current plan they are in when they already have a payment method. Will you include this in this PR?

pablovelezvidal:
> @pablovelezvidal as discussed today, we need to add which is the current plan they are in when they already have a payment method. Will you include this in this PR?

yes, I will update this PR asap with those changes.

components/admin_console/billing/payment_info_details.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+import FormattedMarkdownMessage from 'components/formatted_markdown_message';
+import CardImage from 'components/payment_form/card_image';
+import {GlobalState} from 'types/store';
+
+export interface PaymentInfoDetailsProps {
+    children?: React.ReactNode;
+}
+
+const PaymentInfoDetails: React.FC<PaymentInfoDetailsProps> = ({children}: PaymentInfoDetailsProps) => {
```
nickmisasi, line 16: Nit: payment_info_details (PaymentInfoDetails), "info" and "details" mean basically the same thing. Can we shorten this to payment_info or payment_details? I think it would be a little nicer for readability

pablovelezvidal: ok.

components/admin_console/billing/payment_details.tsx
```diff
@@ -0,0 +1,63 @@
...
+import FormattedMarkdownMessage from 'components/formatted_markdown_message';
+import CardImage from 'components/payment_form/card_image';
+import {GlobalState} from 'types/store';
+
+export interface PaymentDetailsProps {
+    children?: React.ReactNode;
+}
+
+const PaymentDetails: React.FC<PaymentDetailsProps> = ({children}: PaymentDetailsProps) => {
+    const paymentInfo = useSelector((state: GlobalState) => state.entities.cloud.customer);
```
nickmisasi, line 17: paymentInfo as the var name seems a little weird because we're actually getting the customer here. I'm not really suggesting to change it to `customer`,  but maybe add a comment or something?

nickmisasi (approved):
One little suggestion but otherwise LGTM. Good stuff!

annelieseklein:
The current plan tag looks a little low maybe - is it aligned with that Starter plan?

components/admin_console/billing/payment_details.tsx
pablovelezvidal: makes sense, renaming it to customerPaymentInfo will give more context from where is this information coming

pablovelezvidal:
> The current plan tag looks a little low maybe - is it aligned with that Starter plan?

yes, I just pull the badge 5 pixels up and looks more aligned. Thoughts @annelieseklein ?

Captura de pantalla 2021-05-28 a las 0 01 07

annelieseklein:
looks good - thanks pablo!

annelieseklein:
Also the title should be 16px away from the first plan

annelieseklein (approved):

components/admin_console/billing/payment_info_details.tsx
marianunez: This is still pending, thoughts on this?

components/purchase_modal/purchase_modal.tsx
```diff
@@ -59,31 +63,51 @@ type State = {
     billingDetails: BillingDetails | null;
     cardInputComplete: boolean;
     processing: boolean;
-    selectedProduct: Product | null | undefined;
+    editPaymentInfo: boolean;
+    currentProduct: Product | null | undefined;
+    nextProductToUpgrade: Product | null | undefined;
```
marianunez, line 68: I believe it was more clear before as `selectedProduct` given that they may be able to downgrade from where they start the trial in Enterprise and downgrade to Starter for example.

```suggestion
    selectedProduct: Product | null | undefined;
```

marianunez (approved):
LGTM, minor comments from before and one below

components/admin_console/billing/payment_info_details.tsx
pablovelezvidal: the logic using the ! is for negating a condition, saying if NO payment method and NO billing address return null. But the thing is that before checking for the payment_method and the billing_address we need to make sure that the paymentInfo Object exists and is not null, otherwise JS will thrown an error because trying to access a property on a non existing object. The alternative to not using the ! in conjunction with the ? would be to do this same logic as:

`if (paymentInfo && (!paymentInfo.payment_method && !paymentInfo.billing_address)`

stevemudie (approved):
Will test post merge

AGMETEOR (approved):

components/admin_console/billing/payment_info_details.tsx
marianunez: That is the purpose of the [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) operator `?.` it will stop running if the object that is applied to is null or undefined so it's the same as the expression above.

PR #8156, opened 2021-05-25
Cleanup store initialization

lieut-data:
#### Summary
Rework the store initialization methods to lay the groundwork for upcoming side effect libraries. This PR should effectively be a no-op from the application's perspective.

#### Ticket Link
None.

```release-note
NONE
```


packages/mattermost-redux/src/store/configureStore.ts
```diff
@@ -1,33 +1,37 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {createStore, applyMiddleware, Store} from 'redux';
+import {createStore, applyMiddleware, Store, combineReducers} from 'redux';
+import thunk from 'redux-thunk';
+import {composeWithDevTools} from 'redux-devtools-extension/developmentOnly';
```
lieut-data, line 6: `developmentOnly` embeds the necessary `process.env.NODE_ENV` checks so that tree-shaking both disable and removes the unnecessary code.

At some point in the future, I'd love to consider something like `logOnlyInProduction` to give acecss to the tools without any performance overhead. But this PR is meant as a no-op :)

packages/mattermost-redux/src/store/configureStore.ts
```diff
@@ -39,5 +43,19 @@ export default function configureOfflineServiceStore(preloadedState: any, appRed
         persistConfig.persist(store, persistConfig.persistOptions, persistConfig.persistCallback);
     }
 
+    if (module.hot) {
```
lieut-data, line 46: This gets compiled out as well in a production build.

devinbinnie (approved):

packages/mattermost-redux/src/store/helpers.ts
```diff
@@ -1,28 +1,50 @@
...
+    const offlineReducer = (state: GlobalState, action: Action) => {
         if ('type' in action && 'data' in action && action.type === General.OFFLINE_STORE_RESET) {
             return baseReducer(action.data || baseState, action);
         }
 
-        return baseReducer(state, action as any);
+        return baseReducer(state, action as AnyAction);
+    };
+
+    return enableFreezing(enableBatching(offlineReducer as Reducer<GlobalState, Action>));
```
hmhealey, line 29: The `as Reducer<GlobalState, Action>` might not be necessary any more now that we're specifying `offlineReducer`'s arguments properly

packages/mattermost-redux/src/store/helpers.ts
```diff
@@ -1,28 +1,50 @@
...
+export function createReducer(baseState: GlobalState, ...reducers: Reducer[]): Reducer<GlobalState, Action> {
     reducerRegistry.setReducers(Object.assign({}, ...reducers));
     const baseReducer = combineReducers(reducerRegistry.getReducers());
 
     // Root reducer wrapper that listens for reset events.
     // Returns whatever is passed for the data property
     // as the new state.
-    function offlineReducer(state = {}, action: Action) {
+    const offlineReducer = (state: GlobalState, action: Action) => {
         if ('type' in action && 'data' in action && action.type === General.OFFLINE_STORE_RESET) {
```
hmhealey, line 22: At some point, it might be worth investigating how much of this is even needed any more. A lot of it was ported over from mobile where there was more logic required for the offline storage of stuff, and I know I cleaned up a lot of dead code when moving mattermost-redux into the web app. `OFFLINE_STORE_RESET` is referenced in the webapp's store initialization, but I don't know if that branch actually triggers at all.

hmhealey (approved):
Awesome to see this cleaned up further. I didn't know we could combine configureStore.dev.js and configureStore.prod.ts, but that's nice to know tree shaking lets us do that now to avoid that duplication

jgilliam17:
@lieut-data Is there anything specific that needs to be tested, beside the app starting? Thanks

lieut-data:
@jgilliam17, I'd recommend running e2e tests, and double checking an installation with plugins. (I've already tested with Incident Collaboration.)

jgilliam17:
Thank you @lieut-data 
Verified [PR E2E results](https://automation-dashboard.vercel.app/cycles/0c937da9-d0ec-42b8-ba39-e841fb6a4006) against [master E2E run](https://automation-dashboard.vercel.app/cycles/b76c2020-43b2-47a3-9564-4ab28acd4cbb)
4 test cases failed or timed out on the PR, but passed when testing manually
- Files and Attachment/channel files - passed
- Integrations/outgoing webhook overridden username and Profile pic (Using MM UI) - passed
- Team setting team spec/member can view and search members - passed 
- Toast/permalink post view combined with new message toast - passed 

There was an error when logging out (todo errors repro on master) may be related to these changes 
Screen Shot 2021-06-03 at 11 18 05 AM

lieut-data:
Thanks, @jgilliam17! The incident collaboration error looks like the same type as the todo plugin, and is probably something we should handle better in our product. My bet is that it's non-deterministic: can you confirm if you sometimes see it in `master`, or does it reliably only reproduce in this PR?

jgilliam17:
@lieut-data re-tested, my original sample size of two logouts was too small to catch it, it reproduces on master(and other test servers); not isolated to this PR.

jgilliam17 (approved):

PR #8159, opened 2021-05-26
MM-35346, MM-35394: Fix unread selectors/indicators

calebroseland:
#### Summary

The bulk of this PR is extending the patterns employed in #7733—conditionally using normal counts for CRT:OFF, or `_root`-suffixed selectors, where replies are excluded, for CRT:ON.
- `member.mention_count_root` OR  `member.mention_count`, and
- `channel.total_msg_count_root - member.msg_count_root` OR `channel.total_msg_count - member.msg_count`

Unread indicators affected: 

1. Favicon unread dot 

2. Mobile-view, LHS menu button: (unread dot or mention badge)
A. When in Global Threads:
✓ unread channels cause an unread indicator, but 
X unread threads should not cause an unread indicator.
B. When in a Channel:
✓ unread threads: cause an unread indicator, and
✓ other channels cause an unread indicator, but
X current channel should not cause an unread indicator

3. Multiple LHS selector codepaths (new and legacy sidebars) 
A. Unreads filter button, proper channel inclusion/exclusion.
B. Channel-bolding when it's unread.
C. '↑/↓  More unreads' banner for channels (_not_ Threads sidebar link, nor other changes required for `sidebar_channel_list`)

#### Breaking Changes:
1. `packages/mattermost-redux/src/selectors/entities/channels.ts`
  a. ~`getUnreads`~  → `getUnreadStatus`
  b. ~`getUnreadsInCurrentTeam`~ → `getUnreadStatusInCurrentTeam`
Instead of an object, these selectors now return a scalar/primitive `BasicUnreadStatus`—`bool` for existence of plain unreads, or `number` (>=1) for count of unread mentions.  `basicUnreadMeta(unreadStatus)` is used to decode into more usable pieces. 
`unreadMessages: number` was only ever used for `bool` comparisons, and was inconsistently computed prior to this, and `total_unread_threads` is grouped differently—by teams instead of channels.
`getUnreadStatus` was previously read in legacy sidebar, but it was not utilized in any render methods or logic, so it was removed. Now exclusively used for `favicon_title_handler`.
`getUnreadStatusInCurrentTeam` is exclusively used for the mobile-view account menu dot-indicator.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-35346
- https://mattermost.atlassian.net/browse/MM-35394

#### Related Pull Requests
- Continuation of, prior/similar changes: https://github.com/mattermost/mattermost-webapp/pull/7733

#### Release Note

```release-note
NONE
```


esethna:
@calebroseland is this ready for review from all parties? If so can you help add/ping reviews so we can keep the PRs flowing?

calebroseland:
> @calebroseland is this ready for review from all parties? If so can you help add/ping reviewers so we can keep the PRs flowing?

Was just doing final checks from my side before doing so 👍

hmhealey (commented):
I like the looks of this. Making the value into a primitive instead of an object makes a lot of sense, and having an extra helper to "decode" that is an excellent idea.

I'm not 100% sure if we want to mix types for it though. I wonder if using -1 as the special case instead of false would make this easier to use somehow since that's common as a special return value like with indexOf and such. With that, it could make more sense to have -1 be "no unreads", 0 be "unreads but no mentions", and positive numbers be "mentions", but that would require changing up the logic more than -1 being equivalent to the false value in the current code.

All that said, 0/5 about that either way. I like to avoid mixing types, but that's more personal preference. I'm also not really used to having access to union types either where that is properly supported by the language

calebroseland:
> I like the looks of this. Making the value into a primitive instead of an object makes a lot of sense, and having an extra helper to "decode" that is an excellent idea.
> 
> I'm not 100% sure if we want to mix types for it though. I wonder if using -1 as the special case instead of false would make this easier to use somehow since that's common as a special return value like with indexOf and such. With that, it could make more sense to have -1 be "no unreads", 0 be "unreads but no mentions", and positive numbers be "mentions", but that would require changing up the logic more than -1 being equivalent to the false value in the current code.
> 
> All that said, 0/5 about that either way. I like to avoid mixing types, but that's more personal preference. I'm also not really used to having access to union types either where that is properly supported by the language

I agree mixing types isn't the most ideal, and that's not a bad idea—`-1 | 0 | number`. My only hold up is that it doesn't match common truthiness checks, whereas with `false | true | number`, `Boolean(unreadStatus)` will plainly tell you if there are any unreads if you forgo the decoding helper.

esethna:
@hmhealey @calebroseland is this a blocker to approving and merging? @hmhealey we're trying to rally on PRs to get us to the beta release

koox00 (approved):

esethna:
@hmhealey can you help prioritize your review here?

hmhealey (approved):
Oh yeah, being able to use the truthiness like that is pretty nice. I'm not a huge fan of that type of conversion being implicit, but I've had to accept I lost that battle years ago by just using JavaScript 😛

esethna:
@jgilliam17 please help with a prioritized review on this one when you can

jgilliam17 (approved):
Thanks @calebroseland 
Tested, looks good to merge.
- Verified 
**MM-35346**
When unread filter is enabled threads category is not shown on the sidebar, unless it is bolded
Also when Unread channels are grouped separately and unread filter is ON, threads category is not shown on the sidebar, unless it is bolded
**MM-35394**
Verified unread dot in mobile-view 
a) while viewing Global Threads: Verified there is no unread dot on LHS hamburger menu when only threads are unread, but all channels are read.
b) while viewing any Channel: Verified Unread indicator is present when Threads and other channels are unread; unread indicator is not present when current channel is unread
Verified channel bolding, unread filter and LHS More unreads banner/toast all working as expected

esethna:
Sweet, let's merge! Thanks all

esethna:
@jwilander can you help with kicking the build and merging this one?

PR #8166, opened 2021-05-26
[MM-35236] Add 'gifsicle' module to devDependencies

jayaddison-collabora:
#### Summary
The [`gifsicle`](https://www.npmjs.com/package/gifsicle) module appears to be a build-time dependency for some release branches of Mattermost (including [v5.35](https://github.com/mattermost/mattermost-webapp/tree/release-5.35) as targeted by this pull request).

This changeset adds `gifsicle` as an NPM `devDependency` (development-only dependency not required for distribution with the application).

This attempts to resolve some build issues as reported in mattermost/mattermost-server#17420, mattermost/mattermost-server#17452, mattermost/mattermost-server#17521.

cc @hmhealey 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35236

#### Release Note
```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4415,8 +4415,7 @@
       "version": "0.7.0",
       "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.7.0.tgz",
       "integrity": "sha512-ONhaKPIufzzrlNbqtWFFd+jlnemX6lJAgq9ZeiZtS7I1PIf/la7CW4m83rTXRnVnsMbW2k56pGYu7AUFJD9Pow==",
-      "dev": true,
-      "optional": true
+      "dev": true
     },
     "@sinonjs/commons": {
       "version": "1.8.2",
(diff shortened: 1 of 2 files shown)


hmhealey:
Hi @jayaddison-collabora. Is this still an issue you're seeing with the latest master?

I fought with those errors for a while, but I couldn't seem to get it to install those dependencies correctly even if I added gifsicle as a dependency. I finally found out though that recreating the package-lock.json from scratch seemed to fix the issue, so I'm not sure if this is needed any more. I'm glad to take this PR if it is though.

jayaddison-collabora:
Thanks @hmhealey for taking a look!

I've spent a bit of time trying to reproduce the build issues again, using `npm ci --also=dev` (to try to respect the existing `package-lock.json` commitments) and also using `npm install --also=dev`, both on `master` as you suggest and also against `release-5.35` @ 52c518439cd20ad484fad1ff1521c4bc433c6d7b (as configured for this pull request).

The results are:

- NPM CI, `release-5.35`: no repro
- NPM Install, `release-5.35`: no repro
- NPM CI, `master`: no repro
- NPM Install, `master`: no repro

(this was all using NPM v12.21.0 and NPM v7.5.2, FWIW)

So my best guess is that this was a `gifsicle`-related dependency change external to Mattermost that was a problem for a certain window of time during those issue reports.

As a result: no need to include / merge this change; I think it's harmless, but it's also understandable to keep the declared dependencies to a minimum.

jayaddison-collabora:
(note that I'm doing a little more work locally to see if I can find the exact workflow -- it was build/packaging-related, if I remember correctly - that led me to encounter this issue and provide this suggestion.  I'll report back if & when I discover that)

Willyfrog:
thanks for the hard work, please let us know anything you discover! 😃

hmhealey:
The problem also showed up specifically with Node 16/NPM 7 for me. I was switching between versions a lot when we were preparing to upgrade though, so I don't remember when exactly I started seeing it.

PR #8189, opened 2021-06-01
MM-34188 Add basic framework and example for tracking memoization effectiveness

jwilander:
#### Summary
This is an experiment for tracking how effective the memoization in our redux selectors is. The idea is to track the number of times a selector is called compared against how often a recomputation of the selector result takes place. It does not take into account whether the recomputation was necessary or not. That could potentially be a future improvement but it would be more heavyweight and could affect app performance.

Cypress is leveraged to create "benchmark" tests for certain scenarios where we want to test the selector performance. In this PR I created one example of this for posting into an empty channel. I'd see us wanting to add these benchmark tests for many of our primary use cases or anywhere that we've observed impacting performance.

Two reasons why I went with Cypress for the benchmark tests:

1. It can run against a production build and mimics a real user fairly closely
2. This could be expanded to incorporate other performance tracking measurements, such as how often a component re-renders in React

In order to see how the effectiveness of our selectors is impacted by our code changes we need to track not just the current measurement data but also the previous data. The approach I take here is similar to jest snapshots, where the results of the benchmark tests are stored in a `__benchmarks__` directory. I imagine we could then add something to CI to compare the new results versus the ones on the `master` branch, posting the results on the PR and potentially even blocking merge if the diff is egregious.

This PR is still very much work in progress but I wanted to start getting everyone's thoughts on the approach. Let me know what you think :)

Screenshot of what the output data looks like:
Screen Shot 2021-06-01 at 5 24 29 PM

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34188

#### Release Note
```release-note
NONE
```


packages/mattermost-redux/src/selectors/entities/posts.ts
```diff
@@ -264,6 +264,7 @@ function formatPostInChannel(post: Post, previousPost: Post | undefined | null,
 
 export function makeGetPostsInChannel(): (state: GlobalState, channelId: $ID<Channel>, numPosts: number) => PostWithFormatData[] | undefined | null {
     return createSelector(
+        'makeGetPostsInChannel',
```
jwilander, line 267: Unfortunately I couldn't find an alternative to passing in the function names here. I explored `Function.caller` and other related options but none of the reasonable ones I found worked in strict mode.

packages/mattermost-redux/src/selectors/createSelector.js
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// dumpTrackedSelectorsStatistics prints to console a table containing the measurement data on all tracked selectors.
+function dumpTrackedSelectorsStatistics() {
+    const selectors = getSortedTrackedSelectors();
+    console.table(selectors); //eslint-disable-line no-console
+}
+
+// Only track selectors in development builds.
+if (process.env.NODE_ENV === 'development') { //eslint-disable-line no-process-env
```
enahum, line 145: should there be a flag other than development mode to enable this?

Willyfrog (commented):
added a few minor comments

in general I like the idea, but maybe we can make it a bit simpler, since we are only interested in memoization, we could probably just [substitute the memoization function](https://github.com/reduxjs/reselect#q-the-default-memoization-function-is-no-good-can-i-use-a-different-one) so all other behaviour stays the same.

We could probably do even better on only applying this custom changes if webpack detects a variable so most of the time any changes doesn't affect the production build.

e2e/cypress/benchmark/__benchmarks__/Message -- Current user posting message in empty channel.json
```diff
@@ -0,0 +1,122 @@
+[
```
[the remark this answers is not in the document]
jwilander: Agreed, I think having them be CI artifacts makes more sense than committing them

jwilander:
re @Willyfrog 
>We could probably do even better on only applying this custom changes if webpack detects a variable so most of the time any changes doesn't affect the production build.

Right now the cypress tests only run against production builds. We could introduce a new pipeline only for these benchmark tests and build a version of the webapp that includes the changes otherwise we need to include them in the production build

packages/mattermost-redux/src/selectors/createSelector.js
jwilander: Probably, yeah

jwilander:
re @Willyfrog 

>in general I like the idea, but maybe we can make it a bit simpler, since we are only interested in memoization, we could probably just substitute the memoization function so all other behaviour stays the same.

I took a look at simplifying it to just pass in a custom memoization function but you can only pass options on creation of the createSelector function and not on every invocation of the createSelector function. Without that we can't pass the selector name into createSelector and identifying which selector is which becomes harder. So even if we moved more tracking login into the memoization function itself we'd still need to modify `createSelectorCreator` directly I think.

If someone has a way around that or if I'm looking at it wrong let me know

packages/mattermost-redux/src/selectors/entities/posts.ts
hmhealey: Yeah, I fought with that for a while, but I couldn't find any way to do it automatically either. With all of the weird pseudo-metaprogramming things you can do in JavaScript, I'm surprised that doesn't exist at all.

hmhealey (commented):
Since we can't do anything automatic, I like having something we can just swap in when we need it, but I wonder if we should give it a different name to make it easier to grep for? As it is, it'll be relatively hard to find which files use it. We'll also have to swap out createSelector for an entire file which will be a bit of a pain for some of the larger files (like `selectors/entities/channels`)

jwilander:
@hmhealey agree on the naming but I was actually thinking of having it present everywhere all the time in all the selectors so we can monitor all the selector memoization and not just do spot checking

e2e/cypress/benchmark/__benchmarks__/Message -- Current user posting message in empty channel.json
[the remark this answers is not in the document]
saturninoabril: I ran `message_spec.js` multiple times and I'm getting varying results. The difference is not high but might be significant enough when comparing benchmark results. So maybe, we should run the spec in CI multiple times before saving the final benchmark result for comparison/baseline.

e2e/cypress/benchmark/message_spec.js
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Stage: @prod
```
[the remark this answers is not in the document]
saturninoabril: Yes, please remove. We'll put it back (promote to prod) once we find the test stable for 5 consecutive daily runs.

e2e/cypress/benchmark/message_spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+describe('Message', () => {
+    let testTeam;
+    let testChannel;
+
+    before(() => {
+        // # Create new team and new user and visit Town Square channel
+        cy.apiInitSetup({loginAfter: true}).then(({team, channel}) => {
+            testTeam = team;
+            testChannel = channel;
```
saturninoabril, line 23: I don't see `testChannel` used elsewhere.

e2e/package.json
```diff
@@ -59,5 +59,8 @@
     "test": "start-test start:webhook 3000 test:ci",
     "test:ci": "node run_tests.js",
     "uniq-meta": "grep -r \"^// $META:\" cypress | grep -ow '@\\w*' | sort | uniq"
+  },
+  "dependencies": {
+    "tablemark": "^2.0.0"
```
saturninoabril, line 64:
```suggestion
    "tablemark": "2.0.0"
```

saturninoabril (changes requested):
Thanks @jwilander!

Please update https://github.com/mattermost/mattermost-webapp/blob/master/e2e/utils/file.js#L12 to:
```diff
- const TEST_DIR = 'cypress/integration';
+ const TEST_DIR = 'cypress';
```
So that test/s under `cypress/benchmark` will be considered in CI.

Note: [Prod tests](https://mm-cypress-report.s3.amazonaws.com/314381579-b33f0553-webapp-pr-8189-mattermost-enterprise-edition:3a08aab/mochawesome.html) - Passed. Other failed tests were known issues.

packages/mattermost-redux/src/selectors/createSelector.js
```diff
@@ -0,0 +1,155 @@
...
+        selector.resetCalls();
+        selector.resetRecomputations();
+    });
+}
+
+// getSortedTrackedSelectors returns an array, sorted by effectivness, containing mesaurement data on all tracked selectors.
+function getSortedTrackedSelectors() {
+    let selectors = Object.values(trackedSelectors);
+    selectors = selectors.map((selector) => ({name: selector.selectorName(), effectiveness: selector.effectiveness(), recomputations: selector.recomputations(), calls: selector.calls()}));
+    selectors.sort((a, b) => (a.effectiveness > b.effectiveness) ? 1 : ((b.effectiveness > a.effectiveness) ? -1 : 0))
```
lieut-data, line 134: nit: I think this could be simplified to:
```suggestion
    selectors.sort((a, b) => a.effectiveness - b.effectiveness))
```

lieut-data (approved):

e2e/cypress/benchmark/message_spec.js
```diff
@@ -0,0 +1,51 @@
...
+            cy.visit(`/${testTeam.name}/channels/town-square`);
+        });
+    });
+
+    it('Current user posting message in empty channel', () => {
+        cy.wait(1000);
```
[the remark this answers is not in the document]
devinbinnie: Okay, I think you can extend the timeout on those calls but I don't remember how.

e2e/cypress/benchmark/__benchmarks__/Message -- Current user posting message in empty channel.json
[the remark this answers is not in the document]
jwilander: Yeah, running it multiple times might be a good idea. I'm primarily hoping to capture large changes in effectiveness so a little variation shouldn't be a problem. Either way for now I'm going to add `__benchmarks__` to the gitignore and then in a future PR explore using CI artifacts to track the differences

e2e/cypress/benchmark/message_spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+            cy.getLastPostId().then((postId) => {
+                const divPostId = `#postMessageText_${postId}`;
+
+                // * Check that the message was created
+                cy.get(divPostId).find('p').should('have.text', 'One');
+
+                reportBenchmarkResults(cy, win);
+            });
+        });
```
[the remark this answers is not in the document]
jwilander: Agreed, will add some more benchmarks in a follow-up PR

jwilander:
I'll be pushing a commit shortly addressing a bunch of the feedback, thanks all

hmhealey (approved):

jwilander:
Thanks for the reviews all! I've made some updates that cleans up the implementation a little bit more and adds tracking for all our selectors. Please give it another look.

@saturninoabril would you be able to run the E2E tests again?

saturninoabril:
E2E currently running - https://automation-dashboard.vercel.app/cycles/95c82b33-5cb6-4ecc-b2fd-0fdfb38256d1

jwilander:
Storybook is running into a build error because it can't resolve the `reselect` package. Anyone know what I need to do get it to resolve for storybook? It resolves fine for everywhere else.

packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+/* eslint-disable max-lines */
```
lieut-data, line 4: We should just disable this rule altogether :p

lieut-data (approved):

saturninoabril:
Overall looks good to me. However, I'm requesting Joseph to review and to validate [test report](https://github.com/mattermost/mattermost-webapp/pull/8189#issuecomment-859558719) since I'm still out this week.

saturninoabril (approved):

josephbaylon (approved):
LGTM. Retested failed tests that are not known issues from this [test report](https://automation-dashboard.vercel.app/cycles/95c82b33-5cb6-4ecc-b2fd-0fdfb38256d1) and those passed for me locally.

jwilander:
Thanks again, all! I'll be aiming to merge this tomorrow. I'll need to announce that any new uses of `createSelector` after this goes in will need to pass the function name in as a string for the first argument. I'll update with master in and do one last sweep for any new `createSelector` uses before merging.

packages/reselect/src/index.d.ts
```diff
@@ -0,0 +1,1108 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
devinbinnie, line 1: Do we need this file if we're leaving the `reselect` library overrides in JS?

devinbinnie (approved):

packages/reselect/src/index.d.ts
hmhealey: It's presumably needed to stop TS from complaining about us calling `createSelector` incorrectly since we added an extra name parameter to its result

hmhealey:
I'm not opposed to doing it this way, but something about replacing the implementation of the library transparently doesn't feel great to me, especially since it changes the function signatures. I can't think of anything that's actually a problem with doing it this way, but it does feel like it could be confusing in the future. I think I'd feel a lot better about it if the name was passed automatically as well, but that seems to be impossible without something more complicated like some sort of code generation.

packages/reselect/src/index.d.ts
jwilander: Yes, typescript was complaining if I didn't have it unfortunately

jwilander:
@hmhealey I could change the `reselect` package to be named something like `reselect-with-tracking` to make it more clear that it's different. Would that make it better?

jwilander:
I chatted with @hmhealey and we'll leave this as is for now but changing the package name or something else to make it less transparent in the future might be something we do.

PR #8195, opened 2021-06-02
MM-36172 - purchase modal empty address issue

pablovelezvidal:
#### Summary
There was a situation where the purchase modal was trying to show the payment information in the case where the customer has not it set already, which leaded with a component showing empty strings separated by commas. This PR makes sure that when the customer has no payment information, the fill in payment method form is shown.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36172

#### Release Note
```release-note
NONE
```

### DIFF
--- a/utils/cloud_utils.ts
+++ b/utils/cloud_utils.ts
@@ -2,7 +2,6 @@
 // See LICENSE.txt for license information.
 
 import {CloudCustomer} from 'mattermost-redux/types/cloud';
-import {BillingDetails} from 'types/cloud/sku';
 
 export function isCustomerCardExpired(customer?: CloudCustomer): boolean {
     if (!customer) {
--- a/utils/cloud_utils.ts
+++ b/utils/cloud_utils.ts
@@ -2,6 +2,7 @@
 // See LICENSE.txt for license information.
 
 import {CloudCustomer} from 'mattermost-redux/types/cloud';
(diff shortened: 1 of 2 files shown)


stevemudie (approved):
Will test post merge

utils/cloud_utils.ts
```diff
@@ -20,3 +21,17 @@ export function isCustomerCardExpired(customer?: CloudCustomer): boolean {
...
+    }
+    return Boolean(
+        billingDetails?.address &&
+        billingDetails?.city &&
+        billingDetails?.state &&
+        billingDetails?.country &&
+        billingDetails?.postalCode &&
+        billingDetails.name,
+    );
+}
```
marianunez, line 37: I believe we already have a method for this here: https://github.com/mattermost/mattermost-webapp/blob/2f0cc730068c208d53b0e91c2c885db6b4e02a23/types/cloud/sku.ts#L23-L37

marianunez:
@pablovelezvidal @stevemudie just a reminder that we do have the label to setup the test server for the PR connected to CWS for this type of changes where we can do an initial test before merging when applicable.

utils/cloud_utils.ts
pablovelezvidal: got it.

nickmisasi (approved):

PR #8197, opened 2021-06-03
Fix several autocomplete problems

larkox:
#### Summary
The problems addressed are:
- Interactive dialogs and apps forms not getting the focus when executed from a slash command in the center channel. It did work on the RHS. Solved by moving the focus code before the command execution, so the interactive dialog can steal the focus when created.
- Crash when a command is executed by pressing the tab button to select the "Execute command" option on the RHS. Solved by applying the check on persist.
- Inconsistent updates on the draft of the RHS when submitting a command by hitting enter and therefore selecting the "Execute command" option. We just handle the word when it is not the case where the command is finished.
- Inconsistent updates on the draft when a command returned an error on the RHS. Solved by reapplying the draft on command errors.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34726

#### Related Pull Requests
None

#### Screenshots
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/actions/views/create_comment.jsx
+++ b/actions/views/create_comment.jsx
@@ -163,7 +163,12 @@ export function makeOnSubmit(channelId, rootId, latestPostId) {
         if (isReaction && emojiMap.has(isReaction[2])) {
             dispatch(submitReaction(latestPostId, isReaction[1], isReaction[2]));
         } else if (message.indexOf('/') === 0 && !options.ignoreSlash) {
-            await dispatch(submitCommand(channelId, rootId, draft));
+            try {
+                await dispatch(submitCommand(channelId, rootId, draft));
+            } catch (err) {
+                dispatch(updateCommentDraft(rootId, draft));
+                throw err;
+            }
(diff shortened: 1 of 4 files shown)


hmhealey (approved):

hanzei (approved):
LGTM

DHaussermann (approved):
Tested and passed. I tested the following on both main and RHS input as well as Browser and desktop
 
- When modal opens, focus goes to the 1st available field for input (tested from main and RHS)
- Tabbing-out ti the execute option does not cause any CSS errors (tested from main and RHS)
- For the last 2 points regarding drafts:
-- Successfully run commands are always cleared fro the input and never left as drafts
-- When trying to run and invalid subcommand, pressing enter will select the warning (or parsing issue) to clear the invalid portion
-- Forcing an invalid subcommand to run with the command/control key will show a validation error to the UI and clear the input

Huge thanks @larkox for this PR. This adds huge value to the usability for these cases!

PR #8199, opened 2021-06-03
Update NOTICE.txt

amyblais:
Have asked ``sass`` about adding a license https://github.com/sass/sass/issues/3073.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -2912,9 +2912,30 @@ This product contains 'sass' by Sass.
 Sass is an extension of CSS, adding nested rules, variables, mixins, selector inheritance, and more.
 
 * HOMEPAGE:
-  * https://github.com/sass/sass
+  * https://github.com/sass/dart-sass
 
-* LICENSE:
+* LICENSE: MIT
+
+Copyright (c) 2016, Google Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
+permit persons to whom the Software is furnished to do so, subject to
+the following conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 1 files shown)


NOTICE.txt
```diff
@@ -2807,6 +2905,19 @@ SOFTWARE.
...
+## sass
+
+This product contains 'sass' by Sass.
+
+Sass is an extension of CSS, adding nested rules, variables, mixins, selector inheritance, and more.
+
+* HOMEPAGE:
+  * https://github.com/sass/sass
+
+* LICENSE:
```
amyblais, line 2917: @agnivade I've requested the product owners to add a license to the repo: https://github.com/sass/sass/issues/3073. Not sure if you are able to check if they already have a license that just hasn't been added as a license file.

agnivade: Nope, they don't have a license. But other repos under the same org do have though. :man_shrugging:

amyblais: @hmhealey Is this the correct repo for sass https://github.com/sass/sass? They don't list an open-source license, so I'm not sure if we're allowed to use the repo. cc @coreyhulen

agnivade: Maybe @jamiehurewitz will be able to help here.

amyblais: @hmhealey If the ``sass`` repo owners don't respond, can https://github.com/mattermost/mattermost-webapp/pull/8056 be reverted from v5.36?

agnivade: Interesting .. apparently they had an MIT License at some point - https://github.com/sass/sass/issues/2026. But now it does not seem to be there. I cannot track the commit where it was removed since this appears to have been in a different branch.

Hmm .. no wait, that repo isn't the sass implementation. It just contains the sass spec. The real implementation is in https://github.com/sass/dart-sass or https://github.com/sass/libsass. 

@hmhealey - Are we actually using this repo's code in our codebase?

hmhealey: The NPM page for the library ([link](https://www.npmjs.com/package/sass)) links to https://github.com/sass/dart-sass as the repository for the code we're using. sass/sass seems to be for the spec/proposal that the various implementations of sass (dart-sass, libsass, and the old node-sass library that we used to use). 

> @hmhealey - Are we actually using this repo's code in our codebase?

It's used during compilation to transpile sass into plain CSS.

amyblais: I've updated the link and license to https://github.com/sass/dart-sass / MIT.

agnivade (approved):

PR #8204, opened 2021-06-04
Made comment component react to change in external changes in draft

harshilsharma63:
#### Summary
Made comment component react to change in external changes in draft.

This change is to support the working of my plugin https://github.com/harshilsharma63/mattermost-plugin-reply

#### Ticket Link
NA

#### Related Pull Requests
None

#### Screenshots
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/create_comment/create_comment.jsx
+++ b/components/create_comment/create_comment.jsx
@@ -250,7 +250,7 @@ class CreateComment extends React.PureComponent {
             };
         } else {
             // else, initialize the draft
-            updatedState.draft = {...props.draft, caretPosition: props.draft.message.length, uploadsInProgress: []};
+            updatedState.draft = state.draft || {...props.draft, caretPosition: props.draft.message.length, uploadsInProgress: []};
         }
 
         const rootChanged = props.rootId !== state.rootId;
--- a/components/create_comment/create_comment.jsx
+++ b/components/create_comment/create_comment.jsx
@@ -250,7 +250,7 @@ class CreateComment extends React.PureComponent {
             };
         } else {
             // else, initialize the draft
-            updatedState.draft = {...props.draft, caretPosition: props.draft.message.length, uploadsInProgress: []}
+            updatedState.draft = {...props.draft, caretPosition: props.draft.message.length, uploadsInProgress: []};
         }
 
         const rootChanged = props.rootId !== state.rootId;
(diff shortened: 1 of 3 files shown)


hanzei:
@harshilsharma63 Heads up that a couple of tests are failing: https://app.circleci.com/pipelines/github/mattermost/mattermost-webapp/39092/workflows/0c2b75d5-f15e-4c58-8332-3f0196550144/jobs/266201

harshilsharma63:
@hanzei yes I'm working on it.

components/create_comment/create_comment.jsx
```diff
@@ -240,9 +240,19 @@ class CreateComment extends React.PureComponent {
             createPostErrorId: props.createPostErrorId,
             rootId: props.rootId,
             messageInHistory: props.messageInHistory,
-            draft: state.draft || {...props.draft, caretPosition: props.draft.message.length, uploadsInProgress: []},
         };
 
+        if (state.draft && props.draft && props.draft.message && state.draft.message !== props.draft.message) {
```
michelengelen, line 245: Can't we utilize new ES features here? Just asking, since this would be much simpler when we could use 'state.draft?.message !== props.draft?.message'

michelengelen (approved):
I don't know exactly, but I think the whole if-else block could be simplified with optional object spreading. Just an assumption though, this works as well

LGTM

components/create_comment/create_comment.jsx
```diff
@@ -240,9 +240,19 @@ class CreateComment extends React.PureComponent {
...
+            // if we already have a draft, update it's message if draft exist and is changed
+            updatedState.draft = {
+                ...state.draft,
+                message: props.draft.message,
```
larkox, line 249: Should we worry about he caret position?

components/create_comment/create_comment.jsx
larkox, line 245: I wonder about the UX experience here. This opens the door to modifying the input of the user while they are typing (really nice joke plugin).

larkox (approved):
LGTM, just two small considerations.

Willyfrog (commented):
I wonder if after modifying the draf it should:
- update  `this.draftsForPost[this.props.rootId]=<draft>`
- call `this.props.onUpdateCommentDraft(<draft>);`

PR #8205, opened 2021-06-04
MM-35405: hides reply notifications user setting

koox00:
#### Summary

When CRT is on reply notifications user setting makes no sense as it is.
This commit hides that section.
Later on a setting about threads will be added on desktop notifications section.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35405

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/user_settings/notifications/index.js
+++ b/components/user_settings/notifications/index.js
@@ -6,6 +6,7 @@ import {bindActionCreators} from 'redux';
 
 import {updateMe} from 'mattermost-redux/actions/users';
 import {getConfig} from 'mattermost-redux/selectors/entities/general';
+import {isCollapsedThreadsEnabled} from 'mattermost-redux/selectors/entities/preferences';
 
 import UserSettingsNotifications from './user_settings_notifications.jsx';
 
@@ -18,6 +19,7 @@ function mapStateToProps(state) {
     return {
         sendPushNotifications,
         enableAutoResponder,
+        isCollapsedThreadsEnabled: isCollapsedThreadsEnabled(state),
     };
 }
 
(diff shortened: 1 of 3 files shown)


jwilander (approved):
LGTM - other than the linter failing

esethna:
@jwilander @koox00 I confirmed on daily that the setting **does in fact take effect**. ie if you set to "Trigger notifications on messages in reply threads that I start or participate in" or  "Trigger notifications on messages in threads that I start" then you **get a mention badge for every reply on your followed threads** (not just direct @mentions).

Given that, when a user enables CRT, do we need to set this setting to "Do not trigger notifications on messages in reply threads unless I'm mentioned" AND hide it so we get the proper mention badge behaviour

jwilander:
Ok, I think it would be better to handle that completely server-side rather than trying to mess with changing the user's settings from the client

jwilander:
So I think this PR is ok to go in as is but we'll need a follow-up ticket to make the change server-side to ignore the setting when CRT is enabled

esethna:
Okay cool, so @ashishbhate I assume that will be included with this ticket? ^ https://mattermost.atlassian.net/browse/MM-35415

ashishbhate:
>Okay cool, so @ashishbhate I assume that will be included with this ticket? ^ https://mattermost.atlassian.net/browse/MM-35415

Yep

anurag6713 (approved):

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified "Reply notifications" setting has been removed from Account settings when CRT is ON.

PR #8208, opened 2021-06-04
MM-36166 Force recalculate emoji picker position on open

jwilander:
#### Summary
Add the `show` as an input to the memoized functions for determining emoji picker position.

@hmhealey I know you mentioned we had done this in other places differently but I couldn't find any good examples and this worked well enough. Feel free to point out a better way to do it though.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36166

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/emoji_picker/emoji_picker_overlay.jsx
+++ b/components/emoji_picker/emoji_picker_overlay.jsx
@@ -45,8 +45,13 @@ export default class EmojiPickerOverlay extends React.PureComponent {
         enableGifPicker: false,
     };
 
-    emojiPickerPosition = memoize((emojiTrigger) => {
+    emojiPickerPosition = memoize((emojiTrigger, show) => {
         let calculatedRightOffset = Constants.DEFAULT_EMOJI_PICKER_RIGHT_OFFSET;
+
+        if (!show) {
+            return calculatedRightOffset;
+        }
+
         if (emojiTrigger) {
             calculatedRightOffset = window.innerWidth - emojiTrigger.getBoundingClientRect().left - Constants.DEFAULT_EMOJI_PICKER_LEFT_OFFSET;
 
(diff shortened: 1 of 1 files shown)


Willyfrog (approved):

amyblais:
@hmhealey @jgilliam17 This is for v5.36.

jgilliam17:
Thanks @jwilander 
The emoji picker position looks good on the main channel view.
Seeing few issues on RHS
1. On Firefox: Emoji picker opens over the RHS textbox when at least one reply is present
Screen Shot 2021-06-07 at 12 55 53 PM
2. All browsers: On a thread with enough posts to push the comment textbox to the bottom, emoji picker opens slightly over the comment textbox
Screen Shot 2021-06-07 at 1 04 49 PM

jwilander:
Thanks @jgilliam17 do those issues reproduce on 5.35 and master too? Wondering if it's a new issue from my PR or not

jgilliam17:
@jwilander yes, I can reproduce on master and 5.35. It's something Harrison described in repro steps 2 in the ticket.

jgilliam17:
Actually I went back to 3.52 (before your original PR 8092 was merged) and the overlap on RHS reproduces there too. Hope that helps.

jgilliam17:
Splitting RHS issues from this ticket as I can reproduce few versions back https://mattermost.atlassian.net/browse/MM-36261

hmhealey:
I think the first screenshot you posted is correct actually since there's not really room above or below the emoji button for the picker. The second still looks off, but that's totally fine for a separate ticket since it's not a regression.

jwilander:
Any issues I need to address here @jgilliam17 or are you good to approve?

jgilliam17:
No issues. Looks good to merge.

jgilliam17 (approved):

hmhealey (approved):
I would've expected to see this happening in the parent component that controls the emoji picker since that's a bit cleaner (because it doesn't add side effects to render), but this should be fine regardless.

PR #8209, opened 2021-06-04
MM-36068 Update avatars and reply count correctly for unfollowed threads

jwilander:
#### Summary
Update avatars and reply count correctly for unfollowed threads.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36068

#### Release Note
```release-note
NONE
```

### DIFF
--- a/packages/mattermost-redux/src/reducers/entities/posts.ts
+++ b/packages/mattermost-redux/src/reducers/entities/posts.ts
@@ -13,6 +13,7 @@ import {
     PostOrderBlock,
     MessageHistory,
 } from 'mattermost-redux/types/posts';
+import {UserProfile} from 'mattermost-redux/types/users';
 import {Reaction} from 'mattermost-redux/types/reactions';
 import {
     $ID,
@@ -283,6 +284,19 @@ function handlePostReceived(nextState: any, post: Post) {
     // Delete any pending post that existed for this post
     if (post.pending_post_id && post.id !== post.pending_post_id && nextState[post.pending_post_id]) {
         Reflect.deleteProperty(nextState, post.pending_post_id);
+    }
+
+    const rootPost: Post = nextState[post.root_id];
+    if (post.root_id && rootPost) {
+        const participants = rootPost.participants || [];
+        const nextRootPost = {...rootPost};
+        if (!participants.find((user: UserProfile) => user.id === post.user_id)) {
+            nextRootPost.participants = [...participants, {id: post.user_id}];
+        }
+
+        nextRootPost.reply_count++;
+
+        nextState[post.root_id] = nextRootPost;
     }
 
     return nextState;
--- a/packages/mattermost-redux/src/reducers/entities/posts.ts
+++ b/packages/mattermost-redux/src/reducers/entities/posts.ts
@@ -294,7 +294,9 @@ function handlePostReceived(nextState: any, post: Post) {
             nextRootPost.participants = [...participants, {id: post.user_id}];
         }
 
-        nextRootPost.reply_count++;
+        if (post.reply_count) {
+            nextRootPost.reply_count = post.reply_count;
+        }
 
         nextState[post.root_id] = nextRootPost;
     }
--- a/packages/mattermost-redux/src/reducers/entities/threads.ts
+++ b/packages/mattermost-redux/src/reducers/entities/threads.ts
@@ -49,6 +49,14 @@ export const threadsReducer = (state: ThreadsState['threads'] = {}, action: Gene
     }
     case ThreadTypes.FOLLOW_CHANGED_THREAD: {
         const {id, following} = action.data;
+
+        // Delete local copy of thread when unfollowing
+        if (!following && state[id]) {
+            const nextState = {...state};
+            Reflect.deleteProperty(nextState, id);
+            return nextState;
+        }
+
         return {
             ...state,
             [id]: {
--- a/packages/mattermost-redux/src/reducers/entities/threads.ts
+++ b/packages/mattermost-redux/src/reducers/entities/threads.ts
@@ -3,10 +3,12 @@
 
 import {combineReducers} from 'redux';
 
-import {TeamTypes, ThreadTypes, UserTypes} from 'mattermost-redux/action_types';
+import {PostTypes, TeamTypes, ThreadTypes, UserTypes} from 'mattermost-redux/action_types';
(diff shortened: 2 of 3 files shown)


packages/mattermost-redux/src/reducers/entities/threads.ts
```diff
@@ -49,6 +49,14 @@ export const threadsReducer = (state: ThreadsState['threads'] = {}, action: Gene
     }
     case ThreadTypes.FOLLOW_CHANGED_THREAD: {
         const {id, following} = action.data;
+
+        // Delete local copy of thread when unfollowing
+        if (!following && state[id]) {
+            const nextState = {...state};
+            Reflect.deleteProperty(nextState, id);
+            return nextState;
+        }
```
jwilander, line 58: Without deleting the thread on unfollow the reply count and participants wouldn't update anymore locally since the thread selector was returning the unfollowed thread that was not being updated instead of creating a synthetic thread.

jwilander:
Couldn't find any tests that could be easily extended to cover this case. I think it's something that would be suited to a cypress test rather than a jest or snapshot test but I don't think those have been set up yet for CRT (something we should definitely do).

jgilliam17 (approved):
Thanks @jwilander 
Tested, looks good to merge.
- Verified avatars and reply count are updated correctly for unfollowed threads ( also tested during threading sync today)
Please ping me to retest if any changes are requested. 🙂

packages/mattermost-redux/src/reducers/entities/threads.ts
calebroseland: // threadsReducer

// case PostTypes.RECEIVED_NEW_POST
// if newPost.root_id && state[newPost.root_id] && !state[newPost.root_id].is_following
// if ...!participants.find...
// updatedThread.participants = [...]

packages/mattermost-redux/src/reducers/entities/posts.ts
```diff
@@ -285,6 +286,21 @@ function handlePostReceived(nextState: any, post: Post) {
...
     }
 
+    const rootPost: Post = nextState[post.root_id];
+    if (post.root_id && rootPost) {
+        const participants = rootPost.participants || [];
+        const nextRootPost = {...rootPost};
+        if (!participants.find((user: UserProfile) => user.id === post.user_id)) {
+            nextRootPost.participants = [...participants, {id: post.user_id}];
+        }
+
```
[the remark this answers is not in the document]
ashishbhate: Would it be possible to handle removal similar to how we do it here in a reducer?

From the deleted post's rootId we could find the threadId and then update the participants if the deleted post was the only post by the user.

packages/mattermost-redux/src/reducers/entities/threads.ts
jwilander: I'd prefer to clear out stale data to handle it but let me see if it causes issues while viewing the thread in the global view. If it does then maybe it's easiest to do it in the threadsReducer

packages/mattermost-redux/src/reducers/entities/posts.ts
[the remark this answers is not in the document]
jwilander: Unfortunately @ashishbhate, we can't get the thread state in the posts reducer so the only way would be to loop through every single post to check if it's root_id is the same. I don't think that would be too bad since deletes shouldn't happen all that often.

Either way I think that's a for a different ticket, not this one.

packages/mattermost-redux/src/reducers/entities/threads.ts
calebroseland: > I'd prefer to clear out stale data to handle it but let me see if it causes issues while viewing the thread in the global view. If it does then maybe it's easiest to do it in the threadsReducer

Thinking about this again—if Global Threads (thread_item, thread_pane, etc.) switched to using `getThreadOrSynthetic` (instead `getThread`), it would probably work to clear stale data as you have it.

packages/mattermost-redux/src/reducers/entities/posts.ts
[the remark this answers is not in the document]
calebroseland: > Either way I think that's a for a different ticket, not this one.

👍

ashishbhate (approved):
LGTM. The finer points of the implementation I'll defer to @calebroseland

packages/mattermost-redux/src/reducers/entities/threads.ts
jwilander: Ok I'll check that

jwilander: @calebroseland turned out to be a lot more involved to use `getThreadOrSynthetic` and to keep the deletion of the thread on unfollow. Pushed a commit to instead go with your suggestion of adding the new post login to the threads reducer

jwilander:
@jgilliam17 can you give this another quick test? Particularly around following, then un-following a thread to see if the reply count/avatars still update correctly. Also please try unfollowing a thread in the global view to make sure that still works ok.

jgilliam17 (approved):
@jwilander Looks good to merge
Just few notes on the something I observed during testing
1. I can follow and unfollow threads- avatar and reply count update in real time for both. There is an issue with avatars rearranging after post is followed - but I believe Caleb added an issue for that already; repros on daily.
2. When unfollowing a thread on global view, I am seeing one issue where clicking on the reply button in the footer to open unfollowed thread in the RHS - follows the thread again. I'm not able to repro on daily, but Ogi can, so I'll file that separately if we don't have the ticket already.

packages/mattermost-redux/src/reducers/entities/threads.ts
```diff
@@ -57,6 +59,27 @@ export const threadsReducer = (state: ThreadsState['threads'] = {}, action: Gene
             },
         };
     }
+    case PostTypes.RECEIVED_NEW_POST: {
+        const post: Post = action.data;
+        const thread: UserThread | undefined = state[post.root_id];
+        if (post.root_id && thread) {
```
calebroseland, line 65:
```suggestion
        if (post.root_id && thread && !thread.is_following) {
```
If we check for not following, I think it would cut down on the number of double-updates as Followed threads additionally receive `SocketEvents.THREAD_UPDATED` which includes updated participants and reply_count.

calebroseland (approved):

PR #8216, opened 2021-06-06
Migrate 'components/profile_popover' module and associated tests to T…

AshishDhama:
#### Summary
**Migrate 'components/profile_popover' module and associated tests to TypeScript**
Fixes [https://github.com/mattermost/mattermost-server/issues/13478](https://github.com/mattermost/mattermost-server/issues/13478)

#### Ticket Link
[https://mattermost.atlassian.net/browse/MM-20595](https://mattermost.atlassian.net/browse/MM-20595)

#### Related Pull Requests
None

#### Screenshots
Screenshot 2021-06-06 at 6 57 55 PM

#### Release Note
```release-note
- Renamed the files to their associated TypeScript extensions
- Updated components using it to the correct imports
- Added utility function for removing keys from the object.
```


components/profile_popover/index.ts
```diff
@@ -24,15 +24,17 @@ import {areTimezonesEnabledAndSupported} from 'selectors/general';
...
+import ProfilePopover from './profile_popover';
 
-function getDefaultChannelId(state) {
+function getDefaultChannelId(state: GlobalState) {
     const selectedPost = getSelectedPost(state);
     return selectedPost.exists ? selectedPost.channel_id : getCurrentChannelId(state);
 }
 
-function mapStateToProps(state, {userId, channelId = getDefaultChannelId(state)}) {
+function mapStateToProps(state: GlobalState, {userId, channelId = getDefaultChannelId(state)}: {userId: string; channelId?: string}) {
```
devinbinnie, line 37: Nit: usually we define an actual type called `OwnProps` for the second parameter.
Non-blocking though.

components/profile_popover/profile_popover.tsx
```diff
@@ -16,263 +13,270 @@ import {browserHistory} from 'utils/browser_history';
...
+        closeModal: (modalId: string) => Promise<{
+            data: boolean;
+        }>;
+        openDirectChannelToUserId: (userId?: string) => Promise<any>;
+        getMembershipForEntities: (teamId: any, userId: any, channelId: any) => Promise<any>;
+    };
+    intl: IntlShape;
+}
+type ProfilePopoverState = {
+    loadingDMChannel: string | undefined;
```
devinbinnie, line 140: Nit: shortform
```suggestion
    loadingDMChannel?: string;
```

components/profile_popover/profile_popover.tsx
```diff
@@ -16,263 +13,270 @@ import {browserHistory} from 'utils/browser_history';
...
+    enableTimezone: boolean;
+    actions: {
+        openModal: (modalData: {modalId: string; dialogType: any; dialogProps?: any}) => Promise<{
+            data: boolean;
+        }>;
+        closeModal: (modalId: string) => Promise<{
+            data: boolean;
+        }>;
+        openDirectChannelToUserId: (userId?: string) => Promise<any>;
+        getMembershipForEntities: (teamId: any, userId: any, channelId: any) => Promise<any>;
```
devinbinnie, line 135: Most ids are `string`

components/profile_popover/profile_popover.tsx
```diff
@@ -16,263 +13,270 @@ import {browserHistory} from 'utils/browser_history';
...
+     * The overwritten username that should be shown at the top of the popover
+     */
+    overwriteName?: React.ReactNode;
+
+    /**
+     * @internal
+     */
+    enableTimezone: boolean;
+    actions: {
+        openModal: (modalData: {modalId: string; dialogType: any; dialogProps?: any}) => Promise<{
```
devinbinnie, line 128: `dialogType` here should be `React.ComponentType`

components/profile_popover/profile_popover.tsx
```diff
@@ -16,263 +13,270 @@ import {browserHistory} from 'utils/browser_history';
...
-
-        if (this.state.loadingDMChannel !== -1) {
+        if (this.state.loadingDMChannel !== undefined) {
             return;
         }
-
         this.setState({loadingDMChannel: user.id});
-
-        actions.openDirectChannelToUserId(user.id).then((result) => {
+        actions.openDirectChannelToUserId(user.id).then((result: any) => {
```
devinbinnie, line 187: We can type `result` here with an `error` field

components/profile_popover/profile_popover.tsx
```diff
@@ -16,263 +13,270 @@ import {browserHistory} from 'utils/browser_history';
...
+
+    /**
+     * Source URL from the image to display in the popover
+     */
+    src: string;
+
+    /**
+     * Source URL from the image that should override default image
+     */
+    overwriteIcon?: React.ReactNode;
```
devinbinnie, line 42: Shouldn't this also be a `string`?

components/profile_popover/profile_popover.tsx
```diff
@@ -400,45 +380,42 @@ class ProfilePopover extends React.PureComponent {
...
         }
-
         if (this.props.user.is_bot && !haveOverrideProp) {
             dataContent.push(
                 <div
                     key='bot-description'
                     className='overflow--ellipsis text-nowrap'
                 >
-                    {this.props.user.bot_description}
+                    {this.props.user.bot_description as ReactNode}
```
devinbinnie, line 389: I don't think we need this cast here.

components/profile_popover/profile_popover.tsx
```diff
@@ -400,45 +380,42 @@ class ProfilePopover extends React.PureComponent {
...
                 </div>,
             );
         }
-
         if (this.props.user.position && !haveOverrideProp) {
-            const position = this.props.user.position.substring(0, Constants.MAX_POSITION_LENGTH);
+            const position = (this.props.user?.position as string).substring(
```
devinbinnie, line 394: Instead of a potentially dangerous cast, we can coalesce to a blank string.
```suggestion
            const position = (this.props.user?.position || '').substring(
```

components/profile_popover/profile_popover.tsx
```diff
@@ -400,45 +380,42 @@ class ProfilePopover extends React.PureComponent {
...
-                    >
+                    <div className='overflow--ellipsis text-nowrap pt-1 pb-1'>
                         {position}
                     </div>
                 </OverlayTrigger>,
             );
         }
-
-        const email = this.props.user.email;
+        const email = this.props.user.email as string;
```
devinbinnie, line 411: Same here wrt. coalescing to blank string.

components/profile_popover/profile_popover.tsx
```diff
@@ -656,6 +674,6 @@ class ProfilePopover extends React.PureComponent {
     }
 }
 
-delete ProfilePopover.propTypes.id;
+// delete ProfilePopover.propTypes.id;
```
devinbinnie, line 677: We can just remove this line.

devinbinnie (changes requested):
Thanks for the PR @AshishDhama!

components/profile_popover/profile_popover.tsx
AshishDhama: @devinbinnie what about Promise<any> where can I look for it's type

components/profile_popover/profile_popover.tsx
AshishDhama: @devinbinnie Converting to React.ComponentType throwing error of the wrong type. I can see everywhere we have used as any only. Could please share any reference to check

components/profile_popover/profile_popover.tsx
AshishDhama: result: {error: any}

components/profile_popover/profile_popover.tsx
AshishDhama: Done

components/profile_popover/profile_popover.tsx
AshishDhama: Done

components/profile_popover/profile_popover.tsx
AshishDhama: Done

components/profile_popover/profile_popover.tsx
AshishDhama: Done

components/profile_popover/profile_popover.tsx
devinbinnie: Can you try `React.Component`?

components/profile_popover/profile_popover.tsx
devinbinnie: I think `error` is usually of type `ServerError`

components/profile_popover/profile_popover.tsx
AshishDhama: Screenshot 2021-06-07 at 8 55 33 PM
That's why I added them as ReactNode but yeah it does make sense to keep them as a string

AshishDhama: Done

components/profile_popover/profile_popover.tsx
AshishDhama: Done

components/profile_popover/profile_popover.tsx
devinbinnie: In this case we can actually return `void` or `Promise<void>` because this component doesn't use the result of the promise.

components/profile_popover/index.ts
AshishDhama: Done

components/profile_popover/profile_popover.tsx
AshishDhama: Yeah I did

components/profile_popover/profile_popover.tsx
AshishDhama: Done

components/profile_popover/profile_popover.tsx
AshishDhama: Done

components/profile_popover/profile_popover.tsx
AshishDhama: Have checked to assign Type as React.Component or React.ComponentType throw type error as we are assign object of different type inside class

hmhealey: I wouldn't be surprised if there's some weird typing issues going on here, but it looks like we use `React.Component` in some places
https://github.com/mattermost/mattermost-webapp/blob/master/components/sidebar/sidebar_category/sidebar_category_menu/index.ts#L28 
and any in others
https://github.com/mattermost/mattermost-webapp/blob/master/components/sidebar/sidebar_category/sidebar_category_menu/sidebar_category_menu.tsx#L26

Notably though, it seems like we use `React.Component` only in index files currently and `any` in the components themselves, so we might be hitting some weirdness with our redux typing.

components/profile_popover/profile_popover.tsx
AshishDhama: Conflict with other components if we put there type as string

components/profile_popover/profile_popover.tsx
AshishDhama: Screenshot 2021-06-07 at 9 33 57 PM

AshishDhama: @hmhealey @devinbinnie This is the type getting inferred by typescript of the UsersettingModal

components/profile_popover/profile_popover.tsx
AshishDhama: @hmhealey @devinbinnie please suggest what should be the course of action here

Willyfrog (changes requested):
minor request for safety, but LGTM

utils/utils.jsx
```diff
@@ -2254,3 +2267,10 @@ export function stringToNumber(s) {
 
     return parseInt(s, 10);
 }
+
+export function deleteKeysFromObject(value, keys) {
+    for (const key of keys) {
+        delete value[key];
```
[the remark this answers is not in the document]
AshishDhama: @Willyfrog delete will return true even if the key doesn't exist on value. Just double-checked with [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete), So I think we don't need try-catch here. Thanks for the suggestion

components/profile_popover/profile_popover.tsx
devinbinnie: Given the above comments, we can just keep this as `any` for now, though it would be worth looking into why this doesn't work.

AshishDhama: Okay

AshishDhama: I will check if I can find some good reason

components/profile_popover/profile_popover.tsx
AshishDhama: My main concern is that in one Component we are expecting string and in one React.ReactNode can use the Union type of React.ReactNode and String and then cast them accordingly.

devinbinnie: Well, `overwriteIcon` should always be a `string` since it's used as the `src` attribute for an `img` tag. As for `overwriteName`, I think it can stay as `ReactNode` because `UserProfile` expects that you can pass any arbitrary HTML into it as an override. So we can probably also change the code referenced [here](https://github.com/mattermost/mattermost-webapp/pull/8216#discussion_r646703351) to:
`title = this.props.overwriteName`

Willyfrog (approved):

components/profile_popover/profile_popover.tsx
AshishDhama: Type of overwriteIcon is ReactNode in UserProfile is ReactNode, We are using UserProfile at a couple more place so changing that to string doesn't feel like the right choice to me without assessing the risk of that.
Screenshot 2021-06-09 at 2 49 24 PM
Screenshot 2021-06-09 at 2 49 05 PM

One solution comes to my mind as overwriteIcon is optional we can do something like this in UserProfile
```
<ProfilePopover
 className='user-profile-popover'
 userId={userId}
channelId={channelId}
src={profileImg}
isBusy={isBusy}
hide={this.hideProfilePopover}
hideStatus={hideStatus}
isRHS={isRHS}
hasMention={hasMention}
overwriteName={overwriteName}
overwriteIcon={typeof overwriteIcon === 'string' ? overwriteIcon : ''}
 />
```
I think after this we don't need to change the type of overwriteIcon inside UserProfile and we can keep the type safety of typescript as well.

devinbinnie: The root of the issue here is that `overwriteIcon` is incorrectly defined in `UserProfile`, it should always be a `string`. The only usage of `overwriteIcon` is passed into `ProfilePopover` (which does specify that `overwriteIcon` is a string) which is then passed into `Avatar` (also expecting a `string`) which passes it into the `img` tag. I don't see a lot of risk here in changing that definition since it's wrong. I know it increases the scope a bit here but it's definitely the most correct thing to do. We shouldn't need to be doing runtime type checking for stuff like this.

components/profile_popover/profile_popover.tsx
devinbinnie: We can do `{error: ServerError}`

AshishDhama: Okay Thanks

components/profile_popover/profile_popover.tsx
AshishDhama: Great thanks will change it there as well.

devinbinnie (approved):
Thanks @AshishDhama, this is looking good now.

stevemudie (approved):
Did general smoke test around popovers including DMs etc.

Worked as expected. Looks good!

PR #8070, opened 2021-05-05
[34475] Emoji 13.0

Willyfrog:
#### Summary

Update emojis to 13.0, there is a new category and some of the old ones have been renamed

[emoji 13.0 changes](https://emojipedia.org/emoji-13.0/)

#### Ticket Link

[34475](https://mattermost.atlassian.net/browse/MM-34475)

#### Screenshots
Screen Shot 2021-05-06 at 11 25 35
Screen Shot 2021-05-06 at 11 25 44
Screen Shot 2021-05-06 at 11 25 55
Screen Shot 2021-05-06 at 11 26 15

#### Release Note

```release-note
updated emojis and emoji categories to standard 13.0
```

#### Additional Notes
there are a few emojis failing, my guess is that most of those reuse the image from another one. The list of emojis failing are:

| png | link | name |
|--|--|--|
| 1f9d1-200d-1f91d-200d-1f9d1 | https://emojipedia.org/people-holding-hands/ | People holding hands
| 1f469-200d-2764-fe0f-200d-1f48b-200d-1f468 |https://emojipedia.org/emoji/%F0%9F%91%A9%E2%80%8D%E2%9D%A4%EF%B8%8F%E2%80%8D%F0%9F%92%8B%E2%80%8D%F0%9F%91%A8/ | Kiss: Woman, Man |
| 1f469-200d-2764-fe0f-200d-1f468 | https://emojipedia.org/emoji/%F0%9F%91%A9%E2%80%8D%E2%9D%A4%EF%B8%8F%E2%80%8D%F0%9F%91%A8/ | Couple with Heart: Woman, Man |
| 1f468-200d-1f469-200d-1f466 | https://emojipedia.org/emoji/%F0%9F%91%A8%E2%80%8D%F0%9F%91%A9%E2%80%8D%F0%9F%91%A6%E2%80%8D%F0%9F%91%A6/ | Family: Man, Woman, Boy, Boy |
| 2640-fe0f | https://emojipedia.org/emoji/%E2%99%80/ | Female Sign |
| 2642-fe0f | https://emojipedia.org/emoji/%E2%99%82/ | Male Sign |
| 2695-fe0f | https://emojipedia.org/emoji/%E2%9A%95%EF%B8%8F/ | medical sign |
| 0023-fe0f-20e3 | https://emojipedia.org/emoji/%23%EF%B8%8F%E2%83%A3/ | Keycap Number Sign | 
[...]

PR #8220, opened 2021-06-07
[34475] Emoji 13.0 with new library

Willyfrog:
#### Summary
this is a rework of https://github.com/mattermost/mattermost-webapp/pull/8070 but using a new library, so closing that one.

support for skin tones (including multiple skin tones) and emojis up to 13.0

Note that the generated files are in the first commit so you can select the rest of changes for ease of review.

#### Ticket Link
[34475](https://mattermost.atlassian.net/browse/MM-34475)

#### Related Pull Requests
- Has server changes https://github.com/mattermost/mattermost-server/pull/17741

#### Screenshots
Screen Shot 2021-06-07 at 15 25 45

#### Release Note
```release-note
support for emoji standard 13.0
```


components/emoji_picker/emoji_picker.jsx
```diff
@@ -39,88 +38,59 @@ const EMOJI_TO_LOAD_PER_UPDATE = 135;
...
+//     'people & body': createCategory('people & body', 'fa fa-user', 'People & Body', t('emoji_picker.people-body')),
+//     'animals & nature': createCategory('animals & nature', 'fa fa-leaf', 'Animals & Nature', t('emoji_picker.animals-nature')),
+//     'food & drink': createCategory('food & drink', 'fa fa-cutlery', 'Food & Drink', t('emoji_picker.food-drink')),
+//     activities: createCategory('activities', 'fa fa-futbol-o', 'Activities', t('emoji_picker.activities')),
+//     'travel & places': createCategory('travel & places', 'fa fa-plane', 'Travel & Places', t('emoji_picker.travel-places')),
+//     objects: createCategory('objects', 'fa fa-lightbulb-o', 'Objects', t('emoji_picker.objects')),
+//     symbols: createCategory('symbols', 'fa fa-heart-o', 'Symbols', t('emoji_picker.symbols')),
+//     flags: createCategory('flags', 'fa fa-flag-o', 'flags', t('emoji_picker.flags')),
+//     custom: createCategory('custom', 'fa fa-at', 'custom', t('emoji_picker.custom')),
+// };
```
jwilander, line 81: Commented out for use later?

packages/mattermost-redux/src/client/client4.ts
```diff
@@ -2651,7 +2651,8 @@ export default class Client4 {
     };
 
     getSystemEmojiImageUrl = (filename: string) => {
-        return `${this.url}/static/emoji/${filename}.png`;
+        const extension = filename.endsWith('.png') ? '' : '.png';
```
jwilander: Is there an inconsistency with some having the extension and others not?

jwilander (approved):
LGTM - just a couple clarifying questions

packages/mattermost-redux/src/client/client4.ts
Willyfrog: no, but the old files used to not have the extension, I can remove this if you prefer

jwilander: 0/5, just wanted to understand the context

components/emoji_picker/emoji_picker.jsx
Willyfrog: no, I was comparing the new and the old one and left it behind inadvertently, I'll remove it.

Willyfrog:
@hmhealey do you know if there is something I'm missing for the build to work properly?

hmhealey:
It looks like the package-lock.json got messed up again so that it doesn't install the dependencies of mozjpeg right. Last time I fixed it, it was by deleting the package-lock and recreating it from scratch.

build/emoji/emoji_gen.mjs
```diff
@@ -0,0 +1,273 @@
...
+
+function filename(emoji) {
+    return emoji.image.split('.')[0];
+}
+
+function writeFile(fileName, filePath, data) {
+    const promise = Fs.writeFile(filePath, data, writeOptions);
+
+    promise.then(() => {
+        console.log(`${fileName} generated successfuly.`);
```
hmhealey, line 83: Total nitpicking time
```suggestion
        console.log(`${fileName} generated successfully.`);
```

build/emoji/emoji_gen.mjs
```diff
@@ -0,0 +1,273 @@
...
+        return [];
+    }
+    let order = nextOrder;
+    return Object.keys(emoji.skin_variations).map((skinCode) => {
+        // if skin codes ever change this will produce a null_light_skin_tone
+        const skins = skinCode.split('-');
+        const skinShortName = skins.map((code) => skinCodes[code]).join('_');
+        const skinName = skins.map((code) => skinNames[code]).join(', ');
+        const variation = {...emoji.skin_variations[skinCode]};
+        variation.short_name = `${emoji.short_name}_${skinShortName}`;
```
hmhealey, line 103: What are the changes to the format of the emoji.json file? I think Matterhorn uses a copy of it to do their emoji replacement, so I want to be able to let them know what changes there'll be before we actually ship them

hmhealey (changes requested):

esethna:
@Willyfrog just a comment, the picker will only one version of each emoji based on the skin tone selected, correct? Assuming this is just showing multiple versions in the PR description because the UI changes for the skin tone picker haven't been done yet?

matthewbirtch:
> @Willyfrog just a comment, the picker will only one version of each emoji based on the skin tone selected, correct? Assuming this is just showing multiple skin tone versions of each emoji in the PR description because the UI changes for the skin tone picker haven't been done yet?

I wondered the same

Willyfrog:
@esethna @matthewbirtch exactly. Next PR will address the new emojis picker and the skin tones selector

build/emoji/emoji_gen.mjs
Willyfrog: thanks!

build/emoji/emoji_gen.mjs
Willyfrog: `name` is not the usual `my_emoji`, but an angry representation of the same one `MY EMOJI`, for that you should use `short_name`
there is no `alias` but `short_names`, which includes the canonical `short_name` in its first position
instead of filename there is `image` which also contains the file extension

if there are skin variations, it has a field `skin_variations` which contains the needed info for that variation, but in our script we convert those to full emojis and add them (in case they are using `utils/emoji.json`, if they use the one provided by the library you can disregard this), I also aded a skins property which is an array of skin codes, I use the first one (if there are more than one) as the one to filter by

there are other fields included that we are not using, but it might be useful to someone so I left them, you can find more info in the [lib's readme](https://github.com/iamcal/emoji-data#using-the-data)

esethna (approved):

matthewbirtch:
> @esethna @matthewbirtch exactly. Next PR will address the new emojis picker and the skin tones selector

Ok, so what should I review from a UX perspective? Other than the emojis themselves, it seems like there aren't any other UI updates

Willyfrog:
Categories changed (names and there is a new one) and I just wanted to make sure that it was ok with the symbols for them.

build/emoji/emoji_gen.mjs
hmhealey: Cool, I'll let them know then. It seems like the two big changes that would probably affect them are renaming aliases to short_names and possibly the change to filename, but I can't imagine either of those are that big

Willyfrog:
@matthewbirtch seems the makefile was preventing the build

matthewbirtch:
@Willyfrog a few questions:

1. It seems like all the emojis aren't loading properly?

image

2. I don't see the 'Recently used' category for emojis. Is that removed?

3. For the category icons, do we have access to use the compass icons? If so, I'm wondering if we can use icons from that set instead, but let me know if that should be for another PR or wait until we revamp the whole picker UI.

Willyfrog:
For the first one, it might be a network error? I can see them in a cellphone
image

2. Recently used category should appear once you post an emoji as it gets stored in the local store. 


3. I think we can use those if the compass icons are already in use in the webapp. Since I'm already working on the redesign, we can try and add those to the next PR, just DM me the icons you'd like to have for each category

matthewbirtch:
> For the first one, it might be a network error? I can see them in a cellphone

Not sure about that. I'm on a pretty stable, fast internet connection, but seem to be getting broken images for these quite often. I notice it most when I scroll down more or click categories beyond the first two.
 
> 2. Recently used category should appear once you post an emoji as it gets stored in the local store.

Yep, makes sense.

> 3. I think we can use those if the compass icons are already in use in the webapp. Since I'm already working on the redesign, we can try and add those to the next PR, just DM me the icons you'd like to have for each category

I sent you a DM with the compass icon names.

4. I noticed another issue. When I try to react to a message using the emoji pickjer, it doesn't add the reaction when I click an emoji from that context (see GIF below).

Willyfrog:
uhm, the problem seems to be that it is loading as if it were custom emojis, I'll take a look at how to fix that, thanks!

components/emoji_picker/emoji_picker.test.jsx
```diff
@@ -132,14 +132,12 @@ describe('components/emoji_picker/EmojiPicker', () => {
         expect(wrapper.find('.emoji-picker__items').prop('style')).toStrictEqual({overflowY: 'hidden'});
         expect(wrapper.find(EmojiPickerCategorySection).length).toBe(2);
 
-        jest.runOnlyPendingTimers();
-        wrapper.update();
-        jest.runOnlyPendingTimers();
+        setTimeout(() => {
```
Willyfrog, line 135: i'm not really liking this change, but seems that now initial perf is worse by having a two categories first and then moving into having all of them due to only having 1 bigger spritesheet instead of multiple smaller ones.

Willyfrog:
@matthewbirtch now it should load all the emojis using the spritesheet (so no more rejected request), thanks for pointing that out as I didn't notice in my local setup (I don't have a limit there 😆 )

jwilander (approved):

matthewbirtch:
@Willyfrog looks like most of the issues I found are resolved. The only outstanding one is the icon update. Were you wanting to include that here or in another PR?

hmhealey (approved):

Willyfrog:
> @Willyfrog looks like most of the issues I found are resolved. The only outstanding one is the icon update. Were you wanting to include that here or in another PR?

since we plan on doing a redesign, I was planning on implementing all the design changes in the next PR

matthewbirtch (approved):

jgilliam17 (approved):
Thanks @Willyfrog 
Tested, looks good to merge.
- Verified emoji picker new category for People and Body, and other category names - looks good.
- There was some slowness on initial load and search is not as quick on the emoji picker and post emoji autocomplete; although the speed improved as the testing went on and all emoji loaded. 
One other issue - lack of clearance between some emoji when they are part of the same post, but different rows. Post appears to be a bit crowded. If concerning, this can be address in the design change PR you created for follow up.

PR #8222, opened 2021-06-07
MM-35837 Improve batching of channel-related actions on post receipt

hmhealey:
One of the causes of slowdowns when receiving a post is that a lot of components re-render due to changes in `state.entities.channels.channels` (which happens whenever any channel changes). A lot of components depend on this object, particularly the sidebar.

Because of how the post receiving actions are written, we dispatch a few rounds of actions as we're updating various fields containing unread information. It's easier to see this when we add an artificial delay between the batches since you'll see the re-render flicker happen a few times as everything happens.

https://user-images.githubusercontent.com/3277310/121036766-ec581600-c77c-11eb-9071-d16045a921d1.mov

By making the code a bit clunkier, we remove an extra re-render and cut a fair bit of time off the render.

https://user-images.githubusercontent.com/3277310/121038000-e9a9f080-c77d-11eb-9aa7-8939307e3f25.mov

I compared the time taken by each `conn.onmessage` call done when the websocket event for the new post was received, and while the times were a bit inconsistent run to run (I blame my Macbook's thermal throttling), there seemed to be about a 40-50% decrease in time taken when viewing a channel with only a few posts.

Screen Shot 2021-06-07 at 10 48 03 AM

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35837

#### Release Note
```release-note
Reduced time taken re-rendering when a post is received
```

### DIFF
--- a/actions/new_post.ts
+++ b/actions/new_post.ts
@@ -1,12 +1,17 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import * as Redux from 'redux';
 import {batchActions} from 'redux-batched-actions';
 
 import {isCollapsedThreadsEnabled} from 'mattermost-redux/selectors/entities/preferences';
 
 import {
+    actionsToMarkChannelAsRead,
+    actionsToMarkChannelAsUnread,
+    actionsToMarkChannelAsViewed,
     markChannelAsRead,
+    markChannelAsReadOnServer,
     markChannelAsUnread,
     markChannelAsViewed,
 } from 'mattermost-redux/actions/channels';
@@ -47,18 +52,17 @@ export function completePostReceive(post: Post, websocketMessageProps: NewPostMe
                 return result;
             }
         }
+        const actions: Redux.AnyAction[] = [];
 
         if (post.channel_id === getCurrentChannelId(getState())) {
-            dispatch({
+            actions.push({
                 type: ActionTypes.INCREASE_POST_VISIBILITY,
                 data: post.channel_id,
                 amount: 1,
             });
         }
 
-        // Need manual dispatch to remove pending post
-
-        const actions = [
+        actions.push(
             PostActions.receivedNewPost(post, isCollapsedThreadsEnabled(state)),
             {
                 type: WebsocketEvents.STOP_TYPING,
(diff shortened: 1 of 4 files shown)


hmhealey:
As a followup to this, I'd like to look at why exactly everything re-renders when `state.entities.channels.channels` changes. Some things like the outer components of the sidebar re-rendering makes sense since the unread state changing could affect sort order and which DMs are displayed, but each sidebar item appears to be re-rendering which is concerning. That could also tie to how the performance issues seem to have been exacerbated by the new sidebar.

I'm also still unsure of how longer threads tie into this except for long threads just having poorer performance to begin with.

actions/new_post.ts
```diff
@@ -68,57 +69,58 @@ export function completePostReceive(post: Post, websocketMessageProps: NewPostMe
...
-    return (dispatch: DispatchFunc, getState: GetStateFunc) => {
-        const state = getState();
-        const currentUserId = getCurrentUserId(state);
-
-        // ignore system message posts, except when added to a team
-        if (shouldIgnorePost(post, currentUserId)) {
-            return {data: false};
+// setChannelReadAndViewed returns an array of actions to mark the channel read and viewed, and it dispatches an action
+// to asynchronously mark the channel as read on the server if necessary.
+export function setChannelReadAndViewed(dispatch: DispatchFunc, getState: GetStateFunc, post: Post, websocketMessageProps: NewPostMessageProps, fetchedChannelMember: boolean): Redux.AnyAction[] {
```
hmhealey, line 83: While I'm explicitly avoiding extra dispatches here by passing around dispatch and getState, I don't want to get in a habit of doing that. I think it hurts readability, and I don't think it should make a performance difference in most places. The big problem seems to be if we get into a pattern like this where we make multiple changes to the same object in a row.

```javascript
const fooA = () => (dispatch) => {
    dispatch({type: 'CHANGE_A_THING'});
    dispatch(fooB());
};

const fooB = () => (dispatch) => {
    dispatch({type: 'CHANGE_A_THING'});
    dispatch(fooC());
};

const fooC = () => (dispatch) => {
    dispatch({type: 'CHANGE_A_THING'});
};
```

In this example, if fooA, fooB, and fooC all touched different objects which impact different parts of the UI, any performance hit would be much less noticeable.

jwilander (approved):
LGTM, nice work 👍 I wonder if there's other places where we could do more batching and less individual dispatching

jwilander:
> As a followup to this, I'd like to look at why exactly everything re-renders when `state.entities.channels.channels` changes. Some things like the outer components of the sidebar re-rendering makes sense since the unread state changing could affect sort order and which DMs are displayed, but each sidebar item appears to be re-rendering which is concerning. That could also tie to how the performance issues seem to have been exacerbated by the new sidebar.
> 
> I'm also still unsure of how longer threads tie into this except for long threads just having poorer performance to begin with.

@hmhealey my [branch here](https://github.com/mattermost/mattermost-webapp/commits/mm-35509) did some work to prevent re-rendering of a lot of the top-level components when `state.entities.channels.channels` changes. I didn't do much with the sidebar though.

jwilander:
@hmhealey got some conflicts and a test failing too

Willyfrog (approved):
LGTM

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Short smoke test; compared [PR E2Es](https://automation-dashboard.vercel.app/cycles/97334927-bc7a-4ec5-a9a4-8d055b56933a) to [results from master](https://mm-cypress-report.s3.amazonaws.com/317466769-bd7ca2e-master-e20-prod-mattermost-enterprise-edition:master/mochawesome.html) 
Re-tested manually few failed tests that differ from master - passing, no new issues.

PR #8224, opened 2021-06-07
Allow uploading `.jpeg` files on Linux

neallred:
#### Summary
Allow uploading JPEG files ending in `.jpeg` on Chromium on Linux (may also enable other platforms).

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/17728

#### Screenshots
Before:
file-types-before

file-types-after

#### Release Note
```release-note
Fixed ability to upload `.jpeg` files on Linux
```

### DIFF
--- a/components/__snapshots__/picture_selector.test.tsx.snap
+++ b/components/__snapshots__/picture_selector.test.tsx.snap
@@ -5,7 +5,7 @@ exports[`components/picture_selector should match snapshot, default picture prov
   className="PictureSelector"
 >
   <input
-    accept=".jpg,.png,.bmp"
+    accept=".jpeg,.jpg,.png,.bmp"
     aria-hidden={true}
     className="PictureSelector__input hidden"
     data-testid="PictureSelector__input-picture_selector_test"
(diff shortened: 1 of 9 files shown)


devinbinnie (approved):
Thanks @neallred!

hmhealey (approved):
Thanks for the PR! Looks good to me. I forgot that that component filtered accepted values based on file extension.

I also had to go look up why there's both jpg and jpeg file extensions. As usual, it's because DOS and old versions of Windows couldn't handle 4 character file extensions, so the different extension just stuck and is still in use almost 30 years later.

hmhealey:
@jgilliam17 This will apply the same on all OSes, so you can just rename an image to have a .jpeg extension. I'm not sure if it'll vary in behaviour by browser, but I assume it would work the same way on each one

neallred:
> because DOS and old versions of Windows couldn't handle 4 character file extensions, so the different extension just stuck and is still in use almost 30 years later.

No way! I always wondered why there were two extensions for it.

jgilliam17 (approved):
Thank you @neallred 
Tested, looks good to merge.
- Verified profile picture .jpeg files can be uploaded on Linux, and other OSs.

PR #8225, opened 2021-06-07
MM-36095 Fix text overflow in announcement banner

hmhealey:
I moved around some elements to get text overflow working in the announcement banner. I don't know how or when it would've worked before because the text-overflow property wasn't on the element that actually contained the text.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36095

#### Screenshots
Screen Shot 2021-06-07 at 3 33 35 PM

#### Release Note

```release-note
- Fix incorrect text alignment and wrapping in announcement banner
```


e2e/cypress/integration/system_console/site_configuration/announcement_banner_spec.js
```diff
@@ -102,14 +103,13 @@ describe('Announcement Banner', () => {
             contains('a', bannerEmbedLink).
             should('have.attr', 'href', bannerEmbedLink);
 
-        // * Verify only the banner text's first part is visible
-        // and check the complete text length spans more than viewport width (also is hidden as per above)
+        // * Verify only the banner text's first part is visible by ensuring the first link
```
hmhealey, line 106: I had to change this test because the paragraph is now what is truncating the text, not the element that contains the paragraph, so the paragraph is now constrained by the screen while its contents are wider. Since I don't think we can measure arbitrary element contents, I think it works just as well if we check that a child at the start is visible and a child at the end isn't.

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
```diff
@@ -134,7 +134,7 @@ export default class AnnouncementBar extends React.PureComponent<Props> {
                     placement='bottom'
                     overlay={announcementTooltip}
                 >
-                    <span>
+                    <div className='announcement-bar__text'>
```
hmhealey, line 137: This isn't strictly necessary, but I wanted to fix the type (since spans can't contain a lot of the elements that show up in Markdown text) and use a more specific selector than just `> span`

michelengelen (approved):
LGTM

pablovelezvidal (approved):

amyblais:
@jgilliam17 This is for v5.36.

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good tot merge.
- Verified banner long text is properly truncated and vertically aligned.

amyblais:
@hmhealey Please help manually cherry-pick to ``release-5.36`` and ``cloud``.

PR #8228, opened 2021-06-08
Export PurchaseModal component and related utils

agarciamontoro:
#### Summary
In order to support Cloud in Incident Collaboration, we need to implement the upgrade flow triggered when users try to use a locked feature.

To keep the experience consistent, we are using the `PurchaseModal` component in two ways:
- When the user is interacting with the Messaging product (in an incident in a channel, for example), and the modal needs to be shown, we leverage the existing architecture by simply calling `openModal`, which pops up the already mounted component.
- When the user is in the backstage, there is no component already mounted, so we are mounting another one in our components hierarchy, and using the same `openModal` logic to show it.

This PR makes all this possible, exporting the needed elements so the plugins can use them; namely:
- The component itself, `PurchaseModal`.
- The function `openModal` to trigger its visibility.
- The constant `ModalIdentifiers`, which contains the ID required as one of the arguments to `openModal`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35626

#### Release Note
```release-note
NONE
```

### DIFF
--- a/plugins/export.js
+++ b/plugins/export.js
@@ -6,6 +6,10 @@ import {formatText} from 'utils/text_formatting';
 import {browserHistory} from 'utils/browser_history';
 import Textbox from 'components/textbox';
 
+import {openModal} from 'actions/views/modals';
+import {ModalIdentifiers} from 'utils/constants';
+import PurchaseModal from 'components/purchase_modal';
+
 // The following import has intentional side effects. Do not remove without research.
 import {openInteractiveDialog} from './interactive_dialog';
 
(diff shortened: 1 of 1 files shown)


agarciamontoro:
@marianunez, I assigned you as a reviewer for visibility, so you and your team knows we are using this component in Incident Collaboration.

marianunez:
@agarciamontoro this purchase modal only applies for Cloud and depends on having connectivity to CWS (Customer Web Service) which is available for all cloud workspaces. Do you have an alternative of what options to show for upgrade when IC is in a self-managed server?

agarciamontoro:
@marianunez, yes, this is for cloud only. The flow for self-managed is already implemented, this is the last step :)
Also, as this is only for cloud, we can rely on the workspaces version being updated, so the plugin can assume that this is actually exported by the webapp.

lieut-data (approved):

marianunez (approved):

crspeller (approved):

agarciamontoro:
Skipping QA review here, as we're just exporting components, and this will be tested as part of the Incident Collaboration plugin. @prapti, I'll let you know when the PR in the incident-collaboration repo is up.

PR #8231, opened 2021-06-08
Fixes: crash at global threads on root post delete

koox00:
#### Summary

If a root post is deleted and the user is at global threads view,
the app crashes.

This commit fixes that by removing the thread from the threads reducer.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35939

#### Release Note

```release-note
NONE
```


packages/mattermost-redux/src/reducers/entities/threads.ts
```diff
@@ -75,6 +87,29 @@ export const threadsReducer = (state: ThreadsState['threads'] = {}, action: Gene
...
+        }
+
+        const teamId = Object.keys(state).
+            find((id) => state[id].indexOf(post.id) !== -1);
+
+        if (!teamId) {
+            return state;
+        }
+
+        const index = state[teamId].indexOf(post.id);
```
hmhealey, line 103: Would there be any benefit to checking if index is -1 here?

koox00: It wouldn't hurt but since we check it in the find I thought it would be redundant.

```js
const teamId = Object.keys(state).find((id) => state[id].indexOf(post.id) !== -1);
```

hmhealey: Oh, I missed that it was coming from the state already. I assumed it was coming from the action for some reason.

hmhealey (approved):

jgilliam17:
Thanks @koox00 
I am not seeing a crash, but there are few errors in the console after root post is deleted. 
There is a follow up ticket to address the appearance of the deleted thread, let me know if observed issues will be addressed here or separately. Thanks. 
Steps: 
1. Create a post as user1
2. Add few replies as both user1 and user2
3. Delete root post as user 1 while user 2 is viewing global threads

Observed:
Post is removed from global threads for user 1, but not for user2 (observing one)
User2 can can click on deleted thread to reveal "message deleted" text but no longer seeing replies
User2 can see the footer of the deleted post after navigating back to channel where the post originated from, and thread opens empty on the RHS.  Footer is missing for the user1 that deleted the post. 
Screen Shot 2021-06-09 at 5 07 54 PM

koox00:
Thanks @jgilliam17, since this ticket was only about the crash, and the other ticket contains expected functionality it makes more sense to me to address those issues separately.

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified app does not crash when user is deleting a root post while on global threads. 
Other issues raised will be addressed in the follow up ticket.

anurag6713 (approved):

esethna:
@koox00 good to merge?

PR #8236, opened 2021-06-08
[MM-34330] Migrate 'components/at_mention' and associated tests to TypeScript 

majidsajadi:
#### Summary

migrated at_mention to typescript.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/17261

#### Release Note

```release-note
NONE
```


components/at_mention/at_mention.test.tsx
```diff
@@ -7,6 +7,8 @@ import {shallow} from 'enzyme';
 import {General} from 'mattermost-redux/constants';
 
 import AtMention from 'components/at_mention/at_mention.jsx';
+import { UserProfile } from 'mattermost-redux/types/users';
```
hmhealey, line 10: These imports are causing the lint check to fail, so they need to be fixed

```
/root/mattermost/mattermost-webapp/components/at_mention/at_mention.test.tsx
   9:1   error  There should be at least one empty line between import groups                                             import/order
   9:1   error  `components/at_mention/at_mention.jsx` import should occur before import of `mattermost-redux/constants`  import/order
  10:9   error  There should be no space after '{'                                                                        object-curly-spacing
  10:21  error  There should be no space before '}'                                                                       object-curly-spacing
  11:9   error  There should be no space after '{'                                                                        object-curly-spacing
  11:15  error  There should be no space before '}'                                                                       object-curly-spacing
  ```

File: components/at_mention/at_mention.test.tsx [hunk 2]
```diff
@@ -15,19 +17,19 @@ describe('components/AtMention', () => {
         currentUserId: 'abc1',
         teammateNameDisplay: General.TEAMMATE_NAME_DISPLAY.SHOW_NICKNAME_FULLNAME,
         usersByUsername: {
-            currentuser: {id: 'abc1', username: 'currentuser', first_name: 'First', last_name: 'Last'},
-            user1: {id: 'abc2', username: 'user1', first_name: 'Other', last_name: 'User', nickname: 'Nick'},
-            'userdot.': {id: 'abc3', username: 'userdot.', first_name: 'Dot', last_name: 'Matrix'},
+            currentuser: {id: 'abc1', username: 'currentuser', first_name: 'First', last_name: 'Last'} as UserProfile,
+            user1: {id: 'abc2', username: 'user1', first_name: 'Other', last_name: 'User', nickname: 'Nick'} as UserProfile,
+            'userdot.': {id: 'abc3', username: 'userdot.', first_name: 'Dot', last_name: 'Matrix'} as UserProfile,
```
michelengelen, line 22: can't we create valid Objects here instead of type casting what we need?

components/at_mention/at_mention.test.tsx
```diff
@@ -15,19 +17,19 @@ describe('components/AtMention', () => {
...
         },
         groupsByName: {
-            developers: {id: 'qwerty1', name: 'developers', allow_reference: true},
-            marketing: {id: 'qwerty2', name: 'marketing', allow_reference: false},
-            accounting: {id: 'qwerty3', name: 'accounting', allow_reference: true},
+            developers: {id: 'qwerty1', name: 'developers', allow_reference: true} as Group,
+            marketing: {id: 'qwerty2', name: 'marketing', allow_reference: false} as Group,
+            accounting: {id: 'qwerty3', name: 'accounting', allow_reference: true} as Group,
```
michelengelen, line 27: same here

components/at_mention/at_mention.test.tsx
```diff
@@ -227,10 +229,10 @@ describe('components/AtMention', () => {
                 getBoundingClientRect: () => ({
                     top: 400,
                 }),
-            },
+            } as any,
```
michelengelen, line 232: there is a type for this defined in the react package: `RefObject<T>` with the generic `T` being the kind of `HTMLElement` (`HTMLDivElement` for example)

I would want to keep the usage of `any` as low as possible.

components/at_mention/at_mention.test.tsx
```diff
@@ -227,10 +229,10 @@ describe('components/AtMention', () => {
...
         };
 
-        wrapper.instance().handleClick({preventDefault: jest.fn(), target: AtMention});
+        wrapper.instance().handleClick({preventDefault: jest.fn(), target: AtMention} as any);
```
michelengelen, line 235: same here ... create as strictly typed objects as possible please.

components/at_mention/at_mention.test.tsx
```diff
@@ -239,10 +241,10 @@ describe('components/AtMention', () => {
                     top: 200,
                     bottom: 400,
                 }),
-            },
+            } as any,
```
michelengelen, line 244: same here

components/at_mention/at_mention.test.tsx
```diff
@@ -239,10 +241,10 @@ describe('components/AtMention', () => {
...
         };
 
-        wrapper.instance().handleClick({preventDefault: jest.fn(), target: AtMention});
+        wrapper.instance().handleClick({preventDefault: jest.fn(), target: AtMention} as any);
```
michelengelen, line 247: same here

components/at_mention/at_mention.test.tsx
```diff
@@ -251,10 +253,10 @@ describe('components/AtMention', () => {
                     top: 200,
                     bottom: 1000,
                 }),
-            },
+            } as any,
```
michelengelen, line 256: same here

components/at_mention/at_mention.test.tsx
```diff
@@ -251,10 +253,10 @@ describe('components/AtMention', () => {
...
         };
 
-        wrapper.instance().handleClick({preventDefault: jest.fn(), target: AtMention});
+        wrapper.instance().handleClick({preventDefault: jest.fn(), target: AtMention} as any);
```
michelengelen, line 259: same here

components/at_mention/at_mention.tsx
```diff
@@ -1,41 +1,52 @@
...
+    isRHS?: boolean;
+    mentionName: string;
+    teammateNameDisplay: string;
+    usersByUsername: UsernameMappedObjects<UserProfile>;
+    groupsByName: NameMappedObjects<Group>;
+}
+
+type State = {
+    show: boolean;
+    target?: EventTarget;
```
hmhealey, line 35: It looks a little weird to see this as `EventTarget` when I feel like it should be a DOM node. If you specify the type of the event passed to `handleClick` as `React.MouseEvent<HTMLAnchorElement>` to match the type given to the ref, does that let you define target as an `HTMLAnchorElement` and avoid the `as` when passing `this.state.target` into `Overlay` as a prop?

components/at_mention/at_mention.tsx
```diff
@@ -1,41 +1,52 @@
...
+    placement?: string;
+}
```
michelengelen, line 37: could you please sort the properties by mandatory and optional? Thanks!

The way it is done in the `State` type is fine

hmhealey (changes requested):
As always, thanks for the PR, @majidsajadi! 🙂

components/at_mention/at_mention.tsx
```diff
@@ -45,9 +56,13 @@ export default class AtMention extends React.PureComponent {
...
-        const placement = popOverOverlayPosition(targetBounds, window.innerHeight, spaceRequiredForPopOver);
+    handleClick = (e: React.MouseEvent) => {
+        const targetBounds = this.overlayRef.current?.getBoundingClientRect();
+
+        let placement;
+        if (targetBounds) {
+            placement = popOverOverlayPosition(targetBounds, window.innerHeight, spaceRequiredForPopOver);
+        }
 
         this.setState({target: e.target, show: !this.state.show, placement});
```
michelengelen, line 67: this needs to be adjusted, since we call `setState` regardless of a value set for `placement` or not. This could potentially lead to bugs.

components/at_mention/at_mention.tsx
```diff
@@ -108,7 +123,7 @@ export default class AtMention extends React.PureComponent {
                 <Overlay
                     placement={this.state.placement}
                     show={this.state.show}
-                    target={this.state.target}
+                    target={this.state.target as React.ReactInstance}
```
michelengelen, line 126: is this the correct type @hmhealey ?

michelengelen (changes requested):
some changes in typings and one in logic.

components/at_mention/at_mention.tsx
majidsajadi: even with defining `React.MouseEvent<HTMLAnchorElement>` as props type for `handleClick` i had to cast `e.target` as `HTMLAnchorElement` since `React.MouseEvent` extends `BaseSyntheticEvent` and the event type in `BaseSyntheticEvent` is `EventTarget & T`.

components/at_mention/at_mention.test.tsx
majidsajadi: i change this to use `TestHelper` methods here. why not using these methods in evey tests?

components/at_mention/at_mention.test.tsx
majidsajadi: same here.

components/at_mention/at_mention.test.tsx
majidsajadi: any idea for this type?

michelengelen: well ... you are defining it yourself in `at_mention.tsx` => `React.MouseEvent<HTMLAnchorElement>`

michelengelen (commented):
the typing should be adjusted ... otherwise LGTM.

components/at_mention/at_mention.test.tsx
majidsajadi: i have tried `React.MouseEvent<HTMLAnchorElement>` but `{preventDefault: jest.fn(), target: AtMention}` is not a valid `React.MouseEvent<HTMLAnchorElement>` a lot of property missing.

michelengelen: Well, since we probably do not need a full fletched event for testing purposes we leave it like that.

michelengelen (approved):
LGTM

components/at_mention/at_mention.test.tsx
hmhealey: I think some of the tests that aren't using it were tests were created before the TestHelper methods were added, and no one ever went back and updated the old tests.

components/at_mention/at_mention.tsx
hmhealey: Oh, I see. I think I like having the `as HTMLAnchorElement` in the click handler like you've done now though, so this looks good to me

hmhealey (approved):

jgilliam17:
Thank you @majidsajadi 
Mention autocomplete looks good 
Can you please take a look at the test failures that are present after branch update?

jgilliam17:
@majidsajadi `At Mention` tests are failing on CircleCI, can you please take a look. Thanks.

majidsajadi:
@jgilliam17  I tried to run tests on my local machine but I got a lot of errors because of `timestamps`? 
i don't know if it's because of my local time or anything else. i cant check CircleCi because i live in Iran and CircleCI restricted our access. I would give it a try again and check if i can resolve the errors.

hmhealey:
The test failures I'm seeing are because the snapshots for the AtMention component were removed at some point. You should be able to regenerate them by running `npm test -- -u components/at_mention`.

Also that's interesting that CircleCI is blocked for you. I wonder if we might want to come up with a way to echo at least which tests are failing back to GitHub just in case others are located in areas where that's the case.

majidsajadi:
@hmhealey thanks for the tip. fixed. 

also, I guess this would be an amazing idea that makes contributing easier for us.

jgilliam17 (approved):
Thank you @majidsajadi and thanks @hmhealey for your help in fixing test on CircleCI
Tested, looks good to merge.
- Verified at mentions are as before.

PR #8237, opened 2021-06-08
[MM-36704] Add selector specifically for search channel name to handle its cases

devinbinnie:
#### Summary
Due to some weirdness with the way we process channels, we don't have a consistent way of using one field to search for or label our channels. DM/GM channels are searched for commonly by the users inside them, and other channels are search in by name.

Since adding our little search shortcut, I used the `name` of the channel which worked for non DM/GM channels, but for DM/GM channels things were a little inconsistent. In order to make things consistency, and get around some of the weirdness in how we fill channels, I've added a new selector specifically to get the name used for search and am using that to fill in the name when using the `Cmd/Ctrl+Shift+F` shortcut so that its consistent with how the suggestion provider does it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36074

#### Release Note
```release-note
NONE
```

### DIFF
--- a/packages/mattermost-redux/src/selectors/entities/channels.ts
+++ b/packages/mattermost-redux/src/selectors/entities/channels.ts
@@ -292,7 +292,7 @@ export const getCurrentChannelNameForSearchShortcut: (state: GlobalState) => str
 
         // Replace spaces in GM channel names
         if (channel?.type === Constants.GM_CHANNEL) {
-            return `@${channel.display_name.replace(/ /g, '')}`;
+            return `@${channel.display_name.replace(/\s/g, '')}`;
         }
 
         return channel?.name;
--- a/packages/mattermost-redux/src/selectors/entities/channels.ts
+++ b/packages/mattermost-redux/src/selectors/entities/channels.ts
@@ -3,7 +3,7 @@
 
 import {createSelector} from 'reselect';
 
-import {General, Permissions} from 'mattermost-redux/constants';
+import {General, Permissions, Preferences} from 'mattermost-redux/constants';
 import {CategoryTypes} from 'mattermost-redux/constants/channel_categories';
 
 import {getCategoryInTeamByType} from 'mattermost-redux/selectors/entities/channel_categories';
@@ -280,13 +280,12 @@ export const getCurrentChannelNameForSearchShortcut: (state: GlobalState) => str
     getAllChannels,
     getCurrentChannelId,
     (state: GlobalState): UsersState => state.entities.users,
-    getTeammateNameDisplaySetting,
-    (allChannels: IDMappedObjects<Channel>, currentChannelId: string, users: UsersState, teammateNameDisplay: string): string | undefined => {
+    (allChannels: IDMappedObjects<Channel>, currentChannelId: string, users: UsersState): string | undefined => {
         const channel = allChannels[currentChannelId];
 
         // Only get the extra info from users if we need it
         if (channel?.type === Constants.DM_CHANNEL) {
-            const dmChannelWithInfo = completeDirectChannelInfo(users, teammateNameDisplay, channel);
+            const dmChannelWithInfo = completeDirectChannelInfo(users, Preferences.DISPLAY_PREFER_USERNAME, channel);
             return `@${dmChannelWithInfo.display_name}`;
         }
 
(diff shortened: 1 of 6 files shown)


packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -276,6 +276,29 @@ export const getCurrentChannel: (state: GlobalState) => Channel = createSelector
...
+
+        // Only get the extra info from users if we need it
+        if (channel?.type === Constants.DM_CHANNEL) {
+            const dmChannelWithInfo = completeDirectChannelInfo(users, teammateNameDisplay, channel);
+            return `@${dmChannelWithInfo.display_name}`;
+        }
+
+        // Replace spaces in GM channel names
+        if (channel?.type === Constants.GM_CHANNEL) {
+            return `@${channel.display_name.replace(/ /g, '')}`;
```
michelengelen, line 295:
```suggestion
            return `@${channel.display_name.replace(/\s/g, '')}`;
```

Filter for all possible whitespace characters

michelengelen (approved):
just a minor improvement in the regex ... otherwise LGTM!

packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -276,6 +276,29 @@ export const getCurrentChannel: (state: GlobalState) => Channel = createSelector
...
+    getCurrentChannelId,
+    (state: GlobalState): UsersState => state.entities.users,
+    getTeammateNameDisplaySetting,
+    (allChannels: IDMappedObjects<Channel>, currentChannelId: string, users: UsersState, teammateNameDisplay: string): string | undefined => {
+        const channel = allChannels[currentChannelId];
+
+        // Only get the extra info from users if we need it
+        if (channel?.type === Constants.DM_CHANNEL) {
+            const dmChannelWithInfo = completeDirectChannelInfo(users, teammateNameDisplay, channel);
+            return `@${dmChannelWithInfo.display_name}`;
```
hmhealey, line 290: Does this work when the teammate name display setting is for something other than username?

packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -276,6 +276,29 @@ export const getCurrentChannel: (state: GlobalState) => Channel = createSelector
...
+        }
+
+        // Replace spaces in GM channel names
+        if (channel?.type === Constants.GM_CHANNEL) {
+            return `@${channel.display_name.replace(/\s/g, '')}`;
```
hmhealey, line 295: Ditto here. Also, what happens if the GM channel name exceeds the max length? That said, I'm not actually sure if the `in:` search handles that to begin with

hmhealey (commented):
I'm a bit worried about having that selector recalculate often since it'll recalculate a lot by taking in `state.entities.users`. Perhaps instead of passing in the name for the search query, we could pass in an action that searches for posts in the current channel?
```
function searchPostsInCurrentChannel() {
    return (dispatch, getState) {
        const channelName = getCurrentChannelNameForSearchShortcut(state);
        dispatch(updateSearchTerms(`in:${channelName}`));
    };
}
```

packages/mattermost-redux/src/selectors/entities/channels.ts
devinbinnie: Good point.

packages/mattermost-redux/src/selectors/entities/channels.ts
devinbinnie: This one does work, because the logic to replace those names is elsewhere. We use this exact same logic in the search provider to get GMs. It's very confusing and inconsistent tbh :P

hmhealey (approved):

jgilliam17:
Thanks @devinbinnie 
DMs look good.
GMs are not working as expected. 
The string that is pre-filled using Cmd/Ctrl+Shift+F is omitting some users and adding a logged in user. See images.
Screen Shot 2021-06-15 at 12 12 36 PM

Screen Shot 2021-06-15 at 12 12 06 PM

devinbinnie:
@hmhealey Needed to do a little re-work so I'd like another pass if you wouldn't mind :)

hmhealey (approved):
This looks like it should be fine. I'm not a huge fan of adding an extra parameter to `completeDirectGroupInfo` just for this, but I think that might just be me having a problem with that function in general because it mutates the channel.

jgilliam17 (approved):
Thanks @devinbinnie 
Tested, looks good to merge.
- Verified using cmd/ctrl+f+shift while viewing a DM or GM, pre-fills the search field with user names instead of channel IDs.

PR #8238, opened 2021-06-09
MM-35840: fixes up arrow edit in CRT

koox00:
#### Summary
We used to get post ids from the channel to find current user's last
post. This worked in threads when CRT off, but if you turn CRT on
replies are not added in the channel thus not being able to find the
user's last post correctly (was always the root post).

This commit fixes that by getting post ids from the thread*
(`postsInThread` reducer) and not the channel, this way both CRT on/off are
working as expected.

*renamed the selector from `makeGetCurrentUsersLatestPost` to
`makeGetCurrentUsersLatestReply.

`#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35840

#### Release Note

```release-note
NONE
```

### DIFF
--- a/actions/views/create_comment.jsx
+++ b/actions/views/create_comment.jsx
@@ -9,7 +9,7 @@ import {
     makeGetMessageInHistoryItem,
     makeGetCommentCountForPost,
     getPost,
-    getPostIdsInChannel,
+    getPostIdsInThread,
 } from 'mattermost-redux/selectors/entities/posts';
 import {getCustomEmojisByName} from 'mattermost-redux/selectors/entities/emojis';
 import {
@@ -32,7 +32,7 @@ import * as Utils from 'utils/utils.jsx';
 import {Constants, StoragePrefixes} from 'utils/constants';
 
 export function clearCommentDraftUploads() {
-    return actionOnGlobalItemsWithPrefix(StoragePrefixes.COMMENT_DRAFT, (key, value) => {
+    return actionOnGlobalItemsWithPrefix(StoragePrefixes.COMMENT_DRAFT, (_key, value) => {
         if (value) {
             return {...value, uploadsInProgress: []};
         }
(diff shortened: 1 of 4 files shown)


jwilander (approved):

jgilliam17:
Thanks @koox00 
Seeing few issues
1. After user makes multiple back to back replies in a thread, arrow up is not picking up the last post they made, instead it picks up the first they made after another user. Repros with CRT on and off
e.g. User1 posts 3 consecutive posts `7`, `8` and `9` after sysadmin; arrow up brings up edit modal for the first post `7` and not last `9`
Screen Shot 2021-06-09 at 10 32 23 PM
2. There is a 500 server error when following a post 
Screen Shot 2021-06-09 at 10 31 22 PM
3. Unable to use arrow up on a single post in RHS, nothing happens, no errors either. Repros with CRT on and off

koox00:
Thanks @jgilliam17!
1 and 3 should be fixed now, and I think 2 have been fixed on master in another ticket, 
I updated this branch so we shouldn't see it.

jgilliam17:
Thanks @koox00 - 1 and 3 are fixed, looks good. The server 500 error is still present when opening single post on the RHS and starting to follow it, but I just retested and it reproduces on daily. 
There are also few gateway time-out errors.  I don't have exact repro steps, I left the app idle for few minutes to repro errors on daily and upon return found the errors in the console for both users.
Screen Shot 2021-06-10 at 9 43 11 AM
Screen Shot 2021-06-10 at 9 39 47 AM

esethna:
@koox00 can we re-assign reviewers if you're comfortable with it since Caleb is out for the rest of the week?

lieut-data:
I confess I'm not sure I'm in the loop enough to properly review this change -- deferring back to @calebroseland for now.

koox00:
FYI @jgilliam17 the 500 error is unrelated to this PR, I've opened another ticket for that.
The websocket issue seem unrelated too since we saw those yesterday on the testing session (this PR was not tested), but I can't find the cause of that.

jgilliam17 (approved):
Thanks @koox00 for adding a server error ticket
Tested, looks good to merge.
- Verified up arrow opens correct post for edit with CRT on or off.

calebroseland (approved):

PR #8249, opened 2021-06-10
Filter forms and embedded bindings

larkox:
#### Summary
Filter forms to avoid malformed forms giving a bad experience.
There is a filter already done on the server side (https://github.com/mattermost/mattermost-plugin-apps/pull/203 ), but this is a sanity check.

We also do some filtering on embedded bindings, to make sure no empty selects or empty labels get rendered.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33362
https://mattermost.atlassian.net/browse/MM-33436

#### Related Pull Requests
Proxy: https://github.com/mattermost/mattermost-plugin-apps/pull/203
Mobile. Soon

#### Release Note
```release-note
NONE
```


larkox:
After this is approved, I will make the mobile version, since most of the code will be practically the same.

components/suggestion/command_provider/app_command_parser/app_command_parser.ts
```diff
@@ -1183,7 +1183,8 @@ export class AppCommandParser {
             }));
         }
 
-        const items = callResponse?.data?.items;
+        let items = callResponse?.data?.items;
+        items = items?.filter((option) => option.value && !option.value.match(/^[ \t]+$/));
```
hanzei, line 1187: Does it make sense to move the filtering into a helper function to keep the code DRY?

components/post_view/embedded_bindings/select_binding/select_binding.tsx
```diff
@@ -133,15 +158,23 @@ export class SelectBinding extends React.PureComponent<Props, State> {
...
     render() {
+        if (!this.nOptions) {
+            return null;
+        }
+
         const {binding} = this.props;
+        const label = binding.label || binding.location;
+        if (!label) {
+            return null;
+        }
```
hanzei, line 169: I'm hesitant to add such fallback on the client side as the server is not longer the single source of truth. Also, the client's verions might be not synchronized and hence there would be different behavior on mobile and webapp.

components/post_view/embedded_bindings/select_binding/select_binding.tsx
```diff
@@ -43,16 +43,41 @@ type State = {
...
+
+                if (usedValues[b.location]) {
+                    return;
+                }
+
+                usedLabels[label] = true;
+                usedValues[b.location] = true;
+
+                options.push({text: label, value: b.location});
             });
```
hanzei, line 78: What is the advantage of also checking the semantics here?

packages/mattermost-redux/src/utils/apps.ts
```diff
@@ -1,116 +1,188 @@
...
+                return;
+            }
+            break;
+        }
+        case AppBindingLocations.CHANNEL_HEADER_ICON: {
+            if (!b.icon) {
+                toRemove.unshift(i);
+                return;
+            }
+            break;
```
hanzei, line 60: Is this checking also server side?

hanzei (changes requested):

components/post_view/embedded_bindings/select_binding/select_binding.tsx
larkox: The problem we have at the moment is that for embedded bindings, we have no way to validate in the server side. Therefore, we have to check on the client.

packages/mattermost-redux/src/utils/apps.ts
larkox: This is a "webapp only problem". In mobile we don't need the image, because it is always a menu.

larkox: That being said, we can check the context, and filter those on server side if we are serving the webapp. Will do in my server side PR.

components/post_view/embedded_bindings/select_binding/select_binding.tsx
larkox: Not sure what are you asking. Could you elaborate?

packages/mattermost-redux/src/utils/apps.ts
```diff
@@ -1,116 +1,189 @@
...
+                toRemove.unshift(i);
+                return;
+            }
+
+            // Remove leaves without app id
+            if (!b.app_id) {
+                toRemove.unshift(i);
+                return;
+            }
+        }
```
jfrerich, line 85:
```suggestion
        } else {
            // Remove leaves without a call or app id
            if (!b.call || !b.app_id) {
                toRemove.unshift(i);
                return;
            }
        }
```
0/5 You could combine the two

packages/mattermost-redux/src/utils/apps.ts
```diff
@@ -1,116 +1,189 @@
...
+    const channelHeaderBindings = bindings?.filter((v) => v.location === AppBindingLocations.CHANNEL_HEADER_ICON);
+    const postMenuBindings = bindings?.filter((v) => v.location === AppBindingLocations.POST_MENU_ITEM);
+    const commandBindings = bindings?.filter((v) => v.location === AppBindingLocations.COMMAND);
+
+    channelHeaderBindings.forEach((v) => cleanBinding(v, AppBindingLocations.CHANNEL_HEADER_ICON, 0));
+    postMenuBindings.forEach((v) => cleanBinding(v, AppBindingLocations.POST_MENU_ITEM, 0));
+    commandBindings.forEach((v) => cleanBinding(v, AppBindingLocations.COMMAND, 0));
+
+    const hasBindings = (v: AppBinding) => v.bindings?.length;
+    return postMenuBindings.filter(hasBindings).concat(channelHeaderBindings.filter(hasBindings), commandBindings.filter(hasBindings));
```
jfrerich, line 105: Nit, but I would prefer `b` instead of `v` to represent a `binding`

packages/mattermost-redux/src/utils/apps.ts
```diff
@@ -1,116 +1,189 @@
...
+    field.options?.forEach((option, i) => {
+        let label = option.label;
+        if (!label) {
+            label = option.value;
+        }
+
+        if (!label) {
+            toRemove.unshift(i);
+            return;
+        }
```
jfrerich, line 170:
```suggestion
        if (!label) {
            label = option.value;
        }

        if (!label) {
            toRemove.unshift(i);
            return;
        }
```
could you combine these two?

packages/mattermost-redux/src/utils/apps.ts
jfrerich: there are quite a few cases where the `if` statements could be combined to share the common `toRemove.unshift(i)` calls. This is non-blocking and just 0/5 because it's a personal choice :)

jfrerich (approved):
I only have some nits on DRYing up code, but they are only suggestions and non-blocking.

packages/mattermost-redux/src/utils/apps.ts
larkox: I preferred being a bit more verbose here, since we may add or remove more rules here eventually.

packages/mattermost-redux/src/utils/apps.ts
larkox: I was going with `v` for `value` but you are right, `b` for `binding` looks better here.

packages/mattermost-redux/src/utils/apps.ts
larkox: It is weird to combine. We could nest the if, but still would feel weird:
```
if (!label) {
    if (!option.value) {
        toRemove.unshift(i);
        return;
    }
    label = option.value;
}
```
or
```
if (!label) {
    label = option.value;
    if (!label) {
        toRemove.unshift(i);
        return;
    }
}
```
We could turn it around, but it would be pretty similar still:
```
if (!label && !option.value) {
    toRemove.unshift(i);
    return;
}
if (!label) {
    label = option.value;
}
```

components/post_view/embedded_bindings/select_binding/select_binding.tsx
hanzei: Excellent point.

components/post_view/embedded_bindings/select_binding/select_binding.tsx
hanzei: https://github.com/mattermost/mattermost-webapp/pull/8249#discussion_r662359365 make my point obsolete.

hanzei (approved):
LGTM

packages/mattermost-redux/src/utils/apps.ts
jfrerich: That explanation makes sense.  When I glanced at this code I didn't account for `option.value` possibly being undefined.

jfrerich (approved):

larkox:
@DHaussermann Friendly reminder for review.

DHaussermann (approved):
Tested and passed
Deployed this along with https://github.com/mattermost/mattermost-plugin-apps/pull/203
- Availability of the subcommands is working as expected
- In all cases the test steps provided by dev match the behavior I'm seeing
- No issues found

LGTM!

PR #8257, opened 2021-06-11
MM-35917: closes RHS upon navigating to threads

koox00:
#### Summary

Clicking on global threads link closes the RHS.
This is done to reduce the amount of panels open 
in the global threads view upon arriving.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35917

#### Release Note

```release-note
NONE
```


lieut-data (approved):

esethna:
@koox00 Looks good, but one piece is missing I think, which is that when the user switches away from "Threads" we should open the RHS back to whatever content it had open before it was closed. 

In the case where the user executes a search/recent mentions/saved posts while in the threads view then we can leave the RHS open to that when they switch away from Threads

koox00:
> @koox00 Looks good, but one piece is missing I think, which is that when the user switches away from "Threads" we should open the RHS back to whatever content it had open before it was closed.
> 
> In the case where the user executes a search/recent mentions/saved posts while in the threads view then we can leave the RHS open to that when they switch away from Threads

Oh, I forgot that :blush:

koox00:
FYI @esethna restoring the RHS state upon leaving global threads has been added

matthewbirtch:
@koox00 looks like the test server failed, so I'm not able to test.

koox00:
> @koox00 looks like the test server failed, so I'm not able to test.

I had forced push a minor type change in tests so it tried to build for a non-existent commit thus failed.

The changes are in though, so it's OK to test!

esethna (approved):

esethna:
Thanks @koox00!

matthewbirtch (approved):
Looks good @koox00. Approving this, but I just wanted to note one thing below:

You can still get in to a case where the same thread is open in both the RHS and the global threads viewer if you the RHS keyboard shortcut and follow these steps:

1. Go to a channel and open a thread in the RHS
2. Switch to the Global Threads view and open that same thread.
3. On your keyobard use the `Command .` shortcut and it will open the same thread

However, I don't think this is a big deal. We have limited most cases where the two would show side-by-side now, so I think we're good.

esethna:
@matthewbirtch Ya I'm not concerned about that case. If a user executes that shortcut it was their intention to open the RHS

reducers/views/rhs_stash.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {GenericAction} from 'mattermost-redux/types/actions';
+
+import {ViewsState} from 'types/store/views';
+import {ActionTypes} from 'utils/constants';
+
+export default function rhsStash(state: ViewsState['rhsStash'] = null, action: GenericAction): ViewsState['rhsStash'] {
```
lieut-data, line 9: I see code above that appears to handle plugins, but just confirming explicitly that this still works with plugin content in the RHS?

lieut-data (changes requested):
Thanks for a really clear PR, @koox00! One worry I have is needing to embed the actual RHS state into the stash, and the need for all existing RHS reducers to be "stash aware".

I haven't thought this through fully -- so your approach may still prove superior! -- but what if we flipped this around and introduced a new bit in the store: `rhsSuppressed`? This would get set to true when the threads view component mounts, and (unconditionally) set to false when its reducer receives any of the existing RHS actions The view time logic for the RHS would check `rhsSuppressed` before the other checks to decide whether to render. Of course, the threads view would also dispatch to set this bit false when it closes as your code does today.

koox00:
> Thanks for a really clear PR, @koox00! One worry I have is needing to embed the actual RHS state into the stash, and the need for all existing RHS reducers to be "stash aware".

that's a valid worry

> I haven't thought this through fully -- so your approach may still prove superior! -- but what if we flipped this around and introduced a new bit in the store: `rhsSuppressed`? This would get set to true when the threads view component mounts, and (unconditionally) set to false when its reducer receives any of the existing RHS actions The view time logic for the RHS would check `rhsSuppressed` before the other checks to decide whether to render. Of course, the threads view would also dispatch to set this bit false when it closes as your code does today.

That's actually a lot simpler than my solution! On the top of my head seems solid, will try it now, thanks!!

components/threading/global_threads_link/global_threads_link.tsx
```diff
@@ -36,6 +37,10 @@ const GlobalThreadsLink = () => {
     const unreadsOnly = useSelector(isUnreadFilterEnabled);
     const someUnreadThreads = counts?.total_unread_threads;
 
+    const closeRHS = useCallback(() => {
+        dispatch(suppressRHS);
+    }, []);
+
```
Willyfrog, line 43: I'm not really sure I get why do we want to memoize this with no changing arguments, isn't it the same as
```suggestion
    const closeRHS = () => {dispatch(suppressRHS)};
```

Willyfrog (changes requested):
i played with it a bit and found a weird thing:
1- search for a term or click on the at-sign at the top
2- go to threads (rhs closes)
3- click on the search and hit enter (rhs opens)
4- go to the channel
5- go back to threads
observed:
search help popup appears and it is partially hidden

Screen Shot 2021-06-16 at 10 46 14

Willyfrog (approved):

koox00:
> i played with it a bit and found a weird thing:
> 1- search for a term or click on the at-sign at the top
> 2- go to threads (rhs closes)
> 3- click on the search and hit enter (rhs opens)
> 4- go to the channel (rhs stays)
> 5- go back to threads (rhs closes... but)
> observed:
> search help popup appears and it is partially hidden

Good catch, thanks @Willyfrog ! 
To replicate it you have to **not** lose focus on the search input while in threads, and then go back to the channel.

Willyfrog:
> > i played with it a bit and found a weird thing:
> > 1- search for a term or click on the at-sign at the top
> > 2- go to threads (rhs closes)
> > 3- click on the search and hit enter (rhs opens)
> > 4- go to the channel (rhs stays)
> > 5- go back to threads (rhs closes... but)
> > observed:
> > search help popup appears and it is partially hidden
> 
> Good catch, thanks @Willyfrog !
> To replicate it you have to **not** lose focus on the search input while in threads, and then go back to the channel.

yes, kind of hard to get it to happen

koox00:
> i played with it a bit and found a weird thing:
> 1- search for a term or click on the at-sign at the top
> 2- go to threads (rhs closes)
> 3- click on the search and hit enter (rhs opens)
> 4- go to the channel (rhs stays)
> 5- go back to threads (rhs closes... but)
> observed:
> search help popup appears and it is partially hidden

@Willyfrog it should be fixed now, can you please take a look at the fix to see if it makes sense to you.

Willyfrog:
LGTM! thanks for taking care of it :D

reducers/views/rhs_suppressed.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {GenericAction} from 'mattermost-redux/types/actions';
+
+import {ViewsState} from 'types/store/views';
+import {ActionTypes} from 'utils/constants';
+
+export default function rhsStash(state: ViewsState['rhsSuppressed'] = false, action: GenericAction): boolean {
```
lieut-data, line 9: Should we rename this to be `rhsSuppressed` vs. `rhsStash`?

lieut-data (approved):

reducers/views/rhs_suppressed.ts
koox00: thanks for noticing! will do

lieut-data:
Nicely done, @koox00! One quick ask: can you avoid force pushing during a PR? See [Submitting Great PRs](https://developers.mattermost.com/blog/submitting-great-prs/#4-avoid-force-pushing) for a few thoughts (I may be biased here :p), but since we squash everything on merge, it's fine to have "fix" PRs that streamline the review process and won't be seen in `master` when it's all done :)

koox00:
> Nicely done, @koox00! One quick ask: can you avoid force pushing during a PR? See [Submitting Great PRs](https://developers.mattermost.com/blog/submitting-great-prs/#4-avoid-force-pushing) for a few thoughts (I may be biased here :p), but since we squash everything on merge, it's fine to have "fix" commits that streamline the review process and won't be seen in `master` when it's all done :)

Force of habit... Yes of course :slightly_smiling_face:

jgilliam17:
Thanks @koox00 
Looks good.
Seeing one small glitch
 If RHS is expanded (via expand sidebar button) while on global threads it remains expanded on the switch to any other channel.
  I expected it to remain open, but in non-expanded view . If you change the channel again it collapses. See video.
  For comparison if RHS is expanded while viewing a channel, switching to a different channel collapses the RHS immediately.

https://user-images.githubusercontent.com/52937121/122230070-4cd70980-ce87-11eb-8580-d65eef0955e1.mov

koox00:
> Thanks @koox00
> Looks good.
> Seeing one small glitch
>  If RHS is expanded (via expand sidebar button) while on global threads it remains expanded on the switch to any other channel.
>   I expected it to remain open, but in non-expanded view . If you change the channel again it collapses. See video.
>   For comparison if RHS is expanded while viewing a channel, switching to a different channel collapses the RHS immediately.

Thanks @jgilliam17 it should be an easy fix!

koox00:
@jgilliam17 it should be fixed now!

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified opening global threads closes the RHS. 
- Verified if RHS was open prior to visiting Threads, user is returned to viewing the same content on the RHS after leaving Threads.
- Verified RHS can be opened while viewing global threads

PR #8258, opened 2021-06-11
MM-35853 Properly memoize getUserTimezone selector

jwilander:
#### Summary
Added makeGetUserTimezone selector creator and updated getUserTimezone to require the user map being passed in.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35853

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/suggestion/search_date_suggestion/index.ts
+++ b/components/suggestion/search_date_suggestion/index.ts
@@ -4,7 +4,7 @@
 import {connect} from 'react-redux';
 
 import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
-import {getUserTimezone} from 'mattermost-redux/selectors/entities/timezone';
+import {makeGetUserTimezone} from 'mattermost-redux/selectors/entities/timezone';
 
 import {getCurrentDateForTimezone} from 'utils/timezone';
 import {areTimezonesEnabledAndSupported} from 'selectors/general';
@@ -14,26 +14,30 @@ import {GlobalState} from 'types/store';
 
 import SearchDateSuggestion from './search_date_suggestion';
 
-function mapStateToProps(state: GlobalState) {
-    const currentUserId = getCurrentUserId(state);
-    const userTimezone = getUserTimezone(state, currentUserId);
-    const locale = getCurrentLocale(state);
+function makeMapStateToProps() {
+    const getUserTimezone = makeGetUserTimezone();
 
-    const enableTimezone = areTimezonesEnabledAndSupported(state);
+    return (state: GlobalState) => {
+        const currentUserId = getCurrentUserId(state);
+        const userTimezone = getUserTimezone(state, currentUserId);
+        const locale = getCurrentLocale(state);
 
-    let currentDate;
-    if (enableTimezone) {
-        if (userTimezone.useAutomaticTimezone) {
-            currentDate = getCurrentDateForTimezone(userTimezone.automaticTimezone);
-        } else {
-            currentDate = getCurrentDateForTimezone(userTimezone.manualTimezone);
+        const enableTimezone = areTimezonesEnabledAndSupported(state);
+
+        let currentDate;
+        if (enableTimezone) {
+            if (userTimezone.useAutomaticTimezone) {
+                currentDate = getCurrentDateForTimezone(userTimezone.automaticTimezone);
+            } else {
+                currentDate = getCurrentDateForTimezone(userTimezone.manualTimezone);
+            }
         }
-    }
 
-    return {
-        currentDate,
-        locale,
+        return {
+            currentDate,
+            locale,
+        };
     };
 }
 
-export default connect(mapStateToProps)(SearchDateSuggestion);
+export default connect(makeMapStateToProps)(SearchDateSuggestion);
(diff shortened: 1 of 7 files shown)


packages/mattermost-redux/src/selectors/entities/timezone.ts
```diff
@@ -1,10 +1,16 @@
...
+
+import {getUsers} from 'mattermost-redux/selectors/entities/users';
+
 import {GlobalState} from 'mattermost-redux/types/store';
+import {IDMappedObjects} from 'mattermost-redux/types/utilities';
+import {UserProfile} from 'mattermost-redux/types/users';
 
-export function getUserTimezone(state: GlobalState, id: string) {
-    const profile = state.entities.users.profiles[id];
+export function getUserTimezone(users: IDMappedObjects<UserProfile>, id: string): { useAutomaticTimezone: boolean; automaticTimezone: string; manualTimezone: string } {
```
BenCookie95, line 12: Nit: Do you want to make the return type into a Type?

BenCookie95 (commented):
Just curious but do you have metrics on the performance improvement?

packages/mattermost-redux/src/selectors/entities/timezone.ts
jwilander: Yeah, that's probably better. I'll do that

jwilander:
@BenCookie95 no, don't have any metrics. The main reason I started looking at this was because every character typed into the post input box was triggering re-renders in some places where timezones were being pulled and not memoized. That code was part of the status DND timer code but has been temporarily reverted. This will help when that goes back in, however.

jwilander:
Once https://github.com/mattermost/mattermost-webapp/pull/8189 goes in we'll be able to better see the effect these types of changes have

BenCookie95 (approved):
LGTM!

components/suggestion/search_date_suggestion/index.ts
```diff
@@ -14,26 +14,30 @@ import {GlobalState} from 'types/store';
...
-    const currentUserId = getCurrentUserId(state);
-    const userTimezone = getUserTimezone(state, currentUserId);
-    const locale = getCurrentLocale(state);
+function makeMapStateToProps() {
+    const getUserTimezone = makeGetUserTimezone();
 
-    const enableTimezone = areTimezonesEnabledAndSupported(state);
+    return (state: GlobalState) => {
+        const currentUserId = getCurrentUserId(state);
+        const userTimezone = getUserTimezone(state, currentUserId);
```
hmhealey, line 22: 0/5, but it might be helpful to define a `getCurrentTimezone` since I imagine we'll need that more often than we would another user's timezone. It would also reduce the number of copies of this selector by a fair bit.

packages/mattermost-redux/src/selectors/entities/timezone.ts
```diff
@@ -1,10 +1,16 @@
...
+
+import {getUsers} from 'mattermost-redux/selectors/entities/users';
+
 import {GlobalState} from 'mattermost-redux/types/store';
+import {IDMappedObjects} from 'mattermost-redux/types/utilities';
+import {UserProfile, UserTimezone} from 'mattermost-redux/types/users';
 
-export function getUserTimezone(state: GlobalState, id: string) {
-    const profile = state.entities.users.profiles[id];
+export function getUserTimezone(users: IDMappedObjects<UserProfile>, id: string): UserTimezone {
```
hmhealey, line 12: Why not make this just take the singular user we need? That seems like it would make `makeGetUserTimezone` recalculate less often, and we'd probably have the user we want more often anyway.

components/timestamp/index.ts
```diff
@@ -21,29 +21,33 @@ type Props = {
...
-    if (areTimezonesEnabledAndSupported(state)) {
-        timeZone = getUserCurrentTimezone(userTimezone ?? getUserTimezone(state, currentUserId)) ?? undefined;
-    }
+        let timeZone: TimestampProps['timeZone'];
+        let hourCycle: TimestampProps['hourCycle'];
+        let hour12: TimestampProps['hour12'];
 
-    const useMilitaryTime = getBool(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.USE_MILITARY_TIME, false);
+        if (areTimezonesEnabledAndSupported(state)) {
+            timeZone = getUserCurrentTimezone(userTimezone ?? getUserTimezone(state, currentUserId)) ?? undefined;
```
hmhealey, line 35: What's the relationship between `getUserCurrentTimezone` and `getUserTimezone` now? Maybe that's something to solve in another PR, though since it doesn't seem to be a performance concern

hmhealey (changes requested):

components/timestamp/index.ts
hmhealey: Oh, `getUserTimezone` returns the object with the additional information about the user's timezone settings while `getUserCurrentTimezone` breaks that down into their actual timezone which can be used for formatting Dates

packages/mattermost-redux/src/selectors/entities/timezone.ts
jwilander: Yeah, that's probably better. I was just looking to change the signature as little as possible

jwilander:
Closing since it's out of date and I don't have time to work on it near future

PR #8259, opened 2021-06-11
MM-36389 Prevent unnecessary Post children components re-rendering due to comment count propagation

jwilander:
#### Summary
The way the comment count was getting propagated through the many components of the app was causing a lot of unnecessary re-renders which could be impacting performance, especially when there are long threads visible in the channel. This removes a lot of that propagation by hooking components directly up to the redux store where necessary.

I also removed `makeGetReplyCount` since it was practically a duplicate of `makeGetCommentCountForPost`.

Probably easiest to review commit by commit.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36389

#### Release Note

```release-note
NONE
```


components/post_view/post_info/post_info.tsx
```diff
@@ -194,14 +194,13 @@ export default class PostInfo extends React.PureComponent<Props, State> {
...
         const showCommentIcon = fromAutoResponder ||
-        (!isSystemMessage && (isMobile || hover || (!post.root_id && Boolean(this.props.replyCount)) || this.props.isFirstReply));
+        (!isSystemMessage && (isMobile || hover || (!post.root_id && Boolean(this.props.hasReplies)) || this.props.isFirstReply));
         const commentIconExtraClass = isMobile ? '' : 'pull-right';
         let commentIcon;
         if (showCommentIcon) {
             commentIcon = (
                 <CommentIcon
                     handleCommentClick={this.props.handleCommentClick}
-                    commentCount={collapsedThreadsEnabled ? undefined : this.props.replyCount}
```
koox00, line 204: When CRT is "on" we do not show the comment count on the icon, 
maybe we should move this logic to `components/post_view/comment_icon/index.ts`

koox00 (commented):
Awesome!
One comment about CRT on.

packages/mattermost-redux/src/selectors/entities/posts.ts
```diff
@@ -391,13 +393,13 @@ export function makeGetProfilesForThread(): (state: GlobalState, props: {rootId:
     );
 }
 
-export function makeGetCommentCountForPost(): (state: GlobalState, props: {post: Post}) => number {
+export function makeGetCommentCountForPost(): (state: GlobalState, post: Post) => number {
```
koox00, line 396: :heart:

components/post_view/post_info/post_info.tsx
jwilander: Good catch. Let me check if that's needed

koox00 (approved):

packages/mattermost-redux/src/selectors/entities/posts.ts
```diff
@@ -391,13 +393,13 @@ export function makeGetProfilesForThread(): (state: GlobalState, props: {rootId:
...
     return createSelector(
         getAllPosts,
-        (state: GlobalState, {post}: {post: Post}) => state.entities.posts.postsInThread[post ? post.id : ''] || [],
-        (state, props) => props,
-        (posts, postsForThread, {post: currentPost}) => {
-            if (!currentPost) {
+        (state: GlobalState, post: Post) => state.entities.posts.postsInThread[post ? post.root_id || post.id : ''] || [],
+        (state, post: Post) => post,
+        (posts, postsForThread, post) => {
```
hmhealey, line 401: Nothing that needs to change here, but I'm surprised that we can get away with calling this parameter `post` and then using a local variable called `post` in the `forEach` below

utils/post_utils.jsx
```diff
@@ -484,22 +484,6 @@ export function getNewMessageIndex(postListIds) {
     );
 }
 
-export function makeGetReplyCount() {
```
hmhealey, line 487: I very much appreciate removing this since selectors being defined in `utils` kind of annoys me since they have their own folder already 😛

hmhealey (approved):

PR #8263, opened 2021-06-12
[MM-18945] Migrate 'components/activity_log_modal' and associated tests to TypeScript

majidsajadi:
#### Summary

migrated `activity_log_modal` to typescript.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/16743

#### Release Note

```release-note
NONE
```


components/activity_log_modal/activity_log_modal.tsx
```diff
@@ -82,29 +89,24 @@ export default class ActivityLogModal extends React.PureComponent {
     }
 
     render() {
-        let content;
-        if (this.props.sessions.loading) {
-            content = <LoadingScreen/>;
```
devinbinnie, line 87: I don't think we should be removing this logic here. I think it's causing some of the tests to fail by removing this.

components/activity_log_modal/activity_log_modal.test.jsx
```diff
@@ -4,9 +4,10 @@
 import React from 'react';
 import {shallow} from 'enzyme';
 
+import ActivityLogModal from 'components/activity_log_modal/activity_log_modal';
+
 import {General} from 'mattermost-redux/constants';
 
-import ActivityLogModal from 'components/activity_log_modal/activity_log_modal.jsx';
```
devinbinnie, line 9: Can we convert the test file to TS as well?

devinbinnie (changes requested):

components/activity_log_modal/activity_log_modal.tsx
```diff
@@ -2,64 +2,71 @@
...
-    static propTypes = {
+export type Props = {
 
-        /**
-         * The current user id
-         */
-        currentUserId: PropTypes.string.isRequired,
+    /**
+ * The current user id
+ */
```
agarciamontoro, line 18: Can we fix this indent here? I don't know why this is not caught by the linter :thinking: 
```suggestion
    /**
     * The current user id
     */
```

components/activity_log_modal/activity_log_modal.tsx
```diff
@@ -2,64 +2,71 @@
...
-            /**
-             * Function to revoke a particular session
-             */
-            revokeSession: PropTypes.func.isRequired,
-        }).isRequired,
+        revokeSession: (userId: string, sessionId: string) => Promise<{ data: boolean }>;
+    };
+}
+
+type State ={
```
agarciamontoro, line 50: Minor linting problem here:
```suggestion
type State = {
```

agarciamontoro (changes requested):
Thank you for your contribution, @majidsajadi!

Apart from @devinbinnie's comments, I've added a couple more minor ones.

agarciamontoro:
Hi, @majidsajadi, how can we help to move this PR forward? Thanks again for your work!

components/activity_log_modal/activity_log_modal.tsx
majidsajadi: in the index we bind sessions props to this method:
 
 `sessions: getUserSessions(state),`
 
 which returns an array of type `Sessions`.
 
I don't think the code would reach the `if` block. let me know if I'm wrong. @devinbinnie

agarciamontoro: I took a look at the reducer and it looks like it always returns an array, but I may be missing something. I'll defer to @devinbinnie here.

components/activity_log_modal/activity_log_modal.test.jsx
agarciamontoro: I think we still need to convert this test file to Typescript :)

agarciamontoro (changes requested):
Thanks for addressing the linting changes! Devin's requested changes are still pending, though, if I'm not mistaken

components/activity_log_modal/activity_log_modal.tsx
devinbinnie: It does look that way, weird that this code was never removed. If we're removed this code, can we also remove the tests that are looking for this code? Thanks!

components/activity_log_modal/activity_log_modal.test.jsx
```diff
@@ -21,18 +20,6 @@ describe('components/ActivityLogModal', () => {
         locale: General.DEFAULT_LOCALE,
     };
 
-    test('should match snapshot', () => {
-        const wrapper = shallow(
-            <ActivityLogModal {...baseProps}/>,
-        );
-        expect(wrapper).toMatchSnapshot();
-        expect(wrapper.find(LoadingScreen).exists()).toBe(false);
```
devinbinnie, line 29: I don't think we need to completely remove the snapshot tests, we can just update the snapshots and not worry about checking if the `LoadingScreen` is there.

components/activity_log_modal/activity_log_modal.test.jsx
devinbinnie: @majidsajadi We're still looking to have this test file converted to TypeScript. Let us know if you need any help with that.

devinbinnie (changes requested):
Still one comment to be resolved. Thanks for your work on this!

components/activity_log_modal/activity_log_modal.test.tsx
```diff
@@ -1,13 +1,12 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React from 'react';
+import React, { MouseEvent } from 'react';
```
agarciamontoro, line 4: You're using `React.MouseEvent` below, which is fine, so we can remove this import
```suggestion
import React from 'react';
```

agarciamontoro (changes requested):
Yay, thank you! Just one small thing that is making the linter complain.

devinbinnie (approved):
Thanks @majidsajadi, this is looking good now.
We'll need to resolve the conflicts in the branch but otherwise we're good to go.

agarciamontoro (approved):
Looks perfect! Thank you for all your work here, @majidsajadi :)

majidsajadi:
Thanks, sorry i took so long to resolve the issues, i was really busy.

agarciamontoro:
@majidsajadi, no problem at all! Life always gets in the way :)

devinbinnie (approved):

furqanmlk (approved):
@majidsajadi Can you please add QA steps ?

PR #8267, opened 2021-06-14
MM-35769: CRT, shows unread indicator

koox00:
#### Summary

Shows unread indicator when on global threads page as well.

We guard against empty strings for the value of current channel id, but in the case of threads, `currentChannelId` is indeed an empty string. This prevented showing the unread top/bottom indicators, removing the guard on empty strings fixes the issue.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35769

#### Release Note

```release-note
NONE
```


anurag6713 (approved):

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -176,7 +176,7 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
     }
 
     getFirstUnreadChannelFromChannelIdArray = (channelIds: string[]) => {
-        if (!this.props.currentChannelId) {
+        if (typeof this.props.currentChannelId !== 'string') {
```
hmhealey, line 179: Could we maybe remove this block entirely? I can't see any reason that this would cause problems if currentChannelId is undefined or anything. I wonder if we were originally passing in `currentChannel` as an object and this was a null check before using `currentChannel.id` below

hmhealey (changes requested):

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
hmhealey: Yeah, that's what it was. This code originally null checked `this.props.currentChannel` and after going back and forth a few times between a channel and its ID, we eventually settled on passing `currentChannelId`, so this check can just be removed.

calebroseland (approved):

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
Verified more unread toast logic on the sidebar per ticket description.  
- Toast is not present if there are no bolded channels on the sidebar regardless of user  viewing global threads or not, or having unread threads.
- Toast is present when there are bolded channels in both cases, when viewing global threads or when viewing other channels.

hmhealey (approved):

hmhealey:
Echoing https://github.com/mattermost/mattermost-webapp/pull/8272#issuecomment-862696362, I'll merge this as well so we can get it onto community-daily

PR #8271, opened 2021-06-15
Use reply count from post if thread not tracked in postsInThread reducer

jwilander:
#### Summary
Quick fix for CRT where unfollowed posts didn't have reply counts showing up in the center channel.

#### Release Note

```release-note
NONE
```


ashishbhate (approved):

jwilander:
Going to merge this as is since @koox00 and I already talked about it in a DM and it's fixing a glaring issue

PR #8272, opened 2021-06-16
MM-36218,MM-36219: CRT thread read flashes

koox00:
#### Summary

New messages that arrive while the thread is open are marked as read and
the indicators don't flash anymore.

Furthermore if a thread is manually marked as unread it should stay
unread even if new messages arrive while it's open.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36218
https://mattermost.atlassian.net/browse/MM-36219

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/threading/global_threads/thread_menu/thread_menu.tsx
+++ b/components/threading/global_threads/thread_menu/thread_menu.tsx
@@ -11,7 +11,7 @@ import {UserThread} from 'mattermost-redux/types/threads';
 import {get} from 'mattermost-redux/selectors/entities/preferences';
 
 import {setThreadFollow, updateThreadRead} from 'mattermost-redux/actions/threads';
-import {updateThreadLastOpened} from 'actions/views/threads';
+import {manuallyMarkThreadAsUnread} from 'actions/views/threads';
 
 import {
     flagPost as savePost,
@@ -58,6 +58,20 @@ function ThreadMenu({
     } = useThreadRouting();
 
     const isSaved = useSelector((state: GlobalState) => get(state, Preferences.CATEGORY_FLAGGED_POST, threadId, null) != null);
+
+    const handleReadUnread = useCallback(() => {
+        const lastViewedAt = hasUnreads ? Date.now() : unreadTimestamp;
+
+        dispatch(manuallyMarkThreadAsUnread(threadId, lastViewedAt));
+        dispatch(updateThreadRead(currentUserId, currentTeamId, threadId, lastViewedAt));
+    }, [
+        currentUserId,
+        currentTeamId,
+        threadId,
+        hasUnreads,
+        updateThreadRead,
+        unreadTimestamp,
+    ]);
 
     return (
         <MenuWrapper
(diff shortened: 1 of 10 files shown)


jwilander (approved):

esethna:
@hmhealey @jgilliam17 please help with priority reviews if possible

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified no bolding of the thread when replies are received while current user is viewing the thread item (MM-36218)
- Verified when thread is open and manually marked as unread, it stays unread as new replies to this thread are received.(MM-36219)

components/threading/global_threads/thread_menu/thread_menu.tsx
```diff
@@ -59,6 +59,20 @@ function ThreadMenu({
 
     const isSaved = useSelector((state: GlobalState) => get(state, Preferences.CATEGORY_FLAGGED_POST, threadId, null) != null);
 
+    const handleReadUnread = useCallback(() => {
+        const lastViewedAt = hasUnreads ? Date.now() : unreadTimestamp;
+
+        dispatch(manuallyMarkThreadAsUnread(threadId, lastViewedAt));
+        dispatch(updateThreadRead(currentUserId, currentTeamId, threadId, lastViewedAt));
+    }, [
+        currentUserId,
```
hmhealey, line 68: I don't know if it'll have as much impact on this component, but I've been looking a bit at where we pass props into a component purely for calling actions because that leads to the component potentially re-rendering more frequently for data that isn't actually needed for rendering. I'm not sure if this is the case here, but it might be something to keep in mind.

For example, assuming we weren't using `currentUserId` or some of the other props for rendering, we could refactor this into its own action to look more like
```javascript
function markThreadAsUnread(threadId) {
    return (dispatch, getState) => {
        const state = getState();
        const currentTeamId = getCurrentTeamId(state);
        const lastViewedAt = /*...*/;
        
        dispatch(manuallyMarkThreadAsUnread(threadId, lastViewedAt));
        dispatch(updateThreadRead(currentUserId, currentTeamId, threadId, lastViewedAt));
    };
}

function ThreadMenu({threadId}) {
    // ...
}
```
to reduce the props passed into ThreadMenu and keep it down to the ones that are only used for rendering

reducers/views/threads.ts
```diff
@@ -31,7 +31,24 @@ export const lastViewedAt = (state: ViewsState['threads']['lastViewedAt'] | Reco
     return state;
 };
 
+export function manuallyUnread(state: ViewsState['threads']['manuallyUnread'] | Record<string, unknown> = {}, action: GenericAction) {
```
hmhealey, line 34: Why is the `Record<string, unknown>` needed here? It seems like that makes the typing information here much less useful. If nothing else, I think the `unknown` could be `boolean`

hmhealey (approved):
I have a couple suggestions for possible improvements, but neither of them will block this

jwilander:
Going to merge to get this in but @koox00 feel free to create follow-up PRs/tickets if you'd like to address Harrison's comments

PR #8277, opened 2021-06-16
MM-36490 Add github shortcodes for backwards compatibility for emoji aliases

jwilander:
#### Summary
This will fix emojis where the name changed by adding the old GitHub shortcodes as aliases. For example `:woman_facepalming:` should work again now that it is aliased to `:woman-facepalming:`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36490

#### Release Note

```release-note
NONE
```


jwilander:
Waiting on https://github.com/mattermost/mattermost-server/pull/17741 then I need to submit a server PR to update `model/emoji_data.go`

Willyfrog (changes requested):
this will map all the un-skinned emojis, but any emoji with a skin name will be left out so 🤦‍♀️ will be mapped correctly, but :woman_facepalming_dark_skin_tone: will not exist
substituting short_name and short_names before line 120 will probably solve that, as the function for skin variations does add the different skin tones for all possible aliases

jwilander:
@Willyfrog do we want to add all those aliases for the skin tones? I think we only need this aliases for backwards compatibility reasons, right?

Willyfrog:
> @Willyfrog do we want to add all those aliases for the skin tones? I think we only need these aliases for backwards compatibility reasons, right?

We have those in community (stable), they doesn't show in the emoji picker, but there are skin tones available if you know the shortcode.

jwilander:
Oh, so we shipped some emoji skin tones in a release already? Ok I'll update it per your suggestion then

build/emoji/emoji_gen.mjs
```diff
@@ -116,6 +117,39 @@ function genSkinVariations(emoji, index, nextOrder) {
...
+        if (emoji.short_name === newShortName) {
+            return;
+        }
+
+        // Don't duplicate short names
+        if (shortNames.includes(newShortName)) {
+            return;
+        }
+
+        // Don't use short names that are already taken by other emoji
```
Willyfrog, line 145: nice one

Willyfrog (approved):
LGTM!

hmhealey:
Are we supporting both hyphens and underscores in this case? I'd rather just keep the old aliases and not add new ones

jwilander:
That's a good question. I was under the impression that we wanted to use hyphens because that's the new default for emoji 13. If that's not the case and we can stick with underscores then I think that's what we should do. I'll schedule us a quick call to chat about it today

cc @Willyfrog

Willyfrog:
> That's a good question. I was under the impression that we wanted to use hyphens because that's the new default for emoji 13. If that's not the case and we can stick with underscores then I think that's what we should do. I'll schedule us a quick call to chat about it today
> 
> cc @Willyfrog

The original library didn't give a choice, but now with the mapping, we can choose one, the other or both, but let's meet later and discuss it.

jwilander:
Closing this PR as @hmhealey is going to look into switching us to a more backwards compatible emoji standard

PR #8278, opened 2021-06-17
MM-36391: navigates to threads on shortcut ALT+ARROW

koox00:
#### Summary

ALT(+SHIFT)+UP/DOWN switches channels in the left sidebar,
but it won't switch to global threads when CRT is on.

This commit enables switching to global threads as well.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36391

#### Release Note

```release-note
NONE
```


esethna (approved):
Nice! Checked the alt+arrow and alt+shift+arrow

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified using alt+up arrow(for all channels) and alt+shift+up arrow(for unreads) to navigate through channels on the LHS now includes Threads category (when CRT is enabled).

actions/views/threads.ts
```diff
@@ -34,3 +44,17 @@ export function manuallyMarkThreadAsUnread(threadId: string, lastViewedAt: numbe
...
     ]);
 }
+
+export function switchToGlobalThreads() {
+    return (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState() as GlobalState;
+        const teamUrl = getCurrentRelativeTeamUrl(state);
+
+        if (getIsRhsOpen(state)) {
+            dispatch(suppressRHS);
```
koox00, line 54: I would like to move all the `suppressRHS` dispatch, from individual actions to the `global_threads`/`componentDidMount` lifecycle method, so we don't have to care about it at all.

One thing though, it will introduce a slight delay on closing the RHS while navigating to global threads,
is that OK from a UX point of view @matthewbirtch?

matthewbirtch: @koox00 I guess that would depend on how much of a delay we're talking about. How long are we talking?

koox00: just after the global threads are shown, now it hides it just before.

matthewbirtch: @koox00 This is probably ok. Can I test it out on this PR?

esethna: @matthewbirtch @koox00 is that change on the test server though? 

We should be careful about anything that delays opening "threads", even minor delays can have a noticeable impact on usability (it's why we invested in pre-loading unread channels to minimize the short loading time on channel content when executing a channel switch). We already have a significant delay on opening threads on daily (probably because of the non-virtualized list)

koox00: Sorry... it is not, I didn't had the time today, will do tomorrow morning

matthewbirtch: ok @koox00, let us know when this is in so we can see how it feels

koox00: @esethna @matthewbirtch the change is on the test server :)

esethna: @koox00 looks good to me, it doesn't actually delay opening threads it just delays closing the RHS, so my concerns above aren't valid. Thanks!

esethna (approved):

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -288,12 +291,27 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
...
         if (e.altKey && !e.shiftKey && !e.ctrlKey && !e.metaKey && (Utils.isKeyPressed(e, Constants.KeyCodes.UP) || Utils.isKeyPressed(e, Constants.KeyCodes.DOWN))) {
             e.preventDefault();
 
             const allChannelIds = this.getDisplayedChannelIds();
             const curChannelId = this.props.currentChannelId;
+
+            if (this.props.collapsedThreads) {
+                // threads set channel id to ''
+                // add it to allChannelIds
+                allChannelIds.unshift('');
```
hmhealey, line 312: Maybe it would be a bit clearer to have a `global_threads` constant for use in this file instead of just treating the empty string as Global Threads? I could see more non-channel items being added to the sidebar in the future that would need their own IDs for this.

hmhealey (changes requested):
Looks good for the most part, but I have a suggestion for how we might futureproof some of this

matthewbirtch:
Is there something wrong with the test server? Things are quite broken for me with the above link @koox00. Plus it looks like the test server creation failed

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
calebroseland: +1 for a constant

Also an aside mention that once we get a more stable infrastructure for these non-channel pages, that https://github.com/mattermost/mattermost-webapp/blob/master/components/threading/global_threads/global_threads.tsx#L66-L67 should be removed.

koox00: @hmhealey, @calebroseland while I agree 100% with what you are both saying, 
I don't see how we can use a constant when we don't have a concrete architecture for non-channel sidebar items. 
The issue I see is that we need to check against `currentChannelId` (which is `''`) for global threads.

I'll attempt a commit to support non-channel pages

koox00:
> Is there something wrong with the test server? Things are quite broken for me with the above link @koox00. Plus it looks like the test server creation failed

@matthewbirtch can you check again?

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
koox00: @hmhealey, @calebroseland I've added a commit to address the comments, can you take a look and share your thoughts?

matthewbirtch:
> @matthewbirtch can you check again?

yep, seems to be resolved!

matthewbirtch (approved):
Verified that Alt-Up, Alt-Down, Alt-Shift-Up, Alt-Shift-Down now includes threads

jwilander (approved):
LGTM

calebroseland (approved):

esethna:
@hmhealey can you review and merge if looks good?

hmhealey (changes requested):
Sorry, I didn't expect changing that would require changing the `currentChannelId` everywhere in the sidebar. I just meant we could add a constant in `SidebarChannelList` to that's added to the array by `navigateChannelShortcut` and handled by `navigateByChannelId`.

I'd much rather stick with what we had before to avoid complicating the rest of the sidebar code for this. The management of currentChannelId and currentTeamId is a bit funny as-is, so I'd rather not add another field for the selected item in the sidebar.

koox00:
> Sorry, I didn't expect changing that would require changing the `currentChannelId` everywhere in the sidebar. I just meant we could add a constant in `SidebarChannelList` to that's added to the array by `navigateChannelShortcut` and handled by `navigateByChannelId`.

@hmhealey I don't see how to do that without radical changes, 
since we need to check against `currChannelId` to get the current index so we can determine the next index.

```js
if (allChannelIds[i] === curChannelId) {
    curIndex = i;
}
// ...
```

> I'd much rather stick with what we had before to avoid complicating the rest of the sidebar code for this. The management of currentChannelId and currentTeamId is a bit funny as-is, so I'd rather not add another field for the selected item in the sidebar.

I agree with the sentiment, I'll revert the commits no problem.

hmhealey (approved):

hmhealey:
Yeah, I feel like we need to rethink how our sidebar interacts with routing on that. Since the currentChannelId is stored independent of what is in the URL bar, it ends up causing these weird cases where the sidebar should be able to be aware of the route for things like the Threads view, but it isn't currently without some more major reworking.

PR #8279, opened 2021-06-17
MM-36542: getOpenThreadId should consider suppressed RHS

koox00:
#### Summary

Fixes a bug when suppressing RHS via going to global threads,
and then receiving new replies to the thread open in the suppressed RHS,
kept marking them as read.

This commit changes `getOpenThreadId` to consider if RHS is open or not.
Also mounts/unmounts components in the RHS with account to RHS hidden/visible,
this change was made because right now we can suppress the RHS and that action
does not clear the rhsState nor the selectedPostId etc..

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36542

#### Release Note
```release-note
NONE
```

### DIFF
--- a/selectors/views/threads.test.ts
+++ b/selectors/views/threads.test.ts
@@ -37,28 +37,6 @@ describe('selectors/views/threads', () => {
         },
     }) as unknown as GlobalState;
 
-    describe('getOpenThreadId', () => {
-        test('should return selected post id if it exists and rhs is open', () => {
-            const state = makeState(null, 'selected_post_id');
-            expect(selectors.getOpenThreadId(state)).toBe('selected_post_id');
-        });
-
-        test('should return null if selected post id exists but rhs is closed ', () => {
-            const state = makeState(null, 'selected_post_id', false);
-            expect(selectors.getOpenThreadId(state)).toBe(null);
-        });
-
-        test('should return selected thread id is selected post doesn\'t exist', () => {
-            const state = makeState('selected_thread_id', '');
-            expect(selectors.getOpenThreadId(state)).toBe('selected_thread_id');
-        });
-
-        test('should return null when neither selected post nor selected thread exist', () => {
-            const state = makeState(null, '');
-            expect(selectors.getOpenThreadId(state)).toBe(null);
-        });
-    });
-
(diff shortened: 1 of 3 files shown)


selectors/views/threads.test.ts
```diff
@@ -31,16 +31,23 @@ describe('selectors/views/threads', () => {
             },
             rhs: {
                 selectedPostId,
+                isSidebarOpen,
```
michelengelen, line 34: shouldn't we call it `isRhsOpen` to be consistent with the naming?

michelengelen (approved):
LGTM ... just a small remark with no real impact! ✨

selectors/views/threads.test.ts
koox00: I see what you mean, It would make sense to me to name it `isOpen` since it's in `rhs` reducer. 
I guess it's called `isSidebarOpen` to differentiate from  `isMenuOpen` which is also in `rhs` reducer.

calebroseland (changes requested):
While we do hide RHS when going to Threads, it is still possible to have two threads open at the same time via search or at-mentions. I'd propose just letting `isThreadOpen` check both locations.
Something like [MM-36542-review.patch.zip](https://github.com/mattermost/mattermost-webapp/files/6679433/MM-36542-review.patch.zip) perhaps?

koox00:
> While we do hide RHS when going to Threads, it is still possible to have two threads open at the same time via search or at-mentions. I'd propose just letting `isThreadOpen` check both locations.!

Oh yes! thanks @calebroseland 

> Something like [MM-36542-review.patch.zip](https://github.com/mattermost/mattermost-webapp/files/6679433/MM-36542-review.patch.zip) perhaps?

yep looks good!

michelengelen (approved):

jgilliam17 (approved):
Thank you @koox00 
Tested, looks good to merge.
- Verified receiving new replies on a suppressed thread, while user is viewing Threads, is marking the thread as unread. 
- Verified when leaving global threads the RHS is visible again, and the thread is now marked as read as if you just opened the thread.

calebroseland (approved):

PR #8281, opened 2021-06-18
MM-36561: fixes loader shown on socket reconnect

koox00:
#### Summary

The loader was shown in ThreadViewer upon websocket reconnection.
This commit fixes that, so the loader now shows only on mount and upon changing selected thread.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36561

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/threading/thread_viewer/thread_viewer.tsx
+++ b/components/threading/thread_viewer/thread_viewer.tsx
@@ -311,7 +311,7 @@ export default class ThreadViewer extends React.Component<Props, State> {
     // scrolls to either bottom or new messages line
     private onInit = async (reconnected = false): Promise<void> => {
         if (reconnected || this.morePostsToFetch()) {
-            this.setState({isLoading: true});
+            this.setState({isLoading: !reconnected});
             await this.props.actions.getPostThread(this.props.selected.id, !reconnected);
         }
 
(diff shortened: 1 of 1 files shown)


components/threading/thread_viewer/thread_viewer.tsx
```diff
@@ -311,15 +311,15 @@ export default class ThreadViewer extends React.Component<Props, State> {
     // scrolls to either bottom or new messages line
     private onInit = async (reconnected = false): Promise<void> => {
         if (reconnected || this.morePostsToFetch()) {
-            this.setState({isLoading: true});
+            this.setState({isLoading: !reconnected});
```
larkox, line 314: Is there any case where both are true? (and therefore, it should show the loader?

larkox (approved):
LGTM, but I am not familiar with the code.

Just a small question to verify.

components/threading/thread_viewer/thread_viewer.tsx
koox00: A little background: `onInit` gets called on 3 occasions: mount / `selectedTheadChanged` / reconnect.
It should show the loader at mount and `selectedThreadChanged`, and only if the thread is not wholly loaded already. 
So it would be `true` everytime it gets to be set with the exception of re-connection.

anurag6713 (approved):

jgilliam17 (approved):
Thank you @koox00 
Tested, looks good to merge.
- Verified loader is not shown on the thread viewer on websocket reconnect.

PR #8286, opened 2021-06-20
[MM-18496] Private channels should appear in channel link autocomplete

arvinDarmawan:
Summary
This code has dependency with PR https://github.com/mattermost/mattermost-server/pull/17804

Ticket Link
Jira card: https://mattermost.atlassian.net/browse/MM-18496
Github issue: https://github.com/mattermost/mattermost-server/issues/16245

```release-note
Release Note
- Change icon from globe to lock for private channel  
```


hmhealey (changes requested):
This looks good to me code-wise assuming no further changes are needed depending on the server code.

Would you mind undoing the changes to package-lock.json though? It seems you're on an older version of NPM (6.x probably) which caused a bunch of stuff to be removed from the lockfile when you regenerated it. Theoretically, the one generated by NPM 7 should be backwards compatible, but if that's not working, you might want to upgrade.

arvinDarmawan:
Hi @hmhealey ,  I already revert it.

michelengelen:
> Theoretically, the one generated by NPM 7 should be backwards compatible, but if that's not working, you might want to upgrade.

@hmhealey it is not working that well ... I had some problems in the past as well with it on the compass-components repo.

components/suggestion/channel_mention_provider.jsx
```diff
@@ -20,7 +20,13 @@ class ChannelMentionSuggestion extends Suggestion {
...
 
         const channelName = item.channel.display_name;
-        const channelIcon = (<span className='suggestion-list__icon suggestion-list__icon--large'><i className='icon icon--no-spacing icon-globe'/></span>);
+
+        let channelIcon;
+        if (item.channel.type === Constants.OPEN_CHANNEL) {
+            channelIcon = (<span className='suggestion-list__icon suggestion-list__icon--large'><i className='icon icon--no-spacing icon-globe'/></span>);
+        } else if (item.channel.type === Constants.PRIVATE_CHANNEL) {
+            channelIcon = (<span className='suggestion-list__icon suggestion-list__icon--large'><i className='icon icon--no-spacing icon-lock'/></span>);
+        }
```
michelengelen, line 29: I am not a fan of repeated code-blocks in if-clauses. I would suggest using `clsx` for this specific use-case, since the only difference is the icon class.

But since we are not using `clsx` in the webapp I would suggest something like this:

```suggestion
        const channelIcon = (
            <span className='suggestion-list__icon suggestion-list__icon--large'>
                <i className=`icon icon--no-spacing icon-${item.channel.type === Constants.OPEN_CHANNEL ? 'globe' : 'lock'}`/>
            </span>
        );
```

michelengelen (changes requested):
I would suggest a slight improvement here, if you don't mind @arvinDarmawan

components/suggestion/channel_mention_provider.jsx
arvinDarmawan: Hi @michelengelen , I already change it.

hmhealey (approved):
Thanks for reverting that. Looks good now!

hmhealey:
> @hmhealey it is not working that well ... I had some problems in the past as well with it on the compass-components repo.

In that case, I guess we should start recommending everyone upgrade then. It was going to be necessary once we start using workspaces for things like compass-components anyway, so I don't think there's any issue with making people use it now.

michelengelen (approved):
LGTM

michelengelen:
> > @hmhealey it is not working that well ... I had some problems in the past as well with it on the compass-components repo.
> 
> In that case, I guess we should start recommending everyone upgrade then. It was going to be necessary once we start using workspaces for things like compass-components anyway, so I don't think there's any issue with making people use it now.

yeah, we should probably do that.

michelengelen:
cc/ @jgilliam17 for QA review

hmhealey:
FYI, for testing this, it will rely on the server changes being in to see private channels in the autocomplete. The webapp changes are just for changing the icon used for private channels, so it's not a big deal to test those. The server changes will make the private channels actually appear in the dropdown.

jgilliam17 (approved):
Thank you @arvinDarmawan.
This is good to merge based on the testing tip from - Thanks Harrison!
QA will test on the related server PR.

PR #8287, opened 2021-06-21
MM-36590 Re-add old emoji aliases

hmhealey:
additional_shortnames.json contains a mapping from the new shortnames to all of the old ones that we used to support. The script to generate that list is kind of a mess, so I'm not checking it in since hopefully we'll never need to do that again. The biggest changes are that we're re-adding full name aliases for countries (like `:canada:` instead of the `:flag-ca:` used by the new emoji library) and that there are different names for a lot of the people emojis (like `:man_astronaut:` instead of `:male-astronaut:`).

There are still going to be some changes with the new emoji set due to some changes in emojis such as:
- `:cricket:` is now the insect instead of the game which is `:cricket_game:`.
- `:satellite:` is now a satellite in space like GitHub's `:artificial_satellite:`. It was previously a satellite dish.
- We previously had skin toned versions of the family/couple/kiss emojis listed in emoji.json, but the unicode for them was all wrong, so I just left them out since they never rendered correctly anyway. When we figure out how to support skin tones for emojis with multiple people in them, we can re-add them properly.
- There were some other changes like cricket and satellite that aren't really worth mentioning.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36590

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/17806

#### Release Note
```release-note
NONE
```


Willyfrog:
> There are still going to be some changes with the new emoji set due to some changes in emojis such as:
> 
>     * `:cricket:` is now the insect instead of the game which is `:cricket_game:`.
> 
>     * `:satellite:` is now a satellite in space like GitHub's `:artificial_satellite:`. It was previously a satellite dish.

I don't think there is a possible way to treat this unless we want to include an sql migration to change all previous use of the emoji.

>     * We previously had skin toned versions of the family/couple/kiss emojis listed in emoji.json, but the unicode for them was all wrong, so I just left them out since they never rendered correctly anyway. When we figure out how to support skin tones for emojis with multiple people in them, we can re-add them properly.

Family emojis with skin tones should render properly now, in order to try and sort by skin tones I only used the first one appearing in the sense of who is the user, meaning that if you have a skin tone selected it will show you emojis with the first skin tone matching the selected one, but not if it matches the second, third, etc. 

But if you think we should change the approach, we can discuss that.

Willyfrog (approved):
LGTM

hmhealey:
Are the family emoji skin tones something in your new PR? I tried them when working on this PR, and they still had the problem they had before Emoji 13 where they rendered the first man/woman in the emoji instead of the family

hmhealey:
Here's how they currently look on community-daily
Screen Shot 2021-06-21 at 10 33 51 AM
and regular community
Screen Shot 2021-06-21 at 10 33 54 AM

jwilander (approved):

Willyfrog:
> Here's how they currently look on community-daily
> Screen Shot 2021-06-21 at 10 33 51 AM
> and regular community
> Screen Shot 2021-06-21 at 10 33 54 AM

I've just checked the iamcal lib, and it doesn't specify any skins for the family, so I guess we'll have to figure that out in the future

PR #8291, opened 2021-06-21
MM-36592 Fix memoization of getDirectTeammate

hmhealey:
The current version of `getDirectTeammate` returns an empty object in the case where you're not viewing a DM, but that has the usual problem of always returning a new empty object. To fix that, it can now return undefined in that case, so I've added handling for that wherever it might happen.

I also moved it into a selector file since I have a personal vendetta against utils.jsx.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36592

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8292
https://github.com/mattermost/mattermost-webapp/pull/8293

#### Release Note
```release-note
Improved performance of certain components when viewing non-DM channels
```


components/rhs_thread/index.ts
```diff
@@ -25,24 +20,17 @@ function makeMapStateToProps() {
...
             posts = getPostsForThread(state, {rootId: selected.id});
         }
 
-        const previewCollapsed = get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.COLLAPSE_DISPLAY, Preferences.COLLAPSE_DISPLAY_DEFAULT);
-
         return {
             selected,
             channel,
             posts,
-            socketConnectionStatus: socketStatus.connected,
```
hmhealey, line 42: None of these props are actually used by `RhsThread`, so they were safe to just remove.

BenCookie95 (approved):
LGTM!

nevyangelova (approved):
Awesome 🎉

hmhealey:
Skipping QA since there's nothing particular to test here.

PR #8292, opened 2021-06-21
MM-36593 RHS performance improvements

hmhealey:
There's two big changes here:
- The ThreadViewer uses `makeGetPostsForThread` to get its posts, but due to how we store posts, that returns a new array any time a new post is received, loaded, reacted to, etc. Since posts are "immutable", we can use `createIdsSelector` here which shallowly compares the result against the previous one and prevents a new array from being returned if it contains exactly the same posts.
- The SearchResults component constructs a map of channels for use when rendering its children, but it didn't memoize that map, so it would cause constant re-rendering. That's fixed by making the children of SearchResults responsible for getting their own channels from the store.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36593

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8291
https://github.com/mattermost/mattermost-webapp/pull/8293

#### Release Note
```release-note
Minor improvements to performance of posts in the RHS
```


jespino (approved):
LGTM

koox00 (approved):

hmhealey:
Skipping QA since there's nothing really to test

PR #8293, opened 2021-06-21
MM-36595 Reduced re-rendering of Reaction components

hmhealey:
`getSortedUsers` which was used in the mapStateToProps for the Reaction component wasn't memoized, so it caused that component to constantly re-render.

While fixing that, I also moved the sorting and rendering of the tooltip into its own component. `OverlayTrigger` only mounts the tooltip when it's ready to actually show it, so that means that the work of generating the user list and the resulting tooltip now only happens when the tooltip is visible instead of currently how it happens as long as there's a Reaction visible. That also means it should theoretically only happen at most one per render cycle instead of once per Reaction mounted.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36595

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8291
https://github.com/mattermost/mattermost-webapp/pull/8292

#### Release Note
```release-note
Improved performance of components that show reactions on posts
```


jwilander:
@hmhealey a merge from master should fix the type error

koox00 (approved):
Nice!

nevyangelova (approved):
LGTM :)

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified post reaction buttons and tooltips - as before.

PR #8302, opened 2021-06-22
MM-36609 Reduce recomputations of haveIChannelPermission and getMyChannelPermissions selectors during typing and other state changes

jwilander:
#### Summary
This pull request does two things to address the poor performance of the `haveIChannelPermission` and `getMyChannelPermissions` selectors:

1. (first commit) Front loads the role processing into the channels reducers, adding new channel state in redux for `roles`
2. (second commit) Remove unnecessary memoization from `haveIChannelPermission` and `haveITeamPermission` and correctly memoize `getMyChannelPermissions` using a factory

Before the changes these selectors were recomputed a significant amount with low effectiveness:

Screen Shot 2021-06-21 at 3 34 06 PM

And on Firefox `getMyChannelPermissions` was showing as a hotspot in the profiles:

Screenshot 2021-06-21 153757

After these changes, the selectors are much more effective:

Screen Shot 2021-06-22 at 3 46 48 PM

And on Firefox the hotspot looks to be gone:

Screenshot 2021-06-22 155458

With only `haveIChannelPermission` showing far down the list:

Screenshot 2021-06-22 155848

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36609

#### Release Note

```release-note
Some improvements to performance when typing
```


jwilander:
@jgilliam17 can you do a general permissions smoke test? Basically make sure you're able to the normal things in a channel you should always be able to do and check switching to/from channel admin, etc.

packages/mattermost-redux/src/selectors/entities/roles.ts
```diff
@@ -152,29 +140,42 @@ export const getMyTeamPermissions: (state: GlobalState, options: PermissionsOpti
     },
 );
 
-export const getMyChannelPermissions: (state: GlobalState, options: PermissionsOptions) => Set<string> = createSelector(
-    'getMyChannelPermissions',
-    getMyChannelRoles,
-    getRoles,
-    getMyTeamPermissions,
-    (state, options: PermissionsOptions) => options.channel,
```
hmhealey, line 160: Taking a look at the old code, I don't think the `options` object would've actually caused any extra recalculation if these used a proper selector factory. The result function is only re-ran when one of the first n-1 selectors returns a new result. Since `options.channel` is just a string in this case, the result function wouldn't recalculate as long as this was called with an identical object twice in a row.

packages/mattermost-redux/src/utils/channel_utils.ts
```diff
@@ -42,6 +42,10 @@ export function completeDirectChannelInfo(usersState: UsersState, teammateNameDi
     return channel;
 }
 
+export function splitRoles(roles: string): Set<string> {
```
hmhealey, line 45: 0/5, but this could live next to the roles reducer I don't think we need to preemptively export helpers like this for use in other code since it leads to more inter-dependency between files.

hmhealey (approved):
A couple non-blocking comments, but otherwise looks good.

jgilliam17:
Thanks @jwilander 
Seeing few issue (while I wait on E2E results).
1) When logged in as Sysadmin, other users posts and mentions are not received in real time
Screen Shot 2021-06-28 at 2 51 58 PM


2) When editing or deleting others' posts as channel admin, post is not updated for other users, needs refresh or leaving/opening channel again. This delay is not reproducing on master or when editing own posts on the server PR.
Screen Shot 2021-06-28 at 2 35 25 PM

jwilander:
Thanks @jgilliam17 - not sure how this would have caused 1) or 2) but I'll take a look. Sounds more like a websocket connection issue.

jwilander:
@jgilliam17 it seems to be working fine for me on the test server. Can you try again and check the dev console for the sysadmin to see if the websocket is still connected?

jgilliam17:
Thanks @jwilander 
Tested, looks good to merge.
I retested and the sysadmin issues I was seeing before are resolved on the PR server for me too, no longer seeing the delay.
I also retested manually few issues that were flagged by the [cypress test run](https://mm-cypress-report.s3.amazonaws.com/328675383-webapp-pr-8302-webapp-pr-8302-mattermost-enterprise-edition:963c54c/mochawesome.html) and the test cases that differ from the [master test run](https://mm-cypress-report.s3.amazonaws.com/328780962-50af8b2-master-e20-prod-mattermost-enterprise-edition:master/mochawesome.html) are passing on the test server.

jgilliam17 (approved):

Willyfrog (approved):
Looking good, while I agree with @hmhealey that a memoized selector should return the same object, there might be reducers acting on the state triggering new objects with the same values (or the same values regarding what we need), so moving towards simpler structures seems like a good move even if we find the root cause for the cache invalidation of the memoized function.

PR #8304, opened 2021-06-23
[MM-36175] Permalink previews

BenCookie95:
#### Summary
New Permalink preview feature.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36175

#### Screenshots
Screen Shot 2021-06-23 at 10 18 41 AM

#### Related PRs
- https://github.com/mattermost/mattermost-server/pull/17796
- https://github.com/mattermost/enterprise/pull/1005

#### Release Note
```release-note
New permalink preview feature
```


components/post_view/show_more/show_more.tsx
```diff
@@ -32,7 +34,12 @@ export default class ShowMore extends React.PureComponent<Props, State> {
 
     constructor(props: Props) {
         super(props);
-        this.maxHeight = this.props.isAttachmentText ? MAX_ATTACHMENT_TEXT_HEIGHT : MAX_POST_HEIGHT;
+        this.maxHeight = MAX_POST_HEIGHT;
+        if (this.props.attachmentTextOverflowType === 'ellipsis') {
+            this.maxHeight = 100;
+        } else if (this.props.isAttachmentText) {
+            this.maxHeight = MAX_ATTACHMENT_TEXT_HEIGHT;
+        }
```
hmhealey, line 42: Maybe instead of all of this, we pass in maxHeight as a prop? I don't know how it ended up as a field to begin with, but having it potentially derived from multiple props is rather weird.

There's also the possibility of it not being updated right if the props change somehow, although I don't think that can actually happen with how we're using `ShowMore`.

components/post_view/show_more/show_more.tsx
```diff
@@ -18,6 +19,7 @@ type Props = {
     isRHSOpen: boolean;
     text?: string;
     compactDisplay: boolean;
+    attachmentTextOverflowType?: AttachmentTextOverflowType;
```
hmhealey, line 22: 0/5, but maybe we could shorten this name to just `overflowType`?

packages/mattermost-redux/src/actions/posts.ts
```diff
@@ -930,22 +953,48 @@ export function getProfilesAndStatusesForPosts(postsArrayOrMap: Post[]|Map<strin
...
         const userId = post.user_id;
 
+        if (post.metadata && post.metadata.embeds) {
+            post.metadata.embeds.forEach((embed: any) => {
+                if (embed.type === 'permalink' && embed.data) {
+                    if (embed.data.user_id && !profiles[embed.data.user_id] && embed.data.user_id !== currentUserId) {
+                        userIdsToLoad.add(embed.data.user_id);
+                    }
+                    if (embed.data.id && posts && !posts[embed.data.id] && !postsDictionary[embed.data.id]) {
+                        postsToLoad.add(embed.data.id);
```
hmhealey, line 989: Loading these posts asynchronously isn't going to work since I assume the previews aren't a fixed height.

We've had a long history fighting with issues caused by posts changing their height after they are first rendered. It causes a lot of problems around positioning the user in the right place when first loading (such as ensuring the New Messages line is on screen), keeping them in the right place later on (like at the bottom of the channel), and ensuring infinite scrolling works smoothly. Post metadata was originally added mostly to ensure the post components had all the information they needed to size themselves correctly and prevent changing sizes.

I assume this was being done this way to solve the permission issues we identified around not sending posts a user didn't have permissions too, but unfortunately, I think that needs to be addressed on the server side.

packages/mattermost-redux/src/types/posts.ts
```diff
@@ -145,3 +147,10 @@ export declare type OpenGraphMetadata = {
     url?: string;
     images: OpenGraphMetadataImage[];
 };
+
+export declare type PostPreviewMetadata = {
+    id: string;
+    user_id: string;
+    channel_display_name: string;
+    team_name: string;
```
hmhealey, line 155: These names might also be considered confidential as well like, for example, if someone posted a link to a customer channel in a public channel on community.

hmhealey (changes requested):

packages/mattermost-redux/src/types/posts.ts
mkraft: None of the post metadata appears if the user doesn't have permission to view the post.

hmhealey: Oh that's good to hear. I wasn't sure how that was handled since I haven't looked at the server code at all.

components/post_view/post_attachment_container/post_attachment_container.tsx
```diff
@@ -0,0 +1,39 @@
...
+export type Props = {
+    className?: string;
+    children?: JSX.Element;
+    link: string;
+};
+const PostAttachmentContainer = (props: Props) => {
+    const {children, className, link} = props;
+    const history = useHistory();
+    const handleOnClick = useCallback((e) => {
+        const tagName = e.target.tagName;
```
koox00, line 16: Since we are going to add an onClick event on posts in https://github.com/mattermost/mattermost-webapp/pull/8342,
maybe we should also `stopPropagation` here.

components/post_view/post_message_preview/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type Props = {
+    metadata: PostPreviewMetadata;
+}
+
+function mapStateToProps(state: GlobalState, ownProps: Props) {
+    let user = null;
+    let post = null;
+    if (ownProps.metadata.id) {
+        post = getPost(state, ownProps.metadata.id);
```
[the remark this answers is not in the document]
BenCookie95: PostBodyAdditionalContent doesn't actually have `post` in order to pass it down. This `post` is the preview post, not the original post.

I can make `post` optional though and only call it if it isn't in ownProps.

packages/mattermost-redux/src/types/posts.ts
```diff
@@ -27,12 +27,14 @@ export type PostType = 'system_add_remove' |
...
-export type PostEmbedType = 'image' | 'link' | 'message_attachment' | 'opengraph';
+export type PostEmbedType = 'image' | 'link' | 'message_attachment' | 'opengraph' | 'permalink';
+
+export type AttachmentTextOverflowType = 'ellipsis';
 
 export type PostEmbed = {
     type: PostEmbedType;
     url: string;
-    data: Record<string, OpenGraphMetadata>;
+    data?: Record<string, OpenGraphMetadata> | PostPreviewMetadata;
```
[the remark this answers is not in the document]
BenCookie95: I was so confused about why it was setup like this but now I understand. I never knew there was a store in redux with metadata mapped to the url but I don't know why it's like this for the PostEmbed type. There are no instances where the `data` attribute is setup with metadata mapped to the URL. I'm actually going to update this type to:
```
export type PostEmbed = {
    type: PostEmbedType;
    url: string;
    data?: OpenGraphMetadata | PostPreviewMetadata;
}
```

components/post_view/show_more/show_more.tsx
BenCookie95: Made a maxHeight prop

packages/mattermost-redux/src/actions/posts.ts
BenCookie95: We decided to add the whole post object to the metadata. When the webapp receives the posts array we check metadata for each post and if we find a permalink we add the embedded post to the post redux store and remove it from the post object from the metadata (metadata retains the post_id so we can find the post in our redux store).

components/post_view/post_attachment_container/post_attachment_container.tsx
BenCookie95: Done!

components/post_view/post_message_preview/post_message_preview.tsx
```diff
@@ -0,0 +1,89 @@
...
+    metadata: PostPreviewMetadata;
+};
+export default class PostMessagePreview extends React.PureComponent<Props> {
+    render() {
+        const {user, metadata, post} = this.props;
+        if (post && user) {
+            return (
+                <PostAttachmentContainer
+                    className='permalink'
+                    link={`/${metadata.team_name}/pl/${metadata.id}`}
```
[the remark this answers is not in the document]
BenCookie95: Actually I just tested this with subpaths, it will work fine because webpack is aware of the siteURL so it will treat whatever subpath you have configured as the root.

koox00 (approved):
LGTM!

packages/mattermost-redux/src/types/posts.ts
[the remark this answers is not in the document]
hmhealey: That does sound correct. I don't know why there's a Record in the current definition of PostEmbed, but I wouldn't be surprised if the current definition is incorrect since some of our Redux types are wonky.

packages/mattermost-redux/src/actions/posts.ts
hmhealey: Awesome, that sounds perfect to me.

components/post_view/post_message_preview/index.ts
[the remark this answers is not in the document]
hmhealey: I don't know if it's still the case with the changes to what's in the PostEmbed, but if it hasn't, perhaps we should use some different naming to differentiate between the post and preview post? I could see that getting confusing sometimes.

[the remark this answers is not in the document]
BenCookie95: That's a good point, I'll take a look

components/post_view/post_body_additional_content/post_body_additional_content.tsx
```diff
@@ -113,7 +113,15 @@ export default class PostBodyAdditionalContent extends React.PureComponent<Props
                     toggleEmbedVisibility={this.toggleEmbedVisibility}
                 />
             );
-
+        case 'permalink':
+            if (embed.data && 'post_id' in embed.data) {
```
mkraft, line 117: `'post_id' in embed.data` will be `true` even if the value is `null`.

components/post_view/post_message_preview/post_message_preview.tsx
```diff
@@ -0,0 +1,92 @@
...
+            className='permalink'
+            link={`/${metadata.team_name}/pl/${metadata.post_id}`}
+        >
+            <div className='post-preview'>
+                <div className='post-preview__header'>
+                    <div className='col col__name'>
+                        <div className='post__img'>
+                            <span className='profile-icon'>
+                                <Avatar
+                                    username={user?.username}
```
mkraft, line 40: `username` (and a few lines down `id`) are both required fields of `UserProfile` so the optional chaining operator will compile to unnecessary JavaScript conditions. 0/5 whether it affects performance and/or JS compilation size enough to care.

components/post_view/post_message_preview/post_message_preview.tsx
```diff
@@ -0,0 +1,92 @@
...
+                    maxHeight={100}
+                />
+
+                <div className='post__preview-footer'>
+                    <p>
+                        <FormattedMessage
+                            id='post_message_preview.channel'
+                            defaultMessage='Originally posted in ~{channel}'
+                            values={{
+                                channel: metadata?.channel_display_name,
```
mkraft, line 82: `channel_display_name` is a required field.

packages/mattermost-redux/src/actions/posts.ts
```diff
@@ -425,6 +425,29 @@ export function editPost(post: Post) {
...
+        }
+
+        dispatch({
+            type: PostTypes.RECEIVED_POSTS,
+            data: {posts},
+        });
+
+        return {data: {posts}};
+    };
+}
```
mkraft, line 450: Let me know if we're not planning to use this code and we can remove it and the server-side.

packages/mattermost-redux/src/types/posts.ts
```diff
@@ -146,3 +146,10 @@ export declare type OpenGraphMetadata = {
     url?: string;
     images: OpenGraphMetadataImage[];
 };
+
+export declare type PostPreviewMetadata = {
+    post_id: string;
+    post?: Post;
```
mkraft, line 152: Remind me, what's the reason for `post` being optional?

components/post_view/post_body_additional_content/post_body_additional_content.tsx
BenCookie95: This is to check that embed.data is actually using the type: `PostPreviewMetadata`. `data` can also be of the type, `OpenGraphMetadata`. 

I can also add a check for `null` or `undefined` though

packages/mattermost-redux/src/types/posts.ts
BenCookie95: I actually remove post from this object when storing the metadata in state because the post will already exist somewhere else in state. So that's why I marked it as optional

packages/mattermost-redux/src/actions/posts.ts
BenCookie95: You can remove that endpoint

components/post_view/post_message_preview/index.ts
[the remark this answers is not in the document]
BenCookie95: renamed `post` to `previewPost`

mkraft (approved):

hmhealey (approved):
One more non-functional nit, but otherwise looks good.

furqanmlk (approved):
Updated tests details in other PR
https://github.com/mattermost/mattermost-server/pull/17796

esethna:
@BenCookie95 please go ahead and merge if reviews are complete :)

BenCookie95:
> @BenCookie95 please go ahead and merge if reviews are complete :)

Giving it 1 final run through now and will merge after

PR #8311, opened 2021-06-25
Block only file drops, not all the kind of drops (from drag and drops)

jespino:
#### Summary
Block only file drops, not all the kind of drops (from drag and drops).

This is done because react-dnd, used in focalboard, relies on the document drop event, and this was blocking all the cases.

This should help other plugins creator that wants to use react-dnd or any other dnd that follows the same pattern.

#### Ticket Link
[MM-36964](https://mattermost.atlassian.net/browse/MM-36964)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/root/root.jsx
+++ b/components/root/root.jsx
@@ -112,8 +112,10 @@ export default class Root extends React.PureComponent {
 
         // Prevent drag and drop files from navigating away from the app
         document.addEventListener('drop', (e) => {
-            e.preventDefault();
-            e.stopPropagation();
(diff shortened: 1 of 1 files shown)


crspeller (approved):

hmhealey (approved):

jespino:
@ogi-m can you prioritize this, is important for another PR from focalboard.

ogi-m (approved):
LGTM - no change in drag and drop functionality, working as expected

Tried dropping plain text and URLs - it doesn't work but it doesn't work currently either, so the behaviour is unchanged.

PR #8322, opened 2021-06-29
[MM-36618] Save skin tone to preferences

Willyfrog:
#### Summary

User skin tone selection was persisted to localstorage and we are switching that to the user preferences in the server.

#### Ticket Link

[MM-36618](https://mattermost.atlassian.net/browse/MM-36618)

#### Release Note

```release-note
NONE
```

### DIFF
--- a/actions/emoji_actions.js
+++ b/actions/emoji_actions.js
@@ -49,13 +49,13 @@ export function setUserSkinTone(skin) {
     return async (dispatch, getState) => {
         const state = getState();
         const currentUserId = getCurrentUserId(state);
-        const teamOrderPreferences = [{
+        const skinTonePreference = [{
             user_id: currentUserId,
             name: Preferences.EMOJI_SKINTONE,
             category: Preferences.CATEGORY_EMOJI,
             value: skin,
         }];
-        dispatch(savePreferences(currentUserId, teamOrderPreferences));
+        dispatch(savePreferences(currentUserId, skinTonePreference));
     };
 }
 
--- a/actions/emoji_actions.js
+++ b/actions/emoji_actions.js
@@ -4,12 +4,14 @@
 import * as EmojiActions from 'mattermost-redux/actions/emojis';
 import {getCustomEmojisByName} from 'mattermost-redux/selectors/entities/emojis';
 import {getConfig} from 'mattermost-redux/selectors/entities/general';
+import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
 
 import {setRecentEmojis} from 'actions/local_storage';
 import {getEmojiMap, getRecentEmojis, isCustomEmojiEnabled} from 'selectors/emojis';
 import {isCustomStatusEnabled, makeGetCustomStatus} from 'selectors/views/custom_status';
+import {savePreferences} from 'mattermost-redux/actions/preferences';
 
-import {ActionTypes} from 'utils/constants';
+import {ActionTypes, Preferences} from 'utils/constants';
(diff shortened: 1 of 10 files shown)


jwilander (approved):

actions/emoji_actions.js
```diff
@@ -43,13 +45,17 @@ export function incrementEmojiPickerPage() {
...
-        dispatch({
-            type: ActionTypes.SET_RECENT_SKIN,
-            data: skin,
-        });
-        return {data: skin};
+export function setUserSkinTone(skin) {
+    return async (dispatch, getState) => {
+        const state = getState();
+        const currentUserId = getCurrentUserId(state);
+        const teamOrderPreferences = [{
```
hmhealey, line 52: Nit: Could you change the name of this?

reducers/views/emoji.js
```diff
@@ -35,7 +35,7 @@ function shortcutReactToLastPostEmittedFrom(state = '', action) {
     }
 }
 
-function recentSkin(state = '', action) {
+function userSkinTone(state = '', action) {
```
hmhealey, line 38: Did you mean to remove this?

selectors/emojis.js
```diff
@@ -39,18 +41,15 @@ export const getRecentEmojis = createSelector(
...
-
-export function getRecentSkin(state) {
-    const stateSkin = getStateRecentSkin(state);
-    if (stateSkin) {
-        return stateSkin;
-    }
-    const recentSkin = LocalStorageStore.getRecentSkin(getCurrentUserId(state));
-    return recentSkin || 'default';
-}
+// todo rename to getEmojiSkin
```
hmhealey, line 44: Did you want to address this before this PR goes in?

selectors/emojis.js
```diff
@@ -39,18 +41,15 @@ export const getRecentEmojis = createSelector(
...
+export const getUserSkinTone = createSelector(
```
hmhealey, line 45: We might not actually need to use createSelector here since there's nothing complicated going on in calculating this. We could just have
```javascript
export function getUserSkinTone(state) {
    return get(state, Preferences.CATEGORY_EMOJI, Preferences.EMOJI_SKINTONE, 'default');
};
```
0/5 on that though. I don't think this createSelector would actually cause any issues here.

hmhealey (changes requested):

actions/emoji_actions.js
Willyfrog: 🤦 definitely!

selectors/emojis.js
Willyfrog: i did rename all "recentSkin" but went with a different name at the end, I just forgot about this comment which was left in case I forgot to do it.

reducers/views/emoji.js
Willyfrog: yes, I removed the old implementation but this one slipped through, thanks for catching it

selectors/emojis.js
Willyfrog: my original thought was so it shows up in joram's selector table, but it might not be needed as it is really simple. I'll change that.

jgilliam17 (approved):
Thanks @Willyfrog 
Tested, looks good to merge.
- Verified user selected skin tone for emoji is stored and persists between sessions.

selectors/emojis.js
hmhealey: Ideally, I think we'd have less memoized selectors showing up there at all and more pure, simple selectors that we don't need to worry about as much :)

hmhealey (approved):

PR #8323, opened 2021-06-29
MM-36730: permalinks to replies sometimes throw

koox00:
#### Summary

It seemed that `componentDidUpdate` didn't have the `postRef` setted yet,
thus producing an error on scrolling to highlighted reply id.
This commit fixes it by:
  1. Changing the `isInViewport` function signature to
     always expect the element to exist.
  2. Ensure `postRef` exists before scrolling into view.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36730

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/rhs_comment/rhs_comment.jsx
+++ b/components/rhs_comment/rhs_comment.jsx
@@ -160,7 +160,7 @@ class RhsComment extends React.PureComponent {
 
     scrollIntoHighlight = () => {
         window.requestAnimationFrame(() => {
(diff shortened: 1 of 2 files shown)


jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified clicking on a link to jump to thread reply is no longer causing an error.

larkox (approved):
LGTM. It is not clear to me still why this would create a type error before, but the code seems sound.

PR #8327, opened 2021-06-29
Cropping screenshots while uploading

lapaz17:
#### Summary
This feature opens a modal when user pastes a screenshot into the text area if and only if user has 2+ monitors. Why 2+? I use 2 monitors at work and usually send whole screen to my teammates. But I can remove that condition if you want.

 In the modal, he can crop the selected area (by zooming in and out) by clicking "crop" button or select  "Upload full" button to paste whole screenshot.
I did not write the tests since UI Team might ask me to change UI a lot.
#### Ticket Link
There is no ticket for this.

#### Related Pull Requests

None

#### Screenshots
Demo
```release-note



* New features and improvements, including behavioural changes, UI changes and CLI changes

Added modal which is showed when user pastes a screenshot (it must be bigger than current screen size - particularly window.screen.w<screenshot.w || window.screen.h <screenshot.h    -> h = height , w = width ,|| = or logic), where he can zoom in and out at the modal body. 
Added two buttons on modal footer for selection.


```


lapaz17:
Idk why gif does not show. In readme files it's working :(

esethna:
Hi @lapaz17, really appreciate the PR!

This is an interesting feature. My suggestion would be to build this feature as a plugin rather than as part of the core webapp. Reason being is that the common use case from our experience is uploading a pasted image without cropping and we don't want to complicate that flow for the majority of users. 

With that said, I can imagine this kind of a tool being very useful when uploading a **profile image**. Let us know if you'd be interested in helping adapt this concept for that use case and we can provide some UI and UX guidance.

lapaz17:
@esethna could you provide a UI guidance for plugin development? You can send me mocks via community.mattermost.com, my username is @lapaz at there. If you dont want to provide, I wont change the current ui.

esethna:
@lapaz17 feel free to create the plugin any way you'd like!

If you're interested in continuing this work for the profile picture uploads then we can certainly provide some UI/UX guidance since it will be part of the core webapp interface.

lapaz17:
@esethna  I don't want to involve in the profile picture task, so I'm going to close this PR and open a new one for the plugin later. Thanks.

PR #8335, opened 2021-06-30
MM-36692: tracks CRT display setting on/off

koox00:
#### Summary

Tracks collapsed reply threads user setting change.

The category is `settings` the event is `user_settings_update` and the event props are:

```js
{
   field: 'display.collapsed_reply_threads`,
   value: 'on' || 'off'
}
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36692

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/user_settings/display/user_settings_display.tsx
+++ b/components/user_settings/display/user_settings_display.tsx
@@ -10,6 +10,8 @@ import {getTimezoneRegion} from 'mattermost-redux/utils/timezone_utils';
 import {PreferenceType} from 'mattermost-redux/types/preferences';
 import {UserProfile, UserTimezone} from 'mattermost-redux/types/users';
 
+import {trackEvent} from 'actions/telemetry_actions';
+
 import Constants from 'utils/constants';
 import * as Utils from 'utils/utils.jsx';
 import {getBrowserTimezone} from 'utils/timezone.jsx';
@@ -167,6 +169,17 @@ export default class UserSettingsDisplay extends React.PureComponent<Props, Stat
         }
     }
 
+    trackChangeIfNecessary(preference: PreferenceType, oldValue: any): void {
+        const props = {
+            field: 'display.' + preference.name,
+            value: preference.value,
+        };
+
+        if (preference.value !== oldValue) {
+            trackEvent('settings', 'user_settings_update', props);
+        }
+    }
+
     handleSubmit = async () => {
         const userId = this.props.user.id;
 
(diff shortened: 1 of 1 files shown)


hahmadia (commented):
Nice job!

Do we want to start tracking as soon as they log into Mattermost too? Because this only tracks it if they change the setting but what about those who already have toggled the setting (like myself). Don't you want to add a tracking event so as soon as a user logs in, we can know if they have it turned on or not even if they don't go into their user settings to change anything?

esethna:
@hahmadia since we haven’t released the beta yet, no one will have it on. Users will HAVE to go to account settings to turn it on for the first time.

esethna (approved):

hahmadia (approved):
LGTM

components/user_settings/display/user_settings_display.tsx
```diff
@@ -167,6 +169,17 @@ export default class UserSettingsDisplay extends React.PureComponent<Props, Stat
         }
     }
 
+    trackChangeIfNecessary(preference: PreferenceType, oldValue: any): void {
+        const props = {
+            field: 'display.' + preference.name,
```
calebroseland, line 174: `PreferenceType.category` could possibly be used here (potentially making it reusable, and could be hoisted to something like `components/user_settings/utils`?)

calebroseland (approved):
LGTM, just a thought on potential reusability. 

@esethna will this give us the ability to fully confirm active CRT:ON users? What if an admin disables the feature after a number of users turn it on?

koox00:
> LGTM, just a thought on potential reusability. 
> 
> @esethna will this give us the ability to fully confirm active CRT:ON users? What if an admin disables the feature after a number of users turn it on?

@calebroseland admin feature flag settings are already tracked so we are safe in that regard

esethna:
@enelson720 can you help @koox00 test this? Please feel free to get on a zoom call if needed. I would certainly like to get this in for the v5.37 release so needs to be merged asap

enelson720 (approved):
LGTM

components/user_settings/display/user_settings_display.tsx
koox00: @calebroseland I agree upon using constants, but the naming is particular to how we use tracking already.
So `display.collapsed_reply_threads` conforms to our tracking naming in other places as well.
Gathering all these strings into a constants file though is making a lot of sense, 
I'll add another ticket to refactor track events to use constants.

PR #8342, opened 2021-07-01
MM-35722: CRT, post click open thread

koox00:
#### Summary
Clicking, anywhere on a post in the center channel while having CRT on should open the thread in RHS,
even for single messages.

The difficulty in this is that posts may contain other elements that
have `onClick` handlers, or even just links and buttons. In those cases we
should not open the RHS.
Added a `isEligibleForClick` utility function that tries to solve the above
problem.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35722

#### Release Note

```release-note
NONE
```


matthewbirtch:
@koox00 Just played with this and noticed that if I click on a system join message, the screen goes blank. We should probably disable this feature on leave/join messages

koox00:
> @koox00 Just played with this and noticed that if I click on a system join message, the screen goes blank. We should probably disable this feature on leave/join messages

yes I just noticed too, will fix thx!

utils/utils.jsx
```diff
@@ -2250,3 +2259,42 @@ export function deleteKeysFromObject(value, keys) {
...
+
+export function isSelection() {
+    const selection = window.getSelection();
+    return selection.type === 'Range';
+}
+
+// Returns false when the element clicked or its ancestors
+// is a potential click target (link, button, image, etc..)
+// and true in any other case.
+export function isEligibleForClick(event) {
```
hmhealey, line 2271: 0/5, but perhaps there's something that could be consolidated between this PR and @BenCookie95's Permalink Previews PR (#8304)? There's some similar logic [in this function](https://github.com/mattermost/mattermost-webapp/pull/8304/files#diff-13a27994b1a523f8b6c63e5085facd209a800c35c115d56469757cab7f55edb2R15) for determining whether or not you're clicking on a clickable element or somewhere else in the PostMessageView

hmhealey (approved):

utils/utils.jsx
koox00: @hmhealey yes it's pretty similar, we could make this a function factory maybe so in addition to "clickable" elements we could pass additional class selectors depending on the case, what do you think?

*The only notable difference I see is that in this case we need to traverse parents of `event.target`s to ensure  nested elements (`span`s for example) in links or buttons are getting handled properly. If this is desired in the other PR as well we could generalize.

components/post_view/post/post.jsx
```diff
@@ -189,11 +191,24 @@ class Post extends React.PureComponent {
...
+
         if (!post) {
             return;
         }
 
+        const isSystemMessage = PostUtils.isSystemMessage(post);
+        const fromAutoResponder = PostUtils.fromAutoResponder(post);
+
+        if (
+            !e.altKey &&
```
nevyangelova, line 204: nit: do we need this negation here?

nevyangelova (approved):
LGTM

components/post_view/post/post.jsx
koox00: @nevyangelova I think we do, alt+clicking marks the post as unread, so we don't want to open the thread on that occasion.

utils/utils.jsx
BenCookie95: I like this function, it is something I can re-use in my PR. The traversing is something that I need for the next part of the preview feature. Nice work!

BenCookie95 (approved):
LGTM!

matthewbirtch (changes requested):
@koox00 after playing with this a few times I think it's working quite well. The only thing I'm wondering is:
1. Can we make sure links, link previews, permalink previews, attachments or other internal interactive elements do not open the RHS?

koox00:
> @koox00 after playing with this a few times I think it's working quite well. The only thing I'm wondering is:
> 
>     1. Can we make sure links, link previews, permalink previews, attachments or other internal interactive elements do not open the RHS? Right now it looks like clicking anywhere on the link preview container opens the RHS and clicking on attachments also opens the RHS (as well as the gallery view)

Oh, thanks @matthewbirtch! 
That was my concern as well, we cannot be 100% sure something won't "break" it in the future.

calebroseland (approved):
LGTM! Difficult problem to solve "correctly."

Besides any pending fine tuning, my only thought might be to put a maintenance PSA comment somewhere, maybe in `post.tsx`? 

~~`// here be dragons`~~ 
```tsx
/**
 * @see {isEligibleForClick} — When adding clickable targets within a root post, please add to/maintain CLICKABLE_ELEMENTS
 */
 class Post ...
 ```

File: utils/utils.jsx (hunk 1 above)
Comment (koox00) [reply utils/utils.jsx]: @BenCookie95, @hmhealey I've made some small changes, could you take a look?

I've changed `isEligibleForClick` to `makeIsEligibleForClick` which now takes a CSS selector as an argument and returns a `isEligibleForClick` function, so we are able to parametrize which other elements we don't want to handle the event for.
*CLICKABLE_ELEMENTS remain the same for all.

**Comment by matthewbirtch** [comment]
@koox00 A couple things we noticed from the team testing session:
- For the 'Update your status' tag that shows next to your username if you haven't ever used the custom status, clicking this opens up the RHS and shouldn't
- For polls, the cursor shows the whole thing as clickable, but only the poll buttons are actually clickable
- For link previews, the cursor shows the whole thing as clickable, but only the link title is actually clickable. This may be another PR, but just like the proposed permalink previews work, maybe the whole link preview container should actually be clickable

File: utils/utils.jsx [hunk 3]
```diff
@@ -2250,3 +2259,57 @@ export function deleteKeysFromObject(value, keys) {
...
+/*
+ * Returns false when the element clicked or its ancestors
+ * is a potential click target (link, button, image, etc..)
+ * but not the events currentTarget
+ * and true in any other case.
+ *
+ * @param {string} selector - CSS selector of elements not eligible for click
+ * @returns {function}
+ */
+export function makeIsEligibleForClick(selector) {
```
BenCookie95, line 2277: Can we add a default of empty for this?

esethna:
@koox00 can we try to get this merged soon? I'd like to get it in v5.38 so we can get feedback on it before the CRT GA launch

angeloskyratzakos:
@koox00 We are upgrading the test cluster at the moment, so in about 30 mins clusters will be ready. Sorry about that.

angeloskyratzakos:
just FYI @koox00 server is ready ^

koox00:
@matthewbirtch ready to be tested :)

matthewbirtch (approved):
Looks good! I think all the issues are resolved now

esethna:
cc// @jgilliam17, I'd like to try and get this in v5.38 so we can get feedback on it!

hmhealey (approved):

jgilliam17:
Thanks @koox00 
Looks and works great.
I have only two small comments:
- There is one small dead spot on the post, not clickable.
After you save a post, actual saved icon area in the post header is not opening rhs, works fine when clicking around small pin icon. It's really an edge case, and not likely to occur that often 
Screen Shot 2021-07-15 at 2 48 08 PM

- When you open user profile popover, clicking outside of the modal to dismiss it opens which ever post is below on the rhs. Workaround is to click on profile icon again to close or click on text box/header area/LHS. Personally, I'd need to re-train myself 🙂

matthewbirtch:
> * When you open user profile popover, clicking outside of the modal to dismiss it opens which ever post is below on the rhs. Workaround is to click on profile icon again to close or click on text box/header area/LHS. Personally, I'd need to re-train myself 🙂

Good find @jgilliam17

hmhealey:
>     * When you open user profile popover, clicking outside of the modal to dismiss it opens which ever post is below on the rhs. Workaround is to click on profile icon again to close or click on text box/header area/LHS. Personally, I'd need to re-train myself 🙂

That sounds like another thing that'd be fixed by making it so that, whenever a popover/menu is open, you can't interact with anything behind it. We had discussed doing that with a future menu component to stop us from having to deal with positioning menus correctly when the user scrolls behind the menu, but it would apply here since the user's first click outside the profile popover would always close it and never accidentally interact with anything else. That's obviously not something for this PR though.

esethna:
@koox00 on the above ^

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified click on a post opens RHS thread. 
Issue with dismissing profile popover has been filed separately. https://mattermost.atlassian.net/browse/MM-37291

esethna:
@jgilliam17 if we don't have one can we file a ticket for the issues we discussed in planning where this takes effect even if menus/modals etc are open

jgilliam17:
Can we repurpose the ticket I created for the profile popover? https://mattermost.atlassian.net/browse/MM-37291

jgilliam17:
@esethna see above

esethna:
@jgilliam17 sure let's just add details as needed for other cases

PR #8343, opened 2021-07-02
Eslint prettier

komik966:
#### Summary
While working with webapp code, I always spend some time on adjusting code layout (putting new  lines, spaces, etc.)
After adding prettier, this tedious work becomes just one keyboard shortcut.

I'm opening as draft, because it needs to be ensured whether this addition not breaks builds and allows for incrementally fixing style issues across codebase.
I've noticed that `make check-style` takes longer time now.


hmhealey:
Thanks for the PR, @komik966! Sorry that it seems to have slipped through the cracks back when it was submitted.

I'd definitely like to use Prettier across our code base since it'd be nice to not have to debate code style changes as often (similar to how `gofmt` enforces the recommends most of the style in our Go code), but it's something we haven't gotten around to yet since we'd like to make it something we can recommend gets used across all of our JS code, not just this repo. That unfortunately means it's a lot bigger change to make than just setting it up in this one repo and that it'll require a much bigger discussion.

I'm definitely glad to continue this discussion, but I'm going to close this for now since we're not really planning on working on this in the near term.

komik966:
ok, no problem

PR #8350, opened 2021-07-06
MM-36696: global threads list now virtualized

koox00:
#### Summary
Global threads list is virtualized to avoid rendering all thread
items at once.
This reduces the time needed to enter the threads route.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36696

#### Release Note

```release-note
NONE
```


components/threading/global_threads/thread_list/virtualized_thread_list.tsx
```diff
@@ -0,0 +1,91 @@
...
+type Props = {
+    ids: Array<$ID<UserThread>>;
+    selectedThreadId?: $ID<UserThread>;
+};
+
+const ITEMS_ON_SCREEN = 8;
+const OVERSCAN_COUNT_BACKWARD = 4;
+const OVERSCAN_COUNT_FORWARD = 4;
+
+const virtListStyles = {
```
Willyfrog, line 22: do we need to have the style inline?

components/threading/global_threads/thread_list/thread_list.tsx
```diff
@@ -97,7 +110,21 @@ const ThreadList = ({
...
             />
             <div className='threads'>
-                {children}
+                <VirtualizedThreadList
+                    key={`threads_list_${currentFilter}`}
+                    ids={currentFilter === ThreadFilter.unread ? unreadIds : ids}
+                    selectedThreadId={selectedThreadId}
+                />
+                {currentFilter === ThreadFilter.unread && !someUnread && isEmpty(unreadIds) ? (
+                    <NoResultsIndicator
```
Willyfrog, line 119: shouldn't this be either the threadlist or the noresults? why nesting the no result indicator inside a list that won't be populated?

components/threading/global_threads/thread_list/virtualized_thread_list.tsx
```diff
@@ -0,0 +1,91 @@
...
+        );
+    };
+
+    const canLoadMorePosts = useCallback(() => {}, []);
+
+    // dynamic-virtualized-list renders items in reverse order
+    // so we need to reverse the ids before feeding the list
+    // so that we get the correct order afterwards.
+    //
+    // TODO: have dynamic-virtualized-list to maybe accept a direction prop
```
Willyfrog, line 50: nitpick: this should be an issue in the [repo](https://github.com/mattermost/dynamic-virtualized-list) instead of a todo here

Willyfrog (approved):
non-blocking comments

components/threading/global_threads/thread_list/virtualized_thread_list.tsx
koox00: `DynamicSizeList` accepts a `style` prop and not a class one. 
My guess would be that since there are no stylesheets included in the module, there are only [inline styles](https://github.com/mattermost/dynamic-virtualized-list/blob/8b239f63178925163245e2629a6dea339702b4f5/src/DynamicSizeList.js#L367-L374), 
we pass inline styles so we can properly override if needed(?).

components/threading/global_threads/thread_list/thread_list.tsx
koox00: @Willyfrog it's the noresults for the unreads tab. So it's not like nesting in the list, but nesting in the left pane.
Would renaming the component from `thread_list` to `thread_list_pane` make it clearer?

PS: 
We have 3 "noresults" cases:
1. whole global threads page (when not following any thread)
2. right pane (when there is no selected thread)
3. left pane (threads list) unreads tab (when we have no unread threads) <--- the case we are discussing

matthewbirtch:
@koox00 this seems great to me, but I'm not sure the test server is the best place to test on since it doesn't have a lot of data yet.

Is there anything specific from a UX perspective that I should pay close attention to in the review? I'm not seeing anything glaring

esethna:
@matthewbirtch @koox00 do we want to add a loading indicator at the bottom of the list if you scroll down and threads are still fetching? I throttled my connection to "slow 3G" then scrolled down really fast and there was no indication that more threads were loading in until they were returned. In the center channel we have the three dots pulsing animation

esethna:
This is important for slow or unstable connections ^

koox00:
> @matthewbirtch @koox00 do we want to add a loading indicator at the bottom of the list if you scroll down and threads are still fetching? I throttled my connection to "slow 3G" then scrolled down really fast and there was no indication that more threads were loading in until they were returned. In the center channel we have the three dots pulsing animation

@esethna the infinite loading will be added later on, this PR is just to improve rendering performance of the list so that the time to enter threads view is reduced.

esethna:
Ah got it, thanks @koox00

components/threading/global_threads/thread_list/virtualized_thread_list.tsx
hmhealey: Note that the repo was created originally by Sudheer as a fork of react-window, and since he left, I don't think anyone else has been maintaining it.

components/threading/global_threads/thread_list/virtualized_thread_list.tsx
```diff
@@ -0,0 +1,91 @@
...
+        return (
+            <ThreadItem
+                key={itemId}
+                threadId={itemId}
+                isSelected={selectedThreadId === itemId}
+            />
+        );
+    };
+
+    const canLoadMorePosts = useCallback(() => {}, []);
```
hmhealey, line 44: Doesn't this need to return something? I would've expected something with this name to return true/false.

hmhealey (approved):
One non-blocking comment, but otherwise this looks good to me.

components/threading/global_threads/thread_list/virtualized_thread_list.tsx
koox00: @hmhealey well I'd expect the same, but seeing the source code for `dynamic-sized-list` it seems that it's more of a callback and not a guard on something: [L519](https://github.com/mattermost/dynamic-virtualized-list/blob/master/src/DynamicSizeList.js#L519)

But this made realize a bug... the list doesn't get updated automatically on new threads getting added, let me fix that.

calebroseland (commented):
LGTM overall. Seems some memoization (Avatars, maybe others) may be negated DynamicSizeList, when you mark as unread/read, they all seem to re-render

1st half current, 2nd half rerendering
CleanShot 2021-07-09 at 11 42 14

components/threading/global_threads/thread_list/virtualized_thread_list.tsx
koox00: > Edit: Also, that's funny that it works that way since I'm pretty sure the reason it does that is because we send the posts to the client sorted backwards (newest to oldest) which honestly has caused a lot of confusion in the post storage code in the past. I wonder if we should just store posts oldest to newest and then update the list component to never reverse the list.

@hmhealey, it seems that it doesn't work :( I cannot find a way to add items at the top... 
items added at the bottom (0 index) are getting rendered, but items added at the top (last index) don't...

I'd like to add back `react-window` for this case.

koox00:
@hmhealey, @calebroseland I've added some changes to improve performance, can you take a look?

components/threading/global_threads/thread_item/index.ts
```diff
@@ -1,4 +1,53 @@
...
+
+        const post = getPost(state, threadId);
+
+        if (getPostsForThreadProps.rootId !== post.id) {
+            getPostsForThreadProps.rootId = post.id;
+        }
+
+        if (getChannelProps.id !== post.channel_id) {
+            getChannelProps.id = post.channel_id;
+        }
```
koox00, line 37: I'll revert this change, since it doesn't affect memo.

matthewbirtch (approved):
Not seeing anything glaring from my tests. Seems to work/look great.

esethna (approved):

calebroseland (approved):
LGTM!

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified opening long global threads list, list renders as expected.

jgilliam17:
@koox00 I didn't merge this due to conflicts, can you please take a look? Thanks.

PR #6885, opened 2020-10-22
MM-29858: Migrate string refs to functional ones in autosize textarea component

sowmiyamuthuraman:
Fixes https://github.com/mattermost/mattermost-server/issues/16041

JIRA ticket: https://mattermost.atlassian.net/browse/MM-29858

PR #8351, opened 2021-07-06
WIP - Initial conversion to typescript

ewwollesen:
#### Summary

Attempting to convert this file over to Typescript, and I got most of it done I think but there is a second in the middle I'm not really sure about. Lines 124, 139, 146, and 151 are the lines in question, and VSCode is marking the `refs` as deprecated, so I'm not really sure what to do there. I saw another ticket (or maybe a communities post) about ripping out JQuery so I'm not sure if this is related that or what I need to do here, so any help would be appreciated. 

#### Ticket Link

https://github.com/mattermost/mattermost-server/issues/17281


Willyfrog:
seems you run into multiple campaigns in one file 😅  thanks for tackling them!

The issue about refs is because we moved away from string refs, so `ref='some-ref'`is not valid (you can see some linting rules being disabled at the top and the end of the file). For an example on how to fix it there are a few PRs fixing string references (like [this one](https://github.com/mattermost/mattermost-webapp/pull/6885/files)) but feel free to ask any questions regarding that.

ewwollesen:
That's perfect, thank you. I will look at that example PR and hopefully have this finished up soon.

Willyfrog:
@ewwollesen let us know if you have any questions regarding the changes or the example

ewwollesen:
> @ewwollesen let us know if you have any questions regarding the changes or the example

I think I got the refs issues resolved, and now I'm working through on one more props issue that popped up when it's trying to compile. I didn't get much time to work on it over the weekend, but hopefully I'll have that done here in the next day or two.

ewwollesen:
Just wanted to give an update that I'm still working on this and trying to get past an issue but I'm about to be out of town for a week so (and I was distracted by the docathon).

jasonblais:
Thanks @ewwollesen for continuing to work on this! :)

ewwollesen:
@Willyfrog I finally updated this PR (I think) with a new commit and the good news is that assuming I didn't make a total mess of things it actually compiles and I was able to test it and the user list seems to work as it should. I do have a couple questions and outstanding items though.

The first is line 89:
`rowComponentType?: object;`

I'm not sure what the return type for this should be.

The next one was lines 117-123:
```
        renderCount: function(
            count: number,
            total: number,
            startCount: number,
            endCount: number,
            isSearch: boolean
        ) { return null }
```

I don't know what should be the return type for this function should be. I set it to null just so it would work but that's probably not correct.

Willyfrog:
> The first is line 89:
> `rowComponentType?: object;`
> 
> I'm not sure what the return type for this should be.
> 

I'd try with `unknown`, given you don't have to check for it and only pass it along, I wouldn't worry too much unless someone defines an interface for what it is needed to pass. Maybe @devinbinnie knows better than me in this case.

> The next one was lines 117-123:
> 
> ```
>         renderCount: function(
>             count: number,
>             total: number,
>             startCount: number,
>             endCount: number,
>             isSearch: boolean
>         ) { return null }
> ```
> 
> I don't know what should be the return type for this function should be. I set it to null just so it would work but that's probably not correct.

the prop is either a function or undefined, so it should be undefined (no default) otherwise it will always return in line 215, by allowing it to be undefined it can use the code after that line.

https://github.com/mattermost/mattermost-webapp/blob/815818c59c61daeb3c7e01ae41ff485f0e80ceae/components/searchable_user_list/searchable_user_list.tsx#L215

devinbinnie:
> > The first is line 89:
> > `rowComponentType?: object;`
> > I'm not sure what the return type for this should be.
> 
> I'd try with `unknown`, given you don't have to check for it and only pass it along, I wouldn't worry too much unless someone defines an interface for what it is needed to pass. Maybe @devinbinnie knows better than me in this case.

@ewwollesen Can we try using `React.Component` as the type?

PR #8362, opened 2021-07-07
[MM-37074] Encapsulate inserted GIFs in inline image markdown

Roy-Orbison:
#### Summary
Changes Markdown inserted by GIF picker from
```markdown
https://thumbs.gfycat.com/LikableValidAlpineroadguidetigerbeetle-size_restricted.gif
```
to
```markdown
![Beverly Hills Cop - OK gesture](https://thumbs.gfycat.com/LikableValidAlpineroadguidetigerbeetle-size_restricted.gif)
```

#### Ticket Link
As per **Other notes** of [\[MM-12290\]](https://mattermost.atlassian.net/browse/MM-12290)
https://mattermost.atlassian.net/browse/MM-37074

#### Screenshots
link only  
inlined image

#### Release Note
```release-note
Improved default rendering of images inserted via the GIF picker.
```

### DIFF
--- a/components/gif_picker/gif_picker.jsx
+++ b/components/gif_picker/gif_picker.jsx
@@ -59,7 +59,7 @@ export default class GifPicker extends React.PureComponent {
     }
 
     handleItemClick = (gif) => {
-        this.props.onGifClick(gif.max5mbGif);
+        this.props.onGifClick('![' + (gif.title.replace(/]/g, '\\]') + '](' + gif.max5mbGif + ')');
     }
 
     render() {
--- a/components/gif_picker/gif_picker.jsx
+++ b/components/gif_picker/gif_picker.jsx
@@ -59,7 +59,7 @@ export default class GifPicker extends React.PureComponent {
     }
 
     handleItemClick = (gif) => {
-        this.props.onGifClick('![' + (gif.title.replace(/]/g, '\\]') + '](' + gif.max5mbGif + ')');
+        this.props.onGifClick('![' + gif.title.replace(/]|\[/g, '\\$&') + '](' + gif.max5mbGif + ')');
     }
 
     render() {


srkgupta:
Adding myself to check the CodeQL alert created on this PR

Roy-Orbison (commented):
Is this check somehow aware of Markdown syntax, thus expecting the `.replace()` to also escape possible `\` in `title` properties? Or is it balking at the escaped `[` in the regex? Or incorrectly parsing the escaped `\` in the replacement string?

srkgupta:
> Is this check somehow aware of Markdown syntax, thus expecting the `.replace()` to also escape possible `\` in `title` properties? Or is it balking at the escaped `[` in the regex? Or incorrectly parsing the escaped `\` in the replacement string?

@Roy-Orbison the alert did not consider the Markdown syntax and hence it's a false positive. The changes LGTM and I have dismissed the alert.

components/gif_picker/gif_picker.jsx
```diff
@@ -59,7 +59,7 @@ export default class GifPicker extends React.PureComponent {
     }
 
     handleItemClick = (gif) => {
-        this.props.onGifClick(gif.max5mbGif);
+        this.props.onGifClick('![' + gif.title.replace(/]|\[/g, '\\$&') + '](' + gif.max5mbGif + ')');
```
hmhealey, line 62: Also good call escaping the square brackets in the gif title

hmhealey (approved):
Thanks for the PR, @Roy-Orbison! Good catch on the followup to that ticket since it doesn't look like one was created to actually track this separately

components/gif_picker/gif_picker.jsx
Roy-Orbison: It doesn't escape `\` characters a well, but that's because the flavour of markdown in MM simply looks for a `\` preceding a square bracket. So `[\\[x\]](#y)` renders like this on GH [\\[x\]](#y), but the first `[\` of that rendering is included in the link text on MM.

Markdown is a little ill-defined.

Roy-Orbison:
I tested on that build instance, and the gif markdown worked in all three contexts that receive the updated function parameter (create post, edit post, create comment).

components/gif_picker/gif_picker.jsx
hmhealey: We do plan on switching MM's Markdown to CommonMark which should match GitHub more closely, so we'll likely have to start doing that eventually, but I don't think it's a concern for now. The only way I see this becoming a concern is if Gfycat starts sending us bad data anyway.

Roy-Orbison: All user data is bad data ;)

hahmadia (approved):

jgilliam17 (approved):
Thanks @Roy-Orbison 
Tested, looks good to merge.
- Verified GIF inline markdown

PR #8363, opened 2021-07-07
MM-35726 Dependency updates

hmhealey:
Nothing super big in terms of upgrades except I added a file that tracks upgrades using npm-upgrade. That file includes references to the ticket numbers for the bigger upgrades that we've been skipping.

Most of the individual line changes are snapshots changes caused by react-intl renaming some of their components internally.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35726

#### Release Note
```release-note
NONE
```


BenCookie95 (approved):
LGTM, ran fine locally for me. Had a look through and it's 98% snapshot changes. For the npm-upgrade file, why are the packages under an "ignore" key?

hmhealey:
They're being ignored since I wanted to skip packages I knew we weren't upgrading when I was running `npm-upgrade`. As far as I know, that's the only way to tag them with their corresponding upgrade tickets in a convenient way.

sass/layout/_markdown.scss
```diff
@@ -213,6 +213,7 @@ h6 {
         margin: 5px 0;
         border-radius: 0.25em;
         font-size: 13px;
+        overflow-x: auto;
```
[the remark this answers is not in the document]
hmhealey: Yep, that was intentional. The update to highlight.js caused code blocks to overflow horizontally

package.json
```diff
@@ -61,65 +62,65 @@
...
+    "@types/jest": "26.0.24",
     "@types/jquery": "3.5.5",
     "@types/katex": "0.11.0",
-    "@types/lodash": "4.14.168",
+    "@types/lodash": "4.14.170",
     "@types/mark.js": "8.11.5",
     "@types/marked": "0.7.4",
     "@types/moment-timezone": "0.5.30",
-    "@types/node": "14.14.37",
+    "@types/node": "15.12.5",
```
calebroseland, line 123:
```suggestion
    "@types/node": "16.3.0",
```

calebroseland (approved):
LGTM. Though `@types/node` could be bumped to 16 to keep that in sync. And should we add the `engines` field?

```json
"engines": {
    "node": ">=16.4.2",
    "npm": ">=7.18.1"
},
```

Also, these dep-updates are a great way to dethrone @crspeller. Only 158,145 to go 😛https://github.com/mattermost/mattermost-webapp/graphs/contributors?type=a
(disclaimer: joke, not a race)

calebroseland (commented):
Storybook 6.2 -> 6.3

https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#63-deprecations

I think what's breaking the build: ~~`@storybook/addon-knobs`~~ `@storybook/addon-controls`


Also: `webpack`, `@storybook/*`, and `typescript` all have recent minor bumps.

hmhealey:
I downgraded Storybook itself. I tried some other things to fix it, but those weren't working, and I don't understand enough about it too get replace addon-knobs. It sounds like it's just deprecated at this point and should continue working until Storybook 7, so I don't know what the issue is other than that.

As for those other ones, I'm not going to bother keeping the minor versions bumped after the initial one I did last week. We'll grab those next dependency update.

hmhealey:
@jgilliam17 There's nothing particular to test here outside of a regular smoke test. The only notable things to test outside of what CI covers is that the libraries for syntax highlighting and searching in the system console were updated, and I've tested both of those locally already.

.npm-upgrade.json
```diff
@@ -0,0 +1,108 @@
+{
+  "ignore": {
+    "@formatjs/intl-pluralrules": {
+      "versions": "",
+      "reason": "MM-34912"
```
crspeller, line 5: Maybe include a one line description of why here? That way you can look though this list and quickly check if the reason is something you should check up on.

hmhealey: Good call. I went through and added those.

jgilliam17:
Thanks @hmhealey 
I checked [PR E2E failures ](https://mm-cypress-report.s3.amazonaws.com/335656494-webapp-pr-8363-webapp-pr-8363-mattermost-enterprise-edition:b582216/mochawesome.html) against [master run](https://mm-cypress-report.s3.amazonaws.com/335940256-4f94e80-master-e20-prod-mattermost-enterprise-edition:master/mochawesome.html) and there is one issue that reproduces manually on the test server
- Searching for individual plugins in the system console search box, more specifically "antivirus" can't be found even though it is installed and visible on the LHS.

hmhealey:
@jgilliam17 I ended up downgrading the library we're using for searching in the System Console back to its previous version since I wasn't able to figure out how to get the new one to work. It's ready for another review now.

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
-Verified E2E results/smoke test, syntax highlighting and searching in sys. console which now works after latest revert.

PR #8385, opened 2021-07-12
Migrate Create Post and Create Comment to typescript

larkox:
#### Summary
Migrate Create Post and Create comment components to Typescript.

I left the tests out since it simplifies very much the test creation not having to conform with all the typescript needs.

Also, removing certain IDs that were only being used in the tests may have rendered some tests less useful.

This PR will be used as base of another PR for the Apps framework, so it would be great to have this merged soon, to avoid many merge conflicts.

#### Ticket Link
None

#### Related Pull Requests
None

#### Release Note
```release-note
NONE
```


components/create_comment/create_comment.tsx
```diff
@@ -301,7 +334,7 @@ class CreateComment extends React.PureComponent {
     }
 
     componentWillUnmount() {
-        this.props.resetCreatePostRequest();
+        this.props.resetCreatePostRequest?.();
```
hmhealey, line 337: When is this ever undefined?

components/create_comment/create_comment.tsx
```diff
@@ -312,8 +345,8 @@ class CreateComment extends React.PureComponent {
         }
     }
 
-    componentDidUpdate(prevProps, prevState) {
-        if (prevState.draft.uploadsInProgress.length < this.state.draft.uploadsInProgress.length && this.props.scrollToBottom) {
+    componentDidUpdate(prevProps: Props, prevState: State) {
+        if (prevState.draft!.uploadsInProgress.length < this.state.draft!.uploadsInProgress.length && this.props.scrollToBottom) {
```
hmhealey, line 349: Same with draft

actions/views/create_comment.tsx
```diff
@@ -29,43 +29,47 @@ import {getPostDraft} from 'selectors/rhs';
...
         const nextMessageInHistory = getMessageInHistory(getState());
 
         dispatch(updateCommentDraft(rootId, {...draft, message: nextMessageInHistory}));
+        return {data: true};
     };
 }
 
-export function submitPost(channelId, rootId, draft) {
-    return async (dispatch, getState) => {
+export function submitPost(channelId: string, rootId: string, draft: PostDraft & {props?: any}) {
```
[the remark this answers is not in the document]
larkox: Apparently not, since `Post` does not have the `fileInfos` attribute.

components/create_comment/create_comment.tsx
larkox: After a second look, apparently never. Probably wasn't marked as required on the PropTypes, and that is why I left it like optional.

components/create_comment/create_comment.tsx
larkox: It will be undefined the short time between the constructor is called, and the getDerivedStateFromProps is called. We cannot set it during the constructor because getDerivedStateFromProps is relying on it being undefined.

actions/views/create_comment.tsx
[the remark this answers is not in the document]
hmhealey: Ah right. Should PostDraft include the optional props field then instead of adding it just for submitPost?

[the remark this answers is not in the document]
larkox: I don't know enough of the code around it, but I am happy to make it like that.

hmhealey (approved):

larkox:
@hmhealey Re-requesting your review since along the merge I had to remove the TabIndex (which was not used) from the Textbox component.

hmhealey (approved):

hmhealey:
I'm really unclear about whether or not tabIndex is required since I remember that causing some problems with my PR recently. I've also seen warnings about it being required and not provided, so I'm not sure what exactly is up with that.

larkox:
The thing is that Textbox is asking for a tabIndex, but it is not using it. It is assigning the div it is rendering the tabindex 0 and that's it. Since Textbox is not inheriting from anything and is not passing the tabindex to anything, in theory it should be not needed.

larkox:
There has been a change on the Textbox component that I also don't completely get, which is requiring the rootID. I made the change in create post passing and empty string, but it just feel wrong. Not sure if you have any input on that.

hmhealey:
That does feel wrong, and it looks like I'm the one that caused that. If you don't mind, I can go submit a quick PR to add it to the defaultProps instead.

hmhealey:
PR here: https://github.com/mattermost/mattermost-webapp/pull/8510

I ended up just making rootId optional again. I think what confused me was that the component had it as optional previously, but the mapStateToProps required it, so I assumed we could make it required everywhere.

mickmister (approved):
LGTM!

larkox:
@jgilliam17 Notes for testing: Most of the changes are on the Create Post and Create Comment components (which visually are more or less the textbox on the center channel, the RHS, and the thread view on CRT). It is just a refactoring, so no changes in functionality are expected.

jgilliam17 (approved):
Thanks @larkox 
Tested, looks good to merge.
- Verified textbox on the main channel view, RHS and Threads view with CRT on - no change, as before.

PR #8391, opened 2021-07-13
Fix dynamic and static select not being readonly

larkox:
#### Summary
Static and Dynamic selects were not respecting the readonly parameter. Now they are.

#### Ticket Link
None

#### Release Note
```release-note
None
```


hanzei (approved):
Good catch

mickmister (approved):

PR #8394, opened 2021-07-13
MM36866 - mid month proration UX changes to purchase modal

pablovelezvidal:
#### Summary
This PR adds the changes to show the proration information to customers that want to upgrade from Cloud Starter (non-free-trial) to any of the other product offerings.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36866

#### Release Note

```release-note
NONE
```


components/purchase_modal/purchase_modal.tsx
```diff
@@ -281,6 +287,28 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
+        return (
+            <a
+                className='footer-text'
+                onClick={() => {
+                    trackEvent(
+                        TELEMETRY_CATEGORIES.CLOUD_PURCHASING,
+                        'learn_more_prorated_payment',
+                    );
+                }}
+                href={CloudLinks.PRORATED_PAYMENT}
```
AGMETEOR, line 300: Looks like this leads to a dead link

components/purchase_modal/purchase_modal.tsx
```diff
@@ -290,6 +318,72 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
+                >
+                    <div className='tooltipTitle'>
+                        <FormattedMessage
+                            defaultMessage={'Prorated Payments'}
+                            id={'admin.billing.subscription.proratedPayment.tooltipTitle'}
+                        />
+                    </div>
+                    <div className='tooltipText'>
+                        <FormattedMessage
+                            defaultMessage={'If you upgrade to {selectedProductName} from Cloud Starter mid-month, you will be charged a prorated amount for both plans.'}
```
AGMETEOR, line 348: Can't a an upgrade happen during anytime within the month and not necessarily mid-month? I think @annelieseklein could help clarify on this one.

components/purchase_modal/purchase_modal.tsx
```diff
@@ -290,6 +318,72 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
+                values={{
+                    beginDate: getNextBillingDate(),
+                }}
+            />
+        );
+
+        let payment = normalPaymentText;
+        if ((!this.props.isFreeTrial && this.state.currentProduct?.name.toLowerCase().includes('starter')) &&
+                !this.state.selectedProduct?.name.toLowerCase().includes('starter')) {
+            const announcementTooltip = (
```
AGMETEOR, line 335: Just a few UI inconsistencies with the designs in the ticket
Screenshot 2021-07-14 at 11 57 36

- The color of the text after "Prorated Payments" looks different
- The date i.e. "Aug 1, 2021" should begin on a new line

AGMETEOR (changes requested):

components/purchase_modal/purchase_modal.tsx
pablovelezvidal: it is still dead, but it was the link provided, eventually it will work.

components/purchase_modal/purchase_modal.tsx
annelieseklein: yup! anytime within the month @AGMETEOR @pablovelezvidal

pablovelezvidal: I think mid-month is the expression to refer `in the middle of the month` which not exactly refers to the literal middle/half. Is my assumption correct @annelieseklein ?

components/purchase_modal/process_payment_setup.tsx
```diff
@@ -139,6 +143,45 @@ export default class ProcessPaymentSetup extends React.PureComponent<Props, Stat
...
 
+    private sucessPage = () => {
+        const {error} = this.state;
+        if (this.props.isProratedPayment) {
+            const formattedButonText = (
+                <FormattedMessage
+                    defaultMessage={'Return to {team}'}
+                    id={'admin.billing.subscription.proratedPayment.buttonText'}
+                    values={{team: this.props.teamName}}
+                />
```
marianunez, line 154: @annelieseklein given that the payment modal can (and mostly is accessed) from the System Console, the button that says "Return to {team}" seems misleading because it implies that it would take you to back to the team outside of the System Console. This will actually just close the modal and leave you where you were before. Should we reconsider changing this to something else? Before it was just `Let's go`

https://www.figma.com/file/B3YyrLCLAjNfdPUNq0shQq/Cloud-Professional-Trial-Flow?node-id=1753%3A1569

components/purchase_modal/purchase_modal.tsx
```diff
@@ -200,28 +206,36 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
         // if not on trial, only show current plan and those higher than it in terms of price
-        if (!this.props.isFreeTrial && currentPrice) {
-            options = options.filter((option: Option) => {
-                return option.price >= currentPrice;
+        if (!this.props.isFreeTrial && currentProduct.price_per_seat) {
+            userBasedProducts = userBasedProducts.filter((option: RadioGroupOption) => {
+                return option.price >= currentProduct.price_per_seat;
+            });
+            flatFeeProducts = flatFeeProducts.filter((option: RadioGroupOption) => {
+                return option.price >= currentProduct.price_per_seat;
```
marianunez, line 228: Is this suppose to be the `else` of the above `if (!this.props.isFreeTrial && currentProduct.price_per_seat)` ? What happens if it's flat_fee price?

components/purchase_modal/purchase_modal.tsx
```diff
@@ -290,6 +326,72 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
+                id={'admin.billing.subscription.paymentBegins'}
+                values={{
+                    beginDate: getNextBillingDate(),
+                }}
+            />
+        );
+
+        let payment = normalPaymentText;
+        if ((!this.props.isFreeTrial && this.state.currentProduct?.name.toLowerCase().includes('starter')) &&
+                !this.state.selectedProduct?.name.toLowerCase().includes('starter')) {
```
marianunez, line 342: We can generalize this condition to be: `if currentProduct.billing_scheme == flat_rate && selectedProduct,billing_scheme== per_user`

So in general, any time we are going from a flat-fee to user-based there will be prorated charges.

components/purchase_modal/purchase_modal.tsx
```diff
@@ -573,6 +669,9 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
                                         }}
                                         contactSupportLink={this.props.contactSalesLink}
                                         selectedProduct={this.state.selectedProduct}
+                                        teamName={this.props.team?.name}
+                                        isProratedPayment={(!this.props.isFreeTrial && this.state.currentProduct?.name.toLowerCase().includes('starter')) &&
+                                        !this.state.selectedProduct?.name.toLowerCase().includes('starter')}
```
marianunez, line 674: Same as above here, the condition to be prorated is going from a flat-fee to a user-based product. We can generalize this so we don't hardcode any product name

i18n/en.json
```diff
@@ -351,6 +352,12 @@
   "admin.billing.subscription.privateCloudCard.freeTrial.description": "We love to work with our customers and their needs. Contact sales for subscription, billing or trial-specific questions.",
   "admin.billing.subscription.privateCloudCard.freeTrial.title": "Questions about your trial?",
   "admin.billing.subscription.privateCloudCard.upgradeNow": "Upgrade Now",
+  "admin.billing.subscription.proratedPayment.buttonText": "Return to {team}",
+  "admin.billing.subscription.proratedPayment.substitle": "Thank you for upgrading to Cloud Professional. You will be charged a prorated amount  for your Cloud Starter plan and Cloud Professional plan based on the number of days and number of users.",
+  "admin.billing.subscription.proratedPayment.title": "You are now subscribed to Cloud Professional",
+  "admin.billing.subscription.proratedPayment.tooltipText": "If you upgrade to {selectedProductName} from Cloud Starter mid-month, you will be charged a prorated amount for both plans.",
```
marianunez, line 358: Let's remove the hardcoded product names here in all these texts. Let's have them be open based on what they are upgrading. ie.:
```suggestion
  "admin.billing.subscription.proratedPayment.substitle": "Thank you for upgrading to {selectedProductName}. You will be charged a prorated amount  for your {currentProductName} plan and {selectedProductName} plan based on the number of days and number of users.",
```

components/purchase_modal/process_payment_setup.tsx
```diff
@@ -139,6 +143,45 @@ export default class ProcessPaymentSetup extends React.PureComponent<Props, Stat
...
+                        title={t('admin.billing.subscription.proratedPayment.title')}
+                        subtitle={t('admin.billing.subscription.proratedPayment.substitle')}
+                        date={getNextBillingDate()}
+                        error={error}
+                        icon={successSvg}
+                        formattedButonText={formattedButonText}
+                        buttonHandler={this.props.onClose}
+                        className={'success'}
+                    />
+                </>
```
marianunez, line 168: In this success page we are still missing the table of the example of prorated?

Screen Shot 2021-07-15 at 10 25 13 PM

i18n/en.json
marianunez, line 358: We need to remove all the hardcoded name of products in this messages

marianunez (changes requested):
Great work @pablovelezvidal! We may still be missing the prorating table example in the success page

components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -117,6 +95,30 @@ const BillingSubscriptions: React.FC = () => {
...
+        let product: Product;
+        if (products && subscription) {
+            product = products[subscription?.product_id];
+            if (!product) {
+                keys.forEach((key) => {
+                    if (products[key].name.toLowerCase().includes('professional')) {
+                        product = products[key];
+                    }
+                });
+            }
```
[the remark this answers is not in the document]
pablovelezvidal: this was part of the original validations we made for when loading the purchase modal product, so, initially we decided that the fallback product was always going to be professional. This condition is only accomplish when no subscription product_id is associated.

components/purchase_modal/process_payment_setup.tsx
pablovelezvidal: This will be done as part of the ticket https://mattermost.atlassian.net/browse/MM-37164

components/purchase_modal/purchase_modal.tsx
pablovelezvidal: ok!

components/purchase_modal/purchase_modal.tsx
pablovelezvidal: ok

components/purchase_modal/purchase_modal.tsx
pablovelezvidal: ok

i18n/en.json
pablovelezvidal: ok

pablovelezvidal: ok

pablovelezvidal:
Captura de pantalla 2021-07-16 a las 19 55 17

components/admin_console/billing/billing_subscriptions/index.tsx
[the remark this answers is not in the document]
marianunez: Not sure at the time why we were setting this way, but do you see any issues to only falling back to the first product in the list if none is found (as you do below)?

components/purchase_modal/purchase_modal.tsx
```diff
@@ -200,28 +206,39 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
-        const currentPrice = this.state.currentProduct?.price_per_seat;
+        const currentProduct = this.state.currentProduct!;
 
         // if not on trial, only show current plan and those higher than it in terms of price
-        if (!this.props.isFreeTrial && currentPrice) {
-            options = options.filter((option: Option) => {
-                return option.price >= currentPrice;
-            });
+        if (!this.props.isFreeTrial) {
+            if (currentProduct.price_per_seat) {
```
marianunez, line 224: Should we check here on `billing_scheme` ?

components/purchase_modal/purchase_modal.tsx
```diff
@@ -200,28 +206,39 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
+                flatFeeProducts = flatFeeProducts.filter((option: RadioGroupOption) => {
+                    return option.price >= currentProduct.price_per_seat;
+                });
+            }
         }
 
+        // temp commented if current product is user based, do not allow downgrade to flat_fees
+        // if (!this.props.isFreeTrial && currentProduct.billing_scheme === BillingSchemes.PER_SEAT) {
+        //     flatFeeProducts = [];
+        // }
```
marianunez, line 238: Let's integrate this with the conditions above

components/purchase_modal/purchase_modal.tsx
```diff
@@ -290,6 +329,73 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
+                id={'admin.billing.subscription.paymentBegins'}
+                values={{
+                    beginDate: getNextBillingDate(),
+                }}
+            />
+        );
+
+        let payment = normalPaymentText;
+        if ((!this.props.isFreeTrial && this.state.currentProduct?.billing_scheme === BillingSchemes.FLAT_FEE) &&
+                this.state.selectedProduct?.billing_scheme === BillingSchemes.PER_SEAT) {
```
marianunez, line 345: nit: no need for extra parenthesis here
```suggestion
        if (!this.props.isFreeTrial && this.state.currentProduct?.billing_scheme === BillingSchemes.FLAT_FEE &&
                this.state.selectedProduct?.billing_scheme === BillingSchemes.PER_SEAT) {
```

components/purchase_modal/process_payment_setup.tsx
marianunez: Sounds good!

components/purchase_modal/process_payment_setup.tsx
annelieseklein: @marianunez I changed it to "Return to {team}" because I thought that the user might want to return to working within the upgraded product and go to the team instead of the system console. If we want to just have it close (same as the close icon in the corner) then "Let's go" is fine.

marianunez: Thanks Anneliese! @pablovelezvidal see above regarding this button wording. We should be able to remove any team information we are getting that is no longer needed here for the name.

pablovelezvidal:
this PR also includes the fix for a defect found by @marianunez in the initial review of this PR. This is the Defect https://mattermost.atlassian.net/browse/MM-37033 . @AGMETEOR can you please re-review this PR since it has got a lot of changes after Maria's feedback? Another thing to notice is that spinwick doesn't work longer on this PR because database inconsistencies caused during the testing, so the plan is to finish the code static review and then we will open a new PR so @stevemudie and @annelieseklein can take a final look at the working changes with a new working spinwick server.

marianunez (approved):
Thanks for working through all these details for upgrade and product sorting 👍 (And patience with the spinwick too 😄 ) 

Non-blocking comments below

components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -118,6 +95,19 @@ const BillingSubscriptions: React.FC = () => {
     }
 
     useEffect(() => {
+        if (product !== null || isEmpty(products) || !subscription?.product_id) {
+            return;
+        }
+        const keys = Object.keys(products!);
```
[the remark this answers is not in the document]
pablovelezvidal: not completely necessary, but typescript doesn't read the above condition using isEmpty, so it keeps complaining that products could be null.

components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -118,6 +95,19 @@ const BillingSubscriptions: React.FC = () => {
...
+        let tempProduct: Product;
+        tempProduct = products![subscription?.product_id];
+        if (isEmpty(tempProduct)) {
+            tempProduct = products![keys[0]];
+        }
```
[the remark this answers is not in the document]
pablovelezvidal: same here

components/purchase_modal/purchase_modal.tsx
```diff
@@ -194,34 +198,50 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
...
-        if (!this.props.isFreeTrial && currentPrice) {
-            options = options.filter((option: Option) => {
-                return option.price >= currentPrice;
-            });
+        if (!this.props.isFreeTrial) {
+            if (currentProduct.billing_scheme === BillingSchemes.PER_SEAT) {
+                userBasedProducts = userBasedProducts.filter((option: RadioGroupOption) => {
+                    flatFeeProducts = [];
+                    return option.price >= currentProduct.price_per_seat;
+                });
```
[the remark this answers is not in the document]
pablovelezvidal: totally! good catch!

components/admin_console/billing/billing_subscriptions/index.tsx
```diff
@@ -118,6 +95,19 @@ const BillingSubscriptions: React.FC = () => {
...
+        setProduct(tempProduct);
+    }, [products, subscription]);
+
+    useLayoutEffect(() => {
```
AGMETEOR, line 110: Why `useLayoutEffect` and not good ol' `useEffect`?

AGMETEOR: Asking based on what I read here https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

pablovelezvidal: actually that was a leftover after fighting with the subscriptions information that ended up seeming as a chrome debugger problem, but yes, basically there is no difference between both hooks, however useLayoutEffects executes itself before useEffect, so I was using that situation but the end result ended up being the same. Let me set it back to useEffect since it makes no difference. Good catch @AGMETEOR !

pablovelezvidal:
Closing this PR, the new one that contains this same changes is https://github.com/mattermost/mattermost-webapp/pull/8485

pablovelezvidal:
QA Validation approved in this PR https://github.com/mattermost/mattermost-webapp/pull/8485

PR #8396, opened 2021-07-13
Do not save drafts until typing has stopped for X milliseconds

jwilander:
#### Summary
Currently, we were saving drafts for messages being typed into the post and comment textboxes on animation frames which can occur quite frequently. Saving drafts also seems to be less performant on some browser/OS/device combinations than on others. They also fire off a number of dispatches and change redux state which can negatively impact performance too. During some local testing, where I added a console log for each draft save, I was seeing it saved on every single key press which is 1) saving significantly more often than necessary and 2) could be negatively impacting performance, particularly in certain environments.

This is after typing about 100 characters into the post textbox:
Screen Shot 2021-07-13 at 8 28 02 AM

In addition some very useful profiles from the community on Firefox are showing `saveDraftFrame` being called about every 100ms during typing and was taking up the majority of the JS processing time in the profile:
Screen Shot 2021-07-12 at 1 43 29 PM

This PR moves from using animation frames to save the draft to a more controlled `setTimeout` which will only save the draft after 500ms of no typing has occurred (or the component unmounts).

#### Release Note

```release-note
Improve typing performance in affected environments by reducing the frequency at which drafts are saved
```

### DIFF
--- a/components/create_post/create_post.jsx
+++ b/components/create_post/create_post.jsx
@@ -1,6 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+/* eslint-disable max-lines */
+
 import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
@@ -45,6 +47,8 @@ import FormattedMarkdownMessage from 'components/formatted_markdown_message.jsx'
 import MessageSubmitError from 'components/message_submit_error';
 
 const KeyCodes = Constants.KeyCodes;
+
+const CreatePostDraftTimeoutMilliseconds = 500;
 
 // Temporary fix for IE-11, see MM-13423
 function trimRight(str) {
@@ -375,7 +379,7 @@ class CreatePost extends React.PureComponent {
         if (this.saveDraftFrame) {
             const channelId = this.props.currentChannel.id;
             this.props.actions.setDraft(StoragePrefixes.DRAFT + channelId, this.draftsForChannel[channelId]);
-            cancelAnimationFrame(this.saveDraftFrame);
+            clearTimeout(this.saveDraftFrame);
         }
     }
 
@@ -532,7 +536,7 @@ class CreatePost extends React.PureComponent {
             postError: null,
         });
 
-        cancelAnimationFrame(this.saveDraftFrame);
+        clearTimeout(this.saveDraftFrame);
         this.props.actions.setDraft(StoragePrefixes.DRAFT + channelId, null);
         this.draftsForChannel[channelId] = null;
     }
@@ -802,10 +806,10 @@ class CreatePost extends React.PureComponent {
             ...this.props.draft,
             message,
         };
-        cancelAnimationFrame(this.saveDraftFrame);
-        this.saveDraftFrame = requestAnimationFrame(() => {
+        clearTimeout(this.saveDraftFrame);
+        this.saveDraftFrame = setTimeout(() => {
             this.props.actions.setDraft(StoragePrefixes.DRAFT + channelId, draft);
-        });
+        }, CreatePostDraftTimeoutMilliseconds);
         this.draftsForChannel[channelId] = draft;
     }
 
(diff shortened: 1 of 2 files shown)


components/create_post/create_post.jsx
```diff
@@ -46,6 +48,8 @@ import MessageSubmitError from 'components/message_submit_error';
 
 const KeyCodes = Constants.KeyCodes;
 
+const CreatePostDraftTimeoutMilliseconds = 500;
```
jwilander, line 51: I initially started lower at 250ms but during my testing it would still fire when I was typing at my normal speed. At 500ms it would consistently only fire after I had a significant pause in my typing.

@hmhealey I know you wanted to lean on the lower end but for this to have the performance impact I'm hoping it will I think it's important that saving doesn't occur during typing.

jwilander:
@jgilliam17 can you test to make sure that post and comment drafts are still saved correctly?

jwilander:
Found a minor issue with the comment button staying disabled too long that might be related to this change. Going to investigate that

jwilander:
Fixed the issues. @jgilliam17 can you smoke test commenting on posts/threads with this change too?

jwilander:
Also @jgilliam17 (and anyone else) please leave the test server up since I'm going to ask some people to create accounts and see if improves performance for them

jgilliam17:
@jwilander Do we have a Jira ticket for this?

jwilander:
No, I didn't create one since it was a bit reactionary. I can create one if you want, though.

jgilliam17:
I'll add one as a reminded to test again after merge. 🙂

components/create_post/create_post.jsx
devinbinnie: @jwilander Non-blocking, but is there any benefit to having this constant declared here and the same one declared in `create_comment`. Is there any point where we'd want those intervals to be different? If not it might be worth be putting the constant in a constants file.

components/threading/thread_viewer/thread_viewer.tsx
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+/* eslint-disable max-lines */
```
devinbinnie, line 4: Unrelated, but we should really get rid of this...it's more annoying than it helps.

devinbinnie (approved):

components/create_post/create_post.jsx
hmhealey: I was originally thinking we could lean on the side of having it fire periodically as the user was typing, but I guess if saving the draft is taking as long as we've seen, you're probably right that it's better to be on the safe side with this. If we hear people are losing drafts, we can always back off on the interval.

As for the duplicated constants, I don't think it's important enough to move to the constants file. That thing is bloated with a lot of things that are only used in one or two places if they're even still used at all. I think having two values is fine since I can't see us using this anywhere else, and ideally, both the CreatePost and CreateComment would be rewritten to share this logic entirely anyway.

components/threading/thread_viewer/thread_viewer.tsx
hmhealey: Maybe we could increase the limit or disable it for older files, but I'd like to keep it for newer files since it's nice to discourage people from adding new files with too much stuff going on in them. Only 1/5 on that though.

hmhealey (approved):
This looks good to me. I was thinking we might use lodash's debounce utility since that avoids having to manage the timeout manually, but the more I think about it, doing that manually has me less concerned about potentially making mistakes like saving a draft to the wrong channel.

components/create_post/create_post.jsx
jwilander: >As for the duplicated constants, I don't think it's important enough to move to the constants file. That thing is bloated with a lot of things that are only used in one or two places if they're even still used at all. I think having two values is fine since I can't see us using this anywhere else, and ideally, both the CreatePost and CreateComment would be rewritten to share this logic entirely anyway.

This was pretty much my exact reasoning.

jgilliam17 (approved):
Thanks @jwilander 
Tested, looks good to merge.
- Verified post and comment drafts are saved correctly, no issues commenting on threads and single posts.

PR #8398, opened 2021-07-13
Add balancing call to TestHelper.tearDown() in mattermost-redux test

jayaddison-collabora:
#### Summary
This change doesn't fix any particular issue, nor does it add functionality, and I don't think it is required or important.  All it does is make sure that calls to `TestHelper.initBasic` and `TestHelper.tearDown` are balanced in the codebase; one teardown call was missing.

#### Ticket Link
N/A

#### Related Pull Requests
N/A

#### Release Note
```release-note
NONE
```


hmhealey (approved):
Thanks for the PR. I'm not sure if it's strictly necessary for these calls to be matching everywhere since I know nock clears some information between test cases and suites, but it definitely doesn't hurt to include it.

hmhealey:
Skipping QA since the change only affects the unit tests.

nevyangelova (approved):

PR #8399, opened 2021-07-13
ApplyTheme Updates – convert older changeCSS JS to CSS variables part 1

deanwhillier:
#### Summary
This PR consolidates 10 related branches that each deal with converting older `changeCSS` theming applied using JS to newer `CSS Properties` (variables).

This PR also changes from using opacity baked into CSS Variables to using RGBA. For example:

``` css
background: var(--center-channel-color-08);
```

changes to

```css
background: rgba(var(--center-channel-color-rgb), 0.08);
```

It is challenging to deal with each branch separately as there is crossover, and as such, each successive PR is based on the previous so I've elected to consolidate work to date into a single PR.

To help with reviewing/testing, I'm going to list each commit grouped by theme variable name so related changes can be more easily seen in context given this consolidated PR shows all compounded changes together. Additionally, the git description for each commit indicates what is affected within the webapp.

Commits:

- Sidebar BG: https://github.com/mattermost/mattermost-webapp/commit/eb797a6107eb39a36df468deb09ee37efcdcf2b2
- Sidebar Text: https://github.com/mattermost/mattermost-webapp/commit/826c5954c3703e3ef994c7a428337b2924630920
- Sidebar Text Hover GB: https://github.com/mattermost/mattermost-webapp/commit/a7ab7ddbf869ec92b7d9e9339cd7191139a75015
- Sidebar Text Active Color: https://github.com/mattermost/mattermost-webapp/commit/a4b61cdf94326ae5d5055d52225ff3821c5dc919
- Remove Change Opacity:
   - https://github.com/mattermost/mattermost-webapp/commit/589c10618eda90bada63ca45a34ab2aeff5c32a9
   - https://github.com/mattermost/mattermost-webapp/commit/2801e3f3728218a2cc8609de31bd8fdcaff7b6ea
   - https://github.com/mattermost/mattermost-webapp/commit/50ba7dfa6c209d9d2341a0a679a9e1dda7262f24
- Sidebar Header BG: https://github.com/mattermost/mattermost-webapp/commit/0e512fc3784606aae63f842da6e49ef7a2d65138
- Sidebar Header Text Color: https://github.com/mattermost/mattermost-webapp/commit/8011f940edb7985745c048b3d56698a55cfc6412
- DND Indicator: https://github.com/mattermost/mattermost-webapp/commit/f8406061084e27f2c4f5eb1de745b8f28b14a753
- Mention BG: https://github.com/mattermost/mattermost-webapp/commit/f30e1cb830b982dbe6450db2e2c89315b5425701
- Mention Color: https://github.com/mattermost/mattermost-webapp/commit/1bd3463b66a74626174e2178a85cbf4c80a3f8d5

Please reach out to me if further help is needed to review/test as I know the old CSS is not the easiest to navigate!

Edits:
- added baked-in opacity variables back to ensure backwards compatibility for plugins etc.
- removed what looked to be unused legacy compatibility code (thanks @hmhealey!)


sass/components/_badge.scss
```diff
@@ -13,15 +13,23 @@
         line-height: 18px;
     }
 
+    .team-sidebar,
+    .sidebar--left,
+    #SidebarContainer {
+        .active .badge,
+        .badge {
+            background: var(--mention-bg);
+            color: var(--mention-color);
```
nevyangelova, line 22: Is there an rgb variable we can use here?

sass/components/_status-icon.scss
```diff
@@ -47,6 +47,10 @@
                 width: 13px;
                 height: 13px;
                 max-height: initial;
+
+                &.offline--icon {
+                    fill: var(--sidebar-text);
```
nevyangelova, line 52: Here as well?

sass/layout/_headers.scss
```diff
@@ -737,7 +738,7 @@
             position: relative;
             width: 100%;
             padding: 2px 16px 0 0;
-            color: $white;
+            color: var(--sidebar-header-text-color);
```
nevyangelova, line 741: And this

nevyangelova (commented):
LGTM but I think in some places we can replace using the rgba variables

sass/components/_badge.scss
deanwhillier: A little while back, we decided to have two color variables, hex and rgb. We use rgba() function with `*-rbg` variables when a color needs transparency, otherwise we use the direct hex colour.

I'm 0/5 on whether moving forward we exclusively use rgb() and rbga() function and forgo hex colours, but we should discuss.

@hmhealey, thoughts?

hmhealey: I'm 0/5 on one way or another, but we'd have to make sure to provide plugins a timeline to stop using either CSS variable if we were to stop using one ourselves. We'll have to keep supporting hex values in the theme variables on the JS-side since that's what custom themes store in the DB and what any component using the theme utilities will expect, but we don't necessarily need to expose that in the CSS.

The only advantage I see to using two different variables is that it's slightly shorter when we're using full opacity colours, but I'm starting to realize we seem to really like using different opacities a lot 😛

sass/components/_modal.scss
```diff
@@ -250,7 +250,7 @@
         .modal-title {
             overflow: hidden;
             max-width: calc(100% - 80px);
-            color: $bg--gray;
+            color: var(--sidebar-header-text-color);
```
hmhealey, line 253: Does this also need to be applied to `.name` below as per https://github.com/mattermost/mattermost-webapp/pull/8399/files#diff-efa0d8b3bc74afa1fa4a36331dfa53b1549a04392979e2a2d83fdd914ffc2f4aL583?

utils/utils.jsx
```diff
@@ -872,86 +820,27 @@ export function applyTheme(theme) {
             'sidebar-unread-text-rgb': toRgbValues(theme.sidebarUnreadText),
 
             // Hex CSS variables
-            // TODO: phase out changeOpacity() here
             'sidebar-bg': theme.sidebarBg,
             'sidebar-text': theme.sidebarText,
-            'sidebar-text-08': changeOpacity(theme.sidebarText, 0.08),
```
hmhealey, line 878: Have you checked with the teams developing plugins to see if they're using any of these? If compatibility is a problem, perhaps we can leave these variables around until 6.0, include them in the deprecation list, and loudly yell at anyone who tries to use them in the web app in the mean time?

hmhealey (changes requested):

utils/utils.jsx
deanwhillier: Excellent point, let's put them back! 🙂

sass/components/_badge.scss
deanwhillier: @nevyangelova, I'm going to do a combined pass to convert all hex variables uses to rgb in a separate PR, so will leave these as is for now.

sass/components/_modal.scss
deanwhillier: I couldn't find `.name` used anywhere anymore, added a note to that affect to the description for that commit.

nevyangelova (approved):
thanks I think I got your comments(as much as I could from the phone). LGTM!

hmhealey (approved):

deanwhillier:
@michaelgamble, @jgilliam17, test servers are back up and running! 🎉

michaelgamble (changes requested):
After reviewing carefully, the only item that is missing from my perspective is:
- Sidebar Header > Hamburger Menu Icon (or sidebar-header-dropdown__icon) is black when it should be white

deanwhillier:
Thanks @michaelgamble, good catch! Looks like a change from another PR that was merged to master recently messed up the css for that one. Should be fixed now.

michaelgamble (changes requested):
So one last change.. it seems in the original state for mattermost, the hamburger menu's default color has an opacity of 0.8 and then looks like on hover becomes opacity of 1.0

I think once we add that, it should be good to go.

deanwhillier:
Try now? Same reason, different location ... 🤷‍♂️

FWIW, I'm not updating that to use the newer rgba(var(...), ...) approach, just leaving it as is as that is beyond the scope of this PR.

jgilliam17:
Thanks @deanwhillier 
Overall looks good, seeing few smaller issues
- Legacy sidebar dm icons fill color in light themes is off

Screen Shot 2021-07-30 at 12 10 24 AM

- Dark team system console hamburger menu dropdown incorrectly themed
Screen Shot 2021-07-30 at 12 27 37 AM

- DND background color, shows black or white on the main channel
Screen Shot 2021-07-30 at 12 40 20 AM
Screen Shot 2021-07-30 at 12 42 44 AM

- Theme on Add member modal dropdown (this was recently fixed on master)
Screen Shot 2021-07-30 at 12 53 31 AM

- Note sure if this is related to your changes, but About modal header border is missing when accessing system console from dark team and opening about modal there 
Screen Shot 2021-07-30 at 1 15 51 AM

deanwhillier:
Thanks @jgilliam17!

- status icon colours should be fixed now
- I merged master to inherit the other fix
- system console bugs are likely in master (I verified that menu one is at least), can you please double check on community and file separate tickets for the messaging platform if so?

jgilliam17 (approved):
Thanks @deanwhillier 
Tested, looks good to merge.
- Verified changes as listed in description. System console issues - filed separately MM-37669, MM-37670

michaelgamble (approved):
Looks good to me :)

PR #8405, opened 2021-07-15
[MM-37064] - Onboarding flow is showing for existing users

AGMETEOR:
#### Summary
This PR addresses the following issues with the onboarding process:
- Because onboarding was recently opened for self managed accounts, existing users were seeing the onboarding flow as well. In this PR we check if a user account has existed for some time (5 days) and we don't show the onboarding flow. They can instead access it from the main menu
- If onboarding is not showing, switching between threads view and channel view will not show it
- Switching between threads and channel view hides the onboarding flow. Also switching between channels hides the onboarding flow. The user can always access it from the main menu

#### Ticket Link
[MM-37064](https://mattermost.atlassian.net/browse/MM-37064?atlOrigin=eyJpIjoiMGUyYmI5NzNlN2E5NDM0YzgxNWFhNjQyYTQ3MWJjMjIiLCJwIjoiaiJ9)

[MM-37106](https://mattermost.atlassian.net/browse/MM-37106?atlOrigin=eyJpIjoiNTBlMDg0NDBjNDAzNGZmMGEzMTQxNWQyOTk4ZDIwMTYiLCJwIjoiaiJ9)

[MM-37105](https://mattermost.atlassian.net/browse/MM-37105?atlOrigin=eyJpIjoiM2EwZDI4MDZhMWY3NDNkYTk4ZDIwNWY4NjQzMTQ1MzgiLCJwIjoiaiJ9)

## Test 1
- Create an instance and add some users. We need the user that we're going to test to stay on this server for 6 days (we temporarily alter code to make it shorter)
- Add several teams and enable threads from the account settings > display
- For a user that has stayed more than 6 days on the server, the onboarding should not show up

## Test 2 (can be done immediately)
- Now create a new user. and log in with that user. Add that user to all the teams on the server
- When logged in, the onboarding flow should show for this user.
- Test these scenarios. Switching between teams should hide the onboarding and it can then be got again from main menu > Getting started
- Switch to thread view and refresh. Onboarding should show there instead of threads. Switch between thread and channel for the current team. The onboarding should disappear.
- Now switch to channel and refresh. Once again the onboarding shows up. Switch between channels and the onboarding should disappear.

#### Release Note
```release-note
None

```


components/next_steps_view/steps.ts
```diff
@@ -194,3 +196,15 @@ export const nextStepsNotFinished = createSelector(
...
+    'userExistedForSomeTime',
+    (state: GlobalState) => getCurrentUser(state),
+    (currentUser): boolean => {
+        const today = moment(Date.now());
+        const created = moment(new Date(currentUser.create_at));
+        const d = today.diff(created, 'days');
+        const existedForSomeTime = d > NUMBER_OF_DAYS_AFTER_WHICH_TO_CONSIDER_AS_NON_NEW_USER;
+        return existedForSomeTime;
+    },
+);
```
marianunez, line 210: Should we move this to the users selectors here: https://github.com/mattermost/mattermost-webapp/blob/master/packages/mattermost-redux/src/selectors/entities/users.ts#L1

components/next_steps_view/steps.ts
```diff
@@ -194,3 +196,15 @@ export const nextStepsNotFinished = createSelector(
         return !Steps.every(checkPref);
     },
 );
+
+export const userExistedForSomeTime = createSelector(
+    'userExistedForSomeTime',
```
marianunez, line 201: Naming here is a big vague, could we maybe switch it up to check if current user is a new account so it's more clear the intention of this - maybe something like: `isCurrentUserNewAccount` or something similar.

utils/constants.jsx
```diff
@@ -42,6 +42,8 @@ import solarizedDarkCSS from '!!file-loader?name=files/code_themes/[hash].[ext]!
 // eslint-disable-line import/order
 import solarizedLightCSS from '!!file-loader?name=files/code_themes/[hash].[ext]!highlight.js/styles/solarized-light.css'; // eslint-disable-line import/order
 
+export const NUMBER_OF_DAYS_AFTER_WHICH_TO_CONSIDER_AS_NON_NEW_USER = 5;
```
marianunez, line 45:
```suggestion
export const NUMBER_OF_DAYS_CONSIDERED_NEW_USER = 5;
```

components/channel_layout/center_channel/center_channel.tsx
```diff
@@ -58,6 +70,27 @@ export default class CenterChannel extends React.PureComponent<Props, State> {
...
 
+    async componentDidMount() {
+        const {actions, showNextSteps, showNextStepsTips, isOnboardingHidden, userExistedForSomeTime, currentUserId} = this.props;
+        await actions.getProfiles();
+
+        // We call getUser want to make sure we have the current user's profile into the profiles state
+        await actions.getUser(currentUserId);
+        if ((showNextSteps || showNextStepsTips) && !isOnboardingHidden && !userExistedForSomeTime) {
+            actions.setShowNextStepsView(true);
+        }
```
marianunez, line 81: It seems that if they change teams, this will make it show again even though they navigated away from it and it should only be showing in the sticky menu in the LHS

marianunez (changes requested):

components/channel_layout/center_channel/center_channel.tsx
AGMETEOR: I have tested and switching teams doesn't make it show again

AGMETEOR: This makes sure it doesn't show 
```js
    componentDidUpdate(prevProps: Props) {
        const {location, showNextStepsEphemeral, actions, currentUserId} = this.props;
        if (location.pathname !== prevProps.location.pathname && showNextStepsEphemeral) {
            actions.setShowNextStepsView(false);
        }

        // We call getUser want to make sure we have the current user's profile into the profiles state
        actions.getUser(currentUserId);
    }

```

devinbinnie (approved):
Code looks good, but I did have a question about the approach. Is there a way instead that we could check if the user has ever logged in as well? I'm just wondering if there's ever a case where a user could be created a week in advance but the user can't login until later. For example, hiring a new employee that doesn't start for at least a week.

AGMETEOR:
> Code looks good, but I did have a question about the approach. Is there a way instead that we could check if the user has ever logged in as well? I'm just wondering if there's ever a case where a user could be created a week in advance but the user can't login until later. For example, hiring a new employee that doesn't start for at least a week.

@devinbinnie Yes, I considered that scenario, but I looked at the user model and I couldn't see any field or information that points to that direction. I'll try asking the in the server channel.

AGMETEOR:
@stevemudie since this ticket addresses 3 tickets at once. I have added QA steps to PR description above

packages/mattermost-redux/src/selectors/entities/users.ts
```diff
@@ -681,3 +684,15 @@ export function searchProfilesInGroup(state: GlobalState, groupId: $ID<Group>, t
...
 }
+
+export const isCurrentUserNewAccount = createSelector(
+    'isCurrentUserNewAccount',
+    (state: GlobalState) => getCurrentUser(state),
+    (currentUser): boolean => {
+        const today = moment(Date.now());
+        const created = moment(new Date(currentUser.create_at));
+        const d = today.diff(created, 'days');
+        const existedForSomeTime = d > NUMBER_OF_DAYS_CONSIDERED_NEW_USER;
```
AGMETEOR, line 695: Try and use audit tables to see last login. See here https://community.mattermost.com/core/pl/yoad19cwqj89in5rs5h3iswf3a

marianunez:
> Code looks good, but I did have a question about the approach. Is there a way instead that we could check if the user has ever logged in as well? I'm just wondering if there's ever a case where a user could be created a week in advance but the user can't login until later. For example, hiring a new employee that doesn't start for at least a week.

@devinbinnie @AGMETEOR 

Agreed that it can get tricky just checking the last 5 days. The most correct approach here would be to update all existing user preferences and toggle the hide onboarding for them. However, wondering if we can just to a more simple approach and check if the user was created before the release of the feature we just don't show it and any new users after the release will see the on-boarding properly (regardless of when they created the account and if they log in 5 or more days later)

devinbinnie:
> > Code looks good, but I did have a question about the approach. Is there a way instead that we could check if the user has ever logged in as well? I'm just wondering if there's ever a case where a user could be created a week in advance but the user can't login until later. For example, hiring a new employee that doesn't start for at least a week.
> 
> @devinbinnie @AGMETEOR
> 
> Agreed that it can get tricky just checking the last 5 days. The most correct approach here would be to update all existing user preferences and toggle the hide onboarding for them. However, wondering if we can just to a more simple approach and check if the user was created before the release of the feature (5.38) we just don't show it and any new users after the release will see the on-boarding properly (regardless of when they created the account and if they log in 5 or more days later)

I would personally be much more in favour of the migration script over anything. I'm not the biggest fan of looking through audit tables as I assume those queries aren't well optimized. I'm also not a fan of writing temporary code checking for a specific version, I feel like that would be prone to issues as well.

stevemudie:
@AGMETEOR On the Spinwick I’ve logged in using my steve.mudie@mattermost.com email address and have followed the steps you laid out above but wasn’t able to enable threads as it’s not an option in the account settings.

Is that element important for the 6 day test?

AGMETEOR:
> @AGMETEOR On the Spinwick I’ve logged in using my [steve.mudie@mattermost.com](mailto:steve.mudie@mattermost.com) email address and have followed the steps you laid out above but wasn’t able to enable threads as it’s not an option in the account settings.
> 
> Is that element important for the 6 day test?

@stevemudie go the admin console and search threads, you’ll have to enable it there and then that option to enable it as well in the account settings will appear

stevemudie:
> go the admin console and search threads, you’ll have to enable it there and then that option to enable it as well in the account settings will appear

@AGMETEOR
Thanks! Sorry to bother!

I'd searched for it but didn't use "threads" as the search term. 🤦‍♂️

AGMETEOR:
> > go the admin console and search threads, you’ll have to enable it there and then that option to enable it as well in the account settings will appear
> 
> @AGMETEOR
> Thanks! Sorry to bother!
> 
> I'd searched for it but didn't use "threads" as the search term. 🤦‍♂️

No worries

stylianosrigas:
Cleaning the test server as it is failing with the following error:
```
Failed to set plugin key value with options","error":"failed to delete PluginKeyValue: context deadline exceeded","errorVerbose":"context deadline exceeded\nfailed to delete PluginKeyValue\ngithub.com/mattermost/mattermost-server/v5/store/sqlstore.SqlPluginStore.CompareAndSet\n\tgithub.com/mattermost/mattermost-server/v5/store/sqlstore/plugin_store.go:104
```

AGMETEOR:
> > > Code looks good, but I did have a question about the approach. Is there a way instead that we could check if the user has ever logged in as well? I'm just wondering if there's ever a case where a user could be created a week in advance but the user can't login until later. For example, hiring a new employee that doesn't start for at least a week.
> > 
> > 
> > @devinbinnie @AGMETEOR
> > Agreed that it can get tricky just checking the last 5 days. The most correct approach here would be to update all existing user preferences and toggle the hide onboarding for them. However, wondering if we can just to a more simple approach and check if the user was created before the release of the feature (5.38) we just don't show it and any new users after the release will see the on-boarding properly (regardless of when they created the account and if they log in 5 or more days later)
> 
> I would personally be much more in favour of the migration script over anything. I'm not the biggest fan of looking through audit tables as I assume those queries aren't well optimized. I'm also not a fan of writing temporary code checking for a specific version, I feel like that would be prone to issues as well.

@devinbinnie @marianunez 

I have been thinking we can do the opposite of what the migration script will do. I noticed that currently the onboarding will show up for existing users (on prem/ cloud) because in the db, `recommended_next_steps` preferences with names `skip` and `hide` are not explicitly specified. These only get added to the db if a user explicitly clicks `skip getting started` or `hides` when the onboarding shows up.

For example, one of the selectors that determines if we show onboarding or not is this one
```js
export const isOnboardingHidden = createSelector(
    'isOnboardingHidden',
    (state: GlobalState) => getCategory(state, Preferences.RECOMMENDED_NEXT_STEPS),
    (stepPreferences) => {
        return stepPreferences.some((pref) => (pref.name === RecommendedNextSteps.HIDE && pref.value === 'true'));
    },
);
```
From the code above, because `stepPreferences` is an empty, `[].some` evaluates to false and we show the onboarding flow. This is the same for the other selectors like `showNextSteps` and `showNextStepsTips`.

We have 2 issues to solve:
- Not showing onboarding to existing users (> 5 days from time of creation)
- Be able to show the onboarding flow for new users

For the first issue, we know that if the onboarding is showing up for existing users, it's because they have not explicitly stated they want to hide or skip. We can therefore make a check that if they're older than 5 days (can change) and they haven't hidden or skipped. So older than 5 days and stepPreferences are empty array, we hide onboarding.

For the second issue, we can make it a point that, for any new users we create. We also save their next step preferences with skip and hidden false. This would be a change on the server.

Thoughts?

marianunez:
> For the second issue, we can make it a point that, for any new users we create. We also save their next step preferences with skip and hidden false. This would be a change on the server.

@AGMETEOR I think this approach would work best so that we don't have to do a migration of existing users that involved doing an insert for each in the preference table - which could've been expensive for servers with a big user base. 

However, if we are now reverting to show if they have the preference enabled in the db, do we still need to check the 5 day old condition? I think the preference check inverted would be enough

devinbinnie:
> > For the second issue, we can make it a point that, for any new users we create. We also save their next step preferences with skip and hidden false. This would be a change on the server.
> 
> @AGMETEOR I think this approach would work best so that we don't have to do a migration of existing users that involved doing an insert for each in the preference table - which could've been expensive for servers with a big user base.
> 
> However, if we are now reverting to show if they have the preference enabled in the db, do we still need to check the 5 day old condition? I think the preference check inverted would be enough

Agree with @marianunez on dropping the 5 day rule. 

I think one issue I had run across with checking for the inverse was that the preferences may not have loaded completely when the selector calculates, so the onboarding flow would flash in and out upon initial load. Though if that were the case, it should definitely happen here too since if there's no preferences it should show it. 

But I think checking for the absence of the `RecommendedNextSteps.HIDE` pref makes more sense than what we currently have for sure.

marianunez:
> For the second issue, we can make it a point that, for any new users we create. We also save their next step preferences with skip and hidden false. This would be a change on the server.

@AGMETEOR let's go ahead with this proposal so it's more robust for existing and new users. Do you want to do this change on another PR and maybe we can decouple from the issues of not showing when in thread view [MM-37106](https://mattermost.atlassian.net/browse/MM-37106) and coming back after team switch [MM-37105](https://mattermost.atlassian.net/browse/MM-37105)?

AGMETEOR:
> > For the second issue, we can make it a point that, for any new users we create. We also save their next step preferences with skip and hidden false. This would be a change on the server.
> 
> @AGMETEOR let's go ahead with this proposal so it's more robust for existing and new users. Do you want to do this change on another PR and maybe we can decouple from the issues of not showing when in thread view [MM-37106](https://mattermost.atlassian.net/browse/MM-37106) and coming back after team switch [MM-37105](https://mattermost.atlassian.net/browse/MM-37105) already fixed here?

I am going to break them up into smaller PRs tomorrow

AGMETEOR:
Closing in favor of https://github.com/mattermost/mattermost-webapp/pull/8442

marianunez:
@AGMETEOR is there another PR that takes care of https://mattermost.atlassian.net/browse/MM-37106 and https://mattermost.atlassian.net/browse/MM-37105 ?

AGMETEOR:
> @AGMETEOR is there another PR that takes care of https://mattermost.atlassian.net/browse/MM-37106 and https://mattermost.atlassian.net/browse/MM-37105 ?

@marianunez yes https://github.com/mattermost/mattermost-webapp/pull/8437 and https://github.com/mattermost/mattermost-webapp/pull/8438

PR #8406, opened 2021-07-15
[MM-37108] Only make request if server is enterprise ready

streamer45:
#### Summary

Making the `getPrevTrialLicense()` request was causing a panic server side when in team edition. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-37108

#### Release Note

```release-note
NONE
```

#### Related PRs

https://github.com/mattermost/mattermost-server/pull/17933

### DIFF
--- a/components/admin_console/license_settings/license_settings.jsx
+++ b/components/admin_console/license_settings/license_settings.jsx
@@ -67,13 +67,12 @@ export default class LicenseSettings extends React.PureComponent {
     }
 
     componentDidMount() {
-        if (!this.props.enterpriseReady) {
+        if (this.props.enterpriseReady) {
+            this.props.actions.getPrevTrialLicense();
+        } else {
             this.reloadPercentage();
         }
         this.props.actions.getLicenseConfig();
-        if (this.props.enterpriseReady) {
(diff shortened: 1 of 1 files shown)


harshilsharma63 (approved):

components/admin_console/license_settings/license_settings.jsx
```diff
@@ -71,7 +71,9 @@ export default class LicenseSettings extends React.PureComponent {
             this.reloadPercentage();
         }
         this.props.actions.getLicenseConfig();
-        this.props.actions.getPrevTrialLicense();
+        if (this.props.enterpriseReady) {
+            this.props.actions.getPrevTrialLicense();
+        }
```
streamer45, line 76: If the order of these calls is not important I can just add the call to an else block. Let me know.

BenCookie95: I don't think the order of these calls matter, you could move it into the else if you want

BenCookie95 (approved):

furqanmlk (approved):
Working as expected
Details are updated in Server PR

PR #8408, opened 2021-07-16
MM-36756 Add new component for positioning SuggestionList in RHS

hmhealey:
This code is for deciding whether or not the SuggestionList in the RHS should open upwards or downwards. The previous version did this by accessing the DOM in the render method which is against good Redux practices, but it also had performance impacts since accessing the DOM is slow. The new RhsSuggestionList component only does this whenever the list is opened.

This also required moving some of the logic that positions the SuggestionList relative to the keyboard caret when the user is typing since that relied on knowing which direction the SuggestionList would open. It now provides some more information to the SuggestionList that it can use for doing the calculations itself.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36756

#### Release Note
```release-note
NONE
```


utils/utils.jsx
```diff
@@ -1155,16 +1155,17 @@ export function offsetTopLeft(el) {
     return {top: rect.top + scrollTop, left: rect.left + scrollLeft};
 }
 
-export function getSuggestionBoxAlgn(textArea, pxToSubstract = 0, boxLocation = 'top') {
+export function getSuggestionBoxAlgn(textArea, pxToSubstract = 0/*, boxLocation = 'top'*/) {
```
larkox, line 1158: Any reason to keep the commented code?

components/suggestion/suggestion_list.jsx
```diff
@@ -11,11 +11,13 @@ import {Constants} from 'utils/constants.jsx';
 import FormattedMarkdownMessage from 'components/formatted_markdown_message';
 import LoadingSpinner from 'components/widgets/loading/loading_spinner';
 
+// When this file is migrated to TypeScript, type definitions for its props already exist in ./suggestion_list.d.ts.
```
larkox, line 14: Why not migrate this along this change?

larkox (approved):
LGTM. Just two small considerations.

utils/utils.jsx
hmhealey: Good catch. That should've been removed.

components/suggestion/suggestion_list.jsx
hmhealey: I forget exactly why I didn't migrate this when I originally did this a few weeks back, but I think there was just too much going on between this and the other related components to migrate it over.

components/suggestion/suggestion_list.d.ts
```diff
@@ -0,0 +1,43 @@
...
+    // suggestionBoxAlgn is an optional object that can be passed to align the SuggestionList with the keyboard caret
+    // as the user is typing.
+    suggestionBoxAlgn?: {
+        lineHeight: number;
+        pixelsToMoveX: number;
+        pixelsToMoveY: number;
+    };
+}
+
+declare module 'components/post_view/reaction' {
```
calebroseland, line 39:
```suggestion
declare module 'components/suggestion/suggestion_list' {
```

components/textbox/textbox.tsx
```diff
@@ -39,6 +38,7 @@ type Props = {
     onBlur?: (e: FocusEvent) => void;
     supportsCommands: boolean;
     handlePostError?: (message: JSX.Element | null) => void;
+    suggestionList?: React.Component<React.ComponentProps<SuggestionList>>;
```
calebroseland, line 41:
```suggestion
    suggestionList?: React.Component<React.ComponentProps<typeof SuggestionList>>;
```

components/suggestion/rhs_suggestion_list.tsx
```diff
@@ -0,0 +1,43 @@
...
+import AutosizeTextarea from 'components/autosize_textarea';
+import QuickInput from 'components/quick_input';
+
+import Constants from 'utils/constants';
+
+import SuggestionList from './suggestion_list';
+
+type Props = React.ComponentProps<SuggestionList> & {
+    target: React.Ref<QuickInput>;
+}
```
calebroseland, line 15:
```suggestion
type Props = React.ComponentProps<typeof SuggestionList> & {
    inputRef: React.RefObject<QuickInput>;
}
```

components/suggestion/rhs_suggestion_list.tsx
```diff
@@ -0,0 +1,43 @@
...
+            const inputTop = (input?.getInput() as AutosizeTextarea).getDOMNode()?.getBoundingClientRect().top || 0;
+            const newPosition = (inputTop < Constants.SUGGESTION_LIST_SPACE_RHS) ? 'bottom' : 'top';
+
+            if (newPosition !== position) {
+                // This potentially causes a second render when the list position changes, but that's better
+                // than checking the bounding rectangle while rendering.
+                setPosition(newPosition);
+            }
+        }
+    }, [target, otherProps.open]);
```
calebroseland, line 35:
```suggestion
    const [position, setPosition] = useState<Props['location']>('top');

    const {inputRef, ...otherProps} = props;

    useEffect(() => {
        const input = inputRef?.current;

        if (otherProps.open) {
            const inputTop = (input?.getInput() as AutosizeTextarea).getDOMNode()?.getBoundingClientRect().top || 0;
            const newPosition = (inputTop < Constants.SUGGESTION_LIST_SPACE_RHS) ? 'bottom' : 'top';

            if (newPosition !== position) {
                // This potentially causes a second render when the list position changes, but that's better
                // than checking the bounding rectangle while rendering.
                setPosition(newPosition);
            }
        }
    }, [inputRef, otherProps.open]);
```

calebroseland (changes requested):
The declaration path seemed off which caused props to be `any` downstream, and I think it's `inputRef` instead of `target`—I was getting hard-crashes on each keystroke in create_comment. Otherwise seems to be working well 👍.

components/suggestion/suggestion_list.d.ts
hmhealey: How the heck did I manage that...

hmhealey:
I've fixed all that. I'm kind of concerned with how it ended up not working when I definitely tested this. Hopefully I just broke something when splitting it off from the followup PR.

components/textbox/textbox.tsx
```diff
@@ -39,6 +38,7 @@ type Props = {
     onBlur?: (e: FocusEvent) => void;
     supportsCommands: boolean;
     handlePostError?: (message: JSX.Element | null) => void;
+    suggestionList?: React.Component<React.ComponentProps<typeof SuggestionList>>;
```
calebroseland, line 41: Didn't realize this could be simplified last round 😛 (also fixes a type)
```suggestion
    suggestionList?: typeof SuggestionList;
```

calebroseland (commented):
Didn't notice the errors last review, but..

The other type error could be fixed in `apps_form_field.tsx` with `any` in prop `listComponent?: React.ComponentClass<any>;`

There might be a better way, but seems ok and underlying is just propTypes anyway.

components/textbox/textbox.tsx
hmhealey: Oh that's much easier. I know I've used typeof here a few times, but I didn't actually understand how it worked in a typing context.

hmhealey: Okay, I wasn't able to get this to work. I started getting some bizarre errors from AppsFormField when I tried to set this type, so I gave up and essentially set them back to any. Assuming this passes the CI checks finally, I think we might just need to live with this for now.

I don't know if `typeof SuggestionList` works when you're wanting any component with the same props as `SuggestionList`, but perhaps that's because there's so much plain JS code in between that definition and the code using it. We might need to wait until all of this is migrated to be able to resolve those issues.

calebroseland: Big +1 on letting it be with a now-passing CI.

calebroseland (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified autocomplete in the RHS is positioned correctly relative to the position of the textbox (above or below) and caret.

PR #8409, opened 2021-07-16
Standardize name of suggestionListPosition

hmhealey:
I was working on this code last night, and I kept getting confused by the names of these fields since it changes between "style" and "location" for the prop, and I originally thought "location" was whether or not the autocomplete appeared in the RHS or centre channel. Long story short, I renamed a bunch of things.

The notable part of this PR is that plugins like Incident Collaboration use the Textbox, so it still expects the old prop to exist. I worked around that by creating a wrapper component that translates the name of the prop so we can use the new name in the web app without breaking plugins. While this is kind of a silly use case to start with, I imagine this pattern will become a lot more useful once we start making larger changes to components like the Textbox to improve performance in the app.

Since this is based off the actual PR I was supposed to be working on at the time, the changes specific to this one are here: https://github.com/mattermost/mattermost-webapp/compare/MM-36756_suggestion-box-position...hh_suggestion-box-position-rename

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8408

#### Release Note
```release-note
NONE
```


crspeller (approved):
Tested with playbooks and appears to work!

cpoile (approved):
Nice, thanks @hmhealey!

hmhealey:
@jgilliam17 For this, could you double check that the autocompletes for various textboxes look like they're going in the right direction? I did a pass of that myself including the command autocomplete in the Playbooks plugin which Christopher has checked as well.

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified autocompletes for various textboxes, all opening as expected.

PR #8410, opened 2021-07-16
[MM-34895] send timestamp after last_reply_at when viewing thread

ashishbhate:
#### Summary

When the user posts a message in a thread, the server responds with a THREAD_UPDATED event. In the websocket handler for this event, the thread is immediately marked as read if that thread is open. 

Looking at the community server responses on my local machine - the `create_at` of user's post  was `1626446482669`. The webapp then sends a request to mark the thread as read with the timestamp `1626446482566` (`Date.Now()`)

The mark as read timestamp is before the `create_at` of the post created. **Could be this because of some clock skew?** I have updated the code to set the read time as just after the `create_at `of the post, if `Date.Now` is before the `create_at` time of the post.

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-34895

#### Related Pull Requests

- N/A

#### Release Note

```release-note
NONE
```

### DIFF
--- a/actions/websocket_actions.jsx
+++ b/actions/websocket_actions.jsx
@@ -1470,6 +1470,11 @@ function handleThreadUpdated(msg) {
         let lastViewedAt;
         if (isThreadOpen(state, threadData.id) && !isThreadManuallyUnread(state, threadData.id)) {
             lastViewedAt = Date.now();
+
+            // Sometimes `Date.now()` was generating a timestamp before the
+            // last_reply_at of the thread, thus marking the thread as unread
+            // instead of read. Here we set the timestamp to after the
+            // last_reply_at if this happens.
             if (lastViewedAt < threadData.last_reply_at) {
                 lastViewedAt = threadData.last_reply_at + 1;
             }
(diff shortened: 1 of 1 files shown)


actions/websocket_actions.jsx
```diff
@@ -1470,6 +1470,9 @@ function handleThreadUpdated(msg) {
         let lastViewedAt;
         if (isThreadOpen(state, threadData.id) && !isThreadManuallyUnread(state, threadData.id)) {
             lastViewedAt = Date.now();
+            if (lastViewedAt < threadData.last_reply_at) {
+                lastViewedAt = threadData.last_reply_at + 1;
```
jwilander, line 1474: Maybe add a quick comment explaining why this is done?

jwilander (approved):
Good with me - just one comment

koox00 (approved):

actions/websocket_actions.jsx
ashishbhate: Done

ashishbhate:
@jgilliam17 I can't reproduce the issue on my local machine or a server I set up on AWS. I only see this issue on the community server. We should test after merge and deployment to daily.

jgilliam17 (approved):
Thanks @ashishbhate 
Tested, looks good to merge.
- Verified self replies are not marking threads as unread.

PR #8430, opened 2021-07-20
MM-36934 Disable unreads filter when unreads are grouped separately

hmhealey:
Users could previously get into a weird state when they enable the unreads filter in the channel sidebar and then turned on Group Unreads Separately in Account Settings > Display since that hides the option for the user to turn the unreads filter back off.

This PR changes it so the unreads filter is never allowed to be enabled when Group Unreads Separately is also enabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36934

#### Release Note
```release-note
Prevented users from having the unreads filter enabled when the button to toggle it is not shown
```


AshishDhama (approved):

devinbinnie (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified unreads filter is disabled when unreads are grouped separately.

PR #8437, opened 2021-07-21
[MM-37105] - Onboarding comes back after team or thread view switch

AGMETEOR:
#### Summary
This PR ensures that
- Switching switching teams will hide the onboarding flow
- Switching from thread view to channel view will not show the onboarding flow

#### Ticket Link
[MM-37105](https://mattermost.atlassian.net/browse/MM-37105?atlOrigin=eyJpIjoiM2Q3MDJjMDk3ZGRiNGZlOWEwNDQzODg1NGNiNjY2M2IiLCJwIjoiaiJ9)

#### Release Note
```release-note
NONE

```


components/channel_layout/center_channel/center_channel.tsx
```diff
@@ -58,6 +69,27 @@ export default class CenterChannel extends React.PureComponent<Props, State> {
         return {lastReturnTo: nextProps.location.pathname};
     }
 
+    async componentDidMount() {
+        const {actions, showNextSteps, showNextStepsTips, isOnboardingHidden, currentUserId} = this.props;
+        await actions.getProfiles();
+
+        // We call getUser to make sure we have the current user's profile into the profiles state
+        await actions.getUser(currentUserId);
```
pablovelezvidal, line 77: hey @AGMETEOR just curiosity, why exactly do we need to make this call? like, what can potentially happen if we don't do it?

pablovelezvidal (approved):
LGTM! Great work @AGMETEOR ! :shipit:

components/channel_layout/center_channel/center_channel.tsx
marianunez: Possibly left over from the previous 5 day creation date check?

AGMETEOR: When we call `getProfiles()` it fetches a paginated list of user profiles. Because the profiles are paginated, it's possible that the page of profiles returned may not have the current user profile. To ensure it's added to the profiles returned we need to make that call `getUser`. It looks redundant but that's how the redux was set up. I'll ask Harisson if they can add to the redux state an entry for the `currentUserProfile`. Atm, there's only `currentUserId`

pablovelezvidal: ok, that kind of makes sense. Maybe we can left the note that we should create a getProfileById and just use that one with the currentUserId, internally will be exactly the same, but will make the logic and the code simpler to read.

devinbinnie: Huh, I'm surprised we don't make a specific call for this at any point during regular operation. At least not that I could find.

components/channel_layout/center_channel/center_channel.tsx
```diff
@@ -58,6 +69,27 @@ export default class CenterChannel extends React.PureComponent<Props, State> {
...
+    }
+
+    componentDidUpdate(prevProps: Props) {
+        const {location, showNextStepsEphemeral, actions, currentUserId} = this.props;
+        if (location.pathname !== prevProps.location.pathname && showNextStepsEphemeral) {
+            actions.setShowNextStepsView(false);
+        }
+
+        // We call getUser to make sure we have the current user's profile into the profiles state
+        actions.getUser(currentUserId);
```
devinbinnie, line 90: This will be called a LOT of times if we do it everytime the component refreshes. Are we just doing it in case it changes? If so, it might make more sense to have it pulled down by the action that changes it instead of having it update everytime the component does.

components/channel_layout/center_channel/index.ts
```diff
@@ -31,13 +38,36 @@ const mapStateToProps = (state: GlobalState, ownProps: Props) => {
...
     }
     const lastChannelPath = `${ownProps.match.url}/channels/${channelName}`;
+    const currentUserId = getCurrentUserId(state);
     return {
         lastChannelPath,
         lhsOpen: getIsLhsOpen(state),
         rhsOpen: getIsRhsOpen(state),
         rhsMenuOpen: getIsRhsMenuOpen(state),
         isCollapsedThreadsEnabled: isCollapsedThreadsEnabled(state),
+        currentUserId,
```
devinbinnie, line 48: Nit: since we aren't we re-using this, can we just have it in-line?
```suggestion
        currentUserId: getCurrentUserId(state),
```

devinbinnie (changes requested):

stevemudie (approved):
Works great!
Onboarding didn't pop back up regardless of what I did even though I hadn't completed it.

components/channel_layout/center_channel/center_channel.tsx
AGMETEOR: Yes. It's possible that `getProfiles` gets called else where. In fact, if it it'll remove the current user from the profiles. See here https://github.com/mattermost/mattermost-webapp/blob/42ac676894524deb94743e3243653f1c9035b201/packages/mattermost-redux/src/actions/users.ts#L304

components/channel_layout/center_channel/center_channel.tsx
AGMETEOR: `getProfiles` will remove the current user from the profiles. See here https://github.com/mattermost/mattermost-webapp/blob/42ac676894524deb94743e3243653f1c9035b201/packages/mattermost-redux/src/actions/users.ts#L304

components/channel_layout/center_channel/center_channel.tsx
marianunez: Why do we need to get the user for every update here?

components/channel_layout/center_channel/center_channel.tsx
marianunez: Why do we need to `getProfiles` here in the mount of center channel and how does it relate to the onboarding flow?

components/channel_layout/center_channel/center_channel.tsx
AGMETEOR: It’s the only way to ensure the current user profile is part of the profiles fetched by ‘getProfiles’

marianunez: Sure, taking a step back here - why do we need to fetch profiles on the center channel ? Apologies for insisting here, but just want to make the correlation with onboarding and why is something we need now

AGMETEOR: We probably don’t need to fetch profiles. It’s something I took from the previous implementation. But 4/5 we need to getUser

AGMETEOR: Actually, I have remembered, the reason I had put it there was to check for the user creation date from the profile. We're not doing that anymore. I'll take it out :)

AGMETEOR: With the `getProfiles`, since it was used in the previous implementation, I am not sure why they call it, so it's safe to leave it there.

marianunez: Makes sense! Thanks for double checking 👍

AGMETEOR: @devinbinnie sorry for the confusion with the call to `getUser`. I have taken it out.

marianunez (approved):

marianunez:
@devinbinnie gentle ping on this - want to make sure you are good with the fixes before we merge

marianunez:
@AGMETEOR some conflicts to resolve given the related PRs are merged

devinbinnie (approved):

PR #8441, opened 2021-07-21
MM-36765: fixes threads crashing when LEAVE_CHANNEL

koox00:
#### Summary

Leaving a channel removes post items that belong to that channel from
posts, but it doesn't remove threads that belong to that channel, so the
global threads view crashes.

This commit fixes that by removing threads as well upon a LEAVE_CHANNEL
action being dispatched.

The way entities.threads reducer is structured it wasn't possible to
update counts or threads in team upon LEAVE_CHANNEL action. 
So this commit refactors combineReducers to a custom function to enable updating those as well.

PS: ATM when visiting global threads you would see those removed threads
getting added again upon fetching threads. This is going to be fixed at [MM-36769](https://mattermost.atlassian.net/browse/MM-36769)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36765

#### Release Note

```release-note
NONE
```


jwilander (approved):

packages/mattermost-redux/src/reducers/entities/threads.ts
```diff
@@ -243,17 +295,84 @@ export const countsReducer = (state: ThreadsState['counts'] = {}, action: Generi
...
+// enables passing data between reducers
+function reducer(state: ThreadsState = {threads: {}, threadsInTeam: {}, counts: {}}, action: GenericAction): ThreadsState {
+    const extra: ExtraData = {};
+
+    // acting as a 'middleware'
+    if (action.type === ChannelTypes.LEAVE_CHANNEL) {
+        if (!action.data.viewArchivedChannels) {
+            extra.threadsToDelete = getThreadsOfChannel(state.threads, action.data.id);
+        }
+    }
```
jwilander, line 352: Not sure if we've done this type of thing before. Looks good to me but interested to see what @hmhealey thinks

koox00: @jwilander we've done it in the posts reducer, well not quite the same but same pattern.
my opinion is that It's not something we should do often but I couldn't figure how to update counts and threads in team without having that extra data.

hmhealey: I wish we had a better pattern for doing this, but yeah, it's acceptable to do this. There's part of the redux docs that talks about how combineReducer is convenient for some things but limited sometimes.

hmhealey (approved):

hmhealey:
@koox00 Could this use QA testing?

packages/mattermost-redux/src/reducers/entities/threads.ts
koox00: Another way I can think of achieving just the same, is a redux middleware that adds extra data just on this action, which is pretty close to this, but feels like an overkill

jgilliam17:
Thanks @koox00 
There is an error now a bit earlier, when leaving a channel, before accessing threads view.
Screen Shot 2021-07-23 at 4 43 44 PM

koox00:
@jgilliam17 good catch, FYI this error exists on master as well, it's happening when you leave a channel and you have a thread of that channel open in the RHS.

koox00:
> There is an error now a bit earlier, when leaving a channel, before accessing threads view.

@jgilliam17 it should be fixed now :)

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified leaving a channel and accessing global threads afterwards no longer causes a crash.

PR #8452, opened 2021-07-23
[MM-36758] Fix App filter and locations

larkox:
#### Summary
We were considering wrongly the input data for the filter a binding, instead of a list of bindings. Also, on the command parser context, we were filling the location with just "/command" instead of the full location.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36758

#### Related Pull Requests
Mobile: https://github.com/mattermost/mattermost-mobile/pull/5567

#### Release Note
```release-note
Fix error with locations and binding filtering.
```


larkox:
@DHaussermann Instructions to test:
- Locations: The test app should print in console the whole context of any call. You just have to check that the location is properly formed.
- Filtering: The current filtering on client does the following. I put alongside each element the element from the test app you shouldn't be seeing in the client.
  - Remove commands without a label (`/test invalid ERROR_with_no_label`)
  - Remove channel header elements without icon or without label (`ERROR_with_no_label`, `ERROR_with_no_icon`)
  - Remove post actions without label (`ERROR_with_no_label`)
  - Remove elements without a call (`/test invalid ERROR_with_no_call`, post action `ERROR_with_no_call`, `ERROR_with_no_call`)
  - Remove elements without app_id (should never happen)

actions/command.test.js
```diff
@@ -72,10 +72,12 @@ const initialState = {
             bindings: [{
                 location: '/command',
                 bindings: [{
+                    location: '/command/appid',
                     app_id: 'appid',
                     label: 'appid',
                     bindings: [
                         {
+                            location: '/command/appid/custom',
```
catalintomai, line 80: @larkox, asked the same in a diff. context (your blog post) - but do we allow the  user define smth. like '/command/appid/custom'? (like are we protecting against that)?

larkox: We are protecting ourselves of locations lower than the base (i.e. if you call your command location `/command/something`, it will be sent back as `/command/command/something`. Since the locations are something used by the app developer, and we only use the base when we use it, there is no "danger" on whatever the app developer does there.

mickmister (approved):
Much smaller PR with whitespace disabled 🙂 

LGTM!

catalintomai (approved):

DHaussermann (approved):
Tested and passed
- No invalid commands are displayed
- This is working as per test steps provided by dev
- No issues found 

LGTM!

**note** Mobile PR is still in progress here https://github.com/mattermost/mattermost-mobile/pull/5567

PR #8457, opened 2021-07-23
Fix console errors

deanwhillier:
#### Summary
Fixes two console errors currently in master:

- tabfocus on `textbox` component was marked as required, wasn't provided in `create_post`, but also wasn't used in `textbox` ... made it optional, used it in `textbox` and defaulted to `0` (the existing hardcoded value)
- direct handling of focus was removed from `settingitemmin` a while back, but still expected by `themesetting` component to be passed to `settingitemmin` ... removed

#### Ticket Link
No ticket, quicker to just fix and push. :P

#### Release Note
Neither of these should have visible impact beyond the ugly console error in dev, so I don't think release notes are necessary

```release-note
NONE
```


jgilliam17 (approved):
Thanks @deanwhillier 
Tested, looks good to merge.
- Verified errors mentioned in the description are no longer reproducing on the test server.

devinbinnie (approved):

components/textbox/textbox.tsx
```diff
@@ -248,7 +248,7 @@ export default class Textbox extends React.PureComponent<Props> {
 
             preview = (
                 <div
-                    tabIndex={0}
+                    tabIndex={this.props.tabIndex || 0}
```
calebroseland, line 251: Maybe?
```suggestion
                    tabIndex={this.props.tabIndex ?? 0}
```
to allow for `-1`

calebroseland (commented):
LGTM, but should tabIndex also be added to SuggestionBox?

components/textbox/textbox.tsx
deanwhillier: Not sure that's necessary? `-1 || 0` in the console returns -1 for me (-1 isn't falsy) ... as does `-1 ?? 0`. Also not sure if there is a preferred approach here? I feel like this is fine as we want the provided value to be a legit number so any falsy values supplied will default to `0`. Of course, typescript should help out here so back at I'm not sure if one is preferred over the other.

deanwhillier:
Thanks @calebroseland!

My only intent here is to correct a console error where `tabindex` was being expected (even though it wasn't even being used) but not provided. Let me know if I misunderstood!

components/textbox/textbox.tsx
calebroseland: Ahhh, my bad! Makes sense to me now.

calebroseland (approved):

PR #8475, opened 2021-07-26
Removes react-window once again

koox00:
#### Summary

Uses `dynamic-virtualized-list` instead of `react-window` for global threads
list. This way we avoid having fixed height thread items and we can have
variable heights instead.

A bit of context: `react-window` was added because `dynamic-virtualized-list` wouldn't reflect
items added at the top. This was due to:
1. `onItemsRendered` callback guards also against changes so it needs to be added even as a noop for changes to be reflected
2. even with `onItemsRendered` added we still need to scroll to reflect the changes at the top of the list.

There is a hack needed for this to work though...To update the rendered list 
while adding new items at the top we have to initiate a scroll by a any amount (1px) and the list will update.
It can be avoided but the list won't update unless the user scrolls, by any amount.

This commit essentially reverts https://github.com/mattermost/mattermost-webapp/pull/8350/commits/5b59cd2ef4384c8b1a60d464201461597d9b6c61
#### Release Note

```release-note
NONE
```


esethna:
@koox00 how confident are we that this is stable using the hack you mentioned? Testing I found two issues (see below). If the advantage to doing this is only that we can have variable height rows, I'm not sure it's worth it for decreased stability.

https://i.gyazo.com/02f9f0523f8b29d495f62198f81ab958.mp4

1. First issue you can see is that a new thread became unread but the list did not update (you can see the unread bubble on the "Unreads tab" but the thread didn't appear in the "All your threads" list until I scrolled around
2. Second issue seend later in the videa is that when a new thread became unread in the list, it caused the lower half of the list of threads to disappear

PR #8486, opened 2021-07-27
fix whitespace in refresh bar

neallred:
#### Summary
Change whitespace in refresh bar to be an `&nbsp;` instead of a `' '`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37426

#### Screenshots
before:
before-nbsp

after:
after-nbsp

#### Release Note
```release-note
Change whitespace in refresh bar so that it always displays to user.
```

### DIFF
--- a/components/announcement_bar/version_bar/version_bar.tsx
+++ b/components/announcement_bar/version_bar/version_bar.tsx
@@ -59,7 +59,7 @@ export default class VersionBar extends React.PureComponent <Props, State> {
                                 id='version_bar.new'
                                 defaultMessage='A new version of Mattermost is available.'
                             />
-                            {' '}
+                            &nbsp;
                             <a onClick={this.reloadPage}>
                                 <FormattedMessage
                                     id='version_bar.refresh'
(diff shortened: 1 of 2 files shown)


components/announcement_bar/version_bar/version_bar.tsx
```diff
@@ -59,7 +59,7 @@ export default class VersionBar extends React.PureComponent <Props, State> {
                                 id='version_bar.new'
                                 defaultMessage='A new version of Mattermost is available.'
                             />
-                            {' '}
+                            &nbsp;
```
pablovelezvidal, line 62: even when this works just fine, I don't see an extended use of the non breaking space across the app. So far what I have seen is either modify the text so it contains the space where necessary or the use of css margins to get the same result. My personal choice would be to add inline css to the anchor tag with half rem:
`                            <a
                                onClick={this.reloadPage}
                                style={{marginLeft: '.5rem'}}
                            >`

pablovelezvidal (approved):
LGTM! thanks @neallred

neallred:
After converting to margin based, looks like this on my machine in local:
with-margin-left

neallred:
Deferring QA testing to post merge when we have an actual upgrade to test against. The setup to try to force an upgrade banner to display is too high relative to the smallness of the change.

nickmisasi (approved):

PR #8490, opened 2021-07-28
[MM-31004] Highlight whole category bounds when holding a channel above a category name

LauSam09:
#### Summary
This PR removes additional styling for (unsorted) categories that results in them appearing differently to, for example, DMs.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/17736

#### Screenshots
demo-3

#### Release Note
```release-note
Made whole category bounds highlighted when holding a channel above a category name
```


hmhealey (commented):
Thanks for the PR, @LauSam09. This does feel nicer to me now that I'm trying it out. 

The changes you've made look good, but one of the E2E tests still looks for the `autoSortedCategory` class if you wouldn't mind looking into fixing that. I think the test uses it specifically to find the DM category, so maybe there's another way we can do that.

The docs for running those tests are [here](https://developers.mattermost.com/contribute/webapp/end-to-end-tests/) if you haven't run those before. Let me know if you have any questions about doing that or if there's anything I could help with.

nevyangelova (approved):

LauSam09:
No problem @hmhealey, thanks for pointing me to the docs and I hope that change is alright!

hmhealey (approved):
Yep, that looks great. Thanks for making that change.

Even if the test hadn't broken, I actually prefer it this way since the old test was kind of looking for information that it shouldn't have to care about anyway.

jgilliam17:
Thanks @LauSam09 
Looks great.
One small issue, when dropping a channel into new, empty category, bottom of the "Drag channels here..." label is too close to the category bounds. @matthewbirtch can you please advise? Thanks

matthewbirtch:
> Thanks @LauSam09
> Looks great.
> One small issue, when dropping a channel into new, empty category, bottom of the "Drag channels here..." label is too close to the category bounds. @matthewbirtch can you please advise? Thanks
> Screen Shot 2021-08-06 at 12 48 09 PM

Good catch @jgilliam17. @LauSam09 are we able to grow the height of the border while dragging in to this space so that it doesn't sit right against the border?

LauSam09:
@matthewbirtch @jgilliam17 Thanks for the spot & suggestion.

I've looked into this and wanted to check with you that you definitely want the extra spacing around the 'Drag channels here...' label _only_ when dragging. I assume I'm being too literal as that will cause displacement of the other channels during dragging - see the attached GIFs.

Add spacing only during drag:
OnDrag

Always add spacing:
Constant

Thanks!

matthewbirtch:
> I've looked into this and wanted to check with you that you definitely want the extra spacing around the 'Drag channels here...' label _only_ when dragging. I assume I'm being too literal as that will cause displacement of the other channels during dragging - see the attached GIFs.

You're right @LauSam09. Let's go with the 'Always add spacing option'

hmhealey (approved):
That sounds like it should be fine. I'd be worried about adding a div to some of the other parts of the sidebar, but the New Category placeholder doesn't have as much complicated logic involved in it.

I tried restarting the CI job to see if that fixes it too. Hopefully it's just an intermittent issue.

jgilliam17 (approved):
Thanks @LauSam09 
Tested, looks good to merge.
- Verified the whole category bounds are highlighted when dragging and dropping a channel. 
Also verified the fix for new category bounds, looks good
Screen Shot 2021-08-13 at 3 10 48 PM

PR #8503, opened 2021-07-29
MM-32342 Remove old sidebar

hmhealey:
Removing the old sidebar lets us also remove a bunch of now-unused logic, preferences, and config settings from the app as well. I've tried to call out the interesting ones below, and everything else should be visible in the commit logs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32342

#### Release Note
```release-note
Removed legacy channel sidebar
```


actions/user_actions.jsx
```diff
@@ -462,13 +440,3 @@ export function autoResetStatus() {
         return userStatus;
     };
 }
-
-export function trackDMGMOpenChannels() {
```
hmhealey, line 466: This only tracked DMs/GMs open on the old sidebar, so it isn't needed any more.

components/channel_header/channel_header.js
```diff
@@ -447,11 +445,7 @@ class ChannelHeader extends React.PureComponent {
                     popoverSize='lg'
                     style={{maxWidth: `${this.state.popoverOverlayWidth}px`, transform: `translate(${this.state.leftOffset}px, ${this.state.topOffset}px)`}}
                     placement='bottom'
-                    className={classNames(['channel-header__popover',
-                        {
-                            'chanel-header__popover--lhs_offset': this.props.hasMoreThanOneTeam,
-                            'chanel-header__popover--new_sidebar': !isLegacySidebar,
-                        }])}
+                    className={classNames('channel-header__popover', {'chanel-header__popover--lhs_offset': this.props.hasMoreThanOneTeam})}
```
hmhealey, line 448: The different margins with the new sidebar have been rolled into the base class.

components/leave_team_modal/index.ts
```diff
@@ -4,36 +4,50 @@
...
     const currentUser = getCurrentUser(state);
+
     return {
         currentUserId,
         currentTeamId,
         show,
         currentUser,
-        privateChannels,
-        publicChannels,
+        numOfPrivateChannels: getNumOfPrivateChannels(state),
```
hmhealey, line 49: Switching these to pass plain numbers wasn't strictly necessary since the main thing was that I was trying to get rid of the selectors that were used here before.

components/needs_team/needs_team.tsx
```diff
@@ -202,7 +200,6 @@ export default class NeedsTeam extends React.PureComponent<Props, State> {
         }
         if (Date.now() - this.blurTime > UNREAD_CHECK_TIME_MILLISECONDS && this.props.currentTeamId) {
             this.props.actions.fetchMyChannelsAndMembers(this.props.currentTeamId);
-            this.props.actions.loadProfilesForDirect();
```
hmhealey, line 205: Despite the name, `loadProfilesForDirect` would update the prefs to ensure unread DMs/GMs were visible in the sidebar when they were unread. Since we always show unread channels in the new sidebar, that's not needed any more.

components/team_sidebar/team_sidebar.tsx
```diff
@@ -2,44 +2,339 @@
...
+            {...props}
+            className='scrollbar--vertical'
+        />);
+}
 
 export default class TeamSidebar extends React.PureComponent<Props, State> {
-    renderTeam = (team: Team) => {
-        return (
-            <TeamButton team={team}/>
+    constructor(props: Props) {
```
hmhealey, line 83: This all came from replacing the dummy `TeamSidebar` component with all the code from the `LegacyTeamSidebar` that we never ended up re-writing. It's otherwise unchanged

components/user_settings/sidebar/user_settings_sidebar.tsx
```diff
@@ -4,829 +4,72 @@
...
-                                defaultMessage='Favorites grouped separately'
-                            />
-                        </label>
-                    </div>
-                    <div>
-                        <br/>
-                        <FormattedMessage
-                            id='user.settings.sidebar.favoritesDesc'
-                            defaultMessage='Channels marked as favorites will be grouped separately.'
+export default function UserSettingsSidebar(props: Props): JSX.Element {
```
hmhealey, line 21: Turns out this thing is incredibly simple now that there's only two settings and they both manage their own saving and rendering.

packages/mattermost-redux/src/actions/channel_categories.ts
```diff
@@ -291,13 +290,6 @@ export function moveChannelToCategory(categoryId: string, channelId: string, new
             return {error};
         }
 
-        // Update the favorite preferences locally on the client in case we have any logic relying on that
```
hmhealey, line 294: The web app no longer uses prefs to determine if a channel is favourited, so this all could be removed.

hmhealey:
For reviews:
- @nevyangelova and @devinbinnie - Could you focus on the actual sidebar code?
- @BenCookie95 - Could you focus on the last few commits for removing System Console code and the final config-related stuff?

nevyangelova (commented):
I feel like there will be a few more CSS changes that we can sneak into this PR but of course there will be digging. Right now there is a lot of conflicting styles in webapp and a lot are related to global header across all devices. What do you think about us thoroughly removing those in this PR? I can also chip in and help a bit :)

hmhealey:
If there's CSS that's only used by the old sidebar, we can remove that, but I'd rather not get into too much more cleanup since this is already so large. I probably went a bit overboard with what I included in here to begin with.

components/user_settings/sidebar/user_settings_sidebar.tsx
```diff
@@ -4,829 +4,72 @@
...
-    favoriteAtTop: string;
-
-    /**
-     * Setting if the legacy sidebar is enabled
-     */
-    enableLegacySidebar: boolean;
-
-    updateSection?: (section: string) => void;
-    activeSection?: string;
+export interface Props {
```
devinbinnie, line 14: Probably don't need to export this anymore.

devinbinnie (approved):
image

nevyangelova (approved):
LGTM and I am so happy about it lol but I would open a follow up ticket to remove even more CSS if needed ;)

components/admin_console/admin_definition.jsx
```diff
@@ -6040,63 +6040,6 @@ const AdminDefinition = {
                         isDisabled: it.not(it.userHasWritePermissionOnResource(RESOURCE_KEYS.EXPERIMENTAL.FEATURES)),
                         isHidden: it.configIsFalse('FeatureFlags', 'CollapsedThreads'),
                     },
-                    {
```
BenCookie95, line 6043: Shouldn't we also remove `ServiceSettings.EnableLegacySidebar` from admin_definition? (Line 6005)

actions/user_actions.test.js
```diff
@@ -595,10 +583,4 @@ describe('Actions.User', () => {
         expect(UserActions.queue.onEmpty).toHaveBeenCalled();
         expect(UserActions.queue.add).toHaveBeenCalled();
     });
-
```
BenCookie95, line 598: Line 333 & 100 in this file, we no longer need `ExperimentalChannelSidebarOrganization`

BenCookie95 (commented):
I also see `EnableXToLeaveChannelsFromLHS` in `sidebar_base_channel.test.tsx` but apart from that everything looks good!

hmhealey:
Good catch. I know some of this was due to using a case sensitive search, but I don't know how I missed some other ones.

BenCookie95 (approved):
Nice work, LGTM!

jgilliam17:
@hmhealey there is one error I am seeing after re-testing failures from [cypress test run ](https://mm-cypress-report.s3.amazonaws.com/352979702-webapp-pr-8503-webapp-pr-8503-mattermost-enterprise-edition:6687022/mochawesome.html)
Screen Shot 2021-08-16 at 4 29 34 PM

jgilliam17:
One more thing I just noticed - If a user has mention badge on the channel and is removed from that channel by another user, the channel is still available on the sidebar for them, just mentions get removed.

hmhealey:
@jgilliam17 I've fixed the error you encountered with the channel switcher. I can repro the second bug you found on mysql.test, so I'll file a separate ticket for that

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified old sidebar and all settings for old sidebar in system console have been removed.

PR #8509, opened 2021-07-30
[MM-36274] Catch all updates on Global Header with Compass

nevyangelova:
#### Summary
First iteration of the global header with compass components. Changes to be reviewed: 
- Status dropdown moved to the top right corner of the global header. Account settings are in that menu accordingly.
- Main menu renders only team specific items when global header is enabled. 
- Product switcher menu shows the list of the current products as well as system settings. Team/sysadmin permissions apply.
- Search bar and mentions + saved posts buttons are a part of the global header. Search nav bar should be removed from the channel header. Check RHS search bar as well
- When global header is disabled or we are viewing from a mobile device there should be no changes at this time, even if the design shows the product switcher button at the moment.
- Product switcher menu is currently restructured to use the existing menu items for simplicity. Proposal to refactor that when the new Menu common component is released. 

Design reference https://www.figma.com/file/iR1U1mW4y1U8S7qXWT9O1l/MM-35325-Global-Header-MVP?node-id=0%3A1

#### Release Note

-->
```release-note
NONE
```


deanwhillier (commented):
Still need to do a thorough review, but I noticed the below while working on the theming.

michaelgamble (changes requested):
**Status dropdown**
- should have a hover state bounding box
- Probably kill the little dropdown arrow
- Should only be 12px from the right edge of hover / active bounding box to right edge of screen
- I think the profile bounding box is a bit off in terms of overall width
- I would use this screen for reference for right alignment https://www.figma.com/file/iR1U1mW4y1U8S7qXWT9O1l/MM-35325-Global-Header-MVP?node-id=66%3A62770
- The hover state for it should match the same opacity as the other buttons hover state

**Old main menu**
- Icon and vertical alignment is wrong
- This icon should be an icon button instead of the whole row being clickable
- Menu Items
    - Leave team should use the newer destructive red styling (Reference the delete option for categories)

**Product Switcher**
- The product switcher icon is showing an toggle state when it should be an active state
- The padding seems off for the menu’s from what’s in the design
- The “Open…” is too close to the top edge
- The product icon color seems off - its supposed to be buttonBG
- The product icon size is incorrect it should be a 24x24 icon
- The product icon space between icon and text should be 8px
- The hover state seems to bleed out of the bounds
- The separator line opacity seems too dark it should be %8 centre channel
- The non product icons seem too dark - they should be 56% centre channel
- The non product icon space between text should be 12px
- Marketplace red dot location? @matthewbirtch ?

**Product icon/name in header**
- Gap between icon and name should be 8px
- Product name should be using metropolis regular (500 weight)

**Navigation arrows**
- Should have a 1px gap between the buttons

**Search field**
- I’m not 100% but for some reason every once in a while it keeps bugging out with text data of “@sysadmin” in the field which I didn’t add
- The search component is too wide it should be 432 in width
- MORE TO COME ON THIS ONCE I CAN DESIGN THE DIFFERENT STATES

**Help Button**
- It showing a toggle state when it should be an active state

deanwhillier:
@michaelgamble, thanks for the awesome feedback so far!

All, we will be taking note of all the visual feedback and making updates in the next PR for this project rather than this one so we can get it merged and running in daily. For this PR, we will only be dealing with major functional issues and broken user experience. Hope that makes sense!

@nevyangelova, maybe followup with comments saying what you have fixed from feedback and keep track of the rest for our next pass.

All, please assume that if Nevy hasn't mentioned an update to certain items in your feedback when she marks you for a re-review, we will be making those changes in the next PR!

Reach out in the UI-Platform channel for any needed discussion!

calebroseland (changes requested):
Overall looks great! Lots going on, but I think I got through most of it enough. One minor request for event handlers. Leaving out UX commentary, except one question on some tooltips, and a small font/styling glitch that occurs on first-click of any button in global header. 

CleanShot 2021-08-09 at 10 11 07

jasonblais:
I was trying it out and happened to see a few small issues - please don't treat as a complete review, but wanted to report so they can be tracked accordingly,

1 - Clicking on `Channel Preferences` in the global header does nothing
2 - Do Not Disturb time options go off screen - the menu options should open to the left instead. If so, the `>` arrow should probably be adjusted accordingly (0/5 how, would defer to UX)
3 - Each time I click on the back/forward arrows, I get JS errors
image
4 - No tooltip for back/forward arrows, and I was at first confused what they do. I thought it may switch between the products (Channels > Playbooks > Boards)
5 - Let's remove the red dot next to Marketplace in the "top" main menu. Discussed previously with Aaron and Eric Sethna. No longer relevant with the new global header layout.
image


image

components/global/global_header.tsx
```diff
@@ -1,82 +1,131 @@
...
+        margin-left: 8px;
+    }
 `;
 
 const GlobalHeader = (): JSX.Element | null => {
     const enabled = useSelector(getGlobalHeaderEnabled);
     const products = useProducts();
     const currentProductID = useCurrentProductId(products);
 
+    useEffect(() => {
```
crspeller, line 81: This is fragile. Maybe we could use styled components for this?

components/global/global_search_nav/global_search_nav.tsx
```diff
@@ -0,0 +1,99 @@
...
+        if (Utils.cmdOrCtrlPressed(e) && !e.shiftKey && Utils.isKeyPressed(e, Constants.KeyCodes.K)) {
+            if (!e.altKey) {
+                e.preventDefault();
+                toggleQuickSwitchModal();
+            }
+        }
+    };
+
+    return (
+        <Flex
```
crspeller, line 85: Seems strange to have a component for a flex container. Why not just a styled component?

components/global/global_search_nav/global_search_nav.tsx
```diff
@@ -0,0 +1,99 @@
...
+            closeModal(ModalIdentifiers.QUICK_SWITCH);
+        } else {
+            openModal({
+                modalId: ModalIdentifiers.QUICK_SWITCH,
+                dialogType: QuickSwitchModal,
+            });
+        }
+    };
+
+    const handleQuickSwitchKeyPress = (e: KeyboardEvent) => {
```
crspeller, line 75: Is this really the place to handle this? Maybe we can locate this more appropriately.

components/global/global_search_nav/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    actions: bindActionCreators({
+        showMentions,
+        openRHSSearch,
+        closeRightHandSide,
+        openModal,
+        closeModal,
+    }, dispatch),
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(GlobalSearchNav);
```
crspeller, line 48: Again redux hooks! https://react-redux.js.org/api/hooks

components/global/history_buttons/history_buttons.tsx
```diff
@@ -0,0 +1,56 @@
...
+import {browserHistory} from 'utils/browser_history';
+
+const HistoryButtonsContainer = styled.nav`
+    display: flex;
+    align-items: center;
+`;
+
+const HistoryButtons = (): JSX.Element | null => {
+    if (!isDesktopApp()) {
+        // need to re-enable this
```
crspeller, line 20: ?? Guessing this should be disabling in the desktop app.

components/global/history_buttons/history_buttons.tsx
```diff
@@ -0,0 +1,56 @@
...
+        // return null;
+    }
+
+    const goBack = () => {
+        trackEvent('ui', 'ui_history_back');
+        browserHistory.goBack();
```
crspeller, line 26: I think the correct way to do this now is with the `useHistory` hook: https://reactrouter.com/web/api/Hooks/usehistory

components/global/product_switcher/product_switcher.tsx
```diff
@@ -60,6 +66,14 @@ const MenuItem = styled(Link)`
         text-decoration: none;
         color: inherit;
     }
+
+    i {
```
crspeller, line 70: Is this the proper way to style a compass component? Shouldn't we be doing something like:

```js
const StyledIcon = styled(Icon)`
    color: var(--sidebar-bg);
`
```

components/global/product_switcher_menu/product_switcher_menu.tsx
```diff
@@ -0,0 +1,144 @@
...
+    enableOutgoingWebhooks: boolean;
+    canManageSystemBots: boolean;
+    canManageIntegrations: boolean;
+    enablePluginMarketplace: boolean;
+    pluginMenuItems: any;
+    intl: IntlShape;
+    firstAdminVisitMarketplaceStatus: boolean;
+};
+
+class ProductSwitcherMenu extends React.PureComponent<Props> {
```
crspeller, line 43: Functional component?

components/global/product_switcher/product_switcher.tsx
```diff
@@ -102,16 +130,35 @@ const ProductSwitcher = (): JSX.Element => {
...
+                <IconButton
+                    icon={'products'}
+                    onClick={handleClick}
+                    size={'sm'}
+                    compact={true}
+                    toggled={switcherOpen}
+                    inverted={true}
+                    aria-label='Select to open product switch menu.'
+                />
+                <ProductBranding>
```
crspeller, line 146: This should only show for channels right?

components/global/product_switcher/product_switcher.tsx
```diff
@@ -70,11 +84,25 @@ const MenuItemTextContainer = styled.div`
...
-    height: 14px;
-    color: rgba(var(--center-channel-color-rgb), 0.56);
-`;
+const SwitcherNavEntry = (props: SwitcherNavEntryProps) => {
+    return (
+        <MenuItem
+            to={props.destination}
+        >
+
+            <IconButton icon={props.icon}/>
```
crspeller, line 93: props.icon is a component not a string. This will break the current integrations. Can we put this back to be more flexible?

components/global/user_guide_dropdown/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import {getConfig} from 'mattermost-redux/selectors/entities/general';
+
+import {GlobalState} from 'types/store';
+
+import UserGuideDropdown from './user_guide_dropdown';
+
+function mapStateToProps(state: GlobalState) {
+    const {HelpLink, ReportAProblemLink, EnableAskCommunityLink} = getConfig(state);
+    return {
+        helpLink: HelpLink!,
```
crspeller, line 15: Are these really guaranteed to be not null? Or are we cheating typescript here?

components/global/user_guide_dropdown/user_guide_dropdown.tsx
```diff
@@ -0,0 +1,130 @@
...
+        this.setState({
+            buttonActive: menuActive,
+        });
+    }
+
+    askTheCommunityClick = () => {
+        trackEvent('ui', 'help_ask_the_community');
+    }
+
+    renderDropdownItems = (): React.ReactNode => {
```
crspeller, line 51: Whenever I see `render` in a function name for a component I always ask myself if this should just be a separate component or just in-lined.

components/global/user_guide_dropdown/user_guide_dropdown.tsx
```diff
@@ -0,0 +1,130 @@
...
+            >
+                <OverlayTrigger
+                    delayShow={500}
+                    placement='bottom'
+                    overlay={this.state.buttonActive ? <></> : tooltip}
+                >
+                    <IconButton
+                        size={'sm'}
+                        icon={'help-circle-outline'}
+                        onClick={() => {}} // icon button currently requires onclick ... needs to revisit
```
crspeller, line 110: This seems like a design flaw in the icon button component.

components/global/user_guide_dropdown/user_guide_dropdown.tsx
```diff
@@ -0,0 +1,130 @@
...
+                        toggled={this.state.buttonActive}
+                        inverted={true}
+                        compact={true}
+                        aria-label='Select to toggle the help menu.' // proper wording and translation needed
```
crspeller, line 114: Should we not handle this now?

components/main_menu/main_menu.jsx
```diff
@@ -179,7 +181,113 @@ class MainMenu extends React.PureComponent {
             />
         );
 
-        return (
+        return this.props.globalHeaderEnabled && !this.props.mobile ? (
```
crspeller, line 184:
```suggestion
        return (this.props.globalHeaderEnabled && !this.props.mobile) ? (
```
Removes ambiguity.

crspeller (changes requested):

components/global/at_mentions_button/at_mentions_button.tsx
```diff
@@ -0,0 +1,63 @@
...
+            />
+        </Tooltip>
+    );
+
+    return (
+        <OverlayTrigger
+            trigger={['hover']}
+            delayShow={Constants.OVERLAY_TIME_DELAY}
+            placement='bottom'
+            overlay={isRhsOpen ? <></> : tooltip}
```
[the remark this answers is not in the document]
nevyangelova: That is a good question for UX as I just copied the previous behaviour. Maybe @michaelgamble can advice?

nevyangelova:
> I was trying it out and happened to see a few small issues - please don't treat as a complete review, but wanted to report so they can be tracked accordingly,
> 
> 1 - Clicking on `Channel Preferences` in the global header does nothing
> 2 - Do Not Disturb time options go off screen - the menu options should open to the left instead. If so, the `>` arrow should probably be adjusted accordingly (0/5 how, would defer to UX)
> image
> 3 - Each time I click on the back/forward arrows, I get JS errors
> image
> 4 - No tooltip for back/forward arrows, and I was at first confused what they do. I thought it may switch between the products (Channels > Playbooks > Boards)
> 5 - Let's remove the red dot next to Marketplace in the "top" main menu. Discussed previously with Aaron and Eric Sethna. No longer relevant with the new global header layout. Feel free to consider out of scope as a separate ticket for Channels team.
> image

Thank you for the feedback Jason! On these points: 
1. This ticket is in progress as we need to split the account settings modal in two parts and when we have the messaging modal we will connect it to that button. 
2. Will fix in this PR
3. Will fix in this PR although it might be related to some missing plugins 
4. Will fix in this PR
5. Will fix in this PR

components/global/user_guide_dropdown/user_guide_dropdown.tsx
nevyangelova: We can remove it but its good to keep it as a reminder IMO

components/global/user_guide_dropdown/user_guide_dropdown.tsx
nevyangelova: For sure. We will discuss aligning compass components with the needs for webapp in our team meeting today

components/global/at_mentions_button/at_mentions_button.tsx
```diff
@@ -0,0 +1,63 @@
...
+    rhsState: typeof RHSStates[keyof typeof RHSStates] | null;
+    isRhsOpen?: boolean;
+    actions: {
+        showMentions: () => void;
+        closeRightHandSide: () => void;
+    };
+};
+
+const AtMentionsButton = (props: Props): JSX.Element => {
+    const {rhsState, isRhsOpen, actions: {closeRightHandSide, showMentions}} = props;
```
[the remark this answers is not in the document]
nevyangelova: haha I think this is definitely one of those things that are simple but harder to agree on, I much prefer destructuring props all together but since there isn't much repetition in these components I can bring it back.

components/compass_theme_provider/compass_theme_provider.tsx
```diff
@@ -0,0 +1,56 @@
...
+import ThemeProvider, {lightTheme} from '@mattermost/compass-components/utilities/theme';
+
+import {Theme} from 'mattermost-redux/types/themes';
+
+type Props = {
+    theme: Theme;
+    children?: React.ReactNode | React.ReactNode[];
+}
+
+const CompassThemeProvider = ({theme, children}: Props): JSX.Element | null => {
```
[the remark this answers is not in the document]
nevyangelova: @deanwhillier might be better to ask about the theme provider changes.

jasonblais:
@nevyangelova Thank you :+1:

components/global/product_switcher_menu/product_switcher_menu.tsx
nevyangelova: The menus across webapp will be replaced soon with the common component so for now we are just copying the old menus as we are just splitting them for the purpose of the header.

components/global/global_search_nav/global_search_nav.tsx
nevyangelova: you mean is this component the place? I guess we can handle this in the parent but the quick switcher has its own events too so to unify those would be nice but I would say it should be done in another PR

components/global/global_search_nav/global_search_nav.tsx
nevyangelova: The idea for the flex utility is to avoid the need to create the styled components separately and nest divs, this container already gives us a flex parent.

components/global/global_header.tsx
nevyangelova: I guess it was implemented that way due to the fact that the feature flag will be removed soon. I can bring it up for discussion in out team meeting today.

components/global/settings_button/settings_button.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <Tooltip id='channelPreferences'>
+            <FormattedMessage
+                id='channel_header.channelPreferences'
+                defaultMessage='Channel Preferences'
+            />
+        </Tooltip>
+    );
+
+    return (
+        <OverlayTrigger
```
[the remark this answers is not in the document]
nevyangelova: Or just a tooltip. We are planning to use our popover component for tooltips from now on.

components/global/user_guide_dropdown/user_guide_dropdown.tsx
nevyangelova: true, this will be replaced with the menu common component soon as well so we didn't focus on optimisation, just moving functionality.

components/global/product_switcher/product_switcher.tsx
nevyangelova: No I think when we get the other products we will remove the hardcoded values and do this for all of them.

components/global/product_switcher/product_switcher.tsx
nevyangelova: I am not sure on this point since we will replace all existing icons with the compass one anyway which takes a string as the icon glyph. But this shouldn't be an icon button for sure, i will replace it.

components/global/history_buttons/history_buttons.tsx
nevyangelova: ahh this is just so that UX or QA can see it in the test server, it will be removed after the test is done. These history arrows should only show in desktop app.

components/global/history_buttons/history_buttons.tsx
nevyangelova: will have to discuss with desktop app team as well.

components/global/at_mentions_button/at_mentions_button.tsx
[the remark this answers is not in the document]
deanwhillier: I think that may have been related to an issue that came up a while back where tooltips overlapped menus, this may have been the approach to fix it back then.

I agree that it would be good to still have tooltips regardless of whether the RHS is open.

@nevyangelova, I would say that we can remove this in the Global Header (leave whatever functionality is there when the GH is off), but that it can be done with the next PR as it's not functionally critical for this PR.

components/global/global_header.tsx
```diff
@@ -1,82 +1,131 @@
...
-                        globalHeader={true}
+                <HistoryButtons/>
+            </LeftControls>
+            <CenterControls>
+                { currentProductID !== null &&
+                    <Pluggable
+                        pluggableName={'Product'}
+                        subComponentName={'headerComponent'}
+                        pluggableId={currentProductID}
                     />
```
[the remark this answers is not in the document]
deanwhillier: @calebroseland, seems the linter should pick this up, no?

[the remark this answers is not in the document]
calebroseland: I'm 0/5 as to which one, but Indeed—eslint should pick it up :). I see it as the same as `eslint/object-curly-spacing`, but I guess it's in a separate rule `react/jsx-curly-spacing`.

components/compass_theme_provider/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import CompassThemeProvider from './compass_theme_provider';
+
+export default CompassThemeProvider;
```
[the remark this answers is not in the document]
deanwhillier: There was a conversation a while back about with a decision to always use index files to create a public 'API' for all components. I believe @calebroseland and @jespino were part of that conversation as well. We'll leave it like this unless that decision is reverted.

Thoughts @calebroseland and @jespino?

components/compass_theme_provider/compass_theme_provider.tsx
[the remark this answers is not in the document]
deanwhillier: Complicated how?

It's a component to map the webapp theme to the current implementation of the Compass Components theme, which in itself is currently complicated as the Compass Design System theme/colour system isn't directly compatible with the older webapp theming system.

We do plan to revisit this, but need something up and running for now.

components/global/at_mentions_button/at_mentions_button.tsx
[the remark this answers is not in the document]
deanwhillier: I personall agree with the destructuring approach, although it could possibly be done directly in the function parameter list instead of a new const statement.

```javasctipt
const MyFunction = ({prop1, prop2}: Props): JSX.Element => {...}
```

Modern IDE's make it very easy to determine the origin of a variable, though I do understand the momentary uncertainty! The challenge is that we are going to have many different opinions on syntax style preferences and so would need to agree on a standardized approach. I personally like the idea of adopting an existing style guide that is considered by the community in general to be a decent standard to follow vs trying to define one ourselves.

components/global/user_guide_dropdown/index.ts
deanwhillier: Agreed, might be best to check their contents first.

components/global/global_header.tsx
deanwhillier: @crspeller, this is needed to update styles throughout the rest of the app that need to change only when the Global Header is enabled. It is the equivalent of a feature flag for JS but makes it easier to group styles together with the Global Header vs. trying to keep track of and update all the differences in the various SASS files spread throughout the code base and support the GH being on or off. It is not possible to convert everything outside the Global Header that need CSS changes when the GH is on given the available time.

This absolutely should be replaced with applying styling changes directly when the Global Header is no longer a feature.

components/global/global_search_nav/global_search_nav.tsx
deanwhillier: This particular component is an early/foundational implementation of a planned layout component that is intended to:

* make it easer for a dev to layout components without having to know/formulate CSS
* that inherently adheres to the Compass Design layout guidelines for consistency
* and removes the need to constantly (re)write the same spacing/padding CSS in styled components every time there is a need to space components in a consistent manner.

For example, if several icon buttons need to be laid out horizontally with pre-determined 8px spacing based on the Design System (for instance), a dev would be able to do something like this:

```javascript
<Row spacing={8}>
    <IconButton ... />
    <IconButton ... />
    <IconButton ... />
    <IconButton ... />
</Row>
```

Typescript would then suggest/enforce the pre-determined/approved spacing from the design system.

Flex may be the foundation element under the hood that makes something like this possible, but we're not quite there yet.

The one thing I like about styled-components is that you can have a component name that is more semantic to what its purpose is, which is something to consider as well. Though in that case, it would obviously be possible to do something this to help maintain consistency with the Design system and reduce repetition in code:

```javascript
const GlobalSearchNav = styled(Row)``;
```

(Not sure off hand if the back ticks are needed here)

components/global/global_search_nav/global_search_nav.tsx
deanwhillier: I agree that this doesn't belong to the Search Nav component as it's not functionally related.

@nevyangelova, do you know where this was implemented before? Functionally it belongs to the Channel Sidebar where the Quick Switcher UI lives and IMO it should stay there.

components/global/history_buttons/history_buttons.tsx
deanwhillier: This is porting existing functionality. If we want to handle differently moving forward, we should open a ticket and deal with it in a separate PR.

components/global/product_switcher/product_switcher.tsx
deanwhillier: Once we have a functionally integrated plugin for one of the other suites, we will make necessary updates to exactly how this (and other areas) works.

components/global/user_guide_dropdown/user_guide_dropdown.tsx
deanwhillier: One step at a time, this is a placeholder of additional work that still needs to be done.

I created a placeholder ticket: https://mattermost.atlassian.net/browse/MM-37792

nevyangelova:
> @nevyangelova Thank you 👍

Jason updated, we won't be adding the tooltips in this PR as they aren't present and not sure what the text should be. Also the status dropdown looks off at the moment and its good to have a design for it. @michaelgamble can you perhaps check out the current state of the DND status dropdown when its opening left and advice on the UI changes for the next styling PR.

deanwhillier:
@michaelgamble, as discussed in our team meeting this morning, we will be addressing most of your feedback in a subsequent PR that will be more focused around visual fidelity.

Placeholder ticket here: https://mattermost.atlassian.net/browse/MM-37785

deanwhillier:
Checked with @esethna and he will do a PM review of the next PR so we can move forward without on this one.

components/global/product_switcher/product_switcher.tsx
crspeller: @deanwhillier We do. Boards is done, just not on community yet. 
@nevyangelova Maybe we shouldn't break it by adding this now? As written this will show up even though boards or playbooks is active (unless I am missing something).

components/global/product_switcher/product_switcher.tsx
crspeller: @nevyangelova This will break the current implementation that Boards has already merged resulting in a white screen on community. Probably easier to just set this back to `{props.icon}`? Or you can work with them to make sure their implementation is changed.

components/global/user_guide_dropdown/user_guide_dropdown.tsx
crspeller: A ticket is a great reminder to do something later.

esethna:
In case it's not a known issue @nevyangelova 

image

michaelgamble (approved):

components/global/global_search_nav/global_search_nav.tsx
nevyangelova: No it is in the RHSSearchNav which is only rendered instead of the global header so for now it should be present in both. I can refactor it to be handled by channel sidebar but i would say we can open a ticket and do it after as this was just moving functionality.

components/global/user_guide_dropdown/user_guide_dropdown.tsx
nevyangelova: all these tickets are in our active sprint so we can't avoid it :D

nevyangelova:
> In case it's not a known issue @nevyangelova
> 
> image

thank you @esethna I didn't notice that at all! Fixed

components/global/global_search_nav/global_search_nav.tsx
deanwhillier: Interesting. Ya, can you open that ticket and then we can move on as is for now?

nevyangelova: https://mattermost.atlassian.net/browse/MM-37814

deanwhillier (approved):

jgilliam17:
Thanks @nevyangelova seeing few issues
1. Channel header is covering a small portion of the RHS header when viewing in full screen mode. Left arrow on DND menu item is causing a misalignment, I'll add that to the polish PR
Screenshot 2021-08-11 151448

2. Totally an edge case, but when banner is present search in mobile view now has an extra glitchy header
Screen Shot 2021-08-11 at 2 22 32 PM

3. Product Switcher icon in Ubuntu has some extra space, bottom 3 blocks are separated from the top 6. I should probably add this to the follow up PR
Screenshot 2021-08-11 152119

components/global/product_switcher.tsx
```diff
@@ -73,11 +87,25 @@ const MenuItemTextContainer = styled.div`
...
-    height: 14px;
-    color: rgba(var(--center-channel-color-rgb), 0.56);
-`;
+const SwitcherNavEntry = (props: SwitcherNavEntryProps) => {
+    return (
+        <MenuItem
+            to={props.destination}
+        >
+
+            <StyledIcon glyph={props.icon}/>
```
crspeller, line 96: This still needs to be addressed. Ether revert to a raw `{props.icon}` or work with Boards team to make sure we don't break community.

components/global/product_switcher.tsx
```diff
@@ -107,56 +135,55 @@ const ProductSwitcher = (): JSX.Element => {
...
-                </SwitcherMenuDescriptiveText>
-                <SwitcherNavEntry
-                    destination={'/'}
-                    icon={<ChannelsIcon/>}
-                    text={'Channels'}
-                    active={currentProductID === null}
-                />
-                {items}
-            </SwitcherMenu>
+                    <ProductBranding>
```
crspeller, line 151: This needs to be resolved. My suggestion is to remove it for now. 
It's also a bit confusing to be here anyway. It probably should be with the other product specific stuff or at the top level of the global header component? It's a bit confusing to have it lumped with the switcher menu.

crspeller (changes requested):
I don't think this was already reported. Expanding the modified RHS causes overlap: 
image

components/global/product_switcher.tsx
deanwhillier: That one's on me, I did the initial setup and forgot I hadn't updated to dynamically adjust based on product.

components/global/product_switcher.tsx
deanwhillier: It will be better to provide a string of the desired product icon glyph from Compass Icons instead of an icon component to ensure consistency of usage.

For this PR we can revert to the raw `{props.icon}` to avoid issues when merging and can then ensure that we talk with the product teams about the change.

@nevyangelova, lets go with this change for now and we can update in the next PR
```suggestion
            {props.icon}
```

calebroseland: @nevyangelova cc @deanwhillier @crspeller perhaps something along these lines for the time being?

```tsx
const SwitcherNavEntry = (props: SwitcherNavEntryProps) => {
    let icon: React.ReactNode;
    if (isValidElementType(props.icon) && typeof props.icon !== 'string') {
        const ProductIcon = props.icon;
        icon = <ProductIcon/>;
    } else if (React.isValidElement(props.icon)) {
        icon = props.icon;
    } else if (typeof props.icon === 'string') {
        icon = <StyledIcon glyph={props.icon as TIconGlyph}/>;
    }

    return (
        <MenuItem
            to={props.destination}
        >
            {icon}
            <MenuItemTextContainer>
                {props.text}
            </MenuItemTextContainer>
            {props.active &&
                <Icon
                    size={16}
                    glyph='check'
                />
            }
        </MenuItem>
    );
};
```

components/global/global_search_nav/global_search_nav.tsx
```diff
@@ -0,0 +1,96 @@
...
+            }
+            if (Utils.isKeyPressed(e, Constants.KeyCodes.L)) {
+                // just close the modal if it's open, but let someone else handle the shortcut
+                dispatch(closeModal(ModalIdentifiers.QUICK_SWITCH));
+            }
+        }
+    };
+
+    const toggleQuickSwitchModal = () => {
+        const isQuickSwitcherOpen = useSelector((state: GlobalState) => isModalOpen(state));
```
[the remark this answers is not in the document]
deanwhillier: I'll update this, stay tuned ...

calebroseland (approved):
LGTM besides the outstanding ones @crspeller posted

components/global/global_search_nav/global_search_nav.tsx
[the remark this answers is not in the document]
deanwhillier: I added a hook to get modal open state that returns the direct boolean value for regular usage and a ref with the value for usage in callbacks.

deanwhillier:
@crspeller, we'll deal with the "...RHS causes overlap..." in the next PR, thanks.

components/global/hooks.tsx
```diff
@@ -99,3 +101,18 @@ export const useShowTutorialStep = (stepToShow: number, products?: ProductCompon
 
     return showTutorialStep;
 };
+
+/**
+ * Hook that returns the current open state of the specified modal
+ * - returns both the direct boolean for regular use and a ref that contains the boolean for usage in a callback
```
crspeller, line 107: For my own reference, why return a ref here too?

crspeller (approved):
Great! Lots of work here. Can't wait to see the polished product :)

components/global/hooks.tsx
deanwhillier: So you can't use the directly returned value from the hook in the eventListener callback as it will only have access to the original value and not any of the subsequent updates. My research into it a while back indicated you need to use a ref in that case, but if you don't need to reference it from this kind of callback, it's better to use the direct value. Since that is the most likely use case of this kind of hook, it's returned first.

I initially thought that the re-rendering of the component would have provided the callback with the new value, but the component doesn't re-render on native DOM events so ...

I did try again anyway, but no dice.

If you have any ideas for a better way to handle this particular kind of situation, I'm all ears!

jgilliam17 (approved):
Thanks @nevyangelova and @deanwhillier 
Looks good to merge. 
- Verified functional aspects of the feature, while keeping in mind we have a follow up PR in the works to address additional UX, visual issues. I added few last comments to the follow up PR.

PR #8510, opened 2021-07-30
Make rootId prop optional for Textbox

hmhealey:
Something definitely got messed up here, probably by me.

To make it optional, I had to fix the type of the arguments to a couple selectors to avoid making everything weird. I think this makes everything look nicer overall anyway.

#### Release Note
```release-note
NONE
```


deanwhillier (approved):
Lol, one little question mark caused all that :P

devinbinnie (approved):

PR #8515, opened 2021-08-02
MM-36234,MM-37030,MM-37031: CRT, desktop notification settings

koox00:
#### Summary

This commit makes needed changes to desktop notifications depending on
CRT preference.
Adds a new `user.notify_props` setting called `desktop_threads`,
which can be either `ALL` or `MENTION` with a default set to `ALL`.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36234

#### Related Pull Requests

- Has server changes (https://github.com/mattermost/mattermost-server/pull/18088)
- Has mobile changes (https://github.com/mattermost/mattermost-mobile/pull/5618)
- Has desktop changes (https://github.com/mattermost/desktop/pull/1699)

#### Screenshots

image

#### Release Note

```release-note
NONE
```


actions/notification_actions.jsx
```diff
@@ -58,11 +79,17 @@ export function sendDesktopNotification(post, msgProps) {
...
-        if (notifyLevel === NotificationLevels.DEFAULT) {
-            notifyLevel = user && user.notify_props ? user.notify_props.desktop : NotificationLevels.ALL;
+        // if collapsedThreads is enabled and the post received is a Reply
+        // and the thread is not a UserThread (AKA followed thread) do not notify
+        if (isCrtReply) {
+            const thread = await dispatch(getThread(user.id, channel.team_id, post.root_id, false));
+            if (thread.error) {
+                return;
+            }
         }
```
koox00, line 89: This could probably be done better, right now when CRT is on it calls the API on each reply received.
This is done to make sure to get a notification only when the thread is followed.

*when the thread is not followed the API returns an error code.

anurag6713: Good work @koox00 🙂 Have you checked if we are also receiving `post.is_following` for this?

koox00: Thanks @anurag6713, unfortunately it's not, AFAICT we don't return that kind of info on single posts.
So maybe one solution would be to support that server side,
and another solution would be to hook on `thread_updated` event for CRT notifications and `posted` event for non CRT replies and root posts.

hmhealey: The websocket event received here also isn't per-person, so I don't think we could include whether or not the recipient is following the thread on the event.

That said, is there not a way to use the existing data loaded by the app to prevent having to make this check for each reply. That seems rather heavy.

actions/notification_actions.jsx
```diff
@@ -22,10 +25,30 @@ import {stripMarkdown} from 'utils/markdown';
 
 const NOTIFY_TEXT_MAX_LENGTH = 50;
 
+export function getNotifyLevel(isCrtReply, userSettings, channelSettings) {
+    let notifyLevel = userSettings?.desktop;
+
+    if (isCrtReply) {
+        if ([NotificationLevels.NONE, NotificationLevels.ALL].includes(notifyLevel)) {
```
hmhealey, line 32: Is there a way to rewrite this logic in a way that makes it easier to follow? Between the multiple returns and the various `||` operators, I'm having trouble figuring out when each setting is used by looking at this code.

actions/notification_actions.jsx
hmhealey: Alternatively, maybe we could have the server treats replies to followed threads as mentions so the client doesn't need to do as much work here? That's probably a lot larger change though since I don't know how much the notification code is aware of CRT. It would also make this parallel the old reply mentions feature as well.

hmhealey (changes requested):

actions/notification_actions.jsx
hmhealey: 4/5 though that we should try to avoid repeatedly calling `getThread` through. If we could get it so that we're only calling it on the first time a user receives a post in the thread, that seems acceptable to me though.

koox00: @hmhealey `notification.go` is an intimidating module as it does lots of stuff! 
But for the rest of the notifications to work (email, push) with CRT, we would need to modify it, so your proposition makes a lot of sense, thanks!

koox00: @hmhealey logic has been moved to server and `notification_actions` has been reverted, for the most part, as it was :tada:
related server PR https://github.com/mattermost/mattermost-server/pull/18088

hmhealey (approved):
This looks great. Glad to see that moving stuff to the server for push/email notifications did end up reducing the changes needed for the desktop ones.

anurag6713 (approved):

actions/notification_actions.jsx
```diff
@@ -168,6 +198,7 @@ const notifyMe = (title, body, channel, teamId, silent, soundName) => (dispatch,
 
         if (isDesktopApp() && window.desktop && semver.gte(window.desktop.version, '4.6.0')) {
             msg.data = {soundName};
+            msg.url = url;
```
Willyfrog, line 201: 4.6 doesn't understand this parameter, it should at least be version '4.7.2' if we backport the change.

components/logged_in/index.js
```diff
@@ -30,7 +30,15 @@ function mapStateToProps(state, ownProps) {
 }
 
 // NOTE: suggestions where to keep this welcomed
-const getChannelURLAction = (channel, teamId) => (dispatch, getState) => browserHistory.push(getChannelURL(getState(), channel, teamId));
+const getChannelURLAction = (channel, teamId, url) => (dispatch, getState) => {
+    const state = getState();
+
+    if (url) {
+        return browserHistory.push(url);
```
Willyfrog, line 37: we should check that the url belongs to the server (does it contain the SiteUrl? or make sure it is a relative path) otherwise a crafted message might try to navigate the desktop to a different location.

we should also check that it is a valid url, we don't want to add anything random to the browserhistory

Willyfrog (changes requested):
LGTM, but I added a few requests that might be needed.

components/logged_in/index.js
hmhealey: I think it would have to be relative to prevent issues on the community cluster. At the same time, I'm wondering if subpaths might cause problems here too.

Regarding the desktop app navigating to weird places, where do these URLs come from specifically? I wouldn't be too worried if they're just sent by something we control like the desktop app or the server. Even if they could come from a web app plugin, someone writing a malicious plugin could just change `window.location` directly

koox00: The url is a permalink so we could maybe `/^\/pl\/\w+$/.test(url)`?

koox00: Note: `history.push` follows the same origin policy, so you cannot push external URLs, it would result in an error
so that part is covered from the browser.
Nevertheless I'm adding a check to ensure url is ours and a permalink one.

Willyfrog: if we check directly for it being a relative permalink, then I have no concerns :)

Willyfrog (approved):

hmhealey (approved):

esethna:
@amyblais please cherry pick to v6.0

amyblais:
@esethna Dev can use ``/cherry-pick cloud``.

PR #8518, opened 2021-08-02
ESLint overlay trigger error help message

calebroseland:
#### Summary
Eslint error when trying to use OverlayTrigger directly from react-bootstrap, which doesn't forward `Intl.Context` to overlay content.

CleanShot 2021-08-02 at 17 10 43@2x

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -65,6 +65,11 @@
     "no-only-tests/no-only-tests": ["error", {"focus": ["only", "skip"]}],
     "react/style-prop-object": [2, {
       "allow": ["Timestamp"]
+    }],
+    "no-restricted-imports": ["error", {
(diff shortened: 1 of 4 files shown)


agarciamontoro (approved):
Cool! What was the motivation behind this? By the way, should we export this component and use it in Playbooks?

hmhealey:
It's because the old version of react-bootstrap that the web app uses doesn't pass along the Context needed to use react-intl components. If we're running into a spot where Playbooks or plugins start needing it though, I'd almost say we modify our fork of react-bootstrap to pass that properly. What we're doing now is sort of a hack, and I'd rather not spread it to other code bases.

hmhealey (approved):

PR #8522, opened 2021-08-03
[MM-36432] Pass through browser history changes to the desktop app for desktop v5.0+

devinbinnie:
#### Summary
Changes to the webapp to allow it to send any changes in browser history to the desktop app, to determine if the navigation should be forwarded to another browser tab instead of being executed on the same browser tab.

Must use desktop v5.0 or greater. Currently if you build the desktop app it reports as 12.0.0 (ie. the Electron version) so that should work for testing.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36432

#### Related Pull Requests
https://github.com/mattermost/desktop/pull/1680

#### Release Note
```release-note
NONE
```


utils/browser_history.jsx
```diff
@@ -3,4 +3,41 @@
...
+    case 'browser-history-push-return': {
+        const {pathName} = message;
+        b.push(pathName);
+        break;
+    }
+    }
+});
+
+export const browserHistory = {
+    ...b,
```
hmhealey, line 27: Is the object created by `createBrowserHistory` a plain object or is it one that actually has a class and methods? I think this might be fine given how weird classes are in JS, but I'm trying to think if we should be extending a class or something here.

devinbinnie: Looks like it's just a POJO: https://github.com/ReactTraining/history/blob/28c89f4091ae9e1b0001341ea60c629674e83627/packages/history/index.ts#L559

I think either way we're good to just copy over the contents and override the methods we want to override, normally I'd be in favour of inheritance but it doesn't look like that's possible here.

hmhealey (approved):

crspeller (approved):

stafot:
FYI: this test server created from this PR seems failing.

jwilander:
@stafot there's some sort of issue with it when plugins get uploaded. Thread here: https://community-daily.mattermost.com/core/pl/gq1a3f373fnffbcpgqi393bpyr

stafot:
@jwilander Thanks for the clarification!!

stylianosrigas:
@jwilander @stafot I will remove the spinwick for now to stop the alerting. Please feel free to add it again when needed.

devinbinnie:
Just a note, the test server for this PR is located here: http://devin-test-desktop.dev.spinmint.com:8065

@amyblais @lindalumitchell

jgilliam17 (approved):
Thanks @devinbinnie 
Tested, looks good to merge.
- Verified switching products in the desktop app is working as expected.
2 issues with Boards are filed separately:
https://mattermost.atlassian.net/browse/MM-38259
https://mattermost.atlassian.net/browse/MM-38279

PR #8524, opened 2021-08-04
Cypress/E2E: Add flag disabling limit to Docker resource when launching chrome, add "@timeout_error" metadata and only remove plugin by ID

saturninoabril:
#### Summary
This PR attempts to isolate and debug further the timeout error happening on the following specs:
1. recently_used_emoji_spec.js
2. plugin_install_spec.js
3. plugin_startup_fail_spec.js

- The `--disable-dev-shm-usage` for launching Chrome removes the limitation to Docker resource where Cypress is running. Such might solve the issue of hang up if the timeout error is due to some CPU/memory limitation.
- `@timeout_error` test metadata will be used to isolate the said plugins for further investigation in CI.
- The other one related to plugin is to only remove a specific plugin as much as possible.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/e2e/cypress/integration/emoji/recently_used_emoji_spec.js
+++ b/e2e/cypress/integration/emoji/recently_used_emoji_spec.js
@@ -7,7 +7,7 @@
 // - Use element ID when selecting an element. Create one if none.
 // ***************************************************************
 
-// Group: @emoji
+// Group: @emoji @timeout_error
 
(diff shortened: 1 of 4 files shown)


josephbaylon (approved):
LGTM. Tests passed locally.

furqanmlk (approved):
Thanks @saturninoabril 
image

image

image

PR #8537, opened 2021-08-09
[MM-37452] Add query parameter to translate in-product help pages

ivanaairenee:
#### Summary
This PR adds the ability to derive locale from query parameter if user is not logged in (it will derive from user's locale if they're logged in).
The original intention of the issue was only for in-product help pages, but what this PR implements will also affect the pages user can open when not logged in (register page, login page, etc), please advise if this causes any concerns.

Also done in this PR, not related to issue:
- convert `selectors/i18n.js` to `selectors/i18n.ts`, added some type checks

#### Ticket Link
[JIRA Ticket link](https://mattermost.atlassian.net/browse/MM-37452)

Fixes https://github.com/mattermost/mattermost-server/issues/18050

#### Related Pull Requests
N/A

#### Screenshots

Example on help page:
image

Example on login page:
image

#### Release note
```release-note
Translated in-product help pages when opened from desktop app
```


selectors/i18n.test.js
```diff
@@ -25,7 +25,7 @@ describe('selectors/i18n', () => {
             expect(getCurrentLocale(state)).toEqual('fr');
         });
 
-        test('not logged in', () => {
+        test('logged in', () => {
```
ivanaairenee, line 28: CMIIW, I assume this was an unintended typo?

hmhealey: Yep, that looks like an incorrect copy-paste on my part. Good catch

selectors/i18n.test.js
```diff
@@ -68,6 +68,92 @@ describe('selectors/i18n', () => {
...
+                            profiles: {},
+                        },
+                    },
+                };
+
+                setWindowLocaleQueryParameter('ko');
+
+                expect(getCurrentLocale(state)).toEqual('ko');
+
+                resetWindowLocationSearch();
```
hmhealey, line 106: Since you call this in the `afterEach` above, I think we can skip it here so we're not repeating ourselves

hmhealey (changes requested):
Thanks for the PR, @ivanaairenee! This looks great. Thanks for migrating the selector file to TypeScript as well since we've slowly but surely been doing that across the whole codebase.

Could you also add that new query parameter to the Help link below the main textbox in the app? I believe the component for that is called TextboxLinks.

hmhealey:
Regarding the other pages in the app being translated by this, I think that's totally fine. It might actually be helpful for some users even if we're not advertising it as a feature.

selectors/i18n.test.js
ivanaairenee: Forgot to clean this up, great catch! 😊

ivanaairenee:
Hi team,

In the latest commit I made I've made these changes:
- Fixes from code reviews (remove redundant `resetWindowLocationSearch();` from tests)
- Adding locale query param in `TextboxLinks`, in the way, I've also refactored this component to functional component so that we utilize hooks. I hope I'm getting this right that the codebase is also slowly migrating to functional components 🙂
- I also notice that we might want to also preserve the query param locale when navigating through the help pages. I've made some refactor and changes to reduce repetition on rendering the help page navigation as well as well as persisting locale when it exists in the query param.

I know that this PR is becoming bigger than it was intended to be, do let me know if some of the refactors can be done on a separate PR instead.

components/textbox/textbox_links/textbox_links.tsx
```diff
@@ -2,141 +2,145 @@
...
-    };
+const TextboxLinks: React.FC<Props> = ({
+    showPreview,
+    characterLimit,
+    previewMessageLink,
+    updatePreview,
+    message = '',
+    isMarkdownPreviewEnabled,
+}: Props) => {
+    const currentLocale: string = useSelector(getCurrentLocale);
```
hmhealey, line 28: Instead of using `useSelector`, could you pass in the current locale from the `mapStateToProps` in the index file instead so that we're not mixing ways of accessing the store for this component?

hmhealey (changes requested):
Thanks for making those changes. There's one last thing I'd like changed, but overall, this looks great.

We appreciate the extra refactoring and cleanup too. I'd only worry about splitting it out into a separate PR if it wasn't related to the changes you're making, but I think it's fine to include here. We definitely appreciate the help getting some of these long-neglected components cleaned up too.

components/textbox/textbox_links/textbox_links.tsx
ivanaairenee: Sure thing, addressed! 😊

nevyangelova (approved):
Thank you, this is much more readable now :)

hmhealey (approved):

hmhealey:
FYI, I added a release note to the PR description to stop that bot from complaining

jgilliam17 (approved):
Thanks @ivanaairenee 
- Tested, looks good to merge.
- Verified Help pages, when opening from the desktop app, open in the browser and render in the currently selected app language.

PR #8552, opened 2021-08-10
MM-36977 Prevent unnecessary changes to emoji state

hmhealey:
When using the emoji autocomplete, we'll end up frequently either receiving autocomplete results for emojis we've already received and sometimes we'll receive empty autocomplete results. Both of these caused `state.entities.emojis.emojis` to be modified even though they wouldn't have any actual changes on the app, so a bunch of components (mostly instances of the `Markdown` component) would re-render a lot more than they should. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36977

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8480

#### Release Note
```release-note
NONE
```

### DIFF
--- a/packages/mattermost-redux/src/reducers/entities/emojis.ts
+++ b/packages/mattermost-redux/src/reducers/entities/emojis.ts
@@ -1,5 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+
 import {combineReducers} from 'redux';
 
 import {EmojiTypes, PostTypes, UserTypes} from 'mattermost-redux/action_types';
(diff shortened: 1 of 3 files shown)


packages/mattermost-redux/src/reducers/entities/emojis.ts
```diff
@@ -11,22 +12,27 @@ import {IDMappedObjects} from 'mattermost-redux/types/utilities';
 export function customEmoji(state: IDMappedObjects<CustomEmoji> = {}, action: GenericAction): IDMappedObjects<CustomEmoji> {
     switch (action.type) {
     case EmojiTypes.RECEIVED_CUSTOM_EMOJI: {
-        const nextState = {...state};
-        nextState[action.data.id] = action.data;
-        return nextState;
+        const emoji: CustomEmoji = action.data;
```
michelengelen, line 15: why the extra variable init and no immediate return?

packages/mattermost-redux/src/reducers/entities/emojis.ts
```diff
@@ -11,22 +12,27 @@ import {IDMappedObjects} from 'mattermost-redux/types/utilities';
...
+
+        return storeEmoji(state, emoji);
     }
     case EmojiTypes.RECEIVED_CUSTOM_EMOJIS: {
-        const nextState = {...state};
-        for (const emoji of action.data) {
-            nextState[emoji.id] = emoji;
-        }
-        return nextState;
+        const emojis: CustomEmoji[] = action.data;
```
michelengelen, line 20: same question here

michelengelen (commented):
just one question (in two places) ... otherwise LGTM.

AGMETEOR (approved):

packages/mattermost-redux/src/reducers/entities/emojis.ts
hmhealey: I started doing that recently so that we get some clearer typing. Since our reducers aren't very TS-friendly, I'm using it like inline documentation about what data the action includes.

packages/mattermost-redux/src/reducers/entities/emojis.ts
hmhealey: In this case, I think it's actually necessary so that TypeScript knows what the return value of `reduce` is. I might be confusing that with [this case below](https://github.com/mattermost/mattermost-webapp/pull/8552/files/8adc709f83665bf85b3c06bda3bdcbe3ae542a24#diff-205e6890da9eee5d067f7ded407255abd6fef79132519cdfbf2572733708dba4L42) that I also changed.

packages/mattermost-redux/src/reducers/entities/emojis.ts
michelengelen: Fair enough! 👍🏻

michelengelen (approved):

PR #8553, opened 2021-08-10
Mm 37049 cta cloud

neallred:
#### Summary
Change feature discovery so that cloud users are prompted to "Subscribe now" and see the upgrade modal.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37049

#### Related Pull Requests
None

#### Screenshots
In the screenshot, I didn't locally have all the CWS settings set, so the upgrade options aren't right in the modal. But they should be right in test environments.

https://user-images.githubusercontent.com/13738432/128938819-5b01d3b0-ba3e-4ffc-aaca-45331dbeff46.mp4

#### Release Note

```release-note
Launch the cloud upgrade modal in feature discovery if subscription is cloud.
```


AGMETEOR:
Could the test server creation be failing due to some failing tests?

AGMETEOR:
"Note locally I didn't have all the CWS settings set, so the upgrade options aren't right in the modal. But they should be right in test environments."

@neallred to mock the CWS calls you can set `ENABLE_MOCK_CWS=true` when starting the server and enterprise repos. If there're particular calls not yet mocked out, we can add them to the mock code.

AGMETEOR (commented):
Great work on adding the changes. All the code looks good. Just a small thing, with these changes, this line here https://github.com/mattermost/mattermost-webapp/blob/e63a0ac2c5ea59b284854b97a2af67cca914fee0/components/admin_console/feature_discovery/feature_discovery.tsx#L114 would not be coherent if we have cloud license. @marianunez I think we would need to change the wording in that disclaimer.

packages/mattermost-redux/src/selectors/entities/general.ts
```diff
@@ -21,7 +21,7 @@ export function getFeatureFlagValue(state: GlobalState, key: keyof FeatureFlags)
     return getConfig(state)?.[`FeatureFlag${key}` as keyof Partial<ClientConfig>];
 }
 
-export function getLicense(state: GlobalState): any {
```
marianunez, line 24: Wondering if this change may affect any other calling code for this

marianunez (approved):
LGTM. Could we add a test for this new case for cloud in feature discovery here: https://github.com/mattermost/mattermost-webapp/blob/cert-auth-check/e2e/cypress/integration/system_console/feature_discovery_spec.js

packages/mattermost-redux/src/selectors/entities/general.ts
neallred: If everything still type checks, no. It won't change any run time code; any breakages seen would be at compile time.

marianunez: Sound good, thanks for clarifying!

AGMETEOR:
> Great work on adding the changes. All the code looks good. Just a small thing, with these changes, this line here
> 
> https://github.com/mattermost/mattermost-webapp/blob/e63a0ac2c5ea59b284854b97a2af67cca914fee0/components/admin_console/feature_discovery/feature_discovery.tsx#L114
> 
> would not be coherent if we have cloud license. @marianunez I think we would need to change the wording in that disclaimer.
Hello @marianunez , how about this?

neallred:
Ok @stevemudie , it took a while, but this one should be testable now! Steps in https://mattermost.atlassian.net/browse/MM-37049

marianunez:
> Great work on adding the changes. All the code looks good. Just a small thing, with these changes, this line here
> 
> https://github.com/mattermost/mattermost-webapp/blob/e63a0ac2c5ea59b284854b97a2af67cca914fee0/components/admin_console/feature_discovery/feature_discovery.tsx#L114
> 
> would not be coherent if we have cloud license. @marianunez I think we would need to change the wording in that disclaimer.

Missed your message before @AGMETEOR. Agreed this text here does not apply to cloud upgrade. @nathanaelhoun can we just hide that text under the buttons for cloud? Any other disclaimers or terms are shown in the actual upgrade modal so I don't see the need to add anything there

neallred:
Looks like there's at least 2 Nathaniels contributing to mattermost :grin:

marianunez:
> Looks like there's at least 2 Nathan{a,i}els contributing to Mattermost 😁

Oops apologies on incorrect ping here 🙈

stevemudie (approved):
Tested and passed!

- Added a cloud-starter license and observed "Subscribe Now" wording on unavailable features in the System Console
- Removed the license and text reverted to "Start Trial" as expected in all locations

pablovelezvidal (approved):
Nice work @neallred :shipit:

PR #8561, opened 2021-08-11
MM-37600 - disable global header in mobile web views (small viewports)

michelengelen:
#### Summary
disables global header on the mobile web-view

#### Ticket Link
[MM-37600](https://mattermost.atlassian.net/browse/MM-37600)

#### Related Pull Requests
~~CAUTION: #8509 needs to be merged before this PR, since it branches of of it.~~
#8509

#### Release Note
```release-note
disables global header on mobile web-view
```


nevyangelova:
@michelengelen can we change the base branch to MM-37356 so its easier to review?

michelengelen:
> @michelengelen can we change the base branch to MM-37356 so its easier to review? 

I will try that

nevyangelova (approved):

components/user_settings/display/index.ts
```diff
@@ -36,6 +37,9 @@ function mapStateToProps(state: GlobalState) {
     const lockTeammateNameDisplay = getLicense(state).LockTeammateNameDisplay === 'true' && config.LockTeammateNameDisplay === 'true';
     const configTeammateNameDisplay = config.TeammateNameDisplay as string;
 
+    // calling it once here avoids multiple calls to the function in the return
```
crspeller, line 40: Nit: This is clear. Why the comment?

crspeller (changes requested):
Doesn't this just hide the setting and not the actual bar on mobile? Is that the intent?

components/user_settings/display/index.ts
michelengelen: It's just a habit of mine to document stuff inline

michelengelen:
> Doesn't this just hide the setting and not the actual bar on mobile? Is that the intent?

It does both actually

crspeller (changes requested):
Ok had to figure this out since this PR clearly doesn't do what it says it does. 
Turns out, the PR it's based on does it:
https://github.com/mattermost/mattermost-webapp/pull/8509/files#diff-21d89f247b2f5c290a0c392efe0b59fe8435b9acbd9d49b149297aee23e0a64cR10
It doesn't actually disable the bar it just hides it using CSS. That's fine for mobile view. 

This PR actually removes the ability to edit the setting when you are in mobile view. Not sure that is intended. @zefhemel can you clarify the intent of the ticket?

michelengelen:
> Ok had to figure this out since this PR clearly doesn't do what it says it does. 

I partly disagree. It does exactly what it says it does, but additionally it does hide the option in the settings.

> Turns out, the PR it's based on does it:
> 
> https://github.com/mattermost/mattermost-webapp/pull/8509/files#diff-21d89f247b2f5c290a0c392efe0b59fe8435b9acbd9d49b149297aee23e0a64cR10
> 
> It doesn't actually disable the bar it just hides it using CSS. That's fine for mobile view. 

That's not the intent of this PR. As stated in the ticket it should keep the UI to a pre-v6.0 state. So all elements should be in the place as in v5.38.

> This PR actually removes the ability to edit the setting when you are in mobile view. Not sure that is intended. @zefhemel can you clarify the intent of the ticket? 

Having the ability to change the setting would be possible, but a setting that does not change anything visually once you set it does only create confusion, IMO.

zefhemel:
Whether this should also hide the configuration option: honestly I don't think it's really important because will remove this option very soon anyway: https://mattermost.atlassian.net/browse/MM-37821

For now, hiding it on mobile seems sensible.

michelengelen:
gentle ping to re-review @crspeller and @hmhealey

michelengelen:
> Thanks for fixing it so it also does the selector! Not sure if the CSS way should be removed? But not going to block on that.

I have just now found the CSS rules. IMO we can keep it for now, but should remove once we have everything in place and released v6.0.

components/user_settings/display/index.ts
```diff
@@ -36,6 +37,9 @@ function mapStateToProps(state: GlobalState) {
...
+    const isMobile = UserAgent.isMobile();
```
hmhealey, line 41: So this only applies to mobile browsers, but it won't apply to mobile web view on the desktop (which is checked by `Utils.isMobile`. Is that intended?

components/user_settings/display/index.ts
```diff
@@ -52,13 +56,13 @@ function mapStateToProps(state: GlobalState) {
...
-        globalHeaderDisplay: get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.GLOBAL_HEADER_DISPLAY, Preferences.GLOBAL_HEADER_DISPLAY_OFF),
+        globalHeaderDisplay: !isMobile && get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.GLOBAL_HEADER_DISPLAY, Preferences.GLOBAL_HEADER_DISPLAY_OFF),
         messageDisplay: get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.MESSAGE_DISPLAY, Preferences.MESSAGE_DISPLAY_DEFAULT),
         collapseDisplay: get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.COLLAPSE_DISPLAY, Preferences.COLLAPSE_DISPLAY_DEFAULT),
         collapsedReplyThreadsAllowUserPreference: isCollapsedThreadsAllowed(state) && getConfig(state).CollapsedThreads as string !== 'always_on',
         collapsedReplyThreads: getCollapsedThreadsPreference(state),
         linkPreviewDisplay: get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.LINK_PREVIEW_DISPLAY, Preferences.LINK_PREVIEW_DISPLAY_DEFAULT),
-        globalHeaderAllowed: getFeatureFlagValue(state, 'GlobalHeader') === 'true',
+        globalHeaderAllowed: !isMobile && getFeatureFlagValue(state, 'GlobalHeader') === 'true',
```
hmhealey, line 65: Is it necessary to hide the setting in mobile web view, or is it enough to just prevent it from having an effect? 0/5 on that.

michelengelen: Since we are going to remove the setting pretty soon (probably tomorrow) we can leave it like that for now.

components/user_settings/display/index.ts
michelengelen: No, it is not intended. Interestingly enough it works on desktop browsers with a mobile like window size as well. That's the feature I was aiming at providing

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
- Verified global header is hidden in mobile webview; all items are available in this view as before the upgrade to v6.0.

components/user_settings/display/index.ts
michelengelen: FYI: @hmhealey I created a new UserAgent function `isMobileWeb` that checks for **NOT** being IOS or Android (`!isMobile()`) and a max screen width of 768px (defined in the constants file)

michelengelen:
@matthewbirtch and @jgilliam17 you might need to re-review here, since the underlying mechanic changed after I added the new checks as mentioned in the thread with @hmhealey ... 🙇🏼‍♂️

utils/user_agent.tsx
```diff
@@ -103,6 +105,11 @@ export function isMobile(): boolean {
     return isIos() || isAndroid();
 }
 
+// returns true if and only if the user is using Mattermost from a PC with a small viewport size of <= 768px
```
crspeller, line 108: Is this really the intent of the ticket? We want global header to display on desktop and mobile web if you are on a phone, but NOT on small viewports on desktop?

It seems like if we want to hide on mobile we should hide in all cases. (What was happening before with the CSS)

crspeller (changes requested):

utils/user_agent.tsx
michelengelen: oh shoot ... you are absolutely right. I somehow mixed that up. I will change that.

michelengelen:
@crspeller I adjusted the checks, so now it should work as intended: Do not show the global header and keep the app in a pre-v6.0 state, when

1. the viewport is smaller or equal 768px, or ... 
2. it is viewed from a mobile device 

Could you re-review please?

components/user_settings/display/index.ts
michelengelen: I did change the function to check for a small viewport size only and renamed it to `isMobileViewportSize`

selectors/global_header.ts
```diff
@@ -5,12 +5,14 @@ import Constants from 'utils/constants';
...
 
 const Preferences = Constants.Preferences;
 
 export function getGlobalHeaderEnabled(state: GlobalState): boolean {
     const featureFlagEnabled = getFeatureFlagValue(state, 'GlobalHeader') === 'true';
     const userPreferenceEnabled = get(state, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.GLOBAL_HEADER_DISPLAY, Preferences.GLOBAL_HEADER_DISPLAY_DEFAULT) === Preferences.GLOBAL_HEADER_DISPLAY_ON;
+    const isMobileView = isMobileViewPortSize() || isMobile();
 
-    return featureFlagEnabled && userPreferenceEnabled;
+    return !isMobileView && featureFlagEnabled && userPreferenceEnabled;
```
crspeller, line 17: In a vacuum this won't work in the case of a viewport size change, since I don't think this will trigger a re-render on its own. But there is probably enough going on and re-rendering taking place that this will work anyway. Not sure we even really care about the viewport size change case.

crspeller (approved):

selectors/global_header.ts
michelengelen: This is a temporary change anyways and it is working for now. I agree that this is not the best option, but just hiding the GH is not gonna work, since the elements were re-arranged quite a bit and this prevents that from showing.

matthewbirtch (approved):
Looks good to me @michelengelen. The only thing I see is that the focalboard icon is not themed correctly, but I assume that's not in scope for this PR?

image

michelengelen:
> Looks good to me @michelengelen. The only thing I see is that the focalboard icon is not themed correctly, but I assume that's not in scope for this PR?

Correct.

selectors/global_header.ts
hmhealey: One of these days, we should move whether or not the app is in mobile view into redux so that we can easily access them in places like this. We did something like that in the mobile app ages ago, and it made it a lot easier to reference the screen size.

michelengelen: @hmhealey agreed to 100%!

michelengelen:
@jgilliam17 kind reminder to re-review so we can merge this and have it in v6. @hmhealey could you please merge once all reviews are done?

utils/user_agent.tsx
```diff
@@ -103,6 +105,11 @@ export function isMobile(): boolean {
     return isIos() || isAndroid();
 }
 
+// returns true if and only if the user is using Mattermost within a viewport size of <= 768px
+export function isMobileViewPortSize(): boolean {
```
hmhealey, line 109: Since this just checks the screen size again, can we switch back to using `Utils.isMobile` elsewhere? That's assuming I'm not missing some major difference between `innerWidth` and `visualViewport.width` though

hmhealey (changes requested):
Could you also give this PR a more descriptive title since that gets used as the commit message once this is merged?

utils/user_agent.tsx
michelengelen: what do you mean by elsewhere? I did change every occurance in my changes to this check, or am I missing something?

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
- Verified global header is not displayed in mobile web views.

utils/user_agent.tsx
hmhealey: I mean that this function can be removed since we already have `Utils.isMobile` that's used elsewhere.

michelengelen: Not really ... Utils.isMobile will return  false when used in a desktop browser. The new function will check for the size of the viewport regardless of platform (except RN probably)

hmhealey: I think you might be confusing the two `isMobile` functions. `isMobile` in utils.jsx returns true if we're showing mobile view, regardless of the platform, so it can return true on a desktop browser.

It does the same thing as the `isMobileViewPortSize` that you added, so I want to use the existing function in the places that you used `isMobileViewPortSize` instead.

michelengelen: Yes, I did confuse the two functions, you are absolutely right. I did not know of the existence of a separate `isMobile` check in the `utils.jsx` file, so thanks for pointing that out. I will change that to the correct function now. 🙇🏼‍♂️

hmhealey (approved):

PR #8570, opened 2021-08-12
Close products menu on switch

jespino:
#### Summary
Close the products menu when you click on any option (what is expected to switch to that option)

#### Ticket Link
mattermost/focalboard#943

#### Release Note
```release-note
NONE
```


nevyangelova (approved):

components/global/product_switcher.tsx
```diff
@@ -125,7 +125,10 @@ const ProductSwitcher = (): JSX.Element => {
     });
 
     return (
-        <div ref={menuRef}>
+        <div
+            ref={menuRef}
+            onClick={handleClick}
```
michelengelen, line 130: this will also close the menu when you click on any portion of it that is not a clickable menu item.
For example the "Open ..." header, or any other element, even disabled menu items.

The `handleClick` should go into every menu item instead

michelengelen (changes requested):

components/global/product_switcher.tsx
jespino: This is not that simple from some of the entries, for example the Menu.LinkItem was design to relay on the "auto-close" behavior of the current menu system, but you are overwriting that and that generates the problem of not-auto-close behavior, and not onClick handler there. I going to fix it for boards, but for the rest of the links is going to be a bit more complicated.

michelengelen: we can "hack" this by wrapping the menu items in a separate container and handle the clicks on that (But please add an inline TODO so we can remove this later on, when the CC `MenuItem` component is ready). Would that be sufficient?

ogi-m (approved):
Cloud test server creation is failing for some reason, will test post-merge

PR #8575, opened 2021-08-12
only show "integration" and "marketplace" in messaging

michelengelen:
#### Summary
hides the `ProductSwitcherMenu` items "integrations" and "marketplace" in non-messaging products

#### Ticket Link
[MM-37363](https://mattermost.atlassian.net/browse/MM-37363)

```release-note
NONE
```


matthewbirtch:
@michelengelen I can't review this one either since the test server seemed to fail

matthewbirtch:
@michelengelen I don't see focalboard or playbooks in the menu so I can't quite test this to verify the menu items display as expected
image

components/global/product_switcher_menu/product_switcher_menu.tsx
```diff
@@ -81,35 +83,39 @@ class ProductSwitcherMenu extends React.PureComponent<Props> {
...
-                        <Menu.ItemToggleModalRedux
-                            id='marketplaceModal'
-                            modalId={ModalIdentifiers.PLUGIN_MARKETPLACE}
-                            show={!this.props.isMobile && this.props.enablePluginMarketplace}
-                            dialogType={MarketplaceModal}
-                            text={formatMessage({id: 'navbar_dropdown.marketplace', defaultMessage: 'Marketplace'})}
+                    {isMessaging && (
+                        <Menu.ItemLink
+                            id='integrations'
+                            show={showIntegrations}
```
deanwhillier, line 89: Looks like the existing menu item component has a `show` prop that is currently used to indicate whether the menu item should be rendered or not. Thoughts on using the prop instead for both menu items?

```suggestion
                        <Menu.ItemLink
                            id='integrations'
                            show={isMessaging && showIntegrations}
```

deanwhillier (commented):
LGTM with one question/thought.

components/global/product_switcher_menu/product_switcher_menu.tsx
michelengelen: I am typically against the usage of a prop for this, since it will call the function regardless. With the approach I took it won't do that. It is just a small performance improvement.

michelengelen: @deanwhillier I just took a look at the implementation of Menu.ItemLink and IMO it massively over-engineered (several nested functions and classes). All the more reason not even calling.

michelengelen:
> @michelengelen I don't see focalboard or playbooks in the menu so I can't quite test this to verify the menu items display as expected
> image

I am not sure why there is no Playbooks or Boards implemented in the test-server. @jespino do you possibly know why?

components/global/product_switcher_menu/product_switcher_menu.tsx
jespino: Actually, MenuItemLink is build with a HOC at load time, so it is not generating that many calls, actually if show is false, it is not rendering anything, or calling any other internal function, so should be pretty efficient.

Said that, you are right, the menu system has some over-engineering, the plan was to reduce the code repetition of each item using HOC that defines what a menu item is, and nesting the items inside the `Menu` component to give a "namespace" to use them.

jespino:
> I am not sure why there is no Playbooks or Boards implemented in the test-server. @jespino do you possibly know why?
It should be there, but is not that easy to incorporate them because the stability of the main branches probably are not at the mattermost-server/mattermost-webapp standards, and at the end of the day they are plugins, that you can download and install.

My POV is, they should be there ASAP, but we are rushing in other points and probably this is not a priority because we have the workaround.

components/global/product_switcher_menu/product_switcher_menu.tsx
michelengelen: @jespino as far as I can see it, the `MenuItemLink` is being called from `Menu` (namespace). The actual `MenuItemLink` consists of a function that takes in the react-router `Link` component and passes it into a function, that creates a new class. Inside that classes render function is the condition to return `null` if `show` is false. All of this can be prevented by just not calling it! :D

michelengelen:
> > I am not sure why there is no Playbooks or Boards implemented in the test-server. @jespino do you possibly know why?
> 
> It should be there, but is not that easy to incorporate them because the stability of the main branches probably are not at the mattermost-server/mattermost-webapp standards, and at the end of the day they are plugins, that you can download and install.
> 
> My POV is, they should be there ASAP, but we are rushing in other points and probably this is not a priority because we have the workaround.

OK, thanks for the clarification. So do we have some kind of workaround so that @matthewbirtch and @jgilliam17 can test this?

jespino:
Yes, the work around is manually install that in the instance. They need to get the playbooks plugin and the boards plugin and install it as a regular plugin in the test server.

You can download the last RC of boards here: https://github.com/mattermost/focalboard/releases
and the last version from playbooks here: https://github.com/mattermost/mattermost-plugin-playbooks/releases

nevyangelova (approved):

components/global/product_switcher_menu/product_switcher_menu.tsx
jespino: You can prevent some things not calling it, but you are assuming that `MenuItemLink` is calling the `menuItem` HOC funtion, that creates the class, but the reality is `MenuItemLink` is a class that is generated on load time (whenever you load the javascript) calling the `menuItem` HOC function, on top of the `MenuItemLinkImpl`. That means, as soon as you loaded the javascript `MenuItemLink` is an instance of a class that looks like this:

```ts
      class MenuItem extends React.PureComponent<Props & React.ComponentProps<typeof Component>> {
          public static defaultProps = {
              show: true,
          };
   
          public static displayName?: string;
   
          public render() {
              const {id, show, icon, text, to} = this.props;
              if (!show) {
                  return null;
              }
   
              let textProp: React.ReactNode = text;
              if (icon) {
                  textProp = (
                      <React.Fragment>
                          <span className='icon'>{icon}</span>
                          {text}                           
                      </React.Fragment>
                  );
              }                              
   
              return (
                  <li
                      className={classNames('MenuItem', {
                          'MenuItem--with-icon': icon,
                      })}
                      role='menuitem'
                      id={id}
                  >
                      <Link to={to}><span className='MenuItem__primary-text'>{text}</span></Link>
                  </li>
              );
        }
    }
```

So all that calls that you are talking about, are just reduced to the `render` function (and the lifecycle functions of the component).

You can read more about HOC here: https://reactjs.org/docs/higher-order-components.html

michelengelen: I know what a HOC is and I know how to improve performance of React applications (done that several times) and regardless of when the class instance gets called React will go and try to mount the component (even if it eventually is `null`).

That is what I am trying to prevent. I know perfectly well, that this might be just a tiny improvement, but I have seen several apps in the past that were pretty inefficient just because of complexity and small things like these. "Death by a thousand cuts".

If you don't like the approach: fine. I can change that back.

deanwhillier (approved):

matthewbirtch:
@michelengelen how can we push this PR forward? Also, I'm going to add @andrewbrown00 as a reviewer here while I'm out.

michelengelen:
> @michelengelen how can we push this PR forward? Also, I'm going to add @andrewbrown00 as a reviewer here while I'm out.

Afaik the spinwicks should work again. I am going to test it now

andrewbrown00:
@michelengelen let me know when the test server is working, thanks

michelengelen:
@spirosoik isnt the issue with the spinwicks resolved? I cannot get the test servers to work atm. Is there something we can do to get them up and running? Merge master maybe?

spirosoik:
@michelengelen you need to rebase with latest master otherwise your docker image which has been built already pings to the server commit with the bug

michelengelen:
thanks @spirosoik .... i will try that!

andrewbrown00:
Thanks @michelengelen, test server is working 🎉 

Are you able to get **Playbooks** and **Boards** to show up? See https://github.com/mattermost/mattermost-webapp/pull/8575#issuecomment-898266153 for context.

michelengelen:
> Are you able to get **Playbooks** and **Boards** to show up? See https://github.com/mattermost/mattermost-webapp/pull/8575#issuecomment-898266153 for context.

Unfortunately I don't ... I know that @jespino knows how to do that. Could you help here please? I am already OoO.

If it is not done by tomorrow I will have a look again.

jgilliam17:
I was trying to manually upload plugins on the test server, but this appears to be disabled.
Screen Shot 2021-08-16 at 8 06 21 PM

michelengelen:
@jgilliam17 the focalboard plugin seems to be running, but i cannot see it in the product-switcher as well. Maybe @jespino can be of help here again?

Screenshot 2021-08-17 at 09 51 48

jespino:
@michelengelen you are using the focalboard 0.8.2 version, that is and older one, you have to use last 0.9.0 RC. that one is the one that supports the header. Here is the link: https://github.com/mattermost/focalboard/releases/download/v0.9.0-rc5/mattermost-plugin-focalboard.tar.gz

michelengelen:
> @michelengelen you are using the focalboard 0.8.2 version, that is and older one, you have to use last 0.9.0 RC. 

But we cannot upload other versions according to @jgilliam17. I tried as well with the sysadmin user, but the option is not available

jespino:
> > @michelengelen you are using the focalboard 0.8.2 version, that is and older one, you have to use last 0.9.0 RC.
> 
> But we cannot upload other versions according to @jgilliam17. I tried as well with the sysadmin user, but the option is not available

Yes, that is a problem, the spinwicks are not supporting upload plugins, that is correct because the information to access is public and with a plugin you can abuse the resources there. I don't know if there is a good way to test it other than creating a local instance and testing it or just testing after merge.

andrewbrown00:
@michelengelen can I test it with an `ngrok`? cc @jespino

michelengelen:
@andrewbrown00 I dont know if this is possible, but since we do not have it installed on the webapp there is a lot to add here just to allow it to work for a single PR. 

@jespino the current master has the new patch version of focalboard running, right? I will try to merge master again and see if this fixes the issue.

jespino:
@michelengelen we don't have the plugin yet in master. I can do the ngrok thing locally without a problem. Anyway, my suggestion would be, unless you feel strong about not doing it, test it after merge.

michelengelen:
> @michelengelen we don't have the plugin yet in master. I can do the ngrok thing locally without a problem. Anyway, my suggestion would be, unless you feel strong about not doing it, test it after merge.

Oh, if it  can be done locally then i have misunderstood the docs ... i have never used it before, so please bear with my innocence here! :D

andrewbrown00:
Thanks, @michelengelen, let me know what the solution is to allow me to test since I only see **Channels** in the popover.

image

michelengelen:
@andrewbrown00 We will test it after merging it.

PR #8581, opened 2021-08-13
MM-37862: Update product registration

deanwhillier:
#### Summary
Per the conversation and resulting decision [summarized here](https://community-daily.mattermost.com/core/pl/rhcg719ip7g1ufu8djwz3bcg3y), this PR updates the Product registration api to follow this type:

```javascript
export type ProductComponent = {
    id: string;
    pluginId: string;
    switcherIcon: TIconGlyph;
    switcherText: string;
    baseURL: string;
    switcherLinkURL: string;
    mainComponent: React.ReactNode;
    headerCentreComponent: React.ReactNode;
    headerRightComponent: React.ReactNode;
};
```

Specifically, it:
- updates `switcherText` to accept a `string`
- renames `headerComponent` to `headerCentreComponent`
- adds new `headerRightComponent`
- makes both `headerCentreComponent` & `headerRightComponent` optional
- renders anything passed to `headerRightComponent` in the `RightControls` container component

There is also a small styling tweak to help ensure the GH sits on top of the UI of each product.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37862

#### Release Note
```release-note
NONE
```


components/global/global_header.tsx
```diff
@@ -34,6 +34,12 @@ const GlobalHeaderContainer = styled.header`
     border-bottom: solid 1px rgba(var(--center-channel-color-rgb), 0.08);
     color: rgba(var(--global-header-text-rgb), 0.64);
     padding: 0 16px 0 12px;
+    z-index: 2;
+
+    + * {
```
jespino, line 39: Here you are altering the `position` and the `z-index` of all the plugins and the custom routes from the outside an using a specific component styles to do it. It sounds wrong to me.

deanwhillier: It only affects the single sibling element immediately following the GH when the GH is enabled.

See here: https://community-daily.mattermost.com/core/pl/9pkmuzj85trrdqkqtcqdwqm9ne

It's equivalent to changing the GH z-index to a really high number to move it on top of any other possible element with a currently higher z-index, only this is much more controlled and doesn't require sifting through the dozens of defined z-index values to try find the perfect value. 

That said, I'm perfectly fine if individual products would prefer to handle layering themselves as is already done for channels.

jespino: Still, I don't see it at Global Header level, having a component there that affects the siblings is something that looks wrong to me. For me, makes more sense to do it from the container of `GlobalHeader`, that can be a style that says `.GlobalHeader + * {`, but at the parent node, because that one has responsibility over all the implicated nodes. Not a big deal, but affecting sibling components from inside the component is a pattern that I don't like.

Because this is not an stopper in any case, I going to approve the PR.

jespino (approved):

components/global/global_header.tsx
deanwhillier: Ah yes, that is a very fair point and I absolutely agree ... I actually had the original channels-only implementation being applied by the parent, but got caught up in the weeds and forgot about that. 😛

One sec, I'll quickly modify!

crspeller (approved):

jespino (approved):

jgilliam17:
@deanwhillier 
I was able to upgrade one of my test cloud server with this PR commit and manually install boards and playbook plugins to test 
Looks good. I got 2 screenshots right before my server crashed 
Screen Shot 2021-08-13 at 1 47 11 PM
Screen Shot 2021-08-13 at 1 46 55 PM
Crash happened right as I switched to Boards, menus opened there fine, over boards 
 I saw few errors on the console... something about  "unable to get channel membership" unfortunately I failed to capture the screenshot 😞

deanwhillier:
@jgilliam17, I just installed the boards plugin alongside the playbook plugin and I'm able to switch back and forth between them without apparent issue.

There is an `ASSERT` log coming from boards that seems to have something to do with the websocket not being available, but that doesn't seem to prevent it from loading for me. @jespino, I'm assuming you're on top of that websocket assert in the console?

image

jgilliam17 (approved):
Thanks @deanwhillier
Tested partially before crash. 
QA will test after merge. Currently test servers are built without an ability to manually upload plugins.

amyblais:
@deanwhillier Is this a requirement for v6.0? If yes, can you help cherry-pick to the ``cloud`` branch? https://community.mattermost.com/core/pl/cmqefh9ue7dnxgfm1cirbcj9se

deanwhillier:
Thanks @amyblais!

PR #8585, opened 2021-08-13
[MM-37785] - Polish global header

nevyangelova:
#### Summary

This PR addressed the UI changes requested by @michaelgamble in the initial global header PR. Some notes:
1. I decided to keep the user status dropdown arrow because it seemed intuitive to click without it. Design can advice.
2. The active states of the IconButton are implemented but we need to wait for the version to be available on npm before updating. You won't be able to see the change right now but as soon as its up I will update the PR. Because of this we will get a type error for now.
3. Michael there is a discrepancy between the documentation for icons in compass and the product icons, you mentioned it should be 24px but the values we have in compass are 20, 28. Should we add this size to compass?
4. History arrows changes should be addressed in another PR.
5.  Removed the marketplace dot as per PM comments.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37785

#### Release Note
-->
```release-note
NONE
```


components/global/product_switcher.tsx
```diff
@@ -67,7 +67,8 @@ const MenuItem = styled(Link)`
 `;
 
 const StyledIcon = styled(Icon)`
-    color: var(--sidebar-bg);
+    color: var(--button-bg);
+    font-size: 24px;
```
deanwhillier, line 71: Ah yes, the icon sizes are somewhat deceiving because the size of the icon is for the container and not specifically the font-size of the icon itself, which is usually a bit bigger to compensate for the extra padding the font adds around the glyph.

In this case, the font-size for the newer 24px icon is actually 28.8px is Figma. 🤷‍♂️

We absolutely should add the new size to the icon component, but in the mean time, this should probably be more like:

```suggestion
    ::before {
        font-size: 28.8px;
        line-height: 28.8px;
    }
```

components/sidebar/legacy_sidebar_header/dropdown/sidebar_header_dropdown.tsx
```diff
@@ -3,12 +3,14 @@
 
 import React from 'react';
 
+import IconButton from '@mattermost/compass-components/components/icon-button';
+
```
deanwhillier, line 7: I believe this is a left-over remnant that now isn't being used after you merged master?

```suggestion
```

components/widgets/menu/menu_group.scss
```diff
@@ -6,6 +6,6 @@
     &.menu-divider {
         height: 1px;
         margin: 8px 0;
-        background: rgba(var(--center-channel-color-rgb), 0.16);
+        background: rgba(var(--center-channel-color-rgb), 0.08);
```
deanwhillier, line 9: Is this a permanent change whether GH is enabled or not? If it's only meant for when GH is enabled, you should probably move to `global_header.scss` until the feature flag is removed.

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -47,6 +47,8 @@ export type Props = {
     direction?: 'left' | 'right';
     openUp?: boolean;
     styleSelectableItem?: boolean;
+    renderLeft?: boolean;
+    renderRight?: boolean;
```
deanwhillier, line 51: I see that the `direction` prop was originally intended for controlling menu position. I would prefer to keep it that way to avoid confusion between the original prop and the two new props.

Why not just hard-code the arrow on the right regardless of what `direction`'s value is and use the `direction` prop just to control the placement of the menu? @matthewbirtch has already confirmed that the arrow should just stay on the right, so I think this would be appropriate.

Looks like `direction` is also controlling whether the keyboard shortcut uses 'left' or 'right' arrows to open the menu. I would update that to only use the right arrow shortcut regardless of direction as the arrow will now always be pointing right as well. Probably best to get @matthewbirtch to weigh in on that though.

The ideal for the menu placement would of course be to automatically detect if there is space to render the menu right and if not render it left, but that's not in scope at the moment. 🙂

components/custom_status/custom_status.scss
```diff
@@ -153,6 +157,9 @@
         align-items: center;
 
         .status {
+            background: var(--sidebar-header-bg);
+            color: var(--sidebar-header-text-color);
```
deanwhillier, line 161: `global_header.scss` adds GH specific variables for background a text using the appropriate existing variables as confirmed with Design:

```css
    --global-header-background: var(--sidebar-teambar-bg);
    --global-header-text: var(--sidebar-header-text-color);
    --global-header-background-rgb: var(--sidebar-teambar-bg-rgb);
    --global-header-text-rgb: var(--sidebar-header-text-color-rgb);
```

So I'm wondering if maybe we should use those variables instead to be consistent and if we should move this whole chunk of GH specific CSS to `global_header.scss` to keep all styling specifically for when the GH is enabled together?

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -145,7 +147,7 @@ export default class SubMenuItem extends React.PureComponent<Props, State> {
     }
 
     public render() {
-        const {id, postId, text, selectedValueText, subMenu, icon, filter, ariaLabel, direction, styleSelectableItem} = this.props;
+        const {id, postId, text, selectedValueText, subMenu, icon, filter, ariaLabel, direction, styleSelectableItem, renderLeft, renderRight} = this.props;
```
michelengelen, line 150: didn't @matthewbirtch say we will always render the chevron on the right side as an indicator that there is a submenu, not to show in which direction the submenu opens?

michelengelen (changes requested):

deanwhillier (changes requested):
Looking good so far @nevyangelova, good eye overall!

I have a few comments/suggestions.

components/main_menu/main_menu.jsx
```diff
@@ -175,7 +175,7 @@ class MainMenu extends React.PureComponent {
                 })}
                 extraText={formatMessage({
                     id: 'navbar_dropdown.invitePeopleExtraText',
-                    defaultMessage: 'Add or invite people to the team',
+                    defaultMessage: 'Add or invite people to team',
```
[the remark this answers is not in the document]
deanwhillier: Oops, ignore this, you've already dealt with it. :)

components/widgets/menu/menu_items/submenu_item.tsx
matthewbirtch: > @matthewbirtch has already confirmed that the arrow should just stay on the right.

This is correct

> Looks like direction is also controlling whether the keyboard shortcut uses 'left' or 'right' arrows to open the menu. I would update that to only use the right arrow shortcut regardless of direction as the arrow will now always be pointing right as well. Probably best to get @matthewbirtch to weigh in on that though.

Or shouldn't the submenu open when you focus on the menu item (without the need to use `Left` or `Right` arrow keys? Since this menu appears on hover with the mouse interaction, I would expect it would appear on focus with the keyboard interaction. Thoughts?

> The ideal for the menu placement would of course be to automatically detect if there is space to render the menu right and if not render it left, but that's not in scope at the moment.

100%. The intent is to display submenus on the right when there is available space to do so

deanwhillier: > Or shouldn't the submenu open when you focus on the menu item

Quite possibly, yes. That wouldn't be an insignificant lift though and I don't think we have time to implement and properly QA with this round of the GH. The menu system is going to be rebuilt at some point anyway, you okay to consider this later with that effort? Or maybe create a ticket to deal with after this initial GH push?

matthewbirtch: > The menu system is going to be rebuilt at some point anyway, you okay to consider this later with that effort? Or maybe create a ticket to deal with after this initial GH push?

Yep, I'm totally fine with that.

deanwhillier: 👍

components/widgets/menu/menu_items/submenu_item.tsx
matthewbirtch: Yes, I did :)

nevyangelova: yup this change should render it to the right... or you mean remove the option to render to the left? I thought we can keep it just in case but I can also remove it :)

components/widgets/menu/menu_group.scss
nevyangelova: I was actually wondering about this too. Since I worked on another ticket in the past which was about reducing the opacity on all dividers. @michaelgamble is this to be applied in all instances for menu dividers?

michaelgamble (changes requested):
1. Status icon triangle needs to go - instead have the mouse icon switch to the finger when hovering over the whole hit state (that will fix the obviousness problem)
2. No problem
3. It seems the size token is missing for the component
image

please reference this doc: https://zeroheight.com/29be2c109/p/19c648-iconography or here in figma: https://www.figma.com/file/wMJsYg80f6GVjYi3FHuoBj/Foundations-Iconography?node-id=1352%3A1234

4. cool
5. looks good



Additional Items:
- thier seems to be a bug with really long usernames in the menu.. we need to truncate with elipses... 
image

components/global/global_header.scss
```diff
@@ -73,4 +73,12 @@
             padding: 0 6px;
         }
     }
+
+    .MenuItem {
+        height: 36px;
```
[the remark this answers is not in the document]
nevyangelova: @matthewbirtch what do you think? I was kinda thinking that we can unify the menu item heights across the webapp.

components/global/global_header.scss
```diff
@@ -64,7 +64,7 @@
         .channel__wrap {
             &.move--left {
                 .channel-header {
-                    width: calc(100% - 400px);
+                    width: calc(100% - 500px);
```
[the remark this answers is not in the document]
nevyangelova: OK i will remove this change and let this be handled with the PR for that ticket :)

matthewbirtch:
Adding @andrewbrown00 as a reviewer here as both Mike and I are out this week.

andrewbrown00 (commented):
Things are looking good @nevyangelova, a few items to resolve:

- [x] The vertical alignment of the elements is off, see the screenshot below for spacing references
   - There are slight differences to account for the icons container sizes
   - [Here's a link to the design file for reference](https://www.figma.com/file/iR1U1mW4y1U8S7qXWT9O1l/MM-35325-Global-Header-MVP?node-id=1%3A444031)
   - image
- [x] The 'checkmark' should use the `Button BG` variable in the default state
   - image
- [x] The caret is missing from the status on hover
   - image
- [ ] The user popover has a few issues:
   - The space between the icon button and the popover should be 4px
   - The border around the status icon doesn't match the selected background for the icon button
   - The popover opens too far to the right
   - The DND chevron (submenu indicator) is too far to the right
   - [Here's a link to the design file for reference](https://www.figma.com/file/iR1U1mW4y1U8S7qXWT9O1l/MM-35325-Global-Header-MVP?node-id=3%3A19994)
   - image

esethna:
Also re the profile area, I think it's missing the hand cursor when hovering in this entire area? cc// @andrewbrown00 @nevyangelova 

image

esethna:
A few more:

4. Focalboard button is incorrectly themed in mobile view

image

5. "x" icon on custom statuses looks misaligned horizontally and vertically. 0/5 maybe same with the custom status text and subtext needs polish? cc// @andrewbrown00  

image

6. The custom status emoji should have a tooltip about the status when hovering over it, same tooltip as when hovering over it in the channel. We probably want the emoji next to the profile clickable to open the custom status modal, same it used to do before these changes
image

components/global/product_switcher.tsx
nevyangelova: I will open a PR in compass components to add the icon size as Michael said this is a part of the documentation so we can use it directly.

nevyangelova:
@andrewbrown00 resolved all comments minus the one about the chevron icon missing since @michaelgamble suggested we remove it.

@esethna thank you for the additional feedback. I resolved your points minus the one for the "x" icon on custom statuses because I would like to discuss a design solution. And the focalboard button theming issue might be out of scope for this PR

esethna (approved):
Approving to not block this PR, but still seeing some issues with

1. "Files" and "Messages" tags in the search box are totally hidden due to color issues and misaligned (as seen when the box is inactive):
image

image


2. search text doesn't have enough contrast

image

3. profile dropdown is missing the arrow when on hover
image

4. custom status menu item polish (added notes to the ticket): https://mattermost.atlassian.net/browse/MM-37976

5. Clear button theming when the search box is not active:
image

nevyangelova:
> Approving to not block this PR, but still seeing some issues with
> 
> 1. "Files" and "Messages" tags in the search box are totally hidden due to color issues and misaligned (as seen when the box is inactive):
>    image
> 
> image
> 
> 1. search text doesn't have enough contrast
> 
> image
> 
> 1. profile dropdown is missing the arrow when on hover
>    image
> 2. custom status menu item polish (added notes to the ticket): https://mattermost.atlassian.net/browse/MM-37976
> 3. Clear button theming when the search box is not active:
>    image

Added to the notes but as discussed in the previous comments Michael Gamble requested that we remove the profile dropdown arrow.

esethna:
> Added to the notes but as discussed in the previous comments Michael Gamble requested that we remove the profile dropdown arrow.

@nevyangelova Ah I see, but I think the intention of @michaelgamble's comment was to remove the entire hover state on the status bubble (not just the arrow):

> Status icon triangle needs to go - instead have the mouse icon switch to the finger when hovering over the whole hit state (that will fix the obviousness problem)

andrewbrown00 (changes requested):
Hey @nevyangelova, we're getting closer 🙂

Items to resolve:


- [x] Status icon should be visible when hovering over the control, it should not be hidden


image
- [x] 1. Need 1px space between arrow controls
- [x] 2. Color should be `#FFF` using `56% opacity`
- [x] 3. Search placeholder and text should use `font-size: 12px; color: #FFF 64%` (placeholder only). Once the font-size changes, add 3px of top padding to align ‘search’ (input text) in the input field.
   
image
- [x] 1. Missing the ‘status’ icon
- [x] 2. The popover is too far to the left. The right edge of the popover should align with the right edge of the control.
- [x] 3. Align the ‘clear X’ with the status icon on the left
- [ ] 4. Missing the ‘checkmark’ icon to indicate the selected status
- [ ] 5. Fix ‘chevron’ icon alignment
- [ ] 6. Under DND, missing the text “Disables all notifications”
- [ ] 7. Clean up the left alignment of the icons
- [x] 8. Status icons are using the old icon font and should be using the `compass-icon` font
- [ ] 9. Menu items are using the old icon font and should be using the `compass-icon` font
- [ ] 10. Icons should use `rgba(var(--center-channel-color-rgb), 0.56);`

[Here's a link to the design file for reference](https://www.figma.com/file/iR1U1mW4y1U8S7qXWT9O1l/MM-35325-Global-Header-MVP?node-id=3%3A19994).

image
- [x] I believe this should use `rgba(var(--global-header-text-rgb),0.64);`. Seems like it’s using `rgba(var(--center-channel-color-rgb), 0.56);` which is causing the color issue when a user clicks out of the search input.

image
- [x] This should use `rgba(var(--center-channel-color-rgb))`. Seems like it’s using `rgba(var(--center-channel-color-rgb), 0.7)` which is causing the color issue lowering the opactiy of the input text.

image
- [ ] Icons should use `rgba(var(--center-channel-color-rgb), 0.56);`

michelengelen:
@nevyangelova I will review once all items are resolved. Would you be so kind and ping me once that is done? Thanks!

nevyangelova:
thanks @andrewbrown00 for the thorough feedback. Since this PR is focusing more on the header itself and the status dropdown had some unclarity we opened a new ticket to address that https://mattermost.atlassian.net/browse/MM-37976 however I did implement some of your points. The ones I didn't we can discuss separately since there are some problems with the current implementation of the menu and the fact that we are going to implement a new component soon. Any new points you might have on the status dropdown can be added to the ticket.

andrewbrown00 (changes requested):
Thanks, @nevyangelova, we're really close, a few more items focused on the Global Header only, not the Status popover which will be handled in a separate PR per your request 🙂 

image

- [x] The text is too high in the input and needs to be lowered to be aligned in the middle of the field. When I inspected it, looks like we need 3px of top padding which I added to:

   `.search__form .search-bar {
   padding: 3px 2.8rem 2px 3rem;
   }`

   Note, I only checked the alignment on Chrome/macOS.

image

- [ ] I don’t see the change to the ‘clear’ icon, it’s still hard to see. I believe this should use `rgba(var(--global-header-text-rgb),0.64);`. Seems like it’s using `rgba(var(--center-channel-color-rgb), 0.56);` which is causing the color issue when a user clicks out of the search input.

image

- [ ] Looks like the *Download Apps* icon color change was missed.

image

- [ ] The back and forward arrow behavior is broken. The arrows should be dimmed (indicate they are disabled) when the user hits the beginning or the end of the history stack. See Google Chrome browser arrows for reference. The `icon button` should be set to the `disabled` state to apply the correct styling.
   -  [Here’s a link to the design file for reference](https://www.figma.com/file/iR1U1mW4y1U8S7qXWT9O1l/MM-35325-Global-Header-MVP?node-id=1%3A70654). Also, see screenshot with component reference/style in Figma.
   - image

image

- [ ] The *Messages* and *Files* tags are not showing correctly in the search input in the selected state. [Here’s a link to the design file for reference](https://www.figma.com/file/iR1U1mW4y1U8S7qXWT9O1l/MM-35325-Global-Header-MVP?node-id=1235%3A126278).

image

- [ ] There should be 8px of space between the leading icon and the next item, whether it’s a tag (messages or files), modifier text, or the search text placeholder. Once a tag is added, the 8px should be maintained between the tag(s), modifier text, or the search text placeholder.

----------

**Adding this for visibility and not sure if you want to handle this here or in [MM-37976](https://mattermost.atlassian.net/browse/MM-37976)** since it's the popover position and not the actual content.

image

- [ ] Popover is too far to the left. The right edge of the popover should align with the right edge of the control

nevyangelova:
@andrewbrown00 fixed most points. The two I left out: 
- History arrows shouldn't even appear in the browser and I left them out because there is a PR which removes them from the browser and there should be a ticket handling error related to them. 
- Let's solve the menu positioning in the other ticket as one of the issues that the current component has is positioning itself.

deanwhillier:
Per the [thread in the `Feature: Global Header` channel](https://community-daily.mattermost.com/core/pl/gr33p71bnfnwb8hds5s5fttmxr) on community, the following points will have outstanding changes/comments/fixes handled in the corresponding tickets so we can merge this PR.

- User Status Button – MM-37976
- History Navigation Buttons – MM-32371
- RHS positioning – MM-37365

deanwhillier (approved):
LGTM, approving with one suggestion that I will also accept to keep things moving.

andrewbrown00 (approved):
Thanks, @nevyangelova, changes look great 🚀 

* Verified the product switcher aligns with the design and functionality ✅ 
* Verified the search control aligns with the design and functionality ✅ 
* Verified the icon button states are correct for active vs toggled ✅ 
* Verified user/status popover is positioned correctly ✅ 
* Verified themes work correctly and styles are maintained ✅ 

Product switcher | Active vs Toggled | User | Status
----- | ----- | ----- | -----
image | image image | image | image




**Search control**

Default | Selected | Filled | Tags | Modifiers
----- | ----- | ----- | ----- | -----
image | image | image image | image | image

**Dark theme**

image

-----

@nevyangelova I found 1 small issue that may be related to themes and we can handle in a separate PR. The search input text is very light for dark themes:

image

michelengelen (approved):

PR #8589, opened 2021-08-14
[MM-37351] Split settings modal

iomodo:
#### Summary
This PR splits settings modal into two

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-37351

#### Related Pull Requests

#### Screenshots

https://user-images.githubusercontent.com/2340127/129442749-7f10eb22-a533-4e97-8cde-450c4414cee0.mp4

#### Release Note

```release-note
NONE
```


components/user_settings/modal/index.ts
```diff
@@ -31,6 +36,8 @@ function mapStateToProps(state: GlobalState) {
         sendEmailNotifications,
         requireEmailVerification,
         collapsedThreads,
+        globalHeaderEnabled: getGlobalHeaderEnabled(state),
+        isContentChannelPreferences: ownProps.isContentChannelPreferences,
```
devinbinnie, line 40: This line is unnecessary, it should pass this through anyways.

devinbinnie (approved):
One non-blocking comment, but otherwise LGTM

components/user_settings/modal/user_settings_modal.tsx
```diff
@@ -326,10 +339,17 @@ class UserSettingsModal extends React.PureComponent<Props, State> {
...
-                        />
+                        {this.props.globalHeaderEnabled && this.props.isContentChannelPreferences ? (
+                            <FormattedMessage
+                                id='channel_header.channelPreferences'
+                                defaultMessage='Channel Preferences'
+                            />
+                        ) : (
+                            <FormattedMessage
+                                id='user.settings.modal.title'
+                                defaultMessage='Account Settings'
```
nevyangelova, line 350: nit: maybe we should keep the naming consistent, account or profile.

components/user_settings/modal/index.ts
nevyangelova: why do we need this check? It seems like the globalHeaderEnabled could be enough for the use case. Also we don't need to add it to the index.ts if its coming from ownProps

nevyangelova (changes requested):

components/user_settings/modal/index.ts
iomodo: @nevyangelova, @devinbinnie agree on the `ownProps` 

@nevyangelova When `globalHeaderEnabled` is on we can have 2 modals `Channel Preferences` or `Account Settings`. `isContentChannelPreferences`  flag determines which one to show.

components/user_settings/modal/user_settings_modal.tsx
iomodo: `Account Settings` is the Title of the modal. `Profile` is the tab. So was in the figma files.

esethna:
@iomodo confirming that accessing the "Account Settings" works when in Boards or Playbooks?

esethna (approved):
Approving on behalf of UX. QA please confirm the Account Settings are accessible from Boards and Playbooks

nevyangelova (approved):
LGTM :)

jespino:
@iomodo @esethna Not necessarily for this PR, but the theme is for the whole suite, not only Channels, so maybe that should be in the user settings.

jgilliam17:
Thanks @iomodo 
Tested, looks good to merge.
- Verified Old Account Settings have been split into two new settings modals and all existing settings options on both modals are working as before
**_Channel Preference_** --  accessible from global header gear icon; opens with Messaging specific options Notifications, Display, Sidebar and Advanced
**_Account Settings_** -- accessible from global header profile icon; opens with options Profile and Security
_**Boards and Playbooks**_ > QA will test both new settings modals on daily after merge. Currently test servers are not allowing manual plugin upload - not able to verify on the PR server.

jgilliam17 (approved):

PR #8591, opened 2021-08-15
[MM-17615] handle filter/show only Administrators in Members modals

rakhi2104:
#### Summary
This PR allows users to filter members based on roles in Team and Channel members list.

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-server/issues/17615

#### Screenshots

https://user-images.githubusercontent.com/17671710/129484707-a61cbe58-74f1-4db9-8012-d5846198774f.mp4

#### Release Note

```release-note
* Add functionality to filter users in Team and Channel members list 
* Partially refactored code for Team Members List to match that of Channel Members List.
```


rakhi2104:
Apologies for not pushing any tests.
Thought a draft PR wouldn't trigger any tests.

components/member_list_channel/member_list_channel.tsx
```diff
@@ -135,21 +139,38 @@ export default class MemberListChannel extends React.PureComponent<Props, State>
...
     }
 
+    toggleShowOnlyAdminUsers= (shouldShowOnlyAdminUsers: boolean): void => {
+        let filters = {};
+        if (shouldShowOnlyAdminUsers) {
+            filters = {
+                roles: [],
+                team_roles: [],
+                channel_roles: [Constants.PERMISSIONS_CHANNEL_ADMIN],
+            };
```
rakhi2104, line 149: Realized just now, this is messing up a few scenarios. Can I know which roles we want to show/filter?

hmhealey: It sounds like we want all users that are either a system admin, team admin, or channel admin to show with this filter. I think system admin would go under roles and team admin would go under team roles if I remember how that system works correctly.

hmhealey:
Definitely feel free to ignore the CI stuff for now. We can fix that later on.

components/member_list_channel/member_list_channel.tsx
rakhi2104: Unfortunately, using the below is returning mixed results:

```js
filters = {
    roles: [Constants.PERMISSIONS_SYSTEM_ADMIN],
    team_roles: [Constants.PERMISSIONS_TEAM_ADMIN],
    channel_roles: [Constants.PERMISSIONS_CHANNEL_ADMIN],
};
```

image

The `searchProfilesInCurrentTeam` function is returning `UserProfile[]` which contains if the user is a `system_admin` while the `team_admin` role is missing. can we leverage any other selector which returns all the user roles?

hmhealey: The team_admin role would likely be on the TeamMembership object for a given user/team, and I think the same thing applies to channel_admin and the ChannelMembership objects. I think we should already have both of those loaded by the time the modal is loaded since they're shown on the dropdown at the right.

It looks like you might already have access to those objects since they come from `getMembersInCurrentTeam` and `getMembersInCurrentChannel`, so maybe take a look at filtering those out there? The only thing that might cause problems with that is if `searchProfilesInCurrentTeam` doesn't return enough users in a page to include the admins, but hopefully that's not the case since it'll require changes to the API and to some server code.

esethna:
@rakhi2104 let us know when this is ready to test on our test servers and continue reviews

hmhealey:
Hey @rakhi2104. Anything we can help with on this?

esethna:
@rakhi2104 let us know if you're still working on this?

esethna:
Hey @rakhi2104 let us know if you're still working on this! Would love to see this in the product!

esethna:
Hey @rakhi2104 how's this going, are you still working on it?

hmhealey:
Thanks for your work on this, @rakhi2104. We're going to close this to let someone else take a shot at this ticket.

PR #8594, opened 2021-08-16
MM-37916: close product switcher on item click

michelengelen:
#### Summary
implements a new optional prop on the `menuItem` HOC: ~~`onClick`~~  `handleItemClick`.
Adds this ~~`onClick`~~  `handleItemClick` to each element in the product switcher menu to close the switcher when an item is clicked.

#### Ticket Link
[MM-37916](https://mattermost.atlassian.net/browse/MM-37916)

#### Release Note
```release-note
NONE
```


components/widgets/menu/menu_items/menu_item.tsx
```diff
@@ -12,6 +12,7 @@ export default function menuItem(Component: React.ComponentType<any>) {
         id?: string;
         icon?: React.ReactNode;
         text?: React.ReactNode;
+        handleItemClick?: (event: React.MouseEvent<HTMLElement>) => void;
```
jespino, line 15: Why not `onClick`?

jespino: I guess this is because it is colliding with the underneath components. But I think all the underneath components should implement the onClick. That would be more consistent from the interface perspective. Than having an extra `handleItemClick` and in some cases also an `onClick`.

michelengelen: Yeah, you are right. It did have conflicts with the onClick prop of a component. 

Thinking about it now I might need to adjust this a bit, since I did not want it to have it being called twice when named onClick.

I will thinking a better solution. Thanks for pointing that out.

michelengelen: @jespino I added a new solution. Could you re-review and maybe take into consideration (since you are more experienced with this component) if the changes might have unwanted side-effects? It could possibly be, but pretty unlikely, that someone planned for an event to bubble form the component. I just want to make sure that this solution is not breaking anything!

andrewbrown00 (approved):
Looks good @michelengelen, thanks 🙂

Let me know if I need to re-review after the code review. I don't think I will since the functionality shouldn't change, just how you're handling it.

jgilliam17:
@michelengelen For some reason there are 2 integrations option on the product switcher, it closes on item click and that's good, that is fixed 🙂 
Can you take a look if this doubling issue is related to your changes? Thanks.
Screen Shot 2021-08-16 at 4 01 56 PM

michelengelen:
Oh yes! That's my fault ... I added one more item for testing. Will remove tomorrow!

nevyangelova:
There is a conflict with master

nevyangelova (approved):
LGTM!

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
- Verified product switcher closes on item click.

michelengelen:
kind reminder for @jespino to re-review.

jespino (approved):
LGTM

PR #8600, opened 2021-08-16
[MM-37682] Check for landing preference before automatically clearing it

devinbinnie:
#### Summary
We were never checking if the landing preference was checked or not on unload, so it was always clearing it. So now we clear it only if it's supposed to be cleared.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37682

#### Related Pull Requests
- Has server changes: https://github.com/mattermost/mattermost-server/pull/18146

#### Release Note
```release-note
Fixed an issue with the landing page where the landing preference would not save correctly.
```


jwilander (approved):

mkraft (approved):

jgilliam17 (approved):
Thanks @devinbinnie 
Tested, looks good to merge.
- Verified deep link redirect preference is remembered.

PR #8604, opened 2021-08-16
Clean up some unused channel-related code

hmhealey:
Each commit should be fairly self-explanatory, but to provide some more context on the first two:
- Fake channels are an old concept that was used to show channels in the sidebar that didn't exist yet. A very long time ago, new users would automatically have a number of DM channels appear in their sidebar to make Mattermost look more friendly to new users, but those channels weren't created in the DB until the user viewed them. That was later removed during a previous sidebar redesign, and that code was left over from before that.
- `Channel.ExtraUpdateAt` is no longer used on the server as of https://github.com/mattermost/mattermost-server/pull/8762 (https://mattermost.atlassian.net/browse/MM-9739). It's still around and unused, and removing it from the type definition on lets us just ignore it on the client.

This is done in preparation for https://mattermost.atlassian.net/browse/MM-37021 since the rest of that is blocked by https://github.com/mattermost/mattermost-webapp/pull/8503 

#### Release Note
```release-note
NONE
```


lieut-data (approved):
Thank you for the discrete commits!! Looks good, and excited for the next round of changes.

larkox (approved):
LGTM.

hmhealey:
Skipping QA review since this is just dead code removal.

amyblais:
@hmhealey Is this for v6.0?

hmhealey:
Nope, it's not needed for it. Thanks for checking though

PR #8608, opened 2021-08-17
Adds websockets to products

mgdelacroix:
#### Summary
This PR adds the websockets client to the products, and wraps their components in a `LoggedIn` component so the client is correctly initialised.

#### Release Note

```release-note
Added websocket client to products
```


jespino (approved):
LGTM

crspeller (approved):
Looks good. 
Took at look at the LoggedIn component. It seems like there is quite a lot of other random stuff in there besides websockets. Hopefully none of that interferes. (Didn't see anything that looked like trouble.)

PR #8617, opened 2021-08-17
MM-37542: support tooltip react components

lieut-data:
#### Summary
To facilitate changing the tooltip text on playbooks depending on the context, support a React component when registering the tooltip text. This was actually already supported downstream, but wasn't correctly supported in the registration api.

#### Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-37542

#### Release Note
```release-note
Support React component in channel header tooltips registered by plugins.
```

### DIFF
--- a/plugins/registry.js
+++ b/plugins/registry.js
@@ -115,7 +115,7 @@ export default class PluginRegistry {
     // - icon - React element to use as the button's icon
     // - action - a function called when the button is clicked, passed the channel and channel member as arguments
     // - dropdown_text - string or React element shown for the dropdown button description
-    // - tooltip_text - string shown for tooltip appear on hover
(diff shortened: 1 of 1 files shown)


plugins/registry.js
```diff
@@ -125,7 +125,7 @@ export default class PluginRegistry {
             icon: resolveReactElement(icon),
             action,
             dropdownText: resolveReactElement(dropdownText),
-            tooltipText,
+            tooltipText: resolveReactElement(tooltipText),
```
crspeller, line 128: Can we update the comment to state it takes a react element as well?

lieut-data: Mais oui!

crspeller (approved):

agarciamontoro (approved):
Nice!

PR #8622, opened 2021-08-17
MM-37429 Fix detection of manually sized collapsible images

hmhealey:
Part of the issue reported in MM-37429 is that if a user overrode the size of a Markdown image and didn't specify the height like
```markdown
![an image](https://example.com/image.png =50)
```
we wouldn't be able to correctly detect whether or not the image should be collapsible. This leads to either large images not being collapsible or small images being collapsible and shifting onto a new line.

As part of this, I added a `parseImageDimensions` helper to replace the old code for detecting overridden dimensions. The previous code wasn't working as designed since it never actually returned an "auto" width or height. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37429

#### Release Note
```release-note
Fixed detection of certain collapsible images
```


AshishDhama (approved):
Looks good to me

devinbinnie (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified when posting markdown images where size is overridden without specifying the height, collapsibility can still be detected.

PR #8629, opened 2021-08-18
Fix aria label on channel switcher

larkox:
#### Summary
On this version, whenever a screen reader was to read the channels in the channel switcher, would find in the aria label something like "Object". This was because we were passing a whole object to the aria label instead of the string.

This PR should solve this issue.

#### Test steps
Open the channel switcher (CTRL+K) and use a screen reader to see that the names of the channels are read out loud without any problem. Alternatively, you can check the aria labels of the div that defines each row in the list.

#### Ticket Link
None

#### Release Note
```release-note
Fix texts for screen readers on channel switcher.
```

### DIFF
--- a/components/suggestion/switch_channel_provider.jsx
+++ b/components/suggestion/switch_channel_provider.jsx
@@ -86,14 +86,8 @@ class SwitchChannelSuggestion extends Suggestion {
             className += ' suggestion--selected';
         }
 
-        let displayName = (
-            <React.Fragment>
-                {channel.display_name}
-                <div className='mentions__fullname'>
-                    {`~${channel.name}`}
-                </div>
-            </React.Fragment>
-        );
+        let name = channel.display_name;
+        let description = '~' + channel.name;
 
         let icon;
         if (channelIsArchived) {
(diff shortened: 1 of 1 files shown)


amyblais:
@hmhealey This is for v5.37.x ESR dot release.

devinbinnie (approved):

hmhealey (commented):
This looks like it might cause a slight change in behaviour where we no longer read out if users are deactivated. I think that could be fine given that that's not critical information in the channel switcher, but maybe @esethna or @devinbinnie would have a better idea about that.

Also, is there not a ticket accompanying this? If it's something that we're cherry picking, it should probably have a ticket for tracking it across branches.

esethna (approved):

esethna:
Reading if the user is deactivated is not critical info in my mind for the channel switcher

amyblais:
> Also, is there not a ticket accompanying this? If it's something that we're cherry picking, it should probably have a ticket for tracking it across branches. 

Daniel mentioned that it was fixed via https://github.com/mattermost/mattermost-webapp/commit/e70518015faf2ef43d7f26553aba4f030139b1c7 earlier, which was in v5.38, so in that sense it doesn't need a ticket to track across branches. But I can create a ticket for QA testing.

hmhealey:
Oh, I see. I thought this PR was against master and that we'd be cherry-picking it later. That's fine by me then.

hmhealey (approved):

jgilliam17:
Thanks @larkox 
It works as expected on Windows 10 using JAWS reader, channel names are read when using up/down arrows to navigate through the list.
On macOS Catalina and Big Sur when using VoiceOver I am not getting any channels read.  Can you please take a look? Thanks.

larkox:
I have now any to test it on macOS, and I don't see significant differences with the aria labels used in other places. Not sure how to look deeper into this.

PR #8630, opened 2021-08-18
Fix attachment images

larkox:
#### Summary
There were some errors in markdown and attachment images (and related, in embedded bindings).
1. Attachment images were not handling the "onToggle" action. This made the "Show More" "Show Less" to still be visible and not properly working when we collapse the image.
2. Attachments (and many other markdown location) did not provide a "onImageHeightChanged" property. This would, in the end, not set the "onToggle" action, which make the attachment break.

#### Test Steps
In the Jira link you can see test steps to reproduce these issues consistently.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38003

#### Release Note
```release-note
Fix crash when markdown images are present on the message attachments and embedded bindings.
```

### DIFF
--- a/components/markdown_image_expand/markdown_image_expand.tsx
+++ b/components/markdown_image_expand/markdown_image_expand.tsx
@@ -9,7 +9,7 @@ export type Props = {
     children: React.ReactNode;
     isExpanded: boolean;
     postId: string;
-    onToggle: (isExpanded: boolean) => void;
+    onToggle?: (isExpanded: boolean) => void;
     actions: {
         toggleInlineImageVisibility: (postId: string, imageKey: string) => void;
     };
@@ -19,7 +19,7 @@ const MarkdownImageExpand: React.FC<Props> = ({children, alt, isExpanded, postId
     const {toggleInlineImageVisibility} = actions;
 
     useEffect(() => {
-        onToggle(isExpanded);
+        onToggle?.(isExpanded);
     }, [isExpanded]);
 
     const handleToggleButtonClick = () => {
--- a/components/post_view/message_attachments/message_attachment/message_attachment.tsx
+++ b/components/post_view/message_attachments/message_attachment/message_attachment.tsx
@@ -259,6 +259,7 @@ export default class MessageAttachment extends React.PureComponent<Props, State>
                     <Markdown
                         message={field.title}
                         options={markdown}
+                        postId={this.props.postId}
                     />
                 </th>,
             );
@@ -268,7 +269,10 @@ export default class MessageAttachment extends React.PureComponent<Props, State>
                     className='attachment-field'
                     key={'attachment__field-' + i + '__' + nrTables}
                 >
-                    <Markdown message={String(field.value)}/>
+                    <Markdown
+                        message={String(field.value)}
+                        postId={this.props.postId}
+                    />
                 </td>,
             );
             rowPos += 1;
@@ -311,7 +315,10 @@ export default class MessageAttachment extends React.PureComponent<Props, State>
             preTextClass = 'attachment--pretext';
             preText = (
                 <div className='attachment__thumb-pretext'>
-                    <Markdown message={attachment.pretext}/>
+                    <Markdown
+                        message={attachment.pretext}
+                        postId={this.props.postId}
+                    />
                 </div>
             );
         }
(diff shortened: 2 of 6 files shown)


utils/message_html_to_component.jsx
```diff
@@ -130,6 +132,8 @@ export function messageHtmlToComponent(html, isRHS, options = {}) {
                     <MarkdownImage
                         className={className}
                         imageMetadata={options.imagesMetadata && options.imagesMetadata[attribs.src]}
+                        height={parseInt(height, 10)}
```
hmhealey, line 135: Funny that you hit this right when I did in https://github.com/mattermost/mattermost-webapp/pull/8622. These actually need to be strings since they can be either the empty string or "auto". Do my changes fix the issue you were running into with passing them as strings?

components/post_view/message_attachments/message_attachment/message_attachment.tsx
```diff
@@ -77,6 +77,7 @@ export default class MessageAttachment extends React.PureComponent<Props, State>
 
         this.imageProps = {
             onImageLoaded: this.handleHeightReceived,
+            onImageHeightChanged: this.checkPostOverflow,
```
hmhealey, line 80: Are you sure this is needed here? I'm pretty sure we should know the dimensions of the image because the app should have received those in the post metadata.

hmhealey: Oh, I see. This is called when a collapsed image is toggled, is that correct? I guess we never thought to add that when we made Markdown images collapsible.

utils/message_html_to_component.jsx
larkox: I am pretty sure your solution deals with the string problem (those were mainly about annoying warnings in the webapp). I will revert the related changes and focus only on the toggle problem.

components/markdown_image_expand/markdown_image_expand.tsx
```diff
@@ -9,7 +9,7 @@ export type Props = {
     children: React.ReactNode;
     isExpanded: boolean;
     postId: string;
-    onToggle: (isExpanded: boolean) => void;
+    onToggle?: (isExpanded: boolean) => void;
```
nevyangelova, line 12: Is this ok to be optional now that we are passing 'onImageHeightChanged' correctly?

nevyangelova (approved):
otherwise LGTM!

components/markdown_image_expand/markdown_image_expand.tsx
larkox: We are passing it on posts and attachments. Markdown might be used in more places. If we do not make it optional, we might find unexpected crashes in other places.

hmhealey (approved):

DHaussermann:
This PR solves the web app crash.
When regression testing, I saw an odd behavior where collapsing images that are part of an attachment will now collapse images in other posts that also have markdown images in their attachments.
This behavior may not be new to this PR. @larkox will investigate.

larkox:
@DHaussermann The error was that since we were not passing the post id to the different markdown elements, it stored the "collapse" status in the same state variable. I raised another potential issue on [community](https://community.mattermost.com/core/pl/bh8hu5mwq7dxtkc5a7wchg8kuc), but that one is completely out of the scope of this PR.

DHaussermann (approved):
Tested and passed
- Webapp crash resolved
- Issue where expand / collapse was affecting multiple post resolved
- Regression tested default value for expand / collapse is respected 
- Regression tested markdown images not in attachments
LGTM!

amyblais:
@larkox Is this a regression and how likely is it that users see this bug?

larkox:
@amyblais The bug was introduced, if I am not mistaken, by this commit:
https://github.com/mattermost/mattermost-webapp/commit/66b49a432d9b9ed7e64c0e990bdbfad99bdd8464

This was a new feature, and the bug was present since then, so not 100% sure whether it can be classified as regression.

The likelihood of encountering this is hard to quantify, since it is regarding Markdown Images present in the text of the Message Attachments (AKA Slack Attachments). Since this feature is mainly thought for integrations (plugins, webhooks, etc...) it is really hard to tell who might be affected.

AFAIK, our plugins do not do anything like this, so those are not affected, but others (and custom integrations) may be.

larkox:
@amyblais Correcting myself on my previous comment. The new feature had nothing to do with the Markdown Images in the attachment description, so yes, it is a regression.

PR #8605, opened 2021-08-16
[MM-37157, MM-37159] Add new branding graphics welcome to MM, tips & next steps and shared background

deanwhillier:
#### Summary
Updates the "download apps" graphic and background graphic (MM-37159), both with theming. Also updates the graphic for the "Welcome to Mattermost" screen that precedes tips and next steps (MM-37157). Background update is shared by both tickets

The background uses a `<symbol/>` of the pill shape and surrounding 'stipple' and is instanced for each pill visible in the background with appropriate transforms to rotate and position. The 'stipple' is made with an inlined JPG (base64) used as a mask on a `<rect />` to allow for theming while keeping the overall size a fraction of what it would be if the `stipple` was vector-based.

In talking with @matthewbirtch, it was decided that the background would be anchored to the top left and would not scale up past "100%". It does scale down to a min of "50%" however, otherwise the graphic gets lost on smaller screens ... can adjust as needed.

This PR also makes a small tweak to scrolling to remove the permanently visible scrollbar. There was a recent fix for [a bug](https://mattermost.atlassian.net/browse/MM-37065) where scrolling was not working on iOS. The fix uses `-webkit-overflow-scrolling: touch;` and `overflow-y: scroll;` to fix the scrolling issue. Seems that only the `-webkit-overflow-scrolling: touch;` is needed to fix the iOS bug and the other can remain as `overflow-y: auto;` so the scrollbar only shows when there is content to be scrolled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37159
https://mattermost.atlassian.net/browse/MM-37157

### Related Tickets
https://mattermost.atlassian.net/browse/MM-37065

#### Screenshots
image

#### Release Note
```release-note
NONE
```

PR #8631, opened 2021-08-18
[MM-37707] - Move next steps menu item to help dropdown and adjust mo…

nevyangelova:
#### Summary
As part of the new global header in messaging we are moving the next steps menu item to the user help dropdown and we have to adjust the modal backdrop and the modals accordingly. However since now the header container is fluid we cannot definitively position the backdrop cutout. As a proposed solution I have excluded the whole global header from the backdrop so its visible enough for the user. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37707

#### Screenshots
Screenshot 2021-08-18 at 14 35 48
Screenshot 2021-08-18 at 14 35 57

#### Release Note

```release-note
NONE
```


esethna (approved):
@nevyangelova looks good.

0/5 should we fix the styling of the mobile view modal in this PR? We have a separate ticket for it but it's very related to this work: https://mattermost.atlassian.net/browse/MM-37959

image

The rest of that ticket seems to fixed already (ie [Dean's PR](https://github.com/mattermost/mattermost-webapp/pull/8605) fixes the background overlaying the mobile view menu, and this PR fixes the placement of the green arrow in mobile view by hiding it (as per expected)

components/generic_modal.tsx
```diff
@@ -23,6 +23,7 @@ type Props = {
     autoCloseOnCancelButton?: boolean;
     autoCloseOnConfirmButton?: boolean;
     enforceFocus?: boolean;
+    container?: any;
```
michelengelen, line 26: why `any`? cant we be a bit more strict here?

components/global/user_guide_dropdown/user_guide_dropdown.test.tsx
```diff
@@ -29,6 +29,8 @@ describe('components/channel_header/components/UserGuideDropdown', () => {
         helpLink: 'helpLink',
         reportAProblemLink: 'reportAProblemLink',
         enableAskCommunityLink: 'true',
+        showGettingStarted: false,
+        showNextStepsTips: false,
```
michelengelen, line 33: is `showNextSteps` missing here?

components/global/user_guide_dropdown/user_guide_dropdown.tsx
```diff
@@ -19,6 +21,8 @@ type Props = {
     helpLink: string;
     reportAProblemLink: string;
     enableAskCommunityLink: string;
+    showGettingStarted: boolean;
+    showNextStepsTips: boolean;
```
michelengelen, line 25: same here

michelengelen (changes requested):

andrewbrown00 (changes requested):
Thanks, @nevyangelova, a few items to resolve:

- [x] Bump up the illustration to centered over the ‘?’ icon, it sits a bit low.
   
   Current | Expected
   --------- | ---------
    image | image
 
- [ ] Not if this is a bug on my end, but clicking the ‘Tips & Next Steps’ option from the Help menu doesn’t open the Tips & Next Steps view, it does nothing.

   

- [ ] Confirm this should be using the `active` state and not the `toggled` state. Design made the decision to use the `active` state for popovers and the `toggled` state for controls that open panels, i.e. mentions and saved.

   image


- [ ] 1. Missing the illustration and tooltip, tooltip should reflect “Main Menu” for the mobile view
- [ ] 2. I believe the mobile view should still read “Main Menu” since the item will be listed under the main menu. cc @esethna 
- [ ] 3. Missing the buttons

   image

components/generic_modal.tsx
nevyangelova: I was following the type convention for the react bootstrap Modal for consistency :D but we can restrict it to a node for sure.

components/global/user_guide_dropdown/user_guide_dropdown.tsx
nevyangelova: I am not sure we need this here as the initial display is handled by channel_view and the menu item just serves as unhiding it. (see previous functionality)

nevyangelova:
thanks @andrewbrown00 and @esethna, I have implemented the mobile changes as well. I wasn't sure if its part of this scope of this PR so I will leave a comment in https://mattermost.atlassian.net/browse/MM-37959 so that it's not worked on twice. @andrewbrown00 fixed all points, except the one for the toggle state of the help dropdown since this is fixed in the polish global header PR

components/generic_modal.tsx
```diff
@@ -23,6 +23,7 @@ type Props = {
     autoCloseOnCancelButton?: boolean;
     autoCloseOnConfirmButton?: boolean;
     enforceFocus?: boolean;
+    container?: React.ReactNode;
```
michelengelen, line 26: `React.ReactNode | React.ReactNodeArray`

michelengelen (changes requested):
one last change to be made

components/generic_modal.tsx
nevyangelova: does the container need to be an array of nodes? I would add it for the children prop but in this case i feel like its always going to be one element.

esethna:
Thanks @nevyangelova, mobile modal still looks a bit off:

image

nevyangelova:
@esethna fixed

michelengelen (approved):
LGTM

andrewbrown00 (approved):
Thanks, @nevyangelova, changes look good 🙂 

* Verified the 'Tips & Next Steps' has been moved
   *  WebApp moved to the **? help menu**  ✅  
   * Mobile Web View moved to the **main menu**  ✅  
* Verified the overlay is present when a user attempts to dismiss the T&NS
   * WebApp ✅  
   * Mobile Web View ✅  
* Verified once the T&NS is dismissed, accessing from the help menu reopens the view
   * WebApp ✅ 
   * Mobile Web View ✅  
* Verified the language changes based on the view
   *  WebApp (Help Menu)  ✅  
   * Mobile Web View (Main Menu)  ✅  
 
WebApp | Mobile Web View
----- | -----
image | image

esethna:
@deanwhillier @jgilliam17 friendly ping to help review this

deanwhillier (changes requested):
LGTM, one suggestion for variable naming

deanwhillier (approved):

matthewbirtch:
Looks good now @michelengelen. @jgilliam17 the issue you caught should be resolved now

jgilliam17:
Thanks @michelengelen, text looks good in light and dark themes.
Placement of the arrow in the mobile view is now off a bit
Screen Shot 2021-08-30 at 10 24 41 AM

nevyangelova:
@jgilliam17 i can't reproduce, but if you are checking this by shrinking the screen the 'isMobile' check isn't going to be triggered before refresh so you will have to test with the device inspector.

michelengelen:
@nevyangelova we really need some kind of check for that, that listens on size changes and stores it in the redux store to be consistent. I already have an idea to add a certain hook to the app that does exactly that. We can talk this through and maybe get @hmhealey or @crspeller into the boat?

nevyangelova:
I know its good to have this detect screen size change but that would mean refactoring the 'isMobile' check, which I assume is out of scope for this PR, but definitely let's align!

michelengelen:
It definnitely is out of scope for this PR. I just mentioned it here so that everyone knows, that there is a need and I have an idea on how to tackle this! :D

jgilliam17:
@nevyangelova I don't think I understand the refresh instructions for mobile view. Are you saying that if I reduce the screen, refresh and then try to leave the Tips and Next Steps the arrow with help text will be positioned correctly?

nevyangelova:
@jgilliam17 correct. This or using the device inspector in Chrome, the point is it won't happen on resize at the moment. Sorry for the inconvenience!

jgilliam17:
Thanks @nevyangelova On the test servers, it is not positioning for me correctly after refresh. Should we add a follow up ticket to fix this?

nevyangelova:
For which resolution? The 'isMobile' check should work for all screen sizes under 768px?

nevyangelova:
@jgilliam17 nice catch for the multi teams! There was an additional CSS check that didn't take mobile view into account. Fixed now

jgilliam17:
Thank @nevyangelova for figuring it all out. 
Tested, looks good to merge. 
- Verified next steps item has been moved to help menu and that the illustrations are positioned correctly.

jgilliam17 (approved):

PR #8643, opened 2021-08-18
Mm 37462 add channel split

neallred:
#### Related Pull Requests

#### Release Note

```release-note

```

### DIFF
--- a/components/sidebar/channel_navigator/channel_navigator.tsx
+++ b/components/sidebar/channel_navigator/channel_navigator.tsx
@@ -13,9 +13,10 @@ import {isDesktopApp} from 'utils/user_agent';
 import AddChannelDropdown from '../add_channel_dropdown';
 import ChannelFilter from '../channel_filter';
 import InviteMembersButton from '../invite_members_button';
-import {InviteMembersBtnLocations} from 'mattermost-redux/constants/config';
+import {InviteMembersBtnLocations, AddChannelButtonTreatments} from 'mattermost-redux/constants/config';
 
 type Props = {
+    addChannelButton?: AddChannelButtonTreatments;
     canGoForward: boolean;
     canGoBack: boolean;
     canJoinPublicChannel: boolean;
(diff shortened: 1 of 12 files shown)


esethna:
Thanks @neallred,

For the sidebar BG color test, please note the circular button (not just on hover) 
image

It's hard to see the elevation of the buttons, when hovering and not. Is there a way to test them on the server?

neallred:
Keen eyes! :eyes: Test B background was wrong. Will update the videos and update here when the split is ready.

neallred:
Access here: https://mattermost-server-pr-18254.test.mattermost.cloud


Account Type | Username | Password
-- | -- | --
Admin | sysadmin | Sys@dmin123
User | user-1 | User-1@123

Using mattermost-server's spinwick as currently it can pick up webapp branch code, whereas vice-versa webapp is having trouble picking up mattermost-server branch code.

esethna:
@neallred can you please ping @matthewbirtch for a review when this is ready? Not sure how or if it's possible for us to test the variations?

neallred:
Ahh, yeah, this is ready now! Because split.io is based on groups of servers rather than traffic, it will require me manually toggling them for you. Let me know when you would like to change variations. It only takes a second to change, and it updates in roughly 1 minute.

matthewbirtch (changes requested):
@neallred just took a quick look and found the following:

1. I can get in to a scenario where both the team menu and the `+` menu are open at the same time causing overlapping menus. This seems to happen if I click the team name first, then click the `+` button.
image

2. Having the `add members` icon button right under the `+` button seems really redundant. These tests aren't going to happen at the same time are they? Can we confirm that we won't actually see this?

image

3. Can we move the chevron that is now next to the team name a little closer? I think if we remove the left margin on the icon it will do the trick

4. It feels odd to make the blank space between the team name and the plus button all clickable now. Can we isolate this clickable area to the team name and chevron only?

5. I think we do need to come up with a hover state for the team name now. Can we do something similar to what we do for the channel name dropdown? We could use 8% opacity of the sidebar text color for the background with a 4px border-radius. We would also need to add some padding to the container. The active state could change the background to 8% opacity. Let me know if you think we can squeeze this in. 

Here are the states:
image


I also added the states in the figma file: https://www.figma.com/file/hY66Ff8plZD2X6gTFR2Ey1/MM-37462-More-Prominent-Plus-Button?node-id=90%3A37568

components/sidebar/channel_navigator/channel_navigator.tsx
```diff
@@ -78,21 +79,25 @@ export default class ChannelNavigator extends React.PureComponent<Props> {
...
-                canJoinPublicChannel={this.props.canJoinPublicChannel}
-                handleOpenDirectMessagesModal={this.props.handleOpenDirectMessagesModal}
-                unreadFilterEnabled={this.props.unreadFilterEnabled}
-                townSquareDisplayName={this.props.townSquareDisplayName}
-                offTopicDisplayName={this.props.offTopicDisplayName}
-                showTutorialTip={this.props.showTutorialTip}
-            />
-        );
+        let addChannelDropdown = null;
+        if (!this.props.addChannelButton || this.props.addChannelButton === AddChannelButtonTreatments.NONE) {
```
marianunez, line 83: @neallred Here we should check if the Global Header is off? Currently, the `AddChannelButtonTreatments` is set to the default we configured for 6.0 https://github.com/mattermost/mattermost-server/pull/18281 and if `GlobalHeader` is off then it seems to be causing https://mattermost.atlassian.net/browse/MM-38612

neallred: I agree; we should check and I think it is the cause of the bug. I can take the bug up.

PR #8644, opened 2021-08-19
[MM-38214] Added Inline Latex support

KobeBergmans:
#### Summary

Added inline Latex support by changing the marked library (https://github.com/mattermost/marked). Rendered using the katex library (just as the latex blocks). The code must be present in between dollarsigns ($ latex code here $). The rendering of inline Latex is controlled by an option in the admin_console/site_config/posts menu.

#### Ticket Link

Ticket added: https://mattermost.atlassian.net/browse/MM-38214

There is no ticket for this PR. I am a student and I work for my professors who use Mattermost as their main form of communication. Untill today they used an old hack to get inline Latex working in the webapp but it didn't work anymore with the current server build so they asked me to get it working with the current build. This feature is very important for them to enable easy communication of math symbols and equations. There are also a lot of people who share this opinion as there was a suggestion for this back in 2015: https://mattermost.uservoice.com/forums/306457-general/suggestions/11206086-allow-latex-inline-latex-here.

#### Related Pull Requests

- Has server changes: https://github.com/mattermost/mattermost-server/pull/18219
- Has marked changes: https://github.com/mattermost/marked/pull/20

(If the PR is accepted, there needs to be a change so that the marked library points to the mattermost repo again)

#### Screenshots

Screenshot_42

Screenshot_43

#### Release Note

```release-note
Added new config setting: ServiceSettings.EnableInlineLatex. Added inline Latex rendering
```


Willyfrog:
Hello and thanks for the PR, I've asked a review from a PM as we don't have a ticket to validate the PR against, but thanks for taking the effort to do this!

@hmhealey I've also added you as you might also want to look into that :)

esethna:
Curious to hear thoughts from @hmhealey. I'm not opposed to taking a feature that makes inline latex easier to use as long as we aren't compromising the stability, compatibility and feature set of our current markdown.

KobeBergmans:
I don't think the stability, compatibility or feature sets have changed of the current markdown, the feature is just an extra that can be turned off if the user does not want to use it. (The feature is turned off by default)

There is also a problem with the test server because on the current server master branch you can't open the admin_console/site_config/posts page so you can't turn the feature on.

hmhealey:
Thanks for the PR! I'll take a look through the code shortly.

I'm always a bit hesitant about adding new stuff to the Markdown, partly because our version of marked is horribly out of date (I want to eventually replace it with a new library), and I'm also a bit hesitant about adding to Latex since we've wanted to move that to a plugin for quite a while now (which likely requires the new Markdown library). That all said, this has been a feature a lot of people have asked for in the past, so I'd be open to potentially taking this.

hmhealey:
> There is also seems to be a problem with the test server because on the current server master branch you can't open the admin_console/site_config/posts page so you can't turn the feature on.

That's a known issue with a fix currently in the works. I think I can hack the setting on though

utils/markdown/index.ts
```diff
@@ -18,12 +22,15 @@ export function format(text: string, options = {}, emojiMap: EmojiMap) {
...
 export function formatWithRenderer(text: string, renderer: marked.Renderer) {
+    const config = getConfig(store.getState());
+
     const markdownOptions = {
         renderer,
         sanitize: true,
         gfm: true,
         tables: true,
         mangle: false,
+        inlinelatex: config.EnableInlineLatex === 'true',
```
hmhealey, line 33: @KobeBergmans Actually, the test server we set up isn't picking up the correct server branch, so it doesn't know the new setting exists. To get around that, could you temporarily set this to true so we can try it out?

KobeBergmans: Yes i'll do it right away!

KobeBergmans:
> Thanks for the PR! I'll take a look through the code shortly.
> 
> I'm always a bit hesitant about adding new stuff to the Markdown, partly because our version of marked is horribly out of date (I want to eventually replace it with a new library), and I'm also a bit hesitant about adding to Latex since we've wanted to move that to a plugin for quite a while now (which likely requires the new Markdown library). That all said, this has been a feature a lot of people have asked for in the past, so I'd be open to potentially taking this.

I saw the thread about moving Latex into it's own plugin, and i looked into making inline Latex a plugin but i didn't seem to find a way to get that working.

utils/markdown/index.ts
KobeBergmans: @hmhealey the test server now works as intended.

hmhealey:
Yeah, better plugin support is actually one of the things I wanted to get working when we redo it, and it's specifically for things like this. The way the existing Latex blocks are implemented is already kind of a hack since it just overrides how a code block renders, and I'd like to replace those with something cleaner. But that's all for the future.

In the meantime, this looks pretty great, but there's a couple things I think we might want to change:
1. For the syntax to trigger the inline Latex, I was looking around at how different programs handle that. The `$<latex>$` pattern sounds common enough (Pandoc uses it), but there's also `\(<latex>\)` (Rocket.chat) and $\`<latex>\`$ (GitLab). I think I found the most references to the one you used, so it's probably our best option unless we wanted to add support for multiple formats.

    I was able to "unintentionally" trigger it by making a post that lists some prices like "$1.50 $4.35" though, so I think others might run into that. When trying to find how others handled that problem, I did find [this note in the Pandoc manual](https://pandoc.org/MANUAL.html#math) that references some more conditions for the opening and closing dollar signs, so maybe we could do something like that to avoid people using latex by mistake? 
2. I think we can have marked recognize the latex blocks, but I think we'll have to have Mattermost be the one that calls out to katex to render it. Katex is a bit of a heavier library, so we have the `LatexBlock` component lazily load it only when it's needed, but I don't know if we can keep that behaviour if marked has to depend on it itself.

    Instead of putting the latex rendering into marked's renderer, we could leave a dummy render function there and have the renderer in Mattermost handle it like it does with `LatexBlock`.

Thoughts on both of those?

KobeBergmans:
The `$latex$` syntax also felt the most comfortable to me because it is the syntax used when writing latex documents. It could be a nice addition if there was an option to choose the syntax, I don't know if it would work great with the marked library but I'll have a look at that tommorow.

I did also run into the trigger when listing prices, you can escape the inline latex by typing: "$$ $1.50 $$ $4.35". This works because the inline latex only triggers if there is atleast a space in between the dollar signs. But the pandex note seems to be good solution and I will implement that tomorrow.

I don't know a lot about performance in Javascipt but it is indeed probably a better solution to also lazily load it. I don't really know how lazy loading works in Javascript: is it a feature of react components or is it just because the `LatexBlock` code is in another file?

Last week I already made a proof of concept version that did not depend on modifying the marked library: it works by checking the output of marked for dollar signs inside of `<p>` elements. If it finds those it calls Katex to change them. Does this seem like a better option to you? Or is it best to let marked pass on a dummy element that then is changed inside of the mattermost code? (If you are interested in the proof of concept here is the code: https://github.com/KobeBergmans/mattermost-webapp/tree/inline-latex-htmlhijack. It is pretty garbage code but it works 😉)

KobeBergmans:
I just finished making the suggested changes:

I implemented a new regex search for the inline latex wich only detects a latex sequence if the last `$` is not followed by a word character (numbers and letters). This allows latex to be at the end of a sentence but the prices example works just fine now.

I looked into the lazy loading and it is indeed a lot faster to do it that way. I made marked still search for the inline latex syntax  but i made it pass a dummy `<span>` element with the relevant latex code. This code is then rendered inside a new react component `LatexInline`.

hmhealey:
Those changes look good to fix the lazy loading. I think we can also avoid redefining the types by doing something like
```typescript
import type * as Katex from 'katex';

type State = {
    katex?: Katex;
}
```
Does that work?

hmhealey:
Now that the server changes have been merged, the overrides to force this setting on can be removed

KobeBergmans:
> Now that the server changes have been merged, the overrides to force this setting on can be removed

@hmhealey you are aware that the current server master branch still does not display the admin_console/site_config/posts page? Or can you work around this bug?

hmhealey:
Right, I forgot the System Console page was also broken. That's fixed on the latest master now that https://github.com/mattermost/mattermost-webapp/pull/8646 has been merged though.

And that works for me. Anything that makes it easier to keep that working next time I update dependencies is good with me 😛.

You can merge it with PR. I think it's related closely enough to this PR to merge them at the same time.

KobeBergmans:
@hmhealey I merged the commits onto this branch and it seems to work nicely. Is there a way to update the test server so that the system console works?

hmhealey:
You might have forgotten to push the merge commit here since I'm not seeing the needed changes on the branch. Once those are pushed, the test server will pick them up

components/latex_block/latex_block.test.tsx
```diff
@@ -12,28 +12,34 @@ describe('components/LatexBlock', () => {
         enableLatex: true,
     };
 
-    test('should match snapshot', () => {
+    test('should match snapshot', async () => {
         const wrapper = shallow(<LatexBlock {...defaultProps}/>);
+        await import('katex'); //manually import katex
```
hmhealey, line 17: You might be able to change this to
```javascript
await Promise.resolve();
```
I think we just need to await on something to let the `LatexBlock`'s `import(...).then` be run, so it doesn't matter what exactly we're awaiting on here.

components/latex_block/latex_block.test.tsx
```diff
@@ -12,28 +12,34 @@ describe('components/LatexBlock', () => {
...
+        wrapper.update();
```
hmhealey, line 18: Is this needed? I'd expect the snapshot to be updated automatically once the import statement causes the `setState`

hmhealey (changes requested):

KobeBergmans:
> You might have forgotten to push the merge commit here since I'm not seeing the needed changes on the branch. Once those are pushed, the test server will pick them up

@hmhealey sorry for the late answer i was on vacation untill today.

I looked at my local branch and there is no merge commit i need to push. This is probably because the branch i made the changes on was just a branch created of the latest commit of the branch this PR is made from.

I also tried the suggested changes and if I changed the `await import('katex')` the test fail on my machine. The `wrapper.update()` was not needed so i deleted it.

hmhealey:
I'll try merging master into this. That should fix it.

KobeBergmans:
@hmhealey That did the trick, now everything works as intented

hmhealey (approved):

components/latex_inline/latex_inline.tsx
```diff
@@ -0,0 +1,75 @@
...
+        });
+    }
+
+    render(): React.ReactNode {
+        if (!this.props.enableInlineLatex || this.state.katex === undefined) {
+            return (
+                <span
+                    className='post-body--code inline-tex'
+                >
+                    {'$' + this.props.content + '$'}
```
Willyfrog, line 40: should content have any `$` sign escaped? It's been a long time since I used latex so I'm not 100% sure if it could be a problem

Willyfrog (approved):
LGTM! thanks for the PR

Added a comment, but this should be good to merge as it is

components/latex_inline/latex_inline.tsx
KobeBergmans: I am not sure what you mean with this comment. This line of code is executed when the inline latex setting is dissabled or katex is not yet loaded. So this content is not rendered.

Willyfrog: oh, but I thought people would be able to select and copy it and that's why you put the `$` signs around, I'm not sure if that's an actual latex delimiter or just some sort of convention

KobeBergmans: The `$` is just added in the text because marked only returns the content of the latex code. So if latex rendering is dissabled then i just added them back in and displayed the latex code.

jgilliam17:
Thanks @KobeBergmans 
Looks great, just one small issue.
Can we also add a check and prevent triggering latex when using $ on the right of the numeric value? 
for example: `some text with 15$ and 20$`

hmhealey:
That's a really good catch. I guess we need to add a condition to the first $ that it appears before a word similar to how the second one has to appear after a word.

Technically, someone could still accidentally trigger Latex rendering if they mix both styles of dollar values, but I really don't expect people to do that.

KobeBergmans:
@jgilliam17 I made suggested change.

Right now these cases do not trigger:
- `some text with 15$ and 20$`
- `some text with $15 and $20`
- `some text with 15$ and $20`

Only this one triggers: `some text with $15 and 20$`. But i think that is unavoidable

hmhealey:
@KobeBergmans Thanks for all your work on this! The last thing left to do on this is to point marked back to commit `6ca9a6b3f4bdd35dbf58d06f5e53369791e05915` now that that PR has been merged

KobeBergmans:
@hmhealey I changed the reference to the mattermost repo.

Thanks for accepting the changes, it was a pleasure working with you guys!

hmhealey (approved):

hmhealey:
It was great working with you as well!

I'll admit that I've been a bit hesitant about adding inline latex in the past when people have requested it since I'm cautious about making more changes to our Markdown implementation, but it's a lot easier to say yes to when given a nearly-complete PR with tests and everything already completed 😄

PR #8646, opened 2021-08-19
[MM-37985] Fix config names

hanzei:
#### Summary
https://github.com/mattermost/mattermost-server/pull/18128, https://github.com/mattermost/mattermost-server/pull/18103 and https://github.com/mattermost/mattermost-server/pull/18152 changed the field names of some config values. This unintentionally broke the system console as the config json parsing relies on the field name. While touching the code I've also renamed variables, i18n ids and component names to keep them in sync with the config field values.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/18149
https://mattermost.atlassian.net/browse/MM-37985

#### Release Note
```release-note
NONE
```


components/admin_console/custom_url_schemes_setting.jsx
```diff
@@ -11,7 +11,7 @@ import LocalizedInput from 'components/localized_input/localized_input';
 
 import Setting from './setting';
 
-export default class CustomUrlSchemesSetting extends React.PureComponent {
+export default class CustomURLSchemesSetting extends React.PureComponent {
```
hmhealey, line 14: I agree with updating the names of values that come from the server and their corresponding i18n keys and such, but I think we've tended towards the previous style of capitalizing acronyms in class names in the past. I don't think there's really an idiomatic style for naming in JavaScript though, so we could consider changing that here as well.

That all said, I think this change is fine. It is getting me thinking about us taking another crack at standardizing styling throughout the JS code base again though.

e2e/cypress.json
```diff
@@ -13,10 +13,10 @@
         "adminPassword": "Sys@dmin-sample1",
         "allowedUntrustedInternalConnections": "localhost",
         "cwsUrl": "http://localhost:8076",
-        "cwsApiUrl": "http://localhost:8076",
+        "cwsAPIURL": "http://localhost:8076",
```
hmhealey, line 16: Should this be changing? It seems like something that might be needed by Cypress itself

e2e/cypress.json
```diff
@@ -13,10 +13,10 @@
...
         "dbClient": "postgres",
         "dbConnection": "postgres://mmuser:mostest@localhost/mattermost_test?sslmode=disable\u0026connect_timeout=10",
-        "elasticsearchConnectionUrl": "http://localhost:9200",
+        "elasticsearchConnectionURL": "http://localhost:9200",
```
hmhealey, line 19: Same here

e2e/cypress/support/api/cloud_default_config.json
```diff
@@ -260,7 +260,7 @@
         }
     },
     "DisplaySettings": {
-        "CustomUrlSchemes": [],
+        "CustomURLchemes": [],
```
hmhealey, line 263: There's a typo here that's hopefully not affecting too many files

e2e/cypress/support/api/on_prem_default_config.json
```diff
@@ -488,11 +488,11 @@
         "EnableRemoteMarketplace": true,
         "AutomaticPrepackagedPlugins": true,
         "RequirePluginSignature": false,
-        "MarketplaceUrl": "https://api.integrations.mattermost.com",
+        "MarketplaceURL": "https://api.integrations.mattermost.com",
         "SignaturePublicKeyFiles": []
     },
     "DisplaySettings": {
-        "CustomUrlSchemes": [],
+        "CustomURLchemes": [],
```
hmhealey, line 495: Like this one too

e2e/cypress/support/api/system.js
```diff
@@ -123,8 +123,8 @@ export const getDefaultConfig = () => {
 
     if (isCloud) {
         fromCypressEnv.CloudSettings = {
-            CWSUrl: cypressEnv.cwsUrl,
-            CWSAPIUrl: cypressEnv.cwsApiUrl,
+            CWSURL: cypressEnv.cwsUrl,
```
hmhealey, line 126: Opposite my question on `e2e/cypress.json`, if that was intentional, should `cwsUrl` be renamed then?

package-lock.json
```diff
@@ -89,6 +89,7 @@
         "stream-browserify": "3.0.0",
         "styled-components": "5.3.0",
         "superagent": "6.1.0",
+        "timezones.json": "1.5.2",
```
hmhealey, line 92: This should be fixed if you update to the latest master. Git seems to be handling it fine though

hmhealey (changes requested):
Have we looked into how those server changes affect mobile? The changes in the full config.json likely won't affect it since it doesn't have the System Console, but any changes to the client config should affect it. Since the mobile app has to be compatible with older server versions, I could see that causing problems.

Similarly, the client config is used by plugins too, so have we considered how that might affect them?

hmhealey:
Thinking more about this, I think we should revert the name changes made in the client config since I'm really worried about those affecting mobile and plugins as well. For those, the server and client aren't in lockstep for versions, so we have to worry about their older code trying to use a newer server and failing to find the config settings by their new names.

The changes to the server config used in the System Console are probably fine though since I don't think the mobile app or plugins use those at all.

e2e/cypress.json
hanzei: I wanted to keep the cypress env names in sync with the config names. They are used to communicate values to the cypress tests. AFAIK they are only defined here.

e2e/cypress/support/api/cloud_default_config.json
hanzei: Fixed it.

hanzei: Thanks for catching the typo. Actually, `CustomUrlSchemes` is correct as it comes from the client manifest.

e2e/cypress/support/api/system.js
hanzei: Good catch. Fixed in edb0b6fc1.

e2e/cypress/support/api/cloud_default_config.json
hanzei: Thanks for catching the typo. Fixed it.

saturninoabril (approved):
Okay, the issue on tests were related to test servers. New report can be found here - https://automation-dashboard.vercel.app/cycles/b862cbcc-9b94-4f2b-a39b-2b3b8b2ab261

Webapp crash on System Console > Site Configuration > Posts is verified fix and works as expected.

Did random test and everything seemed to work fine. Will going to merge this now and will observe on daily test.

PR #8648, opened 2021-08-19
make `ProductBranding` component clickable

michelengelen:
#### Summary
re-arranged the left portion of global header to allow for the `ProductBranding` to be clickable and open the `ProductSwitcherMenu` as well (alongside the `ProductSwitcher` 'IconButton`)

The `ProductBranding` currently has no visual guidance for it to being a clickable element. It would probably be a good idea to at least give it some kind of `hover` state to make it more clear to the user.

Thoughts @andrewbrown00  ?

#### Ticket Link
[MM-37986](https://mattermost.atlassian.net/browse/MM-37986)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/global/product_branding/product_branding.tsx
+++ b/components/global/product_branding/product_branding.tsx
@@ -15,6 +15,7 @@ type Props = {
 const ProductBrandingContainer = styled.div`
     display: flex;
     align-items: center;
+    cursor: pointer;
 
     > * + * {
         margin-left: 8px;
--- a/components/global/product_branding/product_branding.tsx
+++ b/components/global/product_branding/product_branding.tsx
@@ -8,6 +8,10 @@ import Heading from '@mattermost/compass-components/components/heading';
 
 import {useProducts, useCurrentProductId} from '../hooks';
 
+type Props = {
+    onClick: React.MouseEventHandler<HTMLDivElement>;
+}
+
 const ProductBrandingContainer = styled.div`
     display: flex;
     align-items: center;
@@ -17,13 +21,13 @@ const ProductBrandingContainer = styled.div`
     }
 `;
 
-const ProductBranding: React.FC = (): JSX.Element | null => {
+const ProductBranding: React.FC<Props> = ({onClick}: Props): JSX.Element | null => {
     const products = useProducts();
     const currentProductID = useCurrentProductId(products);
     const currentProduct = products?.find((product) => product.id === currentProductID);
 
     return (
-        <ProductBrandingContainer>
+        <ProductBrandingContainer onClick={onClick}>
             <Icon
                 size={20}
                 glyph={currentProduct && typeof currentProduct.switcherIcon === 'string' ? currentProduct.switcherIcon : 'product-channels'}
--- a/components/global/product_branding/product_branding.tsx
+++ b/components/global/product_branding/product_branding.tsx
@@ -11,7 +11,6 @@ import {useProducts, useCurrentProductId} from '../hooks';
 const ProductBrandingContainer = styled.div`
     display: flex;
     align-items: center;
-    cursor: pointer;
 
     > * + * {
         margin-left: 8px;
--- a/components/global/product_branding/product_branding.tsx
+++ b/components/global/product_branding/product_branding.tsx
@@ -8,10 +8,6 @@ import Heading from '@mattermost/compass-components/components/heading';
 
 import {useProducts, useCurrentProductId} from '../hooks';
 
-type Props = {
-    onClick: React.MouseEventHandler<HTMLDivElement>;
-}
-
 const ProductBrandingContainer = styled.div`
     display: flex;
     align-items: center;
@@ -22,13 +18,13 @@ const ProductBrandingContainer = styled.div`
     }
 `;
 
-const ProductBranding: React.FC<Props> = ({onClick}: Props): JSX.Element | null => {
+const ProductBranding: React.FC = (): JSX.Element | null => {
     const products = useProducts();
     const currentProductID = useCurrentProductId(products);
     const currentProduct = products?.find((product) => product.id === currentProductID);
 
     return (
-        <ProductBrandingContainer onClick={onClick}>
+        <ProductBrandingContainer>
             <Icon
                 size={20}
                 glyph={currentProduct && typeof currentProduct.switcherIcon === 'string' ? currentProduct.switcherIcon : 'product-channels'}
--- a/components/global/product_switcher.tsx
+++ b/components/global/product_switcher.tsx
@@ -11,6 +11,7 @@ import Menu from 'components/widgets/menu/menu';
 import MenuWrapper from 'components/widgets/menu/menu_wrapper';
 
 import {useClickOutsideRef, useCurrentProductId, useProducts} from './hooks';
+import ProductBranding from './product_branding';
 import ProductSwitcherMenu from './product_switcher_menu';
 import ProductSwitcherTip from './product_switcher_tip';
 
@@ -133,6 +134,7 @@ const ProductSwitcher = (): JSX.Element => {
                         aria-label='Select to open product switch menu.'
                     />
                     <ProductSwitcherTip/>
+                    <ProductBranding onClick={handleClick}/>
                 </ProductSwitcherContainer>
                 <Menu
                     className={'product-switcher-menu'}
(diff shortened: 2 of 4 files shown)


andrewbrown00:
Hey, @michelengelen, I don't see the option to enable the **Global Header** in the account settings or in the system console.

image

michelengelen:
it should be enabled by default through a feature flag, or am I missing something @spirosoik ?

andrewbrown00:
> @andrewbrown00 it should be enabled by default through a feature flag, or am I missing something @spirosoik ?

I thought that was the case too, but I don't see it enabled on this test server 🤔

image

spirosoik:
@michelengelen @andrewbrown00 in spinwicks we are not setting anything we just actually call the provisioner or CWS to create the envs. Let me see what are the default values for provisioner in test

spirosoik:
@michelengelen @andrewbrown00 doesn't seem to have this as default value

andrewbrown00:
> @michelengelen @andrewbrown00 doesn't seem to have this as default value

Hey, @spirosoik, who has to set the `default value`? It's blocking me from reviewing a handful of PR's.

cc @michelengelen

spirosoik:
@andrewbrown00 we are syncing the changes in provisioner to have this, I will ping you when it's done

spirosoik:
@michelengelen @andrewbrown00 is this the env variable? `MM_GLOBALHEADER`

spirosoik:
@michelengelen @andrewbrown00 what exactly do we want to see? The global header is forced by Split as enabled

michelengelen:
ok @spirosoik i just tested CWS and the feature flag is not set there as well ... this is the feature flag we set during local server init: `MM_FEATUREFLAGS_GLOBALHEADER=true`

spirosoik:
@michelengelen Split should force it to enable as I see in the test servers. Are we sure that split and global header work as expected?

michelengelen:
@spirosoik I will check that

michelengelen:
@spirosoik on my local machine it works like a charm.
When we set the env variable it gets added to the current App state, like seen in this screenhost:

Screenshot 2021-08-20 at 10 49 58

We then access it with a helper function like this: `const featureFlagEnabled = getFeatureFlagValue(state, 'GlobalHeader') === 'true';`

It accesses the state and gets the property `FeatureFlag{key}`, in our case `FeatureFlagGlobalHeader`.

The only case this is not working is when the feature flag is not set (or the flag is set, but not stored in the app state correctly).

I will try to have a look at the state on the server.

spirosoik:
@michelengelen yes with env variable will definitely work the question is that we have this in Split and overall we are not setting feature flags static for the purpose to be dynamic through split. So something goes wrong with Split I think and it doesn't get the right value.

spirosoik:
In practice all these values are coming from Split on purpose. I don't see global header though at all, like it's not there. Could you rebase just to be sure there is no issue with a wrong version of server? (wondering

Screenshot 2021-08-19 at 9 54 52 PM

michelengelen:
WOW ... that rebase is messy for some reason. :/

michelengelen:
@andrewbrown00 ready for review. Test server is running.

cc/ @spirosoik

spirosoik:
so it was rebase issue

michelengelen:
probably yes

andrewbrown00:
@michelengelen I'm getting the white screen on this PR now. I made sure my cache was clear and tried in Incognito, but no luck. Are you seeing the white screen as well?

cc @spirosoik

michelengelen:
well, not on this one ... strange though

Screenshot 2021-08-20 at 17 59 45

cc/ @spirosoik

andrewbrown00:
@michelengelen this is what I'm seeing. I also include the `console` to see if it helps.

image

michelengelen:
@andrewbrown00 and @jgilliam17 ... I just rebased this branch onto the current master, which incorporates the fix for the "white screen of death" bug. Gentle ping to re-review once the test server is up and running again.

matthewbirtch (approved):
I think this works @michelengelen. We will likely want to revisit whether we give the product brand an active/hover state, but we can look at that post 6.0

matthewbirtch:
@michelengelen one thing that would be good: can you change the cursor to the pointer when hovering over the product brand?

michelengelen:
> @michelengelen one thing that would be good: can you change the cursor to the pointer when hovering over the product brand?

@matthewbirtch done! 👍🏼

components/global/product_switcher.tsx
```diff
@@ -142,6 +143,7 @@ const ProductSwitcher = (): JSX.Element => {
                         aria-label='Select to open product switch menu.'
                     />
                     <ProductSwitcherTip/>
+                    <ProductBranding onClick={handleClick}/>
```
crspeller, line 146: Maybe we can make the ProductSwitcherContainer handle the click so we don't have to pass this to two components?

crspeller (changes requested):

components/global/product_switcher.tsx
michelengelen: i did this to prevent closing of the menu when the divider is being clicked

michelengelen: oh, and it is easier to handle the action on the same level as the state for the opening/closing of the menu sits

michelengelen:
@deanwhillier and @crspeller kind reminder to re-review

components/global/product_switcher.tsx
michelengelen: sry ... my previous comments were for a different kind of component ... I mixed those up 🙇🏼‍♂️ 

But the same kinda goes for this as well: We don't want the space to be clickable, but only those that are visually indicating a potential action and by passing it to the container we would have the effect that the whole area is clickable

crspeller: But the only two things in the area are the two things we want clickable. I think it would be a bug if there was a tiny space between them that was unclickable. Presumably we want the whole area to eventually highlight not just the grid button?

crspeller:
@michelengelen Not tally specific to this PR but you might want to read https://mattermost.com/blog/submitting-great-prs/ it's a great read and you might want to look closely at #4. Makes it much easier on reviewers 😄

michelengelen:
> @michelengelen Not tally specific to this PR but you might want to read https://mattermost.com/blog/submitting-great-prs/ it's a great read and you might want to look closely at #4. Makes it much easier on reviewers 😄

thanks for the link ... I did not know of that. I rebased solely because we recently had issues with spinwicks not getting the latest server changes when we pulled master into the working branch. This might be resolved, but I wanted to make sure that all changes are up to date.

FYI: this is not how i normally do that! :D

components/global/product_switcher.tsx
michelengelen: about the highlighting: that is something for UX to decide upon (although IMO you are right)
the clickable area: it depends ... right now there is only the 2 items with no space in between, so we can certainly apply it to the parent component

selectors/global_header.ts
```diff
@@ -3,9 +3,12 @@
 
 import {GlobalState} from 'types/store';
 import {getFeatureFlagValue} from 'mattermost-redux/selectors/entities/general';
+import * as UserAgent from '../utils/user_agent';
+import * as Utils from '../utils/utils';
 
 export function getGlobalHeaderEnabled(state: GlobalState): boolean {
     const featureFlagEnabled = getFeatureFlagValue(state, 'GlobalHeader') === 'true';
+    const isMobileView = UserAgent.isMobile() || Utils.isMobile();
```
crspeller, line 11: It looks like this unrelated code has appeared. Bad rebase?

crspeller (changes requested):

selectors/global_header.ts
michelengelen: yep ... sadly! will remove

crspeller (approved):

components/global/product_branding/product_branding.tsx
```diff
@@ -11,6 +11,7 @@ import {useProducts, useCurrentProductId} from '../hooks';
 const ProductBrandingContainer = styled.div`
     display: flex;
     align-items: center;
+    cursor: pointer;
```
deanwhillier, line 14: I'm wondering if this should maybe be relocated to `ProductSwitcherContainer` since the `onClick` handler is there so we can keep the functionality and the visual affordance for the functionality "together".

deanwhillier (changes requested):
Apologies for the delayed review @michelengelen, I somehow lost track of this one. LGTM with one possible suggested tweak.

michelengelen:
@deanwhillier feel free to merge once you re-review and approve! 🙇🏼‍♂️

deanwhillier (approved):

PR #8658, opened 2021-08-19
MM-37021 Track message count outside of Channel object

hmhealey:
As an attempt to improve performance, we're trying to make it so that the Channel object changes less frequently in the web app by splitting it up throughout the Redux store. By separating the fields that rarely change (ID, name, type) from those that change a lot (message count, last post at), we're hoping to reduce rendering of components that only care about Channels but not their unread state.

This PR splits out the message count, and there'll be a follow up to move the Channel.LastPostAt out later.

To do this, there's a few important things that I want to note with this PR:
1. Everything that accessed Channel.TotalMsgCount or Channel.TotalMsgCountRoot (with channels in state.entities.channels.channels) now has to access MessageCount.Total and MessageCount.Root (stored in state.entities.channels.messageCounts). Unfortunately, this means that there's a lot of code that went from caring about the Channel and ChannelMember to needing to care about both of those plus a MessageCount, so there's a bump in code complexity, but hopefully it should help performance in the long run.

2. Removing fields from the Channel object requires us to define Channel objects coming from the server differently from those being stored on the client, so we ended up with a separate ServerChannel type that's used only in API responses. Hopefully this pattern works out in case we need to do this more down the line.

    I also had to remove those fields from the Channel objects stored in Redux. That's not exactly new since we already do that with Posts when receiving post metadata, but that leaves behind a bit of data instead of removing the field entirely. There is a small chance that this means we'll end up saving identical Channel objects that change the state if only their TotalMsgCount would have changed, but I don't think that's likely enough to happen that we'd want to do any sort of shallow comparison to prevent that.

3. I had also intended on moving ChannelMember.MsgCount and ChannelMember.MentionCount into their own object, but unfortunately, ChannelMember.UpdateAt also changes whenever those change, so it seems like it won't be possible to move those out unless we stop using ChannelMember.UpdateAt as well.

4. Last but not least, I tried to centralize all the logic around determining whether or not a channel is unread in `calculateUnreadCount`. We seemed to randomly duplicate a lot of the logic around checking if a channel is unread and how muting affects that, so I hope to have cleaned all that up. There's a chance that this introduces some subtle bugs, but I think there's an equal chance that it actually fixes some other minor bugs that might have existed previously.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37021

#### Release Note
```release-note
NONE
```


devinbinnie (approved):

packages/mattermost-redux/src/types/teams.ts
```diff
@@ -52,10 +52,23 @@ export type TeamsState = {
...
     mention_count: number;
-    msg_count: number;
+
+    // The number of unread mentions in root posts in channels on this team, not including DMs and GMs
     mention_count_root: number;
+
+    // The number of unread posts in channels on this team, not including DMs and GMs
+    //
+    // Note that this differs from ChannelMember and ChannelUnread since it tracks unread posts instead of read posts.
+    msg_count: number;
```
calebroseland, line 65: While not strictly necessary, the `/** */` syntax can attach to the type of following member for better discovery.

image

```suggestion
    /**
     * The number of unread posts in channels on this team, not including DMs and GMs
     * @remarks Note that this differs from ChannelMember and ChannelUnread since it tracks unread posts instead of read posts.
     */
    msg_count: number;
```

calebroseland (approved):
(Sorry this took me forever to review.) 
This is excellent, and quite thorough. One comment on type comment formatting, but this LGTM!

packages/mattermost-redux/src/types/teams.ts
hmhealey: Oh interesting. I didn't know there was special syntax to trigger that. I had wondered why that shows up sometimes and not other times.

hmhealey: Comments updated. I wasn't able to use `@link` there since the files don't reference each other, but I did use them in some of the other ones.

jgilliam17:
@hmhealey something is wrong with the test server. Here are 3 errors from the console, server is loading at all for me.

Screenshot 2021-09-02 at 17 38 40 
Screenshot 2021-09-02 at 17 38 30  
Screenshot 2021-09-02 at 17 37 59

jgilliam17:
Disregard, this must be a test server issue. Also seeing the same error on other PRs.

jgilliam17:
@hmhealey Per sresupport, please rebase the branch onto latest master, that should fix the errors on the test server. Thanks 🙂

jgilliam17:
I reviewed [Cypress results](https://automation-dashboard.vercel.app/cycles/0ecc5aa8-63f2-418b-89f8-ad8d4e86f7c8) last week, no unread failures except one test that also fails on master. I just need the test server for manual smoke test. Thanks.

jgilliam17 (approved):
Thank you @hmhealey 
Tested, looks good to merge.
- Verified marking channels read/unread is working as before. Checked E2E results, no unusual failures when compared to master.
Failing test for marking post as unread on RHS that reproduces on master will be filed separately.  https://mattermost.atlassian.net/browse/MM-38413

PR #8662, opened 2021-08-19
MM-37712 Allow text to be drag and dropped into inputs

hmhealey:
So the reason that we didn't allow drag and dropping text into inputs previously was because of how `dragster.js` (originally from https://github.com/catmanjan/jquery-dragster which itself was a port of https://github.com/bensmithett/dragster if anyone is curious) called `preventDefault`. You have to call `preventDefault` on `dragenter` and `dragover` events to keep the drag active (since the default stops the drag), but us calling it on `drop` only needs to be done when we want to prevent the drop from doing anything.

In our case, we wanted to prevent the page from navigating which happens if you drop a file or URL onto the page, and not when you drop text onto an input. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37712
Closes https://github.com/mattermost/mattermost-server/issues/18074

#### Release Note
```release-note
Added support for dragging and dropping text into inputs
```


components/file_upload/file_upload.jsx
```diff
@@ -385,9 +377,19 @@ export class FileUpload extends PureComponent {
         this.props.onFileUploadChange();
     }
 
-    registerDragEvents = (containerSelector, overlaySelector) => {
-        const self = this;
+    handleDropText = (e) => {
+        const targetType = e.target.nodeName.toLowerCase();
```
Willyfrog, line 381: nit: can we rename to targetNode or targetName? given there is an `e.target.type` I would expect targetType to contain that info.

components/file_upload/file_upload.jsx
```diff
@@ -705,3 +711,13 @@ export class FileUpload extends PureComponent {
 const wrappedComponent = injectIntl(FileUpload, {forwardRef: true});
 wrappedComponent.displayName = 'injectIntl(FileUpload)';
 export default wrappedComponent;
+
+// Checks if a data transfer contains files not text, folders, etc..
+// Slightly modified from http://stackoverflow.com/questions/6848043/how-do-i-detect-a-file-is-being-dragged-rather-than-a-draggable-element-on-my-pa
+export function isFileTransfer(files) {
+    if (UserAgent.isInternetExplorer() || UserAgent.isEdge()) {
```
Willyfrog, line 718: do we need to check for edge? seems [it is compatible](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer)

Willyfrog (approved):
LGTM, a couple of nitpicks there, but nice, looking forward to being able to do this :D

components/file_upload/file_upload.jsx
hmhealey: I feel like I had that exact thought at some point, so I can't remember why I didn't change that before.

components/file_upload/file_upload.jsx
hmhealey: Probably not since that check is over 5 years old, but I don't remember enough about it to remove it. I'd rather leave it for now until we end up cleaning up old references to IE11 and Edge throughout the code since we can do a proper test of it then.

BenCookie95:
I've been trying to test this but it doesn't seem to work. I'm dragging text from another text input outside MM but it isn't dropping it into the compose message input

hmhealey:
@BenCookie95 Which browser are you using for that? It's working for me on Firefox/Mac at least.

BenCookie95:
> @BenCookie95 Which browser are you using for that? It's working for me on Firefox/Mac at least.

@hmhealey It was chrome/mac

hmhealey:
Ah, I see it now. I guess there's some other magic we need to do with the drag and drop events to get them to fire correctly. I'll take another look.

PR #8664, opened 2021-08-20
MM-35017: ordering and inclusion of participants

koox00:
#### Summary

Participants' avatars in threads should be shown in reverse reply order,
the one who replied last should be shown first and so forth.

Root poster's avatar should not be shown, except in the case of global
threads, where it is shown always as the first item.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35017

#### Related Pull Requests

- Has server changes (https://github.com/mattermost/mattermost-server/pull/18235)

#### Release Note

```release-note
NONE
```


components/threading/global_threads/thread_item/thread_item.tsx
```diff
@@ -78,7 +78,15 @@ function ThreadItem({
...
 
-    const participantIds = useMemo(() => thread?.participants?.map(({id}) => id), [thread?.participants]);
+    const participantIds = useMemo(() => {
+        const ids = thread?.participants?.flatMap(({id}) => {
+            if (id === post.user_id) {
+                return [];
+            }
+            return id;
+        }).reverse();
+        return [post.user_id, ...ids];
```
anurag6713, line 88: Should we include `post.user_id` by default or only if `post.user_id` is also found in the participants list?

koox00: as per https://github.com/mattermost/mattermost-server/pull/18235 the root poster won't be included in the participants list. So in global threads list we do the inclusion manually.

In thread footer the requirement is that we do not show the root poster if the user has not replied to the thread.

anurag6713: Got it thank you @koox00 🙂

anurag6713 (approved):
LGTM! Thank you 🎉

calebroseland (approved):
LGTM!

PR #8667, opened 2021-08-20
[MM-37814] - Adjust status dropdown according to design

nevyangelova:
#### Summary
Design fixes for the status dropdown that were outstanding in https://github.com/mattermost/mattermost-webapp/pull/8585

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37976

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -160,7 +160,11 @@
             cursor: pointer;
 
             .status {
-                background-image: linear-gradient(0deg, var(--sidebar-text-08), var(--sidebar-text-08));
+                background-image: linear-gradient(
+                    0deg,
+                    var(--sidebar-text-08),
+                    var(--sidebar-text-08)
+                );
             }
         }
 
@@ -168,7 +172,11 @@
             background: var(--sidebar-text-16);
 
             .status {
-                background-image: linear-gradient(0deg, var(--sidebar-text-16), var(--sidebar-text-16));
+                background-image: linear-gradient(
+                    0deg,
+                    var(--sidebar-text-16),
+                    var(--sidebar-text-16)
+                );
             }
         }
 
@@ -218,12 +226,18 @@
             }
         }
 
-        #statusDropdownMenu-timedDND {
+        #statusDropdownMenu-timedDND,
+        #statusDropdownMenu-timedDND #status-menu-dnd-timed, {
             .icon {
                 height: auto;
                 width: auto;
                 padding: 0 12px 0 0;
                 margin: 0;
+                top: unset;
+            }
+
+            .SubMenu__icon-right {
+                padding-top: 0;
             }
         }
 
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -218,6 +218,15 @@
             }
         }
 
+        #statusDropdownMenu-timedDND {
+            .icon {
+                height: auto;
+                width: auto;
+                padding: 0 12px 0 0;
+                margin: 0;
+            }
+        }
+
         .dnd--icon {
             vertical-align: middle;
         }
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -231,9 +231,9 @@
         #statusDropdownMenu-timedDND,
         #statusDropdownMenu-timedDND #status-menu-dnd-timed, {
             .icon {
+                top: unset;
                 width: auto;
                 height: auto;
-                top: unset;
                 padding: 0 12px 0 0;
                 margin: 0;
             }
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -253,14 +253,14 @@
             align-items: center;
 
             .status {
-                background-color: var(--global-header-background);
-                color: var(--global-header-text);
                 position: absolute;
                 top: auto;
                 right: -4px;
                 bottom: -4px;
                 width: 16px;
                 height: 16px;
+                background-color: var(--global-header-background);
+                color: var(--global-header-text);
 
                 &:hover {
                     cursor: pointer;
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -194,6 +194,7 @@
 
         .custom_status__container {
             line-height: 14px;
+            align-items: center;
         }
 
         .username-wrapper {
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -192,6 +192,10 @@
             top: 4px;
         }
 
+        .custom_status__container {
+            line-height: 14px;
+        }
+
         .username-wrapper {
             padding: 0 12px;
 
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -170,7 +170,7 @@
         }
 
         &.active {
-            background: var(--sidebar-text-16);
+            background: rgba(var(--global-header-text-rgb), 0.16);
 
             .status {
                 background-image:
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -2,7 +2,6 @@
 
 // css feature flag: apply styling overrides to legacy layout when global header is enabled
 .feature-global-header {
-    background: red;
     --global-header-background: var(--sidebar-teambar-bg);
     --global-header-text: var(--sidebar-header-text-color);
     --global-header-background-rgb: var(--sidebar-teambar-bg-rgb);
@@ -205,8 +204,8 @@
 
             .MenuItem__primary-text {
                 display: grid;
-                grid-template-columns: auto 1fr auto auto;
                 align-items: center;
+                grid-template-columns: auto 1fr auto auto;
 
                 .SubMenu__icon-right {
                     padding-top: 0;
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -163,8 +163,8 @@
                 background-image:
                     linear-gradient(
                         0deg,
-                        var(--sidebar-text-08),
-                        var(--sidebar-text-08)
+                        rgba(var(--global-header-text-rgb), 0.08),
+                        rgba(var(--global-header-text-rgb), 0.08)
                     );
             }
         }
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -193,8 +193,8 @@
         }
 
         .custom_status__container {
-            line-height: 14px;
             align-items: center;
+            line-height: 14px;
         }
 
         .username-wrapper {
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -175,6 +175,7 @@
         .Menu__content {
             right: -1px;
             left: inherit;
+            width: 264px;
         }
 
         .custom_status__row > #custom_status__clear {
@@ -195,6 +196,15 @@
         .MenuHeader {
             height: 50px;
             padding: 0 16px;
+        }
+
+        .MenuItem > button,
+        #status-menu-dnd-timed {
+            padding: 1px 19px;
+        }
+
+        .dnd--icon {
+            vertical-align: middle;
         }
 
         .status-wrapper {
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -150,7 +150,7 @@
         padding: 4px 8px;
         border-radius: 4px;
 
-        i {
+        .icon > i {
             color: rgba(var(--center-channel-color-rgb), 0.56);
             font-family: "compass-icons";
         }
@@ -200,7 +200,16 @@
 
         .MenuItem > button,
         #status-menu-dnd-timed {
-            padding: 1px 19px;
+            padding: 1px 16px;
+
+            .MenuItem__primary-text {
+                display: grid;
+                grid-template-columns: auto 1fr auto;
+            }
+
+            > .MenuItem__help-text {
+                padding-left: 28px;
+            }
         }
 
         .dnd--icon {
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -160,11 +160,12 @@
             cursor: pointer;
 
             .status {
-                background-image: linear-gradient(
-                    0deg,
-                    var(--sidebar-text-08),
-                    var(--sidebar-text-08)
-                );
+                background-image:
+                    linear-gradient(
+                        0deg,
+                        var(--sidebar-text-08),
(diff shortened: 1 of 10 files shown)


andrewbrown00 (changes requested):
Thanks, @nevyangelova, a few items to resolve:

image

- [ ] 1. Clean up the left alignment of the icons
- [ ] 2. Status icons are still using the old icons, should be using the `compass-icon` font
- [ ] 3. Missing the ‘checkmark’ icon to indicate the selected status
- [ ] 4. Align the DND subtext with the label

image

- [ ] The ‘offline’ icon blends in with the background. We may need to apply the color from the icon buttons in the header so the ‘offline’ icon will display correctly across themes. **Note** the other status icons display correctly.



- [ ] Selecting an option from the popover should set the selection (ie change the status) then close the popover after making a selection. Currently, it stays open until a user clicks outside of the popover.

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -47,6 +47,7 @@ export type Props = {
     direction?: 'left' | 'right';
     openUp?: boolean;
     styleSelectableItem?: boolean;
+    extraText?: string;
```
michelengelen, line 50: why not call it `helpText`?

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -239,6 +240,7 @@ export default class SubMenuItem extends React.PureComponent<Props, State> {
                             aria-label={Utils.localizeMessage('post_info.submenu.icon', 'submenu icon').toLowerCase()}
                         />}
                     {subMenuContent}
+                    {extraText && <span className='MenuItem__help-text'>{extraText}</span>}
```
michelengelen, line 243: `helpText` would actually make way more sense. and you already have that as a class-name, sooooo ... :D

michelengelen: And it would probably be better to structure it like this:

```typescript
<div
    className={classNames([{styleSelectableItemDiv: styleSelectableItem}])}
    id={id}
    aria-label={ariaLabel}
    onMouseEnter={this.show}
    onMouseLeave={this.hide}
    onClick={this.onClick}
    tabIndex={0}
    onKeyDown={this.handleKeyDown}
>
    {extraText ? <div>{textprop}<span className='MenuItem__help-text'>{extraText}</span></div> : textProp}
    {selectedValueText && <span className='selected'>{selectedValueText}</span>}
    {id !== 'SidebarChannelMenu-moveToDivider' &&
        <span
            className={classNames([`fa fa-angle-right SubMenu__icon-right${hasSubmenu ? '' : '-empty'}`, {mobile: isMobile}])}
        />}
    {subMenuContent}
</div>
```

But have the `div` a column character.

michelengelen (changes requested):

michelengelen:
@matthewbirtch I did add the requested changes ... its ready for re-review now! thanks!

matthewbirtch (changes requested):
All the noted issues look to be resolved. However, I noticed a new one:

The Custom Status menu item looks like it vertically is misaligned and is taller than the other menu items for some reason. Could you look in to that?

image

michelengelen:
@matthewbirtch I will have a look tomorrow ... I just noticed, that the structure for the custom status is done differently from the other MenuItems, so I will have to adjust that to align this with the rest of the items. 

That is if @nevyangelova does not want to continue on this! 😸

nevyangelova:
@michelengelen i can resolve this one 👍

matthewbirtch:
oh sorry...I think I was confused about who's PR it was, sorry @nevyangelova. Thanks for taking this!

michelengelen:
@matthewbirtch I just took over the PR for the time @nevyangelova was on PTO

matthewbirtch (changes requested):
@nevyangelova the pulsing dot on the custom status menu item still looks slightly misaligned. It's sitting a bit high still. Can we push it down 2-3px?

michelengelen:
@nevyangelova I still think, since the anatomy of the items is 99% the same, we could nearly copy/paste all of the structure on top of the custom status menu item.

nevyangelova:
Apologies @matthewbirtch I didn't have the dot locally so I missed it. Fixed now. @michelengelen it just needed the align-items prop.

michelengelen (approved):
LGTM

michelengelen (approved):

nevyangelova:
@matthewbirtch the test server is fixed now

matthewbirtch:
@nevyangelova i just realized that the status colors don't seem to be using the theme colors for these statuses. 

For example, the pulsing dot is using the `--online-indicator` theme variable, but the checkmark next to "online" is a different green and doesn't seem to be using the theme colors? 

Can you take a look at these for me? (and make sure the status indicator in the avatar matches this as well?

michelengelen:
@matthewbirtch sry to butt in here.
The status icons are using the theming variables from compass components and the indicator is using the "old" variable. We should probably provide a new component for that, wdyt @nevyangelova ?

nevyangelova:
@michelengelen not sure i understand about the new component? Shouldn't it suffice to change the pulsating dot color?

michelengelen:
> @michelengelen not sure i understand about the new component? Shouldn't it suffice to change the pulsating dot colour? @matthewbirtch in one of the feedback points in this PR there was the request to change the status icons to use the new compass font + colour. Let me know how to proceed with the pulsating dot :)

Yeah, that would be sufficient as well, but not consistent. It is just a consideration for the future. 😅

matthewbirtch:
> @michelengelen not sure i understand about the new component? Shouldn't it suffice to change the pulsating dot colour? @matthewbirtch in one of the feedback points in this PR there was the request to change the status icons to use the new compass font + colour. Let me know how to proceed with the pulsating dot :)

I'm confused too. The status colors should all be using theme colors and it looks like they are not. If I customize my theme colors, only the pulsing dot changes to the custom theme colors.

michelengelen:
> > @michelengelen not sure i understand about the new component? Shouldn't it suffice to change the pulsating dot colour? @matthewbirtch in one of the feedback points in this PR there was the request to change the status icons to use the new compass font + colour. Let me know how to proceed with the pulsating dot :)
> 
> 
> 
> I'm confused too. The status colors should all be using theme colors and it looks like they are not. If I customize my theme colors, only the pulsing dot changes to the custom theme colors.

In that case we might have a theming issue here. Maybe we should consult @deanwhillier, since he built the theming port for the compass components.

michelengelen:
@nevyangelova and @matthewbirtch I found the cause for the color issue we ran into. the `CompassThemeProvider` currently does not set custom colors from the "theming bridge":

```typescript
useEffect(() => {
    setCompassTheme({
        ...compassTheme,
        palette: {
            ...compassTheme.palette,
            primary: {
                ...compassTheme.palette.primary,
                main: theme.sidebarHeaderBg,
                contrast: theme.sidebarHeaderTextColor,
            },
            alert: {
                ...compassTheme.palette.alert,
                main: theme.dndIndicator,
            },
        },
        action: {
            ...compassTheme.action,
            hover: theme.sidebarHeaderTextColor,
            disabled: theme.sidebarHeaderTextColor,
        },
        // here we should have a property called "badges" where the colors for the status badges are defined
        text: {
            ...compassTheme.text,
            primary: theme.sidebarHeaderTextColor,
        },
    });
}, [theme]);
```

nevyangelova:
How should we handle this regarding this PR? Seems like we need to add this to compass and update? Should we bring back the old status icons until this point?

michelengelen:
> How should we handle this regarding this PR? Seems like we need to add this to compass and update? Should we bring back the old status icons until this point? 

No need for an update. We just need to map the values set via the "old" theming to the new badges property. This is done in the webapp

nevyangelova:
Thanks @michelengelen and @matthewbirtch ! i have added the theme icons but I am not sure they are mapped to what its supposed to be, because when I switch between themes they stay the same. Let me know if this is what we are trying to achieve :)

matthewbirtch (approved):
@nevyangelova the theme color appears to be working as expected now. And it updates when I customize the status colors. Thanks!

nevyangelova:
@jgilliam17 @deanwhillier kind reminder for a review.

deanwhillier:
@nevyangelova, @michelengelen, the theme integration with the webapp was only the minimum needed to support the components we had agreed to use for the Global Header as we still need to sort out the best way to properly handle the integration as there are currently some challenges that we can discuss later. Glad you found what you needed for now.

deanwhillier (changes requested):
Looking good @nevyangelova! Apologies for the delay, I've been working on the review for a few days between interruptions as I wanted to make sure I gave it the attention it is due!

There are a few items below for consideration and the following thoughts as we work to remove the Global Header feature flag (a.k.a. not necessarily in this PR).

We have been referring to this menu as the status menu and while that is one of the primary functions the menu still provides, it's actually now more of a user account menu with added functionality beyond just user status. I mention this as it's important for the following:

Semantically and structurally, I think we should consider:

- We might want to update the parent component name to reflect the change to being more of a "User Account" menu and not just the status menu as it was previously.
- We might want to move this component into the Global Header component folder along with the other new Global Header child components
- The button that opens/closes the user account menu should wrap the user avatar as well and the button's aria-label should be updated to reflect that is is more than just about user status. 
- When set, the clickable custom status icon that shows up next to the user account menu button should also be a button with an `aria-label` set and not just a `<span>` for accessibility reasons. 
   - Aside: it would be nice to update the components that generate this but that's lower priority.

I'm thinking instead of the current HTML structure:

```html
<div class="MenuWrapper status-dropdown-menu status-dropdown-menu-global-header">
    <div class="status-wrapper">
        <span>
            <span class="emoticon" alt=":hamburger:" data-emoticon="hamburger" style="background-image: url(&quot;http://localhost:8065/static/emoji/1f354.png&quot;); background-size: 16px; height: 16px; width: 16px; max-height: 16px; max-width: 16px; min-height: 16px; min-width: 16px; overflow: hidden; margin-right: 6px;"></span>
        </span>
        <img class="Avatar Avatar-sm" alt="user profile image" src="http://localhost:8065/api/v4/users/4wkced5b9fdw7f37gxpszknnfw/image">
        <button class="status style--none" aria-label="set status">
            <i size="12" color="inherit" class="sc-dlnjwi eTVAHO sc-hKFxyN liZfdD icon-check-circle"></i>
        </button>
        <span class="status status-edit edit"></span>
    </div>
</div>
```

Possibly something more along these lines of:

```html
<div class="MenuWrapper UserAccountMenu">
    <div class="UserAccountMenu__wrapper">
        <button class="style--none" aria-label="update custom status">
            <span class="emoticon" alt=":hamburger:" data-emoticon="hamburger" style="background-image: url(&quot;http://localhost:8065/static/emoji/1f354.png&quot;); background-size: 16px; height: 16px; width: 16px; max-height: 16px; max-width: 16px; min-height: 16px; min-width: 16px; overflow: hidden; margin-right: 6px;"></span>
        </span>
        <button class="style--none" aria-label="toggle user account menu to set status, change account settings and log out">
            <img class="Avatar Avatar-sm" alt="user profile image" src="http://localhost:8065/api/v4/users/4wkced5b9fdw7f37gxpszknnfw/image">
            <i size="12" color="inherit" class="sc-dlnjwi eTVAHO sc-hKFxyN liZfdD icon-check-circle"></i>
        </button>
    </div>
</div>
```

components/global/global_header.scss
```diff
@@ -197,12 +213,57 @@
             padding: 0 16px;
         }
 
+        .MenuItem > button,
+        #status-menu-dnd-timed {
+            padding: 1px 16px;
+
+            .MenuItem__primary-text {
+                display: grid;
```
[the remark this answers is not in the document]
michelengelen: thats my doing ... it was way easier to align the items with grid.

nevyangelova:
> There are a few items below for consideration and the following thoughts as we work to remove the Global Header feature flag (a.k.a. not necessarily in this PR).
> 
> We have been referring to this menu as the status menu and while that is one of the primary functions the menu still provides, it's actually now more of a user account menu with added functionality beyond just user status. I mention this as it's important for the following:
> 
> Semantically and structurally, I think we should consider:
> 
> * We might want to update the parent component name to reflect the change to being more of a "User Account" menu and not just the status menu as it was previously.
> * We might want to move this component into the Global Header component folder along with the other new Global Header child components
> * The button that opens/closes the user account menu should wrap the user avatar as well and the button's aria-label should be updated to reflect that is is more than just about user status.
> * When set, the clickable custom status icon that shows up next to the user account menu button should also be a button with an `aria-label` set and not just a `<span>` for accessibility reasons.


That was my thinking as well, I think this can be a part of the clean up task and possibly linked to https://mattermost.atlassian.net/browse/MM-37792

jgilliam17:
@nevyangelova Can you please take a look at the "roles" error. Michel mentioned rebase should fix it. Thanks.

nevyangelova:
@jgilliam17 updated

andrewbrown00 (approved):
Changes look good 🙂

components/global/global_header.scss
```diff
@@ -160,27 +160,43 @@
...
+                    linear-gradient(
+                        0deg,
+                        rgba(var(--global-header-text-rgb), 0.08),
+                        rgba(var(--global-header-text-rgb), 0.08)
+                    );
             }
         }
 
         &.active {
             background: var(--sidebar-text-16);
```
deanwhillier, line 173: @nevyangelova, apologies, but I noticed one more variable that should probably be chaned while we are "under the hood".

```suggestion
            background:rgba(var(--global-header-text-rgb), 0.16);
```

deanwhillier (approved):
LGTM with one additional comment.

nevyangelova:
nice catch @deanwhillier thank you

jgilliam17:
Thanks @nevyangelova 
Looks great! Requesting one small tweak for the `@username` part of the profile/status menu as discussed with @matthewbirtch 
- If it is only showing the username, the username should show at full opacity
Screen Shot 2021-09-07 at 3 15 59 PM

nevyangelova:
@jgilliam17 let's handle this separately as it wasn't in the points listed above and its getting a bit hard to re-review :) https://mattermost.atlassian.net/browse/MM-38421

PR #8645, opened 2021-08-19
MM-37821 - remove global header option from user settings

michelengelen:
#### Summary
removes the global header option from the user settings modal. Also changes the `getGlobalHeaderEnabled` selector to only use the feature flag and not the user-preferences object anymore.

replacement for #8583 since the old PR did run into severe Spinwick issues

#### Ticket Link
[MM-37821](https://mattermost.atlassian.net/browse/MM-37821

#### Related Pull Requests
Once this successfully runs a spinwick server for testing the old PR #8583 will be closed in favor of this one.

#### Release Note
```release-note
NONE
```

PR #8671, opened 2021-08-20
[MM-37365] Reposition the RHS when Global Header is enabled

deanwhillier:
#### Summary
This PR repositions, resizes and re-layers the RHS when the Global Header (GH) is enabled, taking into account changes made at various screen-width breakpoints in the existing webapp.

This PR also removes `isMobile()` consideration from the `getGlobalHeaderEnabled` selector as that method indiscriminately applies to tablets as well. The requirement for removing the GH from mobile refers to the webapp running on screens that are 768px or less that use the iPhone look and feel. Until/if we implement a better redux-enabled solution (as suggested by @hmhealey for the future) for determining whether a device is mobile according to our criteria, our best bet is to use CSS media queries to accomplish the requirements for removing the GH from mobile. This was the existing behaviour prior to the GH project.

Finally, this PR also stops considering whether the GH is enabled or not when rendering the RHS search container as it needs to be present when a browser window resizes from desktop width to mobile width (when a React re-render is not triggered) in order to not break the mobile view, which requires the RHS search container. Without it, the user will get stuck with the RHS open and no way to close it without reloading.

Update: See conversation below for CSS vs JS handling of showing/hiding the GH for desktop/mobile screen widths: https://github.com/mattermost/mattermost-webapp/pull/8671#discussion_r694083573

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37365

#### Related Pull Requests
- https://github.com/mattermost/mattermost-webapp/pull/8645 – manually applies changes from here, I will correct any conflicts that may arrise
- https://github.com/mattermost/mattermost-webapp/pull/8561 – changes from the `isMobile` approach to a media query approach

#### Screenshots
**Larger Desktop (> 1680px)**:
(RHS is 500px wide)
image
image

**Desktop (> 960px)**:
(RHS is 400px wide)
image
image
image

**Tablet (<= 960px)**:
(RHS is 400px wide)
image
image

**Mobile (<=768px)**:
(RHS is 100% wide)
image

**Phone (<=375px)**:
(RHS is 100% wide)
image

#### Release Note

```release-note
NONE
```


components/global/global_header.scss
```diff
@@ -7,6 +7,12 @@
     --global-header-background-rgb: var(--sidebar-teambar-bg-rgb);
     --global-header-text-rgb: var(--sidebar-header-text-color-rgb);
 
+    @media screen and (max-width: 768px) {
```
michelengelen, line 10: hiding the global header VS not even rendering it is a discussion that could be held here. IMO the "not rendering" (aka `Utils.isMobile() === true`) approach is the better solution, since renderin it in the first place will not only render the global header itself, but all of its subcomponents as well. Correct me if I am wrong, but this can have an impact on performance.

components/global/global_header.scss
```diff
@@ -47,6 +53,43 @@
...
+                }
+            }
+        }
+
+        // adjust RHS position and height
+        #sidebar-right {
+            z-index: 20;
+            top: 0;
+
+            .sidebar-right-container {
```
michelengelen, line 73: nit (and probably out of scope for this ticket): for consistency-purposes this should be named `.sidebar-right__container`

michelengelen (commented):
3/5 we should not use CSS hiding of the global header and instead refrain from rendering it in the first place.

Thoughts @deanwhillier ?

components/global/global_header.scss
michelengelen: as to your reasoning in the PR description why this should render nevertheless: It is correct, but the check could also be bound to the resizeEvent, so that it is performed (maybe in conjunction with a debounce to not have it hit performance). That way we can have the best from both approaches! :D

deanwhillier: I agree it's not ultimately the best approach. The challenge is that as is, the `const isMobileView = UserAgent.isMobile() || Utils.isMobile();` check causes the following bugs:

1. iOS/Android Tablets with a screen width `>768px` won't get the GH experience when they should be
2. Resizing the screen from `>768px` to `<=768px` or the other way doesn't properly re-render the UI creating user experience issues that in some cases can only be recovered from by reloading.

**1** could be partially corrected by just using `Utils.isMoblie()` and not `UserAgent.isMobile()`, however window resizing would still likely have issues in the current implementation.
**2** and the resizing issue of **1** could be fixed by implementing redux-enabled functionality (or similar) that properly monitors screen resizing and updates a new `isMobile()` setting that could replace/augment the existing isMobile checks

Considerations:
1. Not showing GH on screen sizes `<=768px` is temporary
2. Implementing a redux solution (or similar) will require additional effort to build, review and test and generally make sure the overall experience is as expected.
3. The markup for the initial state of the GH is minimal and with `display: none` isn't processed by the browser during rendering
4. We have very limited time left
5. CSS is a much simpler approach for the short term

If you'd like to tackle an updated responsive JS solution on top of your [previous PR](https://github.com/mattermost/mattermost-webapp/pull/8561) that handles the above bugs in the time we have left (keeping in mind PR reviews and QA), then I'm happy to remove the CSS solution in this ticket. Let me know. :)

components/global/global_header.scss
deanwhillier: Yup. Not really in scope for this PR though as it's an existing class on an existing element. :)

michelengelen (approved):

components/global/global_header.scss
```diff
@@ -47,6 +53,43 @@
...
+                height: 100%;
+            }
+
+            .sidebar--right__header {
+                flex-basis: 63px;
```
michelengelen, line 78: I am not 100% sure, but I know that some browsers have different default values for flex. It would probably be a good idea (to prevent issues) to set all values for the flex property.

michelengelen: And isn't 63px restricting the width as well? Just wondering.

michelengelen (approved):
LGTM ... just some thoughts to consider later on

components/global/global_header.scss
deanwhillier: Good thoughts! I don't generally use `flex-basis`, but in this case needed to override an existing applied style of `flex: 0 0 56px;` and so it should be fine to use `flex-basis: 64px;` directly. `flex-basis` controls the size in the defined `flex-direction` and not both directions so I think we're good.

Although it's not very intuitive by name (IMO), I guess `flex-basis` is helpful for situations where a `flex` container could use either `flex-direction: row` or `flex-direction: column` and where you need the same width or height respectively for either direction. I would normally use width/height otherwise.

michelengelen: yeah ... flex can be qwirky at times. All good ... those were just thoughts on possible problems anyways.

michelengelen:
removing @hmhealey (he is OoO for some days) as reviewer and adding @crspeller instead

esethna (approved):

deanwhillier:
Added @michelengelen's fix for the white screen bug from here: https://github.com/mattermost/mattermost-webapp/pull/8682

matthewbirtch (approved):
Nice work @deanwhillier. All looks good to me

jgilliam17 (approved):
Thanks @deanwhillier 
Tested, looks good to merge.
- Verified RHS positioning with Global Header enabled.

hmhealey (approved):

michelengelen:
@crspeller kind reminder to review.

crspeller (approved):
Awesome! This makes a lot more sense.

PR #8702, opened 2021-08-24
MM-38121 Do not overwrite current team's thread mention count on websocket reconnect

jwilander:
#### Summary
Do not overwrite current team's thread mention count on websocket reconnect and fetch threads again.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38121

#### Release Note

```release-note
Fix re-connecting to the websocket causing thread mentions to be cleared in the UI with collapsed reply threads enabled.
```

### DIFF
--- a/actions/websocket_actions.jsx
+++ b/actions/websocket_actions.jsx
@@ -35,6 +35,7 @@ import {isCollapsedThreadsEnabled} from 'mattermost-redux/selectors/entities/pre
 import {getThread, getThreads} from 'mattermost-redux/selectors/entities/threads';
 import {
     getThread as fetchThread,
+    getThreads as fetchThreads,
     handleAllMarkedRead,
     handleReadChanged,
     handleFollowChanged,
@@ -164,6 +165,9 @@ function reconnectWebSocket() {
     initialize();
 }
 
+window.closeWS = close;
+window.connectWS = reconnect;
+
 const pluginReconnectHandlers = {};
 
 export function registerPluginReconnectHandler(pluginId, handler) {
@@ -195,6 +199,7 @@ export function reconnect(includeWebSocket = true) {
     const state = getState();
     const currentTeamId = state.entities.teams.currentTeamId;
     if (currentTeamId) {
+        const currentUserId = getCurrentUserId(state);
         const currentChannelId = getCurrentChannelId(state);
         const mostRecentId = getMostRecentPostIdInChannel(state, currentChannelId);
         const mostRecentPost = getPost(state, mostRecentId);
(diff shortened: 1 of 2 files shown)


agnivade:
Sorry I don't have much context here. But just wanted to mention that - once you have reliable websockets enabled, from an application point of view, there is no reconnect. Everything is abstracted to a single unbroken event stream. Maybe not now, but I think we should remove the reconnect handler altogether.

jwilander:
@agnivade that's a good discussion we should have with the web platform team. If we can remove some (or all) of this reconnect code that would be awesome and reduce complexity and network usage on the clients a lot.

Do the reliable websockets persist across server restart too? A lot of when this specific issue was being seen was on daily was when it was getting new code pushes hourly

agnivade:
> Do the reliable websockets persist across server restart too? A lot of when this specific issue was being seen was on daily was when it was getting new code pushes hourly

Ah I see. No it doesn't. Yeah, the hourly pushes seem to be a special case.

packages/mattermost-redux/src/actions/teams.ts
```diff
@@ -76,14 +76,39 @@ export function getMyTeams(): ActionFunc {
...
+        }
+
+        if (skipCurrentTeam) {
+            const currentTeamId = getCurrentTeamId(getState());
+            if (currentTeamId) {
+                const index = unreads.findIndex((member) => member.team_id === currentTeamId);
+                if (index >= 0) {
+                    unreads.splice(index, 1);
+                }
+            }
```
Willyfrog, line 100: I'm wondering if this should instead happen on the reducer instead of here, as the reducer is the one in charge of merging the data. 0/5 though.

actions/websocket_actions.jsx
```diff
@@ -164,6 +165,9 @@ function reconnectWebSocket() {
     initialize();
 }
 
+window.closeWS = close;
+window.connectWS = reconnect;
```
Willyfrog, line 169: why do we need to make this available? I don't see it being used in the rest of the PR

Willyfrog (approved):
looks good, but added a couple of comments. approving as it could be merged as it is

actions/websocket_actions.jsx
jwilander: Sorry, bad debug commit. I'll remove.

packages/mattermost-redux/src/actions/teams.ts
jwilander: Yeah, I struggled with where to put but I landed on keeping it here since the comment I wrote has all the context for it and it means it's less spaghetti'ed throughout different files

Willyfrog: yeah, as I said 0/5.

jwilander:
@BenCookie95 mind giving this a review when you get a chance?

BenCookie95 (approved):

BenCookie95:
> @BenCookie95 mind giving this a review when you get a chance?

Sorry was on PTO for the last 2 weeks, LGTM!

jwilander:
Ah, no worries - thanks!

PR #8704, opened 2021-08-24
For select and input fields, if the label is empty, don't add the label element. This allows compacting the fields vertically

jfrerich:
#### Summary
This PR resolves several Jira Tickets.  The three ticket solutions are in separate commits and all three differences can be seen between the two attached images.

#### Ticket Link
~~https://mattermost.atlassian.net/browse/MM-38123 - Add option to hide required star~~ (The star is removed in this PR
https://mattermost.atlassian.net/browse/MM-38126 - Add required field message at the bottom of the modal
https://mattermost.atlassian.net/browse/MM-38128 - Spacing between fields is too large (If a label isn't provided, don't add the label field some the space is reduces between fields)

#### Related Pull Requests
https://github.com/mattermost/mattermost-plugin-apps/pull/244 - Add `hide_required_star` to `Field` struct

#### Screenshots
**Before:** 

**After: (This is the final implementation in the is PR)**

**~~After~~: (After UX discussion, this was decided against and the previous image was implemented**

#### Release Note
```release-note
Added ability to hide App Field required star
```


components/apps_form/apps_form_component.tsx
```diff
@@ -159,6 +159,7 @@ export class AppsForm extends React.PureComponent<Props, State> {
 
         this.setState({fieldErrors});
         if (Object.keys(fieldErrors).length !== 0) {
+            this.setState({formError: 'Please fix all field errors'});
```
catalintomai, line 162: does this get localized?

components/widgets/settings/setting.tsx
```diff
@@ -22,18 +22,25 @@ const Setting: React.FC<Props> = ({
...
             >
                 {label}
             </label>
+        );
+    }
+
+    return (
+        <div
+            data-testid={inputId}
+            className='form-group'
```
catalintomai, line 41: nit: would move this as the first prop of the div (to have same ordering as in prev. file - maintaining a consistent ordering of prop types for div, etc. makes spotting a missing property easier).

catalintomai (approved):

components/apps_form/apps_form_component.tsx
jfrerich: It can be.  I added localization for the error.

components/widgets/settings/setting.tsx
jfrerich: @catalintomai, I'm sorry, but don't see what exactly you are asking. Are you suggesting to swap the ordering of `className` and `data-testid`?

components/apps_form/apps_form_component.tsx
jfrerich: done.

larkox (approved):
LGTM. Has this gone through UX? I want to make sure they are OK with it.

larkox:
Also, do you have within your plans doing the mobile version of this change?

mickmister:
@larkox @jfrerich It seems the unwanted behavior occurs specifically when the label of the field is empty. Notice in the mobile version, the required star is not shown for fields that don't have a label:



Maybe the webapp should just follow suit with the mobile app, and we shouldn't go down the path of introducing the `hide_required_star` property. @matthewbirtch Do you have any thoughts on this? See the PR description for the before/after of having this new change enabled.

larkox:
@mickmister I agree we should be consistent between mobile and web. Whether to hide the optional or required star if there is no label, I have mixed feelings, but I will defer to UX.

jfrerich:
@mickmister, I like the look of the mobile screenshot you shared.  I think that looks clean and better, but some developers might prefer the required star for every required field.

larkox:
@matthewbirtch That solves one part, but the problem is "still the same". Imagine we have a block of inputs like that without label, but the three of them are optional. Should we show `(optional)` for each of them? Should we allow the app developer to hide the "optional" label if he finds it necessary? Or we should automatically hide the "optional" label if there is no label?

matthewbirtch:
> @matthewbirtch That solves one part, but the problem is "still the same". Imagine we have a block of inputs like that without label, but the three of them are optional. Should we show `(optional)` for each of them? Should we allow the app developer to hide the "optional" label if he finds it necessary? Or we should automatically hide the "optional" label if there is no label?

Shouldn't every field have a label though? Or at the very least every input field should have an initial placeholder (which could show 'optional' in the placeholder if we really want to offer fields without labels

matthewbirtch (approved):
Verified that the asterisks are now gone and we're using `(Optional)` to denote the optional fields.

As discussed, there are other issues with the form from a UX perspective, but we'll look to address some of those beyond this PR.

Rina-dsg (approved):

jfrerich:
@larkox, I update the code a bit from the last review and added a test, so I requested another review.

larkox (approved):

DHaussermann (approved):
Tested and passed
I deployed this locally and used a recent build of ZenDesk to validate. 
I see both changes as expected as Jason's screenshot. 

LGTM!

DHaussermann:
@jfrerich This slipped off my radar and the UX was approved months ago. 
Is there any reason why this should not be merged now?

If not, Please merge.

PR #8760, opened 2021-08-31
[MM-38241] Use DOMParser to parse tables from clipboard

streamer45:
#### Summary

We start making use of [`DOMParser`](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) to parse html tables from clipboard.  
I am assuming creating a parser is an efficient operation. Let me know if we should cache it or something like that.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38241

#### Release Note

```release-note
NONE
```

### DIFF
--- a/e2e/cypress/integration/messaging/post_html_table_spec.js
+++ b/e2e/cypress/integration/messaging/post_html_table_spec.js
@@ -1,0 +1,36 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Group: @messaging
+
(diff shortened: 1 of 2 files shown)


srkgupta (approved):
Thanks @streamer45. The changes LGTM. Can you please add one test for this, if possible.

streamer45:
> Thanks @streamer45. The changes LGTM. Can you please add one test for this, if possible.

The best I could come up with is an E2E test to check if the security violation gets triggered. Let me know if that works.

BenCookie95 (approved):

hmhealey (approved):

srkgupta:
> The best I could come up with is an E2E test to check if the security violation gets triggered. Let me know if that works.

That's awesome. Thanks @streamer45. The E2E tests looks great and works fine 👍

amyblais:
@streamer45 Would this be a good cherry-pick candidate for v5.39 or is it risky?

streamer45:
> @streamer45 Would this be a good cherry-pick candidate for v5.39 or is it risky?

I don't think it's risky so we could do that. Should I cherry-pick to 6.0 as well?

PR #8765, opened 2021-08-31
MM-38193 Add missing i18n strings from MoreChannels modal

hmhealey:
These strings weren't detected by `make i18n-extract` before, so they weren't showing up in the translation tools.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38193

#### Release Note
```release-note
NONE
```


components/searchable_channel_list.jsx
```diff
@@ -115,7 +115,7 @@ export default class SearchableChannelList extends React.PureComponent {
                             text={localizeMessage('more_channels.joining', 'Joining...')}
                         >
                             <FormattedMessage
-                                id={shouldShowArchivedChannels ? 'more_channels.view' : 'more_channels.join'}
+                                id={shouldShowArchivedChannels ? t('more_channels.view') : t('more_channels.join')}
```
koox00, line 118: I'm just curious, what's the difference with using `t`?

koox00 (approved):

components/searchable_channel_list.jsx
hmhealey: `make i18n-extract` identifies translatable strings in a few different ways like if they're the `id` of a `FormattedMessage` or if they're the first argument to `localizeMessage`. It can only do that if they're constants though, so it can't detect them in this case.

`t` doesn't do anything on its own (it just returns its argument), but it is detected by `make i18n-extract`, so it works as a marker for anything not automatically detected by `i18n-extract`.

koox00: Oh I see, thanks for the clarification!!

amyblais:
@cpoile Kind reminder for PR review.

cpoile (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified i18n string for "Join" has been added.

PR #8781, opened 2021-09-03
[MM-37897] - added fading to the "new channel category dropzone box"

michelengelen:
#### Summary
this PR adds a fading effect on the "new channel category dropzone box" for a better draggable drop indication.
I will add a short clip of the new fade to this PR

#### Ticket Link
[MM-37897](https://mattermost.atlassian.net/browse/MM-37897)

#### Screenshots
https://user-images.githubusercontent.com/32863416/131969038-3a645296-4d32-407e-ba65-5cc4176f8202.mov

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/sidebar/sidebar_category/sidebar_category.tsx
+++ b/components/sidebar/sidebar_category/sidebar_category.tsx
@@ -201,7 +201,7 @@ export default class SidebarCategory extends React.PureComponent<Props, State> {
                 <div className='SidebarCategory_newDropBox'>
                     <div
                         ref={this.newDropBoxRef}
-                        className={classNames('SidebarCategory_newDropBox-content animating', {
+                        className={classNames('SidebarCategory_newDropBox-content', {
                             collapsed: category.collapsed || (draggingState.type === DraggingStateTypes.CATEGORY && draggingState.id === category.id),
                             isDraggingOver,
                         })}
--- a/components/sidebar/sidebar_category/sidebar_category.tsx
+++ b/components/sidebar/sidebar_category/sidebar_category.tsx
@@ -201,7 +201,7 @@ export default class SidebarCategory extends React.PureComponent<Props, State> {
                 <div className='SidebarCategory_newDropBox'>
                     <div
                         ref={this.newDropBoxRef}
-                        className={classNames('SidebarCategory_newDropBox-content', {
+                        className={classNames('SidebarCategory_newDropBox-content animating', {
                             collapsed: category.collapsed || (draggingState.type === DraggingStateTypes.CATEGORY && draggingState.id === category.id),
                             isDraggingOver,
                         })}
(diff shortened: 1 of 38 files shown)


matthewbirtch (approved):
Nice @michelengelen!

components/sidebar/sidebar_category/sidebar_category.tsx
```diff
@@ -201,7 +201,7 @@ export default class SidebarCategory extends React.PureComponent<Props, State> {
                 <div className='SidebarCategory_newDropBox'>
                     <div
                         ref={this.newDropBoxRef}
-                        className={classNames('SidebarCategory_newDropBox-content', {
+                        className={classNames('SidebarCategory_newDropBox-content animating', {
```
deanwhillier, line 204: `animating` as a verb implies that something is currently animating ... wonder if we should change to `animates` or `animated` to indicate that the element can/will animate ... or maybe no additional class is really needed at all?

If you do actually want to use `animating` as the class name, maybe only add it when the animation needs to happen and then remove once done as a number of popular libraries do for this kind of thing.

deanwhillier (changes requested):
LGTM with one consideration.

components/sidebar/sidebar_category/sidebar_category.tsx
michelengelen: the `animating` was included even before I did my changes. I simple added it .. Although I just checked (and removed the hard coded class): It is actually added and removed automatically and I just did not notice that beforehand. 

Thanks for pointing that out! 👍🏼

deanwhillier: Ah, lol, sounds good! 😄

deanwhillier (approved):

michelengelen:
@nevyangelova and @jgilliam17 kind reminder to review

nevyangelova (approved):

jgilliam17:
@michelengelen Test server is showing the "roles" error can you please take a look? Thanks

michelengelen:
@jgilliam17 I fixed the issue for now, but we definitely need to look into this. A lot of the requests are failing when you enter the server and you have no active session.

@spirosoik and @hmhealey this is happening within redux ... i couldn't really figure out the root cause for this (IMO the failing requests are a big indicator :) ), but why are the APIs returning a HTTP status 401 in the first place? The selector `getCurrentUser` relies on the `state.entities.users.profile` object to be filled, but when there is no userProfile to be found it just returns `undefined`, which then breaks the application.

Maybe it fails at the `getCurrentUserId` selector already. I could not check, since it is working fine locally.

Attached is a screenshot of the console with the 401 errors:
Screenshot 2021-09-06 at 16 32 40

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
- Verified fading of the drag indicator when dropping a channel into new category.

PR #8784, opened 2021-09-03
Open forms in bindings on all locations

larkox:
#### Summary
Previously, bindings in other locations than commands will always submit, ignoring fields like "form". Now, if a binding has a form, it will automatically open the app form without communicating with the app.

#### Related PRs
Mobile: https://github.com/mattermost/mattermost-mobile/pull/5665

#### Ticket Link
None

#### Release Note
```release-note
NONE
```


catalintomai (approved):

components/dot_menu/index.ts
```diff
@@ -138,11 +140,12 @@ type Actions = {
     doAppCall: DoAppCall;
     postEphemeralCallResponseForPost: PostEphemeralCallResponseForPost;
     setThreadFollow: (userId: string, teamId: string, threadId: string, newState: boolean) => void;
+    openAppsModal: (form: AppForm, call: AppCallRequest) => void;
 }
 
 function mapDispatchToProps(dispatch: Dispatch<GenericAction>) {
     return {
-        actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc>, Actions>({
+        actions: bindActionCreators<ActionCreatorsMapObject<any>, Actions>({
```
jfrerich, line 148: why did the concrete type get replace with `any`?

jfrerich (approved):
LGTM!  Just one question regarding the replacement of the concrete type with `any`

components/dot_menu/index.ts
larkox: I did it to simplify the typing, since I feel it does not add much value. Nevertheless, I could have changed it to ActionFunc|Action. Do you prefer it that way?

larkox:
@DHaussermann To test this, the latest version of the test app (yesterday's) has a post option and a channel header option called "formInBinding". Before this change, it will only show an OK ephemeral. After this change, when you click will open a modal, than after submit will show an OK ephemeral.

DHaussermann (approved):
Tested and passed

The `FormInBindings` now shows the form (with `foo` field)
Submit produces an ephemeral `OK` post
No other functional testing needed. 

LGTM!

PR #8787, opened 2021-09-06
MM-35718: new replies banner for thread viewer

koox00:
#### Summary

When a new reply is posted and the user is viewing the thread a
"new replies" banner will appear at top depending on the visibility of
the new messages line or whether the user has scrolled to the bottom or
not.

Fixes positioning of the floating timestamp in mobile view as well.

PS: The banner has been added to the top of the thread viewer due to limitations on the virtualized list.
If this [MM-37671](https://mattermost.atlassian.net/browse/MM-37671)  gets implemented then we could re-place the banner on top of the textbox.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35718
https://mattermost.atlassian.net/browse/MM-37889

#### Screenshots

Desktop view:
Screenshot 2021-09-06 at 1 30 06 PM

Mobile view:
Screenshot 2021-09-06 at 1 30 06 PM

#### Release Note

```release-note
NONE
```


larkox:
@koox00 Maybe it is something wrong with the screenshots, but there seems the shade of blue of the banner is different between the two screenshots. Is that expected?

koox00:
> @koox00 Maybe it is something wrong with the screenshots, but there seems the shade of blue of the banner is different between the two screenshots. Is that expected?

😮 I didn't notice! I took the first screenshot from Firefox and the second from the OSX screenshot utility, maybe that's why...!

larkox:
@koox00 Just for removing any doubt, could you check in Firefox if making the browser screen really narrow (and therefore forcing the mobile view) shows the same shade of blue?

Also, I wonder if we should inform anyone from UX about the difference, in case they are not aware and if they want to do anything about it.

koox00:
> @koox00 Just for removing any doubt, could you check in Firefox if making the browser screen really narrow (and therefore forcing the mobile view) shows the same shade of blue?

Yep, I checked, the color is consistent the screenshot utilities may differ on the way they screenshot.

koox00:
firefox screenshot:
image

OSX screenshot:
image

components/threading/virtualized_thread_viewer/virtualized_thread_viewer.tsx
```diff
@@ -331,6 +364,34 @@ class ThreadViewerVirtualized extends PureComponent<Props, State> {
         return undefined;
     }
 
+    isNewMessagesVisible = (): boolean => {
+        const {visibleStopIndex, visibleStartIndex} = this.state;
+        const newMessagesSeparatorIndex = getNewMessageIndex(this.props.replyListIds);
+        if (typeof visibleStopIndex === 'number' && typeof visibleStartIndex === 'number') {
```
larkox, line 370: Is this to differentiate 0 from undefined? Can we use `x == null` here?

If you are not aware about the `x == null` thing... http://adripofjavascript.com/blog/drips/equals-equals-null-in-javascript.html

reducers/views/threads.ts
```diff
@@ -47,8 +47,19 @@ export function manuallyUnread(state: ViewsState['threads']['manuallyUnread'] |
     return state;
 }
 
+export function toastStatus(state: ViewsState['threads']['toastStatus'] = false, action: GenericAction) {
```
larkox, line 50: Just curiosity. Any idea why are we using this `ViewsState['threads']['toastStatus']` pattern instead of just `boolean` here?

larkox (approved):
LGTM, just a few nits and questions, but nothing blocking.

reducers/views/threads.ts
koox00: I just used it to be consistent with the above functions, it adds no actual benefit

components/post_view/floating_timestamp/index.ts
```diff
@@ -7,12 +7,14 @@ import {getPost} from 'mattermost-redux/selectors/entities/posts';
...
 
 import {getToastStatus} from 'selectors/views/channel';
+import {getThreadToastStatus} from 'selectors/views/threads';
 import {GlobalState} from 'types/store';
 
 import FloatingTimestamp from './floating_timestamp';
 
 type OwnProps = {
     postId: string;
+    isRhsPost: boolean;
```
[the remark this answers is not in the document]
koox00: The `isRhsPost` was added before CRT, this PR just uses it in `index.ts` but the prop existed before.

matthewbirtch:
Thanks @koox00. Nice to have this feature in. A few things:

1. I know you mentioned the change in having to place this banner at the top. However, it feels really odd to me. Can we position the banner with absolute positioning anchored to the bottom of the RHS? I just tried adding the following to the banner element and it seemed to work:

```
position: absolute;
bottom: 16px;
```

Maybe I'm missing something, but why won't this work? If not, I think we can live with this for now, but it would be nice to look in to this further.

2. Also, the original intent for the design was not to span the full width. Can we change this so it only spans the width it needs to accommodate for the text and icons? This is how it was originally designed:
image

Other than that, I think it's working as expected. I tested it in the RHS view and the Global Threads view and it's behaving properly.

koox00:
@matthewbirtch OH, I didn't do that at first so that smaller threads won't show the banner at bottom, without realizing that smaller threads would never ever show the banner.... 😅

koox00:
@matthewbirtch I've re-positioned the banner at the bottom 😉

matthewbirtch:
> @matthewbirtch I've re-positioned the banner at the bottom 😉

Thanks @koox00! That worked. Can we push the banner up slightly? It's a bit too tight to the bottom I think. Rather than 48px from the bottom, can we make it 56px?

Also, sometimes the banner is showing and other times it isn't. I can't seem to replicate it in a repeatable way though. It seems that if I scroll to a point where I am just 1 or 2 posts above the new message line, I don't get a banner. But if I'm scrolled up higher in the thread, I do get the banner. It's tricky to diagnose. I'm not sure if this is a blocker though - we could maybe move it in to a follow-up ticket?

koox00:
> > @matthewbirtch I've re-positioned the banner at the bottom 😉
> 
> Thanks @koox00! That worked. Can we push the banner up slightly? It's a bit too tight to the bottom I think. Rather than 48px from the bottom, can we make it 56px?

@matthewbirtch yep I made it 56px :)

> Also, sometimes the banner is showing and other times it isn't. I can't seem to replicate it in a repeatable way though. It seems that if I scroll to a point where I am just 1 or 2 posts above the new message line, I don't get a banner. But if I'm scrolled up higher in the thread, I do get the banner. It's tricky to diagnose. I'm not sure if this is a blocker though - we could maybe move it in to a follow-up ticket?

I've modified it slightly so now it should disappear right above the new messages line.
FYI 
- if you view all new messages (scrolled to the bottom) then the banner would not appear again (expect when a new message arrives).
- While scrolling between the new messages line and the bottom the banner does not show.

matthewbirtch:
Thanks @koox00 those look fixed. I discovered a new issue though:
- If I click the banner, it scrolls to the top of the thread rather than the bottom. This seems to be the case in both the RHS and the Global Thread View

components/threading/virtualized_thread_viewer/virtualized_thread_viewer.tsx
larkox: Did you plan to change this?

koox00: Yes I did, the reason is that it showed the banner when scrolled between new messages line and bottom.
It should only show when above new messages line.

components/new_replies_banner/new_replies_banner.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+function NewRepliesBanner({
+    hasNewReplies,
+    onClick,
+    onDismiss,
+    width,
+    actions,
+}: Props) {
+    const onClickMessage = Utils.localizeMessage('postlist.toast.scrollToLatest', 'Jump to new messages');
```
hmhealey, line 29: This should use `useIntl` instead of `Utils.localizeMessage`

hmhealey (changes requested):
One nit to pick. Looks good otherwise.

koox00:
> Thanks @koox00 those look fixed. I discovered a new issue though:
> 
>     * If I click the banner, it scrolls to the top of the thread rather than the bottom. This seems to be the case in both the RHS and the Global Thread View
>       

@matthewbirtch this seems to happen because of the virtualized list.

Explanation when you scroll the list caches each elements height, but If you scroll to fast (seeing blank spaces) it cannot know how far down the items are, so scrolls to top. 
I'll attempt to hack this somehow.

koox00:
> It seems like that resolved it. We should test this more robustly through QA, but it seems like it behaves as expected now. Thanks!

I tried to correct the scroll by scroll the latest/earliest rendered item so it will render the next batch, and so on until it reaches the desired item.. not elegant and I put a circuit break so threads more than 900 messages may fail to scroll to the desired item, but not always, so it's an edge-edge case

hmhealey:
> @hmhealey I really wished we gave `dynamic-virtualized-list` some love.

Yeahhhh, unfortunately Sudheer is the only one who knew how that code worked, so we don't actually have anyone here who knows how to work with it any more. At some point, we'll have to dig back into it, but hopefully we can just replace it with something more robust before that happens.

stylianosrigas:
Removing spinwick for maintenance reasons.Please feel free to apply again if needed.

stylianosrigas:
Removing test server for maintenance purposes. Please feel free to add it again if needed.

jgilliam17:
Thanks @koox00 
Looks good. One issue - small popover is out of view when hovering over `X`. Image shows mobile view, but it's the same in standard browser widths.
I assume it shows `close`, but does it also give an option to use `esc` to close the same way "x new messages" does? Checking here since esc didn't work when dismissing and I'm not sure if it's supposed to or not. Let me know. Thanks.

Screen Shot 2021-10-29 at 1 52 10 PM

koox00:
> Thanks @koox00 Looks good. One issue - small popover is out of view when hovering over `X`. Image shows mobile view, but it's the same in standard browser widths. I assume it shows `close`, but does it also give an option to use `esc` to close the same way "x new messages" does? Checking here since esc didn't work when dismissing and I'm not sure if it's supposed to or not. Let me know. Thanks.
> 
> Screen Shot 2021-10-29 at 1 52 10 PM

thanks @jgilliam17, both tooltip placement, and close on ESC are fixed!

jgilliam17:
Thanks @koox00 
Tooltip on new replies banner looks good and it is oriented the right way, but esc is still not working for me on the test server.
Additionally, the `x new message` banner now takes 2 x `esc` to dismiss.

hmhealey (approved):

koox00:
@jgilliam17 ESC to dismiss should be fixed now

koox00:
>  Additionally, the `x new message` banner now takes 2 x `esc` to dismiss.

This happens on master as well, and it happens when you are already in the channel, you have no unreads and then you receive a new message, can we file a different issue for this?

jgilliam17:
@koox00 here are the [E2E results](https://mattermost-cypress-report.s3.amazonaws.com/54911-82328dc-ee-ent-webapp-pr-8787-mm-ee-test:82328dc/mochawesome.html) Only group config and upload files tests are also failing on master, other 4 appear to be related to this PR. Can you please take a look? Thanks
Results for the [master run](https://mattermost-cypress-report.s3.amazonaws.com/54843-ce-ent-master-mm-ee-test:test/mochawesome.html) for comparison.

jgilliam17:
Here's the latest [E2E report ](https://automation-dashboard.vercel.app/cycles/fe46a635-fbe2-4845-88a0-6ee977113f71)

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified new replies banner on threads.
E2Es report - failures do not appear related, re-tested manually.

PR #8791, opened 2021-09-06
Fixes "property.roles" of undefined

koox00:
#### Summary

A logged out user with global header enabled would crash for the root
URL. This was happening because we assumed current user exists when
querying on whether to show or not the tutorials or getting started
page.

This commit fixes that by defaulting those values to false and
guarding on current user.

#### Release Note

```release-note
NONE
```


streamer45 (approved):
Thanks :+1:

Willyfrog (approved):
LGTM @michelengelen please revert once we have a final fix

lindy65 (approved):
Thanks @koox00 - looks goo to me 👍

PR #8793, opened 2021-09-07
MM-36697: global threads infinite scrolling

koox00:
#### Summary

Adds "infinite" scrolling pagination to global threads list.
The pagination is done by asking threads before a thread id.
To be able to achieve that this introduces separating the unread threads in team, from all threads in team.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36697

#### Release Note
```release-note
NONE
```


packages/mattermost-redux/src/reducers/entities/threads.test.js
```diff
@@ -38,9 +38,40 @@ describe('threads', () => {
...
+            id: 't1',
+        });
+        expect(nextState.unreadThreadsInTeam.a).toContain('t1');
+        expect(nextState.threadsInTeam).toBe(state.threadsInTeam);
+    });
+
     test('ALL_TEAM_THREADS_READ should clear the counts', () => {
         const state = deepFreeze({
             threadsInTeam: {},
+            unreadThreadsInTeam: {},
```
hmhealey, line 74: If you want to save some typing here, I realized recently that you can just call the reducer with an empty action object to initialize the empty state. I think that's what redux does internally.

I did that throughout the channels reducer so that I wouldn't have to change every existing test when I added a new field to the state.

packages/mattermost-redux/src/selectors/entities/threads.ts
```diff
@@ -51,7 +67,10 @@ export function getThreads(state: GlobalState): IDMappedObjects<UserThread> {
 }
 
 export function getThread(state: GlobalState, threadId: $ID<UserThread> | undefined): UserThread | null {
-    if (!threadId || !getThreadsInCurrentTeam(state)?.includes(threadId)) {
+    if (
+        !threadId ||
+        !(getThreadsInCurrentTeam(state)?.includes(threadId) || getUnreadThreadsInCurrentTeam(state)?.includes(threadId))
```
hmhealey, line 72: I'm curious why we check that the thread should be viewable in the current context here. I would've expected this to always return the thread if it's loaded.

hmhealey (approved):
I had a couple random thoughts, but this looks good to me

packages/mattermost-redux/src/selectors/entities/threads.ts
koox00: yep me too. but I was reluctant to remove the check...

calebroseland: Hey @koox00, @hmhealey. A side effect of this selector is used in Global Threads to determine if the URL should be cleared or not. 

e.g. if the given threadId belongs to another team—this returns null, and `useThreadRouting().clear()` gets called.

`getThreadOrSynthetic` is the opposite—give it a fallback post and it'll always return something.

calebroseland (approved):
LGTM 🎉

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified scrolling and loading of threads list.

PR #8796, opened 2021-09-07
Bugfix - roles error with global header enabled

michelengelen:
#### Summary
in the last days we had some problems with our test-servers not showing a login screen because of a javascript error. In the global header we were accessing the currently logged in user (although there was none) and that resulted in breaking the application.

The root cause for this is that the global header is rendering regardless if a user is logged in pr not. Re-arranging the components in `root.jsx` did not really work out due to our current routing structure. 

the solution for now is to prevent rendering of the global header inside the global header component, by utilizing a new hook `useIsLoggedIn`.

this PR also reverts changes made in #8781 and #8791 

#### Ticket Link
```
NONE
```

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/global/global_header.tsx
+++ b/components/global/global_header.tsx
@@ -21,7 +21,7 @@ import SavedPostsButton from './saved_posts_button/saved_posts_button';
 import SettingsButton from './settings_button';
 import SettingsTip from './settings_tip';
 
-import {useCurrentProductId, useProducts, useShowTutorialStep} from './hooks';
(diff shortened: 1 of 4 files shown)


koox00 (approved):
🎉

crspeller (approved):

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
- Verified ability to log in, no "roles" errors.

PR #7778, opened 2021-03-29
[MM-33859] - Transition animation smoothly on long post permalinks

nevyangelova:
#### Summary
This PR introduces the investigation of removing redundant CSS and transitioning long post permalinks smoothly. I believe more CSS can be removed but I would need help with testing it first. So far it works for me.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33895

PR #8797, opened 2021-09-07
[MM-34870] - fixing post-highlighting

michelengelen:
#### Summary
Due to a regression from #7778 the post highlighting for pinned and flagged posts did not work anymore.
This PR re-implements the SCSS class-definition and fixes an error with getting the correct colors from the CSS variables

#### Ticket Link
[MM-34870](https://mattermost.atlassian.net/browse/MM-34870)

#### Screenshots
|  before  |  after  |
|-----|-----|
|Screenshot 2021-09-07 at 16 58 15|Screenshot 2021-09-07 at 16 57 17|

#### Release Note
```release-note
NONE
```


Willyfrog:
Doing the update-branch thing as it is currently showing the roles of undefined error.

matthewbirtch (changes requested):
@michelengelen is this using the same opacity of this theme color that it used to use? It seems lighter to me maybe. I'm wondering if we should bump up the opacity a touch. Maybe to .24 or .32

nevyangelova:
I remember the original ticket was dealing with the transition being laggy. We might need to check that adding this doesn't lag when you click on a post permalink and then viewing that post.

michelengelen:
@matthewbirtch it is using the original value, but we can raise the opacity a bit if that would be better. I just have the feeling it might be a bit too much.

in comparison:
|  0.12 opacity  |  0.24 opacity  |  0.32 opacity  |
|-----|-----|-----|
|Screenshot 2021-09-08 at 11 49 49|Screenshot 2021-09-08 at 12 45 36|Screenshot 2021-09-08 at 12 46 14|

michelengelen:
@matthewbirtch thanks to your consideration I found out about a variable that is being generated, which is more consistent and less prone to future bugs 🙇🏼‍♂️

matthewbirtch:
> @matthewbirtch it is using the original value, but we can raise the opacity a bit if that would be better. I just have the feeling it might be a bit too much.
> 
> in comparison:
> 
> 0.12 opacity	0.24 opacity	0.32 opacity
> Screenshot 2021-09-08 at 11 49 49	Screenshot 2021-09-08 at 12 45 36	Screenshot 2021-09-08 at 12 46 14

Thanks for the comparison @michelengelen. I think I'm inclined to go with .24. On dark themes especially it doesn't show up well as highlighted. However, if this is how it was originally, we can leave this change out and consider for another time. Let me know what you'd prefer to do.

michelengelen:
> Let me know what you'd prefer to do.

Either way is fine with me. Changing it now is only changing it in one place ... so whatever you prefer, basically!

matthewbirtch:
> > Let me know what you'd prefer to do.
> 
> Either way is fine with me. Changing it now is only changing it in one place ... so whatever you prefer, basically!

Ok, let's make the change to .24 if we can.

michelengelen:
@matthewbirtch Done! ✨

nevyangelova (approved):

Willyfrog (approved):
LGTM

matthewbirtch (approved):
Thanks @michelengelen. LGTM

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
- Verified pinned and saved post highlighting.

PR #8800, opened 2021-09-07
Make invite members button always show as lhs button and remove featu…

neallred:
#### Summary
Remove invite members button feature flag.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38258

#### Related Pull Requests
- Has server changes (https://github.com/mattermost/mattermost-server/pull/18350)

#### Screenshots
invite-members-button

#### Release Note
```release-note
Make left hand side button for adding team members the default, removing other A/B test code 
```


package-lock.json
```diff
@@ -10,8 +10,8 @@
       "dependencies": {
         "@formatjs/intl-pluralrules": "4.0.6",
         "@formatjs/intl-relativetimeformat": "6.2.3",
```
pablovelezvidal, line 12: are those package-lock json changes intended?

pablovelezvidal (approved):
a minor comment about what seems to be unintended changes to package-lock file, besides that, LGTM! Thanks @neallred !

package-lock.json
neallred: More or less. They're not an intentional part of this change. It was a natural result of a prior commit/PR bumping deps without updating the lock file (https://github.com/mattermost/mattermost-webapp/commit/3f79e3b270982d7e74889da84d1646bfc7c39e90) . At this point, all users running `make run` will trigger the same lockfile update, so figured it makes the most sense to go ahead and do it in this PR. Happy to revert and let someone else make the change if you think that would be better.

AGMETEOR (approved):
Thanks @neallred . LGTM!

PR #8826, opened 2021-09-10
[MM-38442] - add product switcher menu unit tests

nevyangelova:
#### Summary
This PR adds a unit test file for the Product switcher menu. It only checks for menu items and currently cannot connect it to redux because if we use 'mount' it would fail because its missing the compass theme, which the icon button requires. Which made it impossible to test for the redux modals for example at this point. Perhaps its a good topic for discussion @michelengelen @deanwhillier  ? 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38442

-->
```release-note
NONE
```


components/global/product_switcher_menu/product_switcher_menu.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    test('should match snapshot with plugins', () => {
+        const props = {
+            ...defaultProps,
+            pluginMenuItems: [
+                {id: 'plugin-1', action: jest.fn(), text: 'plugin-1-text', mobileIcon: 'plugin-1-mobile-icon'},
+                {id: 'plugin-2', action: jest.fn(), text: 'plugin-2-text', mobileIcon: 'plugin-2-mobile-icon'},
+            ],
+        };
+        const wrapper = shallowWithIntl(<ProductSwitcherMenu {...props}/>);
+        expect(wrapper).toMatchSnapshot();
```
michelengelen, line 82: can we add click tests here please?

components/global/product_switcher_menu/product_switcher_menu.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+            enablePluginMarketplace: true,
+        };
+        const wrapper = shallowWithIntl(<ProductSwitcherMenu {...props}/>);
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should match snapshot with plugins', () => {
+        const props = {
+            ...defaultProps,
+            pluginMenuItems: [
```
michelengelen, line 76: you did remove this prop from the `Props` type.

michelengelen (commented):
only 2 comments so far. Regarding the "theme testing" we should take this into our next catch-up or triage, WDYT?

deanwhillier:
@nevyangelova, I think you should be able to just import the CompassThemeProvider component to wrap around the component you are testing and pass it one of the default themes defined in `packages/mattermost-redux/src/constants/preferences.ts`.

That said, Michel is correct, we need to revisit the theming integration for Compass Components and the webapp, so might not be worth add theme testing just yet.

deanwhillier (approved):

michelengelen:
@deanwhillier is correct ... it should be possible to just add a ThemeProvider and wrap your component with it ... if you want to test other themes we should probably add some kind of switching helper function and provide some default and custom themes to it.

nevyangelova:
@deanwhillier @michelengelen I tried to test using the compass theme provider but when you wrap in the redux <Provider> you can only test the child inside of it and therefore its not finding any elements within the Menu itself. 
```
 const wrapper = mountWithIntl(
            <Provider store={store}>
                <CompassThemeProvider theme={theme}>
                    <ProductSwitcherMenu {...props}/>
                </CompassThemeProvider>
            </Provider>,
        );
        ``` 
        
        This can only find elements within the first layer so we can't reliably test the children. I propose we have a discussion how to test theme provider further.
        
        It also proves quite difficult to test separate menu items because of the nesting but they seem to have covered the redirect tests in the menu items tests.

**Comment by michelengelen** [comment]
@nevyangelova we can always write a new mounting helper function as with `mountWithIntl` ... it is actually pretty straightforward

File: components/global/product_switcher_menu/product_switcher_menu.test.tsx [hunk 3]
```diff
@@ -0,0 +1,122 @@
...
+        enableIncomingWebhooks: false,
+        enableOAuthServiceProvider: false,
+        enableOutgoingWebhooks: false,
+        canManageSystemBots: false,
+        canManageIntegrations: true,
+        enablePluginMarketplace: false,
+        onClick: () => jest.fn,
+    };
+
+    test('should match snapshot with id', () => {
```
michelengelen, line 44: nit and my personal preference: 
the test case naming is lacking expressiveness when using `should`.

IMO this would be better as: `matches snapshot with id`

michelengelen (approved):
just a nitty comment ... LGTM!

components/global/product_switcher_menu/product_switcher_menu.test.tsx
nevyangelova: Good point, I would stick to the convention in Mattermost for now and maybe we can discuss naming on a later stage.

PR #8828, opened 2021-09-10
MM-37951: follow thread on reply

koox00:
#### Summary

ThreadUpdated event doesn't include `is_following` so when a thread is updated because current user posted a new post
we should set `is_following` to true. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-37951

#### Release Note

```release-note
NONE
```


Willyfrog (approved):

calebroseland (approved):

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified when replying on an explicitly unfollowed thread, the thread is Followed immediately and is not marked as unread for the current user.

PR #8829, opened 2021-09-10
MM-38212: remove read channel from unreads when navigating to threads

koox00:
#### Summary

We include the current channel id in the unreads even if it's read,
I see no reason to include the lastUnreadChannel as well.

Removing it fixes the case when you navigate from an unread channel to
global threads and it didn't remove the channel from the sidebar.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38212

#### Release Note

```release-note
NONE
```


koox00:
@hmhealey the tests failing make no sense to me, maybe you have more insight on this.

hmhealey:
The name of `lastUnreadChannel` is misleading. It tracks whether or not the current channel was unread when you switched to it. It's used so that the current channel remains in the the correct spot in the sidebar's Unreads category (when that's enabled) after you've switched to it and it's marked unread.

I think those tests are failing because it's expecting the current channel to still be sorted with the unreads. Perhaps instead of changing `getUnreadChannels`, we should ensure that `lastUnreadChannel` is cleared when switching to the Threads view?

koox00:
> The name of `lastUnreadChannel` is misleading. It tracks whether or not the current channel was unread when you switched to it. It's used so that the current channel remains in the the correct spot in the sidebar's Unreads category (when that's enabled) after you've switched to it and it's marked unread.
> 
> I think those tests are failing because it's expecting the current channel to still be sorted with the unreads. Perhaps instead of changing `getUnreadChannels`, we should ensure that `lastUnreadChannel` is cleared when switching to the Threads view?

what I don't understand is that we are always adding the current channel:

```js
                if (currentChannelId && unreadChannels.findIndex((channel) => channel.id === currentChannelId) === -1) {
                    unreadChannels.push(allChannels[currentChannelId]);
                }
```

hmhealey:
That's specifically when the user is using the unread filter since the same selector is used for that and for the Unreads category. We do that to avoid getting the user "stuck" when they've read all their channels with that filter enabled.

Screen Shot 2021-09-14 at 10 30 47 AM

It's a bit of a different use case than `lastUnreadChannel` since we want to keep showing the current channel even if it wasn't unread when the user switched to it.

AGMETEOR (approved):

hmhealey (approved):

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified no read channels remain in Unreads section when switching to Threads.

PR #8831, opened 2021-09-10
[MM-38349] - clean-up: typings

michelengelen:
#### Summary
this is a purely technical PR which adds typings and slight improvements on other parts.

#### Ticket Link
[MM-38349](https://mattermost.atlassian.net/browse/MM-38349)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/global/product_switcher.tsx
+++ b/components/global/product_switcher.tsx
@@ -71,6 +71,7 @@ const MenuItemTextContainer = styled.div`
 `;
 
 // TODO: extract this to a separate file
+// (ticket: https://mattermost.atlassian.net/browse/MM-38478)
 const SwitcherNavEntry = ({icon, destination, text, active, onClick}: SwitcherNavEntryProps) => {
     return (
         <MenuItem
@@ -133,6 +134,7 @@ const ProductSwitcher = (): JSX.Element => {
                         // we currently need this, since not passing a onClick handler is disabling the IconButton
                         // this is a known issue and is being tracked by UI platform team
                         // TODO@UI: remove the onClick, when it is not a mandatory prop anymore
+                        // (ticket https://mattermost.atlassian.net/browse/MM-38476)
                         onClick={() => {}}
                         compact={true}
                         active={switcherOpen}
--- a/components/global/product_switcher.tsx
+++ b/components/global/product_switcher.tsx
@@ -94,7 +94,7 @@ const SwitcherNavEntry = ({icon, destination, text, active, onClick}: SwitcherNa
     );
 };
 
-const ProductSwitcher: React.FC = (): JSX.Element => {
+const ProductSwitcher = (): JSX.Element => {
     const products = useProducts();
     const [switcherOpen, setSwitcherOpen] = useState<boolean>(false);
     const menuRef = useRef<HTMLDivElement>(null);
--- a/components/global/product_switcher.tsx
+++ b/components/global/product_switcher.tsx
@@ -70,6 +70,7 @@ const MenuItemTextContainer = styled.div`
     line-height: 20px;
 `;
 
+// TODO: extract this to a separate file
 const SwitcherNavEntry = ({icon, destination, text, active, onClick}: SwitcherNavEntryProps) => {
     return (
         <MenuItem
@@ -93,14 +94,14 @@ const SwitcherNavEntry = ({icon, destination, text, active, onClick}: SwitcherNa
     );
 };
 
-const ProductSwitcher = (): JSX.Element => {
+const ProductSwitcher: React.FC = (): JSX.Element => {
     const products = useProducts();
-    const [switcherOpen, setSwitcherOpen] = useState(false);
+    const [switcherOpen, setSwitcherOpen] = useState<boolean>(false);
     const menuRef = useRef<HTMLDivElement>(null);
 
     const currentProductID = useCurrentProductId(products);
 
-    const handleClick = () => setSwitcherOpen(!switcherOpen);
+    const handleClick = (): void => setSwitcherOpen(!switcherOpen);
 
     useClickOutsideRef(menuRef, () => {
         setSwitcherOpen(false);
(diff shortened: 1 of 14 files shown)


crspeller (changes requested):
Not sure I like adding all the unnecessary types that can be inferred.
On Playbooks we converged on not using React.FC with webapp approval. So I don't think we should be adding it here. 

@hmhealey Thoughts?

hmhealey:
I think it makes sense to explicitly state types for things that are exported (basically anything that causes a warning from the [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md), but I don't feel super strongly about it for most of the code base. I would be more worried about it for any pieces of code that were exported for use elsewhere since accidentally changing return type would be a lot harder to detect there.

That said, I also don't know what the benefit of declaring components as `React.FC` instead of just using the automatic `(props: Props) => JSX.Element` type. In particular, it seems like `React.FC` is weird by forcing children to be a prop even if your component doesn't allow children or places some restriction on them.

michelengelen:
When doing this I followed what I found in the codebase. Although there are quiet some differences in approach typings and especially the type `React.FC` is a bit controversial. I like to do that just to be clear about the intention (and in some places in the codebase it is also used), so that might be something considered as personal style.

IMO all other typings should be good to go. I normally don't type out those that can be inferred easily.

deanwhillier (approved):

nevyangelova (approved):
LGTM, I am in favour or using JSX.Element although TS infers that your component is already returning a JSX.Element, so you don't even need to declare that return type. https://github.com/facebook/create-react-app/pull/8177

michelengelen:
@nevyangelova the link you posted handles removal of `React.FC` and not `JSX.Element`. Although I do think that you are right, it is still helpful when exporting stuff like @hmhealey already said. I don't see a con in using a return type on a functional component and it even adheres to our eslint rule to type out returns from functions

michelengelen:
@crspeller I removed all of the `React.FC` types from the components. Would you mind re-reviewing? Thanks! 🙇🏼‍♂️

components/global/product_switcher.tsx
```diff
@@ -70,6 +70,7 @@ const MenuItemTextContainer = styled.div`
     line-height: 20px;
 `;
 
+// TODO: extract this to a separate file
```
crspeller, line 73: Can we add tickets to all of these TODO comments?

components/global/product_switcher_menu/product_switcher_menu.tsx
```diff
@@ -45,27 +46,18 @@ type Props = {
...
-
-    render() {
-        const {currentUser, isMessaging, isMobile, onClick} = this.props;
+    handleEmitUserLoggedOutEvent = (): void => GlobalActions.emitUserLoggedOutEvent();
 
-        if (!currentUser) {
-            return null;
-        }
+    render(): JSX.Element {
+        const {appDownloadLink, isMessaging, isMobile, onClick, intl: {formatMessage}} = this.props;
```
crspeller, line 56: If we are going to de-structure we should at least be consistent.  I would prefer leaving everything as `this.props.*` though.

components/global/product_switcher_tip/product_switcher_tip.tsx
```diff
@@ -114,7 +113,7 @@ export class ProductSwitcherTip extends React.PureComponent<Props, State> {
         });
     }
 
-    render() {
+    render(): JSX.Element | null {
```
crspeller, line 116: This is automatic. It doesn't help anyone to put this here.

crspeller (commented):
I think most of this is kinda unnecessary but I don't do much work in the webapp so my style decisions are coming from playbooks. @hmhealey Should we start making some decisions on style here?

components/global/product_switcher_menu/product_switcher_menu.tsx
michelengelen: IMO it needs to be decided on a per case basis: the rule of thumb for my decision-making on this is always: "If you access the property more than once from `this.props.*` it is better to destructure it." I am open for a discussion about this.

components/global/product_switcher_tip/product_switcher_tip.tsx
michelengelen: if that's the case why do we have the before mentioned eslint rule in place? It is constantly throwing warnings about functions/methods not having a return type. It is done very inconsistently in the codebase, so this should probably be put to discussion as well.

components/global/product_switcher.tsx
michelengelen: yep, will do. Thanks for pointing that out. 👍🏼

components/global/product_switcher_menu/product_switcher_menu.tsx
hmhealey: I've gone back and forth on destructuring everything in the past. I used to be against destructuring entirely since I like verbosity and knowing where stuff comes from, but I've definitely changed my opinion over time since that's not always relevant (unless you're doing something like comparing `props` to `nextProps` where it's still relevant because there's 2 objects with the same fields).

I do tend to prefer going all or nothing on destructuring whenever possible, but that's just me though.

michelengelen: It is definitely easier when using functional components, where you typically destructure everything from within the parameter list itself. As for "old-school" components ... well, this is again very much question of personal preference and/or global style-decisions.

michelengelen:
I am going to close this one, since decisions on styling are not exactly necessary at this point and the benefits of applying these types are minimal. We should revisit these when there are guidelines for cases like these and I heard from @hmhealey that it is planned to have discussions (and possibly decisions) on this in Q4.

//cc @crspeller

PR #8836, opened 2021-09-13
remove bodyOnly prop

darkLord19:
Fixes https://github.com/mattermost/mattermost-server/issues/18360


devinbinnie (approved):
LGTM! Thanks @darkLord19!

hmhealey (approved):
I removed the prop from the MoreDirectChannels modal in https://github.com/mattermost/mattermost-webapp/pull/8821, so this'll conflict with that, but that's probably fine since this is going to get through review much faster than that one.

hmhealey:
Skipping QA review since this is just dead code removal

PR #8839, opened 2021-09-13
[MM-38263] - Add channel category menu to channel header

nevyangelova:
#### Summary
This PR adds the option for users to add the channels to different categories from the channel menu, while simultaneously unifies the function in a separate new channel category component. 

Question for PM/QA: There is a current but on daily with menu alignment and its affecting this PR as well. Should it be fixed here or is there a filed ticket that someone is working on already? cc/ @esethna @jgilliam17 

Questions for UX: Currently the category that is selected is moved out of the list and in the new designs the category should stay present and a checkmark should appear next to it. I have implemented it according to the new design but I want to confirm with design that it should be replicated in the menu as well. cc/ @matthewbirtch 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38263

#### Release Note

```release-note
- Added the channel category menu item to the channel header dropdown. 
```


matthewbirtch:
@nevyangelova I assigned @Rina-dsg (Marina) on this since she did the design work for this ticket.

Rina-dsg:
@nevyangelova it'd be great to replicate the indicator for a selected category in the menu (at the sidebar), so we have the same experience in both cases.

esethna:
> it'd be great to replicate the indicator for a selected category in the menu (at the sidebar) 

+1

> There is a current but on daily with menu alignment and its affecting this PR as well. Should it be fixed here or is there a filed ticket that someone is working on already?

@nevyangelova should be fixed separately. There's a ticket here: https://mattermost.atlassian.net/browse/MM-38507

@nevyangelova

components/category_menu_items/category_menu_items.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    icon={<i className='icon-folder-move-outline'/>}
+                    direction={'right' as any}
+                    openUp={openUp}
+                    styleSelectableItem={true}
+                    selectedValueText={currentCategory?.display_name}
+                    renderSelected={false}
+                />
+            </Menu.Group>
+        </React.Fragment>
+    ) : null;
```
devinbinnie, line 112: Nit: 0/5, but I think for improved clarity, we could check if categories is falsy and return null before we return the actual component.

components/category_menu_items/index.ts
```diff
@@ -0,0 +1,3 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+export {default} from './category_menu_items';
```
devinbinnie, line 3: Do we need this connector?

devinbinnie (approved):

components/category_menu_items/category_menu_items.tsx
```diff
@@ -0,0 +1,115 @@
...
+};
+
+export default injectIntl(CategoryMenuItems);
```
hmhealey, line 115: Since this is a functional component, you can use `useIntl` to get that object instead of having to wrap it with `injectIntl`

components/category_menu_items/category_menu_items.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    if (currentTeam) {
+        categories = useSelector((state: GlobalState) => getCategoriesForCurrentTeam(state));
+        currentCategory = useSelector((state: GlobalState) => getCategoryInTeamWithChannel(state, currentTeam.id, channel.id));
+    }
+
+    const moveToCategory = useCallback((categoryId: string) => {
+        dispatch(addChannelsInSidebar(categoryId, channel.id));
+        trackEvent('ui', 'ui_sidebar_channel_menu_moveToExistingCategory');
+    }, []);
```
hmhealey, line 48: This needs to list `channel` as a dependency of the callback

components/category_menu_items/category_menu_items.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const moveToNewCategory = useCallback(() => {
+        dispatch(openModal({
+            modalId: ModalIdentifiers.EDIT_CATEGORY,
+            dialogType: EditCategoryModal,
+            dialogProps: {
+                channelIdsToAdd: multiSelectedChannelIds.indexOf(channel.id) === -1 ? [channel.id] : multiSelectedChannelIds,
+            },
+        }));
+        trackEvent('ui', 'ui_sidebar_channel_menu_createCategory');
+    }, []);
```
hmhealey, line 59: Similarly, this needs to list `channel` and `multiSelectChannelIds`

components/category_menu_items/category_menu_items.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    const selectedChannels = multiSelectedChannelIds.indexOf(channel.id) === -1 ? [channel] : displayedChannels.filter((c) => multiSelectedChannelIds.indexOf(c.id) !== -1);
```
hmhealey, line 61: @esethna @Rina-dsg Should we remove the multiselect logic when this menu is displayed in the channel header? Since the multiselect is firmly part of the sidebar, it might be confusing to users if it affects the channel header as well

components/widgets/menu/menu_items/submenu_item.tsx
```diff
@@ -235,8 +237,8 @@ export default class SubMenuItem extends React.PureComponent<Props, State> {
                 >
                     <span className='MenuItem__primary-text'>
                         {textProp}
-                        {selectedValueText && selectedValueElement}
-                        {id !== 'SidebarChannelMenu-moveToDivider' &&
+                        {renderSelected && selectedValueElement}
```
hmhealey, line 240: What's this change for? 0/5 on it overall, but I'd assume that we just wouldn't pass `selectedValueText` if we don't want the selection to be rendered.

hmhealey (changes requested):

components/category_menu_items/category_menu_items.tsx
hmhealey: Actually, since this isn't passed to a child component as a prop, I don't think this even needs a `useCallback`

components/category_menu_items/category_menu_items.tsx
```diff
@@ -0,0 +1,115 @@
...
+            return category.type !== CategoryTypes.DIRECT_MESSAGES;
+        }
+        return true;
+    }).map((category: ChannelCategory) => {
+        return {
+            id: `moveToCategory-${channel.id}-${category.id}`,
+            icon: category.type === CategoryTypes.FAVORITES ? (<i className='icon-star-outline'/>) : (<i className='icon-folder-outline'/>),
+            direction: 'right' as any,
+            text: category.display_name,
+            action: () => moveToCategory(category.id),
```
hmhealey, line 78: Since we're now showing the channel's current category in this list, could we add some logic to prevent clicking on the current category from triggering an API request?

hmhealey:
@esethna @Rina-dsg Not to devalue any of the work that's been put into this, but does it really make sense to include this option in the channel header (2/5)? When using categories, I think most people will immediately look at the sidebar for options like this since that's where categories appear, so having them in the header feels really disconnected to me.

The channel header menu can also get pretty long already as well, so I'm not sure if it's the best to add more stuff there. I know a few of these are from plugins, but it's already a decent size for me on community.
Screen Shot 2021-09-16 at 9 24 40 AM

esethna:
@hmhealey I don't think it hurts to have it in both places. To me it's about discoverability and ease of use. Also consider the scenario of a user who has never used categories before, but is familiar with accessing the channel menu. This type of a feature also promotes use of the categories feature in general

Rina-dsg:
@nevyangelova 

- Can we get rid of the icon next to 'Move to...' in the Channel menu?
- The arrow icon should be a bit more on the right

Screenshot 2021-09-17 at 14 57 39
Screenshot 2021-09-17 at 15 03 23

nevyangelova:
@Rina-dsg per discussion the second point had a separate ticket but I have addressed it here https://mattermost.atlassian.net/browse/MM-38507

components/category_menu_items/index.ts
nevyangelova: for import clarify and convention but I am ok with removing it.

components/widgets/menu/menu_items/submenu_item.tsx
nevyangelova: we pass the selectedValueText because if it matched the value in the submenu it would display the checkmark icon but sometimes we don't want it in the parent menuitem as well :/

Rina-dsg (approved):

jgilliam17:
Thanks @nevyangelova 
I added a follow up ticket for "move to" background per our discussion. https://mattermost.atlassian.net/browse/MM-38676
One more thing I am noticing - menu separator line is missing below "move to" 
Screenshot 2021-09-20 130817

esethna (approved):
Approving assuming the outstnading issues will be addressed or follow-up tickets created:

ie arrow misplacement, background hover effect, missing separator line

hmhealey:
@nevyangelova Could you actually add a changelog item for this? This seems like something users would be interested in knowing about.

nevyangelova:
@hmhealey added a release note item.
@jgilliam17 nice catch! Added the divider, but will address the menu item opacity separately as this is not a regression here

jgilliam17 (approved):
Thanks @nevyangelova 
Tested, looks good to merge.
- Verified channel menu item "Move to"
- Menu item background and arrow misplacement were filed separately.
https://mattermost.atlassian.net/browse/MM-38676
https://mattermost.atlassian.net/browse/MM-38507

PR #8840, opened 2021-09-13
Mm 38361 pre written messages

neallred:
#### Summary
Add prewritten messages A/B test. A/B test variations are controlled at a server level in Split and require having someone with access changing the `PrewrittenMessages` value in split. You can ask me to change the value to see the different variations.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38361

#### Related Pull Requests
- server changes: https://github.com/mattermost/mattermost-server/pull/18372

#### Screenshots

control group:
prewritten-none

prewritten messages in tip:
prewritten-tour-point

prewritten messages by create post box:
prewritten-around-input

prewritten messages to self:
prewritten-self

prewritten messages to group:
prewritten-group

#### Release Note

```release-note
Added A/B test for providing user with prewritten messages.
```


annelieseklein:
@neallred looks great! For all instances can we please change the :hand: to :wave: ? For the "Oh hello" can we change "🙌" to " ✌️ "?  Also for test C, the highlight box should be at 8% opacity of the button color

neallred:
Having issues getting the spinwick to pull in the feature flag from split. Will post here when its resolved.

components/create_post/prewritten_chips.tsx
```diff
@@ -0,0 +1,144 @@
...
+        customChip,
+    ];
+}
+
+class PrewrittenChips extends React.PureComponent<Props> {
+    render() {
+        const chips = getChips(this.props.currentChannel, this.props.currentUserId);
+        return (
+            <div
+                style={{
```
pablovelezvidal, line 107: nit: if we are using styled components, we should try to avoid inline styles

components/create_post/create_post_tip.tsx
```diff
@@ -0,0 +1,86 @@
...
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {FormattedMessage} from 'react-intl';
+
+import {Channel} from 'mattermost-redux/types/channels';
+
+import FormattedMarkdownMessage from 'components/formatted_markdown_message';
+import TutorialTip from 'components/tutorial/tutorial_tip';
+import {PrewrittenMessagesTreatments} from 'mattermost-redux/constants/config';
```
pablovelezvidal, line 11: group mattermost-redux imports together

components/create_post/create_post.tsx
```diff
@@ -1316,6 +1329,61 @@ class CreatePost extends React.PureComponent<Props, State> {
...
+            if (this.props.currentChannel.teammate_id === this.props.currentUserId) {
+                id = t('create_post.prewritten.around_input.self');
+                defaultMessage = '**Send your first message** to yourself';
+            } else {
+                id = t('create_post.prewritten.around_input.dm');
+                defaultMessage = '**Send your first message** to your teammate';
+            }
+        }
+        return (
+            <div>
```
pablovelezvidal, line 1361: is this div used to layout the internal structure, or just to accomplish the react rule of only render one tag? if so, maybe you could use a React.Fragment `<></>` instead.

components/create_post/create_post.tsx
```diff
@@ -590,6 +614,10 @@ class CreatePost extends React.PureComponent<Props, State> {
 
         this.props.actions.setDraft(StoragePrefixes.DRAFT + channelId, null);
         this.draftsForChannel[channelId] = null;
+
+        if (this.props.tutorialStep === Constants.TutorialSteps.POST_POPOVER && this.props.prewrittenMessages && this.props.prewrittenMessages !== PrewrittenMessagesTreatments.NONE) {
```
pablovelezvidal, line 618: nit: I think it is a good practice to make code self documented and put this kind of long conditions in a constant value to make code self explanatory and easy to read, for example, in this case, it could be something like `const shouldShowPrewrittenMsgs = this.props.tutoriaStep === Contants.Tuto....`

pablovelezvidal (approved):
Great Work @neallred . Just some minor comments!

neallred:
Got the spinwick updated with the right server code! You can see the pulsating dot version of the test now.

esethna:
Thanks @neallred,

For this version

image
1. Can we update the copy?

Select or type your first message and select **Enter** to send it. 

Use the **Attachments** and **Emoji** buttons to add to your messages.

@annelieseklein feel free to ping me to review copy on future designs ahead of implementation. 

2. @annelieseklein @neallred it looks like the buttons are missing hover states? 

3. Did we add telemetry to each of the buttons, including the "Got it" button and dismissing the modal without using a button (ie esc or click away)? 

4. The designs state that the tips "reengage" after 48 hours, what does that mean?

neallred:
> 1. ...copy?

Sure!

> 2. missing hover states

They were incorrectly coded on my part.

> 3. ... telemetry to each...

The got it button has it, but I still need to add to each of the others.

> 4.  designs state that the tips "reengage" after 48 hours ...

The ticket should probably be updated. It was my understanding this was a potential later refinement that we don't want to implement now.

AGMETEOR:
@neallred I want to approve but I feel like I should wait until all the PM feedback is completed since there would be code changes.

esethna (approved):
Approving pending copy updates

neallred:
@pablovelezvidal re-requested review because there's been some meaningful changes to tracking users' dismissal of tutorial tip.

components/tutorial/tutorial_tip/tutorial_tip.tsx
```diff
@@ -63,6 +63,32 @@ export default class TutorialTip extends React.PureComponent<Props, State> {
...
+            // If a user click on an element inside triggers a click outside,
+            // then react-bootstrap's onHide fires onHide for both elements.
+            // Even if the inside click is for opt out, that is already covered in the _skip event
+            !target.closest('.tip-overlay') &&
+
+            // Clicking the post textbox means a prefilled message was chosen, so it is not a dismissal.
+            target.id !== 'post_textbox'
+        );
+        if (wasOutsideClick) {
+            trackEvent('tutorial', tag);
```
neallred, line 88: Per conversation with @esethna we want to track when the user hides the modal. I was hoping this function could be simpler, but for cases like prewritten messages where we need to generate a click to hide the overlay, this function gets called twice, once for the original element click, and once for the create post textbox being clicked/focused. Also, this function gets called when the user clicks "got it" or opts out of the tips. So I had to do some event sniffing to tell if the user was really dismissing the overlay.

annelieseklein (approved):

annelieseklein:
> The ticket should probably be updated. It was my understanding this was a potential later refinement that we don't want to implement now.

^Yes this was for future state -- I'll update the ticket to remove it.

AGMETEOR (approved):
Awesome job @neallred 🎉 LGTM

components/create_post/chip.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+// This component is a temporary placeholder for use until the authoritative `compass-components` Chip is implemented.
+
+type Props = {
+    onClick: () => void;
+    id: string;
+    defaultMessage: string;
+    values: Record<string, any>;
+
+    // for the "other" option unlike the others, e.g. free form response
```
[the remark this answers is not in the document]
neallred: I web searched it, and I guess its `free-form`: https://www.merriam-webster.com/dictionary/free-form

pablovelezvidal (approved):
Great work @neallred ! LGTM :shipit:

stevemudie (approved):
All of the various variations checked and looking as expected.

Great work @neallred! You get to stay another day!

PR #8844, opened 2021-09-13
MM-24370 Remove manual focus change from cancel button 

hmhealey:
Looking back at the diffs, we might have had some logic in this modal to manually control focus, but it was mostly removed a while back with only this `onBlur` left over. Without it, react-bootstrap's automatic focus trapping seems to work fine to keep the keyboard focus within the modal when tabbing around.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24370

#### Release Note
```release-note
Fixed keyboard focus within New Channel modal
```


koox00 (approved):

deanwhillier (approved):

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified pressing tab on new channel modal switches between all input fields and buttons, including Create Channel button.

PR #8846, opened 2021-09-13
Mm 38063 date to utc

hamzaMM:
#### Summary
Changed date to utc in the data retention policy table

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38063

#### Release Note

```release-note
NONE
```


components/admin_console/jobs/table.jsx
```diff
@@ -286,7 +286,7 @@ class JobTable extends React.PureComponent {
             );
         }
 
-        const date = new Date(millis);
+        const date = new Date(new Date(millis).toUTCString().substr(0, 25));
```
mkraft, line 289: I think we should use a localized version of this date display using moment.js, which is already a JS package we ship with.

mkraft (changes requested):

BenCookie95:
@hamzaMM The package-lock.json changes shouldn't be there

PR #8848, opened 2021-09-14
MM-38140 - show only enterprise for professional upgrade

pablovelezvidal:
#### Summary
When a customer is upgrading to Enterprise from professional, we want to only show the Enterprise option.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38140

#### Related Pull Requests
none

#### Screenshots
|  before  |  after  |
|----|----|
| 
Captura de pantalla 2021-09-14 a las 9 27 30
 | 
Captura de pantalla 2021-09-14 a las 9 27 09
 |

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/purchase_modal/purchase_modal.tsx
+++ b/components/purchase_modal/purchase_modal.tsx
@@ -224,7 +224,13 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
             if (products[key].billing_scheme === BillingSchemes.FLAT_FEE) {
                 flatFeeProducts.push(tempEl);
             } else {
+
+                // only show enterprise to customers upgrading from professional
+                if (currentProduct.sku === 'cloud-professional') {
+                    return;
+                }
                 userBasedProducts.push(tempEl);
+
             }
         });
 
@@ -527,7 +533,7 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
                 </div>
                 <div className='RHS'>
                     <div className='price-container'>
-                        {(this.props.products && Object.keys(this.props.products).length > 1) &&
+                        {(this.props.products && Object.keys(this.props.products).length > 1 && this.state.currentProduct.sku !== 'cloud-professional') &&
                             <div className='select-plan'>
                                 <div className='title'>
                                     <FormattedMessage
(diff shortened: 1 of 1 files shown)


AGMETEOR (approved):

AGMETEOR:
@neallred I want to approve but I feel like I should wait until all the PM feedback is completed since there would be code changes.

stevemudie (approved):
All set @pablovelezvidal 

After the initial purchase of Professional all I could see when upgrading from there was Enterprise as expected.

components/purchase_modal/purchase_modal.tsx
```diff
@@ -527,7 +531,7 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
                 </div>
                 <div className='RHS'>
                     <div className='price-container'>
-                        {(this.props.products && Object.keys(this.props.products).length > 1) &&
+                        {(this.props.products && Object.keys(this.props.products).length > 1 && this.state.currentProduct?.sku !== 'cloud-professional') &&
```
marianunez, line 534: Why do we need to hardcode the condition here to check for product name? If there is more than 1 product then it should have the compare plans. In the case of current sku is `Starter` there should be more than one product so it should show the compare plans and in the case of `Professional` it should only have 1 product and not show this

components/purchase_modal/purchase_modal.tsx
```diff
@@ -224,6 +224,10 @@ export default class PurchaseModal extends React.PureComponent<Props, State> {
             if (products[key].billing_scheme === BillingSchemes.FLAT_FEE) {
                 flatFeeProducts.push(tempEl);
             } else {
+                // only show enterprise to customers upgrading from professional
+                if (currentProduct.sku === 'cloud-professional') {
+                    return;
+                }
```
marianunez, line 230: Why filter here for an specific product if it filters by product price below?

marianunez (changes requested):
We should avoid putting conditions for specific skus - let's review the logic we had before for ordering/filtering by product price

components/purchase_modal/purchase_modal.tsx
pablovelezvidal: ok

components/purchase_modal/purchase_modal.tsx
pablovelezvidal: ok

stevemudie (approved):
Stepped through from Starter to Professional and when I went to upgrade to Enterprise it was the only option displayed as expected.

marianunez (approved):
Thanks @pablovelezvidal!

AGMETEOR (approved):

annelieseklein (approved):

PR #8851, opened 2021-09-14
[MM-36551] - improvements to post edit indicator

michelengelen:
#### Summary
According to the [designs](https://www.figma.com/file/fVLSLjGls2d2LUvCqk2JMS/MM-29137-Message-Editing?node-id=630%3A240624) this PR moves the edit-indicator on posts inline and adds some more functionality in the form of a popover which holds information on when the post was last edited.

For this purpose I had to extend 2 functions that are responsible for rendering: `formatText` and `messageHtmlToComponent`, since the rendering of posts differs quite a bit depending on the content. Those are being called from the `Markdown` component (first `formatText` and afterwards `messageHtmlToComponent`).

**changes to `formatText`:**
I essentially remove a wrapping `<p>` tag if it exists and append an empty `<span>` with an id. The format of the id is `postEdited_${postId}`. This makes the span id unique and can be replaced by the second following function: `messageHtmlToComponent`. In order to do this I added a new optional parameter to the function: `postId` with the default value of `''`.

**changes to `messageHtmlToComponent`:**
The only thing that needed to be added here was a new processing instruction to be handled by the `html-to-react` parser. It processes all `<span>` with a matching Id (the one I set from the `formatText` function and replaces it with a new `PostEditIndicator` component.

The matching condition is:
```typescript
node.type === 'tag' && node.name === 'span' && node.attribs.id && node.attribs.id === `postEdited_${options.postId}`
```

To make this possible I added a new property to the `options` object: `editedAt`, which is being set from the post property `post.edit_at`.

**UI changes**
the edit indicator is now rendered as a combination of a pencil-icon and the text `Edited` according to the designs.

**EDIT:**
addition of an `OverlayTrigger` integrating a more detailed text about the edit date:

|  edited ...  |  today  |  yesterday  |  within last week  |  more than 1 week ago  |
|-----|-----|-----|-----|-----|
|   |Screenshot 2021-09-15 at 17 21 25|Screenshot 2021-09-15 at 17 19 02|Screenshot 2021-09-15 at 17 19 48|Screenshot 2021-09-15 at 17 20 32|

#### Ticket Link
[MM-36551](https://mattermost.atlassian.net/browse/MM-36551)

#### Related Pull Requests
- Has mobile changes: TBD

#### Screenshots
|  type  |  before  |  after  |
|----|----|----|
|  normal post  |Screenshot 2021-09-14 at 15 24 30|Screenshot 2021-09-14 at 15 22 16|
|  markdown table  |Screenshot 2021-09-14 at 15 24 38|Screenshot 2021-09-14 at 15 22 32|

#### Release Note
```release-note
Changed the UI of the edit-indicator of posts and moved it inline
```

### DIFF
--- a/components/markdown/markdown.tsx
+++ b/components/markdown/markdown.tsx
@@ -10,6 +10,7 @@ import {Dictionary} from 'mattermost-redux/types/utilities';
 import messageHtmlToComponent from 'utils/message_html_to_component';
 import EmojiMap from 'utils/emoji_map';
 import {ChannelNamesMap, TextFormattingOptions, formatText, MentionKey} from 'utils/text_formatting';
+import PostEditedIndicator from '../post_view/post_edited_indicator/post_edited_indicator';
 
 type Props = {
 
@@ -100,6 +101,11 @@ type Props = {
      */
     postId?: string;
 
+    /**
+     * When the post is edited this is the timestamp it happened at
+     */
+    editedAt?: number;
+
     channelId?: string;
 
     /**
@@ -116,11 +122,21 @@ export default class Markdown extends React.PureComponent<Props> {
         proxyImages: true,
         imagesMetadata: {},
         postId: '', // Needed to avoid proptypes console errors for cases like channel header, which doesn't have a proper value
+        editedAt: 0,
     }
 
     render() {
+        const {postId, editedAt, message} = this.props;
         if (!this.props.enableFormatting) {
-            return <span>{this.props.message}</span>;
+            return (
+                <span>
+                    {this.props.message}
+                    <PostEditedIndicator
+                        postId={postId}
+                        editedAt={editedAt}
+                    />
+                </span>
+            );
         }
 
         const options = Object.assign({
@@ -133,9 +149,11 @@ export default class Markdown extends React.PureComponent<Props> {
             team: this.props.team,
             minimumHashtagLength: this.props.minimumHashtagLength,
             managedResourcePaths: this.props.managedResourcePaths,
+            editedAt,
         }, this.props.options);
 
-        const htmlFormattedText = formatText(this.props.message, options, this.props.emojiMap);
+        const htmlFormattedText = formatText(message, options, this.props.emojiMap, postId);
+
         return messageHtmlToComponent(htmlFormattedText, this.props.isRHS, {
             imageProps: this.props.imageProps,
             imagesMetadata: this.props.imagesMetadata,
@@ -145,6 +163,7 @@ export default class Markdown extends React.PureComponent<Props> {
             postType: this.props.postType,
             mentionHighlight: this.props.options.mentionHighlight,
             disableGroupHighlight: this.props.options.disableGroupHighlight,
+            editedAt,
         });
     }
 }
--- a/components/markdown/markdown.tsx
+++ b/components/markdown/markdown.tsx
@@ -150,9 +150,10 @@ export default class Markdown extends React.PureComponent<Props> {
             minimumHashtagLength: this.props.minimumHashtagLength,
             managedResourcePaths: this.props.managedResourcePaths,
             editedAt,
+            postId,
         }, this.props.options);
 
-        const htmlFormattedText = formatText(message, options, this.props.emojiMap, postId);
+        const htmlFormattedText = formatText(message, options, this.props.emojiMap);
 
         return messageHtmlToComponent(htmlFormattedText, this.props.isRHS, {
             imageProps: this.props.imageProps,
--- a/components/post_markdown/post_markdown.tsx
+++ b/components/post_markdown/post_markdown.tsx
@@ -59,7 +59,7 @@ export default class PostMarkdown extends React.PureComponent<Props> {
     };
 
     getOptions = memoize(
-        (disableGroupHighlight: boolean, options?: TextFormattingOptions, mentionHighlight?: boolean, editedAt?: number) => {
+        (options?: TextFormattingOptions, disableGroupHighlight?: boolean, mentionHighlight?: boolean, editedAt?: number) => {
             return {
                 ...options,
                 disableGroupHighlight,
@@ -95,8 +95,8 @@ export default class PostMarkdown extends React.PureComponent<Props> {
         }
 
         const options = this.getOptions(
-            post?.props?.disable_group_highlight === true, // eslint-disable-line camelcase
             this.props.options,
+            post?.props?.disable_group_highlight === true, // eslint-disable-line camelcase
             mentionHighlight,
             post?.edit_at,
         );
@@ -111,9 +111,9 @@ export default class PostMarkdown extends React.PureComponent<Props> {
                 options={options}
                 channelNamesMap={channelNamesMap}
                 hasPluginTooltips={this.props.hasPluginTooltips}
-                imagesMetadata={this.props.post && this.props.post.metadata && this.props.post.metadata.images}
-                postId={this.props.post && this.props.post.id}
-                editedAt={this.props.post && this.props.post.edit_at}
+                imagesMetadata={this.props.post?.metadata?.images}
+                postId={this.props.post?.id}
+                editedAt={this.props.post?.edit_at}
             />
         );
     }
--- a/components/post_markdown/post_markdown.tsx
+++ b/components/post_markdown/post_markdown.tsx
@@ -59,11 +59,12 @@ export default class PostMarkdown extends React.PureComponent<Props> {
     };
 
     getOptions = memoize(
-        (options: TextFormattingOptions | undefined, disableGroupHighlight: boolean, mentionHighlight: boolean | undefined) => {
+        (options: TextFormattingOptions | undefined, disableGroupHighlight: boolean, mentionHighlight: boolean | undefined, editedAt: number | null) => {
             return {
                 ...options,
                 disableGroupHighlight,
                 mentionHighlight,
+                editedAt,
             };
         })
 
@@ -97,6 +98,7 @@ export default class PostMarkdown extends React.PureComponent<Props> {
             this.props.options,
             post?.props?.disable_group_highlight === true, // eslint-disable-line camelcase
             mentionHighlight,
+            post?.edit_at || null,
         );
 
         return (
@@ -111,6 +113,7 @@ export default class PostMarkdown extends React.PureComponent<Props> {
                 hasPluginTooltips={this.props.hasPluginTooltips}
                 imagesMetadata={this.props.post && this.props.post.metadata && this.props.post.metadata.images}
                 postId={this.props.post && this.props.post.id}
+                editedAt={this.props.post && this.props.post.edit_at}
             />
         );
     }
--- a/components/post_markdown/post_markdown.tsx
+++ b/components/post_markdown/post_markdown.tsx
@@ -59,7 +59,7 @@ export default class PostMarkdown extends React.PureComponent<Props> {
     };
 
     getOptions = memoize(
-        (options: TextFormattingOptions | undefined, disableGroupHighlight: boolean, mentionHighlight: boolean | undefined, editedAt: number | null) => {
+        (disableGroupHighlight: boolean, options?: TextFormattingOptions, mentionHighlight?: boolean, editedAt?: number) => {
             return {
                 ...options,
                 disableGroupHighlight,
@@ -95,10 +95,10 @@ export default class PostMarkdown extends React.PureComponent<Props> {
         }
 
         const options = this.getOptions(
-            this.props.options,
             post?.props?.disable_group_highlight === true, // eslint-disable-line camelcase
+            this.props.options,
             mentionHighlight,
-            post?.edit_at || null,
+            post?.edit_at,
         );
 
         return (
--- a/components/post_view/post_edited_indicator/post_edited_indicator.tsx
+++ b/components/post_view/post_edited_indicator/post_edited_indicator.tsx
@@ -10,11 +10,11 @@ import {isSameDay, isWithinLastWeek, isYesterday} from '../../../utils/datetime'
 import OverlayTrigger from '../../overlay_trigger';
 
 interface Props {
-    postId?: string | null;
+    postId?: string;
     editedAt?: number;
 }
 
-const PostEditedIndicator = ({postId = null, editedAt = 0}: Props): JSX.Element | null => {
+const PostEditedIndicator = ({postId, editedAt = 0}: Props): JSX.Element | null => {
     if (!postId || editedAt === 0) {
         return null;
     }
--- a/components/post_view/post_edited_indicator/post_edited_indicator.tsx
+++ b/components/post_view/post_edited_indicator/post_edited_indicator.tsx
@@ -38,7 +38,7 @@ const PostEditedIndicator = ({postId, editedAt = 0}: Props): JSX.Element | null 
         date = formatDate(editedDate, {month: 'long', day: 'numeric'});
     }
 
-    const time = formatTime(editedDate, {hour: '2-digit', minute: '2-digit'});
+    const time = formatTime(editedDate, {hour: 'numeric', minute: '2-digit'});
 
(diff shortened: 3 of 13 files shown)


matthewbirtch:
@michelengelen looks like we're having test server issues again

michelengelen:
@matthewbirtch i am currently fixing the CI issues

components/markdown/markdown.tsx
```diff
@@ -133,9 +149,11 @@ export default class Markdown extends React.PureComponent<Props> {
             team: this.props.team,
             minimumHashtagLength: this.props.minimumHashtagLength,
             managedResourcePaths: this.props.managedResourcePaths,
+            editedAt,
         }, this.props.options);
 
-        const htmlFormattedText = formatText(this.props.message, options, this.props.emojiMap);
+        const htmlFormattedText = formatText(message, options, this.props.emojiMap, postId);
```
hmhealey, line 155: Perhaps the postId should also be added to options?

components/post_view/post_edited_indicator/post_edited_indicator.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {FormattedMessage} from 'react-intl';
+import Icon from '@mattermost/compass-components/foundations/icon/Icon';
+
+interface Props {
+    postId?: string | null;
+    editedAt?: number;
+}
+
+const PostEditedIndicator = ({postId = null, editedAt = 0}: Props): JSX.Element | null => {
+    return postId && editedAt > 0 ? (
```
hmhealey, line 14: Could you do an early return when either of these are falsey? It avoids some nesting, and personally I'm not a fan of multi-line ternary statements like this as well.

utils/text_formatting.tsx
```diff
@@ -262,6 +263,16 @@ export function formatText(
         output = `<span class="all-emoji">${output.trim()}</span>`;
     }
 
+    if (postId) {
+        // unwrap the output from an enclosing p tag and add a span that will serve as a
+        // palceholder for `messageToHtmlComponent` function later on
+        if (output.startsWith('<p>') && output.endsWith('</p>')) {
```
hmhealey, line 269: Do we need to check for the starting p tag or is just checking for the ending one sufficient? I'm trying to think of how a post starting with a list or table would look when turned to HTML.

utils/text_formatting.tsx
```diff
@@ -262,6 +263,16 @@ export function formatText(
...
+            output = `${output.slice(0, -4)}<span id='postEdited_${postId}'>(edited)</span></p>`;
```
hmhealey, line 270: Would we be using this ID for anything else? I wonder if using something like `data-edited-id=${postId}` would be a bit better so that we're not adding extra meaning to an ID attribute that might be used elsewhere

components/markdown/markdown.tsx
michelengelen: good idea ... it was messing with the types last time i tried that, but it is a better solution, so i will try once more!

components/post_view/post_edited_indicator/post_edited_indicator.tsx
michelengelen: yep, will do ... as well as an immediate return!

utils/text_formatting.tsx
michelengelen: oh, you are absolutely right on this one ... it was due to my first experiments with that, where I somehow came to the conclusion that the whole text is being wrapped in an additional `p` tag. 

Nice catch 👍🏼

utils/text_formatting.tsx
michelengelen: either way should be fine... I will adjust that using the `dataset`

components/post_view/post_edited_indicator/post_edited_indicator.tsx
michelengelen: haha ... eslint does not let me use immediate returns with ternaries 👍🏼

hmhealey: I thought that actually would've gotten rid of the ternary entirely like
```jsx
if (!postId || editedAt === 0) {
    return undefined;
}

return <Stuff/>;
```

michelengelen: but that is exactly the same ... so why should we use the "old-school" if clause?

components/markdown/markdown.tsx
```diff
@@ -116,11 +122,21 @@ export default class Markdown extends React.PureComponent<Props> {
...
     }
 
     render() {
+        const {postId, editedAt, message} = this.props;
         if (!this.props.enableFormatting) {
-            return <span>{this.props.message}</span>;
+            return (
+                <span>
+                    {this.props.message}
+                    <PostEditedIndicator
```
larkox, line 134: 0/5 Markdown is a component way more general. I wonder if it makes sense to handle this difference somewhere else, and not in this component. Probably not possible, but still, feels weird.

components/post_markdown/post_markdown.tsx
```diff
@@ -59,11 +59,12 @@ export default class PostMarkdown extends React.PureComponent<Props> {
     };
 
     getOptions = memoize(
-        (options: TextFormattingOptions | undefined, disableGroupHighlight: boolean, mentionHighlight: boolean | undefined) => {
+        (disableGroupHighlight: boolean, options?: TextFormattingOptions, mentionHighlight?: boolean, editedAt?: number) => {
```
larkox, line 62: disableGroupHighlight, being a boolean, doesn't change much the semantics whether it is optional or not, unless we want to force the developer to write true or false for that, which I don't think it is the case.

Therefore, to lower the footprint of this pr, I would just make disableGroupHighlight an optional argument.

components/post_markdown/post_markdown.tsx
```diff
@@ -111,6 +113,7 @@ export default class PostMarkdown extends React.PureComponent<Props> {
                 hasPluginTooltips={this.props.hasPluginTooltips}
                 imagesMetadata={this.props.post && this.props.post.metadata && this.props.post.metadata.images}
                 postId={this.props.post && this.props.post.id}
+                editedAt={this.props.post && this.props.post.edit_at}
```
larkox, line 116: Why not `this.props.post?.edit_at`?

components/post_view/post_edited_indicator/post_edited_indicator.tsx
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {FormattedMessage} from 'react-intl';
+import Icon from '@mattermost/compass-components/foundations/icon/Icon';
+
+interface Props {
+    postId?: string | null;
```
larkox, line 9: Any strong reason to null this value? Can it be just undefined?

components/post_view/post_edited_indicator/post_edited_indicator.tsx
```diff
@@ -0,0 +1,33 @@
...
+    editedAt?: number;
+}
+
+const PostEditedIndicator = ({postId = null, editedAt = 0}: Props): JSX.Element | null => {
+    return !postId || editedAt === 0 ? null : (
```
larkox, line 14: 0/5 this is a personal taste, but I would prefer this to be in the form of
```
if CONDITION {
  return null
}
return (COMPONENT)
```

Nevertheless, this is just my personal taste. Not sure if we have other style guidelines that prefer the `x ? a : b` approach.

utils/text_formatting.tsx
```diff
@@ -262,6 +271,16 @@ export function formatText(
         output = `<span class="all-emoji">${output.trim()}</span>`;
     }
 
+    if (options.postId) {
+        // unwrap the output from the closing p tag and add a span that will serve as a
+        // palceholder for `messageToHtmlComponent` function later on
+        if (output.endsWith('</p>')) {
+            output = `${output.slice(0, -4)}<span data-edited-post-id='${options.postId}'></span></p>`;
```
larkox, line 278: 0/5 to avoid magic numbers, I would prefer something like  `output.slice(0, -'</p>'.length)`

larkox (approved):
LGTM. I have several style considerations, but I am ok merging as it is today.

components/post_view/post_edited_indicator/post_edited_indicator.tsx
michelengelen: i am setting a default value and I am not a big fan of setting values manually to undefined like this:

```typescript
const Component = ({ param = undefined, ... }) => { ... }
```

components/post_markdown/post_markdown.tsx
michelengelen: good question ... I was just following the structure in the rest of the props ... changed it!

components/post_markdown/post_markdown.tsx
michelengelen: yep ... good point!

components/post_view/post_edited_indicator/post_edited_indicator.tsx
michelengelen: although I changed the structure a bit in the following commits, so changing this is now totally fine

components/post_view/post_edited_indicator/post_edited_indicator.tsx
michelengelen: yep , that is one of the changes I just made to the component

utils/text_formatting.tsx
michelengelen: seems a bit strange to do it like this. It looks overcomplicated. I did write a comment that should explain what is happening. IMO I would leave it like this for now.

michelengelen:
@esethna and @matthewbirtch cloud server is up now ... feel free to take a look.

@hmhealey and @larkox ... could you please re-review. since it was a WIP branch the technical stuff changed after adding the last features.

matthewbirtch (changes requested):
Hey @michelengelen, a few things I'm seeing at first glance:

1. The pencil icon is overlapping the content
image

2. The icon is too far away from the `Edited` text

3. The text is not italic as intended in the design

4. The 'Edited at' time displays the `0` in front of 1:11 PM. Can we remove the `0`?
image

michelengelen:
@matthewbirtch thanks for your first review/impression

1. you are using Chrome, right? It is due to [MM-38500](https://mattermost.atlassian.net/browse/MM-38500) that the icon is overlapping. On Firefox it is not, since I changed the `direction` property for development.
2. Same as 1.
3. I need to look into that.
4. that should not be a problem!

michelengelen:
@matthewbirtch points 1, 2 and 4 should be fixed now. I am going to look into 3 tomorrow! 🚀

matthewbirtch:
> @matthewbirtch points 1, 2 and 4 should be fixed now. I am going to look into 3 tomorrow! 🚀

Thanks @michelengelen. It now looks like the icon is too close to the 'Edited' text. Can you add 2px of space between the icon and the word?

For the italics, you should be able to just add `font-style: italic`, no?

esethna (approved):

components/post_view/post_message_view/post_message_view.test.tsx
```diff
@@ -62,10 +62,8 @@ describe('components/post_view/PostAttachment', () => {
     test('should match snapshot, on edited post', () => {
         const props = {...baseProps, post: {...post, edit_at: 1}};
         const wrapper = shallow(<PostMessageView {...props}/>);
-        const instance = wrapper.instance() as PostMessageView;
```
larkox, line 65: I understand we cannot any longer test this like this. Any way we can add snapshots or tests to check again the edit functionality?

utils/datetime.ts
```diff
@@ -24,9 +24,7 @@ export function isWithin(
     truncateEndpoints = shouldTruncate.get(unit) || false,
 ): boolean {
     const diff = getDiff(a, b, timeZone, unit, truncateEndpoints);
-    return threshold >= 0 ?
-        diff <= threshold && diff >= 0 :
-        diff >= threshold && diff <= 0;
+    return threshold >= 0 ? diff <= threshold && diff >= 0 : diff >= threshold && diff <= 0;
```
larkox, line 27: 0/5 I always felt the ternary operator too confusing. Here, where the "inside" of each part is kind of complex, even more. I liked better the multiline approach, even if it was a bit "strange". I would even like better the "golang way". Instead of `return x ? a : b` I would go with:
```
if (x) {
  return a
}
return b
```

larkox: Similar comments with other uses of the ternary operator.

larkox (approved):
Some minor comments. Also, should we check the e2e tests? Just a wild guess, but this may cause some of them to fail.

matthewbirtch (approved):
Looks great now to me @michelengelen

matthewbirtch:
> if there's not enough room to display both inline, then push both to the next ~row~ line. @matthewbirtch thoughts?

Great catch @jgilliam17. 100% agree with this

michelengelen:
@jgilliam17 it should be working as expected now.

/cc @hmhealey kind reminder to review

jgilliam17:
@michelengelen 
1. Looks good on Chrome, Safari and Edge (macOS and Windows), but on Firefox the icon is overlapping edited text

2. When 24-hour time display is set in Account Settings>Display>Clock display, this setting is not reflected on the popover. 

Screen Shot 2021-09-17 at 9 43 59 AM

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
- Verified in-line post edited indicator and popover.
Follow up filed for clock format on popover: https://mattermost.atlassian.net/browse/MM-38636
Firefox icon issue already has a ticket open: https://mattermost.atlassian.net/browse/MM-38500

utils/datetime.ts
michelengelen: this is basically a question of personal preference. IMO  this is just as readable as the other approach, but I am well used to this pattern, so it might not be the case for everybody. afaik the ternary does not have any impact on performance, since it is only syntactic sugar.

I am fine with writing it the "golang way".

@hmhealey Thoughts?

components/post_view/post_message_view/post_message_view.test.tsx
michelengelen: I will find a way to add tests again.

components/post_view/post_edited_indicator/post_edited_indicator.tsx
hmhealey: I was thinking it was going to reduce indentation, but I guess the changes I suggested didn't actually do that since you have to indent the JSX either way

hmhealey (approved):

utils/text_formatting.tsx
```diff
@@ -262,6 +271,16 @@ export function formatText(
         output = `<span class="all-emoji">${output.trim()}</span>`;
     }
 
+    if (options.postId) {
+        // unwrap the output from the closing p tag and add a span that will serve as a
+        // palceholder for `messageToHtmlComponent` function later on
+        if (output.endsWith('</p>')) {
+            output = `${output.slice(0, -4)} <span data-edited-post-id='${options.postId}'></span></p>`;
```
saturninoabril, line 278: Hello @michelengelen, sorry to bother you but I wonder if the extra spaces here were intended or can be changed to:
```diff
- output = `${output.slice(0, -4)} <span
+ output = `${output.slice(0, -4)} <span

- output += ` <span
+ output += `<span
```

Though no perceived side-effect in actual UI as observed by the user, our UI automation tests failed a lot with this new format due to the extra space rendered at the last position of the message.

Please take a look and let me know what you think. Thanks!

larkox: I think this is making also some Apps Framework e2e tests to fail.

michelengelen: it is ... i am currently fixing this in #8964

PR #8852, opened 2021-09-14
[MM-31389] Show confirm modal for `@here` mentions

streamer45:
#### Summary

PR makes `@here` mention trigger a confirmation modal, similarly to `@all` and `@channel` when `EnableConfirmNotificationsToChannel` is set.

#### Ticket

https://mattermost.atlassian.net/browse/MM-31389

#### Release Note

```release-note
Included @here mentions in the `EnableConfirmNotificationsToChannel` config setting.
```

#### Screenshot

image

#### Related PRs

https://github.com/mattermost/mattermost-mobile/pull/5678


components/create_comment/create_comment.tsx
```diff
@@ -573,6 +574,7 @@ class CreateComment extends React.PureComponent<Props, State> {
         let channelTimezoneCount = 0;
         let mentions: string[] = [];
         const notContainsAtChannel = !containsAtChannel(draft.message);
+        const containsAtHereMention = notContainsAtChannel && containsAtChannel(draft.message, {checkAllMentions: true});
```
hmhealey, line 577: Since we're treating at-here the same as at-channel now, could we just have `containsAtChannel` return true for at-here as well? Perhaps `containsAtChannel` could also return which of those are matched so that we don't need to work it out below.

components/create_comment/create_comment.tsx
```diff
@@ -1098,21 +1099,29 @@ class CreateComment extends React.PureComponent<Props, State> {
                 />
             );
             if (channelTimezoneCount > 0) {
+                const atHereMsg = 'By using **@here** you are about to send notifications to up to **{totalMembers} people** in **{timezones, number} {timezones, plural, one {timezone} other {timezones}}**. Are you sure you want to do this?';
```
hmhealey, line 1102: This is definitely outside the scope of this ticket, but since this logic is becoming more complicated, perhaps we could move the logic for rendering this modal into a new component called something like `NotifyConfirmModal` that could be controlled with the `openModal` redux action so that `CreateComment` and `CreatePost` don't have to be responsible for it? I'm thinking we could add something like
```javascript
this.props.actions.openModal({
    dialogType: NotifyConfirmModal,
    dialogProps: {
        mentions,
        onConfirm: () => {/*whatever actually submits the post*/},
    },
});
```
to the `handleSubmit` method instead of having it store a bunch of state in the component to make the modal appear.

0/5 on doing that now though

streamer45: I must admin that I've thought about that cause I am not a fan of redundant code but somehow when it comes to front-end work laziness gets the better of me. I'll have a stab at that.

components/create_comment/create_comment.tsx
streamer45: Yeah logic there is getting a bit complex. I'll have a look.

esethna (approved):

pablovelezvidal (approved):
LGTM! thanks @streamer45 !

components/create_comment/create_comment.tsx
hmhealey: Sounds good. If not, that's totally fine too. I'm thinking of making an improved modal system as one of Web Platform's OKRs next quarter, so we could always address that then.

hmhealey (approved):
Awesome, thanks for changing all that

furqanmlk (approved):
LGTM
image

PR #8853, opened 2021-09-14
[MM-36442] More Actions Menu Restructure

jfrerich:
#### Summary
This is a Draft PR for tracking and sharing changes made while implementing 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36442

#### Related Pull Requests

#### Release Note

```release-note

```

### DIFF
--- a/components/actions_menu/actions_menu.tsx
+++ b/components/actions_menu/actions_menu.tsx
@@ -1,0 +1,292 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import classNames from 'classnames';
+import {FormattedMessage, injectIntl, IntlShape} from 'react-intl';
+import {Tooltip} from 'react-bootstrap';
+
+import {Post} from 'mattermost-redux/types/posts';
+import {AppBinding} from 'mattermost-redux/types/apps';
+import {AppCallResponseTypes, AppCallTypes, AppExpandLevels} from 'mattermost-redux/constants/apps';
+import {UserThread} from 'mattermost-redux/types/threads';
+import {Team} from 'mattermost-redux/types/teams';
+import {$ID} from 'mattermost-redux/types/utilities';
+
+import {DoAppCall, PostEphemeralCallResponseForPost} from 'types/apps';
+import {Locations, Constants} from 'utils/constants';
+import OverlayTrigger from 'components/overlay_trigger';
+import * as PostUtils from 'utils/post_utils';
+import * as Utils from 'utils/utils.jsx';
+import Pluggable from 'plugins/pluggable';
+import Menu from 'components/widgets/menu/menu';
+import MenuWrapper from 'components/widgets/menu/menu_wrapper';
+import {PluginComponent} from 'types/store/plugins';
+import {createCallContext, createCallRequest} from 'utils/apps';
+
+const MENU_BOTTOM_MARGIN = 80;
+
+export const PLUGGABLE_COMPONENT = 'PostDropdownMenuItem';
+type Props = {
+    intl: IntlShape;
+    post: Post;
+    teamId?: string;
+    location?: 'CENTER' | 'RHS_ROOT' | 'RHS_COMMENT' | 'SEARCH' | string;
+    handleDropdownOpened?: (open: boolean) => void;
+    isMenuOpen?: boolean;
+    pluginMenuItems?: PluginComponent[];
+    isLicensed?: boolean; // TechDebt: Made non-mandatory while converting to typescript
+    enableEmojiPicker?: boolean; // TechDebt: Made non-mandatory while converting to typescript
+    channelIsArchived?: boolean; // TechDebt: Made non-mandatory while converting to typescript
+    appBindings?: AppBinding[];
+    appsEnabled: boolean;
+
+    /**
+     * Components for overriding provided by plugins
+     */
+    components: {
+        [componentName: string]: PluginComponent[];
+    };
+
+    actions: {
+
+        /**
+         * Function to open a modal
+         */
+        openModal: (postId: any) => void;
+
+        /**
+         * Function to perform an app call
+         */
+        doAppCall: DoAppCall;
+
+        /**
+         * Function to post the ephemeral message for a call response
+         */
+        postEphemeralCallResponseForPost: PostEphemeralCallResponseForPost;
+
+    }; // TechDebt: Made non-mandatory while converting to typescript
+
+    userId: string;
+    currentTeamId: $ID<Team>;
+    threadId: $ID<UserThread>;
+    isFollowingThread?: boolean;
+    threadReplyCount?: number;
+}
+
+type State = {
+    openUp: boolean;
+}
+
+export class ActionMenuClass extends React.PureComponent<Props, State> {
+    public static defaultProps: Partial<Props> = {
+        location: Locations.CENTER,
+        pluginMenuItems: [],
+        appBindings: [],
+    }
+    private buttonRef: React.RefObject<HTMLButtonElement>;
+
+    constructor(props: Props) {
+        super(props);
+
+        this.state = {
+            openUp: false,
+        };
+
+        this.buttonRef = React.createRef<HTMLButtonElement>();
+    }
+
+    // listen to clicks/taps on add reaction menu item and pass to parent handler
+    tooltip = (
+        <Tooltip
+            id='dotmenu-icon-tooltip'
+            className='hidden-xs'
+        >
+            <FormattedMessage
+                id='post_info.dot_menu.tooltip.more_actions'
+                defaultMessage='Actions'
+            />
+        </Tooltip>
+    )
+
+    refCallback = (menuRef: Menu) => {
+        if (menuRef) {
+            const buttonRect = this.buttonRef.current?.getBoundingClientRect();
+            let y;
+            if (typeof buttonRect?.y === 'undefined') {
+                y = typeof buttonRect?.top == 'undefined' ? 0 : buttonRect?.top;
+            } else {
+                y = buttonRect?.y;
+            }
+            const windowHeight = window.innerHeight;
+
+            const totalSpace = windowHeight - MENU_BOTTOM_MARGIN;
+            const spaceOnTop = y - Constants.CHANNEL_HEADER_HEIGHT;
+            const spaceOnBottom = (totalSpace - (spaceOnTop + Constants.POST_AREA_HEIGHT));
+
+            this.setState({
+                openUp: (spaceOnTop > spaceOnBottom),
+            });
+        }
+    }
+
+    onClickAppBinding = async (binding: AppBinding) => {
+        const {post, intl} = this.props;
+
+        if (!binding.call) {
+            return;
+        }
+        const context = createCallContext(
+            binding.app_id,
+            binding.location,
+            this.props.post.channel_id,
+            this.props.teamId,
+            this.props.post.id,
+            this.props.post.root_id,
+        );
+        const call = createCallRequest(
+            binding.call,
+            context,
+            {
+                post: AppExpandLevels.ALL,
+            },
+        );
+
+        const res = await this.props.actions.doAppCall(call, AppCallTypes.SUBMIT, intl);
+
+        if (res.error) {
+            const errorResponse = res.error;
+            const errorMessage = errorResponse.error || intl.formatMessage({
+                id: 'apps.error.unknown',
+                defaultMessage: 'Unknown error occurred.',
+            });
+            this.props.actions.postEphemeralCallResponseForPost(errorResponse, errorMessage, post);
+            return;
+        }
+
+        const callResp = res.data!;
+        switch (callResp.type) {
+        case AppCallResponseTypes.OK:
+            if (callResp.markdown) {
+                this.props.actions.postEphemeralCallResponseForPost(callResp, callResp.markdown, post);
+            }
+            break;
+        case AppCallResponseTypes.NAVIGATE:
+        case AppCallResponseTypes.FORM:
+            break;
+        default: {
+            const errorMessage = intl.formatMessage({
+                id: 'apps.error.responses.unknown_type',
+                defaultMessage: 'App response type not supported. Response type: {type}.',
+            }, {
+                type: callResp.type,
+            });
+            this.props.actions.postEphemeralCallResponseForPost(callResp, errorMessage, post);
+        }
+        }
+    }
+
+    render() {
+        const isSystemMessage = PostUtils.isSystemMessage(this.props.post);
+        const isMobile = Utils.isMobile();
+
+        console.log('pluginMenuItems', this.props.pluginMenuItems);
+        const pluginItems = this.props.pluginMenuItems?.
+            filter((item) => {
+                return item.filter ? item.filter(this.props.post.id) : item;
+            }).
+            map((item) => {
+                console.log('item', item);
+                if (item.subMenu) {
+                    return (
+                        <Menu.ItemSubMenu
+                            key={item.id + '_pluginmenuitem'}
+                            id={item.id}
+                            postId={this.props.post.id}
+                            text={item.text}
+                            subMenu={item.subMenu}
+                            action={item.action}
+                            root={true}
+                        />
+                    );
+                }
+                return (
+                    <Menu.ItemAction
+                        key={item.id + '_pluginmenuitem'}
+                        text={item.text}
+                        onClick={() => {
+                            if (item.action) {
+                                item.action(this.props.post.id);
+                            }
+                        }}
+                    />
+                );
+            }) || [];
+
+        let appBindings = [] as JSX.Element[];
+        if (this.props.appsEnabled && this.props.appBindings) {
+            appBindings = this.props.appBindings.map((item) => {
+                let icon: JSX.Element | undefined;
+                if (item.icon) {
+                    icon = (<img src={item.icon}/>);
+                }
+
+                return (
+                    <Menu.ItemAction
+                        text={item.label}
+                        key={item.app_id + item.location}
+                        onClick={() => this.onClickAppBinding(item)}
+                        icon={icon}
+                    />
+                );
+            });
+        }
+        console.log('appsBindings', appBindings);
+        console.log('pluginItems', pluginItems);
+
+        if (typeof pluginItems !== 'undefined' && pluginItems.length === 0 && isSystemMessage) {
+            return null;
+        }
+
+        return (
+            <MenuWrapper onToggle={this.props.handleDropdownOpened}>
+                <OverlayTrigger
+                    className='hidden-xs'
+                    delayShow={500}
+                    placement='top'
+                    overlay={this.tooltip}
+                    rootClose={true}
+                >
+                    <button
+                        ref={this.buttonRef}
+                        id={`${this.props.location}_button_${this.props.post.id}`}
+                        aria-label={Utils.localizeMessage('post_info.dot_menu.tooltip.more_actions', 'Actions').toLowerCase()}
+                        className={classNames('post-menu__item', {
+                            'post-menu__item--active': this.props.isMenuOpen,
+                        })}
+                        type='button'
+                        aria-expanded='false'
+                    >
+                        <i className={'icon icon-apps'}/>
+                    </button>
+                </OverlayTrigger>
+                <Menu
+                    id={`${this.props.location}_dropdown_${this.props.post.id}`}
+                    openLeft={true}
+                    openUp={this.state.openUp}
+                    ref={this.refCallback}
+                    ariaLabel={Utils.localizeMessage('post_info.menuAriaLabel', 'Post extra options')}
+                >
+                    {/* {pluginItems} */}
+                    {appBindings}
+                    <Pluggable
+                        postId={this.props.post.id}
+                        pluggableName={PLUGGABLE_COMPONENT}
+                    />
+                </Menu>
+            </MenuWrapper>
+        );
+    }
+}
+
+export default injectIntl(ActionMenuClass);
--- a/components/actions_menu/index.ts
+++ b/components/actions_menu/index.ts
@@ -1,0 +1,114 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {ComponentProps} from 'react';
+import {connect} from 'react-redux';
+import {ActionCreatorsMapObject, bindActionCreators, Dispatch} from 'redux';
+
+import {getLicense, getConfig} from 'mattermost-redux/selectors/entities/general';
+import {getChannel} from 'mattermost-redux/selectors/entities/channels';
+import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
+import {getCurrentTeamId, getCurrentTeam} from 'mattermost-redux/selectors/entities/teams';
+import {appsEnabled, makeAppBindingsSelector} from 'mattermost-redux/selectors/entities/apps';
+import {getThreadOrSynthetic} from 'mattermost-redux/selectors/entities/threads';
+import {getPost} from 'mattermost-redux/selectors/entities/posts';
+
+import {AppBindingLocations} from 'mattermost-redux/constants/apps';
+import {isSystemMessage} from 'mattermost-redux/utils/post_utils';
+import {isCombinedUserActivityPost} from 'mattermost-redux/utils/post_list';
+
+import {ActionFunc, GenericAction} from 'mattermost-redux/types/actions';
+
+import {Post} from 'mattermost-redux/types/posts';
+
+import {DoAppCall, PostEphemeralCallResponseForPost} from 'types/apps';
+
+import {GlobalState} from 'types/store';
+
+import {openModal} from 'actions/views/modals';
+import {doAppCall, postEphemeralCallResponseForPost} from 'actions/apps';
+
+import {isArchivedChannel} from 'utils/channel_utils';
+
+import {Locations} from 'utils/constants';
+
+import DotMenu from './actions_menu';
+
+type Props = {
+    post: Post;
+    handleCardClick?: (post: Post) => void;
+    handleDropdownOpened: (open: boolean) => void;
+    isMenuOpen: boolean;
+    enableEmojiPicker?: boolean;
+    location?: ComponentProps<typeof DotMenu>['location'];
+};
+
+const getPostMenuBindings = makeAppBindingsSelector(AppBindingLocations.POST_MENU_ITEM);
+
+function mapStateToProps(state: GlobalState, ownProps: Props) {
+    const {post} = ownProps;
+
+    const license = getLicense(state);
+    const userId = getCurrentUserId(state);
+    const channel = getChannel(state, post.channel_id);
+    const currentTeam = getCurrentTeam(state) || {};
+
+    const systemMessage = isSystemMessage(post);
+
+    const rootId = post.root_id || post.id;
+    let threadId = rootId;
+
+    if (
+        rootId && !systemMessage &&
+        (
+
+            // default prop location would be CENTER
+            !ownProps.location ||
+            ownProps.location === Locations.RHS_ROOT ||
+            ownProps.location === Locations.RHS_COMMENT ||
+            ownProps.location === Locations.CENTER
+        )
+    ) {
+        const root = getPost(state, rootId);
+        if (root) {
+            const thread = getThreadOrSynthetic(state, root);
+            threadId = thread.id;
+        }
+    }
+
+    const apps = appsEnabled(state);
+    const showBindings = apps && !systemMessage && !isCombinedUserActivityPost(post.id);
+    const appBindings = showBindings ? getPostMenuBindings(state) : undefined;
+
+    return {
+        channelIsArchived: isArchivedChannel(channel),
+        components: state.plugins.components,
+        isLicensed: license.IsLicensed === 'true',
+        teamId: getCurrentTeamId(state),
+        pluginMenuItems: state.plugins.components.PostDropdownMenuItem,
+        currentTeamId: currentTeam.id,
+        userId,
+        threadId,
+        appBindings,
+        appsEnabled: apps,
+        ...ownProps,
+    };
+}
+
+type Actions = {
+    openModal: (postId: any) => void;
+    doAppCall: DoAppCall;
+    postEphemeralCallResponseForPost: PostEphemeralCallResponseForPost;
+}
+
+function mapDispatchToProps(dispatch: Dispatch<GenericAction>) {
+    return {
+        actions: bindActionCreators<ActionCreatorsMapObject<ActionFunc>, Actions>({
+            openModal,
+            doAppCall,
+            postEphemeralCallResponseForPost,
+        }, dispatch),
+    };
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(DotMenu);
--- a/components/dot_menu/dot_menu.tsx
+++ b/components/dot_menu/dot_menu.tsx
@@ -271,6 +271,10 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
         </Tooltip>
     )
 
+    getKeyShortCut = (key: string): string => {
+        return (key);
+    }
+
     refCallback = (menuRef: Menu) => {
         if (menuRef) {
             const buttonRect = this.buttonRef.current?.getBoundingClientRect();
@@ -448,6 +452,8 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
                     <Menu.ItemAction
                         show={!isSystemMessage && this.props.location === Locations.CENTER}
                         text={Utils.localizeMessage('post_info.reply', 'Reply')}
+                        leftDecorator={<i className='icon icon-reply-outline'/>}
+                        rightDecorator={this.getKeyShortCut('R')}
                         onClick={this.props.handleCommentClick}
                     />
                     <ChannelPermissionGate
@@ -464,6 +470,7 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
                     <Menu.ItemAction
                         id={`follow_post_thread_${this.props.post.id}`}
                         onClick={this.handleSetThreadFollow}
+                        rightDecorator={this.getKeyShortCut('F')}
                         show={(
                             !isSystemMessage &&
                             this.props.isCollapsedThreadsEnabled &&
@@ -485,12 +492,16 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
                         id={`unread_post_${this.props.post.id}`}
                         show={!isSystemMessage && !this.props.channelIsArchived && this.props.location !== Locations.SEARCH}
                         text={Utils.localizeMessage('post_info.unread', 'Mark as Unread')}
+                        leftDecorator={<i className='icon icon-mark-as-unread'/>}
+                        rightDecorator={this.getKeyShortCut('U')}
                         onClick={this.handleUnreadMenuItemActivated}
                     />
                     <Menu.ItemAction
                         id={`permalink_${this.props.post.id}`}
                         show={!isSystemMessage}
                         text={Utils.localizeMessage('post_info.permalink', 'Copy Link')}
+                        leftDecorator={<i className='icon icon-link-variant'/>}
+                        rightDecorator={this.getKeyShortCut('K')}
                         onClick={this.copyLink}
                     />
                     <Menu.ItemAction
@@ -513,6 +524,8 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
                         id={`pin_post_${this.props.post.id}`}
                         show={!isSystemMessage && !this.props.isReadOnly && !this.props.post.is_pinned}
                         text={Utils.localizeMessage('post_info.pin', 'Pin')}
+                        leftDecorator={<i className='icon icon-pin-outline'/>}
+                        rightDecorator={this.getKeyShortCut('P')}
                         onClick={this.handlePinMenuItemActivated}
                     />
                     {!isSystemMessage && (this.state.canEdit || this.state.canDelete) && this.renderDivider('edit')}
@@ -520,6 +533,8 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
                         id={`edit_post_${this.props.post.id}`}
                         show={this.state.canEdit}
                         text={Utils.localizeMessage('post_info.edit', 'Edit')}
+                        leftDecorator={<i className='icon icon-pencil-outline'/>}
+                        rightDecorator={this.getKeyShortCut('E')}
                         onClick={this.handleEditMenuItemActivated}
                     />
                     <Menu.ItemAction
@@ -527,6 +542,8 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
                         show={this.state.canDelete}
                         text={Utils.localizeMessage('post_info.del', 'Delete')}
                         onClick={this.handleDeleteMenuItemActivated}
+                        leftDecorator={<i className='icon icon-trash-can-outline'/>}
+                        rightDecorator={this.getKeyShortCut('delete')}
                         isDangerous={true}
                     />
                     {((typeof pluginItems !== 'undefined' && pluginItems.length > 0) || appBindings.length > 0 || (this.props.components[PLUGGABLE_COMPONENT] && this.props.components[PLUGGABLE_COMPONENT].length > 0)) && this.renderDivider('plugins')}
--- a/components/dot_menu/dot_menu.tsx
+++ b/components/dot_menu/dot_menu.tsx
@@ -362,6 +362,39 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
         }
     }
 
+    onShortcutKeyDown = (e: KeyboardEvent) => {
+        switch (true) {
+        case Utils.isKeyPressed(e, Constants.KeyCodes.R):
+            this.props.handleCommentClick(e);
+            break;
+
+        // edit post
+        case Utils.isKeyPressed(e, Constants.KeyCodes.E):
+            this.handleEditMenuItemActivated();
+            break;
+
+        // copy link
+        case Utils.isKeyPressed(e, Constants.KeyCodes.K):
+            this.copyLink();
+            break;
+
+        // delete post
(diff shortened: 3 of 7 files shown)


matthewbirtch (changes requested):
Awesome work @jfrerich! I did a first pass UI/UX review and here's what I've found so far. I can do another round of reviews once these are resolved.

1. For admins, System join/leave messages appear to have the `Actions` icon added. Is that intentional? It seems like that shouldn't be there, but I don't know if there is a reason for it.
2. The App actions menu seems to have empty menu items at the top of the menu. There is extra blank space at the top and when I hover up there it displays lines.
image
3. The App Marketplace menu item that displays at the bottom of the App actions menu has a misaligned icon. It appears to be sitting too high. This may be an issue for all menu items using these icons - see note below.
image
4. When I have the `More` menu open, if I hit a key that is not a shortcut key for this menu, the menu closes. I noticed when trying to take a screenshot of the menu, but couldn't because it closes as soon as I hit another key. The menu should stay open.
5. The layout for the 'Unfollow/Follow thread' menu item is a bit messy because of the wrapping of the descriptive text for that menu item. It's also causing the icon to be vertically misaligned. I think we could consider removing the descriptive text as the menu item is fairly self-explanatory. 
image
6. 'Follow/Unfollow thread' menu item is also getting horizontally misaligned - not sure if that's the same issue as above or a different problem causing it.
6. The icons in the menu look like they are slightly misaligned vertically. They appear to be sitting too high. The `Delete` option appears to be particularly off
image

7. For the `delete` menu item, the shortcut text on the right needs be white on the hover state.
8. When an item is pinned, the 'Unpin from Channel' menu item's styling doesn't work well. Let's keep the color the same color as all other icons, but use the filled in pin icon `icon-pin`. I might create a new icon that has the outlined pin with a line through it for this purpose, but let's go with this for now.
image
9. I think we need a bit more space between the icon and the menu item label. Can we add 4px of space there?
10. Also, for the menu items here, the icon colors should be `rgba(var(--center-channel-color-rgb), 0.56);` It looks like the opacity is set to 0.5
11. The shortcuts on the right should also have their opacity set to `rgba(var(--center-channel-color-rgb), 0.56); instead of the opacity at 0.5
12. When I open the `More` menu and then click away, it opens the thread RHS - I think because where I clicked was on the message which will open the thread in the RHS if I click anywhere on the message. However, can we prevent that so that clicking away from the menu simply closes the menu and doesn't open the RHS? NOTE: This is only an issue when Collpased Reply Threads is on.
13. I tested the menu with a non-admin user and still saw the `App Martketplace` menu item. This should be hidden for end users and only visible to admins at this point.
14. Found a significant bug with the shortcuts. Steps to reproduce:
    - Open the `More` menu on a message
    - Click away from the menu on a message in the channel (other than the message that the menu was open for)
    - This will open the thread in the RHS for the message you clicked
    - The keyboard now doesn't work and is still accepting the shortcut keys from the `More` menu. I can't type in the message input
    - NOTE: this doesn't always happen. 


**Things I couldn't test on this server**
- If the server has no app actions that are applicable for messages, hide this menu option
- For admins, if there are no app actions applicable to messages, an empty state displays with a button to visit the app marketplace
image
- Tour point - for first time, I should see a tour point explaining
image

jfrerich:
Thanks for the review, @matthewbirtch! I'll get on these right away.

jfrerich:
Issues addressed from Matt's review.

1.
2.
3.
4.
5.
6.
7.
8.
9.
10.
11.
12.
13.
14.
15.

jfrerich:
> For admins, System join/leave messages appear to have the Actions icon added. Is that intentional? It seems like that shouldn't be there, but I don't know if there is a reason for it.

@matthewbirtch, can you provide a screenshot for this issue?  I'm not sure where you are seeing this occur.

matthewbirtch:
> > 1. For admins, System join/leave messages appear to have the Actions icon added. Is that intentional? It seems like that shouldn't be there, but I don't know if there is a reason for it.
> 
> @matthewbirtch, can you provide a screenshot for this issue? I'm not sure where you are seeing this occur.

Sure thing, here you go

image

jfrerich:
@matthewbirtch regarding your items you couldn't test, I labeled them A, B, and C.

For A & B, you can test them by disabling the plugins through `System Console` > `PLUGINS` > `Plugin Management`.  This will allow you to see the `Visit Marketplace` for sysadmins and the actions icon will be hidden for non-sysadmins.

They test instance is currently setup for this configuration, but you can test change it if needed.

matthewbirtch:
> For A & B, you can test them by disabling the plugins through `System Console` > `PLUGINS` > `Plugin Management`. This will allow you to see the `Visit Marketplace` for sysadmins and the actions icon will be hidden for non-sysadmins.

@jfrerich if plugins are disabled entirely, shouldn't we hide the App Actions from the hover menu altogether?

jfrerich (commented):
> > For A & B, you can test them by disabling the plugins through `System Console` > `PLUGINS` > `Plugin Management`. This will allow you to see the `Visit Marketplace` for sysadmins and the actions icon will be hidden for non-sysadmins.
> 
> @jfrerich if plugins are disabled entirely at the server level, shouldn't we hide the App Actions from the hover menu altogether?

Yes, that seems correct. I'll add that case and a a test for it.

jfrerich:
> > For A & B, you can test them by disabling the plugins through `System Console` > `PLUGINS` > `Plugin Management`. This will allow you to see the `Visit Marketplace` for sysadmins and the actions icon will be hidden for non-sysadmins.
> 
> ~@jfrerich if plugins are disabled entirely at the server level, shouldn't we hide the App Actions from the hover menu altogether?~
> 
> Ignore the comment above - I was confused. I just tested both the admin and end-user with plugins turned off for the server and it's working as expected. Thanks!

@matthewbirtch, your comment actually was valid, but not one that was spec'd out. When a server has plugins disabled, the actions icons should be hidden for everyone, correct?

matthewbirtch:
> @matthewbirtch, your comment actually was valid, but not one that was spec'd out. When a server has plugins disabled (globally), the actions icons should be hidden for everyone, correct?

Actually, ya, that's true. The `visit marketplace` empty state should only really appear for the admin if plugins are enabled, but no plugin actions are applicable at the message level

jfrerich:
> > @matthewbirtch, your comment actually was valid, but not one that was spec'd out. When a server has plugins disabled (globally), the actions icons should be hidden for everyone, correct?
> 
> Actually, ya, that's true. The `visit marketplace` empty state should only really appear for the admin if plugins are enabled, but no plugin actions are applicable at the message level

I'll add that code. Would you mind adding a comment to the figma?

jfrerich:
@matthewbirtch, to see the tutorial point, you can login as a new user.  I have an issue where it isn't easy to click the tutorial point instead of the `Actions` menu, but the logic is in place that will only show the pulsating dot until the user clicks the menu.

components/actions_menu/actions_menu.tsx
```diff
@@ -0,0 +1,381 @@
...
+        }
+    }
+
+    visitMarketplaceTip(): React.ReactElement {
+        return (
+            <SystemPermissionGate
+                permissions={[Permissions.MANAGE_SYSTEM]}
+            >
+                <div
+                    style={{
```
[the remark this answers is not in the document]
jfrerich: done. e29578f

components/actions_menu/actions_menu.tsx
```diff
@@ -0,0 +1,381 @@
...
+                        fontSize: 14,
+                    }}
+                >
+                    <FormattedMarkdownMessage
+                        id='post_info.actions.noActions'
+                        defaultMessage='No Actions currently\nconfigured for this server'
+                    />
+                </div>
+                <div
+                    style={{
```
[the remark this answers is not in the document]
jfrerich: done. e29578f

jfrerich:
> When I open the More menu and then click away, it opens the thread RHS - I think because where I clicked was on the message which will open the thread in the RHS if I click anywhere on the message. However, can we prevent that so that clicking away from the menu simply closes the menu and doesn't open the RHS? NOTE: This is only an issue when Collapsed Reply Threads is on.

@matthewbirtch, interesting find, but I don't believe this is an issue created by this new actions menu.  I tested the following on community-daily and get the same result. 

* enable threaded replies.
* click on any user reply
* observer RHS reply is opened with the message, or thread of messages.

Please confirm.  I understand why this would be the expected result in a threaded reply setup. The user wants to scan all replies in a thread and only needs to click on a thread or message in the center channel to see the replies.

If this is not what a user wants to happen, we should consider adding a user level setting to disable opening the RHS when clicking a message in the center channel.

matthewbirtch:
> @matthewbirtch, to see the tutorial point, you can login as a new user. I have an issue where it isn't easy to click the tutorial point instead of the `Actions` menu, but the logic is in place that will only show the pulsating dot until the user clicks the menu.

Ok, I created a new user on this server and got the pulsing dot to show, but couldn't open the tour point itself (as you mentioned).

However, I'm wondering about the conditions for showing the dot. This shouldn't just show for new users. This should show for all users who haven't yet seen the new menu options and haven't dismissed the tour point. I just want to make sure that logic is in place. Is that already in place?

matthewbirtch:
> @matthewbirtch, interesting find, but I don't believe this is an issue created by this new actions menu. I tested the following on community-daily and get the same result.
> 
> * enable threaded replies.
> * click on any user reply
> * observer RHS reply is opened with the message, or thread of messages.
> 
> Please confirm. I understand why this would be the expected result in a threaded reply setup. The user wants to scan all replies in a thread and only needs to click on a thread or message in the center channel to see the replies.
> 
> If this is not what a user wants to happen, we should consider adding a user level setting to disable opening the RHS when clicking a message in the center channel.

You're right @jfrerich. This is not an issue specific to this PR. The expected I have in mind is: if any menu is open and I click away from that menu, the menu should close and the thread RHS shouldn't open if I happen to click on a message in a channel to click away from the menu.

I'll treat this as a separate issue though and will add it to the CRT epic. No need for you to action on this I don't think

jfrerich:
@matthewbirtch, I am ready for another round of reviews.

matthewbirtch (changes requested):
Thanks @jfrerich! Getting closer for sure. A few more things I'm seeing:

1. Now the icons in the menus seem to be sitting 1px too _low_ to be properly aligned with the label. This may not be an issue with this PR specifically though - let me know. I'm seeing some alignment issues in other menus in the app as well. 
image

2. When I have the `More` menu open, the shortcut keys no longer seem to be working. I think because it seems to be focusing on the message input field and the letters I type are added to the message input field instead of performing the action in the menu. When this menu is open, can we unset the focus on the input field and allow the shortcuts to be performed?

3. For the delete menu item, can we drop the opacity of the shortcut `delete` on the right side to 0.56?

4. Clicking on `App Marketplace` button from the `Actions` menu didn't do anything for me (logged in as system admin)

5. For the Tour point on the `Actions` menu icon, I can't seem to get the actual tour point to open up. If I click on the pulsing dot, it just opens the actions menu and then the tour point is removed as soon as the actions menu is open.

6. In the RHS, I noticed the icon button for `Actions` does not get an active state when the menu is open

7. Also in the RHS context, I noticed that the Actions menu automatically closes on my when I hover over something else. (see gif below)


8. For the empty state on the `Actions` menu, can we adjust the following:
    - Make the button label `font-weight: 600` and `font-size: 12px`
    - Can we add 4px of space between the icon and the label in the button
    - Can we change the border radius on the button to `border-radius: 4px`
    - Can we reduce the space above the button by 6px
    - Can we increase the space under the button by 6px

jfrerich:
@matthewbirtch I'll update this as I resolve your 2nd round of feedback. I picked some low hanging fruit to get started. 

1. Move icons up 1px. 
2. shortcut keys not working. 
3.
4. App Marketplace modal click now works
5.
6. RHS actions state should be active when the menu is open
7.
8.

matthewbirtch:
> 1. Move icons up 1px. [188e526](https://github.com/mattermost/mattermost-webapp/commit/188e5261b73a08b3af92618d26b6e3c4644970d2)

Fixed! Thank you

> 2. shortcut keys not working. [`ba7493d` (#8853)](https://github.com/mattermost/mattermost-webapp/pull/8853/commits/ba7493df9fd7230d61391431221f24ab657ba9ca)

This is working again now, thanks! However, the `F`, `P`, `K` shortcuts seem to add an `f`,`k` or `p` characters in to the message input as well as perform the action.

> 3. Make `delete` shortcut text opaque`. [d4f533e](https://github.com/mattermost/mattermost-webapp/commit/d4f533e9961c782fd414e13738de87684becd824)

Not seeing the change here. The original request was to reduce the opacity on the `delete` shortcut label. I saw your note about needing assistance with the css though. Below is the specificity you'll need, but I'm not quite sure how to properly comment this isn the .scss file.

```
.Menu .MenuItem>button.MenuItem__dangerous span.MenuItem__right-decorator {
    color: rgba(var(--dnd-indicator-rgb),0.56);
}

.Menu .MenuItem>button.MenuItem__dangerous:hover span.MenuItem__right-decorator {
    color: rgba(var(--button-color-rgb),.56);
}
```

> 4. App Marketplace modal click now works [43397d9](https://github.com/mattermost/mattermost-webapp/commit/43397d92688221785efa9a8dbe5d9816209c6e5e)

Fixed!

> 6. RHS actions state should be active when the menu is open

Still not seeing this resolved
image


> 8. For the empty state on the Actions menu, can we adjust the following: [4bd85e6](https://github.com/mattermost/mattermost-webapp/commit/4bd85e6c0262b7bf70210e767464c46880db7a8b)

> i. Make the button label font-weight: 600 and font-size: 12px

So, the `font-weight` and `font-size` seems to have been applied to the text above the button. Can you revert the change to the text and apply to the label within button instead?

> ii. Can we add 4px of space between the icon and the label in the button

Done

> iii. Can we change the border radius on the button to border-radius: 4px

Done

> iv. Can we reduce the space above the button by 6px

Done

> v. Can we increase the space under the button by 6px

Done

jfrerich:
@matthewbirtch, I updated the empty state on the Actions menu. This should now match the figma design 4a55b38

jfrerich:
@matthewbirtch, I fixed issue 7. (RHS Actions modal should remain open when hovering outside the modal)

I cannot repro #6 on the test server.  
image

jfrerich:
> However, the F, P, K shortcuts seem to add an f,k or p characters in to the message input as well as perform the action
@matthewbirtch, I fixed this issue.

esethna (changes requested):
Nice work @jfrerich! Can't wait to get this in! A few things:

1. Bug in existing menus due to the style changes
image
image

2. Incorrectly configured plugin causes thin empty row above the "App Marketplace". For example I just went to the system console and turned the github plugin to "true" without properly configuring the other fields

sys admin user:
image

non-sys admin user:
image

3. I wasn't able to trigger the tip to appear. I see the pulsing dot but can't get the flyout to appear. I would expect the first time I click on the 4 square menu item that the flyout appears. Currently it just opens the menu as far as I can see. The logic for showing the pulsing dot looks good though. 



Design: 
image

matthewbirtch:
> > However, the F, P, K shortcuts seem to add an f,k or p characters in to the message input as well as perform the action

>  @matthewbirtch, I fixed this issue.

Yep, resolved!

matthewbirtch:
> @matthewbirtch, I fixed issue 7. (RHS Actions modal should remain open when hovering outside the modal)
> 
> I cannot repro #6 on the test server.

I am not longer seeing this on the test server anymore either. Consider this resolved

matthewbirtch:
> So, the `font-weight` and `font-size` seems to have been applied to the text above the button. Can you revert the change to the text and apply to the label within button instead?

Looks like the button issues and the font issue in this empty state are now resolved too. 
image

I think that resolves all the issues I identified in my last review.

components/actions_menu/actions_menu.scss
```diff
@@ -0,0 +1,35 @@
...
+    padding-right: 16px;
+    padding-bottom: 16px;
+    padding-left: 16px;
+    border-radius: 4px;
+    font-size: 12px;
+    text-align: center;
+}
+
+.VisitMarketplaceButton {
+    border-radius: 6px;
```
matthewbirtch, line 22: @jfrerich as noted in my comment, this border radius should be `border-radius: 4px`

jfrerich:
@esethna, I'll update as I go, but for now, I have resolved your 2nd item.  please confirm.

1. - [ ] Bug in existing menus due to the style changes

2. - [x] no plugins with post menu items registered will now to the following

- sys admin user: 
  - empty row removed
  - App Marketplace menu item is replaced with visit marketplace modal
- non-sys admin user:
  - Actions menu does not appear 

3. - [ ] I wasn't able to trigger the tip to appear. I see the pulsing dot but can't get the flyout to appear. I would expect the first time I click on the 4 square menu item that the flyout appears. Currently it just opens the menu as far as I can see. The logic for showing the pulsing dot looks good though.

jfrerich:
@larkox, I'm adding you to this PR because it involves a PR you recently merged. 

The dot menu no longer has the plugin menu items or app menu items. This PR adds a new Actions menu that will hold those items.  This commit e348685 contains the changes I made after merging the latest master branch that includes your PR https://github.com/mattermost/mattermost-webapp/pull/8439/files.  I just wanted to get your approval regarding your PR code inside the context of this PR.  Thanks!

components/actions_menu/actions_menu.tsx
```diff
@@ -0,0 +1,403 @@
...
+        let tutorialTip = null;
+        if (this.props.showTutorialTip) {
+            tutorialTip = (
+                <ActionsTutorialTip/>
+            );
+        }
+
+        const hasApps = appBindings.length > 0;
+        const hasPluggables = this.props.components[PLUGGABLE_COMPONENT] && this.props.components[PLUGGABLE_COMPONENT].length > 0;
+        const hasPluginItems = typeof pluginItems !== 'undefined' && pluginItems.length > 0;
```
larkox, line 347: Why not `Boolean(pluginItems?.length)`?

larkox: Same for the previous line.

components/actions_menu/actions_menu.tsx
```diff
@@ -0,0 +1,403 @@
...
+                pluginItems,
+                appBindings,
+                pluggable,
+                marketPlace,
+            ];
+        } else {
+            menuItems = [this.visitMarketplaceTip()];
+        }
+
+        return (
```
larkox, line 366: Do we need any check or logic when menuItems length is 0?

components/post_view/post_info/post_info.test.tsx
```diff
@@ -40,12 +40,18 @@ describe('components/post_view/PostInfo', () => {
...
+            setActionsMenuInitialisationState: jest.fn(),
         },
         shouldShowDotMenu: true,
+        firstTimeActionsMenuOpened: false,
+        shouldShowActionsMenu: true,
         isReadOnly: true,
         collapsedThreadsEnabled: false,
     };
 
+    // test('should not show actions menu, plugins disabled', () => {
```
larkox, line 52: Leaving a not here in to confirm whether this test will be implemented or these comments will be removed.

components/rhs_root_post/rhs_root_post.jsx
```diff
@@ -201,12 +214,28 @@ export default class RhsRootPost extends React.PureComponent {
...
+        });
+    };
+
+    handleActionsMenuOpened = (open) => {
+        this.setState({showActionsMenu: open});
+    };
+
     handleFileDropdownOpened = (isOpened) => {
         this.setState({
             fileDropdownOpened: isOpened,
```
larkox, line 241: It would be nice to unify variable naming and style.

larkox (changes requested):
Thank you for asking the review Jason. I also took the liberty of reviewing the rest of the PR.

There is one big detail to check, that will require also UX input, but the rest seems mainly fine. Just a few style changes here and there.

components/actions_menu/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    const {post} = ownProps;
+
+    const userId = getCurrentUserId(state);
+    const currentTeam = getCurrentTeam(state) || {};
+
+    const systemMessage = isSystemMessage(post);
+
+    const apps = appsEnabled(state);
+    const showBindings = apps && !systemMessage && !isCombinedUserActivityPost(post.id);
+    const appBindings = showBindings ? getPostMenuBindings(state) : undefined;
```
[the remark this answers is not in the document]
matthewbirtch: I'm not sure I full understand the issue, but it sounds like there are performance concerns with this feature in search results specifically (or any lists where posts are across multiple channels). 

I have strong feelings that we cannot show the actions menu if there are no actions in it.

For infinite scroll lists, Can we just fetch the bindings for the search results that are initially loaded? And then on each lazy load when the user scrolls, fetch the bindings for the next results? Just an idea. I have no idea if any of that is feasible.

jfrerich:
@larkox 3e00206 resolves several of your PR feedback items. For some reason, I cannot comment directly to you RP comments

jfrerich:
@larkox, I will not leave any empty tests and will be adding or removing any commented tests.

components/actions_menu/index.ts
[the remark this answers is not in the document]
larkox: That would be better, but still not good. Imagine we load 10 posts. That would be 10 calls to the server. The server then has to get the bindings for those 10 calls, which implies sending a request to all installed apps. If we have 4 installed apps, that is 40 requests. And I am talking with low numbers.

I feel that is we were to go with that approach, it would still be an important performance problem.

components/rhs_root_post/rhs_root_post.jsx
```diff
@@ -226,19 +226,19 @@ export default class RhsRootPost extends React.PureComponent {
...
-    handleDotMenuOpened = (isOpened) => {
+    handleDotMenuOpened = (open) => {
         this.setState({
-            showDotMenu: isOpened,
+            showDotMenu: open,
         });
     };
 
     handleActionsMenuOpened = (open) => {
         this.setState({showActionsMenu: open});
```
larkox, line 236: Can you unify also the style? (whether the setState is a one-liner or several lines)

components/actions_menu/actions_menu.tsx
jfrerich: If there are no pluginItems, apps, or pluggable, menuItems will be set to this.visitMarketplace(). It will never be length === 0

components/actions_menu/actions_menu.scss
jfrerich: resolved in a87d771

components/widgets/menu/menu_items/menu_item.scss
```diff
@@ -19,26 +19,56 @@
...
         .MenuItem__icon {
             width: 20px;
-            margin-right: 8px;
+            margin-top: 4px;
+            margin-right: 12px;
+            font-size: 18px;
             text-align: center;
         }
 
+        .MenuItem__compass-icon-opaque {
```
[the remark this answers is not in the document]
jfrerich: I remove `-opaque`. It wasn't necessary.  3ee054f

components/rhs_root_post/rhs_root_post.jsx
jfrerich: If setState is only setting one field/value, I made it a one-liner 00128f0

components/actions_menu/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const getPostOptionBinding = makeGetPostOptionBinding();
+
+const fetchBindings = makeFetchBindings(AppBindingLocations.POST_MENU_ITEM);
+
+function mapStateToProps(state: GlobalState, ownProps: Props) {
+    const {post} = ownProps;
+
+    const userId = getCurrentUserId(state);
+    const currentTeam = getCurrentTeam(state) || {};
```
koox00, line 47: I think this could be removed, since we also have in the return object `teamId: getCurrentTeamId(state)`

selectors/actions_menu.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {GlobalState} from 'types/store';
+import {get} from 'mattermost-redux/selectors/entities/preferences';
+import {Preferences} from 'mattermost-redux/constants';
+
+export function showActionsMenuPulsatingDotPulsatingDot(state: GlobalState): boolean {
```
koox00, line 8: Is the double PulsatingDot in the naming on purpose or by mistake? 😄

koox00 (changes requested):
just some comments here and there.
Great job 💯

components/actions_menu/index.ts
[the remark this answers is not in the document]
matthewbirtch: Is there another option @larkox? I'd still like to avoid showing the menu if it's empty, but I don't have another idea how we might do that while preventing the performance issues you'd identified.

If this is just isolated to search results, can we show the empty menu only for search results? In the channel or thread context, can we still hide the menu if there are no bindings?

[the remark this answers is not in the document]
larkox: This will be a problem in any place that you have a list of posts along several channels. That means:
- Search results
- Recent mentions (that are actually a search result)
- Saved Posts

Any other context should be fine to hide the actions button.

I cannot think on any other option to solve this.

components/actions_menu/actions_menu.tsx
```diff
@@ -0,0 +1,403 @@
...
+         * Function to get the post menu bindings for this post.
+         */
+        fetchBindings: (userId: string, channelId: string, teamId: string) => Promise<{data: AppBinding[]}>;
+
+    }; // TechDebt: Made non-mandatory while converting to typescript
+}
+
+type State = {
+    openUp: boolean;
+    appBindings?: AppBinding[];
```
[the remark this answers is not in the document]
larkox: Let me chime in for this one.

The idea is that app bindings are available in redux only for the RHS and the central channel. Among searches, each post may have its own set of bindings (since they belong to different channels). Therefore, if there are no bindings in the props, it will fetch them and store them in the state.

selectors/actions_menu.ts
jfrerich: Thanks for catching this one! :) That's an embarrassing function name

jfrerich: fixed ec95b79

components/actions_menu/index.ts
jfrerich: done 3486130

jfrerich:
@esethna, I've been 100% working on gekidou for a while.  This PR is etremely close.  I'll take another look at it tomorrow for open source Friday, fix the conflicts and see what remains to be done.  Does that sound ok?

esethna:
@jfrerich sounds great to me, thanks for getting across the finish line!

jfrerich:
@esethna and @matthewbirtch you can check out the lastest commit and see the new tooltip is implemented. 

1. Tip works when used from the post, RHS Comment and RHS Root Post.
2. The actions menu is disabled until the user has clicked the `Got It` button in the tool tip
3. After clicking the `Got It` button, the pulsating dot is removed from all three locations and the actions menu is now active.

I found the easiest method to test and show the pulsating dot again is to remove the record in the server `Preferences` table for your userID, Category = 'actions_menu' and value of `{"actions_menu_modal_viewed":true}`

esethna:
Thanks @jfrerich! Can you help get the test server up and running?

jfrerich:
@esethna the test server has been created! I tested out the tooltip for the sysadmin user, but the server is open creating new users!

components/actions_menu/actions_menu_tutorial_tip.tsx
```diff
@@ -44,6 +44,7 @@ export const ActionsTutorialTip = ({handleOpen, handleDismiss, handleNext, showT
     return (
         <TourTip
             show={showTip}
+            onHidden={handleDismiss}
```
AshishDhama, line 47: @jfrerich This will create problem with other use cases,
I think would be better if use useEffect hook in the component 
```
  useEffect(() => {
        return () => {
            function here...
        };
    }, []);
    ```
    Have done it here 
    https://github.com/mattermost/mattermost-webapp/blob/175f8ebf5c03c0fae1ca0e808e32d39914c687b8/components/onboarding_tasks/complete_your_profile_tour_tip.tsx#L25

File: components/widgets/tour_tip/tour_tip.tsx [hunk 15]
```diff
@@ -206,6 +208,7 @@ const TourTip = ({
             {show && (
                 <Tippy
                     showOnCreate={show}
+                    onHidden={onHidden}
```
AshishDhama, line 211: Check if above solution works for you, we mostly will be replacing tippy with popper/floating UI so trying to expose minimum of tippy api, if not please ignore above comment but I think that should work I am doing something similar.

matthewbirtch (changes requested):
@jfrerich a couple things I just noticed:

1. I created a new user and don't see the new app actions icon when hovering over a post, but for the system admin user I do see it. Is there some kind of permissions-based display of this option?
2. Because of the above, I was not able to test the tour tip
3. If I try to open the ••• menu on a post that is at the bottom of the viewport, the menu opens down instead of up and is out of view. Not sure if this bug was introduced in a different pr though

esethna:
Looks like there's some funkiness with other menus as well. ie profile pop-over is appearing below the RHS:

https://gyazo.com/e14bb6c976e1e80089691629943d290e

esethna (changes requested):
@jfrerich same issues as Matt identified, and it also looks like there's some funkiness with other menus as well. ie profile pop-over is appearing below the RHS:

https://gyazo.com/e14bb6c976e1e80089691629943d290e

I love the keyboard shortcuts in the post menu though :) Nice work

jfrerich:
@esethna and @matthewbirtch thanks for the feedback.  I will solve these cases asap.  I am also adding tests today

AshishDhama:
> Looks like there's some funkiness with other menus as well. ie profile pop-over is appearing below the RHS:
> 
> https://gyazo.com/e14bb6c976e1e80089691629943d290e

@esethna This issue we had on master, got merged a fix today.  @jfrerich As soon you rebase to master this should not Appear

jfrerich:
Thanks @AshishDhama! I'll merge that fix into this PR soon.

esethna:
@jfrerich did you figure out a path forward for what was blocking you or are we okay to move ahead with merging this and solving it afterwards (ie empty menu issue)?

jfrerich:
@esethna, @koox00 and I have a possible solution, but would like to propose it after merging this first.  We are ok with merging with the empty menu.

esethna:
+1 thanks @jfrerich!

jfrerich:
@esethna, I'm merging latest master right now.  I spent time on trying to fix that empty menu issue and will leave it out for now.  I'll resolve the other issues which are minor, spin up another test server and if you give the green light, we'll merge!

esethna:
Thanks @jfrerich, FYI I'll be out starting next week so please don't block merging on my review :)

jfrerich:
> > Looks like there's some funkiness with other menus as well. ie profile pop-over is appearing below the RHS:
> > [gyazo.com/e14bb6c976e1e80089691629943d290e](https://gyazo.com/e14bb6c976e1e80089691629943d290e)
> 
> @esethna This issue we had on master, got merged a fix today. @jfrerich As soon you rebase to master this should not Appear

This is fixed after latest merge of master :)

jfrerich:
> @jfrerich a couple things I just noticed:
> 
> 1. I created a new user and don't see the new app actions icon when hovering over a post, but for the system admin user I do see it. Is there some kind of permissions-based display of this option?
> 2. Because of the above, I was not able to test the tour tip

@matthewbirtch, can you restest this issue?  It should be resolve now.

components/actions_menu/actions_menu_tutorial_tip.tsx
jfrerich: @AshishDhama, I am going to push this out to another Jira ticket that will request this change and also the issue issue about the empty actions menu.  This way we an merge asap.

jfrerich:
I just created https://mattermost.atlassian.net/browse/MM-42481 to track the two remaining items from this review.  If @matthewbirtch and @esethna approve, we can finally merge.  The test server is up now!

jfrerich:
> Also, since playbooks is on by default, is the only case of the "empty more actions menu" going to be if playbooks is manually disabled (since it's enabled by default)?
That is correct. If playbooks is always installed and on by default the emtpy menu would not appear.

jfrerich:
> @jfrerich thanks! I'm seeing the issue of the menu being hidden below the input (not reproing on master)

Looking into this right now!

jfrerich:
@esethna, Give it another go please!

esethna (approved):
Looks great from my view, thanks @jfrerich, nice work!

jfrerich:
@matthewbirtch, @larkox, @koox00, can y'all do another review when you get a chance.  We are getting close to finally merging this feature :)

matthewbirtch:
@jfrerich it seems that the actions menu items takes a few clicks some times to actually open up the tour point popover. I think it's because the pulsing do is the thing has to be clicked, but the button itself doesn't work while the pulsing dot is on top of it.


Also, can we adjust the positioning slightly for the tour point and pulsing dot?
- Can we move the pulsing do up by 6px? It obscures the icon too much I think.
- The arrow for the tooltip doesn't quite point to the center of the dot. It seems like it needs to bump to the left by a few x.
image

Other than that, I think it's looking good1

jfrerich:
> @jfrerich it seems that the actions menu items takes a few clicks some times to actually open up the tour point popover. I think it's because the pulsing do is the thing has to be clicked, but the button itself doesn't work while the pulsing dot is on top of it.

This is how I currently have it designed.  The actions menu will not work until the user clicks the tutorial tip.  How would you like it to respond?  

We could do:

1) as is actions menu disabled until user clicks tip
2) Action Menu allowed to work even if the user has not acknowleged the tip
3) clicking the Actions Menu or the Pulsating dot will open the tutorial point?  
4) I could merge as is and modifiy after some user feedback :)

jfrerich:
@matthewbirtch please see above comment.  I just pushed a commit that moves the dot up vertically 6px.  I am not seeing the issue with the arror portion of the tip not aligning to the dot.  Please verify if that is still happening.  Thanks!  Almost there!

components/actions_menu/actions_menu.tsx
```diff
@@ -0,0 +1,443 @@
...
+                        show={true}
+                        text={formatMessage({id: 'post_info.marketplace', defaultMessage: 'App Marketplace'})}
+                        icon={Utils.getMenuItemIcon('icon-view-grid-plus-outline')}
+                        onClick={this.handleOpenMarketplace}
+                    />
+                </>
+            );
+        }
+
+        if (isSystemMessage) {
```
larkox, line 360: Does it make sense to get this near the beginning of the function?

larkox (changes requested):
Some minor changes.

components/post_view/post_info/post_info.tsx
```diff
@@ -189,6 +208,29 @@ export default class PostInfo extends React.PureComponent<Props, State> {
...
+        this.setState({showActionsMenu: open});
+    };
+
+    handleActionsMenuTipOpened = (): void => {
+        this.setState({showActionTip: true});
+        this.props.handleDropdownOpened(true);
+    };
+
+    handleActionsMenuGotItClick = (): void => {
+        this.props.actions.setActionsMenuInitialisationState?.(({[Preferences.ACTIONS_MENU_VIEWED]: true}));
```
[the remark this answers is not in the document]
jfrerich: done.

components/post_view/post_info/post_info.tsx
```diff
@@ -206,7 +254,10 @@ export default class PostInfo extends React.PureComponent<Props, State> {
         if (showCommentIcon) {
             commentIcon = (
                 <CommentIcon
-                    handleCommentClick={this.props.handleCommentClick}
+                    handleCommentClick={(e) => {
```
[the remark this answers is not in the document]
jfrerich: done.

components/actions_menu/actions_menu.tsx
jfrerich: yes. done.

components/actions_menu/actions_menu.tsx
```diff
@@ -0,0 +1,443 @@
...
+        if (this.props.showPulsatingDot || this.props.showTutorialTip) {
+            e.preventDefault();
+            e.stopPropagation();
+        }
+    };
+
+    render(): React.ReactNode {
+        const isSystemMessage = PostUtils.isSystemMessage(this.props.post);
+
+        // const isMobile = this.props.isMobileView TODO;
```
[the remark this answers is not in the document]
jfrerich: added to cleanup jira ticket.

jfrerich:
@matthewbirtch, if the pulsating tip is showing, now click on the dot or the actions menu will open the tour tip. 

@AshishDhama do you know why this tip point looks weird? 

image

jfrerich:
For reference, I do not see the tour tip point issue on my local machine.  In both cases, I am viewing on chrome.  This is a screen can on my local box 
image

matthewbirtch:
> @matthewbirtch, if the pulsating tip is showing, now click on the dot or the actions menu will open the tour tip.
> 
> @AshishDhama do you know why this tip point looks weird?
> 
> image

Click works better now @jfrerich. Thanks.

I cleared my cache and I'm not seeing the weird display issue with the arrow anymore, but it's still showing misaligned by a few px.
image

components/actions_menu/actions_menu.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        actions: {
+            doAppCall: jest.fn(),
+            openModal: jest.fn(),
+            openAppsModal: jest.fn(),
+            postEphemeralCallResponseForPost: jest.fn(),
+            fetchBindings: jest.fn(),
+        },
+    };
+
+    // test('should have divider when plugin menu item exists', () => {
```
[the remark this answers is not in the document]
jfrerich: done.

jfrerich:
I have tried many different options to get that arrow to align correctly, but the X direction in the Tippy component does not seem to matter. When providing the Y offset, the tip does move correctly.

AshishDhama:
@jfrerich is it fixed?

jfrerich:
@AshishDhama No, I could not get the arrow to offset in the X direction.  

@matthewbirtch could we push that into another ticket so it doesn't prevent merging?

matthewbirtch:
> @AshishDhama No, I could not get the arrow to offset in the X direction.
> 
> @matthewbirtch could we push that into another ticket so it doesn't prevent merging?

Ya, I'm good with that

matthewbirtch (approved):
Approving. Follow-up ticket to be created for the position of the tour point

jfrerich:
@larkox and @koox00, can y'all check your requests were resolved and approve if satisfactory :) Almost there with this feature

jfrerich:
FYI Clean up Jira ticket is here https://mattermost.atlassian.net/browse/MM-42481

larkox (approved):

koox00 (approved):
Awesome work! excited to have this merged 💪

components/dot_menu/dot_menu.tsx
```diff
@@ -179,36 +163,33 @@ export class DotMenuClass extends React.PureComponent<Props, State> {
...
         this.disableCanEditPostByTime();
     }
 
-    componentDidUpdate(prevProps: Props) {
-        if (!this.state.appBindings && this.props.isMenuOpen && !prevProps.isMenuOpen) {
-            this.fetchBindings();
+    componentDidUpdate(prevProps: Props): void {
+        if (!prevProps.isMenuOpen && this.props.isMenuOpen) {
+            window.addEventListener('keydown', this.onShortcutKeyDown);
+            window.addEventListener('keyup', this.onShortcutKeyUp);
```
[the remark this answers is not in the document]
jfrerich: @koox00, I think you are correct. Does the following look correct?  And for clarification, we only need to add the handlers when the component is mounted and removed them after unmounting.  For this reason, we don't need to keep checking the props AND use componentDidUpdate.

```ts
  componentDidMount(): void {
        this.disableCanEditPostByTime();
        window.addEventListener('keydown', this.onShortcutKeyDown);
        window.addEventListener('keyup', this.onShortcutKeyUp);
    }

    componentWillUnmount(): void {
        window.removeEventListener('keydown', this.onShortcutKeyDown);
        window.removeEventListener('keyup', this.onShortcutKeyUp);
        this.editDisableAction.cancel();
    }
```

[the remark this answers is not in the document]
jfrerich: versus the following...
```ts
    componentDidUpdate(prevProps: Props): void {
        if (!prevProps.isMenuOpen && this.props.isMenuOpen) {
            window.addEventListener('keydown', this.onShortcutKeyDown);
            window.addEventListener('keyup', this.onShortcutKeyUp);
        }

        if (prevProps.isMenuOpen && !this.props.isMenuOpen) {
            window.removeEventListener('keydown', this.onShortcutKeyDown);
            window.removeEventListener('keyup', this.onShortcutKeyUp);
        }
    }
```

aaronrothschild:
@jfrerich Nit: could we change the hover tip text from "Actions" to "Message Actions"?  I'm only worried about conflicting with the new "Channel Actions" that is being added to channel menus.  This would better define what the new menu is... "Message Actions".   cc @matthewbirtch @cwarnermm

jfrerich:
@aaronrothschild, your comment missed the merging boat.  Can you bring this up in some channel or ticket when the agreed upon change is requested?

matthewbirtch:
@jfrerich I saw this was merged, but did it go through QA review? I don't think I saw anything from the QA folks in the PR

furqanmlk:
Lot of e2e are failing

jfrerich:
@saturninoabril, @matthewbirtch, @furqanmlk, y'all are correct and I completely forgot to add the QA label and wait. What is the appropriate action?  Do we need to revert?

saturninoabril:
Thanks @jfrerich for this really good feature!
I have asked @DHaussermann to do QA on community-daily or in a test server since this was already merged. I'll defer to him and maybe reports from community users (depending on severity/occurrence) whether we'll revert the change or not. I've created two tickets and assigned to you to address the following:
- https://mattermost.atlassian.net/browse/MM-42584 - to fix broken Cypress test (priority), and;
- https://mattermost.atlassian.net/browse/MM-42585 - to add new test both in Zephyr and Cypress.

aaronrothschild:
@jfrerich I made a PR on a branch here so we don't lose track of it: [pr-mattermost-webapp-10021](https://github.com/mattermost/mattermost-webapp/pull/10021)

PR #8857, opened 2021-09-15
[MM-38457] - Subscribe now CTA missing from Global header menus

AGMETEOR:
#### Summary
This PR adds Subscribe Now CTA to the global header when cloud instance is still in trial

#### Ticket Link
[MM-38457](https://mattermost.atlassian.net/browse/MM-38457)

#### Screenshots
Screenshot 2021-09-15 at 13 38 37

#### Release Note

```release-note
NONE

```


pablovelezvidal (approved):
LGTM! thanks @AGMETEOR

stevemudie (approved):
Tested and passed.
CTA is in the Global Header as expected!

components/widgets/menu/menu_items/menu_top_notification.tsx
```diff
@@ -2,15 +2,34 @@
...
 import './menu_item.scss';
 
 type MenuTopNotificationProps = {
-    show: boolean;
-    id?: string;
-    children?: React.ReactNode;
+    id: string;
 }
-const MenuTopNotification: React.FC<MenuTopNotificationProps> = ({id, show, children}: MenuTopNotificationProps) => {
+const MenuTopNotification: React.FC<MenuTopNotificationProps> = ({id}: MenuTopNotificationProps) => {
```
marianunez, line 21: `MenuTopNotification` seems like a generic class but now that it is specific to cloud trial prompt. Can we rename to something specific like `MenuCloudTrial`?

components/widgets/menu/menu_items/menu_top_notification.tsx
```diff
@@ -2,15 +2,34 @@
...
+    const subscription = useSelector((state: GlobalState) => state.entities.cloud.subscription);
+    const license = useSelector(getLicense);
+
+    const isCloud = license.Cloud === 'true';
```
marianunez, line 25: We should add a null check here just in case `license` doesn't exist
```suggestion
    const isCloud = license?.Cloud === 'true';
```

components/widgets/menu/menu_items/menu_top_notification.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+                    },
+                },
+            },
+        };
+        const store = mockStore(state);
+        const wrapper = mountWithIntl(<Provider store={store}><MenuTopNotification id='topNotification'/></Provider>);
+        expect(wrapper.find('UpgradeLink').exists()).toEqual(true);
+    });
+
+    test('should NOT render when NOT on cloud license and NOT during free trial period', () => {
```
marianunez, line 40: Great tests! Let's add a test that it should not render if no license as well

marianunez (changes requested):

components/widgets/menu/menu_items/menu_top_notification.test.tsx
AGMETEOR: Added a new test case to cover it

marianunez (approved):
Changes look good for the case the Global Header is always available. However, this will be the case for the Oct. 12th Cloud release. 

Let's merge this AFTER the FF for cloud happens Monday so it does not make it to the Sept 29th cloud release. Adding a `Do Not Merge` for now

marianunez:
@AGMETEOR this is ready to merge for Oct 12th Cloud release. Some tests were failing that need to be checked out

PR #8861, opened 2021-09-15
MM-37187 Migrate Pluggable to TypeScript

hmhealey:
One of the dependencies that I updated (I'm honestly not sure which one) caused the TypeScript compiler to complain about `Pluggable` when running `npm run check-types` because we were passing arbitrary props to it depending on the location of the `Pluggable` (like the `postId` being passed for post-related components).

To solve that, I migrated it to TypeScript which required some changes to the surrounding code, and because I'm me, I ended up rewriting it as a functional component and removing some unused functionality (passing a default component as the `Pluggable`'s child).

From what I can tell between the tests and some brief playing around with the Playbooks plugin (clicking the header button, seeing the RHS, and opening the full screen view), this seems to be working, but I'd like some eyes on this that know the system better.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37187

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/8860

#### Release Note
```release-note
NONE
```


plugins/pluggable/pluggable.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    if (!pluggableName || !Object.hasOwnProperty.call(components, pluggableName)) {
+        return null;
+    }
+
+    // Override the default component with any registered plugin's component
+    // Select a specific component by pluginId if available
+    let content;
+
+    if (pluggableName === 'Product') {
```
crspeller, line 66: Why do we have different behavior here for `Product` now? Can't we just support the subcomponent generically?

crspeller (changes requested):

plugins/pluggable/pluggable.tsx
hmhealey: I had to do that to satisfy the typing because that was the only way for me to tell apart when something is a `ProductComponent` instead of a `PluginComponent`. With the way those types are set up, we can't just access arbitrary fields for arbitrary subcomponents. If data was stored in redux in a different way, that'd be doable, but I don't know what the potential side effects of changing that structure would be.

I'm honestly not 100% sure what the purpose of subcomponents is. It seems to be for letting a plugin register multiple components as a batch which makes sense to me, but the structure of the data in the store ends up being confusing since we're mixing types. I think the structure of `Pluggable` somewhat reflects that, especially in the previous version that also supported completely replacing a component (a feature that we haven't used in a long time).

crspeller: I think I created sub-components as a response to the types being all messed up back there and I couldn't do the solution I wanted. Maybe it would be worth fixing all that. This solution is not ideal because it makes the feature work for only one pluggable, without mentioning that in the docs above.

hmhealey: I've tried figuring out how to reintegrate that code path, but I wasn't able to make arbitrary fields for subcomponents work. I think we'll need to restructure the plugin state to do that. Ideally, ProductComponent and PluginComponent would be similar enough then to offer subcomponents for both, but as it stands, I don't think I can get that working right now, especially since we don't have anywher that we want Pluggables to use subcomponents yet.

crspeller (approved):
Alright. Approving to unblock, but we should at least have a ticket to fix this properly.

hmhealey:
I filed https://mattermost.atlassian.net/browse/MM-38806 to re-add that, and I fixed the reviewers since I could've sworn I had assigned it to someone on Boards previously.

jespino (approved):
LGTM

jespino:
@hmhealey Sorry for the delay, I haven't added the QA for this because I don't know who is the right person.

hmhealey:
@DHaussermann For this, I've migrated the Pluggable component (used wherever plugins inject a component into the web app) to TypeScript which shouldn't result in any changes, so we'd at least want to double check that that still works in places like the sidebar and global header. Let me know if I should be assigning someone else to this though.

PR #8862, opened 2021-09-15
Refactor isMobile to use redux and change name to isMobileView for better clarity.

deanwhillier:
#### Summary
This PR adds window resize tracking to Redux with several selectors for getting the current window size and determining if the current window size is `mobile`, `tablet`, `small desktop` or `desktop` corresponding to our current primary CSS media query sizes.

In addition, a few components are updated to use the new selectors in place of their existing custom window resize listeners. Included in this PR are updates to simple uses of custom window resize listeners. Further PR's will address some of the more complex uses of resize listeners and the uses of `Utils.isMobile()` that the new selectors replace.

The existing `isMobile` name has been changed to `isMobileView` to help with clarity. isMobileView speaks more to describing the MM defined view size(s) rather than an arbitrary device size.

Relevant targeted commits for easier review are as follows:
- **https://github.com/mattermost/mattermost-webapp/commit/fc200fc3ae43374322b49773832387a36b72fd04**: the actual implementation of the redux functionality
- **https://github.com/mattermost/mattermost-webapp/commit/b5b06c64aab323ff8bdb15db225637ebcd9129f0**: Global Header updates
- **https://github.com/mattermost/mattermost-webapp/commit/ab1a9e53f45de44dccc69676fcc3c312fee5145b**: Sidebar updates
- **https://github.com/mattermost/mattermost-webapp/commit/a8530ad2263dfcbd64a893994d4c9437d3f6bef7**: Updates to Sidebar header when Global Header is off.
- **https://github.com/mattermost/mattermost-webapp/commit/78aff665e40b8e4dd7e9aaa5cfbcd1cca15b8334**: RHS Search Nav component updates
- **https://github.com/mattermost/mattermost-webapp/commit/73d919d49590327a75f1305ea4aaacf2c5fb5758**:  File preview modal component updates
- **https://github.com/mattermost/mattermost-webapp/commit/241f66da1f15a7d6d28e6daa76653071db3c7819**: Virtual thread viewer component updates

#### Release Notes

```release-note
NONE
```


components/root/root.jsx
```diff
@@ -248,11 +250,16 @@ export default class Root extends React.PureComponent {
             this.onConfigLoaded();
         });
         trackLoadTime();
+
+        window.addEventListener('resize', this.handleWindowResizeEvent);
+        this.props.actions.emitBrowserWindowResized();
```
crspeller, line 255: Is this going to cause performance issues? Hitting redux like this for every resize event will cause a lot of stuff to re-render or at least a whole bunch of redux calculations. A resize operation could cause a lot of slowdown.

deanwhillier: It certainly would, except that the function is throttled on line 290 :)

deanwhillier: I still kinda like the CSS media query approach because CSS is handled separately from JS by the browser. I understand the concern over having additional markup that is hidden for when in mobile view, but IMO its minimal and given the browser doesn't consider the markup or styling when `display: none;` is used, it's not a huge performance issue. I guess we need to decide what is better for our situation ... again, open to thoughts on this.

deanwhillier: Also, @harrison and I talked about computed selectors vs. booleans stored in redux and we went with computed selectors for the `isMobileView` etc., but now I'm wondering if it might actually be more performant to store them in redux and only update them when needed instead of computing them based on the dispatched window size updates?

Meaning, if I grab the window size in the action and return a boolean for `isMobileView` based on that size and then get the reducer to decide if it needs to update the store based on whether the value is different, wouldn't that be more performant than updating width/height values and computing the boolean for `isMobileView` each time, thereby forcing component updates more frequently?

crspeller: I agree that this seems like something that should be handled by CSS. Having the browser handle it would remove any performance concerns around redux. Why did we not keep that approach? 

You would have to not even tell redux about the change. If you dispatch an action I believe even it it doesn't effect a state change redux still sends an update to everyone. But maybe in the action you could check the current state and not send the update at all if they are the same?

components/global/hooks.tsx
```diff
@@ -85,3 +86,7 @@ export const useIsModalOpen = (modalIdentifier: string): [boolean, React.RefObje
 
     return [modalOpenState, modalOpenStateRef];
 };
+
+export const useIsMobileView = (): boolean => {
+    return Boolean(useSelector(getIsMobileView));
```
michelengelen, line 91: why the extra `Boolean()` cast?

michelengelen (approved):
code-wise this looks good ... I only share the performance concerns @crspeller mentioned. Throttling might not be enough to prevent a bottleneck. I just do not have a better solution right now.

One thing that might have an impact could be to only apply the redux update when the window width changes, since we do not really care about the height. Thoughts?

components/root/root.jsx
hmhealey: I was thinking about just storing `isMobileView` instead of the window dimensions, but I thought we found some cases which needed the actual dimensions. If those are limited though, we could re-add listeners to those components and have them store the page size in the state. Storing both dimensions and `isMobileView` kind of defeats the purpose of that though.

I don't know if triggering redux updates when resizing is a problem at the moment since resizing should be rare (compared to just using the app), and hopefully the rest of the app is performant enough that that's not a problem (since redux actions are firing all the time). If that is a concern, throttling should hopefully help though.

deanwhillier: I'm working on a slight alternative along those lines, will push here soon. :)

hmhealey: Unrelated to all that (apparently I can't start another thread here), but is calling `emitBrowserWindowResized` on mount to set the initial values in the redux state? I wonder if that delay might cause any issues. Accessing the window in the reducer isn't great, but it would solve any potential delay problems

selectors/views/browser.ts
```diff
@@ -0,0 +1,31 @@
...
+import {Constants} from 'utils/constants';
+
+import {GlobalState} from 'types/store';
+import {ViewsState} from 'types/store/views';
+
+export function getBrowserWindowSize(state: GlobalState): ViewsState['browser']['windowSize'] {
+    return state.views.browser.windowSize;
+}
+
+export function getIsDesktopView(state: GlobalState): boolean {
```
hmhealey, line 13: This might be nitpicking, but I tend to leave off the `get` for selectors that get boolean values. That's because I'm used to the Java convention of boolean getters being named like `isEnabled` instead of `getEnabled`

deanwhillier: I'm 0/5 on naming conventions here, just trying to follow existing approach ... though it doesn't appear that there is a single, consistent convention.

Interestingly, [Redux recommends prefexing selectors with `select`](https://redux.js.org/style-guide/style-guide#name-selector-functions-as-selectthing).

deanwhillier:
@crspeller, @hmhealey, @michelengelen, see #8870 for alternative approach that should be more performant.

deanwhillier:
Closing this in favour of #8870

PR #8869, opened 2021-09-16
MM-23881: global drafts page

koox00:
#### Summary

Channel and thread draft messages are now shown in a central drafts page
as well.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23881

#### Related Pull Requests

- Has server changes https://github.com/mattermost/mattermost-server/pull/20614

#### Screenshots

image
image

#### Release Note

```release-note
Added a centralized page for draft messages
```


matthewbirtch (changes requested):
This is really great work @koox00. You nailed the UI with great attention to detail. Thank you for that.

1. The drafts section should only show in the LHS in these two cases:
  a. If I have drafts that are unsent
  b. If I am currently on the drafts view (even if empty because I deleted all my drafts)

2. The drafts item in the LHS needs to follow the selected state of channels. It looks like it's missing the active state border on the left side. See comparison below.
Screen Shot 2021-09-17 at 11 54 21 AM
Screen Shot 2021-09-17 at 11 54 48 AM

3. In the drafts view, if I have a draft to myself, in the header can we make it say 'username (you)' like we do in the LHS?
image

4. We discussed the hover state of the pencil icon in the LHS. If you want to leave that for now, I'm fine with that, but let me know what you decide.

jgilliam17:
Thanks @koox00 I took a quick look and noticed that one of the drafts from "new" team managed to show up on the main PR team global Drafts. It's not reproducing consistently, not sure what I did to cause that it jsut happened after a bit and one page refresh. Could it be that both teams have the same channel e.g. Off Topic or similar?
Screen Shot 2021-09-17 at 5 38 03 PM

Also, we should probably add hover state to Delete and Send message modal buttons 
Screen Shot 2021-09-17 at 5 34 42 PM

koox00:
> Thanks @koox00 I took a quick look and noticed that one of the drafts from "new" team managed to show up on the main PR team global Drafts. It's not reproducing consistently, not sure what I did to cause that it jsut happened after a bit and one page refresh. Could it be that both teams have the same channel e.g. Off Topic or similar?
> Screen Shot 2021-09-17 at 5 38 03 PM
> 
> Also, we should probably add hover state to Delete and Send message modal buttons 
> Screen Shot 2021-09-17 at 5 34 42 PM
> 
> 

thanks for the review@jgilliam17, wasit the channel one or the DM?
DMs and GMs are shared across teams, and so does the drafts.
Is it was the channel one then there is a bug.

koox00:
> Also, we should probably add hover state to Delete and Send message modal buttons

as it seems the new modal (GenericModal) is lacking hover state everywhere,
some places: invite members, create new category etc..
Since the GenericModal component will need to change layout, let's add the hover state of buttons on that task.

matthewbirtch (changes requested):
@koox00 it looks like you've resolved the issues I found earlier. A few other minor polish items I found while testing more thoroughly:

- For drafts that are in DMs or GMs, instead of "In: ", can we change it to "To: " as per the design?
- For drafts that are in channels, instead of "To: ", can we change it to "In: "?
- In the design, I had a background set on the main drafts content area: `background: var(--center-channel-color-04);`
- In the header of an item, the text on the left side should be `font-weight: 600`
- The height of the header for draft items should be 40px, but it appears to be 50px. NOTE: the position of the icon buttons on hover may need to be adjusted when you make this adjustment as well
- The `DRAFT` tag looks like it needs the padding adjusted. It seems like the text is not centered properly within and it might need more padding on the left and right side
- Also, the `DRAFT` should have be `font-size: 10px`
- Can we increase the space between the `DRAFT` label and the time stamp? I think it needs another 2px. I don't know if this is borrowing from styles used elsewhere in the app though, so I don't want to create inconsistencies.
- For the actual post content, it looks like the space between the avatar and the body of the message is too wide. It is set to 24px right now, but should only be 12px;
- Also in the post content, the spacing between the username and the post body might be too much. Are we not borrowing styles and struture from the standard post component? If there is an attachment present, the spacing above the attachment card also looks too big. I assumed that we would be using the same post component here, but just displaying the container and header a bit differently.

koox00:
> Also in the post content, the spacing between the username and the post body might be too much. Are we not borrowing styles and struture from the standard post component? If there is an attachment present, the spacing above the attachment card also looks too big. I assumed that we would be using the same post component here, but just displaying the container and header a bit differently.

@matthewbirtch unfortunately we're not, since the post component has a lot of stuff going on. hover states, actions, on click events etc..

matthewbirtch:
> > Also in the post content, the spacing between the username and the post body might be too much. Are we not borrowing styles and struture from the standard post component? If there is an attachment present, the spacing above the attachment card also looks too big. I assumed that we would be using the same post component here, but just displaying the container and header a bit differently.
> 
> @matthewbirtch unfortunately we're not, since the post component has a lot of stuff going on. hover states, actions, on click events etc..

Ok, I understand. I guess that means we need to maintain two separate style definitions or can they share as much as possible still from a styling perspective?

esethna:
@koox00 this is SO AWESOME! A few things to consider or questions cc// @matthewbirtch 

1. What about existing drafts when this is launches? Will they be shown in this new drafts view or do users "start fresh" for any new drafts they create from that time forwards?
2. Are we allowing the whole card to be clickable in the draft view to open the channel or thread? If the whole card should be clickable we should change the mouse cursor to the hand pointer on hover. I imagine that is what we want since the elevation of the card increases on hover as well 
3. Is the mobile work for this started?
4. What are the next steps to implement the server changes? Do we need to discuss a spec with a server dev?

esethna (approved):

matthewbirtch:
> * Are we allowing the whole card to be clickable in the draft view to open the channel or thread? If the whole card should be clickable we should change the mouse cursor to the hand pointer on hover. I imagine that is what we want since the elevation of the card increases on hover as well

Yes, that was the intent. Thanks for the reminder @esethna. I had this note in the Figma file:
"Clicking on the entire item takes users to the location where the draft is to edit it (channel or thread RHS context)"

koox00:
> > > Also in the post content, the spacing between the username and the post body might be too much. Are we not borrowing styles and struture from the standard post component? If there is an attachment present, the spacing above the attachment card also looks too big. I assumed that we would be using the same post component here, but just displaying the container and header a bit differently.
> > 
> > 
> > @matthewbirtch unfortunately we're not, since the post component has a lot of stuff going on. hover states, actions, on click events etc..
> 
> Ok, I understand. I guess that means we need to maintain two separate style definitions or can they share as much as possible still from a styling perspective?

well I re-used some post classes, but in the end the containers are different and we may need to do some maintenance.

koox00:
> @koox00 this is SO AWESOME! A few things to consider or questions cc// @matthewbirtch
> 
>     1. What about existing drafts in local storage when this launches? Will they be shown in this new drafts view or do users "start fresh" for any new drafts they create from that time forwards?
> 
>     2. Are we allowing the whole card to be clickable in the draft view to open the channel or thread? If the whole card should be clickable we should change the mouse cursor to the hand pointer on hover. I imagine that is what we want since the elevation of the card increases on hover as well
> 
>     3. Is the mobile work for this started?
> 
>     4. What are the next steps to implement the server changes? Do we need to discuss a spec with a server dev?

@esethna 

1. I'm trying to find a solution for older drafts, so we can "transform" them to the new shape.
2. I've just added a commit to fix that!
3. Not yet, but I will start today I hope, when 1 succeeds...
4. @ashishbhate has started a server spec draft so we can discuss with the server platform team 💪

components/label/label.scss
```diff
@@ -0,0 +1,31 @@
...
+
+    &-default {
+        background-color: var(--center-channel-color);
+        color: var(--center-channel-bg);
+    }
+
+    &-primary {
+        background-color: var(--sidedar-header-bg);
+        color: var(--center-channel-color);
+    }
```
koox00, line 30: @matthewbirtch can you help verify/review these colors for the label component?
I had to create one for the drafts header, and I took the liberty to add colors for other ones as well, 
but they might be off.

PS: I'm also OK with deleting the other colors for now and adding only when necessary.

koox00: > Thanks! Can we change the following?
> 
> ```
> &-warning {
>     background-color: var(--away-indicator);
> }
> ```
> 
> ```
> &-info {
>     background-color: var(--sidebar-text-active-border);
> }
> ```
> 
> ```
> &-primary {
>     background-color: var(--button-bg);
> }
> ```

They look great thx!

matthewbirtch: @koox00 this is something we'll want to consider for the design system. I'll create a ticket for UX to review and potentially include this in the component when UI Platform builds it out

components/sidebar/sidebar_channel/channel_pencil_icon/index.ts
```diff
@@ -0,0 +1,32 @@
...
+function hasDraft(draft: any, currentChannelId?: string, channel?: Channel) {
+    return draft && Boolean(draft.message.trim() || draft.fileInfos.length || draft.uploadsInProgress.length) && currentChannelId !== channel?.id;
+}
+
+function mapStateToProps(state: GlobalState, ownProps: OwnProps) {
+    const currentChannelId = getCurrentChannelId(state);
+    const draft = ownProps.channel?.id ? getPostDraft(state, StoragePrefixes.DRAFT, ownProps.channel.id) : null;
+
+    return {
+        draft: hasDraft(draft, currentChannelId, ownProps.channel) ? draft : null,
```
hmhealey, line 28: Could we switch this back to just passing a boolean? That'll save the `ChannelPencilIcon` from re-rendering whenever the draft changes since it doesn't care about what's actually in the draft.

Also, I really like the change of where we're putting the draft icon. This seems a lot nicer than replacing the channel icon.

components/sidebar/sidebar_channel/sidebar_channel_link/__snapshots__/sidebar_channel_link.test.tsx.snap
```diff
@@ -39,6 +39,26 @@ exports[`components/sidebar/sidebar_channel/sidebar_channel_link should match sn
       channel_label
     </span>
   </div>
+  <Connect(Component)
```
hmhealey, line 42: I'm not sure if this is something we can fix easily, but shouldn't this have a component's display name in it? If it's the pencil icon, I hope we don't need to specify a display name for it manually.

selectors/drafts.ts
```diff
@@ -0,0 +1,116 @@
...
+export function makeGetDraftsByPrefix(prefix: string): DraftSelector {
+    return createSelector(
+        'makeGetDraftsByPrefix',
+        (state: GlobalState) => state.storage?.storage,
+        (storage) => {
+            if (!storage) {
+                return [];
+            }
+
+            return Object.keys(storage).flatMap((key) => {
```
hmhealey, line 50: Huh, I didn't know `.flatMap` existed. Is that actually necessary here though? I'd assume `Object.keys` would always return a flat array anyway.

selectors/drafts.ts
```diff
@@ -0,0 +1,116 @@
...
+                    return -1;
+                }
+
+                return 0;
+            }).filter((draft) => myChannels.indexOf(draft.value.channel_id) !== -1);
+        },
+    );
+}
+
+export function makeGetDraftsCount(): DraftCountSelector {
```
hmhealey, line 105: Do we need to do anything here to filter for team here? These selectors seems like they might return drafts for channels that aren't on the current team which seems like it could cause some weirdness.

types/store/rhs.ts
```diff
@@ -24,6 +24,8 @@ export type PostDraft = {
     uploadsInProgress: string[];
     props?: any;
     caretPosition?: number;
+    channel_id: string;
```
hmhealey, line 27: Could you rename that `channelId` so that we're not mixing snake case and camel case here?

hmhealey (changes requested):
This looks great! I have lots of comments/questions, but that's just because there's a lot of code. Overall, I'm excited for this!

matthewbirtch (approved):
Amazing work @koox00. Really excited for this feature. It looks like you resolved all the polish items I identified.

matthewbirtch:
@jgilliam17 I think this is in a good place for you to have a thorough review now.

actions/views/create_comment.tsx
```diff
@@ -46,12 +46,16 @@ export function clearCommentDraftUploads() {
 // we're on will not save the draft quickly enough on page unload.
 export function updateCommentDraft(rootId: string, draft?: PostDraft | null) {
     const key = `${StoragePrefixes.COMMENT_DRAFT}${rootId}`;
+    let updatedDraft = draft;
     if (draft) {
-        localStorage.setItem(key, JSON.stringify(draft));
+        const item = localStorage.getItem(key);
+        const data = item ? JSON.parse(item) : {};
+        updatedDraft = {...draft, createAt: data.createAt || new Date()};
```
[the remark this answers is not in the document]
koox00: yep that sounds better UX, @matthewbirtch what do you say?

components/drafts/channel_draft/channel_draft.tsx
```diff
@@ -0,0 +1,128 @@
...
+}: Props) {
+    const dispatch = useDispatch();
+    const history = useHistory();
+
+    const teamId = useSelector(getCurrentTeamId);
+    const channelUrl = useSelector((state) => getChannelURL(state, channel, teamId));
+
+    useEffect(() => {
+        if (!channel?.id) {
+            dispatch(getChannel(id));
```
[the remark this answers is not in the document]
koox00: Oh yes, will remove

components/drafts/channel_draft/channel_draft.tsx
```diff
@@ -0,0 +1,128 @@
...
+    }, [channel?.id]);
+
+    const onSubmit = useMemo(() => {
+        return makeOnSubmit(channel.id, '', '');
+    }, [channel.id]);
+
+    const handleSubmit = async () => {
+        await dispatch(onSubmit(value));
+        return {data: true};
+    };
```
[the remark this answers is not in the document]
koox00: I can remove `handleSubmit` and use `onSubmit` directly I guess.
the `return {data: true}` was added because ts was complaining...

components/sidebar/sidebar_channel/channel_pencil_icon/index.ts
koox00: We were going to add a tooltip upon hovering the icon to show a preview of the draft message, 
but we can do it later when we implement that change

components/sidebar/sidebar_channel/sidebar_channel_link/__snapshots__/sidebar_channel_link.test.tsx.snap
koox00: Hm, my opinion on snapshot testing is to use `react-test-renderer` as jest proposes https://jestjs.io/docs/snapshot-testing#snapshot-testing-with-jest and always mount the component.
Then we could just mock components we don't want to include in the snapshot so that they render just their children or a placeholder.

Snapshot testing via enzyme is weird on the output.

actions/views/create_comment.test.jsx
```diff
@@ -188,13 +188,19 @@ describe('rhs view actions', () => {
         const draft = {message: 'test msg', fileInfos: [{id: 1}], uploadsInProgress: [2, 3]};
 
         test('it calls setGlobalItem action correctly', () => {
+            jest.useFakeTimers('modern');
```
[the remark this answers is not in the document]
koox00: updating the comment draft might change the `create_at` on the draft, and the test failed.

components/drafts/channel_draft/channel_draft.tsx
```diff
@@ -0,0 +1,128 @@
...
+                        uploadsInProgress={value.uploadsInProgress}
+                        user={user}
+                    />
+                </>
+            )}
+        </Panel>
+    );
+}
+
+export default memo(ChannelDraft);
```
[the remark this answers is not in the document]
koox00: well I've used this here and in parent, but not in items children,
my mental model on memo is if I connect the component to redux memo it,
and don't memo the children that inherit props from parent, since those probably would also change when the parent changes.

components/drafts/panel/panel_body.tsx
```diff
@@ -0,0 +1,80 @@
...
+                    src={imageURLForUser(user.id)}
+                />
+            </div>
+            <div className='post__content'>
+                <div className='PanelBody__right'>
+                    <div className='post__header'>
+                        <strong>{displayName}</strong>
+                    </div>
+                    <div className='post__body'>
+                        <Markdown message={message}/>
```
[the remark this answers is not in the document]
koox00: I'll have a check, but it seems to render just fine, images and all.

components/drafts/thread_draft/thread_draft.tsx
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    const handleOnDelete = useCallback((id: string) => {
+        localStorage.removeItem(id);
+        dispatch(setGlobalItem(id, {message: '', fileInfos: [], uploadsInProgress: []}));
+    }, [id]);
+
+    const handleOnEdit = useCallback(() => {
+        if (channel) {
+            dispatch(selectPost({id, channel_id: channel.id} as Post));
```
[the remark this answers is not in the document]
koox00: cc @matthewbirtch ☝️

components/drafts/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import Drafts from './drafts';
+
+function makeMapStateToProps() {
+    const getDrafts = makeGetDrafts();
+    return (state: GlobalState) => {
+        const user = getCurrentUser(state);
+        const status = getStatusForUserId(state, user.id);
+
+        return {
+            displayName: displayUsername(user, getTeammateNameDisplaySetting(state)),
```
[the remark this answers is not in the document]
koox00: It's because both channel and thread drafts needed this prop to render on the header,
hm, now that I said header maybe this belongs there.
ah, it's getting used in draft_actions as well, I think, let me check.

components/generic_modal.tsx
```diff
@@ -30,7 +30,7 @@ type State = {
     show: boolean;
 }
 
-export default class GenericModal extends React.PureComponent<Props, State> {
+export default class GenericModal extends React.Component<Props, State> {
```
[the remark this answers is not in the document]
koox00: oops :D, was just trying stuff because it wouldn't open, but it was that we didn't handled prop.show changing.

selectors/drafts.ts
koox00: I recently found out it seems that you can map and filter in the same loop,
filtering is done by returning an empty array :tada:
The mobile app uses it extensively.

selectors/drafts.ts
koox00: The filtering happens here `filter((draft) => myChannels.indexOf(draft.value.channel_id) !== -1).length,`
which gets both direct and not channels in current team.

The only pain point is that `makeGetDraftsByPrefix` is not filtered just to return legacy drafts that don't include the `channel_id` so that we can fix that on mount.
I'll change the name and add a comment to denote that.

jgilliam17:
Thanks @koox00 Looks great so far, seeing few issues:
1. We should use the same limit for the number of characters as post. Additionally, I think we would want to only show few lines anyways and truncate the rest. Currently, user is allowed any number of characters in the draft and it will displays all on global drafts. @matthewbirtch thoughts?
2. Clicking on the entire item: Links are actionable, hashtags are not, also attachments can be removed while in the draft area. Not sure if this is intended or if we should disable all and direct the click on draft item to only open in channel/rhs for edit
3. Truncate long channel name in narrow screen widths - it pushes time stamp and draft label to the next row. In the screenshot the window is tiled on 15" macbook screen, also repros in mobile webview.
4. Group message - We should remove current user from the list of names and also truncate the rest if the list is long and will push time stamp and label out to the next row
Screen Shot 2021-09-24 at 4 06 35 PM

5. Mobile webview - there are 2 search icons on the header
Screen Shot 2021-09-24 at 4 13 13 PM
6. Some drafts are failing to show in the global drafts
Screen Shot 2021-09-24 at 4 08 32 PM
7. If the user (non-admin or admin) is removed from the channel by another user, the draft channel with the pencil icon will remain on their sidebar.
Screen Shot 2021-09-24 at 5 03 45 PM
8. Private channel shows public channel icon
Screen Shot 2021-09-24 at 5 09 48 PM
9. Archived channel will remain on the global drafts list if draft was present at the time of archiving.

matthewbirtch:
Great catches @jgilliam17. Thanks for the thorough testing. A few responses from me: 

> 1. We should use the same limit for the number of characters as post. Additionally, I think we would want to only show few lines anyways and truncate the rest. Currently, user is allowed any number of characters in the draft and it will displays all on global drafts. @matthewbirtch thoughts?

@jgilliam17 I opted for showing the full draft since users won't likely have a lot of drafts in their list so it's not a big concern to see the whole thing. I guess we can evaluate this as well go, but for the first iteration I designed it to show the entire post.

How did you get it to show more than the maximum characters a post is limited to?

> 2. Clicking on the entire item: Links are actionable, hashtags are not, also attachments can be removed while in the draft area. Not sure if this is intended or if we should disable all and direct the click on draft item to only open in channel/rhs for edit

I'm ok with links and @mentions, links and hashtags all being clickable. Can we fix the broken hastag links?

@koox00 let's not allow deleting of attachments from a draft in the draft list view. I think we should make all editing happen in the actual message input. 

> 3. Truncate long channel name in narrow screen widths - it pushes time stamp and draft label to the next row. In the screenshot the window is tiled on 15" macbook screen, also repros in mobile webview.

Agree. Maybe we can limit the characters similar to the limit with show in the LHS and then add a tooltip to reveal the full name when hovering. We should definitely avoid any wrapping here

> 4. Group message - We should remove current user from the list of names and also truncate the rest if the list is long and will push time stamp and label out to the next row

Agree

> 5. Mobile webview - there are 2 search icons on the header

@koox00 For mobile webview, we should follow a similar layout as threads. "Drafts" moves up to the title in the top header and the header below that should be hidden altogether.
image

koox00:
> > How did you get it to show more than the maximum characters a post is limited to?
> 
> @matthewbirtch currently there's no limit on draft character count. We only stop users for posting over the limit by displaying the error + character count.

Could we just show an indicator above/below each draft item for `current words / available words`?

matthewbirtch:
> > How did you get it to show more than the maximum characters a post is limited to?
> 
> @matthewbirtch currently there's no limit on draft character count. We only stop users for posting over the limit by displaying the error + character count.

@jgilliam17 I'm going to create a follow-up ticket for UX to include a character count to the message input so users have visibility in to how many characters they have left in the input.

components/drafts/channel_draft/channel_draft.tsx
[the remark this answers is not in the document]
hmhealey: Ah of course it was. We need to fix our custom redux types at some point because they keep making us add extra stuff like that

components/drafts/channel_draft/channel_draft.tsx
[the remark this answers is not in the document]
hmhealey: That makes sense to me. I imagine we'll want to look into usage of memo later on to come up with a "when to use memo" doc, but that's not something to worry about right now.

components/drafts/draft_actions/draft_actions.tsx
```diff
@@ -0,0 +1,155 @@
...
+                name='send'
+                tooltip={(
+                    <FormattedMessage
+                        id='drafts.actions.send'
+                        defaultMessage='Send draft'
+                    />
+                )}
+                onClick={handleSend}
+            />
+            <GenericModal
```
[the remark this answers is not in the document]
koox00: Exactly, the styling made us to choose `GenericModal` component

components/sidebar/sidebar_channel/sidebar_channel_link/__snapshots__/sidebar_channel_link.test.tsx.snap
hmhealey: At some point, we should do a big discussion on testing to figure out what we want to do for it since I'm personally not even a big fan of snapshot testing in general. I prefer either targeted assertions like "this button renders when this is true" or E2E tests since I find snapshots tend to get ignored because people just `npm test -u` them

selectors/drafts.ts
hmhealey: Oh, I see. I thought it flattened the input and then did a map so it could be used for stuff like `[[1, 2], 3, [4, 5, 6]].flatMap`, but flattening after the map makes much more sense. Using it to filter and map at the same time feels a bit counterintuitive, but it's useful enough that I can ignore that.

components/drafts/index.ts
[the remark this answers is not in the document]
hmhealey: Ah, I see where it is in the design. That looks much better than I would've expected

hmhealey (approved):
Code looks good. I'm still wanting to hear back on potentially bumping the draft any time a user edits it, but that's not something that'd block my review

components/file_preview_modal/file_preview_modal_info/file_preview_modal_info.tsx
```diff
@@ -31,6 +31,10 @@ const FilePreviewModalInfo: React.FC<Props> = (props: Props) => {
     const channel = useSelector((state: GlobalState) => selectChannel(state, props.post.channel_id));
     const name = useSelector((state: GlobalState) => displayNameGetter(state, props.post.user_id, true));
 
+    if (!channel) {
+        return <div/>;
```
larkox, line 35: Why empty div and not null?

components/sidebar/sidebar_channel/channel_pencil_icon/index.ts
```diff
@@ -0,0 +1,32 @@
...
+import {Channel} from 'mattermost-redux/types/channels';
+
+import {getPostDraft} from 'selectors/rhs';
+import {StoragePrefixes} from 'utils/constants';
+import {GlobalState} from 'types/store';
+
+import ChannelPencilIcon from './channel_pencil_icon';
+
+type OwnProps = {
+    channel?: Channel;
```
larkox, line 16: Seems like we only need the channelID. Maybe we should just pass the id to avoid unneeded rerenders?

components/sidebar/sidebar_channel/sidebar_channel_icon/sidebar_channel_icon.tsx
```diff
@@ -8,7 +8,6 @@ import {Channel} from 'mattermost-redux/types/channels';
 type Props = {
     icon: JSX.Element | null;
     channel: Channel;
```
larkox, line 10: Seems that we only need to know whether the channel is deleted. Should we change that to avoid unneeded rerenders?

selectors/drafts.ts
```diff
@@ -0,0 +1,116 @@
...
+                return [];
+            }
+
+            return Object.keys(storage).flatMap((key) => {
+                const item = storage[key];
+                if (
+                    key.startsWith(prefix) &&
+                    item !== null &&
+                    item.value != null &&
+                    (item.value.message || item.value.fileInfos.length > 0)
```
larkox, line 56: (0/5, may be more readable being explicit) We can remove the `> 0`, since length will be always >= 0 and 0 will be considered as false.

That being said, how much can we trust that fileInfos will always be an array (will never be undefined) and we will always have length? I guess we can trust, but wanted to make the question either way.

components/drafts/actions/legacy_drafts.ts
```diff
@@ -0,0 +1,85 @@
...
+import {getAllPosts} from 'mattermost-redux/selectors/entities/posts';
+import {makeGetDraftsByPrefix, DraftSelector} from 'selectors/drafts';
+import {GlobalState} from 'types/store';
+import {StoragePrefixes, StorageTypes} from 'utils/constants';
+
+const getChannelDrafts: DraftSelector = makeGetDraftsByPrefix(StoragePrefixes.DRAFT);
+const getThreadDrafts: DraftSelector = makeGetDraftsByPrefix(StoragePrefixes.COMMENT_DRAFT);
+
+// Drafts pre global drafts are missing the channelId in the value
+// thus making it hard to determine in which team they belong.
```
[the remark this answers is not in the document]
larkox: Is this needed to filter the drafts by team (not sure if we want that, since, for example, for searches, saved posts, etc... we have started to move towards showing in all teams) or to get the right link to jump to?

I know that for posts, there is a way to jump to the post without specifying the team. For channels I am not sure.

components/drafts/draft_actions/action.tsx
```diff
@@ -0,0 +1,55 @@
...
+import Constants from 'utils/constants';
+
+import './action.scss';
+
+type Props = {
+    icon: string;
+    id: string;
+    name: string;
+    onClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
+    tooltip: React.ReactNode;
```
larkox, line 18: Is there any way to create the tooltip at this level? (and have simpler props, to avoid "same object reasignation")

larkox: Could be a string "type" and handle the type here.

components/drafts/draft_title/draft_title.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+import {imageURLForUser} from 'utils/utils.jsx';
+import {Constants} from 'utils/constants';
+
+import Avatar from 'components/widgets/users/avatar';
+
+import './draft_title.scss';
+
+type Props = {
+    channel: Channel;
```
larkox, line 20: There seems that you only need the channel type, no need for the whole channel.

components/drafts/draft_title/draft_title.tsx
```diff
@@ -0,0 +1,143 @@
...
+        <>
+            <FormattedMessage
+                id='drafts.draft_title.direct_channel'
+                defaultMessage={'To: {icon} {channelName}'}
+                values={{
+                    icon,
+                    channelName,
+                }}
+            />
+            &nbsp;{you}
```
larkox, line 138: Is this space always needed? I have seen automated tests failing because of extra spaces where they are not needed, so just making sure.

components/drafts/draft_title/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        teammateId = getUserIdFromChannelName(userId, channel.name);
+        teammate = getUser(state, teammateId);
+        channelName = displayUsername(teammate, getTeammateNameDisplaySetting(state));
+    }
+
+    if (channel.type === Constants.GM_CHANNEL) {
+        const memberIds = getUserIdsInChannels(state);
+
+        membersCount = 0;
+        if (memberIds && memberIds[ownProps.channel.id]) {
```
larkox, line 40: You already have the channel in the current context (line 23), so no need to take it back from ownProps.

larkox: Same on next line.

components/drafts/draft_title/index.ts
```diff
@@ -0,0 +1,57 @@
...
+import {displayUsername} from 'mattermost-redux/utils/user_utils';
+import {Constants} from 'utils/constants';
+
+import {Channel} from 'mattermost-redux/types/channels';
+import {GlobalState} from 'types/store';
+
+import DraftTitle from './draft_title';
+
+type OwnProps = {
+    channel: Channel;
```
larkox, line 18: Probably over optimizing, but here you only seem to need the display name, name and id of the channel.

components/drafts/drafts.tsx
```diff
@@ -0,0 +1,91 @@
...
+                    id: 'drafts.heading',
+                    defaultMessage: 'Drafts',
+                })}
+                subtitle={formatMessage({
+                    id: 'drafts.subtitle',
+                    defaultMessage: 'All messages you\'ve started will show here',
+                })}
+                right={globalHeaderEnabled ? null : <RHSSearchNav/>}
+            />
+            <main className='Drafts__main'>
```
larkox, line 62: What is the advantage of using main instead of `<div>`?

My main concern is from what I read at [W3C](https://www.w3schools.com/tags/tag_main.asp):
> There must not be more than one `<main>` element in a document.

components/drafts/panel/panel.tsx
```diff
@@ -0,0 +1,45 @@
...
+    children: ({hover}: {hover: boolean}) => React.ReactNode;
+    onClick: () => void;
+};
+
+const isEligibleForClick = makeIsEligibleForClick('.file-preview.post-image__column, .hljs, code');
+
+function Panel({children, onClick}: Props) {
+    const [hover, setHover] = useState(false);
+
+    const handleMouseEnter = () => {
```
larkox, line 19: Shouldn't we use useCallback to avoid rerenders?

components/drafts/panel/panel_body.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+function Body({
+    channelId,
+    displayName,
+    draftId,
+    fileInfos,
+    message,
+    status,
+    uploadsInProgress,
+    user,
```
larkox, line 39: You only seem to need the username and id.

components/drafts/panel/panel_body.tsx
```diff
@@ -0,0 +1,80 @@
...
+                <div className='DraftPanelBody__right'>
+                    <div className='post__header'>
+                        <strong>{displayName}</strong>
+                    </div>
+                    <div className='post__body'>
+                        <Markdown message={message}/>
+                    </div>
+                    {(fileInfos.length > 0 || uploadsInProgress?.length > 0) && (
+                        <FilePreview
+                            fileInfos={fileInfos}
```
larkox, line 69: This might be a bit of an overkill, but @hmhealey what do you think about having a util function like: `arrayOrEmpty<T>(in: T[]): T[]` that if the array is empty, returns a constant object with the empty list? That way we may (or may not) reduce some re-renderings here and there.

larkox (approved):
Awesome PR. I think feel I learned lots from it.

Lots of comments here and there, but I think nothing is blocking. Many of the comments are in order to "potentially improve performance", but I am not 100% sure they actually accomplish that. Feel free to dismiss any comment you don't agree with ;)

larkox:
Another comment. What happens with Draft to channels / teams where I am no longer a member?

components/drafts/panel/panel.tsx
koox00: This is passed directly to an HTML element so It won't make a difference in re-rendering :)

components/drafts/drafts.tsx
koox00: I thought that if we aren't going to change a root div to a <main> it won't be a problem, but I get what you're saying.
I'll change it to a div :)

components/drafts/draft_actions/index.ts
```diff
@@ -0,0 +1,41 @@
...
+import {displayUsername} from 'mattermost-redux/utils/user_utils';
+import {Constants} from 'utils/constants';
+
+import {Channel} from 'mattermost-redux/types/channels';
+import {GlobalState} from 'types/store';
+
+import DraftActions from './draft_actions';
+
+type OwnProps = {
+    channel: Channel;
```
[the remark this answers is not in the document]
koox00: yep, will do!

actions/views/create_comment.tsx
```diff
@@ -46,12 +46,16 @@ export function clearCommentDraftUploads() {
...
+        localStorage.setItem(key, JSON.stringify(updatedDraft));
     } else {
         localStorage.removeItem(key);
     }
-    return setGlobalItem(key, draft);
+    return setGlobalItem(key, updatedDraft);
```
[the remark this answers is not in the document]
hmhealey: I noticed that recently too. I'm currently working on the drafts code, and I might end up cleaning that up myself once I untangle all of that, so we can leave this as-is for now. Web Platform has plans to redo all of the storage-related code in Q4 anyway.

components/file_preview_modal/file_preview_modal_info/file_preview_modal_info.tsx
koox00: To not break flexboxes alignment

components/drafts/thread_draft/thread_draft.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const handleOnEdit = useCallback(() => {
+        dispatch(selectPost({id, channel_id: channel.id} as Post));
+    }, [channel]);
+
+    const handleOnSend = useCallback(async (id: string) => {
+        await dispatch(onSubmit(value));
+
+        handleOnDelete(id);
+        handleOnEdit();
+    }, [value]);
```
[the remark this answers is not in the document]
koox00: yep you're correct, thanks!

components/drafts/actions/legacy_drafts.ts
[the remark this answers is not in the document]
koox00: Yep considering that I added `channelId`. It could work by adding team id, but the thing is that when we move the drafts from local storage to the DB we are going to store the channel id, so I thought this would make it easier.

components/drafts/draft_actions/action.tsx
koox00: hm, it seemed cleaner this way to me, I could change the variable from tooltip to tooltipText though

components/create_post/index.ts
```diff
@@ -164,12 +164,24 @@ type Actions = {
...
+                ...value,
+                channelId,
+                createAt: data.createAt || new Date(),
+            };
+            localStorage.setItem(key, JSON.stringify(updatedValue));
+        } else {
+            localStorage.removeItem(key);
+        }
+
+        return dispatch(setGlobalItem(key, updatedValue));
```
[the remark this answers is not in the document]
koox00: > Another comment. What happens with Draft to channels / teams where I am no longer a member?
> 
> Edit: And also deleted posts.

the draft get removed from the global drafts, and reappears upon joining again

components/drafts/actions/legacy_drafts.ts
```diff
@@ -0,0 +1,85 @@
...
+        const actions: GenericAction[] = [];
+
+        const legacyChannelDrafts = channelDrafts?.filter((draft) => !draft?.value?.channelId);
+        const legacyThreadDrafts = threadDrafts?.filter((draft) => !draft?.value?.channelId);
+
+        legacyChannelDrafts.forEach((draft) => {
+            const updatedValue = {...draft.value, channelId: draft.id};
+            localStorage.setItem(String(draft.key), JSON.stringify(updatedValue));
+
+            actions.push({
```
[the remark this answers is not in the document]
koox00: this should be an one off, and not be re-used, it's just to add channelId in the draft.value of old drafts.
It will be removed or re-factor again in the future when we are going to move drafts to the DB

components/drafts/panel/panel_body.tsx
larkox: Here might not be a problem, but in many index.ts we have the case of `return x || []`, and try to improve that by passing a "constant empty list". A function like that could help us to avoid extra code around those parts.

components/drafts/draft_actions/action.tsx
larkox: I might be over optimizing with this, so feel free to dismiss.

Passing the "tooltipText" is "annoying", because you would need to translate it, without using the "FormattedText" component, since that would once again create an object type, and we won't resolve anything.

My idea is that in the previous component (where this component is instantiated) the tooltip is based on the type of the action, and (IIRC) it is not dependent on anything from that component, so we can just pass a string "type", and build the whole tooltip in this component.

Again, this may dirty things too much in comparison with the potential performance gain, so feel free to dismiss.

components/drafts/actions/legacy_drafts.ts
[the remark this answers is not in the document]
larkox: @esethna Do you have any thoughts on whether the draft view will eventually be universal and not team scoped?

components/drafts/panel/panel_body.tsx
hmhealey: In those cases, I'd prefer to just pass around the null/undefined value and handle it when the variable is used. It does require null checking, but TS should ensure we do that properly. It can be nice to just pass around an empty array since it works for looping over and such, but we also do that for empty objects sometimes which causes a lot of problems with type checking and how we determine whether or not an object is empty or not.

jgilliam17:
Thanks @koox00 
Most of the listed items are fixed.
Just  few remain:
1. Bolding of the Drafts item, per spec `Drafts item is bolded if it is not the current view and there are unfinished drafts`

Screen Shot 2021-10-18 at 12 15 08 PM

2. When logged in as User-1 on the test channel, there is an item stuck in drafts (one count in drafts, but none on LHS channel list) shows error when opening Drafts on PR 8869 team (seen in the previous image too) Not sure of the steps I took to cause this. 

Screen Shot 2021-10-18 at 12 09 12 PM

3. Hash tag is not opening search for me on Drafts view
4. We can't preview or delete file attachments on the Drafts view as expected, but since they are currently not clickable they take away quite a bit of click area to open file for edit 
I'm 0/5,  we should check with @matthewbirtch if we make click on file attachment = open the post for edit

5. In mobile view, Drafts item height is less then Threads or individual channel

Screen Shot 2021-10-18 at 12 55 09 PM

6. Emoji is lost, if it's the last item added (case where there is no text following the emoji)
Screen Shot 2021-10-18 at 12 31 25 PM
Screen Shot 2021-10-18 at 12 31 18 PM

esethna:
@jgilliam17 @koox00 can we clarify what's the expected behaviour for existing drafts after a server gets updated to support this? I think we discussed that the user should "start fresh" so existing drafts aren't added to this new view, just new drafts that are created after this feature is available to the user.

koox00:
@esethna fixed the responsive issue, this should be ready to merge!! 🎉

matthewbirtch:
> 1. Should we make sure the drafts item in the sidebar doesn't appear or get updated until the user actually leaves the channel or thread? Right now it updates almost immediately when you start typing somthing, which means that a users sidebar will be bouncing around as they are composing and sending messages. I wouldn't really count it as a "draft" until the user leaves the channel or thread (this is Slacks approach as well)

@esethna @koox00 I agree. If we can make it so it's not considered a draft until I leave the currently focused message input - or as soon as I leave the current channel or thread.

koox00:
> @esethna @koox00 I agree. If we can make it so it's not considered a draft until I leave the currently focused message input - or as soon as I leave the current channel or thread.

Changing that would require changing the whole posting message/comment logic unfortunately.
Do we want to do this in this PR?

matthewbirtch:
> > @esethna @koox00 I agree. If we can make it so it's not considered a draft until I leave the currently focused message input - or as soon as I leave the current channel or thread.
> 
> @matthewbirtch, @esethna changing that would require changing the whole posting message/comment logic unfortunately. Do we want to do this in this PR?

Ok, if it's that significant, then I'd be fine with leaving it this way for now and creating a follow-up to address that

esethna (approved):
> I still think it's important enough to find a solution for because it's a bit distracting while you're writing a post

@koox00 @matthewbirtch I agree we need to fix that. I think we need to find a solution to this before we ship the feature actually, but its probably fine if it comes in a separate PR. @koox00 can you help open a ticket and take it on next OSF?

esethna:
@jgilliam17 for QA :)

koox00:
> > I still think it's important enough to find a solution for because it's a bit distracting while you're writing a post
> 
> @koox00 @matthewbirtch I agree we need to fix that. I think we need to find a solution to this before we ship the feature actually, but its probably fine if it comes in a separate PR. @koox00 can you help open a ticket and take it on next OSF?

It could be a simpler solution, I don't know if we want it,
we could display the drafts link event when there are no drafts

@matthewbirtch, @esethna

matthewbirtch:
> It could be a simpler solution, I don't know if we want it, we could display the drafts link even when there are no drafts
> 
> @matthewbirtch, @esethna

The initial spec was not to show the Drafts item in the LHS until there are actually drafts and I think we want to keep it that way. The goal with that was to avoid adding clutter and avoid taking up vertical space in the LHS when its not needed

esethna:
> The initial spec was not to show the Drafts item in the LHS until there are actually drafts and I think we want to keep it that way. The goal with that was to avoid adding clutter and avoid taking up vertical space in the LHS when its not needed

+1 ^

koox00:
@esethna, @matthewbirtch, would saving drafts upon leaving the input text work for you?

matthewbirtch:
> @esethna, @matthewbirtch, would saving drafts upon leaving the input text work for you?

I'm not sure. Does that mean if a user writes a draft and then leaves the app or switches to a different tab in their browser that their draft is not saved until they leave the focus of the input? I think a draft needs to be saved with every keystroke change, personally

I think I need to understand more of what you're proposing and to think through the implications. Could you elaborate a bit on what you mean by "saving drafts upon leaving the input text"?

koox00:
@matthewbirtch so to clarify, you want to save drafts on every keystroke, but show the draft in the drafts link when changing channels?

matthewbirtch:
> @matthewbirtch so to clarify, you want to save drafts on every keystroke, but show the draft in the drafts link when changing channels?

Yes, if possible. The saving of drafts on every keystroke is what I assume it is already doing, so I am not trying to propose something new with that. 

But if I've entered text in to the message input field, as soon as I leave that input field (by either changing the channel or opening a thread or some other pane that causes me to lose focus on the message input field) the drafts item should show in the LHS.

hmhealey:
To clarify, we're currently saving drafts when the user stops typing as far as I know. That code has gone through a few iterations lately so that might not be exactly what's happening, but it's roughly saving them as the user types

koox00:
> Should we make sure the drafts item in the sidebar doesn't appear or get updated until the user actually leaves the channel or thread? Right now it updates almost immediately when you start typing somthing, which means that a users sidebar will be bouncing around as they are composing and sending messages. I wouldn't really count it as a "draft" until the user leaves the channel or thread (this is Slacks approach as well)

@matthewbirtch I've attempted a fix on that, can you please review?

components/create_comment/index.ts
```diff
@@ -98,8 +98,12 @@ function makeMapStateToProps() {
     };
 }
 
-function makeOnUpdateCommentDraft(rootId: string) {
-    return (draft?: PostDraft) => updateCommentDraft(rootId, draft);
+function makeOnUpdateCommentDraft(rootId: string, channelId: string) {
+    return (draft?: PostDraft) => updateCommentDraft(rootId, draft ? {...draft, channelId} : draft);
```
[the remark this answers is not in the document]
koox00: @larkox this is used in the callback of either upload success or failure,
 the reasoning is that you might have left that thread, and opened another one since the upload ends.
so the `rootid` should be the one passed in the callback and not the one in the current props.

actions/views/create_comment.tsx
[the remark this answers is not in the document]
koox00: @hmhealey the order of drafts is now based on draft updated at

components/create_post/index.ts
```diff
@@ -169,12 +169,24 @@ type Actions = {
...
+    return (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        let updatedValue;
+        if (value) {
+            const item = localStorage.getItem(key);
+            const data = item ? JSON.parse(item) : {};
+            const channelId = getCurrentChannelId(getState());
+            updatedValue = {
+                ...value,
+                channel_id: channelId,
+                createAt: data.createAt || new Date(),
```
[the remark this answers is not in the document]
koox00: same as above

actions/views/create_comment.tsx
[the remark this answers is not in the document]
koox00: AFAICT setting it to `null` had no actual effect. It seems that 
`redux-persist` will ignore it and wouldn't update the item in the storage.
So I had to set it to an empty string to remove the draft.

It works right now I think because we are always setting the message to something,
either an empty string or an actual message.

That said, I would like to keep drastic changes in this PR at minimun, since it already has a ton of additions.

components/drafts/actions/legacy_drafts.ts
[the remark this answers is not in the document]
koox00: @larkox this file was deleted so I'm resolving this conversation

matthewbirtch (changes requested):
Awesome work @koox00. This is definitely what I had in mind and it's feeling a lot better this way.

I took a quick look at mobile webview, and it looks like there is a problem with the confirmation modals. However, this looks like it might be a problem out of the scope of this PR as it seems to be an issue on the "Create Category" modal as well. Can you confirm that it's not specific to this PR?

image

image

Other than that, I think we're good to go!

koox00:
> Awesome work @koox00. This is definitely what I had in mind and it's feeling a lot better this way.
> 
> I took a quick look at mobile webview, and it looks like there is a problem with the confirmation modals. However, this looks like it might be a problem out of the scope of this PR as it seems to be an issue on the "Create Category" modal as well. Can you confirm that it's not specific to this PR?
> 
> Other than that, I think we're good to go!

Looks like daily suffers from the same problem, so not an issue with this PR. 
A CSS rule that positions the modal at `50vh - 240px` is not being reset for mobile.

We should fix that on another PR

koox00:
@larkox, @hmhealey can you re-review? I have made some changes due to the UX requirement below.

> Should we make sure the drafts item in the sidebar doesn't appear or get updated until the user actually leaves the channel or thread? Right now it updates almost immediately when you start typing somthing, which means that a users sidebar will be bouncing around as they are composing and sending messages. I wouldn't really count it as a "draft" until the user leaves the channel or thread (this is Slacks approach as well)

The gist of the changes are described in this commit's message: https://github.com/mattermost/mattermost-webapp/pull/8869/commits/93a7d563621430bf0e8e143b2e0af34ea1af54a4

matthewbirtch (approved):
Fantastic work @koox00!

esethna (approved):
Amazing! Can't wait to hook this up to the server to sync the drafts ;)

One very minor thing, I think the number next to drafts needs an opacity reduction?

Test server:
image


Design:
image

components/drafts/draft_title/draft_title.tsx
larkox: My previous comment here is still valid. In this component you don't need the whole channel, just the type of channel.

components/drafts/draft_title/draft_title.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    id='drafts.draft_title.you'
+                    defaultMessage={'(you)'}
+                />
+            </>
+        );
+    }
+
+    let icon = <i className='icon icon-globe'/>;
+
+    if (channel.type === Constants.PRIVATE_CHANNEL) {
```
larkox, line 61: Why not use a switch?

larkox (approved):
LGTM, just small nitpicks.

actions/views/drafts.ts
```diff
@@ -0,0 +1,58 @@
...
+import {PostDraft} from 'types/store/rhs';
+
+export function removeDraft(key: string) {
+    return (dispatch: DispatchFunc) => {
+        localStorage.removeItem(key);
+        dispatch(setGlobalItem(key, {message: '', fileInfos: [], uploadsInProgress: []}));
+    };
+}
+
+export function updateDraft(key: string, value: PostDraft|null) {
```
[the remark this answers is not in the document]
koox00: It might not be visible but this returns an action (`setGlobalItem()`), so it has to be dispatched, `dispatch(updateDraft())`.

e.g:

```js
const action = () => setGlobalItem(value)
// the above is the same as doing 
const action = () => dispatch => dispatch(setGlobalItem(value))

File: components/drafts/channel_draft/channel_draft.tsx [hunk 40]
```diff
@@ -0,0 +1,112 @@
...
+
+    const handleOnEdit = useCallback(() => {
+        history.push(channelUrl);
+    }, [channelUrl]);
+
+    const handleOnSend = useCallback(async (id: string) => {
+        await dispatch(onSubmit(value));
+        handleOnDelete(id);
+        handleOnEdit();
+    }, [value, onSubmit]);
```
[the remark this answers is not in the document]
koox00: Yes that was my thinking, I should be explicit though, I'll update :)

koox00:
> My previous comment here is still valid. In this component you don't need the whole channel, just the type of channel.

`draft_title/index.tsx` needs also `id`, `name`, display_name` and `type` so parent passes `channel` as prop to `draft_title`.
I could use `mergeProps` to omit `channel` and just pass `channelType` but it seems like over-optimization to me.

components/file_preview_modal/file_preview_modal_info/file_preview_modal_info.tsx
koox00: Resolving this comment, since the fix that would crash the file preview (empty `channel/post`) is merged in upstream and these changes are being reverted

selectors/drafts.ts
```diff
@@ -0,0 +1,119 @@
...
+    return createSelector(
+        'makeGetDrafts',
+        getChannelDrafts,
+        getRHSDrafts,
+        getMyActiveChannels,
+        (channelDrafts, rhsDrafts, myChannels) => (
+            [...channelDrafts, ...rhsDrafts]
+        ).
+            filter((draft) => myChannels.indexOf(draft.value.channelId) !== -1).
+            sort((a, b) => new Date(b.value.updateAt).getTime() - new Date(a.value.updateAt).getTime()),
```
hmhealey, line 104: Why wrap these in `new Date` when sorting? Couldn't we just have one `updateAt` minus the other?

types/store/rhs.ts
```diff
@@ -24,6 +24,10 @@ export type PostDraft = {
     uploadsInProgress: string[];
     props?: any;
     caretPosition?: number;
+    channelId: string;
+    createAt: Date;
+    updateAt: Date;
```
hmhealey, line 29: I don't think we can use Date objects here since they're written to storage/the indexed DB which I think requires encoding them as JSON which breaks the date. If that's the case, we could probably use numbers and store the unix timestamps like we do on the server

hmhealey (changes requested):

components/sidebar/sidebar_channel/sidebar_channel_link/sidebar_channel_link.tsx
```diff
@@ -235,7 +236,7 @@ export default class SidebarChannelLink extends React.PureComponent<Props, State
         const content = (
             <>
                 <SidebarChannelIcon
-                    channel={channel}
+                    isDeleted={channel.delete_at !== 0}
```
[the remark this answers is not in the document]
koox00: It was asked in a review comment

components/drafts/thread_draft/thread_draft.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+        return () => Promise.resolve({data: true});
+    }, [channel.id, thread?.id]);
+
+    const handleOnDelete = useCallback((id: string) => {
+        dispatch(removeDraft(id));
+    }, [id]);
+
+    const handleOnEdit = useCallback(() => {
+        dispatch(selectPost({id, channel_id: channel.id} as Post));
```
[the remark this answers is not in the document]
koox00: Well it's a bit confusing, the id of the draft is the id of the root post, so that's selecting the root post to open the RHS

selectors/drafts.ts
koox00: well that was only because I was saving date objects in storage, so the returned was a string representation of the object, so this was needed. Will change to use numbers in storage as you recommended below!

types/store/rhs.ts
koox00: Yep, at first I used whatever was used for the timestamp which was date object, but yep it should be a number, will change, thanks!

components/drafts/thread_draft/thread_draft.tsx
```diff
@@ -0,0 +1,127 @@
...
+    status,
+    thread,
+    type,
+    user,
+    value,
+}: Props) {
+    const dispatch = useDispatch();
+
+    useEffect(() => {
+        if (!thread?.id) {
```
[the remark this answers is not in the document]
koox00: you can land in drafts without having the root post of a thread with a draft loaded.

components/drafts/thread_draft/thread_draft.tsx
```diff
@@ -0,0 +1,127 @@
...
+            dispatch(getPost(id));
```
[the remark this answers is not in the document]
koox00: yep you are right, `getThread` requires team id, while `getPost`'s signature is quite simple, requires only the post id.
that's why I'm just getting the post, and then creating a synthetic thread, since we don't care about the thread per se, but just if the post exists or not.

components/drafts/drafts.tsx
```diff
@@ -0,0 +1,87 @@
...
+    useEffect(() => {
+        dispatch(selectChannel(''));
+    }, []);
+
+    return (
+        <div
+            id='app-content'
+            className='Drafts app__content'
+        >
+            <Header
```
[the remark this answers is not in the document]
koox00: it's a <header> with a title (h#) and a subtitle, seemed more comfortable!

components/drafts/draft_title/index.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (memberIds && memberIds[channel.id]) {
+            const groupMemberIds: Set<string> = memberIds[channel.id] as unknown as Set<string>;
+            membersCount = groupMemberIds.size;
+            if (groupMemberIds.has(userId)) {
+                membersCount--;
+            }
+        }
+
+        if (membersCount === 0) {
+            membersCount = channelName.split(',').length;
```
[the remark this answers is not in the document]
koox00: just a fallback, 2 lines above that one is the same calculation as GM channels in the sidebar

components/drafts/draft_actions/delete_draft_modal.tsx
```diff
@@ -0,0 +1,57 @@
...
+                displayName,
+            }}
+        />
+    );
+
+    return (
+        <GenericModal
+            confirmButtonClassName='delete'
+            confirmButtonText={confirmButtonText}
+            handleCancel={() => {}}
```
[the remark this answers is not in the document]
koox00: I think if you don't pass a `handleCancel` it won't render a `cancel` button, should I change `GenericModal` to support a `showCancelBtn` prop?

components/drafts/draft_actions/action.tsx
```diff
@@ -0,0 +1,55 @@
...
+type Props = {
+    icon: string;
+    id: string;
+    name: string;
+    onClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
+    tooltipText: React.ReactNode;
+};
+
+function Action({name, icon, onClick, id, tooltipText}: Props) {
+    const buttonRef = useRef<HTMLButtonElement>(null);
```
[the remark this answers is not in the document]
koox00: nope! thanks!

components/drafts/draft_actions/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import {displayUsername} from 'mattermost-redux/utils/user_utils';
+import {Constants} from 'utils/constants';
+
+import {Channel} from 'mattermost-redux/types/channels';
+import {GlobalState} from 'types/store';
+
+import DraftActions from './draft_actions';
+
+type OwnProps = {
+    channelDisplayName: Channel['display_name'];
```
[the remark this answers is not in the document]
koox00: I think it was asked in a review comment.

components/drafts/draft_title/index.ts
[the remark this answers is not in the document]
hmhealey: Could we pull this all out into a selector then so that we're not duplicating this fairly complicated logic?

components/drafts/thread_draft/thread_draft.tsx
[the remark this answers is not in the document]
hmhealey: That seems like it's going to be very confusing in the future. Could we rename that prop to `rootId` then?

types/store/rhs.ts
```diff
@@ -23,6 +23,10 @@ export type PostDraft = {
     uploadsInProgress: string[];
     props?: any;
     caretPosition?: number;
+    channelId: string;
+    createAt: number;
+    updateAt: number;
+    show?: boolean;
```
hmhealey, line 29: What's this `show` for again? If it's to differentiate empty drafts from ones that we want to see in the draft view, we could make a helper to check if any of `message`, `fileInfos`, and `uploadsInProgress` are set, and if they are, then we show the draft rather than saving a separate field for it

hmhealey (dismissed):
There's a few things from my previous review that I think were missed, but I've also added another couple questions here

types/store/rhs.ts
koox00: It's for a UX requirement. A draft will be displayed in global draft when you focused out of the input.

hmhealey: Ah, I see. Could we perhaps have that derived by having the Global Drafts menu link be aware of the current channel ID and selected post ID? I'd be worried that storing it will somehow cause the stored value to become out of sync with the rest of the UI or cause some weirdness when you have multiple tabs open viewing different channels

koox00: hm I don't know how it would work.
Let's say you have no drafts and visiting channel A and started writing a message,  
the link "knows" you're on channel A and that the draft belongs there,
so it will be hidden.
Now you visit channel B and the link shows itself, 
but what if you go back to channel A having only that one draft?

hmhealey: That's a good point. I can't think of a way to derive that unless we were to do something with not putting the draft in redux until you blur the textbox, but I'd be worried about that causing the draft to sometimes not save at all.

The only other concern that I'd have with this then is if we're persisting the value of `show` to storage. Is it possible for the user to close the tab without triggering the blur so that we end up with a draft saved with a show value of false? I suppose we could use the `REHYDRATE` action to ensure its always set to `true` when rehydrating if it does become an issue though.

Do you think that's a possiblity? If not, I'd be good leaving this as-is

hmhealey: ^ Bump on this since I realize GitHub doesn't make it immediately obvious when someone responds to a thread without flagging it as a whole review

koox00: > Is it possible for the user to close the tab without triggering the blur so that we end up with a draft saved with a show value of false?

@hmhealey I've tried closing, and refreshing the tab via shortcut blur is triggered, can't think of any other way the user would not blur the input and close.

jgilliam17:
Removing myself as reviewer - @koox00 please add me again when ready for QA - thanks 🙂

rmonga71:
@koox00 let me know when this ready, happy to help keep the reviews moving. I know there's some things to discuss still, but really excited to get this out!

mylonsuren:
Both this and https://github.com/mattermost/mattermost-server/pull/20614 are ready for review, most of the code here is unchanged, especially with regards to the  websocket functionality, selectors, and actions. The main additions are minor bug fixes and handling file attachments in `advanced_create_post` and `advanced_create_comment`, the addition of a feature flag, system setting, and user setting, and all the associated tests.

cc: @esethna @koox00

components/drafts/panel/panel.scss
```diff
@@ -0,0 +1,18 @@
+.Panel {
+    border: 1px solid rgba(var(--center-channel-color-rgb), 0.08);
+    background-color: var(--center-channel-bg);
+    border-radius: 4px;
+    box-shadow: 0 2px 3px rgba(0, 0, 0, 0.08);
```
AshishDhama, line 5: We have `--elevation-1` variables let use that

components/drafts/panel/panel_header.scss
```diff
@@ -0,0 +1,74 @@
...
+    &__sync-icon {
+        display: flex;
+        align-items: center;
+    }
+
+    &__actions {
+        display: none;
+
+        &.show {
+            display: inline-flex !important;
```
AshishDhama, line 62: Assuming we can’t avoid important here.

components/drafts/panel/panel_body.tsx
```diff
@@ -0,0 +1,90 @@
...
+    userId: UserProfile['id'];
+    username: UserProfile['username'];
+}
+
+const OPTIONS = {
+    disableGroupHighlight: true,
+    mentionHighlight: false,
+};
+
+function Body({
```
AshishDhama, line 35: nit: Function name `PanelBody` , we are using `PanelHeader` so may be change here as well

components/drafts/panel/panel.scss
```diff
@@ -0,0 +1,18 @@
...
+    cursor: pointer;
+    transition: box-shadow 0.3s ease-in-out;
+
+    & + .Panel {
+        margin-top: 16px;
+    }
+
+    &:hover {
+        border-color: rgba(var(--center-channel-color-rgb), 0.16);
+        box-shadow: 0 4px 6px rgba(0, 0, 0, 0.12);
```
AshishDhama, line 15: We have `--elevation-2` variables let use that

components/drafts/drafts_link.tsx
```diff
@@ -0,0 +1,74 @@
...
+                <NavLink
+                    to={`${url}/drafts`}
+                    id='sidebarItem_drafts'
+                    activeClassName='active'
+                    draggable='false'
+                    className='SidebarLink sidebar-item'
+                    role='listitem'
+                    tabIndex={0}
+                >
+                    <i
```
AshishDhama, line 56: compass icon

components/drafts/drafts_link.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        data-testid='draftIcon'
+                        className='icon icon-pencil-outline'
+                    />
+                    <div className='SidebarChannelLinkLabel_wrapper'>
+                        <span className='SidebarChannelLinkLabel sidebar-item__name'>
+                            {formatMessage({id: 'drafts.sidebarLink', defaultMessage: 'Drafts'})}
```
AshishDhama, line 62: `FormattedMessage` could be used instead, as it will return a span. It's preferred over `formatmessage`.

AshishDhama (changes requested):

AshishDhama:
@mylonsuren I am not finished yet, but posting what ever I can do today. Thanks

components/threading/global_threads/thread_list/virtualized_thread_list.tsx
```diff
@@ -65,6 +63,8 @@ function VirtualizedThreadList({
         [ids, selectedThreadId, isLoading, addNoMoreResultsItem, total],
     );
 
+    const itemKey = useCallback((index, data) => data.ids[index], []);
+
```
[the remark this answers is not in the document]
mylonsuren: @koox00 Can you comment on this? I'm not too familiar with the virtualized list.

mylonsuren:
@AshishDhama I've pushed changes that address most of your comments. Regarding the use of compass-icons, is there a way to avoid having to manually style each icon where it's used? I played around a bit and it using the compass icons gets rid of all styling and I'm not too keen on going through each one to style them appropriately. 

It's not a blocker for this PR but I'd like to figure this out if possible afterwards when we have some time.

Linking the JIRA ticket here for visibility: https://mattermost.atlassian.net/browse/MM-48568

components/drafts/panel/panel.scss
mylonsuren: Done

components/drafts/panel/panel_body.tsx
mylonsuren: Done

components/drafts/panel/panel.scss
mylonsuren: Done

components/sidebar/sidebar_channel/channel_pencil_icon/channel_pencil_icon.tsx
```diff
@@ -0,0 +1,24 @@
...
+import './channel_pencil_icon.scss';
+
+type Props = {
+    hasDraft: boolean;
+};
+
+function ChannelPencilIcon({hasDraft}: Props) {
+    if (hasDraft) {
+        return (
+            <i
```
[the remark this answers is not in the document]
mylonsuren: See https://github.com/mattermost/mattermost-webapp/pull/8869#issuecomment-1324423919, I can update the components to use compass-icons in a later PR if needed.

components/sidebar/sidebar_channel/sidebar_channel_icon/sidebar_channel_icon.tsx
```diff
@@ -3,29 +3,18 @@
...
-                    className='icon icon-pencil-outline'
-                />
-            );
-        }
-
-        return this.props.icon;
+function SidebarChannelIcon({isDeleted, icon}: Props) {
+    if (isDeleted) {
+        return (
+            <i className='icon icon-archive-outline'/>
```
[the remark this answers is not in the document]
mylonsuren: See https://github.com/mattermost/mattermost-webapp/pull/8869#issuecomment-1324423919, I can update the components to use compass-icons in a later PR if needed.

components/sidebar/sidebar_channel/channel_pencil_icon/channel_pencil_icon.scss
```diff
@@ -0,0 +1,15 @@
+#SidebarContainer .SidebarChannel .SidebarLink {
+    .channel-pencil-icon {
```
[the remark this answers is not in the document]
mylonsuren: See https://github.com/mattermost/mattermost-webapp/pull/8869#issuecomment-1324423919, I can update the components to use compass-icons in a later PR if needed.

components/file_preview/file_preview.tsx
```diff
@@ -105,12 +105,14 @@ export default class FilePreview extends React.PureComponent<Props> {
...
-                                onClick={this.handleRemove.bind(this, info.id)}
-                            >
-                                <i className='icon icon-close'/>
-                            </a>
+                            {Boolean(this.props.onRemove) && (
+                                <a
+                                    className='file-preview__remove'
+                                    onClick={this.handleRemove.bind(this, info.id)}
+                                >
+                                    <i className='icon icon-close'/>
```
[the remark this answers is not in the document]
mylonsuren: See https://github.com/mattermost/mattermost-webapp/pull/8869#issuecomment-1324423919, I can update the components to use compass-icons in a later PR if needed.

components/drafts/drafts_link.tsx
mylonsuren: See https://github.com/mattermost/mattermost-webapp/pull/8869#issuecomment-1324423919, I can update the components to use compass-icons in a later PR if needed.

components/drafts/panel/panel_header.scss
mylonsuren: Yeah, not sure there's another way to get the styling we need here.

components/drafts/draft_actions/action.scss
```diff
@@ -0,0 +1,52 @@
+.DraftAction {
+    &__button {
+        display: inline-flex;
+        width: 28px;
+        height: 28px;
+        align-items: center;
```
AshishDhama, line 6:
```
       align-items: center;
        justify-content: center;
```
FYI: We could use 'place-items: center'

components/drafts/draft_actions/delete_draft_modal.tsx
```diff
@@ -0,0 +1,57 @@
...
+            handleConfirm={onConfirm}
+            modalHeaderText={title}
+            onExited={onExited}
+        >
+            <div>{message}</div>
```
AshishDhama, line 52: Do we need block level element? I think we can get rid of div. or maybe 

```typescript
<FormattedMessage
            id={'drafts.confirm.delete.text'}
             tagName={'div'}
```
This may apply to some other places as well.

AshishDhama (approved):
Great work @koox00 @mylonsuren My comments are non-blocking.

AshishDhama:
> @AshishDhama I've pushed changes that address most of your comments. Regarding the use of compass-icons, is there a way to avoid having to manually style each icon where it's used? I played around a bit and it using the compass icons gets rid of all styling and I'm not too keen on going through each one to style them appropriately.
> 
> It's not a blocker for this PR but I'd like to figure this out if possible afterwards when we have some time.
> 
> Linking the JIRA ticket here for visibility: https://mattermost.atlassian.net/browse/MM-48568

Yeah, I think we can only set size and colour without entering into the CSS world, But I think that should be sufficient. We can discuss in mm

components/drafts/draft_actions/delete_draft_modal.tsx
mylonsuren: Yeah just tested it and doesn't look like we need it :+1:

components/drafts/draft_actions/action.scss
mylonsuren: `place-items: center` doesn't centre the items when I test it locally

AshishDhama: sorry, it's `place-content` I am always confused with this.

esethna:
@mylonsuren just checking, are dev reviews complete here, just blocked on QA's comments from the other PR? If so can we remove the dev review label?

mylonsuren:
> @mylonsuren just checking, are dev reviews complete here, just blocked on QA's comments from the other PR? If so can we remove the dev review label?

I pinged @hmhealey earlier to see if he wants to re-review this based on his comments on the PR from earlier this year. Just waiting on a response and once that's resolved dev reviews would be complete.

esethna:
@mylonsuren @koox00 

Chatted with Harrison, let's please turn the feature flag **off by default** so we can follow the regular process for rolling out to community first, then cloud, then on-prem if there's no issues

cc// @amyblais

jgilliam17 (approved):
QA testing was done on the related server PR.
See comment on the server PR for additional info.

PR #8870, opened 2021-09-16
Refactor isMobile to use redux and change name to isMobileView for better clarity (alternative)

deanwhillier:
#### Summary
This PR adds window size range tracking to Redux with several selectors for determining if the current window size is mobile, tablet, small desktop or desktop corresponding to our current primary CSS media query sizes.

In addition, GlobalHeader and Sidebar components have been updated to use the new selectors.

The existing isMobile name has been changed to isMobileView to help with clarity. isMobileView speaks more to describing the MM defined view size(s) as opposed to an arbitrary device size.

NOTE: This is an alternative approach to #8862 that removes precise window pixel tracking to improve redux performance.  The need to know the precise window size is very limited from what I can tell, so we can probably leave that up to the individual components as needed for now. Components using the selectors now only get re-rendered if the window range size changes (i.e. from mobile to tablet) and not with every window resize event update.

#### Related Pull Requests
Alternative to https://github.com/mattermost/mattermost-webapp/pull/8862

#### Release Note
```release-note
NONE
```


deanwhillier:
NOTE: I have purposely left a few `console.log()` statements to be able to see functional and class components re-rendering as the window size is changed. I will remove before this is finalized. A far as I can tell, both component types only re-render when the value of windowSize in redux changes and not with every dispatched action triggered by window resize events.

components/global/global_header.tsx
```diff
@@ -95,7 +96,17 @@ const GlobalHeader = (): JSX.Element | null => {
...
+    console.log('[GLOBAL HEADER] isMobileView', isMobileView);
+
+    useEffect(() => {
+        console.log('[GLOBAL HEADER] general useEffect: isMobileView', isMobileView);
+    });
+
+    useEffect(() => {
+        console.log('[GLOBAL HEADER] limited isMobileView', isMobileView);
+    }, [isMobileView]);
+
```
michelengelen, line 108: testing? :P

actions/views/browser.ts
```diff
@@ -0,0 +1,35 @@
...
+    case width <= Constants.MOBILE_SCREEN_WIDTH: {
+        windowSize = WindowSizes.MOBILE_VIEW;
+        break;
+    }
+    }
+
+    return {
+        type: ActionTypes.BROWSER_WINDOW_RESIZED,
+        data: windowSize || WindowSizes.DESKTOP_VIEW,
+    };
```
michelengelen, line 34: instead of returning a value and alternative you could just use the default case for that, so that the variable is always "filled"

michelengelen (approved):
2 little comments, but overall I like this approach.

actions/views/browser.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {GenericAction} from 'mattermost-redux/types/actions';
+
+import {Constants, ActionTypes, WindowSizes} from 'utils/constants';
+
+export function emitBrowserWindowResized(): GenericAction {
```
crspeller, line 8: Maybe I am misunderstanding, but I don't think this is much more performant then the previous approach since we still send redux updates for every resize. If there is a memoizing selector on the other end, it helps there to prevent a re-render so we get that. But we still have to go though all the redux stuff every time. It's certainly better than the previous approach though. 

I guess I still don't understand why we need this in redux. It feels like CSS and a custom hook that watches for window resize events should be able to solve any problem.

deanwhillier: Unless I'm missing something, this is only triggering re-renders when the window width crosses one of the 3 window size boundaries and the value stored in redux actually changes andnot every time the window resize event dispatches the action. Both the functional Global Header component and the class-based Sidebar component only get re-rendered when the value stored in Redux changes and not every time the action is dispatch. That's what the console.logs I left in the code are there to demonstrate ... give it a try.

I'm actually not doing this specifically for Global Header but as a replacement for Utils.isMobile(), which is often getting used as if it was state that triggers component re-renders when the window resizes, even though it obviously doesn't. 

I'm on the fence as to whether the JS or CSS approach is better for Global Header. CSS Media Queries are meant for this kind of situation, but not having unnecessary markup if it can be accomplished performantly is also appealing.

components/global/global_header.tsx
deanwhillier: To demonstrate the effect of dispatching the windowSize action from the window resize event on the rendering of components that subscribe to the selectors. I'm going to remove if we decide to move forward with this.

actions/views/browser.ts
deanwhillier: Regardless of this exploration, I'm suggesting we [continue with the CSS approach](https://github.com/mattermost/mattermost-webapp/pull/8740#issuecomment-921919067) for GH.

hmhealey: I think Dean's right. This'll trigger the reducer to run, but since the reducer returns a string, `combineReducers` won't generate a new state object, so Redux won't trigger an update.

actions/views/browser.ts
```diff
@@ -0,0 +1,35 @@
...
+    const width = window.innerWidth;
+
+    let windowSize;
+    switch (true) {
+    case width > Constants.DESKTOP_SCREEN_WIDTH: {
```
hmhealey, line 13: Total nitpicking, but I'm still not a fan of using `switch(true)` instead of an if-else statement. I just don't think it adds enough to justify introducing something new that other devs might have to learn

actions/views/browser.ts
hmhealey: As for using a hook, I could see that also working as long as it's only React that cares about window size and that it's newer components that rely on it as well. We have enough old random code that needs it that I don't think we can abandon that entirely in favour of hooks yet. This could even be a stepping stone on the way there though

hmhealey (commented):
Thinking more about it, I do think this ends up being better than the old one where we were updating constantly during resize. Theoretically that shouldn't have been a huge concern, but this will probably result in noticeably smoother resizing.

actions/views/browser.ts
deanwhillier: IMO it's a totally valid and arguably better way of writing conditional logic in situations with multiple potential options and should be something JS developers should know about. No time like the present to learn!

What is interesting to me is the varying opinions around the different ways of writing conditional logic in JS at MM. I would bet that if you put 5 developers in a room and asked them to present their preferred approach, you would get 5 different approaches. :P Every time I go to write something with conditional logic, I find my self questioning what approach to use that might minimize change requests and pretty much each time someone requests a change. 🤷‍♂️

In the absence of a definitive style guide on how we prefer to use conditional logic at MM or at least linting that catches for situations we would prefer not to use, personal preference is going to win at the end of the day so long is it is technically correct. This is one of mine personal preferences. :)

actions/views/browser.ts
deanwhillier: TL;DR, I'll go with switch's default using fall through.

It's a little frustrating when it comes to conditional logic in JS. Everyone seems to have a different preference/opinion for how to do it "right" :P

In this case you could:
1. Just set the default when defining the variable and be done with it.
2. Use switch's default with fall through
3. Use switch's default without fall through
4. Set the default value at the end if none has been set already

I actually prefer to use switch's default option with fall through in a situation like this, but I've seen many comments where people prefer not to take that approach because of the fall through ... even though fall through is a very viable approach and can be super helpful in the right situation. You could omit the case, and thereby remove fall through altogether given it's the default case, but I prefer to explicitly identify all the cases in something like this so there is no question later on down the road as to what the original intentions were. It's much easier to see what the case is than to have to try to infer it by understanding what all the other cases do first.

components/global/global_header.tsx
deanwhillier: Removing now that everyone has had a chance to look.

actions/views/browser.ts
deanwhillier: I'm going to leave the hook approach to later as a majority of the uses of Utils.isMobile() at the moment are in class-based components. We can absolutely revisit down the line!

actions/views/browser.ts
deanwhillier: Well, I guess fall through is explicitly frowned upon in our linting. :P

crspeller (approved):

actions/views/browser.ts
michelengelen: this discussion is being held often and in many variations. Some might say that the `if/else if/else` approach adds readability or more clear intention, but IMO the `switch (true)` style is the better approach as well.

This is not only a matter of personal preference: there are several tests out there in the wild that prove that `switch` is computed faster than `if/else`

hmhealey: Also when looking for some objective reason to argue around, I made a quick benchmark to compare these. I tried to write it so that stuff doesn't get optimized out by the JS compiler, so hopefully I didn't write something that favours one over the other, but if-else does seem to be a tiny bit faster. That's on the scale of tens of microseconds on my machine though, so it's too small to matter either way. Guess we're stuck with personal preference here then lol.

hmhealey (approved):

jgilliam17 (approved):
Thanks @deanwhillier 
Tested, looks good to merge.
- Verified hiding and showing of the channel sidebar and Global Header when changing the sizing of the browser window above and below 768px.

amyblais:
@deanwhillier  Is this needed for v6.0?

deanwhillier:
@amyblais, I don't believe so. It is likely needed when we strip out the Global Header feature flag, but I don't believe we are doing that for 6.0.

PR #8871, opened 2021-09-16
[MM-25966] - Add new members to group messages with conversation history

ghost:
#### Summary
This PR adds the functionality to add users to group messages while maintaining the conversation history. It also adds the ability to leave the group message. Per discussion with @esethna we will address the Converting to private channel part in a separate ticket, as to keep this simple and reviewable without breaking too many changes both on the server and in the webapp.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-25966

#### Related Pull Requests
- Has server changes https://github.com/mattermost/mattermost-server/pull/18402

#### Release Note

-->
```release-note
NONE
```

### DIFF
--- a/components/more_direct_channels/more_direct_channels.tsx
+++ b/components/more_direct_channels/more_direct_channels.tsx
@@ -10,6 +10,8 @@ import {RelationOneToOne} from 'mattermost-redux/types/utilities';
 import {UserProfile} from 'mattermost-redux/types/users';
 import {GenericAction} from 'mattermost-redux/types/actions';
 
+import AlertBanner from 'components/alert_banner';
+import MultiSelect from 'components/multiselect/multiselect';
 import {browserHistory} from 'utils/browser_history';
 import Constants from 'utils/constants';
 
@@ -20,9 +22,6 @@ import {
     optionValue,
     OptionValue,
 } from './types';
-
-import AlertBanner from 'components/alert_banner';
-import MultiSelect from 'components/multiselect/multiselect';
 
 export type Props = {
     currentUserId: string;


components/add_user_to_group_message_modal/add_user_to_group_message_modal.tsx
```diff
@@ -0,0 +1,90 @@
...
+type Props = {
+    onHide: () => void;
+    actions: {
+        openModal: (modalData: { modalId: string; dialogType: any; dialogProps?: any }) => void;
+    };
+}
+
+const AddUserToGroupMessageModal: React.FC<Props> = (props: Props) => {
+    const dispatch = useDispatch();
+    const channel = useSelector((state: GlobalState) => getCurrentChannel(state));
```
michelengelen, line 27: typing?

components/add_user_to_group_message_modal/add_user_to_group_message_modal.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const currentTeamId = useSelector((state: GlobalState) => getCurrentTeamId(state));
```
michelengelen, line 28: same here: typing?

michelengelen (approved):
only remarks regarding typings in the component ... other than that LGTM! 🙇🏼‍♂️

esethna:
@nevyangelova thanks for working on this!! Is this PR ready for review? If not, can you help outline what has to be done before we can spin up a test server? The test server will pull the server changes as long as they have the same branch name

nevyangelova:
@esethna i believe we can do the first round of reviews! I have left all implementation for converting to private channel so we can have PRs for that separately and that includes the modals. I was planning to add the “convert to private channel” option when we implement that functionality. Let me know if this is what you have in mind as well.

esethna:
@nevyangelova yep, makes sense

esethna:
Nice Nevy! Really great start, a few functional things:

1. When a user chooses to "Start a new group" it should prefill the existing users into the Direct messages modal so the user just has to add more people
2. When a user chooses to "Start a new group" there is a missing modal element. We can just adjust remove the sentence about creating a private channel for now since that's coming later
image
3. If you try to add someone, but the GM already exists then we shouldn't create a new one. Currently it duplicates it in the sidebar:
image

Design:
image

4. Menu item should say "Leave Group Message"
image


5. Leave GM confirmation modal is currently using the leave private channel modal, see design for expected. I think we should fix it in this PR

image

6. Adding a user to a GM doesn't seem to happen over the websocket. ie User A is added, then on user A's screen they don't see the GM in their sidebar until they interact with the DM section (ie expand collapse categories, or similar action)

https://gyazo.com/452e34f7fce174409417043317416c8f

esethna:
After we do a first pass of the above items we can request review from @matthewbirtch

esethna (changes requested):
After we do a first pass of the above items we can request review from @matthewbirtch

nevyangelova:
@esethna addressed your points minus the one for the new modals. I had a discussion with UI team and those new modals are currently not in the codebase and are planned as a common component, after which we will replace all existing modals. @matthewbirtch can you confirm if this information is correct.

On the technical side I am getting some errors with leaving the GM channel it seems that it's memorising it and returning the currentUser to it even if they should wait to be added. Tagging @hmhealey for a review and maybe a hint if I am missing a crucial part of the logic for leaving the channel.

matthewbirtch:
> @esethna addressed your points minus the one for the new modals. I had a discussion with UI team and those new modals are currently not in the codebase and are planned as a common component, after which we will replace all existing modals. @matthewbirtch can you confirm if this information is correct.

Yes, modals have been redesigned: https://zeroheight.com/29be2c109/p/426c9b-modals

Not sure where they are in your roadmap, but the designs for this project utilized the new modal styles. I would be okay with not using the new modal style in this project until they are implemented as a common component though.

esethna:
Thanks @nevyangelova! 

1. When opening the "+" button in the sidebar we shouldn't show the element at the bottom of the modal (only show this when using the "Start a new group" flow)

image

2. Still seeing issue #3 (duplicate GMs)

image

components/alert_banner.tsx
```diff
@@ -9,7 +9,7 @@ import './alert_banner.scss';
 type Props = {
     mode: 'danger' | 'warning' | 'info';
     title?: React.ReactNode;
-    message: React.ReactNode;
+    message?: React.ReactNode;
```
michelengelen, line 12: why did you make that optional? IMO a alert banner should always have a message! :D

components/leave_private_channel_modal/leave_private_channel_modal.tsx
```diff
@@ -96,10 +96,26 @@ export default class LeavePrivateChannelModal extends React.PureComponent<Props,
...
+                    defaultMessage="You won't be able to rejoin this group unless you are invited again"
+                />
+            );
+        }
+
         const buttonClass = 'btn btn-danger';
         const button = (
             <FormattedMessage
-                id='leave_private_channel_modal.leave'
+                id={`${this.state.channel?.type === Constants.GM_CHANNEL ? 'leave_group_message_modal' : 'leave_private_channel_modal'}.leave`}
```
michelengelen, line 118: IMO the differentiation between private and group message could be done right at the beginning of the component. All these conditions seem a bit over the top. Maybe even set the `id`s and `defaultMessage`s right in the beginning. This would certainly improve readability

components/more_direct_channels/more_direct_channels.tsx
```diff
@@ -68,6 +69,7 @@ type Props = {
     recentDMUsers?: Array<UserProfile & {last_post_at: number }>;
     statuses: RelationOneToOne<UserProfile, string>;
     totalCount?: number;
+    prepopulatedValues?: [];
```
michelengelen, line 72: can you type this more strictly? Just an array seems to vague. I think you do know what beforehand what can go in there, right? (I think in this case it is `users`)

components/more_direct_channels/more_direct_channels.tsx
```diff
@@ -550,6 +552,19 @@ export default class MoreDirectChannels extends React.PureComponent<Props, State
...
+                mode='info'
+                title={(
+                    <FormattedMessage
+                        id='more_direct_channels.warning_for_groups'
+                        defaultMessage='This will start a new conversation'
+                    />
+                )}
+            />
+        );
+
```
michelengelen, line 567: OH, that's why you made the text prop optional ... got you!

i18n/en.json
```diff
@@ -3545,6 +3546,9 @@
   "last_users_message.others": "{numOthers} others ",
   "last_users_message.removed_from_channel.type": "were **removed from the channel**.",
   "last_users_message.removed_from_team.type": "were **removed from the team**.",
+  "leave_group_message_modal.leave": "Leave",
+  "leave_group_message_modal.message": "You won't be able to rejoin this group unless you are invited again",
```
michelengelen, line 3550: I would prefer to replacve `won't` with `will not`

michelengelen (changes requested):

nevyangelova:
@esethna after digging deeply into this and having a chat with @hmhealey we can confirm that the logic preventing the user from leaving a GM is quite extended in our current codebase. It will generate a lot of edge cases and there are some use case questions like what to do about duplication. Harrison addressed a lot of the edge cases in the comment here. https://github.com/mattermost/mattermost-server/pull/18402#issuecomment-926134496

If we do decide we wan't to proceed with the option to leave a DM then that would require a large change in architecture and some planning ahead. I will be happy to lead that discussion if the efforts prove to be worth it. But generally we wanted to give you an update on the scope of this. For adding the person to an existing DM this can still be implemented but some of the edge cases need to be discussed.

esethna:
For those following this, we've scheduled a discussion on Wednesday regarding next steps for this feature. Please let me know if you want to be included

stylianosrigas:
Removing spinwick for maintenance reasons.Please feel free to apply again if needed.

PR #8876, opened 2021-09-16
Mm 38364 download apps

neallred:
#### Summary
Add A/B test for displaying download apps action items as a tips and next steps panel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38364

#### Related Pull Requests
- Has server changes (https://github.com/mattermost/mattermost-server/pull/18404)

#### Screenshots
control:
control

as next step:
as_next_step

#### Release Note

```release-note
Add A/B test for displaying download apps action items as a tips and next steps panel.
```


components/next_steps_view/steps/enable_notifications_step/enable_notifications_step.tsx
```diff
@@ -57,12 +59,12 @@ export default function EnableNotificationsStep(props: StepComponentProps) {
...
             <div>
                 <TextCardWithAction
-                    cardBodyMessageId={'next_steps_view.notificationSetupNotificationsDisabled'}
+                    cardBodyMessageId={t('next_steps_view.notificationSetupNotificationsDisabled')}
                     cardBodyDefaultMessage={
                         'Notifications were previously disabled or you may be browsing in private mode. You\'ll need to open your browser settings or turn off private mode to enable notifications.'
                     }
                     buttonMessageId={
-                        'next_steps_view.notificationSetup.skipThisStep'
+                        t('next_steps_view.notificationSetup.skipThisStep')
```
neallred, line 67: Noticed that for `TextCardWithAction` usages, the ids weren't automatically picked up by `make i18n-extract`, so took the liberty of adding explicit markup for them to be included. Which means they are now added to `i18n/en.json`, too.

annelieseklein (approved):
looks good!

esethna:
@annelieseklein @neallred don't these buttons do the same thing? Should we remove one of them? 
image

neallred:
Good question! I'd suggest removing the "Skip" button in the card and keep "Skip Getting Started" at the bottom. It skips over the whole tutorial. We don't really use both the "skip" and "complete" options at the same time in any of the other cards, either.

esethna (approved):

stevemudie (approved):
Worked with @neallred to see both A & B and everything worked as expected. Looks good!

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
```diff
@@ -277,6 +278,20 @@ class InviteMembersStep extends React.PureComponent<Props, State> {
...
+            />
+        );
+        if (this.props.downloadAppsAsNextStep) {
+            finishMessage = (
+                <FormattedMessage
+                    id='next_steps_view.invite_members_step.next_step'
+                    defaultMessage='Next step'
+                />
+            );
+        }
```
marianunez, line 294: Wondering here if there is a way to have this step know if it's the last one and not tied to the specific step we are adding right now. We will be experimenting and changing steps/order so it would be ideal if it was smarter here

components/next_steps_view/steps/enable_notifications_step/enable_notifications_step.tsx
marianunez: Thanks!

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
marianunez: Doesn't need to be part of this PR specifically, just wondering if it's a possible future improvement

components/next_steps_view/steps/download_apps_step/download_apps_step.tsx
```diff
@@ -0,0 +1,40 @@
...
+import classNames from 'classnames';
+
+import {pageVisited} from 'actions/telemetry_actions';
+
+import {getAnalyticsCategory} from '../../step_helpers';
+
+import {StepComponentProps} from '../../steps';
+import DownloadSection from '../../download_section';
+
+export default function EnableNotificationsStep(props: StepComponentProps) {
```
marianunez, line 15:
```suggestion
export default function DownloadAppsStep(props: StepComponentProps) {
```

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
neallred: Yeah, I was thinking about that, too. Its fairly doable; you could pass in a "isLastStep" flag [here](https://github.com/mattermost/mattermost-webapp/blob/master/components/next_steps_view/next_steps_view.tsx#L227) . I think the bigger issue that needs addressing to do that is to standardize the button wording in these components a little more. Most of them are specific to the step at hand.

An alternate approach would be to rename the action in invite_members_step and the new download_apps_step to also be specific to the step at hand.

components/next_steps_view/steps.ts
```diff
@@ -91,6 +93,16 @@ export const Steps: StepType[] = [
...
     },
+    {
+        id: RecommendedNextSteps.DOWNLOAD_APPS,
+        title: localizeMessage(
+            'next_steps_view.downloadDesktopAndMobile',
+            'Download the Desktop and Mobile apps',
+        ),
+        roles: [],
+        component: DownloadAppsStep,
+        visible: true,
```
marianunez, line 104: Would it make sense to add the condition of the treatment here instead of filtering in the selector below?

marianunez (changes requested):
Minor change requests below

components/next_steps_view/steps.ts
neallred: I don't feel strongly either way. I lean 1/5 towards keeping the logic and setup as it is today.

That said, the current setup of defining the whole list of possibilities and filtering at the level of the "smart" redux-connected component is a little different than what I'd typically expect of redux passing down the info needed to filter, and the filtering happening within the component. Performance-wise, its memoized so I think its similar either way.

marianunez: 0/5 just thinking it would be cleaner to keep it all within the step logic given that the filtering is step agnostic. Also, for future code cleanup if we decide to not go with this treatment.

nickmisasi (approved):
LGTM!

components/next_steps_view/steps.ts
neallred: I might not be understanding what you mean by "keep it all within the step logic given that the filtering is step agnostic". What I _think_ you mean is do something like define `const Steps = [...]` and then on the next line do something like `if (hasTreatment) { Steps.push(theDownloadStep) }`.

Steps is defined outside a component so it doesn't have access to the data stored in redux or to any component props. So I think to change this we would need to either:
* place its definition within the component, and let the necessary data filter down from redux to build the list
* make a helper function that takes the args that are used in the selector function, which would be really similar to the selector function.

marianunez: Actually more something like: 
```suggestion
        visible: getDownloadAppsCTATreatment(state) === DownloadAppsCTATreatments.TIPS_AND_NEXT_STEPS,
```

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
marianunez: Let's create a ticket for this to see what we can do to improve this in the future

marianunez (approved):
Thanks @neallred!

components/next_steps_view/steps.ts
neallred: Ahh. That does seem like a good use for `visible` and it does keep it more step agnostic. But with Steps defined outside of a context that has access to redux `state`, I still think we'd need to place it in the component, in the selector function , or make a helper function.

components/next_steps_view/steps/invite_members_step/invite_members_step.tsx
neallred: Created as https://mattermost.atlassian.net/browse/MM-38643

PR #8886, opened 2021-09-17
MM-38579: fix town-square being always read-only

lieut-data:
#### Summary
A recent change to remove the experimental town square feature make town square permanently read-only for non system admins. (We don't notice on community since we were used to this setting being enabled).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38579?filter=15018

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/create_comment/index.ts
+++ b/components/create_comment/index.ts
@@ -80,7 +80,7 @@ function makeMapStateToProps() {
             codeBlockOnCtrlEnter: getBool(state, Preferences.CATEGORY_ADVANCED_SETTINGS, 'code_block_ctrl_enter', true),
             ctrlSend: getBool(state, Preferences.CATEGORY_ADVANCED_SETTINGS, 'send_on_ctrl_enter'),
             createPostErrorId: err.server_error_id,
-            readOnlyChannel: !isCurrentUserSystemAdmin(state) && channel.name === Constants.DEFAULT_CHANNEL,
+            readOnlyChannel: false,
             enableConfirmNotificationsToChannel,
             enableEmojiPicker,
             enableGifPicker,
--- a/components/create_post/index.ts
+++ b/components/create_post/index.ts
@@ -112,7 +112,7 @@ function makeMapStateToProps() {
             latestReplyablePostId,
             locale: getCurrentLocale(state),
             currentUsersLatestPost: getCurrentUsersLatestPost(state, ''),
-            readOnlyChannel: ownProps.readOnlyChannel || (!isCurrentUserSystemAdmin(state) && currentChannel.name === Constants.DEFAULT_CHANNEL),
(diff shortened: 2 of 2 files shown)


components/create_comment/index.ts
```diff
@@ -80,7 +80,7 @@ function makeMapStateToProps() {
             codeBlockOnCtrlEnter: getBool(state, Preferences.CATEGORY_ADVANCED_SETTINGS, 'code_block_ctrl_enter', true),
             ctrlSend: getBool(state, Preferences.CATEGORY_ADVANCED_SETTINGS, 'send_on_ctrl_enter'),
             createPostErrorId: err.server_error_id,
-            readOnlyChannel: !isCurrentUserSystemAdmin(state) && channel.name === Constants.DEFAULT_CHANNEL,
+            readOnlyChannel: false,
```
lieut-data, line 83: Not sure if this is correct, or if there are more changes to be made here.

BenCookie95: I wonder if we can just remove the readOnlyChannel prop from this component (`CreateComment`) and wherever it was used we default to the `false` value of the logic?

lieut-data: Yeah, that's the weird thing: there appears to be no `readOnlyChannel` prop for this component, just `CreatePost`. I wonder if the functionality was just broken to begin with?

BenCookie95: I'm not sure if I understand because I can see the `readOnlyChannel` prop being used in the `CreateComment` component?

lieut-data: Ah, I misunderstood your comment. You're right that we can just remove it: but I'm worried about the divergent behaviour between `CreatePost` and `CreateComment` here, given that the former actually had a `readOnlyChannel` prop that clients of `CreatePost` would sometimes set. Are we 100% certain that `CreateComment` should never provide a read-only experience?

BenCookie95: That's a good point! I'm going to read through the components a bit to understand the `readOnlyChannel` prop a bit better. I'll let you know what I think

BenCookie95: @lieut-data I'm under the impression we can remove that prop from both components. The CreatePost component is only used in one place and that prop is unused. The prop was introduced a few years ago (https://github.com/mattermost/mattermost-webapp/pull/1415) but wasn't even in use back then

lieut-data:
@hmhealey, if you can take a look at this one (or re-assign), it would be great to get this merged ASAP. Playbooks' E2E testing pipeline is currently blocked on this Town Square regression.

BenCookie95 (approved):

cpoile (approved):
👍

PR #8887, opened 2021-09-17
Migrate to map some mapped objects

larkox:
#### Summary
I have migrated some mapped objects to maps, as per the discussion here: https://community.mattermost.com/core/pl/nztwmtnj9jdqfeb5rnyxyaat4w

I will make some notes directly in the code.

#### Ticket Link
None

#### Release Note
```release-note
None
```

### DIFF
--- a/components/add_users_to_team_modal/add_users_to_team_modal.tsx
+++ b/components/add_users_to_team_modal/add_users_to_team_modal.tsx
@@ -26,13 +26,13 @@ type Props = {
     team: Team;
     users: UserProfile[];
     filterExcludeGuests?: boolean;
-    excludeUsers: { [userId: string]: UserProfile };
-    includeUsers: { [userId: string]: UserProfile };
+    excludeUsers: Map<string, UserProfile>;
+    includeUsers: Map<string, UserProfile>;
     onAddCallback: (users: UserProfile[]) => void;
     onHide?: () => void;
 
     actions: {
-        getProfilesNotInTeam: (teamId: string, groupConstrained: boolean, page: number, perPage?: number, options?: Record<string, any>) => Promise<{ data: UserProfile[] }>;
+        getProfilesNotInTeam: (teamId: string, groupConstrained: boolean, page: number, perPage?: number) => Promise<{ data: UserProfile[] }>;
         searchProfiles: (term: string, options?: Record<string, any>) => Promise<{ data: UserProfile[] }>;
     };
 }
@@ -45,7 +45,7 @@ type State = {
     saving: boolean;
     addError: null;
     loading: boolean;
-    filterOptions: {[key: string]: any};
+    filterOptions: Map<string, any>;
 }
 
 export default class AddUsersToTeamModal extends React.PureComponent<Props, State> {
@@ -54,9 +54,9 @@ export default class AddUsersToTeamModal extends React.PureComponent<Props, Stat
     public constructor(props: Props) {
         super(props);
 
-        let filterOptions = {};
+        const filterOptions = new Map();
         if (props.filterExcludeGuests) {
-            filterOptions = {role: 'system_user'};
+            filterOptions.set('role', 'system_user');
         }
 
         this.state = {
@@ -86,7 +86,7 @@ export default class AddUsersToTeamModal extends React.PureComponent<Props, Stat
         let searchResults: UserProfile[] = [];
         const search = term !== '';
         if (search) {
-            const {data} = await this.props.actions.searchProfiles(term, {not_in_team_id: this.props.team.id, replace: true, ...this.state.filterOptions});
+            const {data} = await this.props.actions.searchProfiles(term, {not_in_team_id: this.props.team.id, replace: true, ...Object.fromEntries(this.state.filterOptions)});
             searchResults = data;
         } else {
             await this.props.actions.getProfilesNotInTeam(this.props.team.id, false, 0, USERS_PER_PAGE * 2);
@@ -202,11 +202,11 @@ export default class AddUsersToTeamModal extends React.PureComponent<Props, Stat
         let usersToDisplay: UserProfile[] = [];
         usersToDisplay = this.state.search ? this.state.searchResults : this.props.users;
         if (this.props.excludeUsers) {
-            const hasUser = (user: UserProfile) => !this.props.excludeUsers[user.id];
+            const hasUser = (user: UserProfile) => !this.props.excludeUsers.get(user.id);
             usersToDisplay = usersToDisplay.filter(hasUser);
         }
         if (this.props.includeUsers) {
-            const includeUsers = Object.values(this.props.includeUsers);
+            const includeUsers = this.props.includeUsers.values();
             usersToDisplay = [...usersToDisplay, ...includeUsers];
         }
 
--- a/components/admin_console/data_retention_settings/channel_list/channel_list.tsx
+++ b/components/admin_console/data_retention_settings/channel_list/channel_list.tsx
@@ -200,9 +200,9 @@ export default class ChannelList extends React.PureComponent<Props, State> {
                 );
             }
             return {
-                cells: {
-                    id: channel.id,
-                    name: (
+                cells: new Map<string, React.ReactNode>([
+                    ['id', channel.id],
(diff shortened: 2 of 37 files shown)


components/add_users_to_team_modal/add_users_to_team_modal.tsx
```diff
@@ -26,13 +26,13 @@ type Props = {
...
-    excludeUsers: { [userId: string]: UserProfile };
-    includeUsers: { [userId: string]: UserProfile };
+    excludeUsers: Map<string, UserProfile>;
+    includeUsers: Map<string, UserProfile>;
     onAddCallback: (users: UserProfile[]) => void;
     onHide?: () => void;
 
     actions: {
-        getProfilesNotInTeam: (teamId: string, groupConstrained: boolean, page: number, perPage?: number, options?: Record<string, any>) => Promise<{ data: UserProfile[] }>;
+        getProfilesNotInTeam: (teamId: string, groupConstrained: boolean, page: number, perPage?: number) => Promise<{ data: UserProfile[] }>;
```
larkox, line 35: This is an unrelated change. It is removed since the original function doesn't seem to accept any options.

components/add_users_to_team_modal/add_users_to_team_modal.tsx
```diff
@@ -86,7 +86,7 @@ export default class AddUsersToTeamModal extends React.PureComponent<Props, Stat
         let searchResults: UserProfile[] = [];
         const search = term !== '';
         if (search) {
-            const {data} = await this.props.actions.searchProfiles(term, {not_in_team_id: this.props.team.id, replace: true, ...this.state.filterOptions});
+            const {data} = await this.props.actions.searchProfiles(term, {not_in_team_id: this.props.team.id, replace: true, ...Object.fromEntries(this.state.filterOptions)});
```
larkox, line 89: Expansions are tricky. We could expand the map directly, but then we would be expanding the map object, which is not what we want. We would have to be extra careful with these.

larkox: If we were to use the `{[x: string]: myType | undefined}` approach, we would have also problems, since expanding the object will consider some values undefined.

components/add_users_to_team_modal/add_users_to_team_modal.tsx
```diff
@@ -202,11 +202,11 @@ export default class AddUsersToTeamModal extends React.PureComponent<Props, Stat
         let usersToDisplay: UserProfile[] = [];
         usersToDisplay = this.state.search ? this.state.searchResults : this.props.users;
         if (this.props.excludeUsers) {
-            const hasUser = (user: UserProfile) => !this.props.excludeUsers[user.id];
+            const hasUser = (user: UserProfile) => !this.props.excludeUsers.get(user.id);
             usersToDisplay = usersToDisplay.filter(hasUser);
         }
         if (this.props.includeUsers) {
-            const includeUsers = Object.values(this.props.includeUsers);
+            const includeUsers = this.props.includeUsers.values();
```
larkox, line 209: In comparison with expanding an object, this works fine.

Nevertheless, the migration may give us lots of headaches, since "Object.values" may be "type-correct", and we won't see the problem until testing.

components/admin_console/data_retention_settings/channel_list/channel_list.tsx
```diff
@@ -200,9 +200,9 @@ export default class ChannelList extends React.PureComponent<Props, State> {
                 );
             }
             return {
-                cells: {
-                    id: channel.id,
-                    name: (
+                cells: new Map<string, React.ReactNode>([
```
larkox, line 203: Creating maps seems a bit more troublesome, but doesn't seem a big deal. But I have to admit that migrating these is tedious.

components/admin_console/data_retention_settings/channel_list/channel_list.tsx
```diff
@@ -275,71 +275,77 @@ export default class ChannelList extends React.PureComponent<Props, State> {
...
     onFilter = async (filterOptions: FilterOptions) => {
         const filters: ChannelSearchOpts = {};
-        const {public: publicChannels, private: privateChannels, deleted} = filterOptions.channels.values;
-        const {team_ids: teamIds} = filterOptions.teams.values;
-        if (publicChannels.value || privateChannels.value || deleted.value || (teamIds.value as string[]).length) {
-            filters.public = publicChannels.value as boolean;
-            filters.private = privateChannels.value as boolean;
-            filters.deleted = deleted.value as boolean;
-            filters.team_ids = teamIds.value as string[];
+        const channelsFilter = filterOptions.get('channels');
```
larkox, line 278: These situations become a bit more verbose, but I don't think is a big problem.

components/admin_console/filter/filter.tsx
```diff
@@ -61,14 +55,14 @@ class Filter extends React.PureComponent<Props, State> {
     public constructor(props: Props) {
         super(props);
 
-        let options = {...props.options};
+        let options = new Map(props.options);
```
larkox, line 58: This looks prettier in my opinion, but I guess it is about taste.

components/admin_console/team_channel_settings/channel/details/channel_members/channel_members.tsx
```diff
@@ -153,19 +153,22 @@ export default class ChannelMembers extends React.PureComponent<Props, State> {
...
+        const roles = filterOptions.get('role')?.values;
+        if (!roles) {
+            return;
+        }
         const systemRoles: string[] = [];
         const channelRoles: string[] = [];
         let filters = {};
-        Object.keys(roles).forEach((filterKey: string) => {
-            if (roles[filterKey].value) {
+        for (const [filterKey, filterValue] of roles) {
```
larkox, line 163: This way to handle the loops seems elegant too.

components/admin_console/team_channel_settings/group/group_users/users_to_remove.tsx
```diff
@@ -239,89 +243,85 @@ export default class UsersToRemove extends React.PureComponent<Props, State> {
...
-                        <FormattedMessage
-                            id='admin.user_grid.team_admin'
-                            defaultMessage='Team Admin'
-                        />
-                    ),
-                    value: false,
-                },
-            };
-            filterOptions.role.keys = [GeneralConstants.SYSTEM_GUEST_ROLE, GeneralConstants.TEAM_USER_ROLE, GeneralConstants.TEAM_ADMIN_ROLE, GeneralConstants.SYSTEM_ADMIN_ROLE];
+            roleFilter.values.set(GeneralConstants.TEAM_USER_ROLE, {
```
larkox, line 301: Adding and removing elements now is also a bit different. I feel it is clearer, but not sure.

larkox:
I am not sure about this approach. In one hand, I feel it improves some things, but I foresee a lot of headaches, specially on non typescript files, since we would be changing not only the typing, but the actual object being used. Also, it throws some false positives in the linter that I am not sure how to handle.

The other alternative (`{[x: string]: myType | undefined}`) seem to add also lots of headaches with the typings during expansions. Sincerely, not sure if we should follow this approach or not.

components/add_users_to_team_modal/add_users_to_team_modal.tsx
hmhealey: That's very true. I've definitely ran into a few issues like that in the past when I've tried to use Maps elsewhere like on mobile. It's also really easy to write code like `map['foo'] = 'bar'` which is valid code even if it's not doing what you want. I think that's partly why I gave up trying to use Maps before we had TypeScript

components/admin_console/filter/filter.tsx
```diff
@@ -124,10 +118,15 @@ class Filter extends React.PureComponent<Props, State> {
     }
 
     updateValues = async (values: FilterValues, optionKey: string) => {
-        const options = {
+        const currentValue = this.state.options.get(optionKey);
+        if (!currentValue) {
+            return;
+        }
+
+        const options: FilterOptions = {
```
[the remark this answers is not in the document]
larkox: Probably not. Not sure why it is not complaining here.

PR #8889, opened 2021-09-17
[MM-24461] Show last active time of users when not online

BenCookie95:
#### Summary
Show when a user was last active. I added a new item to `user.props` for the last active time of each user. Update channel header and profile popover to add the new label.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24461

#### Related PRs
- https://github.com/mattermost/mattermost-server/pull/19126

#### Release Note

```release-note
Show when a user was last online
```


packages/mattermost-redux/src/utils/user_utils.ts
```diff
@@ -234,3 +234,17 @@ export function applyRolesFilters(user: UserProfile, filterRoles: string[], excl
     }
     return filterRoles.length === 0 || filterRoles.some(checkUserHasRole.bind(null, user, userIsNotAdminOrGuest, membership));
 }
+
+export function displayLastActiveLabel(status: string, timestamp: number, showLastActive: boolean | undefined) {
+    const currentTime = new Date();
+    const oneMin = 5 * 60 * 1000;
```
devinbinnie, line 240: We might have a global constant for this.

devinbinnie (approved):

esethna:
Friendly ping @mkraft to help review the web and server PRs for this feature

BenCookie95:
> @BenCookie95 I don't see the last active status showing on the test server anywhere. Is there a setting that needs to be enabled?

@matthewbirtch Looks like there is an issue right now with the display settings, the settings page is crashing. I will probably need to sync with master and fix the conflicts

BenCookie95:
@agnivade I just wanted to clarify with server. I added a new user setting (`show_last_active`) to `user.props`, this setting will be viewed by other users in order to determine whether we can display their last active time.

I added the field to `user.props` in order to avoid adding a new column to the users table. Is this the approach server wants us to take?

agnivade:
> I added the field to user.props in order to avoid adding a new column to the users table. Is this the approach server wants us to take?

@BenCookie95 - It all depends on the requirements for the new field. The decision to use a new column or not, is dictated by the query patterns on the data. Keep in mind that any data inside a jsonb blob takes more storage than a dedicated column and the query planner also does not store any statistics on it.

Which essentially means that if you are going to use that field in WHERE clauses, it will degrade performance than a usual column. Arguably, Postgres does allow you to index fields in a jsonb column, but for MySQL, you need to create a generated column and index on that. Supporting MySQL creates more complexity.

My suggestion would be that if this is treated purely as a metadata on the user object, and you are not going to perform any sort of queries on that field, then it's okay to add it as a prop. Otherwise, use a dedicated column.

BenCookie95:
@devinbinnie & @mkraft I'm opening this up for re-review because there have been a number of changes since the last review. I am now using user.props to store the new configuration for each user. Previously I added a new column to the users table but I decided against it.

devinbinnie (approved):

mkraft (approved):

matthewbirtch:
@BenCookie95 let me know when the test server is in a place where I can review this from a UX standpoint. Thanks!

BenCookie95:
@matthewbirtch should be good to go

furqanmlk (commented):
Hi @BenCookie95 

All test conditions are passing except one,

1. If less than 1 minute, don’t show active status ✅ 

2. “Active [N] min(s) ago”, if >= 1 minute and < 60 minutes ✅ 
image

3. “Active [N] hr(s) ago”, if >= 1 hour and < 24 hrs ✅ 
**Note: Should it `1 Hour` or `1 Hours` ?**
Pasted Graphic 15

4. “Active yesterday”, if >= 24 hrs and < 48 hrs ✅ 
Pasted Graphic 16

5. “Active [Friday]”, if >=48 hrs and < 7 days 🔴 --- **FAILED**
image

6. “Active [August 10]”, if >= 7 days ✅ 
image

matthewbirtch (changes requested):
This looks great. Thanks @BenCookie95 

Just a few minor things from me:
1. Can we change the time format to abbreviated? (e.g. 2 hrs ago, 5 mins ago, etc)
2. When a custom status is set, can we add some space to the left of it so that it's more separated? I think it needs about 8px of space to the left of the custom status
Screen Shot 2021-11-12 at 2 06 45 PM
3. In the Settings, can we move the 'Share last active time' setting up under 'Show online availability on profile images'?
4. When I toggled the setting off it hid the last active status from other users as I expected. However, when I toggled it back on, it wouldn't save the setting properly. It seems that I am unable to turn it back on after turning it off.

I think that's it! Nice work.

BenCookie95:
> This looks great. Thanks @BenCookie95
> 
> Just a few minor things from me:
> 
> 1. Can we change the time format to abbreviated? (e.g. 2 hrs ago, 5 mins ago, etc)
> 2. When a custom status is set, can we add some space to the left of it so that it's more separated? I think it needs about 8px of space to the left of the custom status
> 
> Screen Shot 2021-11-12 at 2 06 45 PM
> 
> 3. In the Settings, can we move the 'Share last active time' setting up under 'Show online availability on profile images'?
> 4. When I toggled the setting off it hid the last active status from other users as I expected. However, when I toggled it back on, it wouldn't save the setting properly. It seems that I am unable to turn it back on after turning it off.
> 
> I think that's it! Nice work.

@matthewbirtch I think I have fixed most things from your comments. The only issue is with the abbreviations for minutes/hours. The library we are using seems to shorten to `min.` & `hr.` instead of `mins` & `hrs`

BenCookie95:
> Hi @BenCookie95
> 
> All test conditions are passing except one,
> 
> 1. If less than 1 minute, don’t show active status ✅
> 2. “Active [N] min(s) ago”, if >= 1 minute and < 60 minutes ✅
>    image
> 3. “Active [N] hr(s) ago”, if >= 1 hour and < 24 hrs ✅
> 
> Pasted Graphic 15
> 
> 4. “Active yesterday”, if >= 24 hrs and < 48 hrs ✅
> 
> Pasted Graphic 16
> 
> 5. “Active [Friday]”, if >=48 hrs and < 7 days 🔴 --- **FAILED**
>    image
> 6. “Active [August 10]”, if >= 7 days ✅
>    image

@furqanmlk I have fixed the broken test case for number 5.

matthewbirtch:
@BenCookie95 I can't seem to get the last active time to display. The setting is turned on for both users I'm testing with, but it's not showing the last active time anywhere

image

image

BenCookie95:
> @BenCookie95 I can't seem to get the last active time to display. The setting is turned on for both users I'm testing with, but it's not showing the last active time anywhere
> 
> image
> 
> image

@matthewbirtch I'll take another look, I might have broken it on the last push

BenCookie95:
@matthewbirtch seems to be working fine for me on the test server and locally. Is it ok now?

matthewbirtch:
@BenCookie95 seems to be showing now. Weird

matthewbirtch:
> 2\. When a custom status is set, can we add some space to the left of it so that it's more separated? I think it needs about 8px of space to the left of the custom status

Space looks better here, but it looks like space was added for all cases - not just when custom status is set?

BenCookie95:
> > 2. When a custom status is set, can we add some space to the left of it so that it's more separated? I think it needs about 8px of space to the left of the custom status
> 
> Space looks better here, but it looks like space was added for all cases - not just when custom status is set?

Yeah that's true actually, that's a quick fix

BenCookie95:
> > 2. When a custom status is set, can we add some space to the left of it so that it's more separated? I think it needs about 8px of space to the left of the custom status
> 
> Space looks better here, but it looks like space was added for all cases - not just when custom status is set?

@matthewbirtch Should this padding only be added to the custom status when the last active time is showing? or should it be applied to custom status no matter what is to the left of it?

matthewbirtch:
> > > 2. When a custom status is set, can we add some space to the left of it so that it's more separated? I think it needs about 8px of space to the left of the custom status
> > 
> > 
> > Space looks better here, but it looks like space was added for all cases - not just when custom status is set?
> 
> @matthewbirtch Should this padding only be added to the custom status when the last active time is showing? or should it be applied to custom status no matter what is to the left of it?

Good point, I think it would be good to be there no matter what

BenCookie95:
> > > > 2. When a custom status is set, can we add some space to the left of it so that it's more separated? I think it needs about 8px of space to the left of the custom status
> > > 
> > > 
> > > Space looks better here, but it looks like space was added for all cases - not just when custom status is set?
> > 
> > 
> > @matthewbirtch Should this padding only be added to the custom status when the last active time is showing? or should it be applied to custom status no matter what is to the left of it?
> 
> Good point, I think it would be good to have the extra padding to the left of the custom status no matter what

Ok, that's fixed now

matthewbirtch (approved):
LGTM!

esethna:
@furqanmlk please help with testing when available. Thanks!

esethna:
We probably shouldn't show the last active time for your own DM: 
image

furqanmlk (approved):
Locally passed. Working as expected
image

esethna:
@BenCookie95 sounds like this is ready to merge?

BenCookie95:
> @BenCookie95 sounds like this is ready to merge?

I'm going to merge latest master in first to ensure pipelines haven't broke and then I will merge afterwards

BenCookie95:
@esethna One final thought on this. I wonder if there should be a config to enable/disable this setting globally? i.e a config that make this setting available for the everyone in the org

esethna:
@BenCookie95 good call. I'm not opposed to adding that, maybe at the bottom of the users and teams section of the sys console. 

"Enable last active time"
"When enabled, last active time allows users to see when someone was last online.

BenCookie95:
> @BenCookie95 good call. I'm not opposed to adding that, maybe at the bottom of the users and teams section of the sys console.
> 
> "Enable last active time" "When enabled, last active time allows users to see when someone was last online."
> 
> Defaulted to true

Yeah I can add that on Friday morning!

BenCookie95:
@devinbinnie @mkraft I added a new config to admin console to control the last active label globally. Just though you might want a quick look so I asked for reviews again

devinbinnie (approved):

components/user_settings/display/user_settings_display.tsx
```diff
@@ -184,6 +191,35 @@ export default class UserSettingsDisplay extends React.PureComponent<Props, Stat
...
+            props: {
+                ...user.props,
+                show_last_active: lastActiveDisplay,
+            },
+        };
+
+        actions.updateMe(updatedUser).
+            then((res) => {
+                if ('data' in res) {
+                    this.props.updateSection('');
```
mkraft, line 209: Was this meant to have an empty string parameter?

esethna:
Same here @BenCookie95! Please go ahead an merge if we've got the reviews as needed :)

BenCookie95:
> Same here @BenCookie95! Please go ahead an merge if we've got the reviews as needed :)

I had to do some code cleanup due to some performance issues so that will need a re-review. I also found a memory leak in the existing timestamp component, I think my changes are the only components that use the timestamp in this way so that's why the memory leak is only now visible. 

I'm going to fix the memory leak in this PR as well

esethna:
@BenCookie95 just checking in to see if you're still working on this?

esethna:
Chatted with @BenCookie95. He's going to take another glance at this before merging, but the memory leak issue does not seem to be related to this PR in particular, so reviews should be complete

PR #8900, opened 2021-09-20
Mm 38326 getting started position

neallred:
#### Summary
Fix some lhs sidebar & team sidebar height issues when both the global header and the announcement bar are enabled. In the CWS linked spinwick, users should see the announcement bar from the free trial, enabling testing.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38326

#### Screenshots
before:
after

after:
before

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -79,10 +79,6 @@
             z-index: 20;
             top: 0;
 
-            body.announcement-bar--fixed & {
-                height: 100%;
-            }
-
             .sidebar-right-container {
                 height: 100%;
             }
--- a/components/global/global_header.scss
+++ b/components/global/global_header.scss
@@ -38,15 +38,18 @@
 
             // adjust height of team sidebar
             height: calc(100% - 40px);
+
             body.announcement-bar--fixed & {
                 height: calc(100% - 80px);
             }
+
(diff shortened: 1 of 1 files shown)


components/global/global_header.scss
```diff
@@ -38,12 +38,21 @@
 
             // adjust height of team sidebar
             height: calc(100% - 40px);
+
+            body.announcement-bar--fixed & {
+                height: calc(100% - 80px);
```
pablovelezvidal, line 43: since the webapp renders filling the full viewport and adding specific scroll to internal elements, wouldn't it make more sense  to use vh instead of percentage for this calculation?

stevemudie (approved):
The Global Header was enabled and the Getting Started at the bottom of the LHS was still visible as was the + to add a new team. I also tested it with a banner put in place by the admin and everything worked as expected.

Looks good @neallred!

AGMETEOR (approved):

components/global/global_header.scss
neallred: @pablovelezvidal is this ok as is for now?

pablovelezvidal (approved):
LGTM! 🚢

PR #8914, opened 2021-09-21
fixed messaging test case

furqanmlk:
#### Summary
Fixed messaging test case
#### Screenshots
image
image
image
image

### DIFF
--- a/e2e/cypress/integration/auth_sso/authentication_4_spec.js
+++ b/e2e/cypress/integration/auth_sso/authentication_4_spec.js
@@ -280,7 +280,7 @@ describe('Authentication', () => {
         cy.findByText('Add members or email addresses').type('HosseinTheBestProgrammer@Mattermost.com{downarrow}{downarrow}{enter}');
 
         // # Click invite members button
-        cy.findByRole('button', {name: /invite members/i}).click({force: true});
+        cy.findByRole('button', {name: 'Invite Members'}).click({force: true});
 
         // * Verify message is what you expect it to be
         cy.contains('The following email addresses do not belong to an accepted domain:', {timeout: TIMEOUTS.ONE_MIN}).should('be.visible').and('exist');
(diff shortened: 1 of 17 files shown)


e2e/cypress/integration/auth_sso/authentication_4_spec.js
```diff
@@ -277,10 +277,10 @@ describe('Authentication', () => {
         cy.findByText('Members').click();
 
         // # Input email, select member
-        cy.findByText('Add members or email addresses').type('HosseinTheBestProgrammer@Mattermost.com{enter}{enter}');
+        cy.findByText('Add members or email addresses').type('HosseinTheBestProgrammer@Mattermost.com{downarrow}{downarrow}{enter}');
 
         // # Click invite members button
-        cy.findByText('Invite Members').click();
+        cy.findByRole('button', {name: /invite members/i}).click({force: true});
```
saturninoabril, line 283: I'd suggest to make it case sensitive since it's how the user sees it. I think we can use insensitive for case like aria-label and other similar which is not visible to actual user.  
```suggestion
        cy.findByRole('button', {name: 'Invite Members'}).click({force: true});
```

saturninoabril (changes requested):
Thanks @furqanmlk! This is awesome. Just one request and I'm all good.

saturninoabril (approved):
Thanks @furqanmlk!

e2e/cypress/integration/messaging/message_ephemeral_spec.js
```diff
@@ -11,22 +11,22 @@
 // Group: @messaging
 
 describe('Hide ephemeral message on refresh', () => {
-    let townsquareLink;
+    let offtopiclinkLink;
```
josephbaylon, line 14:
```suggestion
    let offtopiclink;
```

josephbaylon (approved):
LGTM. Just one minor non-blocking comment. Please rename variable below and all affected lines.

PR #8926, opened 2021-09-22
[MM 38105] Bug fix: apply theme in root component instead of needs team component

deanwhillier:
#### Summary
When using a single theme for all teams that isn't the default team (Denim), logging out and in again incorrectly applies the default theme (Denim) until the browser is refreshed, at which point the correct theme is applied. This does not happen when using per-team themes.

For some reason, the `NeedsTeam` component that handles applying the users selected theme when it changes, is not updated following initialization to trigger applying the users theme when sharing a single theme across all teams but it is updated if using per-team themes. In both cases, the users selected theme is correctly passed as a prop to the `NeedsTeam` component.

I spent a fair amount of time trying to track down why this is happening, but didn't get very far so left the exploration to another day. Since we now use the theme in the `Root` component, I elected to move the logic that applies the theme when its prop value changes to the `Root` component, which is always updating when the theme changes regardless of theme type. The Root component is already responsible for applying the default theme when not logged in or when user preferences are loaded when logged in after refreshing, so I didn't see this approach as being a problem.

NOTE: This was not a recent regression and was also happening before the recent system theme updates for v6.0. I did not spend more time trying to determine how far back this bug goes though, so I'm not even sure if it is a regression or not.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38105

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/needs_team/index.ts
+++ b/components/needs_team/index.ts
@@ -8,7 +8,7 @@ import {withRouter} from 'react-router-dom';
 import {fetchMyChannelsAndMembers, viewChannel} from 'mattermost-redux/actions/channels';
 import {getMyTeamUnreads, getTeamByName, selectTeam} from 'mattermost-redux/actions/teams';
 import {getGroups, getAllGroupsAssociatedToChannelsInTeam, getAllGroupsAssociatedToTeam, getGroupsByUserId} from 'mattermost-redux/actions/groups';
-import {getTheme, isCollapsedThreadsEnabled} from 'mattermost-redux/selectors/entities/preferences';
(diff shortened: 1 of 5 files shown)


jgilliam17 (approved):
Thanks @deanwhillier 
Tested, looks good to merge.
- Verified when using non-Denim theme, logging out and in again no longer temporarily applies Denim theme to teams.

saturninoabril (approved):
Thanks @deanwhillier! LGTM. Tested and passed.

hmhealey (approved):
I wonder if the issue is something like a mount vs update thing since we're only checking for the theme change on update. If the theme prop didn't change after the component was mounted, then `applyTheme` would never get called

PR #8929, opened 2021-09-22
MM-38006 Fix collapsed post contents sometimes scrolling

hmhealey:
This PR is primarily to fix a weird bug where opening and closing the file preview modal on a partially-hidden Markdown image in a collapsed post causes the post contents to scroll within an `overflow: hidden` container. Yeah, I don't get why it happens either.

Either way, it's fixed by switching that to `overflow: clip` which is the same except it makes the contents unscrollable. That's not supported on Safari yet so this bug will still exist there, but it's coming soon since it's available in the Safari Technical Preview.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38006

#### Release Note
```release-note
NONE
```


deanwhillier (approved):
LGTM. I think that this is small enough that current versions of Safari not yet supporting the new option is okay. It's too bad we don't use a newer version of React Bootstrap to be able make use of the `restoreFocusOptions` prop for the Modal component to prevent scrolling when the modal restores focus to the image as it is dismissed. 🤷‍♂️

AshishDhama (commented):
@hmhealey LGTM, please update the failing test snapshot

hmhealey:
Fixed. That's what I get for fixing a change that isn't in the ticket 😛

AshishDhama (approved):
LGTM

amyblais:
@hmhealey Is this for v6.0 or v6.1? (ticket and PR milestones are different)

hmhealey:
I think 6.1 is fine. It's a weird case that only happens under some pretty specific circumstances (when the user clicks on an inline image in a collapsed post). I'll update the Jira ticket

PR #8940, opened 2021-09-24
[MM-38344] + [MM-38347] - Clean up: apply style adjustments to component styles

michelengelen:
#### Summary
this PR aims to apply all style adjustments, we had to make during the global header implementation phase (collected in one big `global-header.scss` file), to the respective component level scss files.

It removes the setting of the global header class from the root, since it is not needed anymore.

#### Ticket Link
[MM-38344](https://mattermost.atlassian.net/browse/MM-38344)
[MM-38347](https://mattermost.atlassian.net/browse/MM-38347)

#### Release Note
```release-note
NONE
```


michelengelen:
@matthewbirtch I have added a review request from you as well to be safe that nothing from a UX standpoint broke during this change. Feel free to remove the request when you see no need to have a look.

nevyangelova (approved):
Thanks for handling this! Much cleaner now 👍

matthewbirtch:
> @matthewbirtch I have added a review request from you as well to be safe that nothing from a UX standpoint broke during this change. Feel free to remove the request when you see no need to have a look.

Thanks @michelengelen. I tested across a few key screens. The only thing I noticed is that the global header now displays always - even in mobile webview (<768 viewport widths).

image

michelengelen:
@matthewbirtch I fixed the mis-displaying of global header in mobile view.

matthewbirtch (approved):
Thanks @michelengelen

jgilliam17:
Thanks @michelengelen GH looks as before, not seeing any change.
One thing though - there seems to be a bit more space between team name and channel switcher, not sure if this was an intended.
Screen Shot 2021-09-24 at 12 13 56 PM

matthewbirtch:
> Thanks @michelengelen GH looks as before, not seeing any change.
> One thing though - there seems to be a bit more space between team name and channel switcher, not sure if this was an intended.

I'm not seeing that.
image

michelengelen:
@nevyangelova please re-review

nevyangelova:
@michelengelen quite a few tests are failing

jgilliam17:
@michelengelen Looks good.
Question about mobile webview - Seeing Custom Emoji and Keyboard shortcuts, not sure if we should be surfacing those. Unless this was a recent change I missed
Screen Shot 2021-09-27 at 3 31 11 PM

michelengelen:
@jgilliam17 that is something we should get some input on. I can see even more stuff there:

Screenshot 2021-09-28 at 09 12 24

/cc @esethna

matthewbirtch:
> @michelengelen Looks good. Question about mobile webview - Seeing Custom Emoji and Keyboard shortcuts, not sure if we should be surfacing those. Unless this was a recent change I missed

I think we're good to hide both Keyboard Shortcuts and Custom Emoji on mobile webview.

I'm also wondering about System Console since it's not a mobile-friendly experience

michelengelen:
@matthewbirtch I have removed the mentioned entries from the menu in mobile view. Please re-review, when you have the time

michelengelen:
@matthewbirtch it was not ... but it is not obvious, since the condition for rendering the mobile menu is true, but the prop `show` was the opposite, so it never showed in mobile ... I am going to remove that 👍🏼

matthewbirtch (approved):
Thanks @michelengelen!

components/search_bar/search_bar.tsx
```diff
@@ -112,7 +111,7 @@ const SearchBar: React.FunctionComponent<Props> = (props: Props): JSX.Element =>
         >
             <form
                 role='application'
-                className={classNames(['search__form', {'search__form--focused': isFocused, 'search__form--global': props.globalHeaderEnabled}])}
+                className={classNames(['search__form', 'search__form--global', {'search__form--focused': isFocused}])}
```
hmhealey, line 114: Should the styles from `search__form--global` be rolled into `search__form` now since both classes are always applied at the same time?

components/search_bar/search_bar.tsx
```diff
@@ -149,7 +148,7 @@ const SearchBar: React.FunctionComponent<Props> = (props: Props): JSX.Element =>
                     ref={getSearch}
                     id={props.isSideBarRight ? 'sbrSearchBox' : 'searchBox'}
                     tabIndex='0'
-                    className={classNames('search-bar form-control a11y__region', {'Global__search-bar': props.globalHeaderEnabled})}
+                    className={'search-bar form-control a11y__region Global__search-bar'}
```
hmhealey, line 151: Same with `Global__search-bar` and `search-bar`

components/sidebar/channel_navigator/channel_navigator.tsx
```diff
@@ -165,7 +164,7 @@ export default class ChannelNavigator extends React.PureComponent<Props> {
         );
 
         let layout;
-        if (isDesktopApp() && !this.props.globalHeaderEnabled) {
+        if (isDesktopApp()) {
```
hmhealey, line 167: Shouldn't this whole block be removed if we're always turning on the global header now?

Edit: Yeah, it should be removed since we've got duplicated arrows now in the LHS.

components/threading/global_threads/global_threads.tsx
```diff
@@ -126,7 +123,7 @@ const GlobalThreads = () => {
                     id: 'globalThreads.subtitle',
                     defaultMessage: 'Threads you’re participating in will automatically show here',
                 })}
-                right={globalHeaderEnabled ? null : <RHSSearchNav/>}
+                right={null}
```
hmhealey, line 126: Do we have to explicitly pass null here, or can this prop be left out?

sass/base/_css_variables.scss
```diff
@@ -60,4 +60,10 @@
     --elevation-4: 0 8px 24px 0 rgba(0, 0, 0, 0.12);
     --elevation-5: 0 12px 32px 0 rgba(0, 0, 0, 0.12);
     --elevation-6: 0 20px 32px 0 rgba(0, 0, 0, 0.12);
+
+    // GLobal Header variables
```
hmhealey, line 64:
```suggestion
    // Global Header variables
```

sass/components/_search.scss
```diff
@@ -181,6 +186,65 @@
         right: 25px;
         font-size: 1em;
     }
+
+    // global header style adjustments
+    &.search__form--global {
+        background: var(--sidebar-text-08);
```
hmhealey, line 192: This could be left for a followup ticket since I assume they're just moving from elsewhere, but these should be using rgba to set opacity

hmhealey (changes requested):
It took me a moment to realize this was removing all the non-global header code from the web app. Would you mind changing the ticket title to that so that we can find it in the future when looking back through the git history?

michelengelen:
@hmhealey this is ready for re-review

hmhealey (approved):

michelengelen:
@nevyangelova and @jgilliam17 kind reminder to review! 🙇🏼‍♂️

nevyangelova (approved):

amyblais:
> @michelengelen Needs manual cherry-pick to `release-6.0` and `cloud`.

@michelengelen ^ In case you missed my previous ping.

michelengelen:
@amyblais we need to first cherry-pick the commit 03969d8e7bf1e7a1cd6ca063a93d05fd2b8294bc (PR for that: #8870 ), since it is using a function that has been introduced by that

amyblais:
> @amyblais we need to first cherry-pick the commit [03969d8](https://github.com/mattermost/mattermost-webapp/commit/03969d8e7bf1e7a1cd6ca063a93d05fd2b8294bc) (PR for that: #8870 ), since it is using a function that has been introduced by that

Is this PR needed for v6.0 or can it then be pushed to v6.1? @michelengelen @deanwhillier

michelengelen:
It can be pushed to v6.1 ... it is "only" a housekeeping PR that does not add/extend/alter features

PR #8956, opened 2021-09-24
MM-36932: always call reconnect callback

koox00:
#### Summary

Upon websocket reconnect we are missing messages, this commit attempts
to fix that by always calling reconnectCallback regardless on whether
reliableWebSockets is enabled or not.

Some more info:
We keep track of websocket connection status on redux,
which is getting set to false onerror, and set true upon reconnect,
with reliable websockets enabled it didn't got set to true upon reconnection,
since `reconnectCallback` was responsible for that.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36932

#### Release Note

```release-note
NONE
```

### DIFF
--- a/client/websocket_client.tsx
+++ b/client/websocket_client.tsx
@@ -86,7 +86,7 @@ export default class WebSocketClient {
 
             if (this.connectFailCount > 0) {
                 console.log('websocket re-established connection'); //eslint-disable-line no-console
-                if (!reliableWebSockets && this.reconnectCallback) {
+                if (this.reconnectCallback) {
                     this.reconnectCallback();
                 }
             } else if (this.firstConnectCallback) {


agnivade:
The code is actually correct; the callbacks are slightly different for reliable websockets case. What you are looking for is the `missedEventCallback`, not `reconnectCallback`. For reliable websocket scenario, there is no reconnect from the client perspective, because the server guarantees to send all messages across reconnects.

But in case the server restarts, or client is disconnected for a very long time, we emit a `missedEventCallback`. From the client side, both `missedEventCallback` and `reconnectCallback` do the same thing:

https://github.com/mattermost/mattermost-webapp/blob/2c47234fca223d4948b4e1dd4a6930f9d6efff5a/actions/websocket_actions.jsx#L153-L154

But it seems like in some cases the client depends on `reconnectCallback` and not `missedEventCallback`. I'd suggest to handle both cases.

If you want to have just a single callback, feel free to remove `missedEventCallback` altogether. Although I'm not sure if there are any other implications.

cc @hmhealey @devinbinnie (because you reviewed the original client side reliable websocket changes)

agnivade:
On looking at the code again, I realize the mistake now. It was subtle, and even after careful review, I missed it.

So for a temporary disconnect, a client can miss out on a sequence number. In that case, we just close the websocket, allowing the server to send from the last seen sequence_number.

https://github.com/mattermost/mattermost-webapp/blob/f9edd202fda12be18db82d13f33dcfd822876259/client/websocket_client.tsx#L177-L184

But what I missed was that, when the connection reopens, the reconnectCallback is bypassed obviously, but so will be the `missedEventCallback`, because the connectionId will still be the same.

Hence, always firing `reconnectCallback` is the right solution. Thanks @koox00 for catching this. We should backport this to `release-6.0`. cc @amyblais .

agnivade (approved):

devinbinnie (approved):

ogi-m (approved):
Tested and I can't repro the issue on the PR build - the messages are retrieved on websocket reconnect without the need to refresh.
LGTM 👍🏻

hmhealey (approved):

PR #8806, opened 2021-09-08
[MM-38283] - fixed `SUGGESTION.UNDEFINED` in quick-switch modal

michelengelen:
#### Summary
when using the quick switch modal with results that had different types of items in the results (open, public and archived channels and direct messages) it was showing `SUGGESTION.UNDEFINED` as a divider for the direct messages.

This fix involves setting a new property on the items for direct messages in the `SwitchChannelProvider`.

#### Ticket Link
[MM-38283](https://mattermost.atlassian.net/browse/MM-38283)

#### Screenshots
|  before  |  after  |
|-----|-----|
|Screenshot 2021-09-08 at 15 05 01|Screenshot 2021-09-08 at 15 04 17|

#### Release Note
```release-note
NONE
```

PR #8963, opened 2021-09-27
[MM-38850] - fixed sorting and channel divider translation

michelengelen:
#### Summary
fixes some bugs with the channel switcher modal introduced by #8806 :
- channels now get sorted correctly
- the divider for public channels is now translated correctly
- clicking a suggestion now takes the user to the correct channel

#### Ticket Link
[MM-38850](https://mattermost.atlassian.net/browse/MM-38850)

#### Release Note
```release-note
NONE
```


agnivade (commented):
Wondering if we can we add some unit tests to lock the behavior?

michelengelen:
@agnivade we definitely should, but for now we need to merge this asap ... I will create a follow-up ticket to add unit tests!

jwilander (approved):

jasonblais (changes requested):
I get JS errors when trying to use channel switcher. Attached after hitting CMD+K on Mac/Chrome

image

components/suggestion/suggestion_list.jsx
```diff
@@ -192,40 +191,39 @@ export default class SuggestionList extends React.PureComponent {
...
-            return itemA.type.localeCompare(itemB.type);
-        });
-
         let lastType;
-        for (let i = 0; i < sortedItems.length; i++) {
-            const item = sortedItems[i];
-            const term = this.props.terms[i];
-            const isSelection = term === this.props.selection;
+        for (let i = 0; i < items.length; i++) {
+            const item = handledItems[i];
```
AshishDhama, line 219: @michelengelen shouldn't this be 
```suggestion
            const item = items[i];
```

components/suggestion/suggestion_list.jsx
```diff
@@ -192,40 +191,39 @@ export default class SuggestionList extends React.PureComponent {
...
+        const handledItems = [];
+        if (items.length === 0) {
+            if (!renderNoResults) {
                 return null;
             }
-            items.push(this.renderNoResults());
+            handledItems.push(this.renderNoResults());
         }
 
-        const sortedItems = clonedItems.sort((itemA, itemB) => {
```
AshishDhama, line 209: Should we remove the sorting logic, Is it redundant because I can see couple of components either using or extending ``suggestion_list``, Not sure if they rely on this for ordering. Thanks

michelengelen:
I fixed the sorting and addressed the remarks. We do have a problem though, that boils down to the duplication of dividers ... we need to refactor (parts of) the switch channel provider, since it will receive data from the server late and fires the callback once before it receives the data.

That leads to data being handled incorrectly. I will try to provide a fix for that as well, but this should be addressed sooner rather than later.

Thoughts @hmhealey ?

michelengelen:
I did push a possible solution just now ... could someone from QA and PM have a look if everthing is fine with that? Thanks!

/cc @jgilliam17 @jasonblais @esethna

esethna:
@michelengelen I'm curious why the dividers even exist? I'm pretty sure they didn't exist since this channel switcher enhancement went in in v5.29: https://mattermost.atlassian.net/browse/MM-22854

image

michelengelen:
@esethna interesting ... thats something i was not aware of ... I my original ticket was to fix a missing translation and based on that we had some issues with sorting stuff. 

Question is: do we even care about the dividers? Shall I just remove them?

esethna:
@michelengelen we don't care about the dividers, but I'm more curious about what changed since v5.29 that caused this. There might be more that broke re sorting algorithms

michelengelen:
@esethna I will try to find out why they re-appeared

michelengelen:
OK, the change that brought dividers back is this one: https://github.com/mattermost/mattermost-webapp/commit/f92186e70ceb185bc055453f38d52d8c16dbfc20

esethna:
@michelengelen hmm okay. That PR should not have changed anything for how the channels display or sort AFTER a user types some characters. It should only have affected the initial state of the switcher to show recent channels when it opens.

Any changes that might have been made to the behaviour after the user types characters were likely unintentional.

AshishDhama:
@esethna Yeah if that pr is causing this was unintentional, As we didn’t touch the search logic but as per the requirement of initial state(Recents) have add divider which interestingly weren’t showing up when we types some characters.
Channel

michelengelen:
@esethna i think so as well ... i dont exactly know why this came up again, but we are now in this mess and likely need to either ... 
1. fix the issue with that or ...
2. close this PR and provide a better (potentially more in-depth) solution

Regardless of the way we are going forward on this we should plan for a refactor of the modal and especially the  SwitchChannelProvider, since those have grown considerably since they were originally created. 

A lot of the stuff that is going on in there is not perfect. That is something to be expected from a piece of code this old, but nonetheless this is a bit of a tech debt here we shoud address. Thoughts?

michelengelen:
FYI: I will close this PR and open a new one with only the changes needed to remove the results dividers. There has been so much stuff going on and with the information I have now it is better to not obfuscate stuff without reason. 

The new PR will reference this to keep track of the discussion.

PR #8967, opened 2021-09-27
MM-38747 Fixed crash caused by incorrect prop

hmhealey:
`utils.jsx` isn't typed, so `localizeMessage` returning the wrong type for that prop wasn't caught by the type checker.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38747

#### Release Note
```release-note
NONE
```

PR #8970, opened 2021-09-27
Add type definitions for utils.jsx

hmhealey:
https://github.com/mattermost/mattermost-webapp/pull/8967 was needed because TypeScript couldn't detect that the return type of `Utils.localizeMessage` was invalid where it was being used which reminded me that `utils.jsx` is massive and not in TypeScript. To save us some pain while also avoiding actually migrating it to TypeScript, I added a type definition file that gets us part way there.

Of course, adding type definitions did highlight a bunch of other places that needed changes. There's a few bugs fixed in there due to improper usage of certain functions like `Utils.localizeMessage`, but most of the other changes are to fix other type definitions that were otherwise incorrect. I separated fixes for the more interesting type issues into their own commits in case anyone is curious about those.

#### Release Note
```release-note
NONE
```


hahmadia (approved):
Nice changes Harrison! More precise type definitions are always nice. 👍

pablovelezvidal (approved):
LGTM!

PR #8974, opened 2021-09-28
[MM-35700] Accept drag and dropping on a thread view

Willyfrog:
#### Summary

Allow users to drag and drop files on a thread while on the CRT Panel

#### Ticket Link

[MM-35700](https://mattermost.atlassian.net/browse/MM-35700)

#### Release Note

```release-note
Fixed drag and drop files on a thread while on CRT Panel
```

### DIFF
--- a/components/create_comment/create_comment.tsx
+++ b/components/create_comment/create_comment.tsx
@@ -238,6 +238,7 @@ type Props = {
     channelMemberCountsByGroup: ChannelMemberCountsByGroup;
     onHeightChange?: (height: number, maxHeight: number) => void;
     focusOnMount?: boolean;
+    isThreadView?: boolean;
 }
 
 type State = {
@@ -1249,7 +1250,7 @@ class CreateComment extends React.PureComponent<Props, State> {
                     onUploadProgress={this.handleUploadProgress}
                     rootId={this.props.rootId}
                     channelId={this.props.channelId}
-                    postType='comment'
+                    postType={this.props.isThreadView ? 'thread' : 'comment'}
                 />
             );
         }
--- a/components/file_upload/file_upload.jsx
+++ b/components/file_upload/file_upload.jsx
@@ -180,6 +180,8 @@ export class FileUpload extends PureComponent {
             this.registerDragEvents('.row.main', '.center-file-overlay');
         } else if (this.props.postType === 'comment') {
             this.registerDragEvents('.post-right__container', '.right-file-overlay');
+        } else if (this.props.postType === 'thread') {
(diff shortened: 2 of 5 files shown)


koox00 (approved):

AshishDhama (approved):
No sure about drag-drop overlay taking the fullscreen. IMO overlay should appear only for RHS. Adding @matthewbirtch for UX review.
Rest LGTM!
Thanks for taking this.
Screenshot 2021-09-29 at 11 12 29 AM

matthewbirtch:
> No sure about drag-drop overlay taking the fullscreen. IMO overlay should appear only for RHS. Adding @matthewbirtch for UX review.

Agree, the overlay should only take up the thread itself - the same way we do in the RHS when dragging a file in to a thread viewed in the RHS.

For clarity, this is the area where the overlay should be:
image


Other than that, I think it's good. I'll super glad to get this in!

Willyfrog:
@matthewbirtch is it ok if the thread's title is not covered by the overlay?

Screen Shot 2021-09-30 at 10 52 57

matthewbirtch:
> @matthewbirtch is it ok if the thread's title is not covered by the overlay?

It's fine for now if that makes it easier, however that makes it inconsistent with how it works on the RHS
Screen Shot 2021-09-30 at 9 28 34 AM

Willyfrog:
I've been trying to do an easy fix, but it is built differently than the RHS so it is always the full grid or just the cell. So in order to do it properly we'd need to either change it from grid to something else for positioning or do a custom overlay for this case.

but I feel both cases are a bit too much for the gain. So I think it might make sense to leave it as part of the thread cell and have the title out. Let me know your thoughts on it, but if you agree we could send it to QA and merge.

matthewbirtch:
> I've been trying to do an easy fix, but it is built differently than the RHS so it is always the full grid or just the cell. So in order to do it properly we'd need to either change it from grid to something else for positioning or do a custom overlay for this case.
> 
> but I feel both cases are a bit too much for the gain. So I think it might make sense to leave it as part of the thread cell and have the title out. Let me know your thoughts on it, but if you agree we could send it to QA and merge.

Yep, all good with me

matthewbirtch (approved):

jgilliam17:
Thanks @Willyfrog 
Works great on Thread view. 
Seeing one issue with RHS - no longer shows overlay treatment when dragging and dropping a file.

Willyfrog:
@jgilliam17 should be fixed now

jgilliam17 (approved):
Thanks @Willyfrog 
Tested, looks good to merge.
- Verified Threads view accepts drag and drop files. RHS and main channel view work as before.

PR #8976, opened 2021-09-28
[MM-38850] - only render "recent" divider in quick switch modal

michelengelen:
#### Summary
due to some misunderstanding and a regression coming in by another PR the `SuggestionList` inside the quick switch modal was rendering type-dividers for the returned results.

I had a PR open for fixing it (#8963), but since I did not have all information on channel request/sorting logic within it and the number of commits that were already in that PR it was easier to provide a new PR for a fix.

#### Ticket Link
[MM-38850](https://mattermost.atlassian.net/browse/MM-38850)

#### Screenshots

**showing recent channels with divider**
Screenshot 2021-09-28 at 16 48 36

**after typing the results have no divider**
Screenshot 2021-09-28 at 16 48 44

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/suggestion/suggestion_list.jsx
+++ b/components/suggestion/suggestion_list.jsx
@@ -206,28 +206,18 @@ export default class SuggestionList extends React.PureComponent {
             items.push(this.renderNoResults());
         }
 
-        const sortedItems = clonedItems.sort((itemA, itemB) => {
-            if (!itemA.type) {
-                return 1;
-            }
-            if (!itemB.type) {
-                return -1;
-            }
-            return itemA.type.localeCompare(itemB.type);
-        });
-
-        let lastType;
-        for (let i = 0; i < sortedItems.length; i++) {
-            const item = sortedItems[i];
+        let dividerRendered = false;
+        for (let i = 0; i < this.props.items.length; i++) {
+            const item = this.props.items[i];
             const term = this.props.terms[i];
             const isSelection = term === this.props.selection;
 
             // ReactComponent names need to be upper case when used in JSX
             const Component = this.props.components[i];
 
-            if (this.props.renderDividers && item.type !== lastType) {
+            if (!dividerRendered && item.type === 'mention.recent.channels') {
                 items.push(this.renderDivider(item.type));
-                lastType = item.type;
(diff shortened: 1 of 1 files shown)


esethna (approved):
Approving based on convo with Michel

hmhealey (approved):
Could we prioritize UX/QA review on this to get it onto daily? The quick switcher is currently unusable for me because it seems to be dredging up a bunch of archived channels as well that mess with the search results

michelengelen:
@jgilliam17 @matthewbirtch @nevyangelova @deanwhillier please prioritize this PR to get it into daily asap! Thanks! 🙇🏼‍♂️ 

/cc @hmhealey

deanwhillier (approved):
Not sure there is/should be anything to do about this, but I noticed that when results take a bit longer to return, the no-results graphic flashes briefly. Wonder if there might be a way to slightly delay showing no-results to help prevent the flash?

https://user-images.githubusercontent.com/3245614/135282679-e316fdf3-2ae1-4b8b-a444-5b27b0f24b8a.mp4

nevyangelova (approved):

matthewbirtch (approved):
Seems to work as expected

michelengelen:
> > Not sure there is/should be anything to do about this, but I noticed that when results take a bit longer to return, the no-results graphic flashes briefly. Wonder if there might be a way to slightly delay showing no-results to help prevent the flash?
> > NoResultsFlash.mp4
> 
> It might be doable depending on the order of how we render the no result indicator. I had similar issues with rhs search and the no results page.

IMO this is out of scope for the current PR. I will create a follow-up ticket to handle that.

michelengelen:
follow-up ticket: https://mattermost.atlassian.net/browse/MM-38928

esethna:
@jgilliam17 please prioritize a review here if possible, this is a pretty bad productivity blocker on daily right now!

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge.
- Verified only Recent divider renders on the channel switcher modal. Search on the modal and navigation to selected channel works as before.

PR #8978, opened 2021-09-28
MM-38539 Make rehydrateDrafts idempotent

hmhealey:
Since we've worked out that the performance issues we're seeing around having multiple tabs open are storage-related, this is one attempt at fixing those by making it so that even if we do trigger multiple loads of the drafts (which Joram has a PR to fix), we'll only ever update the storage state when something actually changed. Now, we'll only dispatch actions for items that have changed, if any have changed.

I wish that there was a less fancy word for "making a function produce the same result when run repeatedly" than idempotent.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38539

#### Release Note
```release-note
Improve performance of draft loading
```


actions/storage.js
```diff
@@ -31,12 +34,9 @@ export function removeItem(name) {
...
 export function setGlobalItem(name, value) {
-    return (dispatch) => {
-        dispatch({
-            type: StorageTypes.SET_GLOBAL_ITEM,
-            data: {name, value, timestamp: new Date()},
-        });
-        return {data: true};
+    return {
+        type: StorageTypes.SET_GLOBAL_ITEM,
+        data: {name, value, timestamp: new Date()},
```
jwilander, line 39: 🎉

jwilander (approved):

actions/storage.js
```diff
@@ -84,22 +84,31 @@ export function actionOnItemsWithPrefix(prefix, action) {
...
+            const existing = getGlobalItem(state, key);
+            if (existing && areObjectsEqual(existing, parsed)) {
+                continue;
+            }
+
+            actions.push(setGlobalItem(key, parsed));
+        }
+
+        if (actions.length === 0) {
+            return {data: false};
```
lieut-data, line 108: nit: We never used to do this, even if there were no entries. Are there any potentially negative side effects?

lieut-data (approved):
Looks good: just one question.

actions/storage.js
hmhealey: No, there's no changes from this since we never await on this dispatch. Having to return `{data: 'anything'}` is an artifact of our messed up typing that we really need to fix at some point.

actions/storage.js
hmhealey: I wanted to do this to the rest of the action creators too, but I didn't feel like making more changes to this file at the moment 😛

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified performance for users with 20+ drafts on server with multiple teams - no issues.

PR #8980, opened 2021-09-28
MM-38539 Only rehydrate drafts from localStorage on first load

jwilander:
#### Summary
Companion PR to #8978. Draft rehydration was occurring on any change to localforage when it was only required on first load. This had a side effect for multiple tabs on the same desktop app or browser causing a lot of unnecessary hydration resulting in high CPU usage.

@jgilliam17 for testing can you make sure drafts still persist correctly after refresh and channel switch? Particularly when typing quickly and immediately reloading the page.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38539

#### Release Note
```release-note
Performance improvements for draft storage with multiple tabs open
```

### DIFF
--- a/actions/storage.js
+++ b/actions/storage.js
@@ -83,7 +83,7 @@ export function actionOnItemsWithPrefix(prefix, action) {
 }
 
 // Temporary action to manually rehydrate drafts from localStorage.
-function rehydrateDrafts() {
+export function rehydrateDrafts() {
     return (dispatch) => {
         const actions = [];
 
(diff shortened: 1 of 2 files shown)


lieut-data (approved):
Love the "temporary" comment: probably been there for years :D

jwilander:
That one's actually only a couple months old and we /might/ get to actually fixing it next quarter :)

hmhealey (approved):

hmhealey:
Nothing's more permanent than a temporary fix 😛

jgilliam17 (approved):
Thanks @jwilander 
Tested, looks good to merge.
- Verified current drafts are unaffected after refresh and channel switch.

PR #8989, opened 2021-09-29
MM-T1805 No infinite loading spinner on Select Team page

poflankov:
#### Summary
Check 'No infinite loading spinner on Select Team page'

#### Ticket Link

https://github.com/mattermost/mattermost-server/issues/18178

#### Release Note

```release-note
NONE
```

### DIFF
--- a/e2e/cypress/integration/multi_team_and_dm/multi_team_spec.js
+++ b/e2e/cypress/integration/multi_team_and_dm/multi_team_spec.js
@@ -12,7 +12,7 @@ import * as TIMEOUTS from '../../fixtures/timeouts';
 // Stage: @prod
 // Group: @multi_team_and_dm
 
-describe('Send a DM', () => {
+describe('Multi-Team + DMs', () => {
     let userA; // Member of team A and B
     let userB; // Member of team A and B
     let userC; // Member of team A
@@ -52,9 +52,6 @@ describe('Send a DM', () => {
     beforeEach(() => {
         // # Log in to Team A with an account that has joined multiple teams.
         cy.apiLogin(userA);
-
-        // # On an account on two teams, view Team A
-        cy.visit(offTopicUrlA);
     });
 
     it('MM-T433 Switch teams', () => {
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/multi_team_and_dm/multi_team_spec.js
```diff
@@ -147,4 +144,9 @@ describe('Send a DM', () => {
             cy.get('.badge').should('not.exist');
         });
     });
+
+    it('MM-T1805 No infinite loading spinner on Select Team page', () => {
+        cy.visit('/select_team');
```
saturninoabril, line 149: Should be fine to go directly but could you please satisfy the first 2 steps? Like create a team where the test user could join via UI at select team page, then visit again such page to check no more team to join and that the loading indicator does not exist. 
```
1. Create an account on a server with very few teams
2. Join all available teams teams
```

saturninoabril (changes requested):
Thank you for your contribution @poflankov! Please see comment.

e2e/cypress/integration/multi_team_and_dm/multi_team_spec.js
poflankov: Ah, I thought those steps are the 'arrangement' phase for this test, so should be done by using API.
Should I then move out my test from this suite? Because in the suite that arrangement of user and teams is done in `before`.

saturninoabril: We typically do test preparation directly via API whenever it's not the main focus of test, but prefer UI interaction especially when nothing yet exist to exercise such flow. Yeah, I agree to move into its own spec. Thanks @poflankov!

poflankov: @saturninoabril Thank you too for elaborate answer!
So, to make it clear, should I perform both steps 1 and 2 using UI?
And how should I name the new spec file?

saturninoabril: Step 1 via API then step 2 and the rest via UI. Name it related to test description (short, few words only :) )

PR #8991, opened 2021-09-29
MM-38294 Show browse channels with global header disabled and A/B tests enabled

hmhealey:
I don't know if this is only something we'll hit on community-daily, but we seem to have hit an unfortunate combination of feature flags and A/B tests there that make it so that the dropdown is missing

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38294

#### Screenshots
Without the global header:
Screen Shot 2021-09-29 at 10 45 45 AM

With the global header:
Screen Shot 2021-09-29 at 10 44 55 AM

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/sidebar/channel_navigator/channel_navigator.test.tsx
+++ b/components/sidebar/channel_navigator/channel_navigator.test.tsx
@@ -56,4 +56,11 @@ describe('Components/ChannelNavigator', () => {
         const wrapper = shallow(<ChannelNavigator {...props}/>);
         expect(wrapper.find(AddChannelDropdown).length).toBe(0);
     });
+
+    it('should show AddChannelDropdown when there is an active A/B treatment and the global header is disabled', () => {
+        props.addChannelButton = AddChannelButtonTreatments.BY_TEAM_NAME;
+        props.globalHeaderEnabled = false;
+        const wrapper = shallow(<ChannelNavigator {...props}/>);
+        expect(wrapper.find(AddChannelDropdown).length).toBe(1);
+    });
 });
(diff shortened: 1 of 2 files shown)


michelengelen (approved):

neallred:
I think this is a combination of feature flags we decided we won't be using anywhere: https://community-daily.mattermost.com/core/pl/gpzbhcrzcidopjxcs91hjghigw . /cc @marianunez

marianunez:
> I think this is a combination of feature flags we decided we won't be using anywhere: [community-daily.mattermost.com/core/pl/gpzbhcrzcidopjxcs91hjghigw](https://community-daily.mattermost.com/core/pl/gpzbhcrzcidopjxcs91hjghigw) and we already disabled that A/B test in the prod environment: [community-daily.mattermost.com/private-core/pl/4ngbxn6p1p8q78qhmtf1wg4r1o](https://community-daily.mattermost.com/private-core/pl/4ngbxn6p1p8q78qhmtf1wg4r1o). /cc @marianunez

@hmhealey the feature flag was enabled properly in community now so you should see the button again. Given that global header is no longer optional and is being removed in https://github.com/mattermost/mattermost-webapp/pull/8940 this change should not be required given that `this.props.globalHeaderEnabled` would no longer exist.

hmhealey:
That's fine then, but we should try to make sure that the default case for those still works going forward. Maybe I'm just unlucky for having a bunch of tickets that involved joining channels, but I feel like I have hit this quite a few times in the past month.

PR #8993, opened 2021-09-29
MM-38134 - change first purchase success screen texts

pablovelezvidal:
#### Summary
This PR makes sure that after the initial purchase (when purchasing in free trial), the success page message makes reference to an initial purchase instead of an upgrade as it has been doing so far.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38134

#### Release Note

```release-note
NONE
```


annelieseklein:
LGTM, I would just center the laptop illustration on the page (not the laptop + card) @pablovelezvidal

PR #8996, opened 2021-09-30
[MM-38868] Fixes do not disturb submenu item not taking full width

AshishDhama:
#### Summary
Do not disturb submenu item not taking full width
(Note: we need to enable TimeDND feature flag)
Screenshot 2021-09-30 at 2 49 38 PM

#### Ticket Link
Fixes [https://mattermost.atlassian.net/browse/MM-38868](https://mattermost.atlassian.net/browse/MM-38868)

#### Related Pull Requests
NONE

#### Screenshots
|  before  |  after  |
|----|----|
| image-20210927-235911 | Screenshot 2021-09-30 at 2 47 44 PM |

#### Release Note
```release-note
* UI: Fixes Do not disturb submenu item not taking full width
```

### DIFF
--- a/components/status_dropdown/status_dropdown.scss
+++ b/components/status_dropdown/status_dropdown.scss
@@ -42,7 +42,13 @@
 
 #status-menu-dnd-timed_menuitem {
     .SubMenuItem.MenuItem {
-        width: 100%;
+        flex-grow: 1;
+    }
+    .SubMenuItemContainer{
+        &:hover,
+        &:focus {
+            background-color: unset;
+        }
     }
 }
 
--- a/components/status_dropdown/status_dropdown.scss
+++ b/components/status_dropdown/status_dropdown.scss
@@ -40,6 +40,12 @@
     }
 }
 
+#status-menu-dnd-timed_menuitem {
+    .SubMenuItem.MenuItem {
+        width: 100%;
+    }
+}
+
 .custom_status__row {
     position: relative;
     height: auto;
--- a/components/status_dropdown/status_dropdown.scss
+++ b/components/status_dropdown/status_dropdown.scss
@@ -44,7 +44,8 @@
     .SubMenuItem.MenuItem {
         flex-grow: 1;
     }
-    .SubMenuItemContainer{
+
+    .SubMenuItemContainer {
         &:hover,
         &:focus {
             background-color: unset;


components/status_dropdown/status_dropdown.scss
```diff
@@ -40,6 +40,12 @@
     }
 }
 
+#status-menu-dnd-timed_menuitem {
+    .SubMenuItem.MenuItem {
+        width: 100%;
```
larkox, line 45: Is this how we do it on the rest of the elements? (in this submenu or in the menu before?)

I wonder if we are using in other places flex properties, or something like that.
My point is just to be coherent among the different elements / files.

larkox (approved):
LGTM, just a comment regarding potential coherence.

components/status_dropdown/status_dropdown.scss
larkox: After a second look, I think the problem is not so much about taking 100%, but applying two shades at the same time. Maybe the submenu item has already a background and we are applying an extra background we shouldn't?

Willyfrog (approved):
other than @larkox concerns, LGTM

components/status_dropdown/status_dropdown.scss
larkox: I can see a difference in the shades between the element selected in the previous menu, and the one selected in this menu. Not sure if desired. Maybe ask someone from UX?

AshishDhama: @larkox problem wasn't colouring, when we were clicking on the item nothing was happening. What you are suggesting also make sense.

AshishDhama:
@larkox Please take a look now, have fixed the hover background color issue and instead of width have used flex-grow.
@nevyangelova Could you please go through this as well. If we can remove background from ``SubMenuItemContainer``.

components/status_dropdown/status_dropdown.tsx
```diff
@@ -328,6 +328,7 @@ export default class StatusDropdown extends React.PureComponent <Props, State> {
                     action: index === 4 ? () => setCustomTimedDnd() : () => setDnd(index),
                 } as any;
             }));
+        console.log('dndSubMenuItems', dndSubMenuItems);
```
larkox, line 331: Are both changes on this file for debugging?

larkox (changes requested):
LGTM, just remove the (apparently) debugging code.

Willyfrog (approved):

components/status_dropdown/status_dropdown.tsx
AshishDhama: Thanks, Done

larkox (approved):

nevyangelova (approved):

PR #9012, opened 2021-10-01
MM-38908: Handle /:team/_playbooks/:playbookId/run

agarciamontoro:
#### Summary
Navigating to `/:team/_playbooks/:playbookId/run` will switch to the last viewed channel in the team, defaulting to Town Square, and then will execute the `/playbook run-playbook` command, effectively rendering an interactive dialog to run the playbook.

This is needed to support the Run button in the Playbooks product, fixing a bug in the desktop where the app successfully changes the tab to Channels but still renders the interactive dialog in the Playbooks tab.

The video below shows the use-case: the user is in team `eligendi`, then goes to the Playbooks tab and opens a Playbook in team `minus`, when clicking on the Run button, the desktop app focuses the Channels tab, where the user is redirected to the last channel viewed in the playbook team (in this case `neque` in the team `minus`) and opens the interactive dialog. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38908

#### Related Pull Requests
- Playbooks PR: https://github.com/mattermost/mattermost-plugin-playbooks/pull/780

#### Screenshots
https://user-images.githubusercontent.com/3924815/135624556-f38d661e-e7a1-41a7-bea6-930e78227689.mp4

#### Release Note

```release-note

```

### DIFF
--- a/components/channel_layout/center_channel/center_channel.tsx
+++ b/components/channel_layout/center_channel/center_channel.tsx
@@ -11,6 +11,7 @@ import LoadingScreen from 'components/loading_screen';
 import PermalinkView from 'components/permalink_view';
 import ChannelHeaderMobile from 'components/channel_header_mobile';
 import ChannelIdentifierRouter from 'components/channel_layout/channel_identifier_router';
+import PlaybookRunner from 'components/channel_layout/playbook_runner';
 import {makeAsyncComponent} from 'components/async_load';
 const LazyGlobalThreads = makeAsyncComponent(
     React.lazy(() => import('components/threading/global_threads')),
@@ -115,6 +116,11 @@ export default class CenterChannel extends React.PureComponent<Props, State> {
                             path='/:team/:path(channels|messages)/:identifier/:postid?'
                             component={ChannelIdentifierRouter}
                         />
+                        <Route
+                            path='/:team/_playbooks/:playbookId/run'
+                        >
+                            <PlaybookRunner/>
+                        </Route>
                         {isCollapsedThreadsEnabled ? (
                             <Route
                                 path='/:team/threads/:threadIdentifier?'
--- a/components/channel_layout/playbook_runner.tsx
+++ b/components/channel_layout/playbook_runner.tsx
@@ -1,0 +1,73 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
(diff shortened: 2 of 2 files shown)


lieut-data (approved):
Brilliant!

lieut-data:
Looping @devinbinnie for review from the Desktop's perspective.

agarciamontoro:
Hm, I had also added @calebroseland as reviewer :thinking: Let's go with both, you never have too many reviews in a PR :)

devinbinnie (approved):
LGTM! Nice job guys!

calebroseland (approved):
I love the redirect solution! 

I'm not going to hold this up, but putting this in the `center_channel` flow doesn't quite feel right to me.

Perhaps `<PlaybookRunner/>` could be mounted in route handler/container that is provided by Playbooks via `PluginRegistry.registerCustomRoute` (listening for something like`/plug/playbooks/:playbookId/run`)

https://github.com/mattermost/mattermost-webapp/blob/master/components/root/root.jsx#L403-L414

agarciamontoro:
> Perhaps `<PlaybookRunner/>` could mounted in route handler/container that is provided by Playbooks via `PluginRegistry.registerCustomRoute` (listening for something like`/plug/playbooks/:playbookId/run`)

Wouldn't this keep the bug as it is, with the interactive dialog spawning in the Playbooks tab? Maybe there is a better place to put it, but we need something *in* the Channels product, not in the URLs managed by the plugin.

calebroseland:
> > Perhaps `<PlaybookRunner/>` could mounted in route handler/container that is provided by Playbooks via `PluginRegistry.registerCustomRoute` (listening for something like`/plug/playbooks/:playbookId/run`)
> 
> Wouldn't this keep the bug as it is, with the interactive dialog spawning in the Playbooks tab? Maybe there is a better place to put it, but we need something _in_ the Channels product, not in the URLs managed by the plugin.

5/5 You're right. Please disregard my comments above.

agarciamontoro:
@amyblais, we need to get this into 6.0. Is there anything else I need to do apart from the cherry-pick?

amyblais:
@agarciamontoro Just a cherry-pick to release-6.0.

agarciamontoro:
Skipping QA, I reviewed the fix myself and this will be tested as part of Playbooks in our regular bug bash parties :)

PR #9021, opened 2021-10-03
[MM-38451] move loadProfilesMissingStatus and remove statuses prop

joshalling:
#### Summary

This moves some logic to fetch missing statuses into an action so that we can remove the statuses prop from the MoreDirectChannels component.

#### Ticket Link

https://github.com/mattermost/mattermost-server/issues/18485

#### Related Pull Requests

#### Release Note

```release-note

```

### DIFF
--- a/actions/status_actions.jsx
+++ b/actions/status_actions.jsx
@@ -89,6 +89,25 @@ export function loadStatusesByIds(userIds) {
     };
 }
 
+export function loadProfilesMissingStatus(users) {
+    return (dispatch, getState) => {
+        const state = getState();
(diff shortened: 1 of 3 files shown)


devinbinnie (approved):
LGTM! Thanks @joshalling!

hmhealey (changes requested):
@joshalling These changes look perfect! This is exactly what I was hoping for.

It looks like the unit tests for the MoreDirectChannels component are broken though, probably because the new function isn't being passed in by them. Would you be able to fix those?

PR #9024, opened 2021-10-03
[MM-34334] Migrate 'components/intl_provider' to Typescript

cinlloc:
#### Summary
Migrate 'components/intl_provider' and associated tests to TypeScript

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/17269

#### Release Note
NONE


cinlloc:
@mm-cloud-bot /release-note-none

components/intl_provider/intl_provider.tsx
```diff
@@ -1,27 +1,30 @@
...
-        children: PropTypes.element.isRequired,
-        locale: PropTypes.string.isRequired,
-        translations: PropTypes.object,
-        actions: PropTypes.shape({
-            loadTranslations: PropTypes.func.isRequired,
-        }).isRequired,
+import {ActionFunc} from 'mattermost-redux/types/actions';
+
+type Props = {
+    children: ReactElementLike;
```
devinbinnie, line 19: Just curious, does `React.ReactNode` work here instead? I'd rather not rely on `prop-types` if we don't have to.
```suggestion
    children: React.ReactNode;
```

components/intl_provider/intl_provider.tsx
```diff
@@ -1,27 +1,30 @@
...
+    locale: string;
+    translations: Record<string, string> | Record<string, MessageFormatElement[]> | null;
```
devinbinnie, line 21: I don't think we explicitly check for `null`, can we instead make this optional? (ie. `undefined` instead of `null`)
```suggestion
    translations?: Record<string, string> | Record<string, MessageFormatElement[]>;
```

devinbinnie (approved):
Thanks @cinlloc, this is looking good :)
I have a couple non-blocking comments but overall looks good.

components/intl_provider/intl_provider.tsx
cinlloc: In associated test, we have scenarii that use explicitly `translations` with `null` values, that's why I kept `null`. I could replace tests with `null` by tests with `undefined` (tests pass), if you are sure it's ok.

components/intl_provider/intl_provider.tsx
cinlloc: Good catch, fixed in 344a76a

components/intl_provider/intl_provider.tsx
cinlloc: I gave a try in commit 344a76a, please tell me if you feel it convenient

devinbinnie (approved):

agarciamontoro (approved):
Looks great, @cinlloc, thank you so much for the contribution!

@devinbinnie, do we need QA here?

devinbinnie:
> Looks great, @cinlloc, thank you so much for the contribution!
> 
> @devinbinnie, do we need QA here?

Yeah let's assign QA, they can defer if they want to.

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @cinlloc

PR #9026, opened 2021-10-04
MM-39001 - highlight invite members lhs button  when less 10 users

pablovelezvidal:
#### Summary
This PR adds an extra validation to the condition if highlight the invite members button to only highlight in the scenario where the workspace has less than 10 users

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39001

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/sidebar/invite_members_button.tsx
+++ b/components/sidebar/invite_members_button.tsx
@@ -1,13 +1,17 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React from 'react';
+import React, {useEffect} from 'react';
 
 import {useIntl, FormattedMessage} from 'react-intl';
 
-import {useSelector} from 'react-redux';
+import {useDispatch, useSelector} from 'react-redux';
 
 import {Permissions} from 'mattermost-redux/constants';
+import {DispatchFunc} from 'mattermost-redux/types/actions';
+import {getStandardAnalytics} from 'mattermost-redux/actions/admin';
+
+import {GlobalState} from 'types/store';
 
 import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
 
@@ -25,11 +29,22 @@ type Props = {
 
 const InviteMembersButton: React.FC<Props> = (props: Props): JSX.Element => {
     const intl = useIntl();
-
+    const dispatch = useDispatch<DispatchFunc>();
     const currentTeamId = useSelector(getCurrentTeamId);
+    const analytics = useSelector((state: GlobalState) => state.entities.admin.analytics);
+    const userLimit = 10;
+
+    useEffect(() => {
+        if (!analytics) {
+            (async function getAllAnalytics() {
+                await dispatch(getStandardAnalytics());
+            }());
+        }
+    }, []);
+
     let buttonClass = 'SidebarChannelNavigator_inviteMembersLhsButton';
 
-    if (!props.touchedInviteMembersButton) {
+    if (!props.touchedInviteMembersButton && Number(analytics?.TOTAL_USERS) <= userLimit) {
         buttonClass += ' SidebarChannelNavigator_inviteMembersLhsButton--untouched';
     }
 


pablovelezvidal:
> This works for admins, but non-admins don't have access to this endpoint: I get a `403` when I try to read this data in the web UI as a non-admin user.
> 
> Endpoint: `api/v4/analytics/old?name=standard&team_id=` route definition: `https://github.com/mattermost/mattermost-server/blob/master/api4/system.go#L52` handler definition: https://github.com/mattermost/mattermost-server/blob/master/api4/system.go#L378-L405
> 
> I'm guessing there's data in that payload we don't want everyone able to access. If we do want to continue with this check, maybe we can add a server endpoint that merely returns a boolean of whether or not we should assume this is a new installation?

Good catch, let me see if there is another place where we can get how many users already exists in the current workspace.

esethna (approved):
Approving assuming this works for admins and end user accounts

(ie if >= 10 users on the server, the invite member button should not be bolded in the sidebar)

stevemudie (approved):
Worked as expected.
Added 12 users then refreshed and link was no longer highlighted.

I also deactivated enough to get below 10 users but the link never highlighted again though I'm not sure if that's within the scope of the ticket.

pablovelezvidal:
> Looks like that gets us the data we need! I still see an issue that when I reload the app, there's no initial call to the users/stats api. I think we need to trigger it like you were conditionally triggering the admin call as a `useEffect`.
> 
> I can see the issue locally if I: Spin up a server with less than 10 users. Log in with a user that has not clicked the invite members button. Observer no highlight is present Do something to trigger the call like open the direct messages messages modal (the one you get by typing ctrl + shift + k). `/api/v4/users/stats` is made. Observe that data is now fetched, and highlight appears.

Thank you @neallred . I am Going to do the validation in the component itself as I was doing before for the admin analytics. Will let you know when ready for review.

AGMETEOR:
> This works for admins, but non-admins don't have access to this endpoint: I get a `403` when I try to read this data in the web UI as a non-admin user.
> 
> Endpoint: `api/v4/analytics/old?name=standard&team_id=` route definition: `https://github.com/mattermost/mattermost-server/blob/master/api4/system.go#L52` handler definition: https://github.com/mattermost/mattermost-server/blob/master/api4/system.go#L378-L405
> 
> I'm guessing there's data in that payload we don't want everyone able to access. If we do want to continue with this check, maybe we can add a server endpoint that merely returns a boolean of whether or not we should assume this is a new installation?

Good observation. I asked here https://community.mattermost.com/core/pl/jr6frc7zct81fq11ehtdwzxfnh
Hopefully someone can give us an existing way to avoid having to write a new endpoint

components/sidebar/sidebar_category/index.ts
```diff
@@ -25,11 +26,13 @@ function makeMapStateToProps() {
     const getChannelIdsForCategory = makeGetFilteredChannelIdsForCategory();
 
     return (state: GlobalState, ownProps: OwnProps) => {
+        const stats = getTotalUsersStatsSelector(state) || {total_users_count: 0};
         return {
             channelIds: getChannelIdsForCategory(state, ownProps.category),
             draggingState: getDraggingState(state),
             touchedInviteMembersButton: getBool(state, Preferences.TOUCHED, Touched.INVITE_MEMBERS),
             currentUserId: getCurrentUserId(state),
+            totalUserCount: stats.total_users_count,
```
neallred, line 35: Container components ftw! :smiley:

components/sidebar/sidebar_category/sidebar_category.tsx
```diff
@@ -76,6 +78,9 @@ export default class SidebarCategory extends React.PureComponent<Props, State> {
             this.categoryTitleRef.current.addEventListener(A11yCustomEventTypes.ACTIVATE, this.handleA11yActivateEvent);
             this.categoryTitleRef.current.addEventListener(A11yCustomEventTypes.DEACTIVATE, this.handleA11yDeactivateEvent);
         }
+        if (!this.props.totalUserCount) {
+            this.props.actions.getTotalUsersStats();
```
hmhealey, line 82: I hate to say it, but this isn't going to work here since we're going to be loading the total user count multiple times when the sidebar is mounted because each SidebarCategory is going to call this. I think the InviteMembersButton should be the one responsible for this.

To fix the test errors you were hitting before, couldn't you just mock out the action?

hmhealey (changes requested):

components/sidebar/sidebar_category/sidebar_category.tsx
pablovelezvidal: shouldn't this ` if (!this.props.totalUserCount) {` make sure that it is called only once, and the next time it will grab the prop from the redux-store?

pablovelezvidal:
@hmhealey the tests are fixed now, it seems that the problem was actually the call to `store.getState()`. Thanks!

PR #9029, opened 2021-10-04
MM-39033 Revert change to More DMs modal for new users

hmhealey:
I accidentally made it so that new users would see an empty DM modal instead of a list of all users alphabetically before they had any DMs opened. This fixes that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39033

#### Screenshots
Before:
Screen Shot 2021-10-04 at 10 24 32 AM
After:
Screen Shot 2021-10-04 at 10 46 47 AM

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/more_direct_channels/list/index.test.js
+++ b/components/more_direct_channels/list/index.test.js
@@ -532,7 +532,7 @@ describe('makeGetOptions', () => {
         ]);
     });
 
-    test('should return users without DMs as long as a search term is being used', () => {
+    test('should return users without DMs as long as either there are no recents or a search term is being used', () => {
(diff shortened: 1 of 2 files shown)


devinbinnie (approved):

calebroseland (approved):
Sorry I didn't catch this in https://github.com/mattermost/mattermost-webapp/pull/8821!

hmhealey:
@calebroseland Not at all a problem. I think Jelena noticed it, but I didn't realize it was a change from before. The old logic was really complicated, so I'm not surprised we missed a case.

PR #9039, opened 2021-10-04
MM-39046 Don't modify all drafts on page load

hmhealey:
I think this logic is also related to the performance issues we've been fighting. Every time a `CreatePost` or `CreateComment` is mounted, we go through and clear up all in-progress file uploads (which is another bug for the RHS since it means you have to leave the RHS open while a file uploads or else it'll be cleared), and when we do that, we mutate all the drafts. That, in turn, caused all the other tabs to rehydrate which caused other performance-affecting stuff to happen.

Now, we'll only mutate them if there were actually changes to be made.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39046

#### Release Note
```release-note
Reduce storage-related slow downs on page load
```

### DIFF
--- a/actions/storage.js
+++ b/actions/storage.js
@@ -62,24 +62,9 @@ export function clear(options = {exclude: []}) {
 }
 
 export function actionOnGlobalItemsWithPrefix(prefix, action) {
-    return (dispatch) => {
-        dispatch({
-            type: StorageTypes.ACTION_ON_GLOBAL_ITEMS_WITH_PREFIX,
-            data: {prefix, action},
-        });
-        return {data: true};
-    };
-}
-
-export function actionOnItemsWithPrefix(prefix, action) {
-    return (dispatch, getState) => {
-        const state = getState();
-        const globalPrefix = getPrefix(state);
-        dispatch({
-            type: StorageTypes.ACTION_ON_ITEMS_WITH_PREFIX,
-            data: {globalPrefix, prefix, action},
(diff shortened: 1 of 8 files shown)


actions/storage.js
```diff
@@ -62,24 +62,9 @@ export function clear(options = {exclude: []}) {
...
-    return (dispatch) => {
-        dispatch({
-            type: StorageTypes.ACTION_ON_GLOBAL_ITEMS_WITH_PREFIX,
-            data: {prefix, action},
-        });
-        return {data: true};
-    };
-}
-
-export function actionOnItemsWithPrefix(prefix, action) {
```
hmhealey, line 74: This action wasn't actually used anywhere, so I removed it rather than having to also fix it. I'm not really sure what the difference between "items" and "global items" is anyway

actions/views/create_comment.tsx
```diff
@@ -34,11 +34,12 @@ import {GlobalState} from 'types/store';
 import {DispatchFunc, GetStateFunc} from 'mattermost-redux/types/actions';
 
 export function clearCommentDraftUploads() {
```
hmhealey, line 36: I might move this around later to better consolidate the draft logic in one place, but I don't want to do that before the Global Drafts view PR is in

larkox (approved):
LGTM

jwilander (approved):

PR #9040, opened 2021-10-04
[MM-34339] Migrate 'components/quick_input' to TypeScript

cinlloc:
#### Summary
Migrate 'components/quick_input' and associated tests to TypeScript

#### Note
Didn't manage to avoid using `any` type in some parts. Would be happy to have some suggestions about how to do better. 

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/17264

#### Release Note
```release-note
NONE
```


components/quick_input/quick_input.tsx
```diff
@@ -11,65 +11,75 @@ import OverlayTrigger from 'components/overlay_trigger';
...
+    /**
+     * Whether to delay updating the value of the textbox from props. Should only be used
+     * on textboxes that to properly compose CJK characters as the user types.
+     */
+    delayInputUpdate?: boolean;
+
+    /**
+     * An optional React component that will be used instead of an HTML input when rendering
+     */
+    inputComponent?: ReactComponentLike;
```
devinbinnie, line 25: Same comment here and below wrt. `React.ReactNode` vs. `prop-types`

components/quick_input/quick_input.tsx
```diff
@@ -11,65 +11,75 @@ import OverlayTrigger from 'components/overlay_trigger';
...
+    /**
+     * The optional tooltip text to display on the X shown when clearable. Pass a components
+     * such as FormattedMessage to localize.
+     */
+    clearableTooltipText?: string | ReactElementLike;
+
+    /**
+     * Callback to clear the input value, and used in tandem with the clearable prop above.
+     */
+    onClear?: () => any;
```
devinbinnie, line 47: If `onClear` doesn't return anything, the return should be `void`
```suggestion
    onClear?: () => void;
```

components/quick_input/quick_input.tsx
```diff
@@ -11,65 +11,75 @@ import OverlayTrigger from 'components/overlay_trigger';
...
-        onChange: PropTypes.func,
-
-        /**
-         * When true, and an onClear callback is defined, show an X on the input field even if
-         * the input is empty.
-         */
-        clearableWithoutValue: PropTypes.bool,
-    };
+export default class QuickInput extends React.PureComponent<Props> {
+    private input: any;
```
devinbinnie, line 82: `input` should be of type `HTMLInputElement`
```suggestion
    private input: HTMLInputElement;
```

components/quick_input/quick_input.tsx
```diff
@@ -116,11 +126,11 @@ export default class QuickInput extends React.PureComponent {
         return this.input;
     };
 
-    setInput = (input) => {
+    setInput = (input: HTMLInputElement) => {
         this.input = input;
     }
 
-    onClear = (e) => {
+    onClear = (e: any) => {
```
devinbinnie, line 133: `e` should be of type `React.MouseEvent<HTMLDivElement>`
```suggestion
    onClear = (e: React.MouseEvent<HTMLDivElement>) => {
```

devinbinnie (changes requested):
Thanks for the PR @cinlloc! I've added some comments for your review.

components/quick_input/quick_input.tsx
cinlloc: Unfortunately, I tried to use ReactNode or other classes under React, but it seems that React.createElement line 180 does not accept it. Any idea?

components/quick_input/quick_input.tsx
cinlloc: Yes, it was my first try but I had 2 issues with that : it has to be undefined or defined in constructor (but with which value?) and there is a file that performs a cast to a `AutosizeTextarea`, so i simplified to any. Please tell me if b29e36ef80f1 is on the good path

components/quick_input/quick_input.tsx
cinlloc: You're right, fixed in b29e36ef80f1

components/quick_input/quick_input.tsx
cinlloc: Obviously! Fixed in b29e36ef80f1

components/quick_input/quick_input.tsx
joshchoo: Would `React.ElementType` be a good solution?

devinbinnie: Ahh yeah it's not something we're rendering directly to the DOM. Maybe try `React.ComponentClass` or `React.Component`?

cinlloc: I tried a lot of different combinations as suggested, `React.ElementType`, `React.Component`, `React.ComponentClass`... The latest does almost the job but fails line 183/184 because of `ref` and `default`, and in some other files because `inputComponent` is assigned to other types (cf. `marketplace_modal.tsx:201`, `edit_category_modal.tsx:133`). I'm out of solution to replace this legacy type, my level in Typescript reaches its limit, sorry!

devinbinnie: All good, let's stick with what you have then :) thanks for trying!

components/quick_input/quick_input.tsx
devinbinnie: Yeah I think that works for now.

devinbinnie (approved):
LGTM! Thanks @cinlloc!

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @cinlloc

PR #9047, opened 2021-10-05
MM-39073: fixes threads unread tab

koox00:
#### Summary

Clicking on an unread thread sometimes made the thread disappear.
This happened because the `getThread` selector was looking in both all
threads and in unread threads to verify the thread exists in this team,
but if you landed at unread threads tab after a refresh you don't have
more than 5 threads in all threads reducer, and if you click a thread it
was getting removed from the unread reducer.

This commit fixes that by refactoring the `getThread` selector to use
user's current team channels to check if threads belong in a current
team's channel.

There is still a way to have that buggy behavior even after this fix.
The scenario is if you have left a channel you don't get the threads
removed as of yet, so you could replicate the above buggy behavior in
that case.
Since we are going to fix this in a later task (threads will be removed
from "left" channels) I didn't bother to fix it as well.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39073

#### Release Note

```release-note
NONE
```

### DIFF
--- a/packages/mattermost-redux/src/selectors/entities/threads.ts
+++ b/packages/mattermost-redux/src/selectors/entities/threads.ts
@@ -3,6 +3,7 @@
 
 import {createSelector} from 'reselect';
 
+import {getMyChannels} from 'mattermost-redux/selectors/entities/channels';
 import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
 import {GlobalState} from 'mattermost-redux/types/store';
 import {Team} from 'mattermost-redux/types/teams';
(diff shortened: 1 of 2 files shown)


calebroseland (approved):
Great change!

packages/mattermost-redux/src/selectors/entities/threads.ts
```diff
@@ -66,16 +67,24 @@ export function getThreads(state: GlobalState): IDMappedObjects<UserThread> {
...
 }
 
-export function getThread(state: GlobalState, threadId: $ID<UserThread> | undefined): UserThread | null {
-    if (
-        !threadId ||
-        !(getThreadsInCurrentTeam(state)?.includes(threadId) || getUnreadThreadsInCurrentTeam(state)?.includes(threadId))
-    ) {
-        return null;
-    }
+export const getThread: (state: GlobalState, threadId: $ID<UserThread> | undefined) => UserThread | null = createSelector(
```
hmhealey, line 70: Since we're memoizing this now, `getThread` should be turned into a selector factory called `makeGetThread`.

Alternatively, we'd probably be fine to just remove the `createSelector` entirely since none of this computation is particularly complex, and it's not returning a newly created object.

packages/mattermost-redux/src/selectors/entities/threads.ts
```diff
@@ -66,16 +67,24 @@ export function getThreads(state: GlobalState): IDMappedObjects<UserThread> {
...
+    'getThread',
+    getThreads,
+    (state: GlobalState) => getMyChannels(state).map((c) => c.id),
```
hmhealey, line 73: This would also need its own memoization since it's returning a new array of channel IDs each time this is called, even if the state hasn't changed.

hmhealey (changes requested):

packages/mattermost-redux/src/selectors/entities/threads.ts
koox00: you're right, removed memoization to minimize the changes.

PR #9052, opened 2021-10-05
[MM-38289] - Add product branding tests

nevyangelova:
#### Summary
Adds a simple unit and snapshot test for product branding

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38289

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/global/test_helpers.tsx
+++ b/components/global/test_helpers.tsx
@@ -1,0 +1,19 @@
+import {ProductComponent} from 'types/store/plugins';
+
+type TestProductComponent = Omit<ProductComponent, 'switcherIcon'> & {
+    switcherIcon: string;
+}
+
+export function makeProduct(name: string): TestProductComponent {
+    return {
+        id: name,
+        pluginId: '',
+        switcherIcon: `product-${name.toLowerCase()}`,
+        switcherText: name,
+        baseURL: '',
+        switcherLinkURL: '',
+        mainComponent: null,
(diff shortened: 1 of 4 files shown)


components/global/test_helpers.tsx
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {ProductComponent} from 'types/store/plugins';
+
+type TestProductComponent = Omit<ProductComponent, 'switcherIcon'> & {
+    switcherIcon: string;
+}
+
+export function makeProduct(name: string): TestProductComponent {
```
hmhealey, line 10: If you think this might be used elsewhere, this could be put in `utils/test_helper` instead. We could wait until we actually need to use this elsewhere though.

components/global/test_helpers.tsx
```diff
@@ -0,0 +1,22 @@
...
+    return {
+        id: name,
+        pluginId: '',
+        switcherIcon: `product-${name.toLowerCase()}`,
```
hmhealey, line 14: Instead of having `makeProduct` return a new type and then using `as ProductComponent` everywhere else, what if we just change this line to
```suggestion
        switcherIcon: `product-${name.toLowerCase()}` as ProductComponent['switcherIcon'],
```
to fake that this value is the correct type here instead?

PR #9086, opened 2021-10-07
Update in-product string for v6.0

cwarnermm:
#### Summary
Updates an existing next-steps dialog in-product text string for v6.0 Global Header element details.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-38748

#### Release Note
```release-note
NONE
```


michelengelen:
@cwarnermm ideally you would open a new JIRA ticket and link it with this PR ... but since it is just a rather small change we can just do this without a ticket. objections @esethna ?

esethna:
Works for me @cwarnermm @michelengelen

PR #9088, opened 2021-10-07
[MM-34336] Convert MainMenu component to TypeScript

iOSGeekster:
#### Summary
Migrates 'components/main_menu' and tests from `jsx` to `tsx`

#### Ticket Link
Jira: https://mattermost.atlassian.net/browse/MM-34336
https://github.com/mattermost/mattermost-server/issues/17267

#### Comments
I'm still fairly new to TypeScript, so I'm also taking this as a learning opportunity.

I'm a bit unsure whether there is an existing type we could use for the `pluginMenuItems`
But please fell free to provide any kind of feedback!

```release-note
NONE
```


iOSGeekster:
> Thanks for the PR! Unfortunately, you'll also need to resolve some conflicts since we just removed a bunch of code from those files. Hopefully that makes fixing the other issues easier 🙂

Thanks! I'll do that

> For your questions:
> 
> 1. If you change the type of `state` from the `GlobalState` defined in `mattermost-redux` to the one defined in `types/store`, that should give you the type for `pluginMenuItems`. The one in `types/store` extends the other one to add some more stuff that's specific to the web app since the other one is used elsewhere.

Yup, just find out, makes sense. And thanks for the clarification

> 2. The package-lock changes are because you're using an older version of NPM prior to NPM 7. If you don't want to upgrade NPM right now, everything should still work if you just remove the changes to the package-lock from your PR

Ah great, thanks! Yeah, I'm still some-what new to the whole `npm` thingy. 
Your comments are much appreciated!

iOSGeekster:
> @iOSGeekster Can you resolve conflicts and update? Thanks 🙂

Sure thing! It is updated now 👍

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @iOSGeekster

PR #9104, opened 2021-10-08
MM-38893 CRT: Modify Channel label for DMs since name shows twice

jlram:
#### Summary

In the thread list, hides the name of the person you are messaging with in case they started the thread.
A `.Badge_hidden` class has been added to the `<Badge>` component to switch its display mode if the other user has started the thread.
Also, I've grouped the `post.props?.override_username || displayName` expression into `const postAuthor`in order to make it more readable.
#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-38893

#### Screenshots
|  before  |  after  |
|----|----|
| 3fd797c0-794c-440c-be98-6647a5f3510c | Captura de pantalla 2021-10-08 a las 19 03 47 |

#### Release Note

```release-note
Removed badge in thread list when the thread is started by another user in a DM
```

### DIFF
--- a/components/threading/global_threads/thread_item/__snapshots__/thread_item.test.tsx.snap
+++ b/components/threading/global_threads/thread_item/__snapshots__/thread_item.test.tsx.snap
@@ -11,6 +11,7 @@ exports[`components/threading/global_threads/thread_item should report total num
       Someone
     </span>
     <Memo(Badge)
+      className=""
       onClick={[Function]}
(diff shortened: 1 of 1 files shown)


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @jlram

PR #9122, opened 2021-10-12
[MM-35920, GH-18609]Add arrow up/arrow down navigation support for Collapsed Reply Threads

cleferman:
#### Summary

This commit aims to increase productivity by enabling ease of navigation
through threads via arrow keys.

Changes from current behaviour:
-  the input box is not automatically focused anymore, property
focusOnMount is now set to false
-  the input box now gets focused when the user has a thread selected
and starts typing
-  hitting the ESCAPE key while the input box is focused now blurs it

Modified focusTextboxIfNecessary function to match the create_post
component. Left a comment to start a discussion on that.

Update create_comment.test to test if the input blurs on pressing ESCAPE

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/18609

#### Related Pull Requests

#### Screenshots

mattermost-gif

#### Release Note

```release-note
Added support for navigating through Collapsed Reply Threads via arrow keys.
Added support for focusing the input box in CRT while typing.
Added support for blurring the input box in CRT on pressing Escape.
```

### DIFF
--- a/components/logged_in/index.ts
+++ b/components/logged_in/index.ts
@@ -2,7 +2,13 @@
 // See LICENSE.txt for license information.
 
 import {connect} from 'react-redux';
-import {bindActionCreators} from 'redux';
+import {bindActionCreators, Dispatch} from 'redux';
+
+import {GlobalState} from 'types/store';
+
+import {Channel} from 'mattermost-redux/types/channels';
+
+import {DispatchFunc, GenericAction} from 'mattermost-redux/types/actions';
 
 import {autoUpdateTimezone} from 'mattermost-redux/actions/timezone';
 import {getCurrentChannelId} from 'mattermost-redux/selectors/entities/channels';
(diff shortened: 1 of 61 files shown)


cleferman:
@koox00 Thanks for the info! I was wondering how the files are formatted because auto formatting in VSCode changed a whole lotta stuff. Works now with editorconfig!

cleferman:
I've fixed the scrolling and the list no longer wraps when pressing the UP/DOWN arrow keys as @matthewbirtch suggested.
I had to leave some eslint disable lines and a ts-ignore. Please let me know if I can solve this differently. Didn't find a way.

There a two things I noticed:
- Selecting a thread with the mouse applies a focus on the item (a shade of gray), moving away from that thread with the arrows will leave the shade on the item until you focus something else, e.g: typing something will write in the comment box and the shade disappears. Not sure where it comes from, didn't spend time on it. Let me know if I should. Screenshot below: 
image

- If you have one of the first 5 threads selected  (top -> bottom), reloading the page will 'remember' your selection. If you select any other thread than those first 5 it will not. This happens because the first call loads only 5 threads and for some reason if the selected thread is not among them, the URL route gets changed:

| Before | After Refresh|
| -------| -------|
| image | image |

And also, thanks for the feedback everyone. I'm happy to hear that it will make such a big difference productivity wise :).

calebroseland:
@cleferman I think instead of passing the id+callback and doing the effect in the row, hoisting the effect to `VirtualizedThreadList` should allow it to work all the time.

[arrow-keys-effect.patch.zip](https://github.com/mattermost/mattermost-webapp/files/7392248/arrow-keys-effect.patch.zip)

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @cleferman

PR #9128, opened 2021-10-12
[MM-37887] Fix permalink not highlighting post in thread when rhs was opened

cleferman:
#### Summary

This commit fixes the highlighting of a post when a permalink is clicked
and the RHS window is already opened.

To clarify, the bug would reproduce for any permalinked post when
certain post was clicked more than one time consecutively.
If the user clicked two different permalinks that were out of view in
RHS, the highlighting would work.

Added an extra condition for triggering the highlight using
`selectedPostFocusedAt` prop as used also in create_comment.tsx.

A second issue that was causing the scroll to bottom was the
`focusOnMount` prop. I had to reset the state of `userScrolledToBottom`
to `false` right before `scrollToItem` function is called.
#### Ticket Link

Jira https://mattermost.atlassian.net/browse/MM-37887
Fixes https://github.com/mattermost/mattermost-server/issues/18573
#### Related Pull Requests

#### Screenshots

|  before  |  after  |
|----|----|
| permalink-fix-before | permalink-fix-after |
#### Release Note

```release-note
Fixed a bug where clicking a permalink would not scroll the highlighted post into view inside the RHS panel.
```


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @cleferman

PR #9133, opened 2021-10-13
[MM-38990] - Show updated next steps tips to all sysadmins

ghost:
#### Summary
This PR follows up https://mattermost.atlassian.net/browse/MM-37159 and shows the updated illustration of next steps tips to all sysadmins. A question for QA, some context was missing since I didn't implement the first PR, are there any use cases where there is a difference in what we show to first sysadmin versus all other sysadmins? Did I add the condition correctly in all the necessary places? 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38990

#### Release Note

```release-note
NONE
```


jgilliam17:
@nevyangelova 
> are there any use cases where there is a difference in what we show to first sysadmin versus all other sysadmins?

As far as Next steps tips are concerned, my understanding is that as long as any system admin is allowed to modify everything, they should be presented with the same tips screen

PR #9150, opened 2021-10-14
auto show tutorial tips given the a/b test treatment

neallred:
#### Summary
Add treatment where users are autoshown the tutorial tour points on page load and between finishing tips. If approved, this acould be cherry-picked to the cloud release so that we can start collecting data on the A/B test sooner.

feature flag values: `none` , `auto`
Current feature flag value: `auto`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38941

#### Related Pull Requests
- Has server changes https://github.com/mattermost/mattermost-server/pull/18695

#### Screenshots

https://user-images.githubusercontent.com/13738432/137401375-db87fbad-4f13-416f-b9d2-dc332ba989f4.mp4

#### Release Note

```release-note
Add treatment where users are autoshown the tutorial tour points on page load and between finishing tips.
```


esethna:
cc// @annelieseklein can you help do a review of the FTUE to see if there's other stuff we should address? ^^^

marianunez:
> Edit: Dean on the UI team was able to find a way to do this, he's going to submit an update to this PR :)

@esethna can we treat this improvement outside of this PR and A/B Test? This is increasing the scope of the changes that we need to be able to launch the test for next cloud release and we are already tight on the

neallred:
@esethna with the update it is a lot more doable. I think I could implement it, combined with Dean's changes, by EOD today or mid-day tomorrow. @marianunez does that work for the cherry-pick & release schedule?

marianunez:
> I think this is necessary to run this test - especially given the component to knock out now exists. Remember that users seeing this have never seen the Mattermost UI and we are forcing them through the tour points with the entire UI darkened. The knock outs are the only way users can get situated and understand the UI we are trying to teach them

Completely agree that is a necessary improvements but given that the component to knock out is still an active PR which requires it's own QA and reviews, there is a risk this test won't make it to the cloud release next week. 

In either case, let's merge this PR that has the code review complete and tackle the punch out in other one so it helps keeps these changes separately. Once this is released we can make a decision on whether we want to activate the A/B tests or if we need to wait for the punch out in a future release.

neallred:
Merging.

> Button text will be changed as per comment above

Was changed here https://github.com/mattermost/mattermost-webapp/pull/9150/commits/a02416fe86da1b2d420ac95faeb78b18d72c8372#diff-096d7f8dbc4f08b6237a49f4df18c4b316067bcd4f12b331cd9db480eaf3386fR197

> Punch out will be cherry picked to cloud before this test launches

Opting to implement Dean's punch out in a separate ticket and cherry-pick to cloud before the test launches.

PR #9162, opened 2021-10-16
MM-35175 Don't throw error if user didn't allow notifications

vblz:
#### Summary
Currently if user doesn't allow showing notification in browser, every message still calls `Notifications.showNotification()` method. In the method it checks whether user allowed notifications and if not it throws `Error`. These errors left uncaught and each message goes into server logs and leave there a log entry with `Error` level.
On populated servers it floods server log and makes log's monitoring think there are a lot false-positive incidents.

Since blocking notification is not actually an error, the PR removes throwing errors and does nothing if there is no permission to show notification.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35175
https://github.com/mattermost/mattermost-server/issues/17496

#### Related Pull Requests

None

#### Screenshots

No changes in UI

#### Release Note

```release-note
NONE
```

### DIFF
--- a/utils/notifications.tsx
+++ b/utils/notifications.tsx
@@ -11,7 +11,8 @@ let requestedNotificationPermission = false;
 // showNotification displays a platform notification with the configured parameters.
 //
 // If successful in showing a notification, it resolves with a callback to manually close the
-// notification. Notifications that do not require interaction will be closed automatically after
+// notification. If no error occurred but the user did not grant permission to show notifications, it
+// resolves with a no-op callback. Notifications that do not require interaction will be closed automatically after
 // the Constants.DEFAULT_NOTIFICATION_DURATION. Not all platforms support all features, and may
 // choose different semantics for the notifications.
 


lieut-data:
Thanks for thinking through this deeply, @vblz! I agree with your conclusions on which exceptions to leave and which to modify (as currently done in this PR).

With respect to the documentation, the only suggestion I might have is to clarify that the callback doesn't necessarily do anything, since it currently reads that getting a callback returned is a sign that the notification was shown successfully. Maybe something like:

```
// showNotification displays a platform notification with the configured parameters.
//
// If successful in showing a notification, it resolves with a callback to manually close the
// notification. If no error occurred but the user did not grant permission to show notifications, it 
// resolves with a no-op callback. Notifications that do not require interaction will be closed automatically after
// the Constants.DEFAULT_NOTIFICATION_DURATION. Not all platforms support all features, and may
// choose different semantics for the notifications.
```

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @vblz

PR #9180, opened 2021-10-18
[MM-38941] Add component for tooltip overlay background

deanwhillier:
#### Summary
Per a conversation with @esethna, this PR adds a component to allow for basic punch-out's in the tutorial tip overlay background for highlighting areas of the UI. This is mostly a temporary measure until the tutorial tip common component is developed.

There are two usage examples included, one for the account menu tip in the Global header and one for the channel tip in channel sidebar. The first uses pre-determined values and the second calculates values from an existing DOM element.

This PR is made against the actual PR for the feature work for MM-38941, where the full review will take place so we shouldn't need to go through the full review here.

This PR also adds a fix for pushing the Global Header on top of the channels view so the pulsing dot doesn't get cut off.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/global_header/global_header.tsx
+++ b/components/global_header/global_header.tsx
@@ -42,7 +42,7 @@ const GlobalHeader = (): JSX.Element | null => {
     }
 
     return (
-        <GlobalHeaderContainer className='GlobalHeader'>
+        <GlobalHeaderContainer>
             <LeftControls/>
             <CenterControls productId={currentProductID}/>
             <RightControls productId={currentProductID}/>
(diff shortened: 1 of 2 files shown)


deanwhillier:
Note that this component doesn't update when the window is resized. The values provided to it would need to be recalculated if the window size changes and the resulting punch-out needs to change as well. Happy to help talk though how this could be done if needed.

esethna:
Closing as the changes are brought in here: https://github.com/mattermost/mattermost-webapp/pull/9218#issuecomment-948595455

PR #9188, opened 2021-10-19
[MM-38643] - Improve Onboarding steps consistency

AGMETEOR:
#### Summary
This adds a little bit of more flexibility in terms of detecting which is the last step in the onboarding steps. A calculation is made based on the index of the step in the steps array and then a prop `isLastStep` injected into each step component. It is up to the implementation of the step component to use this prop accordingly.

#### Ticket Link
[MM-38643](https://mattermost.atlassian.net/browse/MM-38643)

#### Screenshots

https://user-images.githubusercontent.com/28563179/137899770-96f5acd6-2c08-49d6-b6fe-7d034d96d8eb.mov

#### Release Note

```release-note
NONE
```


annelieseklein:
@AGMETEOR @marianunez can't see the getting started on the test server - it starts after the getting started steps. Is it possible to restart? but from the description I read above (of determining which step has the finish button), it sounds ok

AGMETEOR:
> > > @annelieseklein It's because of the improvement this PR adds. Only the last step that the user completes should show the finish button. This means all steps need to have some unique text on their "complete" buttons
> > 
> > 
> > @AGMETEOR I see the value in only the last step having the finish button, but I'm not sure it necessitates unique text -- maybe I'm missing the value.
> 
> @annelieseklein i can change back the default button text for those 2 steps to “Finish”. We can always revisit later. cc: @marianunez

@annelieseklein sorry for bringing this back. The issue is that users can complete these steps in any order, so if a user goes to the last step first and clicks on "Finish", they would expect to not be taken to the other steps. This is the point @stevemudie was raising. I have attached a screenshot below to demonstrate the issue and why every step should have a unique text on it's complete button.

https://user-images.githubusercontent.com/28563179/139525032-985d4168-216d-4de3-ae3d-7dd96265facb.mov

annelieseklein (approved):

AGMETEOR:
Hello @marianunez . Gentle ping on this

marianunez:
> @annelieseklein sorry for bringing this back. The issue is that users can complete these steps in any order, so if a user goes to the last step first and clicks on "Finish", they would expect to not be taken to the other steps. This is the point @stevemudie was raising. I have attached a screenshot below to demonstrate the issue and why every step should have a unique text on it's complete button.

I agree with @annelieseklein that having "Finish [step]" is wordy and not needed for each step. 
As you implemented it here, each step should have a text for regular text (as they used to have) and for when it's the last step which should be `Finish`. If it's the last step in the list but others are unresolved it probably makes sense to NOT say `Finish` because it's because it's misleading just the regular text. In the case of `Downlaoad App` step, could we change the regular text to `Completed`?

AGMETEOR:
> > @annelieseklein sorry for bringing this back. The issue is that users can complete these steps in any order, so if a user goes to the last step first and clicks on "Finish", they would expect to not be taken to the other steps. This is the point @stevemudie was raising. I have attached a screenshot below to demonstrate the issue and why every step should have a unique text on it's complete button.
> 
> I agree with @annelieseklein that having "Finish [step]" is wordy and not needed for each step. As you implemented it here, each step should have a text for regular text (as they used to have) and for when it's the last step which should be `Finish`. If it's the last step in the list but others are unresolved it probably makes sense to NOT say `Finish` because it's because it's misleading just the regular text. In the case of `Downlaoad App` step, could we change the regular text to `Completed`?

Yes that's true. For the invite step, it's regular text was "Next Step". So we need to find an appropriate text for it as well. Which text should I put? For the Download apps step I am going to change the button text to `Completed` as suggested.

AGMETEOR:
@stevemudie gentle ping on this

AGMETEOR:
@stevemudie gentle ping on this

stevemudie (approved):
@AGMETEOR this looks and works great!

stylianosrigas:
@AGMETEOR there was a bug in the mattermost server that was causing some panics. The image was based on that and its failing to create the spinwick. I am creating a new image ;)

AGMETEOR:
> @AGMETEOR there was a bug in the mattermost server that was causing some panics. The image was based on that and its failing to create the spinwick. I am creating a new image ;)

Thanks @stylianosrigas .

marianunez (approved):

PR #9190, opened 2021-10-19
[MM-38584]: fixing offline indicator color

michelengelen:
#### Summary
fix for a wrong indicator color in the global header avatar and account menu.
It was using an inherited color for that, but since the other avatar implementations (e.g. in the channels view) still use the "old" avatar it was out of line with them.

I added the `badges.offline` property to the interim theme-adjustments introduced by @deanwhillier so that it now consumes the correct color value. 

We do have a certain inconsistency introduced now as well, that might need some input from UX/UI: In themes where the channel view has a light theme and the surrounding elements (global header and sidebar) are dark the center channel color does not work on the dark background. this can be observed in the [screenshots section](#screenshots)

#### Ticket Link
[MM-38584](https://mattermost.atlassian.net/browse/MM-38584)

#### Screenshots
|  theme  |  in channel view  |  in global header  |
|----|----|----|
|  light  |Screenshot 2021-10-19 at 15 18 04|Screenshot 2021-10-19 at 15 18 23|
|  dark  |Screenshot 2021-10-19 at 15 17 49|Screenshot 2021-10-19 at 15 18 34|

#### Release Note
```release-note
changed offline indicator color to use correct theme color
```

### DIFF
--- a/components/compass_theme_provider/compass_theme_provider.tsx
+++ b/components/compass_theme_provider/compass_theme_provider.tsx
@@ -45,7 +45,6 @@ const CompassThemeProvider = ({theme, children}: Props): JSX.Element | null => {
                 online: theme.onlineIndicator,
                 away: theme.awayIndicator,
                 dnd: theme.dndIndicator,
-                offline: theme.centerChannelColor,
             },
             text: {
                 ...compassTheme.text,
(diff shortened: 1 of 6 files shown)


matthewbirtch:
@michelengelen unfortunately, this doesn't work very well. It's very stark in the channel context and not very visible at all in the global header. This is  I believe that's why a middle-grey color (not part of the theme) was chosen as a hard-coded color in the past.

PR #9191, opened 2021-10-19
[MM-39426]: removing div with sidebar--right__bg class since it has no effect

michelengelen:
#### Summary
fixes a bug where the sidebar could be scrolled sideways, since the `div` with the class `sidebar--right__bg` had a width of `1000%` set to it.

I searched for ocurences and possible use-cases, but could not find any. It seems that this part was used before, but is now a obsolete.

Strangely enough the `div` was set to `visibility: hidden` in web view and to `display: none` in mobile view, so it could not have any effect anyways, since it was never really rendered. :D

#### Ticket Link
[MM-39426](https://mattermost.atlassian.net/browse/MM-39426)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/sidebar_right/sidebar_right.jsx
+++ b/components/sidebar_right/sidebar_right.jsx
@@ -216,10 +216,6 @@ export default class SidebarRight extends React.PureComponent {
                 role='complementary'
                 ref={this.sidebarRight}
             >
-                <div
-                    onClick={this.onShrink}
-                    className='sidebar--right__bg'
-                />
                 <div className='sidebar-right-container'>
                     <Search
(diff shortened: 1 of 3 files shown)


deanwhillier:
@michelengelen, throw the RHS into expanded mode and take a look at that div.

image

Right now on the Community server, the `sidebar--right--expanded` class is added to the parent container (`sidebar--right `) when the RHS is expanded and that makes the `sidebar--right__bg` visible, blocking interactions with the UI underneath it.

michelengelen:
@deanwhillier @matthewbirtch I am closing this issue, since I cannot reproduce the reported issue (other than manually setting `overflow: auto` or `overflow: scroll` to the parent container.

I will post a note on the JIRA issue as well. 

Thanks for your patience!

PR #9220, opened 2021-10-20
MM-32783: header & footer for threads in search

pascalhein:
#### Summary
Display "Thread in $channel" instead of the channel
name above a search result that belongs to a
thread (any message, root or child).

Also show the footer (same as for regular posts),
but only for thread roots.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/18782
JIRA: https://mattermost.atlassian.net/browse/MM-32783

#### Screenshots
image

#### Release Note
```release-note
Added thread replies to search results (when collapsible reply threads are enabled)
```


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @pascalhein

jgilliam17:
Thanks @pascalhein 
This looks great!
Question for UX: @matthewbirtch please let me know your thoughts on the following. Thanks 🙂 
1. This is also added to pinned and saved rhs. Saved and Pinned search results are not specifically mentioned in the ticket, please let me know if this is expected. 
Screen Shot 2021-11-19 at 12 46 56 PM

2. Direct Message header:
- 0/5 I personally expected singular `message` , instead of `messages` in  "Thread in Direct Messages(with user-1)" 
- We should also remove the parenthesis around user name

Screen Shot 2021-11-19 at 1 17 06 PM

3. We are now showing 2 reply arrows on a root post on the RHS, please note they do behave differently and may cause confusion. 
Screen Shot 2021-11-19 at 2 00 26 PM

matthewbirtch:
> This is also added to pinned and saved rhs. Saved and Pinned search results are not specifically mentioned in the ticket, please let me know if this is expected.

@jgilliam17 yes, this is expected

> * 0/5 I personally expected singular `message` , instead of `messages` in  "Thread in Direct Messages(with user-1)"
> * We should also remove the parenthesis around user name

Agree with both of these

> We are now showing 2 reply arrows on a root post on the RHS, please note they do behave differently and may cause confusion.

This is a good catch @jgilliam17. I agree this is confusing and we should fix that. if CRT is one, we should hide the second reply arrow on the right. However, we should keep the behavior from that arrow and apply it to the thread footer because it gives users a way back to the search results/mentions/saved list.

image

I'm not sure if the above is clear, so let me know. Really good catches @jgilliam17 - sorry I missed these in my review.

i18n/en.json
```diff
@@ -4017,6 +4017,8 @@
   "search_item.file_badge.direct_message": "Direct Message",
   "search_item.file_badge.group_message": "Group Message",
   "search_item.jump": "Jump",
+  "search_item.thread": "Thread in {channel}",
+  "search_item.thread_direct": "Thread in Direct Messages (with {username})",
```
jgilliam17, line 4021:
```suggestion
  "search_item.thread_direct": "Thread in Direct Message with {username}",
```

components/search_results_item/search_results_item.jsx
```diff
@@ -199,10 +202,33 @@ export default class SearchResultsItem extends React.PureComponent {
...
 
-        if (channelType === Constants.DM_CHANNEL) {
+        const isDirectMessage = channelType === Constants.DM_CHANNEL;
+        const isPartOfThread = isCollapsedThreadsEnabled && (post.reply_count > 0 || post.is_following);
+
+        if (isDirectMessage && isPartOfThread) {
+            channelName = (
+                <FormattedMessage
+                    id='search_item.thread_direct'
+                    defaultMessage='Thread in Direct Messages (with {username})'
```
jgilliam17, line 215:
```suggestion
                    defaultMessage='Thread in Direct Message with {username}'
```

jgilliam17 (changes requested):
Thanks @matthewbirtch 
@pascalhein let me know if you have any questions about the 3rd point above - remove post menu reply option on thread root post on the RHS and apply it's functionality to thread footer arrow. 
I added a suggestion for the DM header text change. Thanks again 🙂

i18n/en.json
pascalhein: Done

components/search_results_item/search_results_item.jsx
pascalhein: Done

pascalhein:
Hi @jgilliam17 and @matthewbirtch,
thanks for your feedback!

I've uploaded a new version to address your concerns. Please let me know if you have any comments.

matthewbirtch (approved):

jgilliam17:
Thanks @pascalhein 
Manually tested, looks good.
Running [E2Es here](https://automation-dashboard.vercel.app/cycles/96afadda-aea9-41c4-9a01-339d8db8725e), will approve after report is posted.

jgilliam17 (approved):
Thank you @pascalhein 
Tested, looks good to merge.
- Verified "Thread in" displays as part of the title for search results items on the RHS. Resulting posts now also include the threads footer. Beside search, this functionality extends to mentions, pinned and saved posts.

PR #9254, opened 2021-10-25
MM-36765: fixes crash on LEAVE_CHANNEL

koox00:
#### Summary

Getting a thread item checks upon 'myChannels' and returns a thread only
if it belongs to a channel user is a part of.
Getting a list of threads though returned threads from all channels,
even channels the user has left before.
This behavior was causing a crash, since we couldn't get the single
thread.

This commit fixes that by removing the guard on `getThread` selector.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36765

#### Release Note

```release-note
NONE
```


koox00:
@jgilliam17 can you test both leaving, or getting removed from a channel?

packages/mattermost-redux/src/selectors/entities/threads.ts
```diff
@@ -68,19 +67,12 @@ export function getThreads(state: GlobalState): IDMappedObjects<UserThread> {
...
     if (!threadId) {
         return null;
     }
 
-    const thread = threads[threadId];
-    if (!thread || (thread.post && myChannels.indexOf(thread.post.channel_id) === -1)) {
-        return null;
-    }
-
-    return thread;
```
koox00, line 83: @calebroseland I'll be removing the guard on threads.
 I know we discussed this in the past (https://github.com/mattermost/mattermost-webapp/pull/8793#discussion_r709256045) but this guard creates more problems than it solves for us. 

Do you maybe have a way to reproduce the case of a thread from another team is going to appear in another team's global threads?

jgilliam17:
Thanks @koox00 
Seeing few issues 
- There is an error when user is removed from public or private channel 
After user is removed from a public channel, the thread they participated in is still available on the Global Threads, opening the thread shows "This channel is read-only..."

Clicking on the public channel label rejoins the channel and shows some additional errors, for private it also warns that you do not have channel access 

Screen Shot 2021-11-03 at 2 14 56 PM
Screen Shot 2021-11-03 at 2 18 08 PM

- When user is being removed from the private channel, threads remain on the Global threads, but in an empty state and without a channel label. Noting the UI in images, but I think here we just need to hide them, not fix the UI
Screen Shot 2021-11-03 at 2 24 53 PM

Screen Shot 2021-11-03 at 2 23 19 PM

- When leaving any channel, thread disappears from the global threads 
When re-joining or being added to channel, old threads are not immediately available on global threads, they populate on refresh.

koox00:
@jgilliam17 can we create a separate ticket for those, so to fix the crash and then fix those as well?

jgilliam17:
Yes, I can create additional tickets to fix above issues.

jgilliam17 (approved):
Thanks @koox00 
- Verified app is no longer crashing when opening Global Threads after user is removed/leaves any channel, specifically channel that contains threads they participated in.
Remaining Issues noted in my previous comment will be fixed in the follow up ticket https://mattermost.atlassian.net/browse/MM-39861

PR #9269, opened 2021-10-27
[MM-34309] migrate to typescript components/admin_console/jobs entire folder

ArcaneDiver:
#### Summary
This PR migrate to typescript the folder: `components/admin_console/jobs`.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/17241

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/admin_console/data_retention_settings/data_retention_settings.tsx
+++ b/components/admin_console/data_retention_settings/data_retention_settings.tsx
@@ -3,6 +3,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+/* eslint-disable react/jsx-closing-bracket-location */
 /* eslint-disable max-lines */
 
 import React, {createRef, RefObject} from 'react';
@@ -608,10 +609,10 @@ export default class DataRetentionSettings extends React.PureComponent<Props, St
                                                 }}
                                                 menuIsOpen={this.state.showEditJobTime}
                                             /> : <span
-                                                                              className='JobSelectedtime'
-                                                                               >
-                                                                              <b>{this.getJobStartTime()}</b>
-                                                                          </span>
+                                                className='JobSelectedtime'
+                                            >
+                                                <b>{this.getJobStartTime()}</b>
+                                            </span>
                                             }
                                             <a
                                                 className='EditJobTime'
--- a/components/admin_console/data_retention_settings/data_retention_settings.tsx
+++ b/components/admin_console/data_retention_settings/data_retention_settings.tsx
@@ -1,5 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+/* eslint-disable max-lines */
 
 import React, {createRef, RefObject} from 'react';
 import {FormattedMessage} from 'react-intl';
@@ -554,7 +558,7 @@ export default class DataRetentionSettings extends React.PureComponent<Props, St
                                 expanded={true}
                             >
                                 <JobsTable
-                                    jobType={JobTypes.DATA_RETENTION}
+                                    jobType={JobTypes.DATA_RETENTION as JobType}
                                     hideJobCreateButton={true}
                                     className={'job-table__data-retention'}
                                     disabled={String(DataRetentionSettings?.EnableMessageDeletion) !== 'true' && String(DataRetentionSettings?.EnableFileDeletion) !== 'true'}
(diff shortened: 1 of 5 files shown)


components/admin_console/jobs/table.tsx
```diff
@@ -1,82 +1,47 @@
...
+
+export type Props = {
+    jobs?: Job[];
+    getExtraInfoText?: (job: Job) => React.ReactElement;
+    disabled?: boolean;
+    createJobHelpText: React.ReactElement;
+    createJobButtonText: React.ReactElement;
+    jobType: JobType;
+    downloadExportResults?: boolean;
+    actions?: {
```
hahmadia, line 28:
```suggestion
    actions: {
```

Since actions are always defined, we can remove the `?` optional question mark.

hahmadia (changes requested):
Some changes I see that can be made. I've always tagged my colleague Harrison who is much more experienced with Typescript and may have better suggestions.

Thanks for the contribution!

components/admin_console/data_retention_settings/data_retention_settings.tsx
```diff
@@ -1,5 +1,10 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+/* eslint-disable react/jsx-closing-bracket-location */
```
hmhealey, line 6: Could you remove this eslint-disable? I'm okay leaving the one for `max-lines` for now since I know that warning is a bit contentious within the team, but we shouldn't be disabling other rules unless we have a good reason to

components/admin_console/jobs/table.test.tsx
```diff
@@ -80,7 +81,7 @@ describe('components/admin_console/jobs/table', () => {
             status: 'warning',
             type: 'data_retention',
         }],
-    };
+    } as unknown as Props;
```
hmhealey, line 84: Instead of using `as Props`, could you pass mocked objects for the missing fields? If it's because we don't have mocks for `jobs`, you could override the type for that, but we'd rather not use `as Props` in these tests since it'll hide when stuff fails from `props` being incorrect

components/admin_console/jobs/table.tsx
```diff
@@ -1,82 +1,47 @@
...
+        getJobsByType: FuncFromArgAndReturn<typeof getJobsByType>;
+        createJob: FuncFromArgAndReturn<typeof createJob>;
+        cancelJob: FuncFromArgAndReturn<typeof cancelJob>;
     };
```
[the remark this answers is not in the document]
hmhealey: Agreed, although I think we can simplify these further though since we don't actually care about the results of the actions
```suggestion
        getJobsByType: (job: JobType) => Promise<void>;
        createJob: (job: JobTypeBase) => void;
        cancelJob: (jobID: string) => void;
    };
```

components/admin_console/jobs/table.tsx
```diff
@@ -1,82 +1,47 @@
...
+    className?: string;
+    hideJobCreateButton?: boolean;
+}
+
+export type State = {
+    loading: boolean;
+}
+
+class JobTable extends React.PureComponent<Props & { intl: IntlShape }, State> {
```
hmhealey, line 41: Why not include `intl` in `Props`?

hmhealey (changes requested):

components/admin_console/data_retention_settings/data_retention_settings.tsx
ArcaneDiver: Hi,
i had to disable that rules because of a eslint bug (i suppose) because in those lines:
https://github.com/mattermost/mattermost-webapp/blob/25d3f702d5b43bc0b7c853702bfa7b492ea52b2d/components/admin_console/data_retention_settings/data_retention_settings.tsx#L578-L613

Eslint signaled an error in the closure of the component but when I went to solve it a new error appeared and when I went to solve it the first reappeared, essentially it was a vicious circle.

So the only way that i found to fix it was to disable the rule.

components/admin_console/jobs/table.tsx
ArcaneDiver: Sorry i forgot to move it to `Props`

components/admin_console/jobs/table.tsx
```diff
@@ -1,82 +1,47 @@
...
-         * Allows for custom styles on the job table component
-         */
-        className: PropTypes.string,
-
-        /**
-         * Hide the job creation button. This is useful if you want to place the button elsewhere on your page or hide it.
-         */
-        hideJobCreateButton: PropTypes.bool,
+type ArgumentTypes<F extends (...args: any) => any> = F extends (...args: infer A) => any ? A : never;
+type FuncFromArgAndReturn<A extends (...args: any) => any> = (...args: ArgumentTypes<A>) => any;
```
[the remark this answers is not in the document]
ArcaneDiver: Sure

components/admin_console/jobs/table.tsx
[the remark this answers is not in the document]
ArcaneDiver: Hi,
by doing this i get the following error:
```
The types returned by 'getJobsByType(...)' are incompatible between these types.
   Type 'Promise<void>' is not assignable to type 'ActionFunc'.
      Type 'Promise<void>' provides no match for the signature '(dispatch: DispatchFunc, getState: GetStateFunc): ActionResult | Promise<ActionResult | ActionResult[]>'
```

from https://github.com/mattermost/mattermost-webapp/blob/25d3f702d5b43bc0b7c853702bfa7b492ea52b2d/components/admin_console/jobs/index.ts#L32

components/admin_console/jobs/table.tsx
```diff
@@ -318,24 +283,24 @@ class JobTable extends React.PureComponent {
...
+    handleCancelJob = async (e: React.UIEvent) => {
         e.preventDefault();
         const jobId = e.currentTarget.getAttribute('data-job-id');
-        await this.props.actions.cancelJob(jobId);
+        await this.props.actions!.cancelJob(jobId as string);
         this.reload();
     };
 
-    handleCreateJob = async (e) => {
+    handleCreateJob = async (e: React.UIEvent) => {
```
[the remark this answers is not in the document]
ArcaneDiver: It's ok even if on mobile it's `React.TouchEvent`?

components/admin_console/jobs/table.tsx
ArcaneDiver: Sure

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @ArcaneDiver

components/admin_console/data_retention_settings/data_retention_settings.tsx
hmhealey: Maybe we could solve that by moving that out of the final return and using a regular if statement instead of a ternary then?
```javascript
let jobTime;
if (this.state.showEditJobTime) {
    jobTime = (
        <ReactSelect
            ...
        />
    );
} else {
    jobTime = <span .../>;
}
```

Personally, I prefer to split up that sort of code when you're rendering something complicated anyway.

components/admin_console/jobs/table.tsx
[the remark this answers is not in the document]
hmhealey: I guess that's why we've had to tell `bindActionCreators` the types of the actions in the past. If you change the `bindActionCreators` call in the `mapDispatchToProps` to look [like this](https://github.com/mattermost/mattermost-webapp/blob/master/components/new_channel_flow/index.ts#L39), it'll fix those warnings and let you use the simpler types for the actions.

We have some weirdness caused by our custom types for redux actions (like `ActionFunc`) that unfortunately makes them a bit difficult to use, but the pattern in the file I linked to is the best version of that that I'm aware of so far.

components/admin_console/jobs/table.tsx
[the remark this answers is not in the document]
hahmadia: @ArcaneDiver Mobile does not have access to the admin console. No need to worry about mobile.

hahmadia:
Please tag both @hmhealey and I once you are ready to have your PR re-reviewed. Thank you.

devinbinnie:
@ArcaneDiver Did you have any additional questions from us for this PR? Please let us know :)

M-ZubairAhmed:
done in https://github.com/mattermost/mattermost-webapp/pull/9491

PR #9274, opened 2021-10-28
Fixes threads list inconsistencies

koox00:
#### Summary

We stopped populating the threads lists when we receive a new thread
this proved to create inconsistencies in threads list.

This commit adds threads when received and they are newer than already
existing ones, older threads won't be added since those are going to be
added by paginating (infinite scroll).

Furthermore READ_CHANGED_THREAD action changes the unreadThreadsInTeam
reducer. When the thread is read it removes it, and when the thread is
unread it adds it (only if the thread is newer).

This PR is created as a follow up of #9206 
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39452

#### Release Note

```release-note
NONE
```


packages/mattermost-redux/src/reducers/entities/threads/threadsInTeam.ts
```diff
@@ -5,11 +5,22 @@ import {ChannelTypes, PostTypes, TeamTypes, ThreadTypes, UserTypes} from 'matter
...
 
 import {ExtraData} from './types';
 
 type State = ThreadsState['threadsInTeam'] | ThreadsState['unreadThreadsInTeam'];
 
+// return true only if it's 'newer' than other threads
+// older threads will be added by scrolling so no need to manually add.
+// furthermore manually adding older thread will BREAK pagination
+function shouldAddThreadId(ids: Array<$ID<UserThread>>, thread: UserThread, threads: IDMappedObjects<UserThread>) {
+    return ids.some((id) => {
```
hmhealey, line 18: Is it intended that this returns true if the given thread is newer than at least one of the stored threads? Perhaps it should only be added if it's newer than every stored thread? 0/5 on whether that should be the case though.

hmhealey (commented):
One question, but this looks good otherwise

packages/mattermost-redux/src/reducers/entities/threads/threadsInTeam.ts
koox00: The case is a bit weird:
We load threads from newest to oldest (top to bottom) in the list, 
and then paginate asking each time the threads before the bottom one, 
older threads will be shown in the list upon scrolling so no problem there.
if we manually add them they might create gaps in the pagination so we don't*. 

Newer threads (or in-between threads) though, would never have a change to be added via scrolling
since we already are past that scrolling point (that's why they are newer), 
so each thread newer than the oldest one should be manually added in the list or 
we won't have a chance to load them**.

components/threading/global_threads/thread_list/thread_list.tsx
```diff
@@ -116,7 +116,11 @@ const ThreadList = ({
     }, [setFilter]);
 
     const handleLoadMoreItems = useCallback(async (startIndex) => {
-        const before = data[startIndex - 1];
+        let before = data[startIndex - 1];
+
+        if (before === selectedThreadId) {
+            before = data[startIndex - 2];
+        }
```
koox00, line 123: `selectedThreadId` is a thread that might be in the list only because it's
selected, so let's don't take into account this one when we paginate for
results.

packages/mattermost-redux/src/reducers/entities/threads/threadsInTeam.ts
hmhealey: Ah, I see now. That makes sense then.

Not something to change here, but this seems like a very similar problem to how we have to keep track of posts that are loaded in a sparse list where there might be items missing. That code is very post-specific at the moment, and there's some changes I'd like to make to simplify it if we end up ever needing to update it, but I'm wondering now if there's anything that we could do to make it more generic. At the time, it was the only place where we cared about loading part of a list of items that are sorted by time, but loading threads for that list does feel very similar. But at the same time, threads can change order which makes that messier, although thankfully don't have to be loaded sparsely since you're always viewing them in recency order.

Anyway, that's just some rambling thoughts on my part. This looks good to me 😛

hmhealey (approved):

anurag6713 (approved):
LGTM! 👍

koox00:
> I'm unable to login on the test server, updating branch and may end up having to rebuild the test server.

I've had populated the PRs database with sampledata, so the credentials differ, sorry I didn't realized it before.

Username: sysadmin
Password : Sys@dmin-sample1

jgilliam17 (approved):
Thanks @koox00 
Tested, looks good to merge.
- Verified new threads are added to the threads list as they are received.

jgilliam17:
Thanks @koox00 Re-tested, the bug with marking a thread as unread when one thread is selected and unreads are empty is fixed. Looks good to merge.

PR #9288, opened 2021-10-29
Fixed a bug where user status is not displayed on first load of a product

harshilsharma63:
#### Summary
Currently, when you open Boards or Playbooks from Channels, your status is always displayed as offline in the global header.
This PR fixes that.

#### Ticket Link
https://github.com/mattermost/focalboard/issues/1481

#### Related Pull Requests
None

#### Release Note
```release-note
Fixes a bug where the user's status was always as offline on opening a product from Channels.
```


actions/user_actions.jsx
```diff
@@ -422,10 +422,16 @@ export function autocompleteUsers(username) {
     };
 }
 
+export function setUserStatus(currentUserId) {
```
hmhealey, line 425: Instead of adding a new action for this, could we have the StatusDropdown just use the action from mattermost-redux directly?

hmhealey (commented):
How is the user's status currently loaded by the web app when viewing Channels? Are we loading it explicitly, or is it just being loaded by something else? While this is needed when viewing Boards or Playbooks, I want to make sure we're not duplicating requests made on load in Channels since we already make too many requests on load there

hmhealey:
Funnily enough, I just found the code that loads the status currently, and it's kind of out of the way since it's loaded by the [ResetStatusModal](https://github.com/mattermost/mattermost-webapp/blob/master/components/reset_status_modal/reset_status_modal.jsx) calling the `autoResetStatus` action defined just below the new one you added.

Instead of doing that since it's rather out of the way, we should add status loading to [`getMeAndConfig`](https://github.com/mattermost/mattermost-webapp/blob/master/actions/views/root.js#L13) that's called before the app renders to begin with. We could remove the extra call to load the status from `autoResetStatus` then to avoid loading the current user's status twice when viewing Channels then.

components/status_dropdown/status_dropdown.tsx
```diff
@@ -83,6 +84,10 @@ export default class StatusDropdown extends React.PureComponent <Props, State> {
         };
     }
 
+    componentDidMount() {
```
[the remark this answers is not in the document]
harshilsharma63: In the browser when yous witch products from the app switcher, nothing is reloaded and the API call is not made again. In the desktop app, switching products always open them up in new tabs and hence new API calls are made.

actions/user_actions.jsx
harshilsharma63: Done.

harshilsharma63:
> Funnily enough, I just found the code that loads the status currently, and it's kind of out of the way since it's loaded by the [ResetStatusModal](https://github.com/mattermost/mattermost-webapp/blob/master/components/reset_status_modal/reset_status_modal.jsx) calling the `autoResetStatus` action defined just below the new one you added.
> 
> Instead of doing that since it's rather out of the way, we should add status loading to [`getMeAndConfig`](https://github.com/mattermost/mattermost-webapp/blob/master/actions/views/root.js#L13) that's called before the app renders to begin with. We could remove the extra call to load the status from `autoResetStatus` then to avoid loading the current user's status twice when viewing Channels then.

Done.

packages/mattermost-redux/src/actions/users.ts
```diff
@@ -249,6 +249,7 @@ export function loadMe(): ActionFunc {
         const user = getState().entities.users.profiles[currentUserId];
         if (currentUserId) {
             Client4.setUserId(currentUserId);
+            await dispatch(getStatus(currentUserId));
```
hmhealey, line 252: Do we need to wait for this to complete? Since we aren't able to fire off this request while we're loading everything else, perhaps we could just dispatch it without `awaiting` on it

packages/mattermost-redux/src/reducers/entities/users.ts
```diff
@@ -419,6 +419,17 @@ function profilesInGroup(state: RelationOneToMany<Group, UserProfile> = {}, acti
     }
 }
 
+function currentUserStatus(state: RelationOneToOne<UserProfile, string> = {}, action: GenericAction) {
```
hmhealey, line 422: Why store the current user's status separately from the other users' statuses?

packages/mattermost-redux/src/store/initial_state.ts
```diff
@@ -2,6 +2,7 @@
 // See LICENSE.txt for license information.
 
 import {GlobalState} from 'mattermost-redux/types/store';
+import {UserStatuses} from '../../../../utils/constants';
```
hmhealey, line 5: We can't import code outside of mattermost-redux into the package. Do we not have a version of this constant in mattermost-redux?

Ideally, we wouldn't duplicate those constants, but that's something we need to resolve for a lot more than just the statuses

packages/mattermost-redux/src/actions/users.ts
harshilsharma63: Yes, we do need to wait for this. This is because the `ResetStatusModal` component fetches the status only on the component mount (and not updated) and if the status is not found in the store, sets it to offline. Not awaiting here causes the status to always be displayed as offline.

packages/mattermost-redux/src/reducers/entities/users.ts
harshilsharma63: Because the store structure is different. For other users we only store the string representation (`online`, `offline`, 'away` or `dnd`) whereas for the current user we need the full status object which contains custom status details as well.

packages/mattermost-redux/src/store/initial_state.ts
harshilsharma63: No, we don't have this constant in the redux package. I've updated it to import the `../constants` instead, as was done in another import in the redux package.

packages/mattermost-redux/src/reducers/entities/users.ts
hmhealey: Ah right. I forgot that we still only store the status string for other users. There's designs around for using some of the other details in the profile popover, but I guess we didn't do that.

packages/mattermost-redux/src/actions/users.ts
hmhealey: I don't think we can do that since it'll make the first load longer. I think we either need to load the status asynchronously while the other stuff loads (like in the `Promise.all` above) or we could make the modal handle the unloaded state properly

chenilim:
@harshilsharma63, what's the next step on this PR? Is there anything that can be done in the short term (e.g. refresh status after 200ms in Boards), if there are blocking issues on the full fix?

hmhealey:
Removing my review request on this to clear out some notifications. Let me know when this is ready for my review again

PR #8541, opened 2021-08-09
[MM-36444] Manage 6.0 deprecations

BenCookie95:
#### Summary
Removing deprecated configs listed in this ticket https://mattermost.atlassian.net/browse/MM-36444.

There is a small bit of overlap with https://github.com/mattermost/mattermost-webapp/pull/8503

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36444

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/18035

#### Release Note
```release-note
Removing deprecated configs from webapp:
- EnableOnlyAdminIntegrations
- RestrictCustomEmojiCreation
- RestrictPostDelete
- AllowEditPost
- ImageProxyType
- ImageProxyURL
- ImageProxyOptions
- UseExperimentalGossip
- EnableTeamCreation 
- RestrictTeamInvite
- RestrictPublicChannelManagement
- RestrictPrivateChannelManagement
- RestrictPublicChannelCreation
- RestrictPrivateChannelCreation
- RestrictPublicChannelDeletion
- RestrictPrivateChannelDeletion
- RestrictPrivateChannelManageMembers
- DisableLegacyMFAEndpoint
- ExperimentalTownSquareIsReadOnly
- ExperimentalHideTownSquareinLHS
- EnableXToLeaveChannelsFromLHS
```

PR #9307, opened 2021-11-01
MM-39199 Fix permissions for reactions buttons

hmhealey:
This fixes some weirdness with the permissions around emoji reactions. The following weird cases are fixed:
1. Non-admins weren't able to click on existing reactions to add them in the Town Square with a license. This seems to be a side effect of https://github.com/mattermost/mattermost-webapp/pull/8541
2. Users without the permission to react to a post wouldn't be able to click on existing reactions, but they would see the + button with the emoji picker. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39199

#### Release Note
```release-note
Fix emoji reaction buttons on posts not respecting user permissions
```


koox00 (approved):

BenCookie95 (approved):

jgilliam17:
Thanks @hmhealey 
Seeing one issue, +emoji icon is missing after first reaction has been added. This is reproing in all cases, even when all channel permissions are enabled.
Screen Shot 2021-11-03 at 10 21 22 AM

hmhealey:
@jgilliam17 That should be fixed now. It turns out I was mixing up things called `canAddReaction` and `canAddReactions` when passing around permissions.

PR #9310, opened 2021-11-02
Fix several bugs on saved posts and mentions

larkox:
#### Summary
Bugs fixed:
- Permalink always link to current team instead of post team on mentions and saved posts
- Timestamp click always sends to the current team instead of post team on mentions and saved posts
- When moving from saved posts to mentions, saved posts are shown for a short time before showing the mentions results.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39815
https://mattermost.atlassian.net/browse/MM-39816
https://mattermost.atlassian.net/browse/MM-39466

#### Related Pull Requests
None

#### Screenshots
None

#### Release Note
```release-note
None
```


koox00 (approved):

amyblais:
@larkox Some tests are failing.

components/dot_menu/index.ts
```diff
@@ -122,7 +123,7 @@ function mapStateToProps(state: GlobalState, ownProps: Props) {
         pluginMenuItems: state.plugins.components.PostDropdownMenu,
         canEdit: PostUtils.canEditPost(state, post, license, config, channel, userId),
         canDelete: PostUtils.canDeletePost(state, post, channel),
-        currentTeamUrl,
+        teamUrl,
         currentTeamId: currentTeam.id,
```
hmhealey, line 127: We're still using the current team ID for some stuff in the component like following threads which seem like they might break when viewing messages on another team.

Perhaps instead of passing the URL around, we should just pass the team ID around? That may not work for posts in DMs though, so feel free to ignore that part

amyblais:
@larkox Kind note to prioritize for tomorrow's v6.1 code freeze.

larkox:
@jgilliam17 I have found a few potential bugs, so feel free to defer the testing until I push some more changes.

larkox:
@hmhealey I saw what you meant with the use of team ID. I have fixed that (at least for dot menu, that seemed the only place needed for search items). Let me know if you find any other issue.

@jgilliam17 Unless @hmhealey finds any other issue, this should be ready to test.

@amyblais Regarding https://mattermost.atlassian.net/browse/MM-39867, I haven't been able to find the source, so it is not fixed in this PR. I will keep investigating, but I don't think it will make it to the code freeze.

amyblais:
> @amyblais Regarding https://mattermost.atlassian.net/browse/MM-39867, I haven't been able to find the source, so it is not fixed in this PR. I will keep investigating, but I don't think it will make it to the code freeze.

@larkox Please continue to investigate and perhaps we can include the fix for v6.1 later this week.

jgilliam17 (approved):
Thanks @larkox 
Tested, looks good to merge.
- Verified clicking on a timestamp of the saved post redirects to the correct team, permalink to the saved posts directs to the correct team and opening mentions no longer flashes previous RHS content.

amyblais:
@hmhealey Kind note to prioritize for today's v6.1 code freeze.

components/dot_menu/index.ts
```diff
@@ -127,11 +128,11 @@ function mapStateToProps(state: GlobalState, ownProps: Props) {
         components: state.plugins.components,
         postEditTimeLimit: config.PostEditTimeLimit,
         isLicensed: license.IsLicensed === 'true',
-        teamId: getCurrentTeamId(state),
+        teamId: channel.team_id,
```
hmhealey, line 131: I think this might cause an issue with the Add Reactions option [here](https://github.com/mattermost/mattermost-webapp/blob/84417ddbb12d51f1c4ddbc73c35ee6fdceb3ab18/components/dot_menu/dot_menu.tsx#L496) since `teamId` will be empty now for DMs/GMs.

Maybe we could just remove the `currentTeamId` prop entirely and just have this be `channel.team_id || currentTeam.id`?

hmhealey (changes requested):

amyblais:
@larkox Some checks are failing.

hmhealey (approved):
Thanks! Knowing that we had both teamId and currentTeamId before was also bugging me more than it should have, so I'm happy to have that fixed as well 😛

PR #9323, opened 2021-11-04
[MM-38239 & MM-39788] Recent files causing crash

BenCookie95:
#### Summary
The crash was caused by missing posts in the redux state. I added a check to fetch any posts that are missing and I also stopped the crash if a post is missing for whatever reason.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38239
https://mattermost.atlassian.net/browse/MM-39788

#### Related Pull Requests
- https://github.com/mattermost/mattermost-server/pull/18942

#### Release Note

```release-note
NONE
```


packages/mattermost-redux/src/client/client4.ts
```diff
@@ -2037,6 +2037,13 @@ export default class Client4 {
         );
     };
 
+    getPostsByIds = (postIds: string[]) => {
+        return this.doFetch<StatusOK>(
```
hmhealey, line 2041: This return type seems incorrect since I'd assume it's returning an array of posts, not just a `StatusOK`

packages/mattermost-redux/src/actions/posts.ts
```diff
@@ -998,6 +998,30 @@ export function getProfilesAndStatusesForPosts(postsArrayOrMap: Post[]|Map<strin
...
+export function getPostsByIds(ids: string[]) {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        let posts;
+
+        try {
+            posts = await Client4.getPostsByIds(ids);
+        } catch (error) {
+            forceLogoutIfNecessary(error, dispatch, getState);
+            dispatch(batchActions([
+                {type: PostTypes.GET_POSTS_FAILURE, error},
```
hmhealey, line 1010: I think we can omit this action. We used to keep track of a bunch of request state with actions like these, but most of it isn't used any more.

I've got a ticket around to go clean that up eventually since we started removing it at some point before we kind of forgot about it.

hmhealey (changes requested):

hmhealey (approved):
Thanks for changing those. I forgot our janky definition of `DispatchFunc` is what required that `as` to be added. I'm hoping some time tomorrow for Open Source Friday to figure out what the heck is wrong with those

mkraft (approved):
@BenCookie95 Does this need to also handle *deleted* posts, and if so does it do so already?

BenCookie95:
> @BenCookie95 Does this need to also handle _deleted_ posts, and if so does it do so already?

Good question, the crash won't occur again because I've handled the UI to not expect the post to exist but I don't know if files still appear for deleted posts. I'll test some use cases around deleted posts

furqanmlk (approved):
Working as expected
https://user-images.githubusercontent.com/6909890/142287686-1e884d2e-7db3-4f68-875e-22205bfa8d74.mp4

BenCookie95:
Will give this one more run through this afternoon before merging

PR #9333, opened 2021-11-07
fix: set token if not already set

SebastianSpeitel:
Fixes: https://github.com/mattermost/mattermost-redux/issues/1359

Currently the token isn't set on login, so using login actually does nothing and any following request that requires permission fails with 401 `'api.context.session_expired.app_error'`

### DIFF
--- a/packages/mattermost-redux/src/client/client4.ts
+++ b/packages/mattermost-redux/src/client/client4.ts
@@ -679,7 +679,7 @@ export default class Client4 {
         );
     }
 
-    login = (loginId: string, password: string, token = '', deviceId = '', ldapOnly = false) => {
+    login = async (loginId: string, password: string, token = '', deviceId = '', ldapOnly = false) => {
         this.trackEvent('api', 'api_users_login');
 
         if (ldapOnly) {
@@ -697,10 +697,19 @@ export default class Client4 {
             body.ldap_only = 'true';
         }
 
-        return this.doFetch<UserProfile>(
+        const {
+            data: profile,
+            headers,
+        } = await this.doFetchWithResponse<UserProfile>(
             `${this.getUsersRoute()}/login`,
             {method: 'post', body: JSON.stringify(body)},
         );
+
+        if (headers.has('Token')) {
+            this.setToken(headers.get('Token')!);
+        }
+
+        return profile;
     };
 
     loginById = (id: string, password: string, token = '', deviceId = '') => {
--- a/packages/mattermost-redux/src/client/client4.ts
+++ b/packages/mattermost-redux/src/client/client4.ts
@@ -146,6 +146,7 @@ export default class Client4 {
     userId = '';
     diagnosticId = '';
     includeCookies = true;
+    setAuthHeader = true;
     translations = {
         connectionError: 'There appears to be a problem with your internet connection.',
         unknownError: 'We received an unexpected status code from the server.',
@@ -449,7 +450,7 @@ export default class Client4 {
             ...this.defaultHeaders,
         };
 
-        if (this.token) {
+        if (this.setAuthHeader && this.token) {
             headers[HEADER_AUTH] = `${HEADER_BEARER} ${this.token}`;
         }
 
--- a/components/root/root.jsx
+++ b/components/root/root.jsx
@@ -121,6 +121,9 @@ export default class Root extends React.PureComponent {
         // Redux
         setUrl(getSiteURL());
 
+        // Disable auth header to enable CSRF check
+        Client4.setAuthHeader = false;
+
         setSystemEmojis(EmojiIndicesByAlias);
 
         // Force logout of all tabs if one tab is logged out


hmhealey:
Hi @SebastianSpeitel. Thanks for the PR!

That seems really weird that we don't set the token at all since we definitely did that at some point. I know the web app uses the cookie to pass the auth token back to the server, so it doesn't need to provide it as a header, but I'm not sure how non-web apps (or even our mobile app) would handle that. I want to look into why that might've changed to see if it was just a mistake or if there's something we might be breaking by re-adding that.

packages/mattermost-redux/src/client/client4.ts
```diff
@@ -3737,6 +3737,10 @@ export default class Client4 {
         const response = await fetch(url, this.getOptions(options));
         const headers = parseAndMergeNestedHeaders(response.headers);
 
+        if (this.getToken() === undefined && headers.has('Token')) {
```
hmhealey, line 3740: Also, maybe this can be done in response to just the login request [like we used to do before](https://github.com/mattermost/mattermost-redux/blame/a10b6c7f67a5163c15d029defbcd9bccf0eb0412/src/client/client4.js#L450)? It shouldn't make much difference overall, but I think it's nice to keep it with the login method itself.

hmhealey (changes requested):
Alright, so the reason that we don't want this token set in the Mattermost web app is that the header that's used to pass the auth token also disables the CSRF check since we don't want that on requests made from outside the web app. We need some way to not have that header set for the web app and set automatically for everyone else.

Maybe we could add a new `setAuthHeader` field on `Client4` that's set to true by default and controls whether or not the `Authorization` header is added to the request? Then we could have the web app set that to false [here](https://github.com/mattermost/mattermost-webapp/blob/master/components/root/root.jsx#L122) when it configures some other stuff about the client like the URL. That way the web app and its plugins wouldn't send the header to the server, but `login` will be fixed for everyone else using the client?

SebastianSpeitel:
Now the Token is only set on login.

hmhealey:
Thanks for changing that. Did you see my other request around making this behaviour opt-out so that we can turn it off in the main web app?

SebastianSpeitel:
> Did you see my other request around making this behaviour opt-out so that we can turn it off in the main web app?

Yes I did, I will try to add but have to figure out where to implement that.

hmhealey:
I think adding a field to the Client4 would make sense so that the web app can override the default behaviour you added. I was thinking that it would make it so that `this.authToken` isn't set automatically when it was enabled, but I could see the field making it so that the `Authorization` header isn't set instead. Both seem like they'd work equally well.

SebastianSpeitel:
I added an option on `Client4` which configures if the Auth header is set and enabled it by default. 
I probably should disable it in the web app in this PR, but I'm not sure where to look.

hmhealey:
I think [here](https://github.com/mattermost/mattermost-webapp/blob/master/components/root/root.jsx#L122) would be the right place to do it since that's where the web app sets the URL on the Client4.

I'm not sure why it does that through mattermost-redux, but that's fine to access Client4 here directly to do that.

SebastianSpeitel:
It seems the `setUrl` was used to enable mocking:

https://github.com/mattermost/mattermost-webapp/blob/master/components/root/root.test.jsx#L41-L43

hmhealey (approved):
Thanks for that change. This looks good to me for both the Mattermost web app case and the standalone one.

crspeller (approved):
I think this should be alright. Going to ask for a security review.

srkgupta (approved):
Thanks @SebastianSpeitel. LGTM from security perspective.

lindalumitchell:
@SebastianSpeitel @crspeller @hmhealey is there some sort of ticket or issue to help give context and spell out expected behavior for the changes (and possible regression risks to test for)? Not sure I'm getting the picture from reading the comments.

Also is there impact on any existing tests and/or are new tests of any kind needed? :)

hmhealey:
There's no ticket since I don't think we realized anything was broken. A quick explanation is that if someone used the API client to build a bot or integration, it wouldn't be able to log in correctly because it didn't actually save the auth token. The web app never had that issue since it saved the auth token a different way.

For testing, I think as long as you can log in and make a post, we can tell that the web app still authenticates correctly through how it works, so a simple smoke test should be enough here

lindalumitchell:
Thanks @hmhealey! Logging in with email and password works fine. I am getting an error trying to use OpenID SSO (tried both GitLab and Google Apps). `Bad response from token request.` I don't see anything obviously off in my config settings; could these changes affect OpenID login as well?

Screen Shot 2021-12-07 at 9 59 39 PM

hmhealey:
No, they shouldn't. The token that this PR is referring to is a different token than the one in that error message. Digging into the server logs, the error message also includes `redirect_uri_mismatch`, so it seems like there's something configured wrong on the Google side of things

lindalumitchell (approved):
Okay cool, sounds good, thanks. Approving.

hmhealey:
Thanks again for the PR, @SebastianSpeitel!

At the moment, we're not shipping any new versions of mattermost-redux, so this won't be available on npm for some time, but people who want to access it can still include the mattermost web app as a git dependency similar to how some of our plugins do. That requires [a bit of magic](https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/webapp/webpack.config.js#L51) to get imports working right, but it's at least a workaround for that in the mean time

SebastianSpeitel:
Thanks for applying my fix. I don't need it myself anymore, since I use a bot token to authenticate now. But this makes it way easier for playing around with your own credentials without needing a token.

PR #9334, opened 2021-11-08
chore/MM-39653 : Migrate FilePreviewModal to be controlled by the ModalController

M-ZubairAhmed:
#### Summary
Migrated FilePreviewModal to be controlled with the ModalController instead of directly rendering in the respective componenets. Following is where the component modal was inside of render method and now is removed and called via `openModal`
- [x] PostImage
- [x] MarkdownImage
- [x] FileAttachmentList
- [x] SingleImagePreview
- [x] FileSearchResultItem

#### Ticket Link

Jira https://mattermost.atlassian.net/browse/MM-39653

#### Related Pull Requests

#### Release Note

```release-note

```


components/file_preview_modal/file_preview_modal.jsx
```diff
@@ -83,7 +76,6 @@ export default class FilePreviewModal extends React.PureComponent {
             showZoomControls: false,
             scale: Utils.fillArray(ZoomSettings.DEFAULT_SCALE, this.props.fileInfos.length),
         };
-        this.videoRef = React.createRef();
```
M-ZubairAhmed, line 86: this ref isn't used

components/file_preview_modal/file_preview_modal.jsx
```diff
@@ -261,9 +238,9 @@ export default class FilePreviewModal extends React.PureComponent {
     }
 
     handleModalClose = () => {
-        this.props.onModalDismissed();
-        this.setState({scale: Utils.fillArray(ZoomSettings.DEFAULT_SCALE, this.props.fileInfos.length)});
```
M-ZubairAhmed, line 265: we dont have to reset the scale since the component will be remounted when opened and the scales will reset

components/markdown_image/markdown_image.jsx
```diff
@@ -66,17 +71,25 @@ export default class MarkdownImage extends React.PureComponent {
         return parseInt(height, 10);
     }
 
-    showModal = (e) => {
-        if (!this.props.imageIsLink) {
+    showModal = (e, link, imageMetadata, extension, name) => {
+        if (!this.props.imageIsLink && extension) {
```
M-ZubairAhmed, line 75: should we add extensions to the check too, i added since it was in https://github.com/mattermost/mattermost-webapp/pull/9334/files#diff-ececdf452d72bd2b2ad7c33364a8648ae2e29476f3f0be17a1787cdb98c1c532L185

components/markdown_image/markdown_image.jsx
```diff
@@ -152,51 +165,32 @@ export default class MarkdownImage extends React.PureComponent {
 
                     let className = '';
                     if (this.state.loaded) {
-                        className = imageIsLink || !extension ?
-                            `${this.props.className} markdown-inline-img--hover markdown-inline-img--no-border` :
-                            `${this.props.className} markdown-inline-img--hover cursor--pointer a11y--active`;
```
M-ZubairAhmed, line 157: this was eslint warning

components/file_attachment_list/file_attachment_list.tsx
```diff
@@ -94,7 +71,7 @@ export default class FileAttachmentList extends React.PureComponent<Props, State
                         key={fileInfo.id}
                         fileInfo={sortedFileInfos[i]}
                         index={i}
-                        handleImageClick={this.handleImageClick}
+                        handleImageClick={(index) => this.handleImageClick(index, sortedFileInfos, postId)}
```
hmhealey, line 74: Did you happen to figure out why the values from props were undefined in `handleImageClick` after your other changes?

Like I mentioned, it'd be better if we could avoid generating new functions and passing them as props into other redux components so that we can avoid all of the children of a component re-rendering when the parent re-renders

components/markdown_image/markdown_image.jsx
```diff
@@ -152,51 +165,32 @@ export default class MarkdownImage extends React.PureComponent {
...
+                        <SizeAwareImage
+                            alt={alt}
+                            className={className}
+                            src={safeSrc}
+                            height={height === 'auto' ? undefined : height}
+                            width={width === 'auto' ? undefined : width}
+                            title={title}
+                            dimensions={imageMetadata}
+                            showLoader={false}
+                            onClick={(e) => this.showModal(e, safeSrc, imageMetadata, extension, alt)}
```
hmhealey, line 190: Same here

components/post_view/post_image/post_image.jsx
```diff
@@ -47,19 +53,15 @@ export default class PostImage extends React.PureComponent {
...
-                                post={this.props.post}
-                                startIndex={0}
-                                fileInfos={[{
-                                    has_preview_image: false,
-                                    link: safeLink,
-                                    extension: this.props.imageMetadata.format,
-                                    name: this.props.link,
-                                }]}
+                                onClick={(e) =>
+                                    this.showModal(
```
hmhealey, line 57: And here

hmhealey (changes requested):

components/file_attachment_list/file_attachment_list.tsx
M-ZubairAhmed: @hmhealey can you try looking into converting to a normal function instead of arrow function, I did it a way of passing the paramaters as props to component and calling the function with those passed in props https://github.com/mattermost/mattermost-webapp/pull/9334/commits/5310cf9668d064a3ce2beb3265ee13dadbeb8916

M-ZubairAhmed: so it seems normal function was working too , thank you for https://github.com/mattermost/mattermost-webapp/pull/9334/commits/31e93809e071074d27eda3ac98042860727cb03e

components/markdown_image/markdown_image.jsx
M-ZubairAhmed: converted to regular function

components/post_view/post_image/post_image.jsx
M-ZubairAhmed: converted to regular function

components/markdown_image/markdown_image.jsx
devinbinnie: Makes sense to me. I'm not sure what the failure case looks like for when there isn't an extension though, since with this it would just not pop a modal vs. before where it would pop a modal but not load an image.

devinbinnie (approved):

hmhealey (approved):

jgilliam17:
Thanks @M-ZubairAhmed 
Tested, looks good to merge.
- Verified File Preview modal is as before.
[E2E report](https://automation-dashboard.vercel.app/cycles/dcee6044-7105-4fd2-9ff9-1ee5e16d3804) - verified that failures are not related to these chages.

jgilliam17 (approved):

PR #9335, opened 2021-11-08
Add .nvmrc

lieut-data:
#### Summary
Introduce a `.nvmrc` to simplify the setup and workflow for users using the [Node Version Manager](https://github.com/nvm-sh/nvm).

#### Ticket Link
N/A

#### Release Note
```release-note
NONE
```


koox00 (approved):
Nice addition!

hmhealey (approved):

M-ZubairAhmed:
this is nice because i kept that file and never committed it to the repo 😛

M-ZubairAhmed:
may be we can add in docs to just do `nvm install` and it will install correct node based on `.nvmrc` file ?

hmhealey:
Yeah, that'd be a good idea. We already mention how they can install the correct version of Node using NVM, so it'd probably make sense to just always have people use NVM.

If you want to submit a PR to change that, the repo for https://developers.mattermost.com is https://github.com/mattermost/mattermost-developer-documentation/

PR #9346, opened 2021-11-09
POC - move apply theme functionality to `CompassThemeProvider`

michelengelen:
#### Summary
this PR aims at moving "all-things-theming" to the `CompassThemeProvider`. In the process I moved all `changeCss` calls (style/class theming adjustments) to the respective SCSS rules. I had to create a `_global.scss` file for adjustments that either were to specific or had no place in the remaining SCSS files.

Since this is a pretty big PR I will assign more reviewers than usual.

#### Ticket Link
n/a

#### Release Note
```release-note
NONE
```


components/compass_theme_provider/compass_theme_provider.tsx
```diff
@@ -1,11 +1,78 @@
...
+
+        const xmlHTTP = new XMLHttpRequest();
+
+        xmlHTTP.open('GET', cssPath, true);
+        xmlHTTP.onload = function onLoad() {
+            link.href = cssPath;
+
+            if (UserAgent.isFirefox()) {
+                link.addEventListener('load', () => {
+                    // changeCss('code.hljs', 'visibility: visible');
```
hmhealey, line 38: Are these intentionally commented out? I don't remember the reason for them to begin with

components/compass_theme_provider/compass_theme_provider.tsx
```diff
@@ -20,6 +87,135 @@ const CompassThemeProvider = ({theme, children}: Props): JSX.Element | null => {
         noFontFaces: true,
     });
 
+    //     if (!UserAgent.isFirefox() && !UserAgent.isInternetExplorer() && !UserAgent.isEdge()) {
```
hmhealey, line 90: Same here

components/compass_theme_provider/compass_theme_provider.tsx
```diff
@@ -20,6 +87,135 @@ const CompassThemeProvider = ({theme, children}: Props): JSX.Element | null => {
...
+    //         changeCss('body.app__body ::-webkit-scrollbar-thumb', 'background:' + changeOpacity(theme.centerChannelColor, 0.4));
+    //     }
+
+    const CssVariables = createGlobalStyle`
+        :root {
```
hmhealey, line 95: 0/5, but perhaps we might want to consider only applying the CSS variables to the children of this component rather than at the root of the app. That would allow us to nest theme providers in the future if we ever wanted to theme one part of the app differently from others. For example, in a hackathon demo Devin and I did way back, we added a preview of a user's theme in the custom theme settings, and that was coloured using a nested ThemeProvider component

utils/utils.jsx
```diff
@@ -364,314 +361,12 @@ export function toTitleCase(str) {
...
-
-// given '#fffff', returns '255, 255, 255' (no trailing comma)
-function toRgbValues(hexStr) {
-    const rgbaStr = `${parseInt(hexStr.substr(1, 2), 16)}, ${parseInt(hexStr.substr(3, 2), 16)}, ${parseInt(hexStr.substr(5, 2), 16)}`;
-    return rgbaStr;
-}
-
 export function applyTheme(theme) {
-    if (theme.centerChannelColor) {
-        changeCss('.app__body .bg-text-200', 'background:' + changeOpacity(theme.centerChannelColor, 0.2));
```
hmhealey, line 379: :tada:

components/compass_theme_provider/compass_theme_provider.tsx
michelengelen: I suppose it was to prevent flickering of the code blocks, when the theme is in the process of changing. I left it commented out on purpose so I can re-apply it when it proves to be wrong or ugly! :D

components/compass_theme_provider/compass_theme_provider.tsx
michelengelen: that is on purpose as well to see if it adds a regression or not

components/compass_theme_provider/compass_theme_provider.tsx
michelengelen: that would simplify things ... we can still change that when we need to ... its pretty much up to how you want to handle that in the future

components/compass_theme_provider/compass_theme_provider.tsx
hmhealey: Just be aware of if changing these styles causes elements to shift on the page since that would cause scrolling issues. I don't think visibility changes the position, but that's just something I'd like to mention.

sass/components/_emoticons.scss
```diff
@@ -117,6 +118,7 @@
         margin-top: 10px;
     }
 
+    /* this might be obsolete (I cannot find usages anywhere) */
```
larkox, line 121: If there are not references in the code, can we remove it?

sass/components/_popover.scss
```diff
@@ -121,36 +121,37 @@
         }
     }
 
+    /** TODO: are these rules needed? They are hidden in the rules above **/
```
larkox, line 124: I think this could be better organized activating some of the linter rules that are not activated. Any thoughts on that?

sass/components/_popover.scss
```diff
@@ -301,7 +307,7 @@
 .user-popover__role {
     padding: 3px 5px;
     margin-left: 10px;
-    background-color: $gray;
+    background: rgba(var(--center-channel-color-rgb), 0.3);
```
larkox, line 310: Probably out of the scope of this PR, but would it be possible to create some custom lint rule that takes care of enforcing these.

sass/components/_post.scss
```diff
@@ -2197,3 +2251,47 @@
...
     }
 }
+
+@media (min-width: 768px) {
+    .post.post--compact.same--root.post--comment .post__content {
+        border-color: rgba(var(--center-channel-color-rgb), 0.2);
+    }
+
+    .post.a11y--active {
+        background: rgba(var(--center-channel-color-rgb), 0.08);
```
larkox, line 2261: Why do we need to change the background color depending on the width? Are the styles not coherent between mobile view and web view?

sass/layout/_headers.scss
```diff
@@ -228,6 +243,7 @@
     .heading {
         overflow: hidden;
         margin: 0 0 0 2px;
+        color: var(--center-channel-color);
```
larkox, line 246: Why are we adding these many "new" color all over the components? What is the change that make this required?

sass/responsive/_global.scss
```diff
@@ -0,0 +1,66 @@
...
+        fill: var(--center-channel-color);
+    }
+
+    .table > tbody > tr > td,
+    .table > thead > tr > th,
+    .post .MenuWrapper .dropdown-menu button {
+        border-color: rgba(var(--center-channel-color-rgb), 0.12);
+    }
+
+    /** IMO these rules feel like overkill **/
```
Willyfrog, line 52: then we should remove them?

components/compass_theme_provider/compass_theme_provider.tsx
```diff
@@ -1,11 +1,78 @@
...
+        xmlHTTP.onload = function onLoad() {
+            link.href = cssPath;
+
+            if (UserAgent.isFirefox()) {
+                link.addEventListener('load', () => {
+                    // changeCss('code.hljs', 'visibility: visible');
+                }, {once: true});
+            } else {
+                // changeCss('code.hljs', 'visibility: visible');
+            }
```
Willyfrog, line 42:
```suggestion
            } 
```

Willyfrog (changes requested):

components/compass_theme_provider/compass_theme_provider.tsx
```diff
@@ -1,11 +1,78 @@
...
+        }
+    });
+
+    const link = document.querySelector<HTMLLinkElement>('link.code_theme');
+    if (link && cssPath !== link.href) {
+        // changeCss('code.hljs', 'visibility: hidden');
+
+        const xmlHTTP = new XMLHttpRequest();
+
+        xmlHTTP.open('GET', cssPath, true);
```
larkox, line 32: What is this needed for?

components/compass_theme_provider/compass_theme_provider.tsx
larkox: Can we add anything to make sure we remove this piece of code if we don't find any regression? (i.e. before we merge)

larkox (approved):
I am not comfortable enough with this piece of code to give proper feedback, so I will approve it from my side.

That being said, I left some comments to understand better what is being changed and some minor feedback.

sass/components/_emoticons.scss
michelengelen: probably ... the thing is: sometimes these styles are used in a very specific place somewhere in the application ... soooo ... it might not be safe to remove ... unless we are fine by adding a regression (even if it is a small one) and fix it later if it appears! :D

components/user_settings/display/user_settings_theme/user_settings_theme.jsx
```diff
@@ -101,7 +100,7 @@ export default class ThemeSetting extends React.PureComponent {
         let themeChanged = this.state.theme.length === theme.length;
         if (!themeChanged) {
             for (const field in theme) {
-                if (theme.hasOwnProperty(field)) {
+                if (Object.hasOwnProperty(field)) {
```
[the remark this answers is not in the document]
michelengelen: oh, there is actually a small bug here ... normally you would do it like this: `Object.prototype.hasOwnProperty.call(theme, field);`

I am going to fix that!

sass/components/_popover.scss
michelengelen: well ... that would be helpful ... and there probably is ... at least for eslint you can define a regex rule that checks for stuff like this

sass/layout/_headers.scss
michelengelen: the change is that it was previously done with the `changeCss` function inside `applyTheme`

sass/components/_post.scss
michelengelen: good question .. will investigate

sass/routes/_settings.scss
```diff
@@ -374,11 +390,11 @@
                     .theme-elements__body {
                         @include border-radius(0 0 3px 3px);
                         @include legacy-pie-clearfix;
-                        @include alpha-property(background-color, $white, 0.05);
```
nevyangelova, line 393: Do we need to keep using these mixins? Unless we need some type of calculation I believe we can replace all of them with CSS props

sass/routes/_activity-log.scss
```diff
@@ -17,7 +17,7 @@
     display: table;
     width: 100%;
     padding: 15px 0;
-    border-top: 1px solid $light-gray;
```
nevyangelova, line 20: Same as these variables, can we safely remove them after using the rgb ones?

sass/responsive/_global.scss
```diff
@@ -0,0 +1,66 @@
...
+        border-color: rgba(var(--center-channel-color-rgb), 0.1);
+        background: rgba(var(--center-channel-color-rgb), 0.05);
+    }
+
+    .svg-text-color,
+    .icon--body {
+        fill: var(--center-channel-color);
+    }
+
+    #post-create,
```
nevyangelova, line 33: I am generally not a fan of having random selectors bunched up in a global style as it makes it very difficult to trace where the change is coming from when debugging. Can we keep the global style relatively global? And split up component specific styles in the module of that component.

sass/components/_tutorial.scss
```diff
@@ -340,12 +341,12 @@ $tip-index: 1000;
         width: 9px;
         height: 9px;
         margin: 0 5px;
-        background: $black;
+        background: var(--center-channel-color);
 
         &.active {
             @include opacity(1);
 
-            background: $primary-color;
```
nevyangelova, line 348: Nit: remove the unused variables

sass/components/_post.scss
```diff
@@ -1775,12 +1810,12 @@
             top: 1px;
             display: inline-block;
             margin: 0 1px 0 5px;
-            fill: inherit;
+            fill: currentColor;
```
nevyangelova, line 1813: nit: I prefer the use of a color variable here as well as its more consistent.

sass/components/_forms.scss
```diff
@@ -118,7 +118,7 @@
     &.radio-inline,
     &.checkbox-inline {
         label {
-            color: $red;
```
nevyangelova, line 121: nit: can we remove all of those safely?

sass/components/_day-picker.scss
```diff
@@ -69,4 +69,19 @@
     .DayPicker-Day--selected:not(.DayPicker-Day--disabled):not(.DayPicker-Day--outside) {
         background-color: var(--button-bg);
     }
+
+    &:not(.DayPicker--interactionDisabled) {
+        .DayPicker-Day {
```
nevyangelova, line 74: I think these can be refactored for sure.

components/compass_theme_provider/compass_theme_provider.tsx
```diff
@@ -20,6 +87,135 @@ const CompassThemeProvider = ({theme, children}: Props): JSX.Element | null => {
...
+            --away-indicator: ${theme.awayIndicator};
+            --dnd-indicator: ${theme.dndIndicator};
+            --mention-bg: ${theme.mentionBg};
+            --mention-color: ${theme.mentionColor};
+            --center-channel-bg: ${theme.centerChannelBg};
+            --center-channel-color: ${theme.centerChannelColor};
+            --new-message-separator: ${theme.newMessageSeparator};
+            --link-color: ${theme.linkColor};
+            --button-bg: ${theme.buttonBg};
+            --button-bg-hover: ${rgbToHex(changeColor(theme.buttonBg, -0.15))};
```
nevyangelova, line 147: Can we pass the hex color here instead so we avoid having to convert? I am generally against having too many color conversion functions.

components/compass_theme_provider/compass_theme_provider.tsx
```diff
@@ -1,11 +1,78 @@
...
+    for (i = 0; i < 3; i++) {
+        c = parseInt(hex.substr(i * 2, 2), 16);
+        c = Math.round(Math.min(Math.max(0, c + (c * lum)), 255)).toString(16);
+        rgb += ('00' + c).substr(c.length);
+    }
+
+    return rgb;
+};
+
+const stripRGBDeclarator = (color: string) => color.slice(4, -1);
```
nevyangelova, line 73: what's the use case for stripping the declarator? Is it possible to agree on one approach when using Rgb variables so we don't need these extra functions?

nevyangelova (changes requested):
Thanks for handling this! It's a very significant change in the right direction. A few minor suggestions:

components/user_settings/display/user_settings_theme/user_settings_theme.jsx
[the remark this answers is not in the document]
larkox: And why not how it was originally done? (i.e. `theme.hasOwnProperty(field)`)

sass/components/_emoticons.scss
larkox: Does the string `nav-tabs`appear anywhere in the code? If not, it is probably safe to remove it. This is something we have to be more careful with "composed classes" (like `item` and `item--selected`), but the name of this class seems safe to assume that if it does not exist it is not being used.

sass/components/_popover.scss
larkox: @hmhealey Should we create a ticket for web platform? Or what do you think about this?

sass/routes/_settings.scss
michelengelen: good question ... IMO this is certainly possible, but out of scope for this PR (its big enough as it is already)

sass/routes/_activity-log.scss
michelengelen: same as above! :D

sass/responsive/_global.scss
michelengelen: yes, will try to do that

sass/components/_tutorial.scss
michelengelen: same comment as above ... we currently do not know what impact that would have

sass/components/_post.scss
michelengelen: yes, it is more descriptive, but this way you are less prone to regressions, when the color changes on the parent ... it is kinda a stylistic choice

sass/components/_forms.scss
michelengelen: same as above

components/compass_theme_provider/compass_theme_provider.tsx
michelengelen: the problem is, that the `buttonBg` color is being altered programmatically and the `changeColor` helper returns a rgb string ... so no (atm)

components/compass_theme_provider/compass_theme_provider.tsx
michelengelen: the function I use for converting hex to rgb and it returns a full rgb string (`rgb(255, 255, 255)`) and since we only want the numbers i just strip the `rgb(` and closing `)` to get that.

There could be a better solution, but not without refactoring the compass component helper

sass/components/_day-picker.scss
```diff
@@ -69,4 +69,19 @@
     .DayPicker-Day--selected:not(.DayPicker-Day--disabled):not(.DayPicker-Day--outside) {
         background-color: var(--button-bg);
     }
+
+    &:not(.DayPicker--interactionDisabled) {
```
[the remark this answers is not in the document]
michelengelen: I don't think so, but I am not very familiar with how the `DayPicker` component is structured ... it does look weird though, I have to admit that.

Maybe we can create a follow-up ticket for this?

Thoughts @furqanmlk ?

sass/components/_emoticons.scss
michelengelen: I will remove them for now and see if regressions pop up from it to fix!

sass/components/_popover.scss
michelengelen: I just found a stylelint rule that could enforce that by setting specific properties and a regex for the value:

https://github.com/AndyOGo/stylelint-declaration-strict-value

michelengelen: @larkox follow-up ticket here: [MM-40158](https://mattermost.atlassian.net/browse/MM-40158)

sass/components/_post.scss
michelengelen: I cannot find a good reason for that ... but since it was done this way before I would opt for leaving it that way for now

sass/routes/_settings.scss
larkox: Can we create tickets for those? Or are there already?

components/compass_theme_provider/compass_theme_provider.tsx
michelengelen: I did move this specific rule to the `_global.scss` file

components/compass_theme_provider/compass_theme_provider.tsx
michelengelen: I just tried it and it seems like the change is quite quick ... I removed them and if there are any regressions we can re-enable that

nevyangelova (approved):
LGTM just needs some conflict resolving

Willyfrog (approved):

components/user_settings/display/user_settings_theme/user_settings_theme.jsx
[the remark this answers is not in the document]
larkox: I still don't understand what does this change accomplish.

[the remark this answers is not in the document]
michelengelen: @larkox this is a security measurement. The explanation from eslint is a pretty good read: https://eslint.org/docs/rules/no-prototype-builtins

sass/components/_day-picker.scss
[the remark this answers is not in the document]
larkox: Follow-up ticket for this sounds good to me.

michelengelen:
> Are there tickets for all the raised issues that are out of the scope for this PR?

@larkox I still need to add one for removing the mixins.

hmhealey (approved):

jgilliam17:
@michelengelen There is an error on the test server, happens on login, prevents channels from loading
Screen Shot 2022-01-06 at 10 52 30 AM

michelengelen:
Ok, I know what's happening and will fix that tomorrow! 👍🏻

michelengelen:
@jgilliam17 it is working now. Sorry for the inconvenience there! 🙇🏼

jgilliam17:
Thanks @michelengelen 
Seeing few theming issues 
1. Create a team modal buttons invisible on hover, also seen on date selector, probably same issue
Screen Shot 2022-01-07 at 1 36 56 PM
Screen Shot 2022-01-07 at 1 57 20 PM

2. Team icon selected state and also channel selected state 
Screen Shot 2022-01-07 at 2 23 10 PM

3.  When changing theme on one team, some styles are affected on other team themes, can be cleared with refresh.
Steps: Have few teams; Open one team and apply quartz theme and save, switch to other team and observe sidebar icons and channel names
Screen Shot 2022-01-07 at 2 32 35 PM

michelengelen:
@jgilliam17 all of the findings from above should now be fixed

jgilliam17:
Thanks @michelengelen 
1. is partially fixed, create a team button is no longer invisible, but date still disappears on hover.
3. is not fixed for me.

michelengelen:
Hi @jgilliam17 ... it seems as if something (I am suspecting a plugin here) is writing old styles onto the `html` element. That is what causes these issues ... I will try and lock down those changes and see if i can fix that

michelengelen:
Hi @jgilliam17 ... I found the culprit: It is the boards plugin that causes the issue and writes some variables onto the `html` element. I will try to find someone from the boards team and see if we can fix that issue asap!

michelengelen:
> @michelengelen Just following up - has the boards issue been fixed?

afaik @jespino is working on that.

michelengelen:
@jgilliam17 all issues should now be resolved ... I merged current master in as well

jgilliam17:
@michelengelen Sorry about delay, I missed the mention 
1. dates on hover are still misaligned
Screen Shot 2022-03-10 at 9 51 33 AM

2. Channel header items are using the wrong theme, system post text, textbox icons, date line, some borders. etc. It is fixed by refreshing after changing the theme.
Screen Shot 2022-03-10 at 9 55 57 AM

michelengelen:
@jgilliam17 
1. I will have a look into
2. is still being influenced by boards plugin ... I am going to take a look if i can get in a PR for them to fix that.

jgilliam17:
Removing myself as reviewer - @michelengelen  please add me again when ready for QA - thanks 🙂

lieut-data:

PR #9358, opened 2021-11-11
Mm 38799 invite to team

neallred:
#### Summary
Convert invite to team modal from a fullscreen modal to an in-screen modal with a backdrop. Note that the possible treatments are the in-screen modal with radio buttons and the in-screen modal with the toggle. We are not keeping the fullscreen modal as a control in this test.

This is a long PR. 8K of the additions are adding the an svg for the "no permissions" screen (see other callouts section). Some of it is false LOC inflation from converting from js to ts, and git counting it as sadded deleted files instead of changed files. About 2K LOC deletions are from removing the fullscreen modal, and the new modal is also about 2K LOC. I could maybe break out up to 1K LOC worth of typescript conversions, but then that would be one more PR with additional risks of having to make sure the newly typescripted files play nice with the existing fullscreen modal, and that they work with the new files when I convert to in-screen modal.

Some caveats:

* I wasn't able to find a slider like the one shown in the designs. We had previously discussed it was acceptable to use something close like it if it existed. If someone does know of an existing slider more like the designs, I'm happy to see if its doable to use it instead.
* The Radio component that matches the designs is in Compass. It needs some further work before it can support labels with more than one color of text. I made a first effort using and changing it, but I don't think we'll be ready to use that Radio component until after we hope to merge this PR (Nov 15 for the Nov 24 release). So I opted to use the Radio group already in use in other places in webapp.

Things still needing decision:

* In the fullscreen modal, we show a confirm modal on top of the fullscreen modal if the user exits out with unsaved progress. (How) should we prevent unsaved progress in the new experience? For example, if they've entered 10 users and 5 channels, and the accidentally click outside the modal and did not mean to, that is a fair amount of lost work. What I implemented as the easiest option was that if they have users or channels inputted or text in the search box, then we do not close the modal on outside clicks, only on pressing escape key or clicking the close button on the top right. More options for preventing lost work are documented in the Jira ticket.

Other callouts:

I don't think there's ever a hard server error in our scenario; instead, they all just get shown as "invite not sent". So I didn't implement a hard server error screen. However, I did see that user's **might** be able to invite members and guests, or that they might only be able to invite members. It's not clear to me whether it can happen that they can't invite either members or guests, so I implemented a "can't invite either" screen that looks similar to the hard server error screen, in which the only choice is to close the modal. Let me know what you think! If someone knows for sure that it can NEVER happen that a user doesn't have permissions to invite users and that all entry points to this modal are hidden when they have no permissions to invite, then we might not need this screen. I haven't looked into how to force this scenario, but for the time being if you type "nopermissions" into the users search textarea, you will get the "no permissions" screen.

#### Ticket Link
Jira: https://mattermost.atlassian.net/browse/MM-38799
Figma: https://www.figma.com/file/fheJeoKw0LBTlvmHpAqpls/%E2%9C%85Invite-to-channel?node-id=27%3A208034

#### Related Pull Requests
- [Server PR](https://github.com/mattermost/mattermost-server/pull/18987)

#### Screenshots

Several of the interactions:
https://user-images.githubusercontent.com/13738432/141336234-1768c8cc-d8a7-49d1-8d2a-3aa18d6f4047.mp4

View when only member invites are enabled:
only-invite-members

no permissions view:
no-permissions

#### Release Note
```release-note
* Convert invite to team modal from fullscreen modal to in-screen modal.
* Track when invite users to team button is clicked.
```


stevemudie (approved):
@neallred this looks good and works as expected. Nice job!

I'm approving it from a QA perspective as is but did note that the wording when inviting a person as a guest (when the toggle is used) feels a bit scary. Since guests can later be transformed into full members what's the thinking behind the wording that says that it can't be "undone"?

image

actions/invite_actions.test.ts
```diff
@@ -0,0 +1,538 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
neallred, line 1: In the new and old version of the test file, the only real change is that the return format for the actions is now wrapped in an object with a `data` key. This accompanying change in result format was done in the implementation to make the actions compatible with `ActionResult`, which in turn made converting all the places that use this action to Typescript easier.

components/invitation_modal/invite_view.tsx
```diff
@@ -0,0 +1,339 @@
...
+        return;
+    }
+    if (modalBody.style.minHeight) {
+        modalBody.style.minHeight = '';
+    }
+}
+
+const FOOTER_HEIGHT = 63;
+
+const ensureInvitesVisible = debounce((options: UserProfile[]) => {
```
neallred, line 88: `ensureInvitesVisible` and `clearMinHeight` was the best way I could think of to overcome the `overflow: hidden` hidden restriction on modal body content not being able to display outside of modal body. In the fullscreen modal, this was not a problem because the modal had the whole screen to work with. It makes some effort to not make the modal taller than the screen.

clearMinHeight *should* be invoked everywhere so we don't get a goofily long looking modal once the dropdown closes or has no options, but extra scrutiny is warranted here.

components/invitation_modal/invite_view.tsx
```diff
@@ -0,0 +1,339 @@
...
+    if (options.length < 2) {
```
neallred, line 89: It seemed like across desktop screen sizes, if there were 2 or less options, the height adjustment wasn't needed.

components/invitation_modal/invite_view.tsx
```diff
@@ -0,0 +1,339 @@
...
+                </div>
+                <UsersEmailsInput
+                    {...errorProperties}
+                    usersLoader={(value: string, cb: (users: UserProfile[]) => void) => {
+                        if (value === '') {
+                            clearMinHeight();
+                        }
+                        return props.usersLoader(value, cb);
+                    }}
+                    onUsersLoad={ensureInvitesVisible}
```
neallred, line 279: By invoking this once the data is loaded, it should avoid the problem of failing to adjust size because of slow networks.

components/invitation_modal/invite_view.tsx
```diff
@@ -0,0 +1,339 @@
...
+        const menuRect = menu.getBoundingClientRect();
+
+        if (menuRect.bottom > bodyRect.bottom) {
+            let betterHeight = menuRect.bottom - bodyRect.top;
+            if ((bodyRect.top + betterHeight + FOOTER_HEIGHT) > window.innerHeight) {
+                betterHeight = window.innerHeight - bodyRect.top - FOOTER_HEIGHT;
+            }
+            modalBody.style.minHeight = `${betterHeight}px`;
+        }
+    }, 125);
```
neallred, line 114: A guess/compromise that for most devices, this will be enough time to render given that the data has already been fetched from the network. In the worst case, we sometimes fail to make this tall enough on slower devices. As an enhancement, we could build in retry logic, but wanted to avoid that if we can.

components/invitation_modal/invitation_modal.tsx
```diff
@@ -0,0 +1,396 @@
...
+                usersEmailsSearch,
+            },
+        }));
+    }
+
+    getBackdrop = () => {
+        // 'static' means backdrop clicks do not close
+        // true means backdrop clicks do close
+        // false means no backdrop
+        if (this.state.view === View.RESULT || (!this.props.canAddUsers && !this.props.canInviteGuests) || this.state.invite.usersEmailsSearch === 'nopermissions') {
```
neallred, line 308: This and other `'nopermissions'` related code is a hack to make it easy to see this view for testing purposes. It should be removed prior to merging.

components/widgets/inputs/channels_input.tsx
```diff
@@ -67,8 +76,8 @@ export default class ChannelsInput extends React.PureComponent {
         }
     }
 
-    optionsLoader = (input, callback) => {
-        const customCallback = (options) => {
+    optionsLoader = (_input: string, callback: (options: Channel[]) => void) => {
```
marianunez, line 79: TIL about the underscore in variables 👍

e2e/cypress/integration/team_settings/invite_user_to_closed_team_spec.js
```diff
@@ -63,30 +57,31 @@ describe('Team Settings', () => {
...
         cy.wait(TIMEOUTS.HALF_SEC);
-        if (isLicensed) {
-            // # Click "Invite members"
-            cy.findByTestId('inviteMembersLink').should('be.visible').click();
-        }
+
+        // if (isLicensed) {
+        //     // # Click "Invite members"
+        //     cy.findByTestId('inviteMembersLink').should('be.visible').click();
+        // }
```
marianunez, line 84: Is his intended?

marianunez (approved):
Awesome work here, great tests and typescript migration! Left some comments but all minor and non-blocking!

marianunez:
> I don't think there's ever a hard server error in our scenario; instead, they all just get shown as "invite not sent". So I didn't implement a hard server error screen.

There is an open ticket (and PR) that has been around for a bit that should address this https://mattermost.atlassian.net/browse/MM-25763 I've moved the ticket into our team and we can work on that separately. 

> . (How) should we prevent unsaved progress in the new experience? 

I tried your approach here on only allowing to close on click if there is no changes in progress and seemed good to me. 👍 

One minor UX detail I did notice was that there seems to be extra space between the title and the `To:` section below it: 

Screen Shot 2021-11-12 at 11 26 54 PM

components/widgets/inputs/users_emails_input.tsx
```diff
@@ -1,14 +1,14 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import PropTypes from 'prop-types';
 import React from 'react';
 import {FormattedMessage} from 'react-intl';
-import AsyncSelect from 'react-select/lib/AsyncCreatable';
-import {components} from 'react-select';
+import {components, /*ValueType, ActionMeta,*/ InputActionMeta, FormatOptionLabelMeta} from 'react-select';
```
pablovelezvidal, line 6: the commented ones could be deleted?

components/widgets/inputs/channels_input.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+                    } as Channel,
+                    {
+                        id: 'test-channel-2',
+                        type: 'P',
+                        display_name: 'test channel 2',
+                    } as Channel,
+                ]}
+            />,
+        );
+        expect(wrapper).toMatchInlineSnapshot(`
```
pablovelezvidal, line 35: just curious why do we need to use inline snapshots here for this test?

components/invitation_modal/result_table.tsx
```diff
@@ -0,0 +1,162 @@
...
+import {isGuest} from 'mattermost-redux/utils/user_utils';
+
+import EmailIcon from 'components/widgets/icons/mail_icon';
+import AlertIcon from 'components/widgets/icons/alert_icon';
+import GuestBadge from 'components/widgets/badges/guest_badge';
+import BotBadge from 'components/widgets/badges/bot_badge';
+import Avatar from 'components/widgets/users/avatar';
+
+import './result_table.scss';
+type InviteNotSent = {
```
pablovelezvidal, line 18: not sure why linter does not catch blank line after import statements, maybe not mandatory but definetely looks more readable :)

components/invitation_modal/result_table.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+let props: Props = {
+    sent: true,
+    rows: [],
+};
+type WithIntlProps = {
+    children: React.ReactNode | React.ReactNodeArray;
+}
+
+const WithIntl = (props: WithIntlProps) => {
```
pablovelezvidal, line 26: we already have a function called mountWithIntl for these scenarios `import {mountWithIntl} from 'tests/helpers/intl-test-helper';
`

components/invitation_modal/invite_view.tsx
pablovelezvidal: nit: other option would be show a loading spinner and once data is loaded (check in componentUpdate lifecycle), show the modal content data. Is just me not being a fan of using timeOuts :p

components/invitation_modal/invite_view.tsx
pablovelezvidal: or maybe as a self descriptive constant name?

components/invitation_modal/invite_view.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+});
+
+let props = defaultProps;
+
+describe('InviteView', () => {
+    beforeEach(() => {
+        props = defaultProps;
+    });
+
+    it('shows invite as UI when user can choose to invite guests or users', () => {
```
pablovelezvidal, line 90: what does invite as UI stands for

components/invitation_modal/invite_view.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+import CompassThemeProvider from 'components/compass_theme_provider/compass_theme_provider';
+
+import InviteAs, {As} from './invite_as';
+import InviteView, {Props} from './invite_view';
+
+type WithIntlProps = {
+    children: React.ReactNode | React.ReactNodeArray;
+};
+const WithProviders = (props: WithIntlProps) => {
```
pablovelezvidal, line 21: maybe set this as a helper test component?

pablovelezvidal: `tests/helpers/` folder would be a good place

components/invitation_modal/invite_as.tsx
```diff
@@ -0,0 +1,118 @@
...
+import React from 'react';
+
+import {FormattedMessage} from 'react-intl';
+
+import {InviteToTeamTreatments} from 'mattermost-redux/constants/config';
+
+import FormattedMarkdownMessage from 'components/formatted_markdown_message';
+import RadioGroup from 'components/common/radio_group';
+
+import './invite_as.scss';
```
pablovelezvidal, line 13: good practice to set the css imports as last group import

components/invitation_modal/invitation_modal.tsx
```diff
@@ -0,0 +1,396 @@
...
+            return true;
+        }
+
+        const emptyInvites = this.state.invite.usersEmails.length === 0 && this.state.invite.usersEmailsSearch === '';
+        if (this.state.invite.as === As.MEMBER && !emptyInvites) {
+            return 'static';
```
pablovelezvidal, line 314: should this value be store in a constant instead of plain string?

components/invitation_modal/invitation_modal.tsx
```diff
@@ -0,0 +1,396 @@
...
+    cloudUserLimit: string;
+    canAddUsers: boolean;
+    canInviteGuests: boolean;
+    intl: IntlShape;
+}
+
+export const View = {
+    INVITE: 'INVITE',
+    RESULT: 'RESULT',
+} as const;
```
pablovelezvidal, line 69: just curious, why the `as const`?

components/invitation_modal/add_to_channels.tsx
```diff
@@ -0,0 +1,139 @@
...
+    currentChannelName: string;
+    titleClass?: string;
+}
+
+const RENDER_TIMEOUT_GUESS = 100;
+
+export default function AddToChannels(props: Props) {
+    useEffect(() => {
+        if (props.customMessage.open) {
+            setTimeout(() => {
```
pablovelezvidal, line 53: some time ago in a different project, I wrote a custom hook for checking if an element was already rendered in the DOM. it used IntersectionObserver function. Let's go this way, but let's talk about recreating this custom hook again (I don't longer have access to that code :p) for possible future scenarios like this.

pablovelezvidal (approved):
Excelcious work, thanks! @neallred. Most of my comments are either suggestions or questions, nothing blocking the merge for this huge PR. :shipit:

components/invitation_modal/invite_view.tsx
neallred: :+1: on both

actions/invite_actions.ts
```diff
@@ -94,9 +106,9 @@ export async function sendGuestInviteForUser(dispatch, user, teamId, channels, m
     try {
         await dispatch(addUsersToTeam(teamId, [user.id]));
         for (const channel of channels) {
-            const member = members && members[channel] && members[channel][user.id];
+            const member = members && members[channel.id] && members[channel.id][user.id];
             if (!member) {
-                await dispatch(joinChannel(user.id, teamId, channel)); // eslint-disable-line no-await-in-loop
+                await dispatch(joinChannel(user.id, teamId, channel.id, channel.name)); // eslint-disable-line no-await-in-loop
```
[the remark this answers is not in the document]
neallred: yep! This was something that Typescript highlighted as an error during the conversion. [`joinChannel`](https://github.com/mattermost/mattermost-webapp/blob/master/packages/mattermost-redux/src/actions/channels.ts#L653) explicitly asks for the id and name.

components/common/hooks/useCopyText.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const DEFAULT_COPY_TIMEOUT = 4000;
+
+export default function useCopyText(options: CopyOptions): CopyResponse {
+    const [copiedRecently, setCopiedRecently] = useState(false);
+    const [copyError, setCopyError] = useState(false);
+    const timerRef = useRef<NodeJS.Timeout | null>(null);
+
+    let successCopyTimeout = DEFAULT_COPY_TIMEOUT;
+    if (options.successCopyTimeout !== null && options.successCopyTimeout !== undefined) {
```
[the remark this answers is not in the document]
neallred: I want to support 0 here. I could make that clearer and shorter with `options.successCopyTimeout || options.successCopyTimeout === 0`

components/common/radio_group.tsx
```diff
@@ -7,7 +7,7 @@ import Badge from 'components/widgets/badges/badge';
 
 type RadioGroupProps = {
     id: string;
-    values: Array<{ key: string; value: string}>;
+    values: Array<{ key: React.ReactNode | React.ReactNode; value: string}>;
```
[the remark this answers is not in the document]
neallred: Oops! That was supposed to be `React.ReactNode | React.ReactNodeArray`.

components/invitation_modal/invitation_modal.tsx
```diff
@@ -0,0 +1,396 @@
...
+            }));
+        }
+    }
+
+    invite = async () => {
+        const inviteAs = this.state.invite.as;
+        if (inviteAs === As.MEMBER && this.props.isCloud) {
+            trackEvent('cloud_invite_users', 'click_send_invitations', {num_invitations: this.state.invite.usersEmails.length});
+        }
+        trackEvent('invite_users', 'click_invite');
```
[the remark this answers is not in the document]
neallred: @marianunez Consolidating sounds good to me. Who should I reach out to make sure I'm not breaking any dashboards or reports or things like that?

components/invitation_modal/invitation_modal.tsx
```diff
@@ -0,0 +1,396 @@
...
+                this.props.currentTeam.id,
+                this.state.invite.inviteChannels.channels,
+                users,
+                emails,
+                this.state.invite.customMessage.open ? this.state.invite.customMessage.message : '',
+            );
+            invites = result.data;
+        }
+
+        if (this.state.invite.usersEmailsSearch !== '') {
```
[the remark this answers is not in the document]
neallred: This is existing logic from porting from fullscreen modal. There, the intent was that if there was an invite that they didn't finish typing, and the clicked invite, we would show this as someone that failed to be invited, because it was never selected as a valid user or email option.

components/invitation_modal/invitation_modal.tsx
neallred: Not sure how it would be in that state, either, but wanted these protections in case something goes wrong with the guards.

components/invitation_modal/invitation_modal.tsx
```diff
@@ -0,0 +1,396 @@
...
+        this.setState((state: State) => ({
+            ...state,
+            invite: {
+                ...state.invite,
+                usersEmailsSearch,
+            },
+        }));
+    }
+
+    getBackdrop = () => {
```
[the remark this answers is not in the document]
neallred: That's a good call. I think so.

components/widgets/inputs/channels_input.test.tsx
neallred: This was a js -> ts file conversion that got counted as an add and delete. Not sure why the original author made it inline versus its own file. In general I don't like snapshots because I think they do a really poor job capturing what you're asserting on.

components/invitation_modal/invitation_modal.tsx
neallred: As a way to narrow the types from being strings to being only the strings listed. Some more details:
https://rlee.dev/stop-misusing-typescript-string-enums
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

e2e/cypress/integration/team_settings/invite_user_to_closed_team_spec.js
neallred: I wasn't sure at first so I commented the code, but then forgot to address. Thinking more, it is intentional. This screen does not exist any more, and members is selected by default if it is an available option.

components/invitation_modal/invitation_modal.tsx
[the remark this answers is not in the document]
marianunez: Good point, we don't have anyone right now that could help us easily determine this. Let's keep the tracking the same as it currently is

components/invitation_modal/invitation_modal.tsx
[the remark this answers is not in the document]
marianunez: Thanks for clarifying! wasn't familiar with the modal implmentation

annelieseklein:
looks good @neallred ! 
1. I see Maria mentioned the spacing issue between the modal title and the "To". 
2. 2. There should also be 8px space between "Set a custom message" and the sub copy. (and subsequently) also 8px between Custom Message and the input field
3. There should also be 24px padding on each side of the input fields. 
4. The cancel/close icon at the top of the modal (and also for the custom message input field) should have a 56% opacity
5. The Invite button should be 24px away from the sub text above it
6. I notice when i have a lot of emails inputted and then add a custom message, the modal shifts a bit, I think to include the scroll bar but it also seems like it shifts a bit spatially.
Screen Shot 2021-11-15 at 10 50 14 AM

neallred:
@annelieseklein good finds! All the points have been addressed. For the modal moving spatially, I noticed there was a second scrollbar for the modal as a whole (as opposed to to the modal body) that showed up on the far right or the screen, hidden under the backdrop shadow. I was able to disable that scroll bar, so the spatial jump shouldn't happen any more.

annelieseklein:
@neallred Thanks! a few more:

1. In terms of modal placement, can we center the modal on the page
2. When we have a long list of autocomplete menu, can we overlay the menu on the modal (so as to not change the actual length of the modal) ? I can provide an updated design here, but a similar design is attached.
3. Eric found a black outline for the copy link click effect, which i didn't find. See attached
Image Pasted at 2021-11-15 13-34

Screen Shot 2021-11-15 at 2 26 58 PM

annelieseklein:
Also I noticed we have a disabled state for the CTA that is different from what we have in the figma, so we should align that as well.

neallred:
@annelieseklein @esethna could you take another look? I was able to:
    * place the modal on the center of the page
    * Have the autocomplete menu overlay the modal instead of expanding the modal to try and accomodate the autocomplete menu.
    * I didn't change the drop shadow, but I think what was seen was the shadow getting cut off because the modal was expanding, and not by enough.
    * fix the disabled state of the invite button
    * Remove the shadow on clicking the copy invite link as I saw it on my machine & browsers (Linux, Firefox & Chromium). Not sure if this addresses what Eric saw; I wasn't able to reproduce the exact shadow he shared.

annelieseklein:
Looks great @neallred ! One final thing - can we cap the max lines in the invite modal to 5 1/4 lines of invitees? Currently we are able to see a long list of emails. I'd like to cap it at 5 with a little peek of a 6th line 


Screen Shot 2021-11-16 at 11 41 12 AM

neallred:
Sure!

neallred:
@annelieseklein changed. Does this look right?
five-plus-lines

annelieseklein:
@neallred yep looks great!

annelieseklein (approved):

esethna (approved):
@annelieseklein @neallred a few thoughts, non-blocking: 

1. Still seeing the black outline on the copy button. @annelieseklein can you help define what is supposed to be the expected hover and click effects on those buttons for both states (before and after the user clicks it)
2. Is this text necessary? Seems kind of alarming to me.
image
3. Is the button using the correct color? Shouldn't it match the button color? 
image
4.  Spacing here looks off:
image

neallred:
Addressed 3 & 4. Open to addressing 1 & 2 but need to merge this so it has enough time for testing prior to release.

PR #9361, opened 2021-11-11
Reducing bundle size (related to languages from @formatjs) from 35.68MB to 31.14

jespino:
#### Summary
Reducing bundle size (related to languages from @formatjs) from 35.68MB to 31.14.

This is because we were including all the locales related to formatjs for all
the languages, and now we are only including the languages that we support in
the application.

#### Ticket Link
[MM-40066](https://mattermost.atlassian.net/browse/MM-40066)

#### Release Note
```release-note
NONE
```


hmhealey (approved):

hmhealey:
Are these polyfills being included in the main bundle? That code would've lead me to believe that they're lazy loaded, but I might be reading it wrong.

If they are in the main bundle, we might want to look at removing them entirely based on what our minimum browser requirements are since most browsers released in the past 2 years support them

https://caniuse.com/intl-pluralrules
https://caniuse.com/?search=intl-relativetimeformat

jespino:
@hmhealey yes, that is in the main bundle, we should verify what browsers do we support and remove the not needed polyfills.

larkox (approved):

jespino:
@hmhealey maybe we can just remove it entirely in this PR, let me do the change and send it to QA.

lindy65:
Hi @jespino - I've tested all languages in Chrome, Edge and Firefox according to instructions in the ticket and did not find any issues.  I've asked in the QA channel for someone on a mac to test in Safari.

I did notice that the languages in (Alpha) as well as Chinese do not have the string below translated but I'm sure a separate PR is required for that:

image

image

hmhealey:
Comparing the supported versions and those charts, it looks like our minimum versions of Safari and Edge aren't high enough, but they're also really old, so we should be able to update those as long as the PMs are okay with it. I'd say we bump the minimum version of Edge to 89 to match Chrome, and Safari to 14 which is about a year old at this point and supported back to Mojave, so it's probably safe as well.

jespino:
@hmhealey probably that means that certain things are not going to be translated in that version, but I guess is not going to crash, so I think is good enough.

@furqanmlk About the translations, some languages are not 100% translated so that is normal, it is not a code problem, is a problem related to the percentage of translated strings. And that is expected so I wouldn't even consider that a bug.

larkox (approved):

hmhealey:
I'd prefer if we bumped up the minimum versions rather than potentially breaking things for users. I don't think that should be a problem, but we should make sure PMs are aware of it and docs are updated to reflect that

jespino:
@hmhealey ok, then, who would be the PM on charge of that? Sound like @wiersgallak, or at least she can now who can take care of that?

@wiersgallak to give you some context, we want to bump a bit the minimum supported version of Edge to 89, and Safari to 14.

jespino:
@furqanmlk Would you mind to test it on `Safari < 14` and `Edge < 89` and verify if that doesn't crashes? my expectations is to not crash, but also not translate properly the dates.

wiersgallak:
@esethna Please weigh in on this as it would be web platform related.  Thanks!

esethna:
@jespino @hmhealey @furqanmlk yes please share what happens if using Safari < 14 and Edge < 89.

I agree that if it is incorrectly translated dates then we can go ahead with bumping the minimum version. If the consequences are more impactful, like crashes or anything that makes Mattermost unusable, then we need to look into this further and determine the number of affected users.

As a note, Slack supports Safari 14+ (Released Sept 2020) and Edge 89+ (Released Feb 2021) as is proposed

hmhealey:
I'd also add that bumping the required version of Edge that high will mean we don't have to support pre-Chromium versions of Edge any more which would be nice since it's basically one less browser both devs and QAs need to deal with

jespino:
@jgilliam17  @furqanmlk any news about this?

furqanmlk:
> @jgilliam17  @furqanmlk any news about this?

I am re testing. Will update shortly

jespino:
@furqanmlk great, thanks! Please, pay special attention to the comment here: https://github.com/mattermost/mattermost-webapp/pull/9361#issuecomment-975741594

furqanmlk:
image

I tested using Safari 12. It's working as expected without any issues.

furqanmlk:
Tested Edge Version 88.0.705.68 (Official build) (64-bit) 
Looks good to me

https://user-images.githubusercontent.com/6909890/147675954-94046e49-949b-4fb2-adf8-0ebbaccb1713.mp4

furqanmlk (approved):
Tested using both Safari < 14 and Edge < 89.
No issue found

jespino:
@hmhealey PTAL

hmhealey (approved):
Awesome! Glad to know that, even if we're missing something on those versions, it still fails smoothly. That's a lot nicer than the issues we used to have with older versions of IE and Safari where they'd just fail to load the page because we were missing something

PR #9364, opened 2021-11-12
MM-33003 Remove a bunch of unneeded mixins used in the sass code

jespino:
#### Summary
A lot of mixins are used for backward CSS3 compatibility, but most of the
browsers already supports most of the CSS3 features, so we are no longer
needing that mixins. The resulting css code is longer without giving us
anything. Also I guess the compile takes a bit more because the need os pass
though the compass mixins.

Also, compass-mixins is unmaintained (not a big deal) so we can get rid of that
dependency if we remove all the mixins used from there.

**UPDATE**: I already removed all the mixins so I removed the compass-mixins dependency.

#### Some interesting numbers.

Before the changes, the css generate is 35020 lines (740K aprox) and is taking 1.6 seconds of CPU to execute.
After the changes, the css generated is 33884 lines (700K aprox) and is taking 1.4-1.5 seconds of CPU to execute.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33003

#### Release Note
```release-note
NONE
```


hmhealey:
I attached this to the ticket I had for upgrading or removing compass-mixins. Feel free to make a new ticket instead if you had one you wanted to make

hmhealey (approved):
Looks good to me! I'm sure most of that stuff is from back before we dropped support for IE10 at least

sass/base/_structure.scss
```diff
@@ -71,7 +71,7 @@ body {
 }
 
 .container-fluid {
-    @include legacy-pie-clearfix;
+    @include pie-clearfix;
```
deanwhillier, line 74: I'm not familiar with this, do you know what it is doing? What's the difference between `pie-clearfix` vs `clearfix`?

sass/routes/_settings.scss
```diff
@@ -271,9 +267,8 @@
 
                     .group--code {
                         select {
-                            @include appearance(none);
-
                             padding-right: 25px;
+                            -webkit-appearance: none;
```
deanwhillier, line 271: Non blocking: from what I can tell from caniuse, `appearance: none` is supported by all our supported browsers/versions meaning the `-webkit ...` version may not be needed anymore.

sass/routes/_settings.scss
```diff
@@ -365,23 +360,22 @@
...
-
                         max-height: 0;
                         padding: 0 0 0 24px;
                         margin: 0 20px 0 0;
+                        border-radius: 0 0 3px 3px;
                         overflow-y: hidden;
                         transition: all 0.4s ease-out;
 
+                        @include pie-clearfix;
+                        @include alpha-property(background-color, $white, 0.05);
```
deanwhillier, line 377: Non blocking: white is white 😛. Worth just replacing with this?

```suggestion
                        background-color: rgba(255, 255, 255, 0.05);
```

sass/utils/_mixins.scss
```diff
@@ -72,3 +71,18 @@
     width: fit-content;
     width: -moz-fit-content;
 }
+
+@mixin clearfix {
+    overflow: hidden;
+    *zoom: 1;
+}
+
+@mixin pie-clearfix {
```
deanwhillier, line 80: Okay, I see now.

sass/base/_structure.scss
deanwhillier: n/m, found! :)

deanwhillier (approved):
LGTM. Some non-blocking comments for now or later :)

jgilliam17:
Thanks @jespino 
Tested, looks good to merge.
- Verified in app styles - as before. E2E report looks good, no PR related failures. 
https://automation-dashboard.vercel.app/cycles/f866ff68-ae0c-4145-a7c5-889f5e0855fa

jgilliam17 (approved):

PR #9365, opened 2021-11-12
Fix for isMobile hitting window.innerWidth

jwilander:
#### Summary
The function `isMobile` in utils.js is called very often in many parts of the app. On app load it's called more than 2.5 thousand times. When a character is typed it's called 50-100 times. It's a very small, one-line function which was doing a basic comparison which normally wouldn't be a problem to call so frequently but it was accessing `window.innerWidth` each time it's called. `window.innerWidth` is one of the browser APIs that can cause style recalculation [per this gist](https://gist.github.com/paulirish/5d52fb081b3570c81e3a).

One of our customers is having typing performance issues that have profiles showing a significant amount of style recalculation during typing. This should hopefully help with that issue and improve performance for that customer and others facing similar issues while typing.

This PR cherry-pick's recent changes from @deanwhillier to store the isMobile view state in redux for selectors to use. The second commit in this PR leverages that to remove the constant calls to `window.innerWidth`.

The PR is opened against `release-5.38` because that's the version the customer is having the issue on. If this fix proves fruitful for them then we can bring these changes to `master` and other versions too.

#### Release Note

```release-note
General performance fix for loading the web application and typing.
```


hmhealey (approved):

amyblais:
FYI: v5.38 is going out of support as of today. https://docs.mattermost.com/upgrade/release-lifecycle.html (meaning we won't backport fixes to 5.38 anymore).

jwilander:
Good call @amyblais, I'll point it at release-5.39

jwilander:
@jgilliam17 can you do a quick smoke test to confirm that changing screen size to mobile dimensions and opening the app in a mobile browser work fine?

jgilliam17 (approved):
Thanks @jwilander 
Tested, looks good to merge.
- Verified changing screen size and opening in mobile webview works as expected.

PR #9312, opened 2021-11-02
[MM-39057] -  Remove Tips & Next Steps page

AGMETEOR:
#### Summary
This PR removes Tips and Next Steps page and all related code from the onboarding

#### Ticket Link
[MM-39057](https://mattermost.atlassian.net/browse/MM-39057)

#### Release Note
```release-note
NONE

```

PR #9367, opened 2021-11-12
Add lazy loading to some components

jespino:
#### Summary
Adding some lazy loading to certain components that are not shown always, like the invitation modal, purchase modal, or marketplace, for example.

The previous root bundle (loaded always): 6.5MB, 3.66MB gzipped
The new root bundle: 4.74MB, 2.52MB gzipped.

#### Ticket Link
Not yet

#### Release Note
```release-note
NONE
```


components/file_preview_modal/index.js
```diff
@@ -1,14 +1,16 @@
...
 import {connect} from 'react-redux';
 
 import {getConfig} from 'mattermost-redux/selectors/entities/general';
 import {getPost} from 'mattermost-redux/selectors/entities/posts';
+import {makeAsyncComponent} from 'components/async_load';
 
 import {canDownloadFiles} from 'utils/file_utils.jsx';
 
-import FilePreviewModal from './file_preview_modal';
+const FilePreviewModal = makeAsyncComponent(React.lazy(() => import('./file_preview_modal')));
```
M-ZubairAhmed, line 13: i am thinking do we still need to do this because of the new ModalController implementation, For FilePreviewModal PR is here https://github.com/mattermost/mattermost-webapp/pull/9334

jespino: This is not strictly related to that. This make the FilePreviewModal asyncronously loaded, that means all the javascript and dependencies are loaded in another bundle with the first time you render the component. That helps to speed up the initial load by not downloading and parsing javascript code that you are not using yet.

M-ZubairAhmed: Oh i thought It was being lazy loaded on the current implementation that it is always mounted in the component tree.

hanzei (commented):
Great work :+1: 

My only question is, if all of these components need a loading state now? Should we re-use something like 'LoadingScreen` for that?

jespino:
@hanzei yes, we can do that, but probably we should do that as part of the Suspense component already there. I don't think is strictly  needed, but but can be added later.

hanzei:
> @hanzei yes, we can do that, but probably we should do that as part of the Suspense component already there. I don't think is strictly needed, but but can be added later.

Sounds good to me. Should there be a ticket for that?

hanzei (approved):

components/dnd_custom_time_picker_modal/index.ts
```diff
@@ -10,8 +11,9 @@ import {setStatus} from 'mattermost-redux/actions/users';
 import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
 
 import {GlobalState} from 'types/store';
+import {makeAsyncComponent} from 'components/async_load';
 
-import DndCustomTimePicker from './dnd_custom_time_picker_modal';
+const DndCustomTimePicker = makeAsyncComponent('DndCustomTimePicker', React.lazy(() => import('./dnd_custom_time_picker_modal')));
```
hmhealey, line 16: Not something that needs to be done now, but I suppose we could eventually make this done automatically by the modal system somehow

hmhealey (approved):

jgilliam17:
@jespino can you please update branch. I'd like to run E2Es after update. Thanks.

jespino:
@jgilliam17 done!

jespino:
Thanks @jgilliam17 :ok_hand:

jgilliam17:
Thanks @jespino 
[E2E report](https://automation-dashboard.vercel.app/cycles/427388ed-b4df-46c6-924b-a47fd022d529) looks good, one issue 
3 tests are failing on "Add user to channel" it appears that multi-select list is displayed within the modal, see image
Screen Shot 2021-12-03 at 5 21 52 PM

jespino:
@jgilliam17 interesting, it is failing in my branch but not in master, so... the problem is related to how we do the lazy loading there. I'll take a look, thanks!

jespino:
@jgilliam17 Fixed! can you take another look?

jgilliam17:
Thanks @jespino Here's the new [E2E report](https://automation-dashboard.vercel.app/cycles/09714c03-49c8-4a8c-a4bb-3962f44d8cd4)
Re-tested manually and the issue with add member drop down not overflowing out of the modal is resolved
There is an issue with Getting Started "Complete your profile" section, on first load it is not expended. If you skip that step and expand "Setup notifications" and come back to Profile it will expand as expected.
Screen Shot 2021-12-06 at 2 26 03 PM

jespino:
@jgilliam17 really interesting bug, I'll take a look. Thanks!

jespino:
@jgilliam17 I removed the lazy loading from that part because it is built in a way that is hard to fix the problem, so I reverted that part of the changes.

jgilliam17:
Thanks @jespino 
Tested, looks good to merge.
- Verified listed components are loading as expected on first try. [E2E report](https://automation-dashboard.vercel.app/cycles/62da2ada-1ba0-49b4-aa8e-730d0a944fe4) looks good, no PR related failures.
One question before merge:
Is the `next_steps` component for Tips and Next Steps? I know that we still have Getting Started, and Tips and Next Steps were removed from the app recently. Just want to make sure I understand which one is changed here. Let me know. Thanks

jespino:
@jgilliam17 the changes in this PR are just making things load lazily, which means, only gets loaded on first use. That makes a lot of sense for things like modals, or things that are only shown once during onboarding, for example.

The last changes has reverted part of that (just the lazy loading part), because some animations were depending the existence of that component with the real final size (that wasn't happening with this lazy load) so it was generating some issues, I just reverted the change to do it later.

I'm not sure if I'm answering your question here.

jgilliam17:
My question is related to [this PR](https://github.com/mattermost/mattermost-webapp/pull/9312/files) that removed Next Steps, just trying to understand if we need to make any changes to component we no longer use

jgilliam17 (approved):

jgilliam17:
@M-ZubairAhmed do you still need to take a look?

jespino:
@jgilliam17 I think that is just removing one specific step on that form, so it shouldn't affect this PR, specially now that we are removing the lazy loading there.

M-ZubairAhmed (approved):

PR #9371, opened 2021-11-12
Lazy loading pdf.js library

jespino:
#### Summary
This adds lazy loading to the pdf.js library, that is used only for previewing pdf but it is a heavy library. Also it changes how is exposed to the plugins to make it lazy.

#### Ticket Link
[MM-40323](https://mattermost.atlassian.net/browse/MM-40323)

#### Release Note
```release-note
NONE
```


components/pdf_preview.jsx
```diff
@@ -147,7 +146,10 @@ export default class PDFPreview extends React.PureComponent {
     }
 
     getPdfDocument = () => {
-        PDFJS.getDocument(this.props.fileUrl).then(this.onDocumentLoad).catch(this.onDocumentLoadError);
+        import('pdfjs-dist').then((PDFJS) => {
```
M-ZubairAhmed, line 149: This looks great to me, i only have one small convern however that the function `getPdfDocument` is also being called in [componentDidUpdate](https://github.com/mattermost/mattermost-webapp/blob/669f0aa9f07b3d803461b2ed3be187e84317375f/components/pdf_preview.jsx#L78) do you think it is fine to import the library once again on every update or can we store the copy of it in ref or something? just curious

jespino: AFAIK that is handled by webpack, the first import load the module and insert in the list of existing modules, after that the module is reuse on every import.

hmhealey: It does look a bit weird to be doing this setup of setting `disableWorker` repeatedly, but I don't think calling it multiple times should cause a problem. We do that in the `LatexBlock` component with `katex`, and even though that's in `componentDidMount`, it'll still be called repeatedly when more of those are mounted and unmounted.

We could save the library in `this.state.pdfjs` like we do in `LatexBlock` for consistency, but I don't think it'd make a functional difference

plugins/export.js
```diff
@@ -29,7 +29,10 @@ window.ReactRedux = require('react-redux');
 window.ReactBootstrap = require('react-bootstrap');
 window.ReactRouterDom = require('react-router-dom');
 window.PropTypes = require('prop-types');
-window.PDFJS = require('pdfjs-dist');
+window.importPDFJS = () => import('pdfjs-dist').then((m) => {
```
hmhealey, line 32: Which plugin uses `PDFJS` again? Are we going to run into any compatibility problems by changing how this is imported? Perhaps we could do some magic with `Object.defineProperty` to define `PDFJS` as a getter function that only does the import when it's called?

Also, we might want to add a comment to this one to flag it as unsafe like with the components below that are defined on the window

components/pdf_preview.jsx
M-ZubairAhmed: OK if you both feel like re importing doesn't make any difference whatsoever i am good with it 👍🏽

plugins/export.js
jespino: Still the problem with that, is that is going to be async, so... you end up having a different API anyway.

components/pdf_preview.jsx
jespino: Yes,  I think this is clean enough, adding state when is not needed doesn't looks like very useful.

M-ZubairAhmed (approved):

plugins/export.js
hmhealey: That's a good point, but we should still know the impact of this before making a breaking change for plugin devs

components/pdf_preview.jsx
M-ZubairAhmed: If I were to do it I wouldn't have added state for it but ref would have been fine. With out both is also fine

plugins/export.js
jespino: I don't think this is a "breaking change" because this is not included as part of the official API. I think I added that for my file search support as a plugin, and is no longer needed, because that proof of concept ends up being part of the product itself.

hmhealey: Even if it's undocumented, I'd still consider it a breaking change, but I do agree that it's not one we need to follow any sort of deprecation policy on.

In that case, could we just remove this entirely instead of adding something new that might not be used? We can always re-add it if someone notices that it's missing

jespino: Good point, lets remove it then.

jespino: Done

hmhealey (approved):
Cool. I'm all for exporting less stuff to plugins since it's less stuff for us to have to support while we get a better system figured out for it 😀

hmhealey:
Is this something we'd want QA to test? We could defer QA testing for it, but I think that requires a Jira ticket so that QA know they have something to test

jespino:
@hmhealey yes, I think at least we should test the case where the user open a PDF file preview.

jgilliam17:
Thanks @jespino 
Here's the [E2E report ](https://automation-dashboard.vercel.app/cycles/9152d82c-e00f-40f0-ab58-6c8136f70f47) -looks good, no related failures.
Manually tested PDF file preview, works as before - looks good to merge.

jgilliam17 (approved):

PR #9376, opened 2021-11-16
Cypress/E2E: Fix cloud test related to max file size

saturninoabril:
#### Summary
- changed default max file size to 100MB
- make the text verification depends on what's with the current config instead of hard-coded value

#### Screenshots
Screen Shot 2021-11-16 at 9 08 35 AM

#### Release Note
```release-note
NONE
```

### DIFF
--- a/e2e/cypress/utils/file.js
+++ b/e2e/cypress/utils/file.js
@@ -1,0 +1,27 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// Converts a file size in bytes into a human-readable string of the form '123MB'.
+export function fileSizeToString(bytes) {
+    // it's unlikely that we'll have files bigger than this
+    if (bytes > 1024 ** 4) {
+        // check if file is smaller than 10 to display fractions
+        if (bytes < (1024 ** 4) * 10) {
+            return (Math.round((bytes / (1024 ** 4)) * 10) / 10) + 'TB';
+        }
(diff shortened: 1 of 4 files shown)


e2e/cypress/utils/file.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// Converts a file size in bytes into a human-readable string of the form '123MB'.
+export function fileSizeToString(bytes) {
+    // it's unlikely that we'll have files bigger than this
+    if (bytes > 1024 ** 4) {
+        // check if file is smaller than 10 to display fractions
```
josephbaylon, line 8: Quick question: are fractions not displayed if file is greater than 10?

josephbaylon (approved):
LGTM. Tested and passed locally. Just a minor question below

e2e/cypress/utils/file.js
saturninoabril: This is an exact copy of https://github.com/mattermost/mattermost-webapp/blob/master/utils/utils.jsx#L1230-L1253 which I missed to mention or indicate somewhere here. So, it will have the same behavior as the webapp itself.

Phrynobatrachus (approved):
passing locally :+1:

saturninoabril:
As discussed offline, I'll leave this as is so that Cass will be able to check merging a PR once given GH permission/access.

PR #9385, opened 2021-11-16
fix/MM-39870 : ADA Accessibility: NVDA does not read out all elements in a channel header, nor does it TAB to all elements in the channel header

M-ZubairAhmed:
#### Summary
Adds `aria-label` to plugin channel buttons.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39870

#### Related Pull Requests

#### Screenshots

https://user-images.githubusercontent.com/17708702/141994903-807711db-4513-444c-8bee-a3d7b700046b.mov

#### Release Note

```release-note

```

### DIFF
--- a/components/channel_header/components/header_icon_wrapper.tsx
+++ b/components/channel_header/components/header_icon_wrapper.tsx
@@ -22,7 +22,7 @@ type Props = {
     iconComponent: React.ReactNode;
     onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
     tooltipKey: string;
-    tooltipText?: string;
+    tooltipText?: React.ReactNode;
     isRhsOpen?: boolean;
 }
 
@@ -105,6 +105,11 @@ const HeaderIconWrapper: React.FC<Props> = (props: Props) => {
         );
     }
 
+    let ariaLabelText;
+    if (ariaLabel) {
+        ariaLabelText = `${localizeMessage(toolTips[tooltipKey].messageID, toolTips[tooltipKey].message)}`;
+    }
+
     let tooltip;
     if (tooltipKey === 'plugin' && tooltipText) {
         tooltip = (
@@ -115,13 +120,12 @@ const HeaderIconWrapper: React.FC<Props> = (props: Props) => {
                 <span>{tooltipText}</span>
             </Tooltip>
         );
+
+        if (typeof tooltipText === 'string') {
+            ariaLabelText = tooltipText;
+        }
     } else {
         tooltip = getTooltip(tooltipKey);
-    }
-
-    let ariaLabelText;
-    if (ariaLabel) {
-        ariaLabelText = `${localizeMessage(toolTips[tooltipKey].messageID, toolTips[tooltipKey].message)}`;
     }
 
     if (tooltip) {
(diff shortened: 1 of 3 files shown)


M-ZubairAhmed:
@larkox @nevyangelova i had to use the labels from channel header itself because we have two different store for mobile channel and regular channel header.

image

components/channel_header/components/header_icon_wrapper.tsx
```diff
@@ -22,7 +22,7 @@ type Props = {
     iconComponent: React.ReactNode;
     onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
     tooltipKey: string;
-    tooltipText?: React.ReactNode;
+    tooltipText?: string;
```
nevyangelova, line 25: I would say keep both because I think there was a reasoning behind that type. 

```suggestion
    tooltipText?: React.ReactNode | string;
```

ghost (changes requested):
Thanks @M-ZubairAhmed

ghost (approved):

components/channel_header/components/header_icon_wrapper.tsx
larkox: I have seen this before for being able to send over the `FormattedText` component. In this particular case, we sadly accept according to the documentation any react component can be passed here :(

https://developers.mattermost.com/integrate/plugins/webapp/reference/#registerChannelHeaderButtonAction

I would deprecate that, things I don't think any plugin is using a component there, but we cannot be sure.

larkox: A possible solution is to check whether it is a string or not, and if it is not a string, use the plugin id as a fallback.

larkox (changes requested):
Nevy's concerns are valid. I proposed a possible solution, but let me know what do you think.

components/channel_header/components/header_icon_wrapper.tsx
M-ZubairAhmed: Thanks @larkox for referencing the docs, i see the point now. @nevyangelova `ReactNode` combination of many types including `String`.
I updated as per suggestion.

M-ZubairAhmed:
cc @deanwhillier

components/channel_header/components/header_icon_wrapper.tsx
```diff
@@ -105,6 +105,11 @@ const HeaderIconWrapper: React.FC<Props> = (props: Props) => {
         );
     }
 
+    let ariaLabelText;
```
larkox, line 108: Am I missing something, or there are still cases where the aria label text is not defined? (if no `ariaLabel`, and not `tooltipText` or `tooltipText` not being a string)

larkox (changes requested):
One small detail to confirm.

M-ZubairAhmed:
Hey after discussion it looks like although the approach in this pr works it's not the best one and it has to be implemented on per plug-in basis. @aaronrothschild this might be good one for Demansol. CC @jwilander

M-ZubairAhmed:
that sounds good 👍🏽

PR #9389, opened 2021-11-16
[MM-39816] Fix mentions permalink

larkox:
#### Summary
Fix mentions permalink not showing the correct team.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39816

#### Related Pull Requests
None

#### Screenshots
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/dot_menu/index.ts
+++ b/components/dot_menu/index.ts
@@ -79,7 +79,7 @@ function mapStateToProps(state: GlobalState, ownProps: Props) {
     const userId = getCurrentUserId(state);
     const channel = getChannel(state, post.channel_id);
     const currentTeam = getCurrentTeam(state) || {};
-    const team = getTeam(state, post.channel_id);
+    const team = getTeam(state, channel.team_id);
     const teamUrl = `${getSiteURL()}/${team?.name || currentTeam.name}`;
 
     const systemMessage = isSystemMessage(post);


jgilliam17 (approved):
Thank you @larkox 
Tested, looks good to merge.
- Verified permalink created from aa saved post opens on the correct team.
Please ping me for re-test if changes are requested. Thanks

nevyangelova (approved):

components/dot_menu/index.ts
```diff
@@ -79,7 +79,7 @@ function mapStateToProps(state: GlobalState, ownProps: Props) {
     const userId = getCurrentUserId(state);
     const channel = getChannel(state, post.channel_id);
     const currentTeam = getCurrentTeam(state) || {};
-    const team = getTeam(state, post.channel_id);
+    const team = getTeam(state, channel.team_id);
```
hmhealey, line 82: Do DMs and GMs need to fall back to the current team ID here?

hmhealey (changes requested):

components/dot_menu/index.ts
larkox: The variable team is only used in the next line. Therefore, if it is a DM, it will use the currentTeam name.

hmhealey: Ah, I see. I also didn't realize that the old code was just incorrect since it was requesting a team using a channel ID

hmhealey (approved):

PR #9420, opened 2021-11-23
chore/MM-37187 : Updated only definetlyTyped packages; no parent package updates - Nov 21

M-ZubairAhmed:
#### Summary
Updates all those type definition packages without their parent updates.
i.e
```
@types/foo -> updated
foo -> no new version
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37187

#### Related Pull Requests

#### Screenshots
none

#### Release Note

```release-note

```


package.json
```diff
@@ -116,37 +115,38 @@
...
+    "@types/lodash": "4.14.177",
     "@types/luxon": "2.0.2",
-    "@types/mark.js": "8.11.5",
+    "@types/mark.js": "8.11.6",
     "@types/marked": "0.7.4",
     "@types/moment-timezone": "0.5.30",
     "@types/node": "16.3.1",
     "@types/react": "16.9.49",
-    "@types/react-beautiful-dnd": "^13.0.0",
+    "@types/react-beautiful-dnd": "^13.1.2",
```
jespino, line 132: Not related to this PR but we should use fixed versions instead of `^`

jespino (approved):

package.json
M-ZubairAhmed: Totally agree on this point. how should we go about doing this @hmhealey

hmhealey: If you just do `npm install --save-dev --exact @types/react-beautiful-dnd@13.1.2`, it should set that correctly. I'm surprised it wasn't set already since we have the `.npmrc` set up to prefer exact versions to begin with

hmhealey (approved):

package.json
M-ZubairAhmed: Yes i have removed the `^` from dependency, also checked other but none have the `^` or `~`.

PR #9426, opened 2021-11-23
MM-40050 Don't load channel-specific data on user update

hmhealey:
Some customers are reporting an issue where their server is periodically bombarded with requests for data that causes their database to lock up. We're pretty sure that's caused by one user updating something about themselves (such as their custom status) which causes all other connected users to attempt to load a bunch of information needed for some other features (more on that below). This only occurs when either:

1. The updated user is a guest user
2. The server has timezones enabled and is capable of supporting LDAP groups (ie E20)

The data that's fetched when that occurs are things that are possible side effects of a user update which we don't have a good way to reflect otherwise. It's used for two things which will be affected by these changes:

1. Showing/hiding the "This channel has guests" indicator in the channel header
2. Keeping track of the number of users/timezones within an LDAP group so that we can show a warning notification when a user attempts to `@group` which will notify a bunch of people at once

Since we don't have a good way of handling loading data for side effects, I've decided to just remove the real-time syncing of this since we already request this data on channel change. That means that the indicator in the channel header will only show/hide when you change channels and that the notification may be incorrect, but both of those only matter when those things change which should be rare

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40050

#### Release Note
```release-note
Removed real-time updates of a couple features to prevent overloading server on user update
```


hmhealey:
@mkraft @BenCookie95 I've added you to this PR since you're familiar with the features that are affected by this. Let me know if you're okay with those, and I can open it up fully for review

BenCookie95:
@hmhealey I'm ok with these changes, nice investigation work!

BenCookie95 (approved):

mkraft (approved):
Code looks good. What do you think about adding a hyperlink to this PR in place where the deleted? I'm just thinking of a way to avoid someone else coming along if these features get re-requested and they rewrite the code and accidentally reintroduce the perf. issue.

jgilliam17:
Thanks @hmhealey 
There are 3 tests that are failing on cypress, when testing manually I am not able to repro.
1. LHS DM sort order
2. Guest badge in mention autocomplete 
3. Invite guest existing team
E2E report: https://automation-dashboard.vercel.app/cycles/9763f655-838a-4887-b2bc-a049595ae2ff

hmhealey:
I can't reproduce the first 2 locally. The 3rd is reproducing, but I don't think it's related to my changes

jgilliam17 (approved):
Thanks for checking @hmhealey 
Tested, looks good to merge.
- Verified guest user feature and group mentions with timezone enabled - as expected.

hmhealey:
This has been cherry-picked as far back as 6.0. I'll do 5.39 and 5.37 later to complete the set.

PR #9432, opened 2021-11-24
Encapsulating react-bootstrap tooltip in a component

jespino:
#### Summary
The idea behind this PR is to be able to get rid of the react-bootstrap Tooltip
component in the near future using a compass-components replacement, or
anything else.

This is not changing anything but is making the work easier later

This changes reduce the amount of imports of react-bootstrap allover the application from 167 to 104

Another good idea is to also encapsulate the Modal component and subcomponets to replace them eventually.

#### Ticket Link
[MM-40358](https://mattermost.atlassian.net/browse/MM-40358)

#### Release Note
```release-note
NONE
```


components/tooltip.tsx
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jespino, line 1: Here is the key part. We just wrap the React Bootstrap component and received the minimun amount of properties to satisfy used properties in the rest of the code.

Take into consideration that `style` and `placement` are used only in 1 place, and `positionLeft` is only used in other 2 places, probably the component should be better if we just make that 2 cases something like the `type` of tooltip, or something like that.

jespino:
@michelengelen do we have something already in compass components to replace this? if not, would it be a good thing to add?

michelengelen:
> @michelengelen do we have something already in compass components to replace this? if not, would it be a good thing to add?

We do sort of have something in the pipeline using popperJs. Nothing that can be shown atm, but definitely on our roadmap for the near future

jespino:
@michelengelen awesome, so this can gets in, whenever, and then we can replace this adapter with an adapter for compass-components implementation :)

michelengelen (approved):
LGTM

components/tooltip.tsx
AshishDhama: @jespino should we add a few simple test cases as well

AshishDhama (approved):
LGTM! @hmhealey Will wrapping other react-bootstrap components will help us in upgrading to the latest version of react-bootstrap.

components/tooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+type Props = {
+    id?: string;
+    className?: string;
+    style?: CSSProperties;
+    children?: React.ReactNode;
+    positionLeft?: number;
+    placement?: string;
+};
+
+export default function Tooltip(props: Props) {
```
hmhealey, line 16: We could consider wrapping this in a `React.forwardRef` so that you can still access the underlying tooltip, but that's not really necessary unless we actually use that ref anywhere. 0/5 on changing anything now, but we might consider that if we start to do this in more places

hmhealey (approved):
This feels a bit like it's a lot of changes that do nothing at the moment, but I suppose it'll help out when we finally get around to replacing these with our custom component. Hopefully it's easy enough to just drop that on once it exists.

hmhealey:
> LGTM! @hmhealey Will wrapping other react-bootstrap components will help us in upgrading to the latest version of react-bootstrap.

It might make it a bit easier, but the bulk of the upgrade is still figuring out how to upgrade (or just flat out remove) regular Bootstrap since that's supposedly blocking us updating React-Bootstrap.

components/tooltip.tsx
jespino: I tried to make it as minimal as possible, only exposing what is needed now.

jespino:
> > LGTM! @hmhealey Will wrapping other react-bootstrap components will help us in upgrading to the latest version of react-bootstrap.
> 
> It might make it a bit easier, but the bulk of the upgrade is still figuring out how to upgrade (or just flat out remove) regular Bootstrap since that's supposedly blocking us updating React-Bootstrap.

If we wrap all react-bootstrap inside our own components, we can just modify what is needed in one single place, instead of everywhere. That is not the case for CSS that Is going to be a pain anyway.

components/tooltip.tsx
hmhealey: Oh yeah, that makes sense. I guess this isn't like where we'd be forwarding a ref to a DOM element.

components/tooltip.tsx
jespino: I can create an snapshot test, but because this is just a wrapper, is not going to get much advantage from testing for now. In the future, if we change the API or the logic to convert from outside properties to the underlaying properties, probably makes more sense, for now I would leave it without tests. But I'm happy to add them if you see value in it.

jgilliam17:
Thanks @jespino, apologies for delay in getting to this 
Smoke tested around tooltips in the app - looks good
Running E2Es [here](https://automation-dashboard.vercel.app/cycles/3b648c6f-5b78-404e-9a1d-9dd830353288), will review this evening.

jgilliam17:
@jespino I'm re-running the E2Es, the first report showed an unusual number of failed tests due to server failing to start. 
https://automation-dashboard.vercel.app/cycles/fb5235b6-c69f-4cf4-bd9d-9e8c25f66834

jgilliam17:
Verified latest E2E report, no PR related failures - looks good to merge.

jgilliam17 (approved):

PR #9434, opened 2021-11-24
Custom groups

BenCookie95:
#### Thing to test/watch out for
1. Mentions for custom groups and ldap groups at professional and E20 license levels. This includes highlighting in the UI
2. Check permissions on different user levels by toggling them in system console.
3. Tryout custom group mentions in ldap restricted teams/channels.
4. How does mobile perform with these groups? The webapp was not fetching groups while typing. There could be a performance issue on mobile if they fetch all groups on load.

#### TODO
- [x] Implement roles and permissions for group menu items. Also check if they are ldap or custom. (Martin) Search for `// TODO: Check Permissions here` in the code
- [ ] Need to add tests to create_post and create_comment
- [x] Add new group permission to system console (Martin)
- [ ] There are some UI improvements that I (Ben) needs to do at some stage in the new year
- [x] Hide _Create new user group_ item from the channel dropdown if the instance isn't licensed
- [x] `disable_group_highlight` post prop being set when group mention created with Professional license.
- [x] Join Group link in group view modal doesn't add current user to the members list
- [ ] group yellow highlighting only appears after opening the user groups modal

#### Summary
Custom user groups WIP PR. Summary of the changes:

- Added 5 new modals:
  -  AddUserToGroupModal - Adds new users to an existing group
  - CreateUserGroupsModal - Creates a new group
  - ViewUserGroupModal - View the details of a group
  - UpdateUserGroupModal - Update name or display name of group
  - UserGroupsModal - List of Groups (Custom and ldap)
- Updated compass-icons to `0.1.19`
- Changed `myGroups` type in redux to be an array of groupIds. It was previously duplicating data from the `allGroups` store.
- We now fetch groups when typing for @ mentions. Previously it would only search the redux store with a selector.
- We no longer load all groups onLoad, just the first 60.
- We have a feature flag and a config for enabling/disabling custom user groups
- Feature will only work with new professional and enterprise licenses

#### Related Pull Requests

- https://github.com/mattermost/mattermost-server/pull/18839
- https://github.com/mattermost/enterprise/pull/1097

#### Release Note

```release-note
Adding the chat facing UI for custom user groups
```


components/create_comment/index.ts
```diff
@@ -71,9 +71,10 @@ function makeMapStateToProps() {
         const canPost = haveIChannelPermission(state, channel.team_id, channel.id, Permissions.CREATE_POST);
         const useChannelMentions = haveIChannelPermission(state, channel.team_id, channel.id, Permissions.USE_CHANNEL_MENTIONS);
         const isLDAPEnabled = license?.IsLicensed === 'true' && license?.LDAPGroups === 'true';
-        const useGroupMentions = isLDAPEnabled && haveIChannelPermission(state, channel.team_id, channel.id, Permissions.USE_GROUP_MENTIONS);
```
mkraft, line 74: We still want group mentions to work for _ldap_ groups even if _custom_ group isn't enabled.

components/add_user_to_group_multiselect/add_user_to_group_multiselect.tsx
```diff
@@ -0,0 +1,271 @@
...
+    renderOption = (option: UserProfileValue, isSelected: boolean, onAdd: (user: UserProfileValue) => void, onMouseMove: (user: UserProfileValue) => void) => {
+        let rowSelected = '';
+        if (isSelected) {
+            rowSelected = 'more-modal__row--selected';
+        }
+
+        return (
+            <div
+                key={option.id}
+                ref={isSelected ? this.selectedItemRef : option.id}
```
jespino, line 195: I'm not 100% sure about this, I don't know if this ref is needed (looks like a leaky abstraction on the multiselect component), also the case where is not selected, why are we setting a ref equal to `option.id`, shouldn't it be something like `null` or `undefined`? (Anyway, this patterns looks like copied from the multiselect component, so I don't think this is an stopper.

components/add_user_to_group_multiselect/add_user_to_group_multiselect.tsx
```diff
@@ -0,0 +1,271 @@
...
+}
+
+type State = {
+    values: UserProfileValue[];
+    term: string;
+    saving: boolean;
+    loadingUsers: boolean;
+}
+
+export default class AddUserToGroupMultiSelect extends React.PureComponent<Props, State> {
```
jespino, line 69: nit: This component has 270 LOC, I don't love big components, I don't know if there is any way to reduce that, maybe moving the render option in its own component. Again, this is not an stopper, it is just a perception that this component is a bit bigger than what I would like.

components/add_users_to_group_modal/add_users_to_group_modal.tsx
```diff
@@ -0,0 +1,171 @@
...
+}
+
+type State = {
+    show: boolean;
+    savingEnabled: boolean;
+    usersToAdd: UserProfile[];
+    showUnknownError: boolean;
+}
+
+export default class AddUsersToGroupModal extends React.PureComponent<Props, State> {
```
jespino, line 41: nit: We are heading more usage of functional components for components that has less logic, probably this is a good candidate for a functional component.

components/create_user_groups_modal/create_user_groups_modal.tsx
```diff
@@ -0,0 +1,265 @@
...
+    mention: string;
+    savingEnabled: boolean;
+    usersToAdd: UserProfile[];
+    mentionUpdatedManually: boolean;
+    mentionInputErrorText: string;
+    nameInputErrorText: string;
+    showUnknownError: boolean;
+}
+
+export default class CreateUserGroupsModal extends React.PureComponent<Props, State> {
```
jespino, line 45: nit: Again, not an stopper, but we could subdivide this on smaller functional components?

components/multiselect/multiselect.tsx
```diff
@@ -495,10 +506,24 @@ export default class MultiSelect<T extends Value> extends React.PureComponent<Pr
                 </div>
                 {this.props.saveButtonPosition === 'bottom' &&
                 <div className='multi-select__footer'>
+                    {
+                        typeof this.props.backButtonClick === 'function' &&
```
jespino, line 510: Is there any reason why we can't use just `this.props.backButtonClick`, I would expect it to be a function or `undefined` (and that is what the type check expects too).

components/create_user_groups_modal/create_user_groups_modal.tsx
```diff
@@ -0,0 +1,265 @@
...
+                show={this.state.show}
+                onHide={this.doHide}
+                onExited={this.props.onExited}
+                role='dialog'
+                aria-labelledby='createUserGroupsModalLabel'
+                id='createUserGroupsModal'
+            >
+                <Modal.Header closeButton={true}>
+                    {
+                        typeof this.props.backButtonCallback === 'function' ?
```
jespino, line 167: Is there any reason to not use just `this.props.backButtonCallback ? ...`. Looks like the type definition assures that this is or a function or an `undefined`.

components/update_user_group_modal/update_user_group_modal.scss
```diff
@@ -0,0 +1,42 @@
+.modal {
```
jespino, line 1: Same here, I would expect a component name (with upper case) as the root class to get this scoped properly.

components/update_user_group_modal/update_user_group_modal.tsx
```diff
@@ -0,0 +1,224 @@
...
+    show: boolean;
+    name: string;
+    mention: string;
+    hasUpdated: boolean;
+    mentionInputErrorText: string;
+    nameInputErrorText: string;
+    showUnknownError: boolean;
+}
+
+export default class UpdateUserGroupModal extends React.PureComponent<Props, State> {
```
jespino, line 45: nit: can this be rewritten as a functional component?

components/user_groups_modal/user_groups_modal.scss
```diff
@@ -0,0 +1,347 @@
+@charset 'UTF-8';
+
+.modal {
```
jespino, line 3: Again, this is not scoped to a component class.

components/user_groups_modal/user_groups_modal.tsx
```diff
@@ -0,0 +1,469 @@
...
+    page: number;
+    myGroupsPage: number;
+    loading: boolean;
+    show: boolean;
+    selectedFilter: string;
+    allGroupsFull: boolean;
+    myGroupsFull: boolean;
+}
+
+export default class UserGroupsModal extends React.PureComponent<Props, State> {
```
jespino, line 72: This component is huge, can we split this in subcomponents (probably functional components)?

components/view_user_group_modal/view_user_group_modal.scss
```diff
@@ -0,0 +1,156 @@
+.modal {
```
jespino, line 1: Again, the root class is not based on component name.

components/view_user_group_modal/view_user_group_modal.tsx
```diff
@@ -0,0 +1,437 @@
...
+
+type State = {
+    page: number;
+    loading: boolean;
+    show: boolean;
+    selectedFilter: string;
+    memberCount: number;
+}
+
+export default class ViewUserGroupModal extends React.PureComponent<Props, State> {
```
jespino, line 68: Another huge component, is there any way to subdivide this in simpler components?

jespino (changes requested):
I give it a first review, there are some tiny things that can be improved (in my opinion).

components/create_user_groups_modal/create_user_groups_modal.scss
```diff
@@ -0,0 +1,21 @@
+.modal {
```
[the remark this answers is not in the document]
BenCookie95: Root class is now related to the component name

components/multiselect/multiselect.tsx
BenCookie95: Fixed, not sure why I was using `typeof this.props.backButtonClick === 'function'`

components/create_user_groups_modal/create_user_groups_modal.tsx
BenCookie95: Fixed

components/update_user_group_modal/update_user_group_modal.scss
BenCookie95: Root class is now related to the component name

components/user_groups_modal/user_groups_modal.scss
BenCookie95: Root class is now related to the component name

components/view_user_group_modal/view_user_group_modal.scss
BenCookie95: Root class is now related to the component name

components/add_user_to_group_multiselect/add_user_to_group_multiselect.tsx
BenCookie95: split this out into another component called `MultiSelectOption`, I removed the option.id as a ref and it's setup the same as `components/more_direct_channels/list_item/list_item.tsx`.

components/add_user_to_group_multiselect/add_user_to_group_multiselect.tsx
BenCookie95: Moved the render option part into its own functional component, this component is now 244 LOC. I don't think there is much else I can do to reduce it.

components/add_users_to_group_modal/add_users_to_group_modal.tsx
BenCookie95: Agreed, changed this to a functional component

components/update_user_group_modal/update_user_group_modal.tsx
BenCookie95: This is now a functional component

mkraft (commented):
test review comment

components/user_groups_modal/user_groups_modal.tsx
BenCookie95: Moved the modal header and group list out into their own smaller functional components. Reduced the size of this file by about 100 LOC

components/view_user_group_modal/view_user_group_modal.tsx
BenCookie95: Separated this out into smaller functional components

BenCookie95:
@jespino I have addressed most of your comments and separated stuff out into some smaller functional components!

mkraft (approved):

components/add_users_to_group_modal/add_users_to_group_modal.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+    const addUserCallback = (users: UserProfile[]): void => {
+        setUsersToAdd(users);
+    };
+
+    const deleteUserCallback = (users: UserProfile[]): void => {
+        setUsersToAdd(users);
+    };
+
+    const addUsersToGroup = async (users?: UserProfile[]) => {
```
jespino, line 55: All this functions are regenerated in every single render, triggering unneeded re-renders in the underneath components. Not an stopper, but something too fix sooner than later. I would use `useCallback` in most of this functions.

components/add_users_to_group_modal/add_users_to_group_modal.tsx
```diff
@@ -0,0 +1,154 @@
...
+            role='dialog'
+            aria-labelledby='createUserGroupsModalLabel'
+            id='addUsersToGroupsModal'
+        >
+            <Modal.Header closeButton={true}>
+                <button
+                    type='button'
+                    className='modal-header-back-button btn-icon'
+                    aria-label='Close'
+                    onClick={() => {
```
jespino, line 92: This anonymous function makes no sense, we can pass just `goBack` directly and has exactly the same effect, and also, if we make the `goBack` function not changing every single render, we can avoid re-rendering this button in every render too.

components/add_users_to_group_modal/add_users_to_group_modal.tsx
```diff
@@ -0,0 +1,154 @@
...
+                        <div className='group-add-user'>
+                            <AddUserToGroupMultiSelect
+                                multilSelectKey={'addUsersToGroupKey'}
+                                onSubmitCallback={addUsersToGroup}
+                                focusOnLoad={false}
+                                savingEnabled={isSaveEnabled()}
+                                addUserCallback={addUserCallback}
+                                deleteUserCallback={deleteUserCallback}
+                                groupId={props.groupId}
+                                searchOptions={{
```
jespino, line 128: This object is going to change in every render, probably we can generate a make it stable with a `useMemo`

components/add_users_to_group_modal/index.ts
```diff
@@ -0,0 +1,46 @@
...
+type Actions = {
+    addUsersToGroup: (groupId: string, userIds: string[]) => Promise<ActionResult>;
+    openModal: <P>(modalData: ModalData<P>) => void;
+}
+
+type OwnProps = {
+    groupId: string;
+}
+
+function makeMapStateToProps() {
```
jespino, line 27: You don't need the `makeMapStateToProps` you can just use `mapStateToProps` and get rid of the "return a function" approach. (you can see it in other `index.ts` files).

The point of `makeMapStateToProps` is to generate selectors at mount time that are reused and take advantage of the cache in runtime, if you are not generating that selectors, this is just more complex code without any need.

components/create_user_groups_modal/create_user_groups_modal.tsx
jespino: Looks like it is not fixed

components/global_header/left_controls/product_menu/product_menu_list/product_menu_list.tsx
```diff
@@ -103,6 +120,22 @@ const ProductMenuList = (props: Props): JSX.Element | null => {
                         />
                     }
                 />
+                <Menu.ItemToggleModalRedux
```
jespino, line 123: I guess that is about UX/QA, but is this always visible? to anybody? there is no permission needed? or license? or config?

components/update_user_group_modal/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import {openModal} from 'actions/views/modals';
+import {getGroup} from 'mattermost-redux/selectors/entities/groups';
+
+import UpdateUserGroupModal from './update_user_group_modal';
+
+type OwnProps = {
+    groupId: string;
+}
+
+function makeMapStateToProps() {
```
jespino, line 23: Again, we don't need here to use `makeMapStateToProps`, we can just use `mapStateToProps`

components/update_user_group_modal/update_user_group_modal.tsx
```diff
@@ -0,0 +1,221 @@
...
+    };
+
+    const updateMentionState = (e: React.ChangeEvent<HTMLInputElement>) => {
+        const value = e.target.value;
+        setHasUpdated(true);
+        setMention(value);
+        setMentionUpdatedManually(true);
+    };
+
+    const patchGroup = async () => {
```
jespino, line 74: Again, all this callback functions could take advantage of something like `useCallback`

components/user_groups_modal/user_groups_list/user_groups_list.tsx
```diff
@@ -0,0 +1,153 @@
...
+            onScroll={onScroll}
+            ref={ref}
+        >
+            {(groups.length === 0 && searchTerm) &&
+                <NoResultsIndicator
+                    variant={NoResultsVariant.ChannelSearch}
+                    titleValues={{channelName: `"${searchTerm}"`}}
+                />
+            }
+            {groups.map((group) => {
```
jespino, line 78: nit: Probably you can change this slightly to something like:

```js
{groups.map((group) => (
    <div>
        ....
    </div>
)}
```

It is equivalent, but remove some indentation and looks a bit nicer.

You can also think about the possibility of moving that code in another functional component, because looks like a ton of JSX that can be extracted easily.

components/user_groups_modal/user_groups_modal_header/user_groups_modal_header.tsx
```diff
@@ -0,0 +1,61 @@
...
+    canCreateCustomGroups: boolean;
+    onExited: () => void;
+    backButtonAction: () => void;
+    actions: {
+        openModal: <P>(modalData: ModalData<P>) => void;
+    };
+}
+
+const UserGroupsModalHeader = (props: Props) => {
+    const goToCreateModal = () => {
```
jespino, line 23: This is created for every single render, so it is going to force the re-render of the underneath component in every render too. a `useCallback` can help.

components/view_user_group_modal/view_user_group_list_item/view_user_group_list_item.tsx
```diff
@@ -0,0 +1,80 @@
...
+    };
+}
+
+const ViewUserGroupListItem = (props: Props) => {
+    const {
+        user,
+        group,
+    } = props;
+
+    const removeUserFromGroup = async (userId: string) => {
```
jespino, line 31: You can use `useCallback` here and create a function with 0 parameters and use the `user.id` from the context, that way you can pass that memoized callback to the button component avoiding re-rendering it on every render.

components/view_user_group_modal/view_user_group_modal_header/view_user_group_modal_header.tsx
```diff
@@ -0,0 +1,229 @@
...
+                        id='user_groups_modal.addPeople'
+                        defaultMessage='AddPeople'
+                    />
+                </button>
+            );
+        }
+        return (<></>);
+    };
+
+    const subMenuButton = () => {
```
jespino, line 142: Why is not this a component? Looks like a component, but is not taking advantage of it. It is just more complexity here and probably more rendering work.

This also apply to the other 2 functions that renders something similar to a component, but they are not components.

jespino (approved):
I don't see nothing that stops us from merging this, but there are a lot of things that can be improved. A lot of miss usage of functional components, and a lot of unneeded re-renders for sure. I would pay special attention to one mapStateToProps that is generating a new object every single time, what is going to trigger a render for every single store change while that component is mounted. Other than that, a lot of space for improvement.

My suggestion is to create a follow up ticket with all my improvement suggestions, and another "with more priority" about the mapStateToProps problem.

furqanmlk (approved):
Working as expected.

PR #9409, opened 2021-11-22
Removing remaining jquery imports

jespino:
This leave the jquery dependency in the package.json because it is required by
bootstrap. We need to eventually remove bootstrap to be able to definitively
get rid of jquery.

#### Summary
Removing remaining jquery imports

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27444
https://mattermost.atlassian.net/browse/MM-27445
https://mattermost.atlassian.net/browse/MM-29031
https://mattermost.atlassian.net/browse/MM-29032

#### Release Note
```release-note
NONE
```

PR #9444, opened 2021-11-25
MM-27450 Remove remaining references to jQuery 🎉

hmhealey:
I was sad to see that Jesús had removed all of jQuery without me having a chance to get involved, so I'm glad to have found a way to contribute to that

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27450

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/9409

#### Release Note
```release-note
NONE
```


devinbinnie (approved):
YES

jespino:
I tried that, but I think bootstrap was a problem because depends on jQuery being present.

hmhealey:
It turns out we can just fix that by not importing Bootstrap 😀

We still rely on Bootstrap's CSS in a some places, but it looks like we haven't relied on the JS part of it since we finally got everything switched over to React-Bootstrap at some point. Comparing this test server with one on my other PR, we're reducing the bundle size by about 40KB here too

jespino (approved):
Love it!!

package.json
```diff
@@ -39,7 +39,6 @@
     "html-to-react": "1.4.5",
     "inobounce": "0.2.0",
     "jasny-bootstrap": "3.1.3",
-    "jquery": "3.6.0",
```
M-ZubairAhmed, line 42: This single line of code gives me so much happiness 😊

PR #9445, opened 2021-11-26
Migrating some files to typescript

jespino:
#### Summary
Migrating some files to typescript

#### Ticket Link
Not yet

#### Release Note
```release-note
NONE
```


reducers/plugins/index.ts
```diff
@@ -5,9 +5,15 @@ import {combineReducers} from 'redux';
...
+import type {GenericAction} from 'mattermost-redux/types/actions';
+import {IDMappedObjects} from 'mattermost-redux/types/utilities';
+import {ClientPluginManifest} from 'mattermost-redux/types/plugins';
+
+import type {PluginComponent, ProductComponent, PostPluginComponent, AdminConsolePluginComponent} from 'types/store/plugins';
+
 import {ActionTypes} from 'utils/constants';
 
-function hasMenuId(menu, menuId) {
+function hasMenuId(menu: any, menuId: string) {
```
devinbinnie, line 16: Could we add an in-line type for this?
```suggestion
type Menu = {
    id: string:
    subMenu?: Menu[];
};

function hasMenuId(menu: Menu, menuId: string) {
```

reducers/plugins/index.ts
```diff
@@ -24,9 +30,9 @@ function hasMenuId(menu, menuId) {
     return false;
 }
 
-function buildMenu(rootMenu, data) {
+function buildMenu(rootMenu: any, data: any) {
```
devinbinnie, line 33: Could do the same here, not sure if this will work exactly but it should be something like this:
```suggestion
function buildMenu(rootMenu: Menu, data: Menu) {
```

reducers/plugins/index.ts
```diff
@@ -294,10 +309,10 @@ function adminConsoleCustomComponents(state = {}, action) {
         const key = action.data.key.toLowerCase();
 
         const nextState = {...state};
-        if (!nextState[pluginId]) {
-            nextState[pluginId] = {};
+        let nextArray: any = {};
```
devinbinnie, line 312: Do we need to type this? If so, can we remove any and try this?
```suggestion
        let nextArray: Record<string, AdminConsolePluginComponent> = {};
```

reducers/views/channel_selector_modal.ts
```diff
@@ -3,12 +3,14 @@
...
+import type {GenericAction} from 'mattermost-redux/types/actions';
+
 import {ChannelTypes, UserTypes} from 'mattermost-redux/action_types';
 
-function channels(state = [], action) {
+function channels(state: string[] = [], action: GenericAction) {
     switch (action.type) {
     case ChannelTypes.RECEIVED_ALL_CHANNELS:
-        return action.data.map((v) => v.id);
+        return action.data.map((v: any) => v.id);
```
devinbinnie, line 13: Wouldn't `v` be of type `Channel` here?
```suggestion
        return action.data.map((v: Channel) => v.id);
```

selectors/general.ts
```diff
@@ -18,17 +20,18 @@ export function areTimezonesEnabledAndSupported(state) {
...
-export function getBasePath(state) {
+export function getBasePath(state: GlobalState) {
     const config = getConfig(state) || {};
 
     if (config.SiteURL) {
         return new URL(config.SiteURL).pathname;
     }
 
-    return window.basename || '/';
+    return (window as any).basename || '/';
```
devinbinnie, line 30: We can add a type declaration here for the `window` to avoid the any cast, it looks like this:
```
declare global {
    interface Window {
        basename: string
    }
}
```

devinbinnie (changes requested):

reducers/plugins/index.ts
jespino: I can try, that is not exactly the valid type. Let me take a look to this.

jespino: Fixed, this required more work that I expected, but now I think looks good :)

jespino:
@devinbinnie PTAL

devinbinnie (approved):
Thanks for the changes @jespino, looks great now :)

reducers/plugins/index.ts
```diff
@@ -49,7 +55,16 @@ function sortComponents(a, b) {
     return 0;
 }
 
-function removePostPluginComponents(state, action) {
+type ComponentsState = {
```
hmhealey, line 58: If you wanted to avoid redefining these, you could import `PluginsState` from `types/plugins` and type things like `PluginsState['components']` and `PluginsState['postTypes']`. Same with the various other reducers in here

reducers/views/channel.ts
```diff
@@ -96,7 +98,7 @@ function mobileView(state = false, action) {
 }
 
 // lastUnreadChannel tracks if the current channel was unread and if it had mentions when the user switched to it.
-function lastUnreadChannel(state = null, action) {
+function lastUnreadChannel(state: (Channel & {hadMentions: boolean}) | null = null, action: GenericAction) {
```
hmhealey, line 101: If we don't have a proper definition for this in the state, we could use this here:
```suggestion
function lastUnreadChannel(state: ({channelId: string; hadMentions: boolean}) | null = null, action: GenericAction) {
```
since this only has the ID and not the whole object

reducers/views/search.ts
```diff
@@ -15,7 +17,9 @@ function modalSearch(state = '', action) {
     }
 }
 
-function modalFilters(state = {}, action) {
+type ModalFiltersState = {roles?: string[]; channel_roles?: string[]; team_roles?: string[]};
+
+function modalFilters(state: ModalFiltersState = {}, action: GenericAction) {
```
hmhealey, line 22: Getting the type from the store type would definitely help here

reducers/views/settings.ts
```diff
@@ -3,7 +3,9 @@
 
 import {ActionTypes} from 'utils/constants';
 
-export default function settings(state = {}, action) {
+import type {GenericAction} from 'mattermost-redux/types/actions';
+
+export default function settings(state: {activeSection: string; previousActiveSection: string} = {activeSection: '', previousActiveSection: ''}, action: GenericAction) {
```
hmhealey, line 8: And here

selectors/general.ts
```diff
@@ -29,6 +37,7 @@ export function getBasePath(state) {
 }
 
 export const getCurrentUserTimezone = createSelector(
+    'getCurrentUserTimezone',
```
hmhealey, line 40: I'm a bit worried that this didn't break before now. This seems like this should've broken months ago when Joram added the custom version of `createSelector`. At least it's fixed now though

selectors/storage.ts
```diff
@@ -1,31 +1,35 @@
...
 // See LICENSE.txt for license information.
 
+import {Dictionary} from 'mattermost-redux/types/utilities';
+
 import {getPrefix} from 'utils/storage_utils';
 
-export const getGlobalItem = (state, name, defaultValue) => {
+import type {GlobalState} from 'types/store';
+
+export const getGlobalItem = (state: GlobalState, name: string, defaultValue: any) => {
```
hmhealey, line 10: I don't know if it's worth changing this now since I'm hoping to rewrite all this, but we could possibly parameterize these functions so that the user can override the return type and keep TypeScript happy.
```suggestion
export const getGlobalItem = <T = any>(state: GlobalState, name: string, defaultValue: T) => {
```

hmhealey (changes requested):

reducers/plugins/index.ts
jespino: Nice one, I'm changing it.

reducers/views/channel.ts
jespino: Done

reducers/views/search.ts
jespino: Done

reducers/views/settings.ts
jespino: Done

selectors/storage.ts
jespino: Done

hmhealey (approved):

PR #9451, opened 2021-11-29
MM-40071 Only calculate suggestion list max height when first opened

hmhealey:
As a performance improvement, we're wanting to not inspect the DOM during rendering which is an anti-pattern anyway. One place we do this is when calculating the max height for the `SuggestionList`, which currently happens regardless of whether or not that list is visible. Instead of doing it during the render, we're now calculating it in `componentDidUpdate`, but only when the `SuggestionList` first becomes visible.

Also while doing this, I had to get the DOM node that makes up the textbox and instead of going through a series of refs and calling methods to find the actual DOM node, I had to change it so that the ref was properly forwarded. The original version only worked on the post textbox and not the search box because one used an `AutosizeTextarea` and the other does not.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40071

#### Release Note
```release-note
Improved performance while typing by moving some autocomplete layout calculation
```

### DIFF
--- a/components/edit_post_modal/edit_post_modal.tsx
+++ b/components/edit_post_modal/edit_post_modal.tsx
@@ -448,7 +448,6 @@ export class EditPostModal extends React.PureComponent<Props, State> {
     handleEntered = () => {
         if (this.editbox) {
             this.editbox.focus();
-            this.editbox.recalculateSize();
         }
     };
 
--- a/components/suggestion/suggestion_list.jsx
+++ b/components/suggestion/suggestion_list.jsx
@@ -74,7 +74,11 @@ export default class SuggestionList extends React.PureComponent {
     }
 
     updateMaxHeight = () => {
-        const inputHeight = this.props.inputRef.current?.input.clientHeight ?? 0;
+        if (!this.props.inputRef?.current) {
+            return;
+        }
+
+        const inputHeight = this.props.inputRef.current.input.clientHeight ?? 0;
 
         this.maxHeight = Math.min(
             window.innerHeight - (inputHeight + Constants.POST_MODAL_PADDING),
--- a/components/suggestion/suggestion_list.jsx
+++ b/components/suggestion/suggestion_list.jsx
@@ -51,6 +51,10 @@ export default class SuggestionList extends React.PureComponent {
         this.currentItem = {};
     }
 
+    componentDidMount() {
+        this.updateMaxHeight();
+    }
+
     componentDidUpdate(prevProps) {
         if (this.props.selection !== prevProps.selection && this.props.selection) {
             this.scrollToItem(this.props.selection);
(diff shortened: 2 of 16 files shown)


components/suggestion/suggestion_list.jsx
```diff
@@ -58,12 +59,29 @@ export default class SuggestionList extends React.PureComponent {
         if (!isEmptyObject(this.currentItem)) {
             this.generateLabel(this.currentItem);
         }
+
+        if (this.props.items.length > 0 && prevProps.items.length === 0) {
+            this.updateMaxHeight();
+        }
```
hmhealey, line 65: This is the important part that was added

components/textbox/textbox.tsx
```diff
@@ -244,25 +244,17 @@ export default class Textbox extends React.PureComponent<Props> {
...
         let textWrapperClass = 'textarea-wrapper';
-        let wrapperHeight;
         if (this.props.emojiEnabled) {
             textboxClassName += ' custom-textarea--emoji-picker';
         }
         if (this.props.badConnection) {
             textboxClassName += ' bad-connection';
         }
-        if (this.wrapper.current) {
-            wrapperHeight = this.getInputBox()?.clientHeight;
```
hmhealey, line 264: And the important part that was removed

components/edit_post_modal/edit_post_modal.tsx
```diff
@@ -448,7 +448,6 @@ export class EditPostModal extends React.PureComponent<Props, State> {
     handleEntered = () => {
         if (this.editbox) {
             this.editbox.focus();
-            this.editbox.recalculateSize();
```
hmhealey, line 451: I had to remove this since the `EditPostModal` no longer has access to the `AutosizeTextarea`'s methods, but that's fine since this worked correctly without it (ie. it sized the modal and scrolled the textarea to the end on long posts)

jwilander (approved):
LGTM

jwilander:
@hmhealey updated the milestone to 5.39 since 5.38 is out of support

devinbinnie (approved):

jgilliam17:
@hmhealey 
Few tests for channel switcher are failing, results list is hidden, verified manually see image
Screen Shot 2021-12-01 at 1 18 46 PM

jgilliam17:
There seems to be an issue with the server, getting 504 an 502 server errors

hmhealey:
Alright, the following should be fixed now:
Screen Shot 2021-12-01 at 4 52 02 PM
Screen Shot 2021-12-01 at 4 53 36 PM
Screen Shot 2021-12-01 at 4 54 08 PM
Screen Shot 2021-12-01 at 4 55 57 PM
Screen Shot 2021-12-01 at 4 57 44 PM

The rest of the failures looked unrelated to this PR

jgilliam17:
Thanks @hmhealey 
Tested, looks good to merge.
- Verified autocomplete and channel switcher suggestion list layout.

jgilliam17 (approved):

PR #9452, opened 2021-11-30
Forward ref through QuickInput

hmhealey:
This PR is based off https://github.com/mattermost/mattermost-webapp/pull/9451. Changes specific to it can be found here: https://github.com/mattermost/mattermost-webapp/compare/MM-40071_suggestion-list-height...MM-40071_style-recalc

This ended up not being needed for the previous PR, but I figured it was good to add anyway. Since we have to use refs to access the various textboxes, this'll make those easier to use by returning the DOM node for the input instead of returning the React component that wraps it.

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/9451

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/quick_input/quick_input.test.tsx
+++ b/components/quick_input/quick_input.test.tsx
@@ -4,7 +4,7 @@
 import React from 'react';
 import {mount} from 'enzyme';
 
-import QuickInput from 'components/quick_input/index';
+import {QuickInput} from './quick_input';
 
 describe('components/QuickInput', () => {
     test.each([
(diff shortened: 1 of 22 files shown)


components/quick_input/quick_input.test.tsx
```diff
@@ -4,7 +4,7 @@
 import React from 'react';
 import {mount} from 'enzyme';
 
-import QuickInput from 'components/quick_input/index';
+import {QuickInput} from './quick_input';
```
M-ZubairAhmed, line 7: Question : should we prefer named exports over defaults

hmhealey: I think we should still generally prefer the default export. The reason I'm doing the named export here is to bypass the `forwardRef` so that the snapshot still shows the body of the component

M-ZubairAhmed (approved):

PR #9459, opened 2021-12-01
chore/MM-39785 : Migrate EditPostModal to be controlled by the ModalController

M-ZubairAhmed:
#### Summary
- [x] Migrated `edit_post_modal/index.js` to typescript
- [x] Made EditPostModal to be controlled by ModalController
- [x] Migrated from string ref to createRef in EditPostModal
- [x] Removed editpostmodal states from Store

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39785

#### Related Pull Requests
None

#### Screenshots
None

#### Release Note

```release-note
NONE
```


components/create_post/create_post.tsx
```diff
@@ -1135,16 +1132,23 @@ class CreatePost extends React.PureComponent<Props, State> {
...
-            type = Utils.localizeMessage('create_post.comment', Posts.MESSAGE_TYPES.COMMENT);
-        } else {
-            type = Utils.localizeMessage('create_post.post', Posts.MESSAGE_TYPES.POST);
-        }
         if (this.textboxRef.current) {
             this.textboxRef.current.blur();
         }
-        this.props.actions.setEditingPost(lastPost.id, 'post_textbox', type);
+
+        let modalTitle;
```
hmhealey, line 1139: Instead of passing the title to the modal, perhaps the modal could work this out itself?

hmhealey (changes requested):
Looks great! I especially appreciate the other cleanup here like removing the string refs and getting rid of some of that textbox state that is in redux for whatever reason.

I just have one suggestion to avoid repeating ourselves a bit when opening the modal

components/create_post/create_post.tsx
M-ZubairAhmed: ✅  Updated as per request/discussion

hmhealey:
I think this is fine to still merge this in the meantime. Inline post editing seems big enough that I imagine it'll be behind a feature flag for a bit while we let it soak for a bit, won't it?

michelengelen:
You are probably right with this... still... it will eventually be an effort in vain, since most of the stuff changed will be removed soon. And it will create another massive merge conflict. 😂

M-ZubairAhmed:
@michelengelen as per the new inline edit feature we are closing this one out

PR #9466, opened 2021-12-01
MM-40071 Attempt to improve performance by reducing direct DOM usage

hmhealey:
Similar to the related PRs, we're trying to improve some performance issues that a customer was reporting where the JS thread would frequently get blocked on style recalculations. As part of an effort to combat that, I've been removing stuff that we think blocks on the UI based on [this list](https://gist.github.com/paulirish/5d52fb081b3570c81e3a).

There's technically two sets of changes, but I'm trying to save some work because of how much cherry-picking is required for these. For the two commits:
1. The first one is avoiding accessing dimensions of stuff during render. I have a theory this is the worst culprit for those issues.
2. The second one is avoiding accessing that when its doing ref callbacks since I assume that's similarly bad. All of the cases I saw where we were doing that was for positioning menus when they're opened. I *think* doing this in `componentDidUpdate` after the render is better, but I feel like the way we were doing it before was kind of abusing ref callbacks a bit

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40071

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/9365
https://github.com/mattermost/mattermost-webapp/pull/9451

#### Release Note
```release-note
Improve performance by reducing DOM usage during render
```

### DIFF
--- a/components/dot_menu/__snapshots__/dot_menu.test.tsx.snap
+++ b/components/dot_menu/__snapshots__/dot_menu.test.tsx.snap
@@ -4,7 +4,7 @@ exports[`components/dot_menu/DotMenu should match snapshot, canDelete 1`] = `
 <MenuWrapper
   animationComponent={[Function]}
   className=""
-  onToggle={[MockFunction]}
+  onToggle={[Function]}
 >
   <OverlayTrigger
     className="hidden-xs"
@@ -141,7 +141,7 @@ exports[`components/dot_menu/DotMenu should match snapshot, on Center 1`] = `
 <MenuWrapper
   animationComponent={[Function]}
   className=""
-  onToggle={[MockFunction]}
+  onToggle={[Function]}
 >
   <OverlayTrigger
     className="hidden-xs"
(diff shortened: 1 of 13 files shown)


devinbinnie (approved):

nevyangelova (approved):
LGTM!

hmhealey:
@jgilliam17 For testing this one, the big thing to test is the positioning of the following menus: the post menu, channel sidebar menus, and the menu on file attachments. I moved the code that makes them open up/down and positions submenus in the sidebar ones, so there's a chance those broke.

The stuff I mentioned in the first point in the description actually has no effect here since the code isn't used on master. I just left it in to make cherry-picking easier

jgilliam17:
Thanks @hmhealey 
[E2E report](https://automation-dashboard.vercel.app/cycles/61e6f54c-be75-4a89-829b-3b8af2c2e550) 
Re-tested failures, none repro manually 
MM-T1684 - this test timed out waiting for channel header to be available so channel menu can be opened - would this be related to your changes?

hmhealey:
No, that wouldn't be related. Nothing in the channel header was affected by these changes

jgilliam17:
Great! Thanks for checking @hmhealey 
Tested, looks good to merge.
-Verified post menu, channel sidebar menus, and the menu on file attachments open as expected up/down depending on their position in the browser.

jgilliam17 (approved):

PR #9467, opened 2021-12-01
Switch to using `window.matchMedia` instead of the window `resize` event to determine current media query breakpoint.

deanwhillier:
#### Summary
This PR switches the mechanism for determining the current window breakpoint size to use `window.matchMedia`.  `window.matchMedia` only fires its `change` event a single time when the window size crosses a defined media query boundary. With this change, the `emitBrowserWindowResized` action is only called once when the active window breakpoint changes instead of every 100ms whiles the window is resizing.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/root/root.jsx
+++ b/components/root/root.jsx
@@ -6,7 +6,6 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import FastClick from 'fastclick';
 import {Route, Switch, Redirect} from 'react-router-dom';
-import throttle from 'lodash/throttle';
 
 import {rudderAnalytics, RudderTelemetryHandler} from 'mattermost-redux/client/rudder';
 import {Client4} from 'mattermost-redux/client';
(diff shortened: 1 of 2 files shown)


components/root/root.jsx
```diff
@@ -96,6 +95,11 @@ const LoggedInRoute = ({component: Component, ...rest}) => (
     />
 );
 
+const desktopMediaQuery = window.matchMedia(`(min-width: ${Constants.DESKTOP_SCREEN_WIDTH + 1}px)`);
```
jwilander, line 98: Just double checking that `window.matchMedia` is supported on all the browsers (+electon) we support? I only ask because `window.matchMedia` is unfamiliar to me and I know we've had issues with browser API support before.

jwilander (approved):
LGTM - only one question

components/root/root.jsx
deanwhillier: Good question! I did a [cursory check](https://caniuse.com/?search=MediaQueryList) before using and support is surprisingly good! A more detailed look shows that a more recent version of Safari (less than v14) needed to use `addListener` instead of `addEventListener`. Updated to support fallbacks including the throttled window.resize event.

deanwhillier:
Tweaked a bit to be able to test ... JSDOM (used by Jest) doesn't support window.matchMedia.

jwilander (approved):
LGTM

hmhealey (approved):
The `addListener` vs `addEventListener` thing is a bit annoying to me, but I guess we have to live with it until we update the minimum version of Safari

jgilliam17:
@deanwhillier can you please update? I'd like to trigger an E2E run and check the report before approving. Thanks.

jgilliam17 (approved):

PR #9494, opened 2021-12-05
Splitting the OpenInvite section of team_general_tab into another component

jespino:
#### Summary
Splitting the OpenInvite section of team_general_tab into another component

#### Ticket Link
[MM-40556](https://mattermost.atlassian.net/browse/MM-40556)

#### Release Note
```release-note
NONE
```


jespino:
@hmhealey @michelengelen This is my idea, gradually migrate this way. This have multiple advantages.

1. We migrate to typescript part of that big files.
2. We migrate to functional components part of that big files.
3. We reduce the original file, so later we are going to be able to migrate that to typescript and functional component way easier.

The idea is to keep doing this small PRs all over the application, reducing that files to a more declarative, functional and simple approach.

components/team_general_tab/open_invite.tsx
```diff
@@ -0,0 +1,148 @@
...
+    isActive: boolean;
+    isGroupConstrained?: boolean;
+    allowOpenInvite?: boolean;
+    onToggle: (active: boolean) => void;
+    patchTeam: (patch: Partial<Team>) => Promise<ActionResult>;
+};
+
+const OpenInvite = (props: Props) => {
+    const intl = useIntl();
+    const [serverError, setServerError] = useState('');
```
michelengelen, line 25: nit: 1/5 I would prefer to destructure the relevant properties from `props`, but I am not certain how this is being handled over the codebase in general ... I tend to use my personal rule-of-thumb: "if you write `props.x` or `this.props.x` it is best to destructure `x` from it"

components/team_general_tab/team_general_tab.jsx
```diff
@@ -823,7 +708,20 @@ export default class GeneralTab extends React.PureComponent {
                         </>
                     }
                     <div className='divider-light'/>
-                    {openInviteSection}
+                    <OpenInvite
+                        teamId={this.props.team.id}
+                        isActive={this.props.activeSection === 'open_invite'}
+                        isGroupConstrained={this.props.team.group_constrained}
+                        allowOpenInvite={this.props.team.allow_open_invite}
+                        onToggle={(active) => {
```
michelengelen, line 716: Right now this is fine, but if you plan to extract more of those parts it would probably a good idea to have some kind of generator function, like this:

```typescript jsx
const getOnToggle = (section) => (active) => this.handleUpdateSection(active ? section : '');
```

and then use it like this: 

```typescript jsx
...
onToggle={getOnToggle('open_invite')}
...
```

michelengelen (approved):
LGTM ... just some considerations!

components/team_general_tab/team_general_tab.jsx
hmhealey: Note that both the current way we're doing it and the proposed way generate new functions and cause the OpenInvite to re-render whenever the TeamGeneralTab does. I don't think this code is exactly performance-sensitive though, and we can always memoize that with `useCallback` once all of this is made into a functional component though

hmhealey (approved):
Looks good! We've already been doing that with some of the settings items in other modals, but the team settings modal is kind of neglected since most people don't see it, so I definitely appreciate the work on it.

Hopefully we'll also be able to simplify each section's render method in the future to avoid having to write so much raw HTML (like the fieldset tag and such) in the future too

components/team_general_tab/team_general_tab.jsx
michelengelen: You can also wrap the component into `React.memo` to prevent unnecessary re-renders. But the proposed way of adding `useCallback` should do the trick as well

components/team_general_tab/open_invite.tsx
jespino: fair enough, I'll do the changes.

michelengelen: i just noticed i forgot an important piece from my rule-of-thumb: "if you write `props.x` or `this.props.x` more than twice it is best to destructure `x` from it"

Sorry for missing that while writing the comment! 🙇🏼‍♂️

components/team_general_tab/team_general_tab.jsx
hmhealey: `React.memo` on its own won't fix the case where the `TeamGeneralTab` re-rendering causes the `OpenInvite` to re-render since it's still receiving a new version of `onToggle` each time

michelengelen: @hmhealey you are absolutely right.

jgilliam17:
Thanks @jespino 
Tested, looks good to merge.
- Verified general team settings modal and open invite functionality - as before.
- [E2Es report](https://mattermost-cypress-report.s3.amazonaws.com/98347-20d5e30-onprem-ent-webapp-pr-9494-mm-ee-test:20d5e30/mochawesome.html) looks good - no PR related failures.

jgilliam17 (approved):

PR #9495, opened 2021-12-06
MM-40418: fix leave/archive channel

koox00:
#### Summary

Leaving or Archiving a channel, while having a thread that belongs in
that channel open in the RHS, crashes the app.

This commit fixes the crash, and also closes the RHS, if it has to, upon aforementioned actions.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40418

#### Release Note

```release-note
Fixes crash upon leaving a channel while having a thread of that channel open
```


amyblais:
@hmhealey  @nevyangelova FYI, code freeze for v6.2 is coming this Thursday, please help review by then if you can.

actions/views/channel.test.js
```diff
@@ -103,6 +104,9 @@ describe('channel view actions', () => {
                 loadingPosts: {},
                 postVisibility: {current_channel_id: 60},
             },
+            rhs: {
+                selectedPostId: '',
```
hmhealey, line 108: Could you add a couple tests for this new behaviour to ensure it doesn't get broken accidentally in the future? They could be unit tests in here or E2E tests

components/rhs_header_post/index.ts
```diff
@@ -33,15 +34,21 @@ import {matchUserMentionTriggersWithMessageMentions} from '../../utils/post_util
...
 function mapStateToProps(state: GlobalState, {rootPostId}: OwnProps) {
     const root = getPost(state, rootPostId);
     const currentUserMentionKeys = getCurrentUserMentionKeys(state);
-    const rootMessageMentionKeys = allAtMentions(root.message);
-    const thread = getThreadOrSynthetic(state, root);
-    const isMentionedInRootPost = thread.reply_count === 0 &&
-        matchUserMentionTriggersWithMessageMentions(currentUserMentionKeys, rootMessageMentionKeys);
+
+    let isMentionedInRootPost = false;
+    let thread = {} as UserThreadOrSythetic;
```
hmhealey, line 45: Instead of making a fake object here, could we perhaps make it so that `getThreadOrSynthetic` can handle the case where `root` is undefined? This could be making the `RhsRootPost` render constantly by passing a new object on each state change, and I'd be worried about other things failing because the fields of this empty object are undefined.

hmhealey (changes requested):

nevyangelova (approved):
LGTM -might be a good idea to add some unit tests for this :)

components/rhs_header_post/index.ts
koox00: Changing `getThreadOrSynthetic` would need to change every use to handle the `null` case,
so I'll refactor a bit to handle the new object.

koox00: BTW, the root cause of this is that `getSelectedPost` can return a fake post which is desired.
https://github.com/mattermost/mattermost-webapp/blob/a468fdb80e61c876462ed56f133c562fc95db737/selectors/rhs.ts#L64-L85

actions/views/channel.test.js
koox00: Now that we are at it, I want to bring something on the table, 
we have many snapshot tests, that are not testing anything at all...
Snapshot tests IMO should use the `react-test-renderer` and never `shallow` mount.

e.g: https://github.com/mattermost/mattermost-webapp/blob/a468fdb80e61c876462ed56f133c562fc95db737/components/rhs_thread/__snapshots__/rhs_thread.test.tsx.snap#L3-L40

The above test would have caught the regression, but did not.
I get that mounting components requires extra work, but I'm positive that it would be beneficial.

koox00: I've added a new e2e test for leaving a channel while RHS is open.
@jgilliam17 can you create a zephyr test case for 
https://github.com/mattermost/mattermost-webapp/pull/9495/files#diff-c308c925a9f30488189bdc1fa95155f822d2135aad3abe2d34ffa95b1fb07640R38

e2e/cypress/integration/channel/leave_channel_spec.js
```diff
@@ -0,0 +1,110 @@
...
+                testChannel = channel;
+            });
+        });
+    });
+
+    beforeEach(() => {
+        cy.visit(`/${testTeam.name}/channels/${testChannel.name}`);
+    });
+
+    it('Leave a channel while RHS is open', () => {
```
koox00, line 44: @jgilliam17 can you add a zephyr test case for this one?

e2e/cypress/integration/channel/leave_channel_spec.js
```diff
@@ -0,0 +1,110 @@
...
+            // * RHS should not be visible
+            cy.get('#rhsContainer').should('not.exist');
+
+            // * Assert that user is redirected to townsquare
+            cy.url().should('include', '/channels/town-square');
+            cy.get('#channelHeaderTitle').should('be.visible').and('contain', 'Town Square');
+        });
+    });
+
+    it('Leave a channel while RHS is open and CRT on', () => {
```
koox00, line 76: @jgilliam17 and for this one?

jgilliam17 (commented):
@koox00 test case created in Zephyr and IDs suggested.
I'll re-run E2Es after update.
Current E2E report looks good, no PR related failures. https://automation-dashboard.vercel.app/cycles/d98b093a-cf12-4806-ab06-5aee3eda22e1

jgilliam17:
Thanks @koox00 
Tested, looks good to merge.
-Verified when RHS is open, leave or archive channel is not causing a crash. Tested with and without CRT.
Re-running E2Es here https://automation-dashboard.vercel.app/cycles/b4c9ee5d-89d0-4632-9ad4-a02b00a7a55b
🤦 I missed the passing tests in my previous report search
Screen Shot 2021-12-08 at 1 48 54 PM
I will verify that test IDs are added in the new report.

jgilliam17 (approved):

koox00:
@jgilliam17 can you re-run the E2E tests?

actions/views/channel.test.js
hmhealey: I don't know too much about `react-test-renderer`, but I assume it's similar to Enzyme's `mount`. Is that correct?

Regardless, I'm not a fan of snapshot tests myself since I find they either break due to unrelated changes (especially when using snapshots and `mount` together) or they make don't make it clear what they're looking for. I do like using `mount` and the enzyme functionality as a lighter weight version of what the E2E tests provide though.

koox00: `react-test-renderer` is like `mount` but also offers a `toJSON`[2] so that snapshot[2] testing makes sense.
Searching a bit it seems there is a plugin[3] for `enzyme` to offer serialize a component to JSON, making snapshot output a little more meaningful. 

[1]: jest snapshot testing https://jestjs.io/docs/snapshot-testing#snapshot-testing-with-jest
[2]: react-test-renderer https://reactjs.org/docs/test-renderer.html#testrenderertojson
[3]: enzyme-to-json https://github.com/adriantoine/enzyme-to-json

components/rhs_header_post/index.ts
hmhealey: I meant that we could use the fake thread returned by `getThreadOrSynthetic` instead of making a new empty object here by adding an extra null check. I'm not the biggest fan of fake objects, but at least the one returned by `getThreadOrSynthetic` has all the fields provided so that we won't accidentally cause errors in code using the object.

```
export function getThreadOrSynthetic(state, rootPost) {
    const thread = getThreads(state)[rootPost?.id]; // Adding ?. here lets rootPost be undefined

    if (thread?.id) {
        return thread;
    }

    return {/* something else fake but with all the required fields*/};
}
```

hmhealey (approved):

jgilliam17:
Thanks @koox00 
Re-tested - looks good to merge.  
Reviewed latest E2E report, no PR related failures (tested manually 2 tests that differ from the master run report - not reproducing)

jgilliam17 (approved):

PR #9500, opened 2021-12-07
fix/MM-40381, MM-40381, MM-40381 : Emoji picker issues with custom emojis

M-ZubairAhmed:
#### Summary
- [x] Fixes the app crash when recently used custom emoji is deleted
- [x] Fixes the UI of the post with only one custom emoji reaction later is deleted
- [x] Fixes custom emojis not showing up in recent emojis area
- [x] Fixes custom emoji getting on top of one another in custom emoji category

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40381
https://mattermost.atlassian.net/browse/MM-40512
https://mattermost.atlassian.net/browse/MM-40132

#### Related Pull Requests
None

#### Screenshots
|  before  |  after  |
|----|----|
| image | image |

#### Release Note

```release-note
NONE
```


amyblais:
@M-ZubairAhmed There are some failing tests.

selectors/emojis.ts
```diff
@@ -54,10 +54,7 @@ export function isCustomEmojiEnabled(state: GlobalState) {
 export const getOneClickReactionEmojis = (state: GlobalState) => {
     const recentEmojis = getRecentEmojis(state).slice(-3);
     const emojiMap = getEmojiMap(state);
-    const emojis = [];
-    for (let i = 0; i < recentEmojis.length; i++) {
-        emojis.push(emojiMap.get(recentEmojis[i]));
-    }
 
-    return emojis.reverse();
+    const emojis = recentEmojis.slice(-3).map((recentEmoji) => emojiMap.get(recentEmoji)).filter(Boolean).reverse();
```
hmhealey, line 58: Did you mean to slice this a second time?

Also, can you convert this whole thing into a memoized selector? It's returning a new array each time currently which'll likely affect performance

packages/mattermost-redux/src/selectors/entities/posts.ts
```diff
@@ -484,7 +495,7 @@ export const getLastPostPerChannel: (state: GlobalState) => RelationOneToOne<Cha
             }
 
             const postId = recentBlock.order[0];
-            if (allPosts.hasOwnProperty(postId)) {
+            if (Object.prototype.hasOwnProperty.call(allPosts, postId)) {
```
hmhealey, line 498: Not something we need to change here, but I think we might want to disable this ESLint warning. It's only really useful if you're worried about people overwriting `hasOwnProperty` or messing up the prototype of an object, and I think we can probably agree that this syntax is uglier than the old one.

packages/mattermost-redux/src/selectors/entities/posts.ts
```diff
@@ -63,16 +65,27 @@ export function getReactionsForPosts(state: GlobalState): RelationOneToOne<Post,
...
-            return reactions[postId];
-        }
+export function makeGetReactionsForPost(): (state: GlobalState, postId: Post['id']) => Record<string, Reaction> | undefined | null {
+    return createSelector(
+        'makeGetReactionsForPost',
+        getEmojiMap,
+        getReactionsForPosts,
+        (_: GlobalState, postId: string) => postId,
+        (emojiMap, reactions, postId) => {
+            if (reactions[postId]) {
```
hmhealey, line 75: Minor nit, we could do an early return if `!reactions[postId]` to avoid indenting as much of this

components/emoji_picker/emoji_picker.jsx
```diff
@@ -554,43 +554,8 @@ export default class EmojiPicker extends React.PureComponent {
...
-
-        return this.filterDistinctEmojis(emojis);
-    }
-
-    getEmojiByUnicode = (unicode) => {
-        const emojiIndex = Emoji.EmojiIndicesByUnicode.get(unicode.toLowerCase());
-        return Emoji.Emojis[emojiIndex];
-    }
-
-    convertRecentEmojiToUserSkinTone = (category, emojiId) => {
```
hmhealey, line 568: Is this reverting the code that makes recent emojis always fit the user's selected skintone, or does that still work?

hmhealey (changes requested):

components/emoji_picker/emoji_picker.jsx
M-ZubairAhmed: i improved the old code for recent ones to use skintone

selectors/emojis.ts
M-ZubairAhmed: - Removed the double slice
- So i had converted this to memoized selector in the start, but it wont fire again when there is change in the recent emoji select as it has to get the new from the local storage

packages/mattermost-redux/src/selectors/entities/posts.ts
M-ZubairAhmed: Reverted ↩️

components/emoji_picker/emoji_picker.jsx
M-ZubairAhmed: Yes i improve the code to selected skin tone only in recent section of emoji picker

packages/mattermost-redux/src/selectors/entities/posts.ts
M-ZubairAhmed: ✅  Updated as per request/discussion

components/emoji_picker/emoji_picker.jsx
M-ZubairAhmed: also due to git diff, code seems to be all the place, please check out the file in single mode if its gets too confusing with code block end etc

M-ZubairAhmed: please try to check out the file in the single mode, since git diff shows too much mess around

selectors/emojis.ts
hmhealey: Shouldn't `getRecentEmojis` return a new result in that case though? Maybe I'm missing something because I know the storage code doesn't always work the way it should

hmhealey (approved):

packages/mattermost-redux/src/selectors/entities/posts.ts
```diff
@@ -41,6 +41,8 @@ import {
 } from 'mattermost-redux/utils/post_utils';
 import {getPreferenceKey} from 'mattermost-redux/utils/preference_utils';
 
+import {getEmojiMap} from 'selectors/emojis';
```
M-ZubairAhmed, line 44: > Instead of getEmojiMap, you can use the getSystemEmojis and the getCustomEmojis selectors separately there. I can't remember why system emojis are part of the web app, but to allow mattermost-redux to still access those, we had the web app save them as a global variable that mattermost-redux can see. And getSystemEmojis basically returns that variable

M-ZubairAhmed: @hmhealey created a new `makeGetUniqueReactionsToPost` inside the webapp, which uses internally `makeGetReactionsForPost`. Since the latter resides in `mattermost-redux` and doesnt have access to emoji map,  i had to create a new one. Let me know if this is ok and placement of the function is in correct file.

components/emoji_picker/emoji_picker.jsx
```diff
@@ -538,6 +539,17 @@ export default class EmojiPicker extends React.PureComponent {
         ];
     }
 
+    convertEmojiToUserSkinTone = (emoji, emojiSkin, userSkinTone) => {
+        if (emojiSkin === userSkinTone) {
+            return emoji;
+        }
+
+        const newEmojiId = userSkinTone === 'default' ? emoji.unified.replace(`-${emojiSkin}`, '') : emoji.unified.replace(emojiSkin, userSkinTone);
```
Willyfrog, line 547: for the default sking, you can get it directly from the `source_index` property we created in `getSkinVariations` function.

When I added the code for skintones, I always thought it was going to be a one switch process, and not having to move across skintones, but given this I would suggest to add a mapping for skin variations in https://github.com/mattermost/mattermost-webapp/blob/4789a05e90fd1c000edcbca91fe2ef25695c6b50/build/emoji/emoji_gen.mjs#L165 in that loop we get the default and the skin variations, so maybe we could generate the mapping once and attach it to the emoji so we could get the variations without trying to calculate something like `emoji.variation['medium_light']` and that would return the index of the variant.

I'm not sure if this change is too big for the PR, though, 0/5

Willyfrog (approved):
added  a suggestion, but otherwise looks good

components/emoji_picker/emoji_picker.jsx
M-ZubairAhmed: Hey lets to that in seperate PR since we already have a ticket open for improving `emoji.mjs` file

jgilliam17:
E2Es running here https://automation-dashboard.vercel.app/cycles/587cb79a-3783-4af9-9ddf-4f08d94dad43

hmhealey (approved):

jgilliam17 (approved):
Thanks @M-ZubairAhmed 
Tested, looks good to merge. E2E report looks good too, no PR related failures.
- Verified removing a custom emoji, that was previously used is no longer crashing the app. (MM-40381)
- Verified, once removed in system console, custom emoji is removed from the post reactions and + to add reactions is also removed. (MM-40512)
- Verified after use, custom emoji are present in the recent emoji area and quick reactions on post hover, also all custom emoji display in the custom category on emoji picker. (40132)

PR #9511, opened 2021-12-08
[MM-35298] Send RootId with add to channel request

ashishbhate:
#### Summary

- Correctly update mention counts and follow state of a thread when a user is added to a channel from a thread mention in the RHS. To do this we send the RootID with the Add to Channel API request

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-35298

#### Related Pull Requests

- Has server changes - https://github.com/mattermost/mattermost-server/pull/19311

#### Release Note

```release-note
NONE
```


components/threading/virtualized_thread_viewer/thread_viewer_row.tsx
```diff
@@ -64,6 +67,15 @@ function ThreadViewerRow({
...
     default:
+        if (PostListUtils.isCombinedUserActivityPost(listId)) {
+            return (
+                <CombinedUserActivityPost
+                    combinedId={listId}
+                    previousPostId={getPreviousPostId}
+                    isLastPost={isLastPost}
+                />
+            );
+        }
```
ashishbhate, line 78: @koox00 I added this here to support `UserActivityPost` in the RHS.  Does this look ok to you? 

Should the "XXX added to the channel by you." be in the RHS, or in the center channel? It appears in the RHS now because the channel join request now includes the `rootId`, so the webapp considers it part of the thread.

An alternative fix would be to just send all `UserActivityPost` to the center channel instead.

koox00: We should show it in the RHS, you might mention somebody from the global threads view

Willyfrog (approved):

jgilliam17:
@ashishbhate 
I am not seeing a mention badge on the channel user was added to via rhs reply mention
On the image `new` is the channel with original thread 
mention badge+bolding  is present on the global threads items
Screen Shot 2022-01-25 at 4 17 04 PM

ashishbhate:
@jgilliam17 The reason the channel has no badge is because the "XXX was added to channel" message is posted in the thread, and not a root post in the channel. Mentions in thread replies won't cause in a channel badge. 
I'm not sure if there's a way to work around this, except to double post the channel join message in the channel and the thread.

// cc @koox00

koox00:
> @jgilliam17 The reason the channel has no badge is because the "XXX was added to channel" message is posted in the thread, and not a root post in the channel. Mentions in thread replies won't cause in a channel badge.
I'm not sure if there's a way to work around this, except to double post the channel join message in the channel and the thread.

IMO the expected would be to bold the threads and not the channel, since that's were the user should focus.

// cc @esethna

esethna:
@jgilliam17 I think there's an unrelated bug here where the plus button is missing for a new user. This is really bad, as a new user is missing all the functions that would be under that button (new category, browse channels, etc). Can you help isolate this issue? We've had some reports of this before but never consistently could repro as far as I know:

image

jgilliam17:
Not reproducing for me. I added two users, one invited & other signed up on login, and both had the + icon with options. Will have to look into that further. If this is something other users are experiencing it won't be exclusive to this PR.

esethna:
Ah understood. Let's merge this with the current behaviour and iterate if needed @jgilliam17

jgilliam17:
Thanks @ashishbhate 
Tested, looks good to merge.
- Verified mentioning a user in a thread and adding them to channel on the RHS, autofollows the thread for that user.
- [E2E report](https://mattermost-cypress-report.s3.amazonaws.com/148602-2c33539-onprem-ent-webapp-pr-9511-mattermost/mm-ee-test2c33539/mochawesome.html) looks good, no PR related failures.

jgilliam17 (approved):

PR #9513, opened 2021-12-09
MM-40378 Plugin is incorrectly displayed as enabled after re-install

frnkshin:
#### Summary
- Added call to dispatch DISABLE*_PLUGIN* actions to update state of the removed plugin to sync with server-side API handler's behavior.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/19108

#### Release Note
```release-note
Reinstalling a previously-enabled plugin now correctly reports enabled status as false.
```


components/admin_console/plugin_management/plugin_management.tsx
```diff
@@ -700,12 +700,17 @@ export default class PluginManagement extends AdminSettings<Props, State> {
     handleRemove = async () => {
         this.setState({lastMessage: null, serverError: null});
         if (this.state.removing !== null) {
-            const {error} = await this.props.actions.removePlugin(this.state.removing);
-            this.setState({removing: null});
+            const {disableError} = await this.props.actions.disablePlugin(this.state.removing)
```
hanzei, line 703: There is a small linter issue:
```suggestion
            const {disableError} = await this.props.actions.disablePlugin(this.state.removing);
```

hanzei, line 703: I don't think the bug in https://github.com/mattermost/mattermost-server/issues/19108 is caused by a wrong state on the server side i.e. the plugin is in fact disabled after an uninstall. It's more of an sync issue between the webapp and server. If you take a look at http://localhost:8065/api/v4/plugins/statuses after the plugin has been re-installed, it states that the plugin is in fact disabled i.e. has `state` `0`.

hanzei (changes requested):
Thanks for working on this @frnkshin !

components/admin_console/plugin_management/plugin_management.tsx
frnkshin: @hanzei
Thanks for your input! I wasn't aware that this calls the API. I updated the code with calling the dispatch with `DISABLE*_PLUGIN*` when `removePlugin` is called in the client to update the status.

hanzei (approved):
LGTM :+1:

components/admin_console/plugin_management/plugin_management.tsx
hanzei: Sounds reasonable to me :+1:

larkox (approved):
LGTM.

hanzei:
@DHaussermann Gentle reminder to review this PR

DHaussermann (approved):
Tested and passed 
- Confirmed the correct inactive state is always reflected in plugin config without a browser refresh
- Confirmed the changes is functional with multiple plugins including one that enabvle with Feature flag such as Boards
- Confirmed this changes also solves the issue on HA servers

LGTM!

Huge thanks @frnkshin for this PR 🎉  This issue was annoying. Your work to fix it is much appreciated!

PR #9524, opened 2021-12-12
[MM-41321] Add keyboard shortcut to mark channel as read

3ach:
#### Summary

Adds a keyboard shortcut (Alt/Option-L) to mark a channel as read. 

#### Ticket Link

Jira ticket: https://mattermost.atlassian.net/browse/MM-41321
The substantive changes here are under 20 lines, though the overall pull request is larger for documentation + test updates. 

#### Related Pull Requests

N/A

#### Screenshots

|  before  |  after  |
|----|----|
|  |  |

#### Release Note

```release-note
Added a new keyboard shortcut to mark a channel as read.
```


3ach:
Sorry, this isn't ready for review yet, I was just putting it up so that I could get some idea of what tests and things will need fixing.

hmhealey:
Thanks for the PR, @3ach! I've added @esethna as a reviewer since he's got more of an idea of what features we want to add. I'm not sure I understand the use case for marking the current channel as read though. That's done automatically when viewing a channel (as long as the user hasn't manually marked it as unread), so I don't know when a user would want to do that. 

If that's not something we want to add though, I could see taking this in a slightly different direction though by either changing it to a shortcut to mark all channels as read since we have [a ticket for that](https://mattermost.atlassian.net/browse/MM-2541) (and [a second one](https://mattermost.atlassian.net/browse/MM-12111) actually, or making it a shortcut that either marks the channel as read or unread. Both of those would obviously require additional work though, so I'm curious to know if I'm missing something first with how you'd want to use this.

esethna:
+1 to Harrison's thoughts @3ach. If you're keen to adapt this contribution to either of these tickets we'd be happy to accept it.

- Shortcut to mark all channels as read: https://mattermost.atlassian.net/browse/MM-2541
- "Mark as unread" added to the channel ... menu in the sidebar: https://mattermost.atlassian.net/browse/MM-28016

If not, can you help us understand your use case a bit more?

3ach:
The specific workflow I have here is for monitoring a single channel with a fairly high amount of chatter. When I first enter the channel, I have an orange line marking the top of messages sent since I was last in the channel. That line moves up as more messages enter the channel, but I wanted a way to mark that I've read the channel to its bottom _now_, and would like the marker to reappear if more messages come in (so if I flip to a different desktop, then back, I can easily catch up). 

Right now, the only way I've found to do this is to click into a new channel, and come back to my original one. This works but is a little annoying; I am currently migrating from IRC and I know that a similar shortcut exists in at least four separate clients. I also then either need to make sure I read all the messages I'm marking as read in the second channel (by opening it) or find a channel with no unread messages as the secondary channel. 


It's not the same as marking all channels as read (though I'm happy to add a shortcut to do so, in this PR or another.) I wouldn't consider mark-all as a replacement for mark-this. 

I can also add a 'mark unread' to the ... menu in the sidebar, which would do the same thing, but I was kind of hoping for a keyboard shortcut for this. Overall, I'd love to add more keyboard shortcuts throughout Mattermost, and if there's a good place to discuss the overall plan around those, I'd be happy to join.

hmhealey:
Okay, I can understand that. I don't know if I've run into that specific case in Mattermost myself, but it's probably because we have so many channels that I'm switching a lot to constantly clear that line. I think I might've also wanted that feature in Discord in the past though since I'm used to using fewer channels there.

Is the alt+L shortcut taken from somewhere else? I'm wondering if there's a standard hotkey for this sort of functionality across other apps. Slack and Discord just use ESC for it, but we'd have to make sure closing modals and autocompletes don't also mark the channel as read if we go with that.

3ach:
There's not a consistent hotkey across other clients that I know of. I didn't know Slack and Discord used ESC, we could do that (though it's a little overloaded... and my intuition for escape would be that it closes the current channel somehow.) Your points about making sure ESC is not misinterpreted also seems tricky, since right now keyboard shortcuts are kind of scattered all over the codebase for the webapp. It would be awesome if they were all in one place, and much easier to avoid conflicts.

I mostly only have experience with IRC clients, and it's all over the map (Cmd+L/Cmd+M are the most common. The reason I chose Alt+L instead of Cmd/Ctrl+L was that Cmd/Ctrl+L already has an extremely well-accepted meaning in browsers -- focus the address bar. So, Alt+L.

esethna:
Thanks for explaining your use case, makes a lot of sense now! +1 to using ESC given it's common usage in Discord and Slack. Since we also plan to implement the "mark all channels as read" with Shift + ESC, I think it works well to use ESC for this (as long as we can make sure it only takes effect if no modals or menus are open). 

@3ach feel free to also take on the [mark all channels as read shortcut](https://mattermost.atlassian.net/browse/MM-2541) if you're keen to continue this theme after this improvement

hmhealey:
Alright, in that case, I think we should switch the hotkey to the escape key for consistency since many of our users comes from Slack, Discord, and such. For handling any potential conflicts which could come from menus, autocompletes, or modals, I think we might be able to handle that by either:
1. Having those stop propagation of the event or using `preventDefault` which the mark as read shortcut watches for
2. In the mark as read shortcut handler, have it look at `document.activeElement` and walk up the DOM tree to see if it's a child of any of those components. For example, I think our menus that close on escape are all wrapped with a `MenuWrapper` class.
I think either way of doing that would work, so you could go with whichever seems better to you.

As for the mark all as read hotkey, I think we can leave that for a separate PR. I think it'll require adding a new API to the server, so it'd be better to handle that separately.

3ach:
OK, ready for review again. I've split this into two commits: one adding `stopPropagation` to handlers which may consume the event first, so that their running will prevent the read marker from being cleared. Then, the second commit actually adds the marker.

components/widgets/menu/menu_wrapper.tsx
```diff
@@ -68,6 +68,10 @@ export default class MenuWrapper extends React.PureComponent<Props, State> {
     }
 
     private keyboardClose = (e: KeyboardEvent) => {
+        if (this.state.open) {
+            e.stopPropagation();
+        }
```
hmhealey, line 73: I think this one should probably be moved into the if block below just to avoid swallowing other key presses that we might want handled while the menu is still open

hmhealey (changes requested):
Nice job! I hadn't realized some of these elements like the NewRepliesBanner actually used the escape key. I just have one suggestion for something to change, but the rest looks good to me

3ach:
@hmhealey d'oh. Sorry, updated, thanks for catching it.

Yeah, I do sort of worry about more collisions if we continue to want more keyboard shortcuts. These are spread all over the components, there might be a better way to centralize/coordinate them to avoid having weird behavior in the future. Happy to discuss further if you're interested :)

AshishDhama:
> @hmhealey d'oh. Sorry, updated, thanks for catching it.
> 
> Yeah, I do sort of worry about more collisions if we continue to want more keyboard shortcuts. These are spread all over the components, there might be a better way to centralize/coordinate them to avoid having weird behavior in the future. Happy to discuss further if you're interested :)

@3ach We are actually looking forward to doing something like this mostly with the help of Awesome community members like you and maybe using a library like https://www.npmjs.com/package/react-hotkeys. But still, open for discussion. Would love to discuss this further on our [community](https://community.mattermost.com/) server (Drop me a message @ashish.dhama). Thanks a lot.

AshishDhama (changes requested):
@3ach There are a few Lint issues could you fix those. Otherwise looks good to me. Thanks for your contribution.

esethna:
We probably need QA on this. It didn't seem to work for me. Is cleared the new message line on ESC, but the sidebar didn't update (to unbold the channel after marking a post as unread and then hitting ESC), and then when i returned to the channel the new messages line reappeared

cc// @jgilliam17

esethna (approved):
Issue I mentioned above has been addressed. Thanks @3ach!

3ach:
(force-push to fix tests)

hmhealey (approved):
Thanks for making that change. I'm just going to update the branch to the latest master to fix that failing CI step. We've actually removed that one on master

AshishDhama:
@3ach Could you please update the failing test, Otherwise looks good to me.

3ach:
Rebased + fixed snapshot.

AshishDhama (approved):
LGTM!!

jgilliam17:
Thanks @3ach 
- Thanks you for your contribution, would you be able to add an automated cypress test for this? 
I created a test case [MM-T4608](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/MM-T4608) MM-T4608 should be used as the ID for the automated test.
- [E2E repor](https://mattermost-cypress-report.s3.amazonaws.com/108113-4c9a72e-onprem-ent-webapp-pr-9524-mattermost/mm-ee-test4c9a72e/mochawesome.html)t looks good, no PR related failures.
- Seeing one issue when using `esc` 
When using esc to closing modals it is also marking the channel as read

jgilliam17:
@3ach Let me know if you have any questions about one remaining issue mentioned [here](https://github.com/mattermost/mattermost-webapp/pull/9524#issuecomment-1024379346). Thanks

3ach:
I will submit a test, I have just gotten busy. I should have time in the second half of this or early next week.

jgilliam17:
Hello @3ach Let us know if you need help getting this to finish line 🙂

PR #9533, opened 2021-12-13
Proposed changes to "Contact sales" area title

jasonblais:
Based on PMM proposal, adjust `Need to increase your headcount?` to `Need to increase user count?`

 - "headcount" is not a preferred term over seats or users
 - shortening text slightly to fit within one line

image

```release-note
NONE
```

cc @annelieseklein for your review, plus if we'd need to reflect it in our design file


pablovelezvidal (commented):
thanks @jasonblais , this change requires also a text modification in `i18n > en.json` line 1225. https://github.com/mattermost/mattermost-webapp/blob/master/i18n/en.json#L1225

jasonblais:
@pablovelezvidal thanks! Good catch, updated

pablovelezvidal (approved):
LGTM thanks @jasonblais! :shipit:

stevemudie (approved):
New text appears as expected.

AGMETEOR (approved):

annelieseklein (approved):

PR #9536, opened 2021-12-14
MM-39595 Banner Scenarios of trial period

vinod-demansol:
#### Summary
Banner scenario's of trial period

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39595

#### Release Note
```release-note
NONE
```


e2e/cypress/integration/enterprise/cloud/billing/banner_trial_spec.js
```diff
@@ -0,0 +1,117 @@
...
+
+    describe('Banner - UI validations', () => {
+        it('MM-T4422 Cloud trial prompts for admin login', () => {
+            // # Login as sysadmin
+            cy.apiAdminLogin();
+
+            cy.apiInitSetup().then(({team}) => {
+                // # Go to town square
+                cy.visit(`/${team.name}/channels/town-square`);
+                cy.waitForResources();
```
hmhealey, line 27: What are we waiting for when we call this with no arguments? And why do we need to wait for background requests when we could just wait for the app to render like other tests do?

e2e/cypress/support/ui/cloud_billing.js
```diff
@@ -7,3 +7,55 @@ Cypress.Commands.add('uiGetPaymentCardInput', () => {
...
+                    );
+                    if (found.length === 0) {
+                        return false;
+                    }
+                    return !resource.number || found.length >= resource.number;
+                },
+            );
+
+            if (allFilesFound) {
+                if (resourcesLoaded.length === resourcesLengthPrevious) {
```
hmhealey, line 40: I don't think I understand this logic. Why aren't we just resolving immediately once all the files are found?

hmhealey (approved):
Thanks for the PR! I have a few questions about `waitForResources`, but otherwise, this looks good to me

e2e/cypress/integration/enterprise/cloud/billing/banner_trial_spec.js
```diff
@@ -0,0 +1,117 @@
...
+            // # Create a new user
+            cy.apiCreateUser({prefix: 'nonAdminUser'}).then(({user}) => {
+                // # Add user to the team
+                cy.apiAddUserToTeam(testTeam.id, user.id).then(() => {
+                    // # Login as new team admin
+                    cy.apiLogin(testUser);
+
+                    // # Go to town square
+                    cy.visit(`/${testTeam.name}/channels/town-square`);
+                    cy.waitForResources();
```
pablovelezvidal, line 99: same question here, is this some kind of way to wait for DOM elements to be/not be available?

e2e/cypress/integration/enterprise/cloud/billing/banner_trial_spec.js
vinod-demansol: Idea here is to wait for network idle and all the background http calls (scrpti,xhr,fetch) to get complete so that I can confidently run the other commands of cypress.

hmhealey:
Coming back to this now that the holidays are over, I'd prefer us to not use `waitForResources` if possible. If there's something specific that we need it for in this test, I'd be okay adding it, but I think we could get similar behaviour by just asserting that the banner is visible. That's also a bit more "realistic" of an E2E test since it's acting more like a user would instead of acting like an automated test by waiting for something that's mostly invisible to the user.

hmhealey:
Hi @vinod-demansol. Is there anything we can help with on this?

vinod-demansol:
> Hi @vinod-demansol. Is there anything we can help with on this?

I have removed the "waitForResources" method in my script , request for review

hmhealey (changes requested):
Thanks! Could you also remove the changes to `cloud_billing.js` and `could_billing.d.ts`? We can always re-add `waitForResources` if we find we need that later.

vinod-demansol:
> Thanks! Could you also remove the changes to `cloud_billing.js` and `could_billing.d.ts`? We can always re-add `waitForResources` if we find we need that later.

I have removed the waitForResource in the above mentioned files. Request to rereview it.

hmhealey (approved):
Thanks! Looks good to me now. We definitely appreciate the help on getting these tests automated

pablovelezvidal (approved):
LGTM, thanks @vinod-demansol 🚀

hmhealey:
@stevemudie This is ready for your review now

stevemudie (approved):
Looks good! Nice work @vinod-demansol!

PR #9540, opened 2021-12-14
remove black outline around copy test link on Microsoft browsers

neallred:
#### Summary
Microsoft Windows adds a thick black outline to the copy text button when `:focus`ed, and for some reason requires a very high specificity to override the outline.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40611

#### Screenshots
Before:
windoze-10

after:

https://user-images.githubusercontent.com/13738432/146081915-70915a41-43cd-406f-bbd3-c1a1a7cf4ed5.mp4

#### Release Note

```release-note
Fix Windows black outline.
```


michelengelen (approved):
LGTM ... pretty strange that it needs to be that specific though 😄

esethna:
Thanks @neallred. Looks like this button is also missing a hover and click effect. @annelieseklein please help provide guidance here based on the design system as applicable

neallred:
Ok, the feedback was to [use the tertiary style button](https://mattermost.atlassian.net/browse/MM-40611?focusedCommentId=120416). It has now been applied.

components/invitation_modal/invite_view.scss
```diff
@@ -10,13 +17,53 @@
     }
 
     &__copyLink {
+        &.tertiary-button {
```
pablovelezvidal, line 20: great button design implementation. I think this tertiary button is going to be widely used with the new designs, I wonder if we could put this in a mixin into `_mixins.scss` so it can be reused?

neallred: Thanks to the [playbooks](https://github.com/mattermost/mattermost-plugin-playbooks/blob/ef3ae7be3baeca825a8f16017cd8715e6b64cf86/webapp/src/components/assets/buttons.tsx#L77) people! Will do. I think the long term hope is to consume this through a compass component that encapsulates it, but having it as a mixin is better in the interim.

pablovelezvidal: thanks!

pablovelezvidal (approved):

stevemudie (approved):
Tested on Windows 11 using Edge browser and it worked as expected. Looks good @neallred!

stevemudie (approved):

PR #9542, opened 2021-12-15
MM-40630 Update mmjstools so i18n-extract supports TypeScript

hmhealey:
`make i18n-extract` used to have issues with TypeScript files. It turns out that that was because it was trying to parse all files as if they used Flow, not TypeScript, and it would silently fail as soon as it hit any TS code that wasn't also valid Flow.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40630

#### Related Pull Requests
https://github.com/mattermost/mattermost-utilities/pull/56

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/widgets/users/avatars/avatars.tsx
+++ b/components/widgets/users/avatars/avatars.tsx
@@ -21,10 +21,6 @@ import Avatar from 'components/widgets/users/avatar';
 import './avatars.scss';
 import {getMissingProfilesByIds} from 'mattermost-redux/actions/users';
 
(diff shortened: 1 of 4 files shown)


levb (commented):
LGTM for the apps string, no context on the other changes

hmhealey:
Yep, that's fine. Jesús already reviewed the actual changes to the tool in the linked PR

calebroseland (approved):
LGTM, though I do think it would be beneficial to migrate to https://formatjs.io/docs/tooling/cli at some point in the future.

PR #9543, opened 2021-12-15
MM-40633 Remove Storybook-related CI steps

hmhealey:
As far as I know, no one actually uses the web app's Storybook stuff since the widgets components aren't in active development any more, so this'll speed up CI a bit. We have another ticket to fully remove the Storybook code from the web app repo, but that'll be done separately since it'll require actual code changes.

This won't affect any of the Storybook in other repos or used by other teams, and we can definitely reintroduce it later on, but we don't need it in this repo for the time being.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40633

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,37 +159,6 @@ jobs:
           paths:
             - mattermost-webapp/mattermost-webapp.tar.gz
 
-  build-storybook:
-    executor:
-      name: default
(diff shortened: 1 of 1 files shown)


metanerd (approved):
Thank you for taking care @hmhealey !

M-ZubairAhmed (approved):

PR #9548, opened 2021-12-20
invalid-pr

M-ZubairAhmed:
#### Summary
Pulled out emoji search into separate component. Since search component was previously inside of render method of EmojiPicker, it rendered as many times as the EmojiPicker, in case of opening picker 15 times. Separating out only renders Search 1 time.
QA : Everything around emoji picker should work as it was before.

#### Ticket Link
Part of https://mattermost.atlassian.net/browse/MM-40296

#### Related Pull Requests
None

#### Screenshots
No UI changes

#### Release Note

```release-note
NONE
```


components/emoji_picker/emoji_picker.jsx
```diff
@@ -350,27 +347,27 @@ export default class EmojiPicker extends React.PureComponent {
...
             e.preventDefault();
             this.selectNextEmoji(EMOJI_PER_ROW);
-            this.searchInput.focus();
+            this.searchInputRef?.current?.focus();
             break;
         }
     }
 
-    handleKeyDown = (e) => {
+    handleSearchKeyDown = (e) => {
```
jespino, line 370: I think all this logic can be moved in the new component, probably we would need to remove the onKeyDown and add something like `onSelectEmoji`.

components/emoji_picker/emoji_picker.jsx
```diff
@@ -486,7 +482,7 @@ export default class EmojiPicker extends React.PureComponent {
         }
     }
 
-    onSkinSelected = (skin) => {
+    handleSkinChange = (skin) => {
         this.props.actions.setUserSkinTone(skin);
```
jespino, line 486: Probably this can be moved to the the other component using a useDispatch, but if not, we can at least pass directly the `this.props.actions.setUserSkinTone`.

components/emoji_picker/components/emoji_picker_search.tsx
```diff
@@ -0,0 +1,73 @@
...
+                userSkinTone={userSkinTone}
+                onSkinSelected={onSkinChange}
+            />
+        </div>
+    );
+});
+
+EmojiPickerSearch.displayName = 'EmojiPickerSearch';
+
+function propsAreEqual(prevProps: Props, nextProps: Props) {
```
jespino, line 66: Is there any reason to not use the default props comparison for memo here?

jespino (changes requested):
I love it. Some improvement suggestions. But it is looking great.

PR #9606, opened 2022-01-03
POC for a reusable, top-level menu component

nevyangelova:
#### Summary
This PR introduced a POC for a reusable, top-level, multi-device generic menu component. Per new design the menu on desktop and mobile should look significantly different, with their own transitions and styling in place, but rendering the same data. I have observed a group based styling in the docs here https://zeroheight.com/29be2c109/p/314add-popover-menu

I propose the following structure: 
- `MenuItem` - the smallest, reusable part of the component, that takes an `onClick` handler and performs an action. It has the ability to contain icons and other elements, as well as multiple labels and descriptions. Design - https://zeroheight.com/29be2c109/p/1092d5-menu-item
- `MenuPopover` - We utilise popper.js here to insure flexible placement as well as z-indexing above all other elements. The menu popover only handles the initial placement and is rendered by the `MenuData` wrapper. 
- `MenuData` - This is the container serving as a Menu or Submenu, and its main purpose is to serve as a panel, having the ability to slide up/down/left/right depending on the user device. Transition prototype here - https://www.figma.com/proto/zVNYaShAXDK8i1RdpuEkr8/%E2%9C%85Channel-Sidebar---All-Phases?node-id=3138%3A58308&scaling=min-zoom&page-id=1843%3A22750&starting-point-node-id=3139%3A928&show-proto-sidebar=1 
- `Menu` - This is the wrapper that handles the menu + submenu interaction and is the component we would call when we need to customise the menus. It takes menu groups and trigger as props. Menu groups should have the following structure: 

```
    const menuGroup: MenuGroup = [
        {
           title?: "Main group",
            menuItems: [
                <MenuItem
                   ...
                    leadingElement={<i className="icon-plus" />}
                    trailingElementLabel="selected"
                    trailingElement={<i className="icon-chevron-right" />}
                />,
                <MenuItem
                    destructive={true}
                   ...
                />,
                <MenuItem
                   ...
                />,
            ],
        },
    ];
// dividers are added after each separate group.
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33560
https://mattermost.atlassian.net/browse/MM-30324
https://mattermost.atlassian.net/browse/MM-30343

#### Screenshots
Screenshot 2022-01-12 at 16 15 19
Screenshot 2022-01-12 at 16 15 26
Screenshot 2022-01-12 at 16 15 38
Screenshot 2022-01-12 at 16 15 43

#### Release Note
-->
```release-note
NONE
```


devinbinnie:
Thanks for the PR @tweeak! Can we get you to sign the CLA here for us before we review your changes?

http://www.mattermost.org/mattermost-contributor-agreement/

components/global_header/hooks.tsx
```diff
@@ -23,15 +26,35 @@ export const useProducts = (): ProductComponent[] | undefined => {
...
             }
+
+            if (!Array.isArray(refOrRefs)) {
+                if (!refOrRefs.current?.contains(target)) {
+                    handler();
+                }
+                return;
+            }
+
+            const someRefContainTarget = refOrRefs.filter((ref) => ref.current).some((ref) => ref.current?.contains(target));
```
michelengelen, line 51: current should always be present after you filtered out those that are not

```suggestion
            const someRefContainTarget = refOrRefs.filter((ref) => ref.current).some((ref) => ref.current.contains(target));
```

components/menu/menu/Menu.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+        &.fade-exit-active {
+            background-color: rgba(0, 0, 0, 0);
+        }
+
+        &.fade-exit-done {
+            background-color: rgba(0, 0, 0, 0);
+        }
+    `;
+});
```
michelengelen, line 64: You can use the "easy" syntax here, since you are not using any props inside:

```suggestion
const Overlay = styled.div`
    background-color: rgba(0, 0, 0, 0);
    display: flex;
    align-items: flex-end;
    justify-content: center;
    height: 100%;
    min-height: 100%;
    left: 0;
    right: 0;
    top: 0;
    position: fixed;
    overflow: auto;
    -webkit-overflow-scrolling: touch;
    overscroll-behavior: contain;
    pointer-events: auto;
    -ms-scroll-chaining: none;
    transition: 1s;
    transition-property: background-color;
    transition-timing-function: ease-in-out;

    &.fade-enter {
        background-color: rgba(0, 0, 0, 0);
    }

    &.fade-enter-active {
        background-color: rgba(0, 0, 0, 0.5);
    }

    &.fade-enter-done {
        background-color: rgba(0, 0, 0, 0.5);
    }

    &.fade-exit {
        background-color: rgba(0, 0, 0, 0.5);
    }

    &.fade-exit-active {
        background-color: rgba(0, 0, 0, 0);
    }

    &.fade-exit-done {
        background-color: rgba(0, 0, 0, 0);
    }
`;
```

components/menu/menu/Menu.tsx
```diff
@@ -0,0 +1,225 @@
...
+    // and can hold any value, similar to an instance property on a class
+    const ref = useRef<T>();
+
+    // Store current value in ref
+    useEffect(() => {
+        ref.current = value;
+    }, [value]); // Only re-run if value changes
+    // Return previous value (happens before update in useEffect above)
+
+    return ref.current;
```
michelengelen, line 99:
```suggestion

    // Return previous value (happens before update in useEffect above)
    return ref.current;
```

components/menu/menu/Menu.tsx
```diff
@@ -0,0 +1,225 @@
...
+}
+
+interface MenuProps {}
+
+
```
michelengelen, line 104: nit: please remove that extra empty line

components/menu/menu/Menu.tsx
```diff
@@ -0,0 +1,225 @@
...
+    }, []);
+
+    useClickOutsideRef([buttonReference, menuRef, submenuRef], () => {
+        if (!open || isMobile) {
+            return;
+        }
+        closeMenu();
+    });
+
+    const closeSubmenuDown = Boolean(prevOpen && prevSubmenuOpen && !submenuOpen && !open);
```
michelengelen, line 129: there is no need for `Boolean`

components/menu/menu/Menu.types.tsx
```diff
@@ -0,0 +1,55 @@
...
+    destructive?: boolean;
+    disabled?: boolean;
+    leadingElement?: React.ReactNode;
+    trailingElementLabel?: string;
+    trailingElement?: React.ReactNode;
+    onClick?: () => void;
+    onHover?: () => void;
+}
+
+export type MenuGroup = {
```
michelengelen, line 18: why are you mixing `interface` and `type` ... just curious

components/menu/menu/MenuData.tsx
```diff
@@ -0,0 +1,171 @@
...
+import React, {ForwardedRef} from 'react';
+import styled, {css} from 'styled-components';
+import classNames from 'classnames';
+
+import MenuPopover from './MenuPopover';
+import {MenuDataProps} from './Menu.types';
+
+const Divider = () => (
+    <div style={{height: '1px', width: 'auto', backgroundColor: '#e0e0e0'}}/>
+);
```
michelengelen, line 13:
```suggestion
const Divider = () => <div style={{height: '1px', width: 'auto', backgroundColor: '#e0e0e0'}}/>;
```

components/menu/menu/MenuData.tsx
```diff
@@ -0,0 +1,171 @@
...
+                    >
+                        {isSubmenu && isMobile && (
+                            <i
+                                className='icon icon-arrow-back-ios'
+                                style={{
+                                    margin: '0 auto 0 23px',
+                                    color: '#3D3C40',
+                                }}
+                                onClick={closeSubmenu}
+                            />
```
michelengelen, line 154: you could use `styled` to add the styles here as well

components/menu/menu/MenuItem.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+import {MenuItemProps} from './Menu.types';
+
+const MenuItemLabel = styled.span(
+    () => css`
+        font-size: 14px;
+        line-height: 20px;
+        margin: 0 8px 0 0;
+    `,
+);
```
michelengelen, line 16:
```suggestion
const MenuItemLabel = styled.span`
    font-size: 14px;
    line-height: 20px;
    margin: 0 8px 0 0;
`;
```

components/menu/menu/MenuItem.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+const MenuItemDescription = styled.span(
+    () => css`
+        font-weight: 400;
+        font-size: 12px;
+        line-height: 16px;
+        color: rgba(var(--center-channel-color-rgb), 0.56);
+    `,
+);
```
michelengelen, line 25:
```suggestion
const MenuItemDescription = styled.span`
    font-weight: 400;
    font-size: 12px;
    line-height: 16px;
    color: rgba(var(--center-channel-color-rgb), 0.56);
`;

```

components/menu/menu/MenuItem.tsx
```diff
@@ -0,0 +1,137 @@
...
+                        justifyItems: 'flex-start',
+                    }}
+                >
+                    {trailingElement && trailingElementLabel && (
+                        <MenuItemDescription>
+                            {trailingElementLabel}
+                        </MenuItemDescription>
+                    )}
+                    {trailingElement}
+                </div>
```
michelengelen, line 132: use `styled` for this as well to be consistent

michelengelen (changes requested):
looking good so far ... I only have some minor suggestions

components/menu/menu/MenuItem.tsx
nevyangelova: yes :D some leftover code I am sorry

stylianosrigas:
Removing cloud test server for maintenance reason. Please feel free to add it again if needed.

nevyangelova:
Ready for first QA/UX spin as well as i fix the test and my favourite type errors:)

components/menu/MenuPopover.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const mobileStyle = isMobile && {
+        position: 'fixed',
+        bottom: 0,
+        left: 0,
+        right: 0,
+        transition: 'visibility 300ms 0ms step-end',
+    };
+
+    const style = {
+        zIndex: isVisible ? 20 : -1,
```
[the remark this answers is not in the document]
nevyangelova: Unfortunately it is... due to scrollbars and other elements fighting for z-index and other issues. We really need to focus on the layout overhaul, if even portals aren't top level :D 

On another note I am pretty sure that we will have to increase that z-index for other menus.

components/menu/MenuData.tsx
```diff
@@ -0,0 +1,169 @@
...
+                                        margin: '0 auto 0 23px',
+                                        color: '#3D3C40',
+                                    }}
+                                    onClick={closeSubmenu}
+                                />
+                            )}
+                            {title && <MenuTitle>{title}</MenuTitle>}
+                        </MenuHeader>
+                        {groups.map((group, i) => (
+                            <>
```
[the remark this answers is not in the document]
nevyangelova: Linter throws an error (there are quite a few of those in master) when each child does not have a unique key prop.

components/menu/Menu.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+        &.fade-exit-done {
+            background-color: rgba(0, 0, 0, 0);
+        }
+    `;
+});
+
+const useIsMobile = (): boolean => {
+    const [isMobile, setIsMobile] = useState<boolean>(false);
+    const mobileMediaQuery = window.matchMedia('(max-width: 699px)');
```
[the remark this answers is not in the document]
nevyangelova: That would work when shrinking the screen, otherwise you need a re-render

components/menu/Menu.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const prevSubmenuOpen = usePrevious(submenuOpen);
+    const isMobile = useIsMobile();
+    const isOverlayVisible = isMobile && open;
+
+    useClickOutsideRef([trigger.ref, menuRef, submenuRef], () => {
+        if (!open || isMobile) {
+            return;
+        }
+        if (overlayCloseHandler !== undefined) {
+            overlayCloseHandler();
```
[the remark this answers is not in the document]
nevyangelova: Good catch!

components/menu/Menu.tsx
```diff
@@ -0,0 +1,172 @@
...
+    } = props;
+
+    const menuRef = useRef(null);
+    const submenuRef = useRef(null);
+    const prevOpen = usePrevious(open);
+    const prevSubmenuOpen = usePrevious(submenuOpen);
+    const isMobile = useIsMobile();
+    const isOverlayVisible = isMobile && open;
+
+    useClickOutsideRef([trigger.ref, menuRef, submenuRef], () => {
```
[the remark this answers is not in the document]
nevyangelova: I know we talked about this but I wanted to ask a UX person just to make sure we are aligned, @michaelgamble do we want to prevent interaction with the app background when the menu is open?

components/menu/MenuData.tsx
```diff
@@ -0,0 +1,169 @@
...
+// See LICENSE.txt for license information.
+
+import React, {ForwardedRef} from 'react';
+import styled, {css} from 'styled-components';
+import classNames from 'classnames';
+
+import {MenuDataProps} from './Menu.types';
+import MenuPopover from './MenuPopover';
+
+const Divider = () => (
```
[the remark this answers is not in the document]
nevyangelova: IMO its small enough and doesn't need prop based styles but i am 2/5 on this so i can update

components/menu/MenuData.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+const MenuItems = styled.div<{isMobile?: boolean; isSubmenu: boolean}>(
+    ({isMobile, isSubmenu}) => {
+        return isMobile ?
+            css`
+                  background-color: rgba(var(--sidebar-text-rgb));
+                  min-height: 340px;
+                  height: auto;
+                  border-radius: 8px 8px 0 0;
+                  max-height: ${document.documentElement.clientHeight}px;
```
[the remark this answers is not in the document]
nevyangelova: oh i think i was calculating this based on removing some pixel values too, clientHeight - 20px for example. I can update but wanted to make sure UX is happy with this as well.

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.test.tsx
```diff
@@ -2,238 +2,42 @@
...
+        const state = {
         };
+        const store = mockStore(state);
+        const dummyDispatch = jest.fn();
+        useDispatchMock.mockReturnValue(dummyDispatch);
 
-        const wrapper = shallowWithIntl(
-            <SidebarChannelMenu {...props}/>,
+        const wrapper = shallow(
+            <reactRedux.Provider store={store}><SidebarChannelMenu {...baseProps}/></reactRedux.Provider>,
```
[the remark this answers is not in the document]
nevyangelova: This is not the final test it was just a quick fix so that nothing fails, i have the ticket to write proper tests here https://mattermost.atlassian.net/browse/MM-44901

Didn't want this to be too cluttered :P

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -1,284 +1,502 @@
...
+    const isDM =
+        channel.type === Constants.DM_CHANNEL ||
+        channel.type === Constants.GM_CHANNEL;
+    const firstMenuGroup = {
+        menuItems: [
+            ...(isUnread ?
+                [
+                    <MenuItem
+                        id={`markAsRead-${channel.id}`}
+                        key={`markAsRead-${channel.id}`}
```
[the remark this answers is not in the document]
nevyangelova: I am actually not sure if we need to keep the ID's i suppose its for E2e tests but when I write new ones these can be obsolete.

package.json
```diff
@@ -52,7 +52,7 @@
     "popper.js": "1.16.1",
     "process": "0.11.10",
     "prop-types": "15.7.2",
-    "react": "16.13.1",
+    "react": "16.13.1 || ^17.0.1",
```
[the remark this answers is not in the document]
nevyangelova: This was for adding the 'ForwardedRef' type xD i can remove it

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -1,284 +1,502 @@
...
+                })}
+                trigger={menuTrigger}
+                submenuTrigger={subMenuTrigger}
+                groups={[firstMenuGroup, submenuItem, secondMenuGroup]}
+                submenuGroups={[submenuGroup]}
+                open={isMenuVisible}
+                submenuOpen={isSubMenuVisible}
+                overlayCloseHandler={closeAllMenus}
+                closeSubmenu={toggleSubmenu}
+                submenuTitle={intl.formatMessage({
```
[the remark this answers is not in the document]
nevyangelova: That's a good point, we should just have two separate components for the menu and submenu.

components/menu/Menu.tsx
michelengelen, line 125: this can be handled by "non-null assertion" `overlayCloseHandler?.();`

components/menu/Menu.tsx
```diff
@@ -0,0 +1,172 @@
...
+        if (!open || isMobile) {
```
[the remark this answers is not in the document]
michelengelen: this is only the outside click handler and does not have a relation to the rendering

components/menu/MenuData.tsx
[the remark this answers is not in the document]
michelengelen: IMO it should be

components/menu/MenuData.tsx
```diff
@@ -0,0 +1,169 @@
...
+import {MenuDataProps} from './Menu.types';
+import MenuPopover from './MenuPopover';
+
+const Divider = () => (
+    <div style={{height: '1px', width: 'auto', backgroundColor: '#e0e0e0'}}/>
+);
+
+const MenuItems = styled.div<{isMobile?: boolean; isSubmenu: boolean}>(
+    ({isMobile, isSubmenu}) => {
+        return isMobile ?
```
michelengelen, line 17: could we consolidate shared styles and only append some for the different use-cases?

components/menu/MenuData.tsx
```diff
@@ -0,0 +1,169 @@
...
+                        >
+                            {isSubmenu && isMobile && (
+                                <i
+                                    className='icon icon-arrow-back-ios'
+                                    style={{
+                                        margin: '0 auto 0 23px',
+                                        color: '#3D3C40',
+                                    }}
+                                    onClick={closeSubmenu}
+                                />
```
michelengelen, line 151: could we use the `IconArrowIcon` component from `@mattermost/compass-icons/components` here please?

components/menu/MenuData.tsx
```diff
@@ -0,0 +1,169 @@
...
+                        isSubmenu={Boolean(isSubmenu)}
+                        className={classNames({
+                            open,
+                            active,
+                            closeSubmenuDown,
+                        })}
+                    >
+                        <MenuHeader
+                            className={classNames({
+                                hasArrow: isSubmenu && isMobile,
```
michelengelen, line 139: what is this class being used for? If we are leveraging `flex` or `grid` this should not be necessary

components/menu/MenuItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+    font-size: 12px;
+    line-height: 16px;
+    color: rgba(var(--center-channel-color-rgb), 0.56);
+`;
+
+const LabelWrapper = styled.div<{leadingElement: boolean}>(
+    ({leadingElement}) => css`
+        display: flex;
+        flex-direction: column;
+        padding-left: ${leadingElement ? '10px' : 0};
```
michelengelen, line 27: this feels overly complicated. This could be made simpler by using `flex` and the `gap` property on the parent component or by using `grid`

components/menu/MenuItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+                {...rootProperties}
+            >
+                {leadingElement}
+                <LabelWrapper leadingElement={Boolean(leadingElement)}>
+                    <MenuItemLabel>{label}</MenuItemLabel>
+                    {description && (
+                        <MenuItemDescription>{description}</MenuItemDescription>
+                    )}
+                </LabelWrapper>
+                <div
```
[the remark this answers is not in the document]
michelengelen: there actually is no need for that wrapping `div` when we would use `grid`

components/menu/MenuPopover.tsx
```diff
@@ -0,0 +1,86 @@
...
+    children,
+}: MenuPopoverProps): JSX.Element | null => {
+    const [popperElement, setPopperElement] =
+        React.useState<HTMLDivElement | null>(null);
+
+    const {
+        styles: {popper},
+        attributes,
+        update,
+    } = usePopper(triggerRef.current, popperElement, {
```
michelengelen, line 46: IMO we should be using floatingUI instead for 2 reasons:

1. popper is a discontinued package
2. floatingUI is a bit more comfortable to use

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -1,284 +1,502 @@
...
-        this.props.actions.markChannelAsRead(this.props.channel.id);
+    const markAsRead = () => {
+        dispatch(markChannelAsRead(channel.id));
         trackEvent('ui', 'ui_sidebar_channel_menu_markAsRead');
-    }
+    };
 
-    favoriteChannel = () => {
-        this.props.actions.favoriteChannel(this.props.channel.id);
+    const favorite = () => {
```
michelengelen, line 132: `favorite` and `unfavorite` could be consolidated into a single `toggleFavorite` handler

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -1,284 +1,502 @@
...
+    const unfavorite = () => {
+        dispatch(unfavoriteChannel(channel.id));
         trackEvent('ui', 'ui_sidebar_channel_menu_unfavorite');
-    }
+    };
 
-    unmuteChannel = () => {
-        this.props.actions.unmuteChannel(this.props.currentUserId, this.props.channel.id);
-    }
+    const unmute = () => {
```
michelengelen, line 142: `mute` and `unmute` could also be consolidated into a `toggleMute` handler

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -1,284 +1,502 @@
...
+        })}
+        leadingElement={
+            location === 'sidebar' ? (
+                <i className='icon-folder-move-outline'/>
+            ) : null
         }
+        onMouseEnter={toggleSubmenuDesktop}
+        onClick={toggleSubmenuMobile}
+        trailingElementLabel='selected'
+        trailingElement={<i className='icon-chevron-right'/>}
```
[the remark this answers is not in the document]
michelengelen: also passing it in like this will cause a lot of unnecessary re-renders

and we can use the icon-component from `@mattermost/compass-icons/components` which makes aligning a lot easier

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -1,284 +1,502 @@
...
+                [
+                    <MenuItem
+                        id={`unfavorite-${channel.id}`}
+                        key={`unfavorite-${channel.id}`}
+                        label={intl.formatMessage({
+                            id: 'sidebar_left.sidebar_channel_menu.unfavoriteChannel',
+                            defaultMessage: 'Unfavorite',
+                        })}
+                        leadingElement={<i className='icon-star'/>}
+                        onClick={unfavorite}
```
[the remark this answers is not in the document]
michelengelen: lol ... i had the same idea and wrote a comment about it on the handlers! :D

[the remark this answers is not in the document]
michelengelen: I would even go on and separate the whole component into a `FavoriteMenuItem` and define it  before passing it in. If we would do that for all the items in this array we could also just pass in null when the item should not exist and reduce the amount of complexity in this array definition

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -1,284 +1,502 @@
...
+                        key={`favorite-${channel.id}`}
+                        label={intl.formatMessage({
+                            id: 'sidebar_left.sidebar_channel_menu.favoriteChannel',
+                            defaultMessage: 'Favorite',
+                        })}
+                        onClick={favorite}
+                        leadingElement={<i className='icon-star-outline'/>}
+                    />,
+                ]),
+            ...(isMuted ?
```
michelengelen, line 308: same comment as above on the favorite `MenuItem`

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
```diff
@@ -1,284 +1,502 @@
...
+                    key={category.id}
+                    onClick={() => moveToCategory(category.id)}
+                    label={category.display_name}
+                    leadingElement={
+                        category.type === CategoryTypes.FAVORITES ? (
+                            <i className='icon-star-outline'/>
+                        ) : (
+                            <i className='icon-folder-outline'/>
+                        )
+                    }
```
michelengelen, line 449: define this before the return in a `const` and pass it into the prop. That way we can reduce re-renders

components/menu/MenuData.tsx
nevyangelova: its adding an arrow on mobile view only so it is

components/menu/MenuItem.tsx
nevyangelova: I am not sure how its overly complicated, please elaborate?

components/menu/MenuPopover.tsx
nevyangelova: this would go against the idea of only introducing new packages when absolutely necessary. Popperjs is widely used in mattermost webapp __

components/menu/Menu.tsx
```diff
@@ -0,0 +1,172 @@
...
+    useEffect(() => {
+        mobileMediaQuery.addEventListener('change', handleChange);
+        return () => {
+            mobileMediaQuery.removeEventListener('change', handleChange);
+        };
+    }, [handleChange]);
+
+    useEffect(() => {
+        setIsMobile(mobileMediaQuery.matches);
+    }, []);
```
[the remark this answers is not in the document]
nevyangelova: its also used when shrinking screen size so its necessary atm

components/menu/MenuData.tsx
nevyangelova: I am not sure about using compass since we don't know the future of the framework? Perhaps @zefhemel can advice?

components/menu/Menu.tsx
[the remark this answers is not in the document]
michelengelen: but this useEffect call would only run on render, so why is that firing on resize?

components/menu/MenuData.tsx
michelengelen: i am not suggesting using compass-components here, but the compass-icons package

nevyangelova: oops my bad

components/menu/MenuItem.tsx
michelengelen: of course

If you are using a `display: flex` on the parent you can set the a `gap` (shorthand for `column-gap` and `row-gap`) as well which then renders a whitespace between two elements (columns or rows).

So when you are defining the property on the parent component each item will be separated by that gap.

You could even go one step further to define `display: grid` and a layout on the parent, which essentially reduces the manual addition of styling and wrapping elements to make that layout possible. This helps in reducing the clutter on the DOM as well, since we don't need the elements (and additional styling on them) to align the contents of it inside that layout.

Here are some useful links:
- [gap explanation](https://css-tricks.com/almanac/properties/g/gap/)
- [grid cheatsheet](https://css-tricks.com/snippets/css/complete-guide-grid/)

If you want (and have time for it) we can go over this in more detail

components/menu/MenuPopover.tsx
michelengelen: I implemented floatingUI into another PR already #10529 ... so we can safely use it here as well. It is just a suggestion though and if we decide to pick that up we definitely need to replace all usages pf popper with floatingUI.

IMO the floatingUI package is a lot better in terms of usability, adjustability and style.

components/menu/MenuData.tsx
michelengelen: so ... kudos for using `grid`, but you actually do not need to use a separate class for that. Just add this to the parent and everything works as before:

```
grid-template-columns: repeat(3, auto);
justify-content: space-between;
```

This is the automated version ... for a more sophisticated version where each item gets their position defined manually (this is more precise) this would look a little different.

components/menu/Menu.tsx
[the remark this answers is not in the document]
hmhealey: I meant more that, if this handler is registered even when the menu isn't open, then I'm worried that means we're going to be doing a lot of processing and rendering for stuff in the menu that isn't visible most of the time.

Going back to this handler though, this is actually going to be a performance regression as-written. We had something like this with the old menu where it registered a click outside handler to close the menu, and just having a few dozen to a few hundred handlers that were called any time you clicked anywhere in the app was causing noticeable lag even if none of them actually did anything with that click.

While I still think an invisible div would have other benefits, if we stick with this sort of click outside handler, we should be sure to only have that registered when the menu is actually visible.

components/menu/Menu.tsx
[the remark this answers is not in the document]
hmhealey: I'd say I'm 3/5 on preventing background interaction. Apart from accidentally interacting with background stuff (which is really bad with click-to-reply turned on), keeping everything looking right when it comes to scrolling is a huge pain. It's nice to see this actually works pretty well in this current version, but when I was playing around with it, it breaks for submenus
Screen Shot 2022-06-13 at 1 23 18 PM

And even though you can make scrolling work nicely, I'd say I'm even 5/5 about that case that it's not worth the effort. There's a reason why a lot of apps either prevent scrolling with a menu open (Slack, Firefox, Chrome, Notes) or just don't bother making it so that menus move when you scroll c (Discord, Steam).

components/menu/MenuData.tsx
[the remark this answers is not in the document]
hmhealey: I'm pretty sure you can actually do `calc(100vh - 30px)` with mixed units too

components/menu/MenuPopover.tsx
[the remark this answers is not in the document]
hmhealey: Interesting. Are other elements like scroll bars mounted at the top level too? I would've thought most of those would be contained in their parents

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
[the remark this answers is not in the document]
hmhealey: Yeah, that might be nice as well. It'd certainly reduce the size of this function too

components/sidebar/sidebar_channel/sidebar_channel_menu/sidebar_channel_menu.tsx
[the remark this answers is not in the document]
hmhealey: Yeah, if we could do that, that'd be great. If we moved the processing for the submenu to its own component, then as long as that component isn't mounted when it isn't visible, it'll save us from doing any of that processing automatically

components/menu/MenuPopover.tsx
hmhealey: Let's stick with popper here for now. We need to figure out where we're going with these since, even if floating-ui is better, we need to stick with a single dependency for some of these things

nevyangelova:
Most of the complexities are stemming from the mobile view animations. In order for the submenu to push the parent to the left on mobile we need to keep track of both menu's open and active states. After I had a conversation with @hmhealey and @zefhemel it became clear that the mobile web view will be eventually deprecated. It doesn't make sense to keep the complexity of the component for the mobile view. So until the view is deprecated, we will have a simplified version of the mobile animations, the submenu will drag on top of the main, without pushing it aside. 

I will close the PR and reopen with the simplified structure shortly. Everyone who would like to review, i will post the link here shortly.

nevyangelova:
new PR https://github.com/mattermost/mattermost-webapp/pull/10614

PR #9318, opened 2021-11-02
MM-39804 - Remove license expiration checks for cloud

marianunez:
#### Related Pull Requests

- Has server changes (please link here)
- Has mobile changes (please link here)

#### Release Note

```release-note

```

PR #9607, opened 2022-01-03
MM-40790 - Do not show license expired banner for Cloud

marianunez:
#### Summary

The license check and invalidation was removed server side for Cloud installations in https://github.com/mattermost/mattermost-server/pull/18921 and https://github.com/mattermost/mattermost-webapp/pull/9318 . However, it was not removed in the banner checks for already expired licenses. This PR will not show the license expired banner for any cloud installation. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40790

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/announcement_bar/configuration_bar/__snapshots__/configuration_bar.test.tsx.snap
+++ b/components/announcement_bar/configuration_bar/__snapshots__/configuration_bar.test.tsx.snap
@@ -40,6 +40,8 @@ exports[`components/ConfigurationBar should match snapshot, expired 1`] = `
 />
 `;
 
+exports[`components/ConfigurationBar should match snapshot, expired, cloud license, show nothing 1`] = `""`;
+
 exports[`components/ConfigurationBar should match snapshot, expired, in grace period 1`] = `
 <Connect(AnnouncementBar)
   message={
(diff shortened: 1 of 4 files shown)


AGMETEOR (commented):
Looks good to me. Maybe unit tests for these util functions?

marianunez:
@stevemudie ping me when ready for licenses to be about to expire (see ticket QA steps)

marianunez:
> Looks good to me. Maybe unit tests for these util functions?

Thanks @AGMETEOR! I've added the unit tests to cover these new cases. 👍

AGMETEOR (approved):

stevemudie:
@marianunez The Cloud license did not show an expired banner as expected so that's all set.

However, the non-cloud license never showed an expired banner - even after refresh - and removing it after 11:30 then re-uploading it failed as it was expired after 11:30 by design.

marianunez:
> However, the non-cloud license never showed an expired banner - even after refresh - and removing it after 11:30 then re-uploading it failed as it was expired after 11:30 by design.

Just looked into this and the expired banner will show the day after it is expired - it does not check the time. Let's differ this test for post-merge so we can do a hot fix for cloud

stevemudie (approved):
Halfway there as discussed above.

Deferring to test post merge to get the fix out the door.

PR #9610, opened 2022-01-04
POC: pa11y integration

michelengelen:
#### Summary
adds pa11y to the project, which lets us test the a11y conformity in CI

#### Release Note
```release-note
NONE
```


Willyfrog (commented):
Looks good, but once we start using this, how do we plan to have all the tests organized? I understand this is a simple POC, but having a giant array with actions doesn't look very good. Any plans on how to structure that?
Do we know how good is the compatibility of this with JAWS and NVDA?

michelengelen:
> Looks good, but once we start using this, how do we plan to have all the tests organized? I understand this is a simple POC, but having a giant array with actions doesn't look very good. Any plans on how to structure that? Do we know how good is the compatibility of this with JAWS and NVDA?

Thanks for checking this out... well to answer your questions:

1. `pa11y` runs by scanning URLs (and having a specific set of actions it can perform), so IMO the structure should conform to that, meaning we could set up a dedicated folder that has all of the test-files in it structured by paths in the webapp. We can also nest the structure a bit (but should keep it as flat as possible) to test specific behaviours like modals, prompts, etc.

2. i plan to add a set of helper functions to add actions to the tests. Something of the likes: `logIn()`, `logInAsAdmin()`, `navigateToTeam()`, `postMessage()`, `postReply()`, `openSearch()`, etc. What they do is returning an array of predefined actions. It could look something like this:

``` typescript
{
    // other config items
    actions: [
        ...loginAsAdmin(),
        ...navigateToSystemConsole(),
        ...openSettings(SETTINGS.Compliance),
    ],
}
```

3.`pa11y` can be configured with different runners, where the default runner is `htmlcs` ([HTML Code Sniffer](https://squizlabs.github.io/HTML_CodeSniffer/)). It has 2 runners pre-installed: `htmlcs` and `axe` ([aXe-Core](https://www.axe-core.org/)), but supports more if they are installed.

I am currently learning about `pa11y` ans the possibilities as well, so there might be more insights I do not know about yet later on. Maybe I can find some best practices as well and see if that fits our needs.

hmhealey:
This sounds cool to me. It'd be nice if the actions were a similar syntax to Cypress to keep the tests that require navigating a real UI looking the same, but these seem easy enough to follow

michelengelen:
@hmhealey there is a separate package that includes `pa11y` and `Lighthouse` into `cypress` called (`cypress-audit`)[https://github.com/mfrachet/cypress-audit], but we (rather @metanerd did) talked to @streamer45 and came to the conclusion that the cypress tests as they are are already bloated enough so adding another step to them would be making them too complex and I totally agree with that.

For now, as mentioned above, I plan to add a series of helper functions to add certain sets of actions for recurring actions in a helper so we don't have to repeatedly write them.

michelengelen:
I found a nice summary of testing tools: https://alphagov.github.io/accessibility-tool-audit/

So the outcome is that every automated a11y testing should be accompanied by a good manual testing step from QA, since the best testing library could only find about 40% of the issues. 

As a learning from that I will change the runner from `htmlcs` to `aXe`.

metanerd:
Thank you for looking into it @michelengelen  and going the extra mile! I actually talked to @saturninoabril instead of @streamer45 about accessibility testing via cypress, but if you want @streamer45 , we can have a chat about it :). Saturn said, he will comment later, if it makes sense to use the cypress routing also for a11y setup after he had more consideration time on how to integrate it in a more isolated fashion.

michelengelen:
Ooops ... @metanerd I somehow mixed them up ... sry about that @streamer45 ... I don't think we need to bother you with this considering you might have your hands full with getting calls prod ready! :D

PR #9615, opened 2022-01-04
MM-40751 Properly set Content-Type on API requests

hmhealey:
Turns out all API requests we make to the server from the client are made with `Content-Type: plain/text` even though we're always passing JSON through `Client4`. I looked through every Client4 call that passes a body, and every one of them calls `JSON.stringify` on an object before passing it as the `body` option

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40751
Closes https://mattermost.atlassian.net/browse/MM-40751

#### Release Note
```release-note
Changed Client4 to properly set Content-Type as application/json on API calls
```

### DIFF
--- a/packages/mattermost-redux/src/client/client4.ts
+++ b/packages/mattermost-redux/src/client/client4.ts
@@ -123,6 +123,7 @@ import {TelemetryHandler} from './telemetry';
 const FormData = require('form-data');
 const HEADER_AUTH = 'Authorization';
 const HEADER_BEARER = 'BEARER';
+const HEADER_CONTENT_TYPE = 'Content-Type';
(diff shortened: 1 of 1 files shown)


packages/mattermost-redux/src/client/client4.ts
```diff
@@ -467,6 +469,10 @@ export default class Client4 {
             headers[HEADER_USER_AGENT] = this.userAgent;
         }
 
+        if (options.body) {
+            headers['Content-Type'] = 'application/json';
```
catalintomai, line 473: should we add `Content-Type` as a header constant similar to the other headers (e.g. `HEADER_USER_AGENT`) ?

catalintomai (approved):

agnivade (approved):

PR #9617, opened 2022-01-05
Add workaround for mmjstool integrity check failing

hmhealey:
Copies a workaround for the integrity check issue that we've had people reporting since we upgraded mmjstool

See https://github.com/mattermost/mattermost-plugin-playbooks/pull/868 for more details

#### Release Note
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,8 @@ i18n-extract: ## Extract strings for translation from the source code
 node_modules: package.json package-lock.json
 	@echo Getting dependencies using npm
 
+	node skip_integrity_check.js
+
 	npm install
 	touch $@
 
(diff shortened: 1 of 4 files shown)


lieut-data:
Thanks, @hmhealey! I ended up patching the first PR in https://github.com/mattermost/mattermost-plugin-playbooks/pull/940/files as we found the preinstall/postinstall hooks weren't reliable enough for use in this context. Would it make sense to include the changes from that PR as well?

hmhealey:
Yep, I can add those changes to this PR. I forgot how flaky those hooks can be since they used to cause so many problems with mattermost-redux not building before

lieut-data (approved):

M-ZubairAhmed (approved):

PR #9638, opened 2022-01-11
[MM-40591]Admin console team override scheme page going into an infinite loading state

AshishDhama:
#### Summary
Admin console team override scheme page going into an infinite loading state

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-40591

#### Related Pull Requests
``None``

#### Screenshots
|  before  |  after  |
|----|----|
| https://www.loom.com/share/691fe8bf460143cea5de25d94290bda6 | https://www.loom.com/share/73f53038beb24ea59558064a73f2f496 |

#### Release Note

```release-note
* Bugfix: Admin console team override scheme page going into an infinite loading state
```


AshishDhama:
@crspeller Let me know if I need to change the role condition instead of loading ``run_admin`` to ``run_member``

marianunez (approved):

crspeller (changes requested):
Thanks for looking at this!
Loading the run admin role is actually unnecessary. The error I think is on line 164 where we check for the existence of run_admin instead of checking for run_member.

crspeller (approved):

saturninoabril (approved):
Thanks @AshishDhama! Tested and passed.

PR #9654, opened 2022-01-15
MM-31051 - unify cc expired sentence

pablovelezvidal:
#### Summary
This PR unifies several sentences in a single sentence so it contains all the context and it is easier to translate to other idioms.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31051

#### Release Note
```release-note
NONE
```


components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
```diff
@@ -290,24 +279,13 @@ export const paymentFailedBanner = () => {
...
-                        />
-                    </BlockableLink>
-                    <FormattedMessage
-                        id='billing.subscription.info.mostRecentPaymentFailed.description.avoidAnyDisruption'
-                        defaultMessage=' to avoid any disruption.'
-                    />
-                </>
+                <FormattedMessage
+                    id='billing.subscription.info.mostRecentPaymentFailed.description.mostRecentPaymentFailed'
+                    defaultMessage='It looks your most recent payment failed because the credit card on your account has expired. Please  <link>update your payment information</link> to avoid any disruption.'
```
neallred, line 284: Neat! Didn't know you could specify tags to wrap content with arbitrary JSX elements like this.

neallred (approved):

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
pablovelezvidal: me neither, special thanks to @hmhealey for that advice: @neallred this is the reference link https://formatjs.io/docs/react-intl/components/#rich-text-formatting

stevemudie (approved):
I used an expired test card in a variety of languages and they all formatted properly and looked good.

Nice work @pablovelezvidal!

hmhealey (approved):

components/admin_console/billing/billing_subscriptions/billing_subscriptions.tsx
hmhealey: Yeah, it's a great feature. It's sort of like the older way we did things where we'd inline HTML, but it's easier to use since it lets us use React components and hide the HTML while also being safer since it prevents potentially malicious translators from breaking anything.

PR #9657, opened 2022-01-17
MM-40181: fixes adding GM/DM threads to all teams

koox00:
#### Summary

Threads in a GM/DM belong in all teams. 
Currently, we add a DM/GM thread only to the current team (when receiving the websocket event).

This commit fixes it by adding those threads to all teams currently in store.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40181

#### Release Note

```release-note
NONE
```

### DIFF
--- a/packages/mattermost-redux/src/reducers/entities/threads/threads.test.js
+++ b/packages/mattermost-redux/src/reducers/entities/threads/threads.test.js
@@ -488,9 +488,13 @@ describe('threads', () => {
         const state = deepFreeze({
             threadsInTeam: {
                 a: ['t1', 't2'],
+                b: ['t1'],
+                c: [],
             },
             unreadThreadsInTeam: {
                 a: ['t2'],
+                b: [],
+                c: [],
             },
             threads: {
                 t1: {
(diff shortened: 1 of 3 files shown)


packages/mattermost-redux/src/reducers/entities/threads/threads.test.js
```diff
@@ -507,28 +511,48 @@ describe('threads', () => {
...
+            ['a', {id: 't3', last_reply_at: 5, unread_mentions: 1, unread_replies: 1}, {a: {all: ['t1', 't2'], unread: ['t2']}, b: {all: ['t1'], unread: []}}],
+            ['a', {id: 't2', last_reply_at: 40, unread_mentions: 0, unread_replies: 0}, {a: {all: ['t1', 't2'], unread: ['t2']}, b: {all: ['t1'], unread: []}}],
+            ['a', {id: 't2', last_reply_at: 40, unread_mentions: 1, unread_replies: 1}, {a: {all: ['t1', 't2'], unread: ['t2']}, b: {all: ['t1'], unread: []}}],
+            ['a', {id: 't2', last_reply_at: 40, unread_mentions: 0, unread_replies: 1}, {a: {all: ['t1', 't2'], unread: ['t2']}, b: {all: ['t1'], unread: []}}],
+            ['a', {id: 't2', last_reply_at: 5, unread_mentions: 0, unread_replies: 0}, {a: {all: ['t1', 't2'], unread: ['t2']}, b: {all: ['t1'], unread: []}}],
+            ['a', {id: 't2', last_reply_at: 5, unread_mentions: 1, unread_replies: 1}, {a: {all: ['t1', 't2'], unread: ['t2']}, b: {all: ['t1'], unread: []}}],
+
+            // DM/GM threads
+            [undefined, {id: 't3', last_reply_at: 40, unread_mentions: 0, unread_replies: 0}, {a: {all: ['t1', 't2', 't3'], unread: ['t2']}, b: {all: ['t1', 't3'], unread: []}}],
```
larkox, line 526: Why aren't we following a similar approach as channels? Channels are either from a team (private or public) or from no team (dms and gms). That way, when we have to query them it is only `thread.team === team || !thread.team`. I foresee this becoming a nightmare with instances with many teams, and potentially giving problems for visibility of dms threads. A way to deal with this is having a key in the state for undefined (i.e. `threadsInTeam: {a: ['t1', 't2'], b: ['t1'], c: [], undefined: ['t3']}`). `undefined` will be transformed to a string to work as a key, and it will be always different to the team id, so it should be safe to use. Another option is to use a specific constant.

packages/mattermost-redux/src/reducers/entities/threads/threadsInTeam.ts
```diff
@@ -45,8 +45,20 @@ function handlePostRemoved(state: State, action: GenericAction) {
...
 
-export function handleReceivedThread(state: State, action: GenericAction, extra: ExtraData) {
-    const {thread, team_id: teamId} = action.data;
+// adds thread to all teams in state
+function handleAllTeamsReceivedThread(state: State, thread: UserThread, teamId: Team['id'], extra: ExtraData) {
+    const teamIds = Object.keys(state);
+
+    let newState = {...state};
+    for (const teamId of teamIds) {
+        newState = handleSingleTeamReceivedThread(newState, thread, teamId, extra);
```
larkox, line 54: If we are going this route, shouldn't we also add these threads to a new team if/when we join a new team? (therefore applying the changes in the websocket events and around the client calls to join a team)

larkox (changes requested):
I have a few concerns about this approach. Please, take a look to the comments.

packages/mattermost-redux/src/reducers/entities/threads/threadsInTeam.ts
koox00: if the store holds no threads in a team we cannot add any single thread in the global threads, 
the reason is that we could break pagination, old.
When you first visit global threads in a team we call for getUserThreads, and then we can deduce whether to a add a single thread or not. 
The criteria is that we can add threads newer than the last thread we have in store, but not threads older than that one.

koox00: older threads will get added via scrolling/paginating

packages/mattermost-redux/src/reducers/entities/threads/threads.test.js
koox00: Not sure that it's a straight forward approach.
when you receive threads inside global threads you get all threads together DM/GM and all Channels. 
Whereas for posts in channels you only get posts in that specific channel which makes it easier to reason/handle.

larkox: I think that the logic behind channels is similar. When you call getChannelsInTeam (or something like that) you also get DM/GMs, but they are handled different.

I can see that there is some extra complexity in terms of "separating and merging" the DM/GMs from the other threads, but still I think it pays off the redundancy cost your changes are including.

That being said, I don't understand enough of how CRT is architecture on the inside, so feel free to dismiss this idea.

packages/mattermost-redux/src/reducers/entities/threads/threadsInTeam.ts
larkox: Let me know what happens in the following scenario:
- I have 3 threads on DM/GMs with mentions
- I join team B
- Threads show:
  - Option A: Nothing, since threads are not yet loaded
  - Option B: A badge with the number 3, corresponding to the 3 DM/GMs unread

From what I understand from what you explained, what happens is Option A, and I would consider that a bug. Even if we try to solve it with this approach, it would seem really error prone. Having DMs/GMs where they belong (in no team) instead of repeated on every team seems more reliable.

Another thing to consider is what happens if a thread is removed? (Root post is deleted, or anything of that sort)

packages/mattermost-redux/src/reducers/entities/threads/threads.test.js
hmhealey: The way @larkox described it is how we do it elsewhere, except instead of undefined, we use the empty string. If you dig through the code around the channel sidebar, it shows channels where either `channel.team_id === currentTeamId || channel.team_id === ''`, and the selectors are the ones in charge of that. Similarly, DMs/GMs also have their IDs stored in `state.entities.channels.channelsInTeam['']`.

We don't have to worry about paging channels though so I don't know if that'd have an effect on this piece of code at least.

packages/mattermost-redux/src/reducers/entities/threads/threadsInTeam.ts
koox00: @larkox Option B will happen. Counts store is separate from threads store. 
Imagine just refreshing the app, you have no threads in store, but we still need show the mention count.

packages/mattermost-redux/src/reducers/entities/threads/threads.test.js
koox00: Another difference is that in channels we don't care about ordering, 
but in threads we do, AKA we don't group threads from channels and DM/GMs separately.

That being said, I'll attempt to make this change since it is a very reasonable one.
 I just wanted to point out my considerations first.

koox00: According to the above we need to do some book-keeping, which is the same as pushing this change.

hmhealey: Ah Right, threads have paging unlike channels. That definitely makes this trickier, so this sounds good to me then

hmhealey (approved):

jgilliam17:
Thanks @koox00 
GM/DMs are appearing in Threads on both teams as expected.
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/204627-ffc4c37-onprem-ent-webapp-pr-9657-mattermost/mm-ee-testffc4c37/mochawesome.html) looks good, no PR related failures
Seeing few errors when manually testing, blank rhs/screen and taking a long time to load rhs (loading spinner+errors)
Screenshot 2022-06-08 115358
Screenshot 2022-06-08 115429
Screenshot 2022-06-08 120416

jgilliam17:
Discussed with Kyriakos, the errors are no longer reproducing after retesting and appear to be to network issue not PR related.
Looks good to merge.

esethna:
@amyblais please forward to v7.0

PR #9666, opened 2022-01-18
Growth spike/add basic accordion

pablovelezvidal:
#### Summary
This PR adss the basic reusable common accordion component and some changes to the texts that Michel initially added so it looks more accurated to the design.

The idea for the container is to be a common reusable component, and as you can see, from dashboard.css we can just modify the standard way it looks to make that accordion look the way we need for the design.

#### Ticket Link
N/A

#### Related Pull Requests
N/A

#### Screenshots
Captura de pantalla 2022-01-18 a las 16 54 55

#### Release Note
```release-note
NONE
```


components/common/accordion/accordion.tsx
```diff
@@ -0,0 +1,46 @@
...
+export type AccordionData = {
+    title: string;
+    description: string;
+    items: React.ReactNode;
+};
+
+type Props = {
+    items: AccordionData[];
+    openMultiple?: boolean;
+    headerClick: <T>(...args: T[]) => T | void;
```
pablovelezvidal, line 17: The accordion will contain:
- items = will be all the possible accordions elements (li) to show.
- openMultiple = allow only one accordion open at a time, or multiple accordions open at a time.
- headerClick = a cb fn that will return a ref obj pointing to the accordion, so you can use to manipulate that from where you are using the common component.

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -137,29 +123,38 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
...
+            ));
+            accData.push({
+                title: data[key].title,
+                description: data[key].description,
+                items,
+            });
+        });
+
+        return accData;
+    };
```
michelengelen, line 150: we can extract this to the FC logic rather than have a function handling that.

Just use it like this:
```suggestion
    const accData: any = [];
    const dataKeys = Object.keys(data);

    dataKeys.forEach((key: string) => {
        const items = data[key].items.map((item) => (
            <AccordionItem
                key={`${key}-item_${item.id}`}
                iconColor={getItemColor(item.status)}
            >
                <h5><i className={classNames('icon', {'icon-check-circle-outline': item.status === 'none', 'icon-alert-outline': item.status === 'warning', 'icon-alert-circle-outline': item.status === 'error'})}/>{item.title}</h5>
                <p>{item.description}</p>
            </AccordionItem>
        ));
        accData.push({
            title: data[key].title,
            description: data[key].description,
            items,
        });
    });
```

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -137,29 +123,38 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
         setTimeout(() => testSiteURL(onSuccess, onError, location.origin), 1000);
     }, []);
 
-    const dataKey = Object.keys(data);
+    const prepareDataForAccordion = () => {
+        const accData: any = [];
+        const dataKeys = Object.keys(data);
+
+        dataKeys.forEach((key: string) => {
+            const items = data[key].items.map((item) => (
```
michelengelen, line 133: nice catch: that way we can add items independent from this specific use-case! 👍🏼

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -137,29 +123,38 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
...
-                            iconColor={getItemColor(item.status)}
-                        >
-                            <h5><i className={classNames('icon', {'icon-check-circle-outline': item.status === 'none', 'icon-alert-outline': item.status === 'warning', 'icon-alert-circle-outline': item.status === 'error'})}/>{item.title}</h5>
-                            <p>{item.description}</p>
-                        </AccordionItem>
-                    ))}
-                </Accordion>
-            ))}
+            <Accordion
+                items={prepareDataForAccordion()}
```
michelengelen, line 157:
```suggestion
                items={accData}
```

components/common/accordion/accordion.scss
```diff
@@ -0,0 +1,61 @@
+.Accordion {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+
+    .accordion-item {
+        border: 1px solid #ccc;
```
michelengelen, line 7: nit: can we use a variable here to have support for a future theming approach (if we ever decide to implement that)?

components/common/accordion/accordion.scss
```diff
@@ -0,0 +1,61 @@
...
+          }
+  
+          &:after {
+            content: "";
+            flex-shrink: 0;
+            opacity: 0.5;
+            width: 18px;
+            height: 18px;
+            margin-left: auto;
+            background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill='%23212529'%3e%3cpath fill-rule='evenodd' d='M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708z'/%3e%3c/svg%3e");
```
michelengelen: i suppose this is for the chevron, right? Why don't we just use an icon there and change the chevron direction when the accordion is open?

components/common/accordion/accordion.scss
```diff
@@ -0,0 +1,61 @@
...
+  
+      .accordion-item.active .accordion-item-header::after {
+          transform: rotate(-180deg);
+      }
+  
+      .accordion-item-container {
+          transition: height 0.2s ease-in-out;
+          overflow: hidden;
+          &__content {
+            border-top: 1px solid #cccccc;
```
michelengelen: don't forget to remove the border for the first child to prevent duplicate borders

components/common/accordion/accordion.tsx
```diff
@@ -0,0 +1,46 @@
...
+};
+
+const Accordion = ({
+    items,
+    openMultiple,
+    headerClick,
+}: Props): JSX.Element | null => {
+    const [currentIdx, setCurrentIdx] = React.useState(-1);
+    const btnOnClick = (idx: number) => {
```
michelengelen, line 26: could you rename `btnOnClick` to `onButtonClick` and `idx` to `index`, please?

components/common/accordion/accordion.tsx
```diff
@@ -0,0 +1,46 @@
...
+        setCurrentIdx((currentValue: number) => (currentValue === idx ? -1 : idx));
+    };
+
+    return (
+        <ul className='Accordion'>
+            {items.map((item, idx) => (
+                <AccordionItem
+                    key={idx.toString()}
```
michelengelen, line 34: this is ok for now, but we will need a better key value here

components/common/accordion/accordion.tsx
```diff
@@ -0,0 +1,46 @@
...
+                    data={item}
+                    isOpen={idx === currentIdx}
+                    btnOnClick={() => btnOnClick(idx)}
+                    openMultiple={Boolean(openMultiple)}
```
michelengelen, line 38: we don't need the extra `Boolean` call here

components/common/accordion/accordion.tsx
```diff
@@ -0,0 +1,46 @@
...
+    openMultiple?: boolean;
+    headerClick: <T>(...args: T[]) => T | void;
+};
+
+const Accordion = ({
+    items,
+    openMultiple,
+    headerClick,
+}: Props): JSX.Element | null => {
+    const [currentIdx, setCurrentIdx] = React.useState(-1);
```
michelengelen, line 25:
```suggestion
    const [currentIndex, setCurrentIndex] = React.useState(-1);
```

components/common/accordion/accordion.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return (
+        <ul className='Accordion'>
+            {items.map((item, idx) => (
+                <AccordionItem
+                    key={idx.toString()}
+                    data={item}
+                    isOpen={idx === currentIdx}
+                    btnOnClick={() => btnOnClick(idx)}
+                    openMultiple={Boolean(openMultiple)}
+                    headerClick={headerClick}
```
michelengelen, line 39:
```suggestion
                    onHeaderClick={onHeaderClick}
```

components/common/accordion/accordion.tsx
```diff
@@ -0,0 +1,46 @@
...
+type Props = {
+    items: AccordionData[];
+    openMultiple?: boolean;
+    headerClick: <T>(...args: T[]) => T | void;
+};
+
+const Accordion = ({
+    items,
+    openMultiple,
+    headerClick,
```
michelengelen, line 23:
```suggestion
    onHeaderClick,
```

components/common/accordion/accordion.tsx
michelengelen:
```suggestion
    onHeaderClick: <T>(...args: T[]) => T | void;
```

components/common/accordion/accordion_item.tsx
```diff
@@ -0,0 +1,86 @@
...
+    isOpen: boolean;
+    btnOnClick: () => void;
+    openMultiple: boolean;
+    headerClick: <T>(...args: T[]) => T | void;
+}
+
+const AccordionItem = ({
+    data,
+    isOpen,
+    btnOnClick,
```
michelengelen, line 20:
```suggestion
    onButtonClick,
```

components/common/accordion/accordion_item.tsx
```diff
@@ -0,0 +1,86 @@
...
+    openMultiple,
+    headerClick,
```
michelengelen, line 22:
```suggestion
    onHeaderClick,
```

components/common/accordion/accordion_item.tsx
```diff
@@ -0,0 +1,86 @@
...
+}: Props): JSX.Element | null => {
+    const contentRef = useRef<HTMLDivElement>(null);
+    const itemRef = useRef<HTMLDivElement>(null);
+
+    const [height, setHeight] = useState(0);
+
+    const [open, setOpen] = React.useState(false);
```
michelengelen, line 29:
```suggestion
    const [open, setOpen] = useState(false);
```

components/common/accordion/accordion_item.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    const toggle = () => {
+        if (openMultiple) {
+            setOpen((currentValue: number) => !currentValue);
+        } else if (btnOnClick) {
+            btnOnClick();
+        }
```
michelengelen, line 36:
```suggestion
        } else if (onButtonClick) {
            onButtonClick();
        }
```

components/common/accordion/accordion_item.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const contentRef = useRef<HTMLDivElement>(null);
+    const itemRef = useRef<HTMLDivElement>(null);
+
+    const [height, setHeight] = useState(0);
+
+    const [open, setOpen] = React.useState(false);
+
+    const toggle = () => {
+        if (openMultiple) {
+            setOpen((currentValue: number) => !currentValue);
```
michelengelen:
```suggestion
            setOpen(!open);
```

components/common/accordion/accordion_item.tsx
```diff
@@ -0,0 +1,86 @@
...
+        } else if (btnOnClick) {
+            btnOnClick();
+        }
+
+        if (headerClick) {
+            headerClick(itemRef);
+        }
```
michelengelen, line 40:
```suggestion
        if (onHeaderClick) {
            onHeaderClick(itemRef);
        }
```

michelengelen (commented):
couple of comments for now ... some nitpickies ... looks good so far! 🚀

components/common/accordion/accordion_item.tsx
pablovelezvidal: since for this change we relay on the previous step, it is better to use the prevState callback function.

components/common/accordion/accordion_item.tsx
pablovelezvidal: for mocking the state values, it is necessary to use React.useState. It will allow you to mock the useState return using jest without using the hooks-testing-framework.

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -141,32 +151,38 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
         setTimeout(() => testSiteURL(onSuccess, onError, location.origin), 1000);
     }, []);
 
-    const dataKey = Object.keys(data);
+    const accData: AccordionDataType[] = [];
+    const dataKeys = Object.keys(data);
+
+    dataKeys.forEach((key: string) => {
```
mkraft, line 157: Another option is something like this:
```js
Object.entries(data).map(([accordionKey, accordionData]) => {
    console.log('accordionKey', accordionKey);
    console.log('accordionData', accordionData);
});
```

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -141,32 +151,38 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
...
+                iconColor={getItemColor(item.status)}
+            >
+                <h5><i className={classNames('icon', {'icon-check-circle-outline': item.status === 'none', 'icon-alert-outline': item.status === 'warning', 'icon-alert-circle-outline': item.status === 'error'})}/>{item.title}</h5>
+                <p>{item.description}</p>
+            </AccordionItem>
+        ));
+        accData.push({
+            title: data[key].title,
+            description: data[key].description,
+            icon: data[key].icon,
```
mkraft, line 170:
```suggestion
        const {title, description, icon} = data[key];
        accData.push({
            title,
            description,
            icon,
```

components/common/accordion/accordion.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+type Props = {
+    items: AccordionDataType[];
+    openMultiple?: boolean;
+    onHeaderClick?: <T>(ref: RefObject<HTMLLIElement>) => T | void;
+};
+
+const Accordion = ({
+    items,
+    openMultiple,
```
mkraft, line 23: I'm a bit confused about the `openMultiple` prop. I don't see `<AccordionItem>` components having any collapse/expand states in the designs, but yet they're receiving the prop.

Is this a way to configure whether multiple `<Accordion>` components are allowed to be expanded at once? If so I would question whether a `<Accordion>` (let alone an `<AccordionItem>`) should have knowledge about the state of other sibling components, and would suggest perhaps an `<AccordionList>` that contains that logic.

I may be totally misunderstanding something though.

mkraft (commented):
This is coming along really well!

michelengelen (approved):
LGTM

components/admin_console/workspace-optimization/dashboard.tsx
michelengelen: @mkraft is right ... the `entries` method might be the better option here

components/admin_console/workspace-optimization/dashboard.tsx
michelengelen: with the beforementioned usage of `Object.entries()` you would not even need to destruct from a member access but just from the data object that is being passed in from the `entries` method

components/common/accordion/accordion.tsx
michelengelen: @mkraft you are right about the `Accordion` component the siblings don't have any insight on the open/closed state of others though

pablovelezvidal: I don't see <AccordionItem> components having any collapse/expand states in the designs

> Captura de pantalla 2022-01-21 a las 13 52 28
> hey @mkraft , so,  as shown above, left side is collapsed state, right side is expanded state.


Is this a way to configure whether multiple <Accordion> components are allowed to be expanded at once?

> Yes, this defines if only one item open at a time or multiple. This is covered in the testing for accordion too. I moved the control for this logic to the Accordion and the tests passed ok, since the test focus on the behavior and not in the internal implementation https://github.com/mattermost/mattermost-webapp/pull/9666/files#diff-5dc32f53bc2f17a06eb337f21adeca4ec4d8fc25c3548edb212899ca3679eec7R79

If so I would question whether a <Accordion> (let alone an <AccordionItem>) should have knowledge about the state of other sibling components, and would suggest perhaps an <AccordionList> that contains that logic

> Accordion now have the control over which AccordionItems are expanded/Collapsed

pablovelezvidal:
Captura de pantalla 2022-01-23 a las 23 45 13
Validate all the svgs look ok

PR #9289, opened 2021-10-29
[MM-29622] Collapsed reply threads tour

AshishDhama:
#### Summary
This will add a Collapsed reply threads tour when the admin switches on the feature based on the following document https://mattermost.atlassian.net/wiki/spaces/Threads/pages/1074757633/Desktop+Webapp+Onboarding we have added the trigger logic. As we have modified the existing tutorial tip Component so that will also require some testing to verify everything is working as usual.
Here is a video on how to reset tour settings via API https://www.loom.com/share/3340fc15d24a4324a6841b69e5c0627a
(Categories are  ``crt_thread_pane_step``,  ``crt_tutorial_step`` ,``crt_tutorial_triggered``)

Note: For the CRT tour to trigger you have to finish/skip the onboarding tour and at least have 1 thread.

#### Ticket Link
   Fixes https://mattermost.atlassian.net/browse/MM-29622

#### Related Pull Requests
``None``

#### Screenshots
|  before  |  after  |
|----|----|
| None | https://www.loom.com/share/c30ed9581ee44d8aaa342b14c442a7c0 |

#### Release Note
```release-note
* Adding CRT tour functionality, after modifying tutorial tip component to support multiple tours.
```

PR #9675, opened 2022-01-19
[MM-40486]: fixes "non-showing" confirm button in CRT modal

michelengelen:
#### Summary
CRT modal now shows the "Got It" Button in the Modal Footer. 
The Modal expects the button to have a handler function, that's why it was not showing!

#### Ticket Link
[MM-40486](https://mattermost.atlassian.net/browse/MM-40486)

#### Release Note
```release-note
NONE
```


koox00 (approved):

matthewbirtch (approved):
Verified that the button now displays properly. Thanks!

image

AshishDhama:
@esethna do we still need this change at least on master(merged CRT tour branch) we have changed this modal.
Screenshot 2022-01-25 at 1 20 55 PM

michelengelen:
@matthewbirtch i think we can close this in favor of #9289

esethna:
@AshishDhama @michelengelen is there a way to keep both modals (the known issues warning + the CRT tour modal) until CRT goes into GA?

jgilliam17 (approved):
Thanks @michelengelen 
Tested, looks good to merge - if we decide to keep this PR in addition to CRT tour.
- Verified button is now present on early access CRT modal.

michelengelen:
@esethna I will have a look

michelengelen:
from what I have seen in my research it should be possible to keep both ... we only need to somehow determine a flag of sorts to check if the CRT has reached GA ... that might prove to be difficult ... 

What is the ETA for GA @esethna ?

I am asking because we can cherry pick this fix onto an earlier release and keep the change on master as is. We might need a follow-up to tidy up the old implementation though

esethna:
@michelengelen CRT may not be in GA until mid this year

Willyfrog:
@esethna @michelengelen what are the NS on this?

michelengelen:
@Willyfrog it proves to be difficult to keep both modals at the same time. This is due to the fact, that @AshishDhama did overwrite the component and he was 100% right to do that. But now we sit between two states where we either separate the 2 modals into different components or stick with a "broken" state until we reach GA for the improved modal.

WDYT @Willyfrog @esethna ?

Willyfrog:
I'm ok with broken until GA as long as users cand dismiss it somehow, is there some extra work we need to do before GA?

michelengelen:
nothing I am aware of ... the users can just dismiss the modal by clicking on the backdrop.

esethna:
> or stick with a "broken" state until we reach GA for the improved modal

I'm not sure I'm following. Didn't Dhama remove the modal completely? What configuration would cause the old warning dialog to show at all? I couldn't get it to appear at all.

michelengelen:
@esethna I don't know why the modal is not showing anymore ... typically it should appear when you enable CRT after closing the settings modal. I can only assume there have been some changes that prevent that from happening since it is also not opening on my local instance (on master) anymore

AshishDhama:
@michelengelen @esethna there was a setting in user settings, which was triggering CRT warning modal.
@michelengelen If you don’t mind will take it over I know what needs to be reverted, I think reverting changes done here https://github.com/mattermost/mattermost-webapp/pull/9289/files#diff-a5b7fdfad9904bb72d9a8fe6dea68ea532feccaf11a088cc9378bf9c1e87083d. will do it.

michelengelen:
@AshishDhama sure, go ahead! Anything off my plate is welcomed and appreciated! 🙇🏼

AshishDhama:
Closing in favour of https://github.com/mattermost/mattermost-webapp/pull/9910

PR #9676, opened 2022-01-19
MM-40527 Make reducers responsible for cleaning up on logout

hmhealey:
As part of getting redux-persist upgraded and used properly, I'm doing a bunch of cleanup of the store initialization logic so that it's less complicated and easier to understand. More cleanup will come in the redux-persist upgrade PR, but I'm splitting it up a bit to make things easier to review

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40527

#### Release Note
```release-note
NONE
```


packages/mattermost-redux/src/reducers/entities/search.ts
```diff
@@ -71,7 +71,7 @@ function matches(state: Record<string, string[]> = {}, action: GenericAction) {
     }
     case SearchTypes.REMOVE_SEARCH_POSTS:
     case UserTypes.LOGOUT_SUCCESS:
-        return [];
```
hmhealey, line 74: A few of these changes are actually fixing some places where we changed the type of the state accidentally

packages/mattermost-redux/src/store/helpers.ts
```diff
@@ -15,18 +13,7 @@ export function createReducer(baseState: GlobalState, ...reducers: Reducer[]): R
     reducerRegistry.setReducers(Object.assign({}, ...reducers));
     const baseReducer = combineReducers(reducerRegistry.getReducers());
 
-    // Root reducer wrapper that listens for reset events.
-    // Returns whatever is passed for the data property
-    // as the new state.
-    const offlineReducer = (state: GlobalState, action: Action) => {
-        if ('type' in action && 'data' in action && action.type === General.OFFLINE_STORE_RESET) {
-            return baseReducer(action.data || baseState, action);
```
hmhealey, line 23: This is the first part of the important change in this PR. On logout, we used to overwrite the whole state tree which seems incorrect when there's some state we might want to save like information about the window size and the client config.

I think the original reason for doing this might have been that there was also some state outside our control which came from redux-offline (which is why this code refers to "offline" a few times). Since we haven't used that in a while, we don't need that part either

store/index.js
```diff
@@ -109,11 +108,6 @@ export default function configureStore(initialState) {
                             // used by the application such as extra and redirect_to.
                             window.location.href = `${basePath}${window.location.search}`;
 
-                            store.dispatch({
```
hmhealey, line 112: This is the other important part. This is the action that's dispatched to cause the state tree to be overwritten

M-ZubairAhmed (approved):
Its coming up brilliantly cant wait to see the follow pr for rest of the part

koox00 (approved):

jgilliam17:
Thanks @hmhealey 
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/107989-a9e0b5b-onprem-ent-webapp-pr-9676-mattermost/mm-ee-testa9e0b5b/mochawesome.html) looks good, no PR related failures.
Let me know specifically what needs to be tested manually here? Thanks.

jgilliam17:
@hmhealey Logging in/out works as expected. There was only one issue I mentioned when logging out on desktop app, a warning is present, disappears quickly.
Let me know if this is related. I am able to log back in without experiencing any issues or errors despite the previous warning.
Screen Shot 2022-01-25 at 3 20 48 PM

hmhealey:
@jgilliam17 I'm not able to repro that myself, but I don't think it's that big of an issue. The login/logout flow isn't great to begin with, so I'm not convinced it was something specifically introduced by this PR.

jgilliam17:
Thanks for checking @hmhealey - It's good to merge then.

jgilliam17 (approved):

PR #9677, opened 2022-01-19
[MM-40590] Fix inviting a guest guest to a channel where he's a member.

JulienTant:
#### Summary
This PR fixes the bug that caused inviting a guest to a channel where he's already a member to show a "Success" message instead of a message saying the guest is already present.

Unrelated to the ticket but fixed here: during the development, I noticed that if the inviter is not in a channel when trying to invite a people (e.g. if he is on the /tips page during the onboarding process), then clicking on the "Invite As: Guest" makes the webapp crash. I pushed to fix to use the "Town Square" when there's no active channels and added the tests for it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40590

#### Related Pull Requests
N/A

#### Screenshots
N/A

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/invitation_modal/add_to_channels.test.tsx
+++ b/components/invitation_modal/add_to_channels.test.tsx
@@ -2,14 +2,13 @@
 // See LICENSE.txt for license information.
 
 import React from 'react';
-import {IntlProvider} from 'react-intl';
-import {render, screen} from '@testing-library/react';
+import {screen} from '@testing-library/react';
 
 import {mountWithIntl} from 'tests/helpers/intl-test-helper';
+import {renderWithIntl} from 'tests/react_testing_utils';
 import deepFreeze from 'mattermost-redux/utils/deep_freeze';
 import {Channel} from 'mattermost-redux/types/channels';
 import CloseCircleIcon from 'components/widgets/icons/close_circle_icon';
-import en from 'i18n/en.json';
 
 import AddToChannels, {Props} from './add_to_channels';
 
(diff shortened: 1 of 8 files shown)


neallred (approved):
Good find and fixes! If I understand the issue, it is that the channel membership data may or may not be loaded yet, so when members was selected, it didn't necessarily show that some guests already belonged to the channel. This PR changes that by making sure the relevant data is loaded first.

components/invitation_modal/add_to_channels.test.tsx
```diff
@@ -28,15 +30,45 @@ const defaultProps: Props = deepFreeze({
...
+    </IntlProvider>
+);
+
 describe('AddToChannels', () => {
     beforeEach(() => {
         props = defaultProps;
     });
+
+    describe('placeholder selection', () => {
+        it('should use townSqureDisplayName when not in a channel', () => {
```
neallred, line 53: Nice tests!

neallred (approved):

saturninoabril (changes requested):
Thanks @JulienTant! Tested and passed. Left minor comments but all good to me.

saturninoabril (approved):
Thanks @JulienTant. Tested and passed. LGTM.

stevemudie (approved):
Unable to invite a guest guest twice to the same channel as described in the PR title. 😛

Looks good @JulienTant!

PR #9653, opened 2022-01-15
chore/MM-40559 : Add memoization to getOneClickReactionEmojis

M-ZubairAhmed:
#### Summary
Added memoized selector for `getOneClickReactionEmojis` redux selector

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40559

#### Related Pull Requests
None

#### Screenshots
None

#### Release Note

```release-note
NONE
```

PR #9682, opened 2022-01-20
Removing unnecessary re-renders of PostInfo due to lack of memoization of recent emojis

jwilander:
#### Related Pull Requests

#### Release Note

```release-note

```

### DIFF
--- a/selectors/emojis.ts
+++ b/selectors/emojis.ts
@@ -1,6 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import deepEqual from 'fast-deep-equal';
+
 import {createSelector} from 'reselect';
 
 import {getCustomEmojisByName} from 'mattermost-redux/selectors/entities/emojis';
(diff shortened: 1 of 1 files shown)


selectors/emojis.ts
```diff
@@ -51,10 +54,21 @@ export function isCustomEmojiEnabled(state: GlobalState) {
...
-    const emojis = recentEmojis.map((recentEmoji) => emojiMap.get(recentEmoji)).filter(Boolean).reverse();
+    // Temporary deep equal check to prevent PostInfo components from re-rendering in many unnecessary scenarios.
+    // The proper fix is to turn this function into a proper selector but that's more complicated than it seems
+    // because getRecentEmojis never returns a new array and the array it returns originally is modified mutably
+    // when a new recent emoji is added, meaning that the selector would never recalculate. Fixing that is another
+    // rabbit hole that would make more sense to do with the other storage refactor.
+    const emojis = threeMostRecentEmojis.map((recentEmoji) => emojiMap.get(recentEmoji)).filter(Boolean).reverse();
+    if (deepEqual(emojis, prevThreeMostRecentEmojis)) {
+        return prevThreeMostRecentEmojis;
+    }
```
jwilander, line 70: @hmhealey I was doing some profiling and I noticed that the PostInfo components were getting re-rendered a lot more often than they should and because of weird events. For example, receiving a 'user_typing' event, regardless of what channel it was in, would cause all the PostInfo components to re-render.

Using why-did-you-render I saw that it was the `recentEmojis` prop causing the problem:

Screen Shot 2022-01-20 at 3 37 44 PM

The short version is it's doing that because `getOneClickReactionEmojis` isn't memoized as a proper selector. Digging into it more revealed it wasn't a simple fix (see my code comment for why). Is this something that might get fixed properly by your storage system changes? I didn't want to keep going if this was something the new storage changes would resolve.

cc @M-ZubairAhmed for visibility

M-ZubairAhmed: Hey we did memoized this in https://github.com/mattermost/mattermost-webapp/pull/9653

jwilander: Oh, awesome. Good stuff! I'll close this.

PR #9684, opened 2022-01-21
[MM-41115] Add option to colorize usernames in compact mode

3ach:
#### Summary

This pull-request adds a new option which is available in compact-display mode. Usernames are colorized according to a consistent hash. This means that all users will see the same color for a given user. 

The option defaults to 'true', but is persistent if changed. If a user disables colorization, modifies their overall message display preference, and then returns to the compact view, their usernames will remain uncolorized. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41115
fixes mattermost/mattermost-server#19368

#### Related Pull Requests

- Has server changes: https://github.com/mattermost/mattermost-server/pull/19382

#### Screenshots

|  before  |  after  |
|----|----|
|  |   |
|  |  |

#### Release Note

```release-note
* Added option to colorize usernames in compact display mode. 
```


3ach:
Whoops, this is missing a change from my computer. One moment...

esethna:
Thanks @3ach! Few comments:

1. What's the logic that determines what color each user will be and the minimum amount of differentiation between the colors? For example will two users names user1 and user2 have very similar colors?
2. It looks like the coloring is affecting standard view as well (it should only add color in compact mode as discussed)
3. Can we try to match the styling of the account settings with the pipe separator as per the mock design (similar to Desktop Notification settings)
image

3ach:
1. I use the [color-hash](https://www.npmjs.com/package/color-hash) library. There are no current guarantees about the differentiation between colors. 
2. Yes, this is the change coming shortly that I forgot to amend in when I put this up for review.
3. Yes, will fix styling now.

3ach:
OK. Updated the view (and my screenshots) to match the other UI.

esethna (approved):
Looks good from my perspective. @matthewbirtch any comments or concerns before we proceed with this feature enhancement?

Willyfrog:
fixes https://github.com/mattermost/mattermost-server/issues/19368

matthewbirtch:
> Looks good from my perspective. @matthewbirtch any comments or concerns before we proceed with this feature enhancement?

I do have concerns about the contrast and legibility with the colors here. Does the library account for proper contrast ratios? Also, how does this apply across the different themes? I imagine it would need to be a different set of colors for dark themes for example.

The other concern I have from a design perspective is the visual noise created by the rainbow of colors that could get introduced by this change with channels that have a lot of active users. I guess it's ok if it is a configurable setting. @esethna was this proposal vetted by the design team before the ticket was created?

3ach:
The library currently takes no other input than the username, so it doesn’t offer guarantees about contrast. We can constrain its color range using information about the background to make the colors more readable.

3ach:
This type of feature really shines when you do use it in a large, high traffic channel. If usernames tend to be around the same length, another quick visual difference helps scan through quickly.

matthewbirtch:
I'm okay with the idea of this, but we need to try and make sure the contrast reaches a ratio of 4.5 (or more) comparing the text color to the background. Is that possible?

esethna:
@matthewbirtch apologies this didn't reach the design team before implementation. @3ach was keen and the idea was solid, so we pushed forward.

+1 and good call to constraining the colors to ensure the contrast with the channel BG is sufficient

devinbinnie:
@3ach @matthewbirtch Is there anything further that needs to be worked out here?

matthewbirtch:
> Is there anything further that needs to be worked out here?

@devinbinnie, a solution is needed to ensure colors meet the contrast ratio. I think that's it.

3ach:
Sorry, I will work on this, I have just gotten very busy on other things and haven't had time to patch this up. I hope to have the time in the second half of this week or early next.

esethna:
Great, thanks @3ach!

calebroseland:
Removing myself for now; feel free to add me again once it's ready 🙂.

darias416:
> Sorry, I will work on this, I have just gotten very busy on other things and haven't had time to patch this up. I hope to have the time in the second half of this week or early next.

Hi @3ach how are things coming along with the contrast ratio? Is there anything we can do to help?

3ach:
Thanks for your patience.
I tried several approaches here, and I'm honestly not an enormous fan of the one I landed on. Basically, check the hashed contrast from the theme background, and if it's not high enough, salt the username, rehash, and check again. Repeat until satisfied. Obviously this is not the most efficient way to do this, but I there were two constraints which I think make something like this the only option:
1. usernames are consistent, if given the same background, someone will have the same color across machines/users/whatever.
2. viewing users can have different backgrounds, so storing a single color per-user doesn't work.

Intuitively, I think that there should be some way to instead constrain a color space based on the background, and then generate a single hash within those constraints. But after reading quite a lot of online information about colors... I couldn't figure out how to do that.

esethna:
Thanks @3ach, I'm going to re-request dev review give this approach to see if there's any concerns cc// @devinbinnie

matthewbirtch (approved):
Contrast seems much better with this now. I'll leave it to the developer reviewers to comment on the technical implementation of it though.

components/user_profile/user_profile.tsx
```diff
@@ -54,6 +60,20 @@ export default class UserProfile extends PureComponent<UserProfileProps> {
         this.overlay = ref;
     }
 
+    generateColor = (username: string, background: string): string => {
+        let userColor = background;
+        let userAndSalt = username;
+        const checker = new ContrastChecker();
+        const colorHash = new ColorHash();
+
+        while (!checker.isLevelCustom(userColor, background, 4.5)) {
```
devinbinnie, line 69: I'm concerned about the time added to the `render()` function by this `while` loop.
From what I can tell (correct me if i'm wrong) we are basically just rolling the dice over and over using the color hash and adding strings as we go until we get a color that we deem acceptable.

This feels like on a larger scale (think like 1000+ different users in one channel for example) that we might end up introducing some poor performance by the `user_profile` component.

I'm wondering can we add some options to the `ColorHash` object (see here: https://github.com/zenozeng/color-hash#custom-hue) that dictate what the HSV values are bound within? That way we could ahead of time calculate what HSV values would the minimum/maximum required to provide adequate contrast? It would at least mitigate the amount of loops that the code go through before it finds a colour.

cc @hmhealey in case he has any ideas.

3ach: Yep, I considered this. I spent a lot of time playing around with the lightness constraints that ColorHash offers. Unfortunately lightness of the color is not the same as relative luminance, and this led to basically indistinguishably-dark or -light versions of the colors (think extreme pastels or earth tones.)

One mitigating factor is that users don't affect each other -- eg calculating the color constant for each user is dependent only on themselves and the background color. So there's not going to be quadratic explosion in the number of users, it's linear. 

I could try strategies like caching the background color + user so that it's only calculated once per theme refresh. I ran an experiment on a few thousand randomly generated usernames and the default background colors and I think it took at most ~4 iterations to find an acceptable color. I can go find the data from that if it helps.

devinbinnie: Thanks for the insight @3ach.
I think that its promising that it likely won't take many iterations to find the correct colour, and I definitely like the idea of caching this information so that we don't update it unless absolutely necessary, that's certainly a step in the right direction.

I'm still a little concerned what the impact might be on lower-end systems with thousands of users. I would like to see what the impact would be with some CPU throttling added with a few thousand users in the channel. Fortunately, it's unlikely we'll have 1000+ users rendered all at once since the center channel is designed not to render too many post components.

3ach: Okay, sure, I'm happy to add the caching. I have no idea how I could run a test with CPU throttling... is there a guide or other accepted practice you are aware of?

devinbinnie: I believe there's a way you can do it in Developer Tools on Chrome (there's a CPU throttling dropdown somewhere, I forget exactly where). As for measurements, I think just a profile comparison (before and after caching) might yield some insights for us.

hmhealey: Perhaps we could add a cap on that where we just give up and accept that a colour is a duplicate after a certain number of retries? We've done something similar in a few other places where something could theoretically take a very long time in rare cases, and while that case shouldn't come up in practice, it's nice to know it'll always perform at least okay

3ach: @devinbinnie - I can go ahead and do a perf test as you described.
@hmhealey - The iterations are to ensure high contrast between background and a given username, not differentiation between usernames. If it's all right to give up on the contrast after a certain number of iterations, I'm happy to add that in.

esethna:
Hey @3ach, how's this going? Are you blocked on anything?

3ach:
I tried gathering profiles in Firefox and Google Chrome and was unable to get a good handle on the perf of this `generateColor` function. Here is a bounded + cached version for review.

esethna:
@3ach is there a missing attachment in your post above?

3ach:
No, I meant that I'd updated my PR to contain bounding (only 10 attempts to hash) and caching (per-username) for review.

esethna:
Ah thanks. @devinbinnie @hmhealey ^

package.json
```diff
@@ -15,6 +15,15 @@
     "chart.js": "2.9.4",
     "classnames": "2.3.1",
     "core-js": "3.21.1",
+    "@types/color-hash": "1.0.2",
+    "@types/country-list": "2.1.0",
+    "@types/react-overlays": "1.1.3",
+    "bootstrap": "3.4.1",
```
devinbinnie, line 21: I think some of these packages were purposely removed, can we rebase from master and make sure we're only adding the packages needed for this change?

components/user_profile/user_profile.tsx
```diff
@@ -54,6 +61,29 @@ export default class UserProfile extends PureComponent<UserProfileProps> {
...
+        if (cachedColor !== null) {
+            return cachedColor;
+        }
+
+        let userColor = background;
+        let userAndSalt = username;
+        const checker = new ColorContrastChecker();
+        const colorHash = new ColorHash();
+
+        let tries = 10;
```
devinbinnie, line 76: Personally I'd be in favour of less tries here (2-3 would be my feeling) but I'm not sure to what degree that will affect performance.

I'm happy with the change to this though, so only 1/5 on the actual number.

devinbinnie (changes requested):
Thanks for handling the changes @3ach!
There was only one blocking comment I had below.

components/user_profile/user_profile.tsx
3ach: 3 sounds good.

package.json
3ach: Oh duh. I actually thought I fixed these. I'll rebase and fix again.

3ach:
Not sure what's up with the CI tests, it seems like they're missing dependencies like eslint and tsc?

devinbinnie (approved):
Thanks @3ach! This is looking good now.

jgilliam17:
Thanks @3ach 
Looks great! Waiting for E2E report to complete.
Manually tested and seeing few small issues.
1. Usernames on RHS are not colorized 
Screen Shot 2022-04-29 at 2 19 28 PM

2. With Collapsed Reply Threads enabled, usernames are not colorized on global Threads inbox
Screen Shot 2022-04-29 at 2 26 19 PM

jgilliam17:
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/189315-c17c379-onprem-ent-webapp-pr-9684-mm-ee-test:c17c379/mochawesome.html) had some skipped tests. Updating and re-running.

jgilliam17:
Thank you  @3ach 
[E2E report](https://automation-dashboard.vercel.app/cycles/f582c7cb-d739-487e-b8d5-896dc3c69042) looks good, no PR related failures.
There are few conflicts that need to be resolved. Can you please take a look and fix - thanks 🙂

esethna:
@3ach are able to help with the conflicts?

3ach:
Rebased

esethna:
cc// @jgilliam17 for QA

hmhealey:
@AshishDhama is actually the other dev we're waiting for reviews on. I just show up in the Reviewers list since I commented on the PR to answer a question

AshishDhama (approved):
LGTM

esethna:
Thanks @jgilliam17 please review and merge if this all looks good. Thanks!

jgilliam17:
@3ach I am not seeing colorization of names on the replies and threads in compact mode. See my previous [comment](https://github.com/mattermost/mattermost-webapp/pull/9684#issuecomment-1113608999)
There are also few conflicts that need to be resolved. Thanks 🙂

esethna:
@3ach, let us know if you're still working on this? Thanks!

3ach:
All right, I've fixed up most the issues. There remains one spot uncolorized, which I would like some input on before moving forward. 

The global threads view (defined in `components/threading/global_threads/thread_item/thread_item.tsx`) doesn't use `UserProfile` like everywhere else does to display a username. It simply yanks the author from the post and uses that. I don't know why it's designed like that, otherwise I would replace the span [here](https://github.com/mattermost/mattermost-webapp/blob/9f519001f689e59a99d6c1f8f9413a1560d66904/components/threading/global_threads/thread_item/thread_item.tsx#L166) with an instance of `UserProfile`, which would resolve the issue and make things more standard. It could also be fine to leave this instance uncolorized, since I don't think this view changes at all between compact and normal. Up to you folks. 

Screenshot 2022-06-17 at 11-26-31 Threads - Zach's Team Mattermost

Willyfrog:
@koox00 can you help with the question?

matthewbirtch:
> All right, I've fixed up most the issues. There remains one spot uncolorized, which I would like some input on before moving forward.
> 
> The global threads view (defined in `components/threading/global_threads/thread_item/thread_item.tsx`) doesn't use `UserProfile` like everywhere else does to display a username. It simply yanks the author from the post and uses that. I don't know why it's designed like that, otherwise I would replace the span [here](https://github.com/mattermost/mattermost-webapp/blob/9f519001f689e59a99d6c1f8f9413a1560d66904/components/threading/global_threads/thread_item/thread_item.tsx#L166) with an instance of `UserProfile`, which would resolve the issue and make things more standard. It could also be fine to leave this instance uncolorized, since I don't think this view changes at all between compact and normal. Up to you folks.
> 
> Screenshot 2022-06-17 at 11-26-31 Threads - Zach's Team Mattermost

Let's leave the colonization out of this context @3ach and just keep it to the messages themselves.

3ach:
Ok, great. Then I think this PR is complete.

jgilliam17 (approved):
Thanks @3ach for the fix
Tested, looks good to merge.
- Verified settings option to colorize usernames when using compact message display. Usernames will be colorized on main channel view and threads, but not on global threads inbox.
- [E2E report](https://mattermost-cypress-report.s3.amazonaws.com/209925-634b86a-onprem-ent-webapp-pr-9684-mattermost/mm-ee-test634b86a/mochawesome.html) looks good, no PR related failures

3ach:
@jgilliam17 I'm not really sure if I've done it right, but I copied how the other strings were done in this file.

jgilliam17:
Thanks @3ach 
I believe another help text string also needs to be added to translation file  "Use colors to distinguish users in compact mode"
@devinbinnie do you mind giving i18n changes a quick look? Thanks

3ach:
OK. Translation done.

jgilliam17:
@hmhealey We are waiting for the related server PR checks to be fixed so we can merge at the same time

PR #9688, opened 2022-01-21
[MM-37245] Format the number of users in enterprise license details

JulienTant:
#### Summary
Format the number of users attached to a license in "System Console > About > Edition and License".

Also applied formatting in the Upload license Modal confirmation message.

**Note:** A comment in the code reminds us that the labels must not be translated (because the labels are linked to the license that must stay in English) - that's why in the screenshot below, the French locale still has English labels.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37245

#### Related Pull Requests
N/A

#### Screenshots
Locale = US (separator is `,`)
image

Locale = FR (separator is a space)
image

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/admin_console/license_settings/enterprise_edition/enterprise_edition_left_panel.test.tsx
+++ b/components/admin_console/license_settings/enterprise_edition/enterprise_edition_left_panel.test.tsx
@@ -1,0 +1,49 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+import {mountWithIntl} from 'tests/helpers/intl-test-helper';
+
+import EnterpriseEditionLeftPanel, {EnterpriseEditionProps} from './enterprise_edition_left_panel';
+
+
+describe('components/admin_console/license_settings/enterprise_edition/enterprise_edition_left_panel', () => {
+    const license = {
+        IsLicensed: 'true',
+        IssuedAt: '1517714643650',
+        StartsAt: '1517714643650',
+        ExpiresAt: '1620335443650',
+        SkuShortName: 'Enterprise',
+        Name: 'LicenseName',
+        Company: 'Mattermost Inc.',
+        Users: '1000000',
+    };
+
+    const props = {
+        license,
+        openEELicenseModal: jest.fn(),
+        upgradedFromTE: false,
+        isTrialLicense: false,
+        issued: <></>,
+        startsAt: <></>,
+        expiresAt: <></>,
+        handleRemove: jest.fn(),
+        isDisabled: false,
+        removing: false,
+    } as EnterpriseEditionProps;
+
+    test('should format the Users field', () => {
+        const wrapper = mountWithIntl(
+            <EnterpriseEditionLeftPanel {...props}/>,
+        );
+
+        const item = wrapper.find('.item-element').filterWhere((n) => {
+            return n.children().length === 2 &&
+                   n.childAt(0).type() === 'span' &&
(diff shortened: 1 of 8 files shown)


pablovelezvidal (approved):
Thanks! Great Work @JulienTant :shipit:

components/admin_console/license_settings/enterprise_edition/enterprise_edition_left_panel.test.tsx
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+import {mountWithIntl} from 'tests/helpers/intl-test-helper';
+
+import EnterpriseEditionLeftPanel, {EnterpriseEditionProps} from './enterprise_edition_left_panel';
+
```
pablovelezvidal, line 9: please, remove extra blank line. Before pushing changes, if you run `make fix-style` it will do it for you or will warn you about what needs to be fixed.

JulienTant: Fixed!

components/admin_console/license_settings/enterprise_edition/enterprise_edition_left_panel.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+        isDisabled: false,
+        removing: false,
+    } as EnterpriseEditionProps;
+
+    test('should format the Users field', () => {
+        const wrapper = mountWithIntl(
+            <EnterpriseEditionLeftPanel {...props}/>,
+        );
+
+        const item = wrapper.find('.item-element').filterWhere((n) => {
```
JulienTant, line 40: Is there a better way to target the content here? Rendered HTML code looks like this:

```html
<div class="licenseElements">
    <div class="item-element">
        <span class="legend">START DATE:</span>
        <span class="value"><span>6/7/2021</span></span>
    </div>
[...]
    <div class="item-element">
        <span class="legend">USERS:</span>
        <span class="value"><span>200,000</span></span>
    </div>
[...]
</div>
```

I could not find a semantic way to tackle this

stevemudie:
@JulienTant The thousands separator is in place as expected. Nice!

However, I noticed in the pop-up (new since this ticket was written last July) still lacks it.

Is this something that can be fixed in this PR?

If it needs a new ticket just let me know.
Missing comma in pop-up

JulienTant:
Don't bother with a new ticket I'll add this here!

JulienTant:
Hey @stevemudie, I pushed a fix for the modal. It appears that the modifications for this had to be made in the translation files. I also applied the change to all languages that contained this translation string already. I'm not sure if changing other locales was OK to do from the code, so I did it in a separate commit that I can drop if this is not desired.

components/admin_console/license_settings/enterprise_edition/enterprise_edition_left_panel.test.tsx
pablovelezvidal: well, this seem to work, but I agree that it's too elaborate for getting an element. If you like, you could check if something like `wrapper.find({children: "USERS:"]})` works or set a `data-testid` attribute to use it to target that element. If any of those work for you, what you have already set is ok.

AGMETEOR (approved):

stevemudie (approved):
Looks good! Thanks for the fix to the modal as well @JulienTant

PR #9690, opened 2022-01-22
MM-36708 Using moment.js to format dates

mhblatch:
#### Summary
I used documentation from react-day-picker to change the output of dates to DD-MM-YYYY instead of M-D-YYYY as it current is. I used this link to change the code: https://react-day-picker.js.org/examples/input-moment/

#### Ticket Link
Custom Status Expiry: Date Format #19358 https://github.com/mattermost/mattermost-server/issues/19358 

#### Release Note

Imported moment.js to parse selected date using the code outlined in the react-day-picker documentation. Read more here: https://react-day-picker.js.org/examples/input-moment/

### DIFF
--- a/components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.tsx
+++ b/components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.tsx
@@ -17,6 +17,13 @@ import MenuWrapper from 'components/widgets/menu/menu_wrapper';
 import './dnd_custom_time_picker_modal.scss';
 import {toUTCUnix} from 'utils/datetime';
 
+import MomentLocaleUtils, {
+    formatDate,
+    parseDate,
+  } from 'react-day-picker/moment';
+  
+
+
 type Props = {
     onExited: () => void;
     userId: string;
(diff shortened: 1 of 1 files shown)


hmhealey (changes requested):
Thanks for the PR, @mhblatch! This looks good to start with, but I have a few suggestions for things to be changed:
1. The `lint` check that's currently failing is because of some code styling issues if you could take a look at those
    ```
    /root/mattermost/mattermost-webapp/components/dnd_custom_time_picker_modal/dnd_custom_time_picker_modal.tsx
    6:20  error  Strings must use singlequote        quotes
    184:44  error  Unexpected usage of doublequote     jsx-quotes
    185:44  error  There should be no space after '='  react/jsx-equals-spacing
    
   ✖ 3 problems (3 errors, 0 warnings)
    3 errors and 0 warnings potentially fixable with the `--fix` option.
    ```
2. Similarly, the `type-check` check is failing because the types for some of those fields is incorrect. In particular, I think we've defined `this.state.selectedDate` to be a Date, not a string, so you'd want to change that.
3. Separate from those, we've actually been trying to move away from moment.js since it's very large and we're not using enough of its features to be worth that. Could you perhaps change this to use [Luxon](https://moment.github.io/luxon/#/) instead? We're working on switching moment for Luxon throughout the code base right now.
4. Finally, could you change the date format to pad the day/month with a 0 so that they're always two digits long as per linked ticket?

Also, we aren't able to accept your submission unless you sign our CLA. If you take a look at the message from Mattermod above, it has instructions on how to do that.

mylonsuren:
Hey @mhblatch, just checking in to see if you're still working on this or need any help?

lieut-data:

PR #9691, opened 2022-01-23
MM-39510 : Quick reactions in the hover menu and emoji picker should be based on most frequently used emoji instead of recently used

darkLord19:
Fixes mattermost/mattermost-server#18603
Jira: https://mattermost.atlassian.net/browse/MM-39150
#### Release Note
```release-note
Changed recent emojis to be sorted by number of times used and store them on the server
```


esethna:
@darkLord19 does this also change the emoji picker to use the frequently used emoji as well?

Are you also able to make the same update to our mobile app in the long press menu and emoji picker?

actions/emoji_actions.js
```diff
@@ -79,16 +80,27 @@ export function addRecentEmoji(alias) {
...
 
         let updatedRecentEmojis;
-        if (recentEmojis.includes(name)) {
-            // If the emoji is already in the recent list, remove it and add it to the front
-            const recentEmojisFiltered = recentEmojis.filter((recentEmoji) => recentEmoji !== name);
-            updatedRecentEmojis = [...recentEmojisFiltered, name].slice(-MAXIMUM_RECENT_EMOJI);
+        const currentEmojiInRecentList = recentEmojis.find((recentEmoji) => recentEmoji.name === name);
+        if (currentEmojiInRecentList) {
+            // If the emoji is already in the recent list, remove it and add it to the front with updated usage count
+            const recentEmojisFiltered = recentEmojis.filter((recentEmoji) => recentEmoji.name !== name);
```
larkox, line 86: (0/5, may be premature optimization) If we already looked for the currentEmoji in the list, cannot we get the index, and just take that index out of the list with splice? That would make so we only traverse the recentEmojis list once.

actions/emoji_actions.js
```diff
@@ -79,16 +80,27 @@ export function addRecentEmoji(alias) {
...
+            const currentEmojiData = {
+                name,
+                usageCount: 1,
+            };
+            updatedRecentEmojis = [...recentEmojis, currentEmojiData].slice(-MAXIMUM_RECENT_EMOJI);
         }
 
         LocalStorageStore.setRecentEmojis(currentUserId, updatedRecentEmojis);
 
+        _dispatch(updateRecentEmojis(updatedRecentEmojis));
```
larkox, line 102: Do we need both?

actions/websocket_actions.jsx
```diff
@@ -1028,6 +1029,11 @@ export async function handleUserUpdatedEvent(msg) {
     if (user && user.props) {
         const customStatus = user.props.customStatus ? JSON.parse(user.props.customStatus) : undefined;
         dispatch(loadCustomEmojisIfNeeded([customStatus?.emoji]));
+
+        const recentEmojis = user.props.recentEmojis ? JSON.parse(user.props.recentEmojis) : undefined;
```
larkox, line 1033: Any chance `JSON.parse` may throw an exception in this context?

packages/mattermost-redux/src/actions/users.ts
```diff
@@ -1552,3 +1554,12 @@ export default {
     enableUserAccessToken,
     checkForModifiedUsers,
 };
+
+export function updateRecentEmojis(recentEmojis: RecentEmojiData[]): ActionFunc {
```
larkox, line 1558: If we are not updating anything on redux, why do we need this?

packages/mattermost-redux/src/types/emojis.ts
```diff
@@ -11,8 +11,7 @@ export type EmojiCategory = (
     | 'objects'
     | 'symbols'
     | 'flags'
-    | 'custom'
-);
+    | 'custom';
 export type CustomEmoji = {
```
larkox, line 15: I think we could use a line break here.

larkox (approved):
No blocking issues, just a few comments.

larkox:
Is this planned for mobile? What are the expectations there? CC @enahum @Rina-dsg

enahum:
@larkox not sure, I guess that is a question for UX/PM cc: @esethna 

on mobile the post options menu show a quick reaction list based on the 5 most recent reactions used, if there aren't 5, we have a list set of the more common ones.

Rina-dsg:
@esethna @enahum if we are planning to implement frequently used reactions for mobile, the solution may be to rank the emoji from the most frequently used to the least and show them accordingly. What do you think? if there aren't 5 frequently used reactions we can put the most common ones instead (the same logic as we used previously).
Screenshot 2022-02-01 at 19 37 28

enahum:
> @esethna @enahum if we are planning to implement frequently used reactions for mobile, the solution may be to rank the emoji from the most frequently used to the least and show them accordingly. What do you think? if there aren't 5 frequently used reactions we can put the most common ones instead (the same logic as we used previously). Screenshot 2022-02-01 at 19 37 28

Make a decision in terms of how it works and in which app does this goes: v1, v2 or both.. then file the appropriate tickets.

But I'm curious as of why it was not thought about before this ticket was created/worked on. Also curious why the feature team did not caught the fact that this is a feature being introduced that was not considered for mobile, so the real question here is Does this makes sense to do in mobile ?

esethna:
@enahum @Rina-dsg This was always considered for both mobile and webapp in the [ticket](https://mattermost.atlassian.net/browse/MM-39150), since the same logic applies to both.

image

I also asked @darkLord19 above if he's interested in making the mobile PR for this to fully complete the feature: https://github.com/mattermost/mattermost-webapp/pull/9691#issuecomment-1020464634

enahum:
And then coordinate with @avinashlng1080 as he is porting the existing feature to v2

esethna:
@enahum thanks good call, I let him know

darkLord19:
@esethna I can most probably work on mobile app feature once webapp pr is properly done. In the meanwhile if someone else is interested then they can take it up.

hmhealey:
Apologies for the delay reviewing this. I haven't been able to make the time for this since I want to make sure I spend enough time on it since it's potentially a sensitive piece of code. I'll be taking a look at it shortly

actions/emoji_actions.js
```diff
@@ -79,15 +78,26 @@ export function addRecentEmoji(alias) {
...
-            updatedRecentEmojis = [...recentEmojis, name].slice(-MAXIMUM_RECENT_EMOJI);
+            const currentEmojiData = {
+                name,
+                usageCount: 1,
+            };
+            updatedRecentEmojis = [...recentEmojis, currentEmojiData].slice(-MAXIMUM_RECENT_EMOJI);
         }
 
-        LocalStorageStore.setRecentEmojis(currentUserId, updatedRecentEmojis);
+        _dispatch(savePreferences(currentUserId, [{category: Constants.Preferences.RECENT_EMOJIS, name: currentUserId, user_id: currentUserId, value: JSON.stringify(updatedRecentEmojis)}]));
```
hmhealey, line 100: Could you rename `_dispatch` back to `dispatch` since it's no longer unused?

@M-ZubairAhmed I think you had changed this a few weeks back when we stopped using dispatch in this function. I think we should probably keep the existing name in that situation since we'll end up having to just change it back later

selectors/emojis.test.js
```diff
@@ -22,13 +20,11 @@ describe('Selectors.Emojis', () => {
         assert.deepEqual(Selectors.getRecentEmojis(testState), []);
 
         const recentEmojis = ['rage', 'nauseated_face', 'innocent', '+1', 'sob', 'grinning', 'mm'];
-        LocalStorageStore.setRecentEmojis(userId1, recentEmojis);
         setTimeout(() => {
```
hmhealey, line 23: These tests should be updated so that the state changes between calls to the selector. The `setTimeout`s also aren't needed any more since they were only to allow for us to write to storage

selectors/emojis.ts
```diff
@@ -23,23 +22,26 @@ export const getEmojiMap = createSelector(
...
-            return [];
-        }
+export const getShortcutReactToLastPostEmittedFrom = (state: GlobalState) =>
+    state.views.emoji.shortcutReactToLastPostEmittedFrom;
 
-        const recentEmojisArray: string[] = JSON.parse(recentEmojis);
-        return recentEmojisArray;
-    },
-);
+export function getRecentEmojis(state: GlobalState): RecentEmojiData[] {
```
hmhealey, line 28: This needs to go back to being wrapped in `createSelector` so that its results are memoized. Otherwise, we're returning a new array from JSON.parse each time this is called which will cause various any React components relying on this to re-render constantly

actions/emoji_actions.test.js
```diff
@@ -4,7 +4,7 @@
 import configureStore from 'store';
 import {getRecentEmojis, getEmojiMap} from 'selectors/emojis';
 import * as EmojiActions from 'actions/emoji_actions';
-import LocalStorageStore from 'stores/local_storage_store';
+import preferences from 'mattermost-redux/action_types/preferences';
```
hmhealey, line 7: Shouldn't this import look more like `import * as PreferenceActions from 'mattermost-redux/actions/preferences';`? You're spying on a set of constants currently

selectors/emojis.ts
```diff
@@ -56,6 +58,14 @@ export const getOneClickReactionEmojis = createSelector(
             return [];
         }
 
-        return recentEmojis.map((recentEmoji) => emojiMap.get(recentEmoji)).filter(Boolean).slice(-3).reverse();
+        const sortedRecentEmojis = recentEmojis.sort(
```
hmhealey, line 61: Do you think it would make any difference performance-wise if we stored a sorted array of emoji to begin with instead of sorting it here? I could go either way on that since I don't imagine it'll make a difference in the long run.

That said, there's a bug here that might show up, particularly if `recentEmojis` is memoized like we should be doing. `recentEmojis.sort(...)` actually mutates `recentEmojis`, so we should copy `recentEmojis` before sorting it

hmhealey (changes requested):

hmhealey:
Something else that's missing here is that all users will have their recent emojis cleared with this since there's no migration going on.

@esethna Have we thought at all about how we should migrate that? Since each user's recent emojis are stored on the device and they could have different sets of recent emojis depending on the device, I think the best thing we could do here is that the first time a user uses Mattermost with these changes applied (ie they don't have any recent emojis stored on the server), we just copy the ones from local storage into their preferences and leave them that way.

esethna:
> I think the best thing we could do here is that the first time a user uses Mattermost with these changes applied (ie they don't have any recent emojis stored on the server), we just copy the ones from local storage into their preferences and leave them that way.

@hmhealey that works for me!

avinashlng1080:
@enahum  thanks for the heads up.

@esethna  -  I went through the ticket and saw that it mentioned **mobile** and it was assigned to the community. 
The new elements in here are: 
1. synchronization with the server;  Will we hit an endpoint to store the user's  most frequent reactions and then call it on mobile?  AFAIK, on Gekidou, we are storing the recent reactions locally.
2. sorting by most frequently used and keeping track of it.
3. implementing it on Gekidou/v2

Given that I'm porting the feature as it is from v1, can you please create a ticket specially to address the above three points and then plan it in the proper release phase?

hmhealey:
@darkLord19 See my above comment for ideas on how to do a migration then.

@avinashlng1080 This can be added to that ticket, but since these are stored in Preferences, you should get the syncing for free already since Preferences are used for many other things as well (theme, various settings, flagged posts, DM visibility, etc). The thing changes that would need to happen on mobile would be:
1. Have it pull recent emojis from Preferences instead of from wherever it stores them currently.
2. Have it write recent emojis back to Preferences in the same way this PR does.
3. Potentially add a similar migration step to the one I commented on above.

darkLord19:
@hmhealey have updated PR with fixes please re-review.

darkLord19:
@hmhealey have updated PR with fixes please re-review. Also about migration, do you have some example in codebase for handling migrations in webapp. I was thinking we can add a check in getRecentEmojis to populate preferences from local store if preferences are empty.

selectors/emojis.ts
```diff
@@ -23,23 +22,42 @@ export const getEmojiMap = createSelector(
...
+export const getShortcutReactToLastPostEmittedFrom = (state: GlobalState) =>
+    state.views.emoji.shortcutReactToLastPostEmittedFrom;
 
 export const getRecentEmojis = createSelector(
     'getRecentEmojis',
-    (state: GlobalState) => LocalStorageStore.getRecentEmojis(getCurrentUserId(state)),
-    (recentEmojis) => {
+
+    // if getSimplifiedData is true then return array of emoji names instead of RecentEmojiData object
+    (state: GlobalState, getSimplifiedData = true) => {
```
hmhealey, line 32: This unfortunately won't work for a couple reasons.

First off, having this input selector return an object literal will mean that the final result function recalculates which would mean that a new array is returned each time this is called. Normally, you could solve that by just reverting this input selector and adding a separate one that just returns the argument to the result function, but since you could have two components calling this selector with a different value for `getSimplifiedData`, they'll end up overwriting each others memoized result and causing the same problem.

I think we're going to have to either split this into two selectors (maybe keep this and a `getRecentEmojiNames` to replace the existing one?) and/or update the other code that uses this to accept an array of emojis.

hmhealey (changes requested):

hmhealey:
For the migration, I don't think we have any good examples of doing that, but I don't think we can do it in `getRecentEmojis` since selectors are supposed to be pure, and even if we were to work around that, I don't know how we could write back to preferences there. We don't have anywhere super clean to put that sort of startup logic, but I guess we could put it in `components/root/root.jsx` with where it calls `setSystemEmojis` just to keep it somewhat near other emoji code. It's not great, but the whole startup logic of the app needs cleanup

hmhealey:
Hey @darkLord19. As usual, let us know if there's anything we can help with on this

M-ZubairAhmed:
@darkLord19 heads up quite a few things changed in `components/emoji_picker/*`. I dont know why it gives me permission error when i try to update the branch.

darkLord19:
@M-ZubairAhmed I have already reverse merged latest master into my branch

selectors/emojis.ts
```diff
@@ -50,12 +76,16 @@ export function isCustomEmojiEnabled(state: GlobalState) {
 export const getOneClickReactionEmojis = createSelector(
     'getOneClickReactionEmojis',
     getEmojiMap,
-    getRecentEmojis,
-    (emojiMap, recentEmojis) => {
+    getRecentEmojisNames,
+    (emojiMap, recentEmojis: RecentEmojiData[] | string[]) => {
```
hmhealey, line 80: Which of those types is now used for `recentEmojis`? I assume it's either one or the other at this point

hmhealey (changes requested):
That new code looks great! I've got a couple small things to pick at, but otherwise, this looks good to go

selectors/emojis.ts
darkLord19: nice catch. I missed to update types here. have fixed it

packages/mattermost-redux/src/actions/emojis.ts
```diff
@@ -250,3 +262,29 @@ export function autocompleteCustomEmojis(name: string): ActionFunc {
         return {data};
     };
 }
+
+export function migrateRecentEmojis(): ActionFunc {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const state = getState() as GlobalState;
+        const currentUserId = getCurrentUserId(state);
+        let recentEmojisFromPreference: RecentEmojiData[] = [];
+        try {
```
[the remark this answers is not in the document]
darkLord19: I suppose you are right. I have removed it.

hmhealey (approved):

hmhealey:
@darkLord19 There's one more lint error, but this looks good otherwise.

@M-ZubairAhmed Did you still want to give this a look before we hand it off to QA? It's got 2 reviews, so feel free to give it a super quick one

hmhealey:
@darkLord19 I was testing the migration locally since it's going to be awkward to test otherwise, and it seems like it's not being triggered. Could you look into that? I see the action's not being called, so either there's some error occurring or it's not being dispatched or something.

darkLord19:
@hmhealey I am trying to get current user id like this
```
        const state = getState() as GlobalState;
```
but it is returning '' empty string and state is also not having `entities.users` key not sure why.

components/root/root.jsx
```diff
@@ -140,6 +140,7 @@ export default class Root extends React.PureComponent {
         Client4.setAuthHeader = false;
 
         setSystemEmojis(EmojiIndicesByAlias);
+        store.dispatch(migrateRecentEmojis());
```
hmhealey, line 143: I don't actually know why we're using the global store directly in this file, but this should be passed in from this file's index.js in the `mapDispatchToProps` to connect it to the store. 

I wonder if that might fix the issue you were having with the current user ID being empty?

darkLord19: I have moved it to mapDispatchToProps but still it's coming empty

hmhealey:
Oh, I see what's happening. By doing that in the constructor for the Root component, it's happening before we've loaded the current user. The correct place to call `migrateRecentEmojis` then is in `onConfigLoaded`. It probably makes sense to do it there right before `loadRecentlyUsedCustomEmojis`.

That one should also be dispatched properly without accessing the global store, but we don't need to change that now.

darkLord19:
Ohh that makes sense. Thanks! :)

hmhealey:
Awesome, that fixed it. The migration looks like it works locally as well now.

If you could take a look at it, there's a test failing now for the Root component, but I assume that's just caused by a something simple like a missing prop. Unfortunately, the test that's failing is async so CI doesn't actually say what the issue is.

jgilliam17:
Thanks @darkLord19 
Here's the [E2E report](https://mattermost-cypress-report.s3.amazonaws.com/171721-3e5a421-onprem-ent-webapp-pr-9691-mattermost/mm-ee-test3e5a421/mochawesome.html), has few failures related to changes in this PR
Screen Shot 2022-03-17 at 4 25 15 PM
Screen Shot 2022-03-17 at 4 25 43 PM

jgilliam17:
Beside cypress failures, everything else works as expected.
- Verified the quick reactions options in the post hover menu are based on most frequently used reactions, not recently used.

darkLord19:
@jgilliam17 that's great. I'll take a look at e2e failures and update pr

darkLord19:
@jgilliam17 can you share e2e test results as I am not able to access the result url and in my local machine lots of unrelated tests are failing due to timeouts and stuff

darkLord19:
@jgilliam17 also quite few tests you posted in earlier ss are also failing on master so don't think related to this change. can you get list of failing tests on master and then we can find out which one are extra failures in this pr

jgilliam17:
Hello @darkLord19 
Here's the link to [E2E report for the PR ](https://automation-dashboard.vercel.app/cycles/f546d6b5-a346-4b3a-9b56-23e39cc64488) and one for [master](https://automation-dashboard.vercel.app/cycles/02c92043-53b8-4d53-a464-1a861652048c) for comparison.
Let me know if you need anything else. 🙂

esethna:
@darkLord19 just circling back on this. Please let us know if you're still working on it?

darkLord19:
@esethna hey yes I am still working on this!

esethna:
@darkLord19 great! Let us know if you need any help or if you're ready for QA review from @jgilliam17

jgilliam17:
Thanks @darkLord19 
Tested, looks good to merge - pending e2e report 
- Verified quick reactions options are based on most frequently used and not most recently used.

jgilliam17:
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/202253-75e2802-onprem-ent-webapp-pr-9691-mattermost/mm-ee-test75e2802/mochawesome.html) shows one failure related to this PR MM-T1480
Please take a look, and also please update, I wasn't able to do it earlier. Thanks 🙂

jgilliam17:
Thanks @darkLord19 
Looks good to merge.
- [E2Es report ](https://automation-dashboard.vercel.app/cycles/2e1406a8-095e-4856-a5f3-bb8756ea1478) - no PR related failures.

jgilliam17 (approved):

hmhealey:
Thanks again for the PR, @darkLord19! Looking forward to having this in

PR #9694, opened 2022-01-24
MM-39622: jittery typing on thread explode

koox00:
#### Summary

Continuous posting of replies to a thread makes text input jittery.
It seems that when a thread is followed, or when the thread is open in
the RHS the problem worsens.

This PR is an attempt to fix that.

- when we receive new posts, ask for the post's thread only once
 
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39622

#### Release Note

```release-note
NONE
```


hmhealey (commented):
0/5, but we could slightly simplify this logic by breaking the loop in two like
```javascript
const rootsSet = new Set();

posts.forEach((post) => {
   if (/*we don't have root post*/) {
       rootsSet.add(post.root_id);
   }
});

rootsSet.forEach((rootId) => {
    promises.push(/*get the thread*/);
});
```

Otherwise, that sounds good to me

hmhealey (approved):

jgilliam17 (approved):

michelengelen (approved):
LGTM

PR #9696, opened 2022-01-25
[MM-40920] - Enterprise Trial Activation Tour Point

AGMETEOR:
#### Summary
In the tutorial points flow, this PR adds a step to show tell only admin users about the trial license

#### Ticket Link
[MM-40920](https://mattermost.atlassian.net/browse/MM-40920)

#### Screenshots

https://user-images.githubusercontent.com/28563179/151004506-12f08af7-e07e-4345-bb4b-9a5ec4c6d2b0.mov

#### Release Note

```release-note
NONE

```


utils/tutorial_steps.ts
```diff
@@ -0,0 +1,43 @@
...
+    POST_POPOVER: 0,
+    CHANNEL_POPOVER: 1,
+    ADD_CHANNEL_POPOVER: 2,
+    MENU_POPOVER: 3,
+    PRODUCT_SWITCHER: 4,
+    SETTINGS: 5,
+    FINISHED: 999,
+};
+
+export const getTutorialSteps = createSelector(
```
neallred, line 34: I like the intent of what this is doing of declaratively controlling which tips the user sees, but I don't think we can safely do it until we migrate the data so that each tip is its own independent piece of data in the database.

So for now, I'd 3/5 recommend having a single set of tutorial steps instead of splitting what the numbers mean per user/role. To avoid non-admins seeing the start trial tip, we can instead control seeing or skipping the tip in whichever component is most relevant to control it from. We have a pattern for doing this in the product switcher tip: https://github.com/mattermost/mattermost-webapp/blob/master/components/global_header/left_controls/product_menu/product_menu_tip/product_menu_tip.tsx#L60-L90

Some reasons to avoid splitting the meaning of the step preference number until we migrate the data:
* User role is mutable: non-admins can become admins, admins can stop being admins. When they do, their tip preference data will be interpreted incorrectly
* When we add on new tips, we need to maintain these two lists
* If other tips become conditional on other criteria than admin status, we need even more lists
* If we go the split list route today, it make migrating more complicated in the future because the migration also needs to take admin status into account, and any users whose admin role changed in the meantime will be migrated incorrectly.

The long term solution (probably outside the scope of this ticket and PR) is to migrate this data out of a single user preference into something that supports each tip's viewed status as its own prop, and we control view and ordering from the UI (like is being done here). We could use a boolean preference per tip, or maybe we could store tips in the sql `users.props` jsonb column.

AGMETEOR: @neallred agreed. Thanks for the detailed explanation. I am going to make edits

AGMETEOR:
Hey @pablovelezvidal and @neallred gentle ping on this

components/tutorial/tutorial_tip/tutorial_tip.tsx
```diff
@@ -256,8 +281,24 @@ export default class TutorialTip extends React.PureComponent<Props, State> {
...
-    private getLastStep(tutorialSteps: Record<string, number>) {
-        return Object.values(tutorialSteps).reduce((maxStep, candidateMaxStep) => {
+    private getLastStep(tutorialSteps: Record<string, number>, category: string) {
+        let steps = tutorialSteps;
+
+        // temporary fix for filtering tutorial_step category based on role
+        // however this could work if TutorialSteps are converted to an enum and tips for only admins added last to the enum
+        if (category === Preferences.TUTORIAL_STEP) {
+            steps = Object.keys(steps).filter((key) => {
+                if (!this.props.isAdmin && key === 'START_TRIAL') {
```
pablovelezvidal, line 291: maybe we can here just return this condition evaluation `this.props.isAdmin && key === 'START_TRIAL'`

components/tutorial/tutorial_tip/tutorial_tip.tsx
```diff
@@ -256,8 +281,24 @@ export default class TutorialTip extends React.PureComponent<Props, State> {
...
     }
 
-    private getLastStep(tutorialSteps: Record<string, number>) {
-        return Object.values(tutorialSteps).reduce((maxStep, candidateMaxStep) => {
+    private getLastStep(tutorialSteps: Record<string, number>, category: string) {
+        let steps = tutorialSteps;
+
+        // temporary fix for filtering tutorial_step category based on role
+        // however this could work if TutorialSteps are converted to an enum and tips for only admins added last to the enum
+        if (category === Preferences.TUTORIAL_STEP) {
```
pablovelezvidal, line 289: thinking this better, if the idea is just remove the preference `START_TRIAL` property and value from the steps object, I think that  just check for the condition if the user is admin and then delete the property from the object with `delete Obj.property`, WDYT @AGMETEOR ?

components/tutorial/tutorial_tip/tutorial_tip.tsx
```diff
@@ -286,14 +327,23 @@ export default class TutorialTip extends React.PureComponent<Props, State> {
...
-            buttonText = (
-                <FormattedMessage
-                    id={'tutorial_tip.finish_tour'}
-                    defaultMessage={'Finish tour'}
-                />
-            );
+            if (this.props.customLastStepButtonText) {
+                buttonText = (
+                    <FormattedMessage
+                        id={this.props.customLastStepButtonText.id}
```
pablovelezvidal, line 335: will this work well with the `make i18n-extract` action. I mean, if someone runs that command later won't it delete whatever value you have set to en.json?

components/widgets/menu/menu_items/menu_start_trial.tsx
```diff
@@ -56,45 +59,83 @@ const MenuStartTrial = (props: Props): JSX.Element | null => {
...
 
+    const title = (
+        <FormattedMessage
+            id='start_trial.tutorialTip.title'
+            defaultMessage='Try our premium features for free'
+        />
+    );
+
+    const screen = (
+        <>
```
pablovelezvidal, line 87: is this fragment needed?

components/tutorial/tutorial_tip/tutorial_tip.tsx
AGMETEOR: That's much better. Thanks :)

e2e/cypress/integration/onboarding/finish_tutorial_spec.js
```diff
@@ -13,6 +13,82 @@
 import * as TIMEOUTS from '../../fixtures/timeouts';
 import {generateRandomUser} from '../../support/api/user';
 
+describe('Onboarding ADMIN', () => {
```
pablovelezvidal, line 16: Awesome tests ❤️

pablovelezvidal (approved):
Excellent work @AGMETEOR . Just a couple of questions. Thanks 🚀

components/tutorial/tutorial_tip/tutorial_tip.tsx
AGMETEOR: Removed this code

components/tutorial/tutorial_tip/tutorial_tip.tsx
AGMETEOR: @pablovelezvidal 3/5 it wouldn't affect. Have you seen this happen before?

pablovelezvidal: it happened to me, but don't remember if it is the same exact situation. You can check it if you run the command and verify if it deletes the entry you added in the  en.json file

components/widgets/menu/menu_items/menu_start_trial.tsx
AGMETEOR: Removed it

components/tutorial/tutorial_tip/tutorial_tip.tsx
AGMETEOR: I already ran it

components/tutorial/tutorial_tip/tutorial_tip.tsx
```diff
@@ -226,7 +248,10 @@ export default class TutorialTip extends React.PureComponent<Props, State> {
             const tag = this.props.telemetryTag + '_next';
             trackEvent('tutorial', tag);
         }
-        if (this.getLastStep(TutorialSteps[this.props.tutorialCategory || Preferences.TUTORIAL_STEP]) === this.props.currentStep) {
+
+        const category = this.props.tutorialCategory || Preferences.TUTORIAL_STEP;
+
+        if (this.getLastStep(TutorialSteps[category], category) === this.props.currentStep) {
```
neallred, line 254: This works for non-admins right now because the last step happens to be an admin step, but if we add a non admin step after the admin START_TRIAL step, it will be broken for non-admins because they'll be pushed into the admin step, not see the tip for it, and not be able to progress to see the last step.

I'd suggest that we include logic that checks if the user is an admin or not, and if the upcoming step is only for admins and the current user is not an admin, we push them to whatever the next non-admin step is.

AGMETEOR: @neallred that's true. I documented that this is just a temporary fix on that check in the `getLastStep`

AGMETEOR: Maybe it's time we do that rework of the tutorial tips

neallred: The use case growth spike is using the tutorial tip, too, so regardless of who implements the change, it might cause too much re-coding to refactor and migrate this now *and* finish the growth spike on time.

stevemudie (commented):
@AGMETEOR This works but there's something odd going on with the positioning of the tutorial point.

In the video example only Channels is shown and the point is right where I would expect.
Channels only

However, in my testing of it I see Channel, Boards & Playbooks and the point isn't where it should be.
Channel, Boards   Playbooks

components/tutorial/tutorial_tip/tutorial_tip.tsx
AGMETEOR: @neallred I added the check

components/tutorial/tutorial_tip/tutorial_tip.tsx
```diff
@@ -159,15 +173,39 @@ export default class TutorialTip extends React.PureComponent<Props, State> {
...
+                    stepValue += 1;
+                } else {
+                    break;
+                }
+            }
         } else if (nextStep === false) {
             stepValue -= 1;
+
+            // if the previous tip step/steps are for only admins, skip them for non admins
+            for (const tipName of AdminOnBoardingTutorialStep) {
```
neallred, line 201: Nice solution to skipping admin steps!! For this backing up steps code path, I think we need to iterate in reverse order, which we could do with

```ts
for (const tipName of [...AdminOnBoardingTutorialStep].reverse()) {
...
```
My code example creates a new array because reverse mutates the original array, which we do not want to do.

AGMETEOR: @neallred why do we need to reverse the array?

neallred: If there are two steps in a row that are admin steps and you are backing up and iterating from start to finish, when the first admin step is reached, in the for loop iteration, the user will still be on the second admin step. Then, once the loop reaches the second admin step, the step will be backed up once. But even though the user is still on an admin step, iteration over the step the user is on is already finished, so the user is incorrectly placed on an admin step.

That's kind of a lot to step through, so I made a minimal code repro where you can see it happening.

```js
const TutorialSteps = {
    ADD_FIRST_CHANNEL: -1,
    POST_POPOVER: 0,
    CHANNEL_POPOVER: 1,
    ADD_CHANNEL_POPOVER: 2,
    MENU_POPOVER: 3,
    PRODUCT_SWITCHER: 4,
    SETTINGS: 5,
    START_TRIAL: 6,
    NEW_ADMIN_STEP: 7,
    OTHER_NEW_STEP: 8,
    FINISHED: 999,
};

const ADMIN_STEPS = ["START_TRIAL", "NEW_ADMIN_STEP"]

function getKeyByValue(obj, value) {
    return Object.keys(obj).find((key) => obj[key] === value);
}

function backupStepsNoReverse(currentStep, isAdmin) {
    let newStep = currentStep - 1;
    if (isAdmin) {
        return newStep;
    }
    for (const adminStep of ADMIN_STEPS) {
        if (getKeyByValue(TutorialSteps, newStep) === adminStep) {
            newStep -= 1;
        }
    }
    return newStep
}

function backupSteps(currentStep, isAdmin) {
    let newStep = currentStep - 1;
    if (isAdmin) {
        return newStep;
    }
    for (const adminStep of [...ADMIN_STEPS].reverse()) {
        if (getKeyByValue(TutorialSteps, newStep) === adminStep) {
            newStep -= 1;
        }
    }
    return newStep
}

console.log(backupStepsNoReverse(TutorialSteps.OTHER_NEW_STEP, false));
console.log(backupSteps(TutorialSteps.OTHER_NEW_STEP, false));
```

If you run that as a file in node, you get:
```
node tmp.js
6
5
```

AGMETEOR: Hmmm... no problem changing it but currently I have a test like so, if we are currently on step 8, but step 7 and 6 are admin steps, then we expect to land on step 5

tutorial_tip.test.tsx
```js
    test('handleSavePreferences skip previous admin steps for non admins', () => {
        const props = {
            currentUserId,
            title: <>{'title'}</>,
            step: 8,
            currentStep: 8,
            autoTour: false,
            isAdmin: false,
            actions: {
                closeRhsMenu: jest.fn(),
                savePreferences: jest.fn(),
                setFirstChannelName: jest.fn(),
                setProductMenuSwitcherOpen: jest.fn(),
            },
            screen: <></>,
            placement: 'right',

        };
        const wrapper = shallow<TutorialTip>(
            <TutorialTip {...props}/>,
        );

        wrapper.instance().handleSavePreferences(false, false);

        // current tip is 7, but tip 6, START_TRIAL is only for admins and is skipped to tip 5
        expect(props.actions.savePreferences).toHaveBeenCalledWith('currentUserId', [{category: 'tutorial_step', name: 'currentUserId', user_id: 'currentUserId', value: '5'}]);
    });
```

and with constants like so
```js
export const TutorialSteps = {
    ADD_FIRST_CHANNEL: -1,
    POST_POPOVER: 0,
    CHANNEL_POPOVER: 1,
    ADD_CHANNEL_POPOVER: 2,
    MENU_POPOVER: 3,
    PRODUCT_SWITCHER: 4,
    SETTINGS: 5,
    START_TRIAL: 6,
    NEW_ADMIN_STEP: 7,
    FINISHED: 999,
};

export const AdminTutorialSteps = ['NEW_ADMIN_STEP', 'START_TRIAL'];
```

AGMETEOR: nvm, I actually see what you mean

AGMETEOR: I should probably document that the order the tips are added to `AdminTutorialSteps` matters

neallred (approved):
LGTM, thanks @AGMETEOR !!

AGMETEOR:
> LGTM, thanks @AGMETEOR !!

Thanks @neallred for all the feedback on this one

stevemudie (approved):
Looks good @AGMETEOR!

neallred:
@AGMETEOR looks like there were some tests that still needing fixing in PR at the time it was merged; could you fix them and open a PR?

https://app.circleci.com/pipelines/github/mattermost/mattermost-webapp?branch=master&filter=all
https://app.circleci.com/pipelines/github/mattermost/mattermost-webapp/49179/workflows/3e500a0b-1bbb-4583-a858-6c7f0dda7b32

```
components/tutorial/tutorial_tip/tutorial_tip handleSavePreferences next skip admin steps for non admins
components/tutorial/tutorial_tip/tutorial_tip handleSavePreferences skip previous admin steps for non admins
```

AGMETEOR:
@neallred yes, to fix we need to pass prop ‘singleTip=true’ in those test cases

marianunez:
@AGMETEOR unfortunately this PR needs to be reverted given that the new onboarding will replace the old tutorial tip flows. There is a direct conflict and this prompt to start trial would need to be moved somewhere else within the new onboarding flow/checklist. 

Thank you for your hard work on this and apologies for this miscommunication.

AGMETEOR:
@marianunez no worries.

PR #9701, opened 2022-01-25
[MM-36190] Permalink Nested Previews

mylonsuren:
#### Summary

Adds nested previews for URLs, files, and images for permalinks. 

#### Ticket Link

Fixes [MM-36190](https://mattermost.atlassian.net/browse/MM-36190)

#### Related Pull Requests

- Has server changes https://github.com/mattermost/mattermost-server/pull/19413

#### Screenshots

URL:
image

Image:
image

Files:
image

#### Release Note

```release-note
Adds nested previews for permalinks
```


components/single_image_view/single_image_view.tsx
```diff
@@ -211,16 +220,18 @@ export default class SingleImageView extends React.PureComponent<Props, State> {
...
-                                onClick={this.handleImageClick}
-                                className={minPreviewClass}
-                                src={previewURL}
-                                dimensions={this.state.dimensions}
-                                fileInfo={this.props.fileInfo}
-                                onImageLoaded={this.imageLoaded}
-                                showLoader={this.props.isEmbedVisible}
-                                handleSmallImageContainer={true}
-                            />
+                            <div className={permalinkClass}>
```
BenCookie95, line 223: lets use `classnames` on this

packages/mattermost-redux/src/reducers/entities/files.ts
```diff
@@ -131,6 +150,25 @@ export function fileIdsByPostId(state: Record<string, string[]> = {}, action: Ge
     }
 }
 
+function storeAllFilesIdsForPost(state: Record<string, string[]>, post: Post) {
```
BenCookie95, line 153: This is very similar to `storeAllFilesForPost`, I wonder if we can just use one function?

packages/mattermost-redux/src/utils/post_utils.ts
```diff
@@ -217,6 +229,11 @@ export function shouldUpdatePost(receivedPost: Post, storedPost?: Post): boolean
             return true;
         }
 
+        if (!storedPost.metadata && receivedPost.metadata) {
```
BenCookie95, line 232: Would there be a situation where metadata would exist in `storedPost` but receivedPost would have updated or different metadata?

packages/mattermost-redux/src/reducers/entities/posts.ts
```diff
@@ -271,45 +271,59 @@ export function handlePosts(state: RelationOneToOne<Post, Post> = {}, action: Ge
     }
 }
 
-function handlePostReceived(nextState: any, post: Post) {
-    if (!shouldUpdatePost(post, nextState[post.id])) {
-        return nextState;
+function handlePostReceived(nextState: any, post: Post, nestedPermalinkLevel?: number) {
+    let currentState = nextState;
```
BenCookie95, line 275: Why not just call the function parameter `currentState`

BenCookie95 (commented):
Nice work, just a few comments!

packages/mattermost-redux/src/reducers/entities/files.ts
mylonsuren: Yeah both these functions are the same, only difference is that `state` is of a different type in each. I looked at doing a type check and overloading `storeAllFilesForPost` to handle both but I couldn't find an efficient or readable solution that would work in this case. Main issue stems from `state` being of type `Record` which makes it difficult to check the type of its objects at run time.

packages/mattermost-redux/src/utils/post_utils.ts
mylonsuren: Yes this is a possibility, but if `receivedPost` has different or updated metadata, then it's `update_at` field should be different as well and `shouldUpdate` will return true. We don't need to consider the case where the post is a permalink because each post is processed individually, so both the original post and permalink will be covered.

packages/mattermost-redux/src/reducers/entities/posts.ts
mylonsuren: The linter doesn't like when we reassign function parameters (occurs a few lines below this), so we need to declare it within the function.

components/post_view/post_attachment_opengraph/post_attachment_opengraph.tsx
```diff
@@ -195,24 +200,24 @@ export default class PostAttachmentOpenGraph extends React.PureComponent<Props>
         if (data.description || imageUrl) {
             body = (
                 <div className={'attachment__body attachment__body--opengraph'}>
-                    <div>
+                    <div className={this.props.isInPermalink ? 'permalink__body--opengraph' : ''}>
```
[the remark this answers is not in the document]
mylonsuren: Done 👍

components/post_view/post_attachment_opengraph/post_attachment_opengraph.tsx
```diff
@@ -195,24 +200,24 @@ export default class PostAttachmentOpenGraph extends React.PureComponent<Props>
...
                     </div>
-                    {imageUrl && hasLargeImage && this.renderLargeImage(imageUrl, imageMetadata)}
+                    {!this.props.isInPermalink && imageUrl && hasLargeImage && this.renderLargeImage(imageUrl, imageMetadata)}
                 </div>
             );
         }
 
         return (
-            <div className='attachment attachment--opengraph'>
+            <div className={this.props.isInPermalink ? 'attachment attachment--opengraph permalink' : 'attachment attachment--opengraph'}>
```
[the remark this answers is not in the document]
mylonsuren: Thanks, I've reverted this back to its original classname as the css that I had added for this is no longer required.

components/single_image_view/single_image_view.tsx
mylonsuren: Done 👍

esethna:
Awesome, thanks @mylonsuren! cc// @michaelgamble for UX review

michaelgamble (changes requested):
A few minor things to fix:
- the vertical space between content and "Only visible to users in ~channel-name >> this should only be 8px
- The previews sub link preview is not the right visual style >> remove the blue border on the left of it, make sure its the same rounded corners as link previews with the same shadow effects before and after hover
- When expanding and collapsing "show more" and "show less" the file attachment spacing / alignments seem to have a visual glitch of slight content shifting

BenCookie95 (approved):
Code looks good to me!

mylonsuren:
> A few minor things to fix:
> 
> * the vertical space between content and "Only visible to users in ~channel-name >> this should only be 8px
> * The previews sub link preview is not the right visual style >> remove the blue border on the left of it, make sure its the same rounded corners as link previews with the same shadow effects before and after hover
> * When expanding and collapsing "show more" and "show less" the file attachment spacing / alignments seem to have a visual glitch of slight content shifting

@michaelgamble Thanks for the feedback!

* Vertical spacing between content and the preview footer has been fixed.
* Link preview styling updated to match the design in Figma. 
  - For the shadow effects, are you saying to add a shadow effect for the URL preview itself? Or should the URL preview remain the same when hovered as it is currently.
* I noticed that slight content shifting yesterday as well after you mentioned it, but it seems to have gone away after updating Chrome. I checked Safari and Firefox as well and I don't see that content shifting. Let me know if you can still reproduce it and I'll continue to look into it.

michaelgamble:
@mylonsuren so similar to how the cards for attached files have thier own hover state
image

I would treat the whole sub link preview card in a similar manner

mkraft (approved):

esethna:
@michaelgamble can you give this a rereview when you can?

michaelgamble:
> 

@mylonsuren So talked to @esethna and we agreed that the best plan for the sublink styling is to keep it how it was originally > inheriting the original global link preview style and for us to break that out as a seperate body of work that will live in its own Jira ticket and PR.

So the plan forward for this PR should be to revert the changes to the sublink styling and then we should be good to go.

I did however notice one last visual change:
 - the spacing between the avatar and the username should be 8px
 - the spacing between the username and the date should be 6px

michaelgamble:
So im not sure if this is some sort of bug in the feature, but i notice when i load the page the first time the preview content shows as blank until i go back to the original channel to see the original posts:
image

mylonsuren:
> So im not sure if this is some sort of bug in the feature, but i notice when i load the page the first time the preview content shows as blank until i go back to the original channel to see the original posts: image

Yeah this looks like a bug, I'll look into this right now and get a fix in for it.

mylonsuren:
@michaelgamble I couldn't reproduce this issue locally so I'm going to use this PR to test some changes. There may be some additional commits while I test some fixes, but I'll let you know when it's ready for another round of review.

michaelgamble (approved):
LGTM!

esethna (approved):

esethna:
Looks great, thanks @mylonsuren! Just needs QA from @furqanmlk

mylonsuren:
@furqanmlk Just an FYI when testing, I'm not sure the test server created on this PR includes the changes from server, so I'd recommend using the test server created on that PR https://github.com/mattermost/mattermost-server/pull/19413.

furqanmlk (approved):
Working as expected

**URL**
image

**IMAGE**
image

**FILES**
image

esethna:
@mylonsuren looks like this can be merged :)

PR #9705, opened 2022-01-26
[MM-22841] Add copy URL, download source image icon buttons, to image thumbnail

plant99:
#### Summary

Adds the following functionality:

`On hover over an image preview, icon buttons appear on the top right of a thumbnail mage so the user doesn't have to open preview to copy link to the image or download.`

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19369
Jira ticket: https://mattermost.atlassian.net/browse/MM-22841
#### Related Pull Requests

NA

#### Screenshots

*before*

no_buttons_old

*after*

 after_copy  

new_copied

#### Release Note

```release-note

Added copy URL, and download buttons over image thumbnail for quicker interactions.
```


hmhealey:
Hi @plant99. Thanks for the PR! Apologies for the notification spam while we get everything set up on our end.

Since this is marked as a Draft, I assume you're still working on it. Feel free to let us know if you have any questions about anything around it

plant99:
Hey @hmhealey no worries at all! Yes, since this is my first PR to this org, it's taking some time for me to get the hang of CI checks and tests. Once I get them all to pass, I'll tag it ready for review, and mention Eric and @ AshishDhama for review.

plant99:
Hi @AshishDhama this PR is ready for review.

michelengelen (changes requested):

michelengelen:
@plant99 thanks for your contribution ... I have some suggestions and changes to be done before we can move on.

matthewbirtch:
@plant99 it looks like some checks failed that are blocking the cloud test server from being created.

Also, I added a case for smaller images in the [Figma design file](https://www.figma.com/file/xqAJ4U380TJhcQnZP6Q5zf/MM-22841-Image-Hover-Actions?node-id=912%3A26179).

plant99:
@matthewbirtch sure I'll modify this according to the figma file for small images.

Also regarding the failing checks, it's generally taking a long time to do the `Setup a remote docker engine process`. For other PRs it's generally in seconds, also in this one, there [are cases](https://app.circleci.com/pipelines/github/mattermost/mattermost-webapp/48895/workflows/a7577699-2c54-4a5d-b3e6-7d4e6fb8c904/jobs/340099) where it succeeded although took a longer time than usual. 

@michelengelen thanks for the review, I'll make sure to fix these issues when I open this PR for review again, after accommodating changes for small images.

esethna:
Thanks for the PR @plant99! Let us know when it's ready for another round of reviews

sass/components/_files.scss
```diff
@@ -265,6 +265,67 @@
...
+                            color: var(--denim-status-online);
+                        }
+                    }
+
+                    // handle 'copy successful' styles
+                    button.size-aware-image__copy_link--recently_copied {
+                        background: rgba(var(--denim-status-online-rgb), 0.16);
+
+                        &:hover {
+                            background: rgba(var(--denim-status-online-rgb), 0.16);
```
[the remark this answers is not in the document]
plant99: I think https://github.com/mattermost/mattermost-webapp/pull/9705/files#diff-8b31960d5e28fadc1d6dc0764c1fb8e8340902e0347883dacaa0770de5f0295fR294 takes precedence so without adding this, even if I hover over a clicked `Copy` button, it shows me `--center-channel-color-rgb`

plant99:
Hi, @matthewbirtch I don't think the design files consider the `collapse` button, the left offset of 8px overlaps with the collapse button. Could you please update the file? For now, I'm working with a value that doesn't make the two buttons overlap. It'd be a simple change later.

intersecting_with_collapse

matthewbirtch:
> Hi, @matthewbirtch I don't think the design files consider the `collapse` button, the left offset of 8px overlaps with the collapse button. Could you please update the file? For now, I'm working with a value that doesn't make the two buttons overlap. It'd be a simple change later.

Ahh, good catch @plant99. I forgot about the collapse icon. I've updated the Figma file to allow for this now.

also, don't forget the border and the shadow on the container

<100px wide
image

>100px wide
image

plant99:
Understood, thanks @matthewbirtch , I've made the necessary changes.

One last thing that remains is the definition of small image. In the existing code, `small-image__container` is only rendered when either dimension is less than https://github.com/mattermost/mattermost-webapp/blob/master/components/size_aware_image.jsx#L10

In the figma file, 100px is mentioned as the threshold for design change.

Which one should I follow for an image of dimension say 52x52 px? (Clearly won't render image inside small-image__container, if I've understood this correctly)

matthewbirtch:
> In the figma file, 100px is mentioned as the threshold for design change.

Ya, I recommended his change because moving the hover actions inside the image container only works above 100px.

sass/components/_files.scss
```diff
@@ -265,6 +265,83 @@
                 min-height: 40px;
                 text-align: left;
             }
+
+            // styles for utility buttons i.e download, and copy source link in image preview
+            .image-preview-utility-buttons-container {
+                // greater z-index than img
+                position: absolute;
+                z-index: 3;
+                right: 2px;
```
matthewbirtch, line 274: right: 8px;
top: 8px;

sass/components/_files.scss
```diff
@@ -265,6 +265,83 @@
...
+                display: none;
+                padding: 2px;
```
matthewbirtch, line 276: padding: 4px;

sass/components/_files.scss
```diff
@@ -265,6 +265,83 @@
...
+                border: 1px solid rgba(v(center-channel-color-rgb), 0.16);
+                margin-top: 4px;
```
matthewbirtch, line 278: margin-top: 0px

sass/components/_files.scss
```diff
@@ -265,6 +265,83 @@
...
+                margin-right: 2px;
```
matthewbirtch, line 279: margin-right: 0px

sass/components/_files.scss
```diff
@@ -265,6 +265,83 @@
...
+                border-radius: 4px;
+                box-shadow: $elevation-1;
+
+                button.size-aware-image__copy_link,
+                a.size-aware-image__download {
+                    display: inline-block;
+                    width: 28px;
+                    height: 28px;
+                    padding: 2px;
+                    margin: 2px;
```
matthewbirtch, line 290: margin-right: 0px

matthewbirtch (changes requested):
Thanks @plant99 it's looking great. A few comments (along with some CSS changes recommended/requested)

In addition to the CSS comments, can we reduce the delay for the checkmark icon to return back to the link icon? I think 1000ms or 1500ms should be long enough.

twig-aidash:
Thanks @matthewbirtch for the review, I'll ask for a review again when I've fixed these issues.

matthewbirtch:
> @matthewbirtch Do we need a collapse icon I mean you have added one which I think can replace the icon (marked in red box) as right now we don't image name?

The latest doesn't show the image name in the expanded state if I remember correctly

matthewbirtch:
One addition note @plant99, can we show these controls when hovering over the image only rather than hovering over the entire row?

plant99:
Hi @matthewbirtch, the recent commits,
 - move buttons inside the image only after width = 100 px
 - address the css changes you commented.
 - show buttons on image hover, than row hover
 
Please review again?

mkraft:
@plant99 Just a note on process which you may already know: once you believe you've addressed all of the comments of a given reviewer please click the "Re-request review" icon beside their name so that the PR re-appears in their _Review requests_ list. Thanks.

sass/components/_files.scss
```diff
@@ -265,6 +265,86 @@
...
+                border-radius: 4px;
+                box-shadow: $elevation-1;
+
+                button.size-aware-image__copy_link,
+                a.size-aware-image__download {
+                    display: inline-block;
+                    width: 28px;
+                    height: 28px;
+                    padding: 2px;
+                    margin: 2px 0 2px 2px;
```
matthewbirtch, line 289: remove this margin

matthewbirtch:
Looks great @plant99 - just one small change noted in the CSS (I think I missed that in the last round). Should be good after that from a UI/UX standpoint

sass/components/_files.scss
plant99: Fixed this.

hmhealey:
@matthewbirtch I have a couple questions about the design for this which unfortunately I think might be tough to fix in this PR:
1. Is the copied link supposed to be a public link that's viewable by users outside Mattermost? If so, that's off-by-default, and we'll have to add a state for when that button isn't visible.
2. If that is the case, we also don't have the public link available on the post list when this is shown, so we might need to make some changes to the API on the server which we may have to run past the security team. Public links for files are only generated on demand, so we wouldn't have that available without asking the server for it.
    Ideally, I think we could get the public link automatically as part of the data sent from the server when loading information about files, but there's a chance that that's a security issue since we currently track when people access the public link for files in the Auditing system. Alternatively, we could fetch the public link when the user clicks on the "copy link" button, but there may be a delay there as we request it from the server, so that could be a UX concern

components/size_aware_image.jsx
```diff
@@ -170,6 +182,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+                placement='top'
+                overlay={copyLinkTooltip}
+                rootClose={true}
+            >
+                <button
+                    key='copy-link'
+                    className={classNames('style--none', 'size-aware-image__copy_link', {
+                        'size-aware-image__copy_link--recently_copied': this.state.linkCopiedRecently,
+                    })}
+                    aria-label='Copy Link to Asset'
```
hmhealey, line 213: This should be translatable like the other text in this component. Since this is a class component, our best bet for this is `Utils.localizeMessage(<id>, <defaultMessage>)`

components/size_aware_image.jsx
```diff
@@ -170,6 +182,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+        const copyLink = (
+            <OverlayTrigger
+                className='hidden-xs'
+                delayShow={500}
+                placement='top'
+                overlay={copyLinkTooltip}
+                rootClose={true}
+            >
+                <button
+                    key='copy-link'
```
hmhealey, line 209: Nitpicking, but this key isn't needed since this isn't in an array

components/size_aware_image.jsx
```diff
@@ -178,18 +263,57 @@ export default class SizeAwareImage extends React.PureComponent {
 
             return (
                 <div
-                    onClick={this.handleImageClick}
-                    className={className}
-                    style={this.state.imageWidth > MIN_IMAGE_SIZE ? {
-                        width: this.state.imageWidth + 2, // 2px to account for the border
-                    } : {}}
+                    className={classNames('small-image-utility-buttons-wrapper')}
```
hmhealey, line 266: More nitpicking, could you remove the use of `classNames` where the class name is a constant?

components/size_aware_image.jsx
```diff
@@ -178,18 +263,57 @@ export default class SizeAwareImage extends React.PureComponent {
...
+                        {image}
+                    </div>
+                    <span
+                        className={classNames('image-preview-utility-buttons-container', 'image-preview-utility-buttons-container--small-image')}
+                        style={this.state.imageWidth > MIN_IMAGE_SIZE ? {
+
+                            // for every pixel the image is wider than MIN, add that to left shift of buttons container
+                            // 20px is the width of collapse button.
+
+                            left: 26 + (this.state.imageWidth - MIN_IMAGE_SIZE),
```
hmhealey, line 284: This might not be something we can improve easily, but is it not possible to position this using a `right` value instead of doing this math here? Constants like this always have me worried that something will change slightly in the CSS and subtly break this without anyone noticing

components/single_image_view/single_image_view.tsx
```diff
@@ -217,6 +217,7 @@ export default class SingleImageView extends React.PureComponent<Props, State> {
                                 src={previewURL}
                                 dimensions={this.state.dimensions}
                                 fileInfo={this.props.fileInfo}
+                                fileURL={fileURL}
```
hmhealey, line 220: Regarding my other comment, this might not be the URL we actually want to use here. Nothing to change here at the moment, but I wanted to make a note of it

hmhealey (changes requested):
Thanks for the PR, @plant99! This looks good so far minus some nitpicking and my questions about the design that hopefully won't cause too many changes here

components/size_aware_image.jsx
plant99: absolute_right

When I started out with right, this happened.

I'll try again later today, what you said definitely makes sense to me. Thanks for the review. I've done other changes.

matthewbirtch:
> * Is the copied link supposed to be a public link that's viewable by users outside Mattermost? If so, that's off-by-default, and we'll have to add a state for when that button isn't visible.

@hmhealey It's intended to be the equivalent of this that currently exists in the gallery - so I guess it needs to be treated the same
https://community-daily.mattermost.com/files/np37pjdc87d6mq7qngut4kiwic/public?h=_o66hNt6WmVWQ_fujI8lYouNov_wMogJGT5gRw7AqTc

> 2\. If that is the case, we also don't have the public link available on the post list when this is shown, so we might need to make some changes to the API on the server which we may have to run past the security team. Public links for files are only generated on demand, so we wouldn't have that available without asking the server for it.
>     Ideally, I think we could get the public link automatically as part of the data sent from the server when loading information about files, but there's a chance that that's a security issue since we currently track when people access the public link for files in the Auditing system. Alternatively, we could fetch the public link when the user clicks on the "copy link" button, but there may be a delay there as we request it from the server, so that could be a UX concern

Can we just make it behave exactly the same as it does in the full screen gallery view?

components/size_aware_image.jsx
plant99: You were 'right' :D , when I explicitly defined position of the parent container to relative, `right` tag started working. Please review the latest commit.

plant99:
Although waiting for a confirmation from @hmhealey   on @matthewbirtch 's comments , could someone please confirm if https://github.com/mattermost/mattermost-webapp/blob/master/components/file_preview_modal/file_preview_modal.tsx#L260-L272 can be used to know if link-sharing is `true` ? Or if there's a part of the state which should be looked at?

esethna:
> Although waiting for a confirmation from @hmhealey on @matthewbirtch 's comments , could someone please confirm if https://github.com/mattermost/mattermost-webapp/blob/master/components/file_preview_modal/file_preview_modal.tsx#L260-L272 can be used to know if link-sharing is `true` ? Or if there's a part of the state which should be looked at?

cc// @hmhealey on the above

hmhealey:
@plant99 `enablePublicLink` is actually the field you want to look at there since that's the one that comes from the config. `showPublicLink` is used to only show that button for time attachments and not Markdown images or the previews that show if you send a message that contains a link to an image.

Its value comes from https://github.com/mattermost/mattermost-webapp/blob/master/components/file_preview_modal/index.ts#L34

plant99:
Thanks for the information @hmhealey  will work on this, and I'll re-request review when this logic is changed.

hmhealey:
Thanks very much! Let me know if you have any questions or need any help on this since I could definitely see loading the public link for those images on-demand being a bit tricky with our current APIs

plant99:
Hi @hmhealey , could you please check https://github.com/mattermost/mattermost-webapp/pull/9705/commits/ca4c16a0fb15642c455c948af03c642d0e211c6b?

Also @matthewbirtch this needs a UX review again, since copy URL button is hidden for some cases.

michelengelen (changes requested):

matthewbirtch (approved):

esethna (changes requested):
Looks like the download and copy link buttons might be missing an on-click effect. See the post hover menu for examples:

https://gyazo.com/dd4ce25d9c68219d13abb52a467f1b3b

cc// @matthewbirtch

matthewbirtch:
> Looks like the download and copy link buttons might be missing an on-click effect. See the post hover menu for examples:
> 
> https://gyazo.com/dd4ce25d9c68219d13abb52a467f1b3b
> 
> cc// @matthewbirtch

You're right @esethna. Good catch. @plant99 can we add the `active` state? Should follow the same style as the post hover icon buttons as noted in the example @esethna provided

components/single_image_view/single_image_view.tsx
hmhealey: Coming back to this, the URL we're passing in is the private URL that requires you to have access to the channel to see, so that won't work here. I think we can just remove this prop then and make the `SizeAwareImage` responsible for loading the public link

components/size_aware_image.jsx
```diff
@@ -230,6 +349,20 @@ export default class SizeAwareImage extends React.PureComponent {
         );
     }
 
+    copyLinkToAsset = () => {
+        const fileURL = this.props.fileURL;
+        copyToClipboard(fileURL ?? '');
```
hmhealey, line 354: I think what we need to do here is that once the user clicks on this button, we call the `getFilePublicLink` action (as seen [here](https://github.com/mattermost/mattermost-webapp/blob/master/components/file_preview_modal/file_preview_modal_main_actions/file_preview_modal_main_actions.tsx#L49)) to get the link and then once that's done, we continue on and copy it.

We have to wait until the user clicks on this button to load the link or else we'll end up making a bunch of requests for those links for every file when we likely won't need it most of those times.

hmhealey (changes requested):
> Hi @hmhealey , could you please check [ca4c16a](https://github.com/mattermost/mattermost-webapp/commit/ca4c16a0fb15642c455c948af03c642d0e211c6b)?

That works for showing/hiding the Copy Link button, but now we're getting to the other thing I mentioned where we have to call another API to get the public link. I'll try to explain below

components/size_aware_image.jsx
plant99: On it, thanks for describing here so clearly.

plant99: Hi @hmhealey could you please check https://github.com/mattermost/mattermost-webapp/pull/9705/commits/509f8930e9301c951421da285d98dc82b21a0bba commit? 

TL;DR: Unlike file preview component, the api call couldn't happen in component mount because they'd be unnecessary. 

So the click handler triggers `getFilePublicLink` and [another function](https://github.com/mattermost/mattermost-webapp/pull/9705/commits/509f8930e9301c951421da285d98dc82b21a0bba#diff-83c6734e3d6d99f358f1bbc40bd17e5a478d44eed57d06e7e1af20f4c31aaef0R354) which runs on render function everytime(on every update) to see if `filePublicLink` is updated. 

If yes, it copies to clipboard and resets `filePublicLink` in global state.

hmhealey: Yep, having the download use the internal link makes sense to me. We know the user can see the file, so we don't need to worry about permissions with it.

Unfortunately, that way of copying the link after getting the URL doesn't work. We can't have side effects like that in the render method since React requires it to be pure. It should actually be easier to do that in the click handler itself though. Instead of storing anything in state, I think you should be able to make the click handler itself wait to receive the response from the server by doing something like
```tsx
onClick = () => {
    // You can still set some "copying..." state here if we want to show that to the user
    this.props.actions.getPublicLink(fileId).then((data) => {
        copyToClipboard(data); // You should get the URL returned from the server here, but I'm not sure of the exact format
        // TODO Set state to show it was copied
    });
}
```

That should also reduce the number of props that we need to pass into the `SizeAwareImage` that's specific to this use case. I'm not a huge fan of adding all this functionality into that component since it was fairly self-contained and focused just on sizing and placeholders before, so I'd like to limit the amount we're passing into it for this new functionality.

hmhealey (changes requested):
See https://github.com/mattermost/mattermost-webapp/pull/9705#discussion_r812157869

components/size_aware_image.jsx
plant99: I agree, that was unnecessarily complex. I've pushed a commit to fix that.

esethna (changes requested):
The on-click effect doesn't look quite right @plant99:

Expected: 
image

Observed: Notice the icon color
image

esethna:
@plant99 I think there may be a bug where copying the image link, then pasting it in the text input and posting is causing a crash

esethna (approved):

hmhealey:
We should definitely look into that crash. It looks like it has to do with the `SizeAwareImage` not having a `fileInfo` passed into it for external images. It should be able to handle when `fileInfo` is missing

plant99:
I see that in case of image link embeds, the parent is not `SingleImageView`, but `post_view/message_attachment`.

Also there are other parents namely `post_view/post_attachment_opengraph`, `markdown_image/markdown_image.jsx`, `components/post_view/post_image/post_image`.

I'll modify `SingleImageView` to account for other sources of images while rendering copy and download buttons.

Meanwhile, the crash was due to a [reference to `fileInfo` without checking if it's valid. ](https://github.com/mattermost/mattermost-webapp/pull/9705/files#diff-83c6734e3d6d99f358f1bbc40bd17e5a478d44eed57d06e7e1af20f4c31aaef0R263)

sass/components/_files.scss
```diff
@@ -306,8 +306,10 @@
 
                     &:active {
                         background: rgba(var(--button-bg-rgb), 0.08);
-                        color: v(button-bg);
-                        fill: v(button-bg);
+
+                        .icon {
+                            color: v(button-bg);
```
michelengelen, line 311: the usage of `v()` is deprecated in the webapp. Could you please replace that with `var(--button-bg)`? Thanks!

michelengelen (changes requested):

plant99:
> I see that in case of image link embeds, the parent is not `SingleImageView`, but `post_view/message_attachment`.
> 
> Also there are other parents namely `post_view/post_attachment_opengraph`, `markdown_image/markdown_image.jsx`, `components/post_view/post_image/post_image`.
> 
> I'll modify `SizeAwareImage` to account for other sources of images while rendering copy and download buttons.
> 
> Meanwhile, the crash was due to a [reference to `fileInfo` without checking if it's valid. ](https://github.com/mattermost/mattermost-webapp/pull/9705/files#diff-83c6734e3d6d99f358f1bbc40bd17e5a478d44eed57d06e7e1af20f4c31aaef0R263)

Continuing on this, for images of which the sources are external, like from link attachments or markdown image tags.

1. Should copy link functionalities be available for them? Download button can use `fileURL` / `src` props.
2. If yes, are there recommended ways to distinguish these two types of images from any prop? One can always use class names, wanted some opinions. Since  getting a public url from server won't be necessary for images with external source.


cc: @hmhealey

hmhealey:
I'd say they probably should. They're different, but we've been slowly but surely making them work the same over time. That also applies to Markdown images (from typing something like `alt text`).

I think the biggest difference between them would probably be whether or not that `postInfo` prop is provided since it would be provided for image attachments, but not for external images

components/size_aware_image.jsx
```diff
@@ -170,6 +196,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+                placement='top'
+                overlay={downloadTooltip}
+                rootClose={true}
+            >
+                <a
+                    href={this.isInternalImage() ? fileURL : src}
+                    className='style--none size-aware-image__download'
+                    target='_blank'
+                    rel='noopener noreferrer'
+                    download={ariaLabelImage}
```
hmhealey, line 263: I missed this earlier, but passing a value for this changes the name of the file being downloaded which I don't think we want. I think just passing `download={true}` here will cause the original name of the file to be preserved, but we should confirm that that works for both internal and external images

components/size_aware_image.jsx
```diff
@@ -170,6 +196,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+
+        const copyLinkTooltip = (
+            <Tooltip
+                id='copy-link-tooltip'
+                className='hidden-xs'
+            >
+                <FormattedMessage
+                    id='single_image_view.copy_link_tooltip'
+                    defaultMessage={this.state.linkCopiedRecently ? 'Copied' : 'Copy link'}
+                />
```
hmhealey, line 211: We can't change just the `defaultMessage` here when `this.state.linkCopiedRecently` changes. We also need to change the ID to have it pick up a different string from the translation files which are stored separately.

Also, when you do that, you'll need to wrap the IDs in the `t` function from `utils/i18n` so that they're detected by our translation tools. Those tools can pick up an ID automatically from a `FormattedMessage` if it's a constant, but since we're changing the ID, we need to use `t` to do that for us.

After that, it'll look something like
```jsx
<FormattedMessage
    id={this.state.linkCopiedRercently ? t('single_image_view.copy_link_tooltip') : t('single_image_view.copied_link_tooltip')}
    defaultMessage={this.state.linkCopiedRecently ? 'Copied' : 'Copy link'}
/>
```

components/size_aware_image.jsx
```diff
@@ -170,6 +196,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+                delayShow={500}
+                placement='top'
+                overlay={copyLinkTooltip}
+                rootClose={true}
+            >
+                <button
+                    className={classNames('style--none', 'size-aware-image__copy_link', {
+                        'size-aware-image__copy_link--recently_copied': this.state.linkCopiedRecently,
+                    })}
+                    aria-label={`${ariaLabelImage}${localizeMessage('copy url')}`}
```
hmhealey, line 226: This actually has to be done a bit differently. We can't  concatenate translated strings like this since some languages may structure text differently. That said, I think we don't need to include the image name in this label either since we'd just be repeating ourselves, so I think we can just have this aria label be something like "Copy link" without the rest of it.

Also, `localizeMessage` works like `FormattedMessage` where you need to pass in two arguments, an ID and a default message, so it would need to be called as `localizeMessage('single_image_view.copy_link_tooltip', 'Copy link')`

components/size_aware_image.jsx
```diff
@@ -170,6 +196,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+                overlay={downloadTooltip}
+                rootClose={true}
+            >
+                <a
+                    href={this.isInternalImage() ? fileURL : src}
+                    className='style--none size-aware-image__download'
+                    target='_blank'
+                    rel='noopener noreferrer'
+                    download={ariaLabelImage}
+                    aria-label={`${ariaLabelImage}${localizeMessage('download image')}`}
```
hmhealey, line 264: The same thing should be done with this aria-label. Similarly, I think we can just use the tooltip here

hmhealey (changes requested):
Those changes fixed the error which is great, but I found a couple other things we had previously missed, particularly around the translatable strings in the UI.

Also, if you could please avoid rebasing a PR after it's been reviewed, that'd be really helpful for us. Force pushing makes it hard to see what we've already reviewed and what's new so merging in updates from master is preferred

components/size_aware_image.jsx
michelengelen: @hmhealey is right about this one, but to improve performance we should probably define 2 different `FormattedMessage` components based on the state:

```
this.state.linkCopiedRercently ? (
    <FormattedMessage
        id={t('single_image_view.copy_link_tooltip')}
        defaultMessage={'Copied'}
    />
) : (
    <FormattedMessage
        id={t('single_image_view.copied_link_tooltip')}
        defaultMessage={'Copy link'}
    />
)
```

components/size_aware_image.jsx
```diff
@@ -170,6 +196,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+                >
+                    <i className='icon icon-download-outline style--none'/>
```
michelengelen, line 266: nit (since this works as well): We can use icon components from the `@mattermost/compass-icons` now (renders plain `svg`), which are a lot easier to align correctly.
example: 

```jsx
import {DownloadOutlineIcon} from '@mattermost/compass-icons/components'

...

<DownloadOutlineIcon
    size={24} // any number will be use to set width and height
    color={'#FFF'} // any valid CSS color string (even variables are possible)
/>
```

All other props will be spread onto the `svg` as well, so using aria attributes do work.

components/size_aware_image.jsx
```diff
@@ -170,6 +196,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+                </a>
+            </OverlayTrigger>
+
```
michelengelen, line 269: please remove the empty line here

components/size_aware_image.jsx
```diff
@@ -170,6 +196,79 @@ export default class SizeAwareImage extends React.PureComponent {
...
+        const download = (
+            <OverlayTrigger
+                className='hidden-xs'
+                delayShow={500}
+                placement='top'
+                overlay={downloadTooltip}
+                rootClose={true}
+            >
+                <a
+                    href={this.isInternalImage() ? fileURL : src}
```
michelengelen, line 259: do you open the image or do you download it in any case?

in case of a download we should add the `role='button'` attribute for semantics.

components/size_aware_image.jsx
```diff
@@ -230,6 +356,43 @@ export default class SizeAwareImage extends React.PureComponent {
         );
     }
 
+    isInternalImage = () => {
+        return this.props.fileInfo !== undefined;
+    }
```
michelengelen, line 361: this can just be a boolean

```suggestion
    isInternalImage = this.props.fileInfo !== undefined;
```

components/size_aware_image.jsx
michelengelen, line 259: see comment regarding `this.isInternalImage` below. I don't think we need to wrap it in a method.

sass/components/_post.scss
```diff
@@ -1651,6 +1651,98 @@
...
+    // styles for utility buttons i.e download, and copy source link in image preview
+    .image-preview-utility-buttons-container {
+        // greater z-index than img
+        position: absolute;
+        z-index: 3;
+        top: 8px;
+        right: 8px;
+        display: none;
+        padding: 4px;
+        border: 1px solid rgba(v(center-channel-color-rgb), 0.16);
```
michelengelen, line 1663:
```suggestion
        border: 1px solid rgba(var(--center-channel-color-rgb), 0.16);
```

sass/components/_post.scss
```diff
@@ -1651,6 +1651,98 @@
         opacity: 0.6;
     }
 
+    // styles for utility buttons i.e download, and copy source link in image preview
+    .image-preview-utility-buttons-container {
+        // greater z-index than img
+        position: absolute;
+        z-index: 3;
```
michelengelen, line 1658: is it in the same stacking context? If yes a value of `1` should be sufficient.

michelengelen (changes requested):

plant99:
Thanks @hmhealey for the suggestions and reviewing again, and apologies about the rebase I had realized later, but it won't happen again :)

Also thanks @michelengelen for reviewing again, I've understood both of your suggestions and will implement them asap.

components/size_aware_image.jsx
hmhealey: In that case, you actually don't need the `t` function since the tooling automatically picks up translation IDs from `FormattedMessage`s

components/size_aware_image.jsx
plant99: I read that we can't start download on external image urls. Please correct me if I'm wrong. So this <a> just opens an image in a new tab.

sass/components/_post.scss
plant99: also this change puts the buttons behind the image, will test it with 2, then 3.

plant99: this is weird, it's happening for some images in test server, but for the same image in my local setup, the buttons stack correctly.

matthewbirtch:
Seems like there might be a z-index issue now on smaller images:
image

sass/components/_post.scss
plant99: cc: @matthewbirtch I’ll fix that.

plant99: https://github.com/mattermost/mattermost-webapp/blob/master/sass/components/_files.scss#L222 will make `z-index: 3`

plant99: That seems to have fixed the issue.

matthewbirtch (approved):

jgilliam17:
Thanks @plant99 
few small issues

1. on some markdown images, collapse button is behind the image
Screen Shot 2022-03-17 at 2 39 04 PM

2. wide images are not displaying correctly, container is too wide, copy/download icons are out of view
Screen Shot 2022-03-17 at 3 04 19 PM

3.  smaller markdown image links have the buttons slightly cut off on the top 
Screen Shot 2022-03-17 at 3 16 16 PM

[E2E report](https://automation-dashboard.vercel.app/cycles/89e43a2d-2f46-43b8-a8a8-628880986d5a) has high number of failures (54) vs. [master](https://automation-dashboard.vercel.app/cycles/6971bf26-a818-4037-b694-c8724225da1f) (36), not sure yet if many are related to this PR. I will re-run after above is fixed.

plant99:
1. Thanks for pointing it out, have fixed this, please check.
2. There is a constraint on width of images in master, it's mostly something I've changed, will push a fix asap.
3. @matthewbirtch need some inputs on this(case of  smaller markdown images) again, should we reduce the size of buttons in this case, to accommodate the size of image? Or should I look into making the post at least the size of these buttons?

hmhealey:
I thought the minimum size of those images was larger so that that sort of thing wouldn't happen? I'm surprised that the minimum is only 50px tall and not more like 100-150px.

components/size_aware_image.jsx
plant99: > This might not be something we can improve easily, but is it not possible to position this using a `right` value instead of doing this math here? Constants like this always have me worried that something will change slightly in the CSS and subtly break this without anyone noticing

 @hmhealey this alignment had broken the small image wrapper. I have used left alignment again in the latest commit.

plant99:
@jgilliam17 could you please run E2E tests again? I've fixed the issues you had mentioned above along with Matthew's suggestions.

components/size_aware_image.jsx
```diff
@@ -293,7 +292,8 @@ export default class SizeAwareImage extends React.PureComponent {
             if (this.state.imageWidth < MIN_IMAGE_SIZE) {
                 className += ' small-image__container--min-width';
             }
-
+            // 24 is the offset on a 48px wide image, for every pixel added to the width of the image, it's added to the left offset to buttons
+            let wideImageButtonsOffset = 24 + this.state.imageWidth - MIN_IMAGE_SIZE;
```
michelengelen: this can be a `const`

components/size_aware_image.jsx
```diff
@@ -309,30 +309,44 @@ export default class SizeAwareImage extends React.PureComponent {
...
                         className={classNames('image-preview-utility-buttons-container', 'image-preview-utility-buttons-container--small-image', {
-                            'image-preview-utility-buttons-container--small-image-no-copy-button': !enablePublicLink || !this.isInternalImage,
+                            'image-preview-utility-buttons-container--small-image-no-copy-button': !enablePublicLink,
                         })}
+                        // since there is a max-width constraint on images, a max-left clause follows.
+                        style={this.state.imageWidth > MIN_IMAGE_SIZE ? (
+                            // cases with no copy link buttons has different offsets than with copy buttons
+                            !enablePublicLink ? {left : `min(${wideImageButtonsOffset + 8}px, calc(100% - 24px)`}
+                            :  {left : `min(${wideImageButtonsOffset}px, calc(100% - 32px)`}
+                         ) : {}}
```
michelengelen: please try to avoid nested ternary expressions, especially when you are using object literals directly in props, since it will have no reference to a stored object and results in unnecessary re-renders.

Better declare it above (just below line 296) and use it here. That way you even have more methods to choose from (e.g. a switch/case).

But I would rather add something like this:

```
const modifier = !enablePublicLink ? 8 : 0;
const styles = this.state.imageWidth > MIN_IMAGE_SIZE ? {
    left: `min(${wideImageButtonsOffset + modifier}px, calc(100% - ${32 - modifier}px)`
} : {};
```

components/size_aware_image.jsx
```diff
@@ -309,30 +309,44 @@ export default class SizeAwareImage extends React.PureComponent {
...
                         {download}
                     </span>
                 </div>
             );
         }
 
+        let utilityButtonsWrapper = null;
+        // handling external small images
+        if (this.state.isSmallImage && !this.isInternalImage) {
+            utilityButtonsWrapper = <></>;
```
michelengelen, line 343: this is essentially the same as passing in `null`

Please re-consider using a ternary and `const` instead

michelengelen (changes requested):

components/size_aware_image.jsx
plant99: thank you @michelengelen for the feedback, some are already present in the latest commit, https://github.com/mattermost/mattermost-webapp/pull/9705/files#diff-83c6734e3d6d99f358f1bbc40bd17e5a478d44eed57d06e7e1af20f4c31aaef0R305 but this is neater, the code-block you suggest, I'll reflect this change there.

jgilliam17:
Thanks @plant99 
E2E report pending
Seeing one issue, copy/download button on wide images is positioned slightly off the image
Screen Shot 2022-03-30 at 9 49 41 AM

plant99:
Oh @jgilliam17 I think it's a requested feature by @matthewbirtch only for images treated as 'small'.

matthewbirtch:
> Oh @jgilliam17 I think it's a requested feature by @matthewbirtch only for images treated as 'small'.

@plant99 @jgilliam17 yes, this is as expected. However, maybe if the image is small in height, but wider than 100px, we can adjust the positioning so it doesn't have to sit so far off the right edge?

matthewbirtch (changes requested):
I just noticed markdown images have the hover buttons positioned differently from regular attached images

**Markdown**
image

**Attached**
image

Other than that, and the previous comment, I think we're good

michelengelen (changes requested):
a few smallish suggestions

michelengelen (approved):

matthewbirtch (approved):

jgilliam17:
Thanks @plant99 
Tested, looks good to merge.
- Verified copy/download button on images
- [E2E report](https://automation-dashboard.vercel.app/cycles/d1d5b64a-9f9f-4c65-8092-3d9a6bbf38b1) looks good, no PR related failures.

jgilliam17 (approved):

michelengelen:
@hmhealey gentle ping on this.

components/size_aware_image.jsx
hmhealey: I'm actually surprised that this wasn't picked up be ESLint. It shouldn't allow nested ternaries, but maybe it has difficulty with that happening in JSX

hmhealey (approved):

hmhealey:
Thanks again for the PR, @plant99! Definitely looking forward to having this in since it'll definitely help those images feel polished

plant99:
Thank you for helping folks, it was my first major PR to this repository and learned a lot from it. Would definitely help me make meaningful changes quicker for the upcoming ones!

PR #9706, opened 2022-01-26
Cypress/E2E: fixed tests (on custom emojis & integrations) failing due to not finding 'back to mattermost' button

Phrynobatrachus:
#### Summary
failures all look similar

image

The top nav is slightly out of view. Consistently reproduced locally. I tried a few different scrolling solutions as well as `findByRole` but none quite worked, so I just changed it to `visit` for now since it otherwise isn't relevant for these tests. There was one other alternative that _did_ work though, which was:

``` Javascript
// shrink the view which ends up in top-left over the button
cy.viewport(200, 200);
// same query as before
cy.findByText('Back to Mattermost').click().wait(TIMEOUTS.FIVE_SEC);
// resize back to defaults (width from cypress.json, height from cy default)
cy.viewport(1300, 600);
```

I mention that because I don't think any of these are quite right. This `visit()` fix won't  work if the element you're having trouble selecting isn't a link, and generally it seems like there's some inconsistency in scroll behavior (whether from Cypress or webapp idk) that should be accounted for. Suggestions/thoughts welcome. 

Ticket
https://mattermost.atlassian.net/browse/MM-41274

#### Release Note
```release-note
NONE
```


e2e/cypress/integration/emoji/custom_emoji_1_spec.js
```diff
@@ -88,7 +88,7 @@ describe('Custom emojis', () => {
         cy.get('.backstage-form__footer').findByText('Cancel').click().wait(TIMEOUTS.FIVE_SEC);
 
         // # Go back to home channel
-        cy.findByText('Back to Mattermost').should('exist').and('be.visible').click().wait(TIMEOUTS.FIVE_SEC);
+        cy.visit('/');
```
M-ZubairAhmed, line 91: I am 0/5 on this but do you think we should visit the team channel directly instead of `/${team.name}/channels/off-topic`

Phrynobatrachus: it's probably a bit more correct yeah, will push up

Phrynobatrachus:
there is one more I almost missed, fixing now

Phrynobatrachus:
@M-ZubairAhmed after the most recent commit, the test `MM-T4463` in `recently_used_emoji_spec.js` still fails on the assertion:

```
cy.get('#recent_reaction_0').should('exist').then((recentReaction) => {
            // * Assert that custom emoji is present as most recent in quick reaction menu
            expect(recentReaction[0].style.backgroundImage).to.include('api/v4/emoji');
        });
```
And it indeed doesn't appear to be there.

Screenshot from 2022-01-26 14-58-19

Seems like an actual bug?

M-ZubairAhmed:
> @M-ZubairAhmed after the most recent commit, the test `MM-T4463` in `recently_used_emoji_spec.js` still fails on the assertion:
> 
> 
> 
> ```
> 
> cy.get('#recent_reaction_0').should('exist').then((recentReaction) => {
> 
>             // * Assert that custom emoji is present as most recent in quick reaction menu
> 
>             expect(recentReaction[0].style.backgroundImage).to.include('api/v4/emoji');
> 
>         });
> 
> ```
> 
> And it indeed doesn't appear to be there.
> 
> 
> 
> Screenshot from 2022-01-26 14-58-19
> 
> 
> 
> Seems like an actual bug? 
> 
> 
> 
> 

Oh let me check it

M-ZubairAhmed (approved):

saturninoabril:
Oh no. Since Cypress renders the page in an iframe, there might be some styling conflict between Cypress and the webapp that caused discrepancy in rendering the page. This makes Cypress unreliable in some cases unfortunately :(

Changes look good to me, thanks Cass!

furqanmlk:
> Oh no. Since Cypress renders the page in an iframe, there might be some styling conflict between Cypress and the webapp that caused discrepancy in rendering the page. This makes Cypress unreliable in some cases unfortunately :(
> 
> Changes look good to me, thanks Cass!

Looks like /integrations/* tests will have same issue.

furqanmlk:
@Phrynobatrachus 
This test case is still failing 
`MM-T4438 Changing the skin of emojis should apply the same skin to emojis in recent section`

https://mattermost-cypress-report.s3.amazonaws.com/109060-1ca004b-onprem-ent-webapp-pr-9706-mattermost/mm-ee-test1ca004b/screenshots/e2e/cypress/integration/emoji/recently_used_emoji_spec.js/Recent%20Emoji%20--%20MM-T4438%20Changing%20the%20skin%20of%20emojis%20should%20apply%20the%20same%20skin%20to%20emojis%20in%20recent%20section%20(failed).png


https://mattermost-cypress-report.s3.amazonaws.com/109060-1ca004b-onprem-ent-webapp-pr-9706-mattermost/mm-ee-test1ca004b/mochawesome.html

furqanmlk (approved):
LGTM

image

PR #9707, opened 2022-01-26
MM-41319 Reduced number of listeners registered by MenuWrapper

hmhealey:
Currently, every instance of MenuWrapper registers two listeners on the document that respond to every keyup and click event made in the app. With only a couple channels, this is already 12 listeners running on each key press or click, but when you have more MenuWrappers mounted (notably each sidebar channel and category uses one even if they're not visible), that starts to add up the amount of time taken to process those events.

Since both of those listeners are used to close the menu when it's open, I've changed those to only be registered once the menu gets opened and then removed when it's closed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41319

#### Related Pull Requests
TBD

#### Screenshots
I added log statements to the listeners to show when they trigger. Here's a before and after with a brand new account

Before:

https://user-images.githubusercontent.com/3277310/151253855-49d84f8b-50f9-4bb2-9882-c888366f8f1c.mov

After:

https://user-images.githubusercontent.com/3277310/151253968-d7f7054a-1c14-47ec-9985-952e06ad15c0.mov

#### Release Note
```release-note
Reduce the number of menu components listening for keyboard and mouse events
```


components/widgets/menu/menu_wrapper.tsx
```diff
@@ -62,7 +63,26 @@ export default class MenuWrapper extends React.PureComponent<Props, State> {
...
+    }
+
     public componentWillUnmount() {
+        if (this.state.open) {
+            this.removeEventListeners();
+        }
+    }
+
+    private addEventListeners() {
+        document.addEventListener('click', this.closeOnBlur, true);
```
larkox, line 81: nit: out of the scope of the pr, but what about moving the `click` and `keyup` events names to a constant? (May be in this file)

larkox (approved):
Looks great!

M-ZubairAhmed (approved):

components/widgets/menu/menu_wrapper.tsx
hmhealey: Yeah, we could consider that. I don't personally think it's that helpful though since they're standard values that come from the browser. I'm not opposed to doing that, but I'm too lazy to go back and do it here 😛

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- Verified channel and category menus in the LHS are working as expected.
- [E2Es look](https://automation-dashboard.vercel.app/cycles/242a151e-c935-403d-b882-c1b1868b115f) good, no PR related failures.

PR #9708, opened 2022-01-26
MM-42320 Remove excess keyboard listeners from post components

hmhealey:
Similar to #9707, each post renders a keyboard handler to detect whether or not alt is pressed because that switches the cursor to show that you can click on a post to mark it as unread. This happens once for each post on the screen which can add a bit of processing time when a user is typing a message.

By only registering those handlers when the user hovers over a post, we ensure that we only have one registered at a time at most which should hopefully speed up typing performance slightly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41320

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/9707

#### Screenshots
Here's a before and after with log statements added to the listeners.

Before:

https://user-images.githubusercontent.com/3277310/151258929-8e1f1ab6-202e-4a88-97b6-9164f0993c37.mov

After:

https://user-images.githubusercontent.com/3277310/151258946-dce6d5be-a030-4032-9dff-2d45bfe6f450.mov

#### Release Note
```release-note
Reduce the number of post components listening for keyboard events
```


michelengelen (approved):
LGTM ... although I feel like all of these components do basically the same thing and we can probably extract it into one component that is being used anywhere a text input is needed.

WDYT @hmhealey @matthewbirtch ?

PS: this is definitely OOS for this PR, but we should definitely think about it!

M-ZubairAhmed (approved):

hmhealey:
Yeah, that's 100% something I'd like to do. The post components (and post textboxes) have so much duplicated code, but they're not something we've had a chance to get to yet. We do have a goal for the year to refactor some areas of heavy tech debt, and those are both high up for me on that list.

Alternatively, I feel like that could make a good custom hook, but we can't do that either until these components are rewritten to be functional

jgilliam17:
//update-branch

jgilliam17 (approved):
Thanks @M-ZubairAhmed 
Tested, looks good to merge.
- Verified cursor turns to a hand pointer when holding alt and hovering over a post. Additionally, when holding alt key while the mouse cursor is not hovering over any post, the cursor is not changing to hand pointer.
- [E2E report ](https://mattermost-cypress-report.s3.amazonaws.com/110147-2680292-onprem-ent-webapp-pr-9708-mattermost/mm-ee-test2680292/mochawesome.html) looks good, no PR related failures.

PR #9714, opened 2022-01-27
Add overall score header

pablovelezvidal:
#### Summary
This PR adds the code for creating the overall score header along with a couple of reusable components:
- Chip List (could be enhanced for other needs). Right now works perfect for what is needed in the workspace health dashboard.
- Circular Chart

#### Ticket Link
N/A

#### Screenshots

Captura de pantalla 2022-01-27 a las 18 54 51

#### Release Note

```release-note
NONE
```


components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -325,41 +333,11 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
...
-            return (
-                <Chip
-                    key={chipKey}
-                    id={id}
-                    defaultMessage={`${defaultMessage}: ${count}`}
-                    className={chipKey}
-                />
-            );
-        });
+        const chipsList = (<ChipsList chipsData={accordionDataChips}/>);
```
michelengelen, line 340: no need to create a variable for that ... you can just assign that directly in the return:

```typescript
return {
    title,
    description,
    icon,
    items,
    extraContent: <ChipsList chipsData={accordionDataChips}/>,
};
```

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -370,17 +348,23 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
...
-            />
+            <div className='admin-console__wrapper'>
+                <div className='admin-console__content'>
+                    <OverallScore chips={overallScoreChips}/>
+                    <Accordion
+                        accordionItemsData={accData}
+                        expandMultiple={true}
+                    />
+                </div>
+            </div>
```
michelengelen, line 367: do we really need both `div`s?

components/common/circular_chart/circular_chart.scss
```diff
@@ -0,0 +1,47 @@
...
+            stroke: var(--denim-status-online);
+        }
+
+        &.info .circle {
+            stroke: var(--sidebar-text-active-border);
+        }
+
+        &.error .circle {
+            stroke: var(--error-text);
+        }
```
michelengelen, line 20: to be consistent we should use the `--sys-*` variables here as well

components/common/circular_chart/circular_chart.tsx
```diff
@@ -0,0 +1,55 @@
...
+import React from 'react';
+
+import './circular_chart.scss';
+
+type CircularChartProps = {
+    value: number;
+    isPercentage: boolean;
+    width: number;
+    height: number;
+    type: 'info' | 'warning' | 'error' | 'success';
```
michelengelen, line 12: nit: we could probably define a global type for this and use it here and with the chips ... WDYT?

components/common/circular_chart/circular_chart.tsx
```diff
@@ -0,0 +1,55 @@
...
+                viewBox='0 0 36 36'
+                className={`circular-chart ${type}`}
+                width={width ? width.toString() : '36'}
+                height={height ? height.toString() : '36'}
+            >
+                <path
+                    className='circle-bg'
+                    d='M18 2.0845
+                a 15.9155 15.9155 0 0 1 0 31.831
+                a 15.9155 15.9155 0 0 1 0 -31.831'
```
michelengelen, line 34:
```suggestion
                    d='M18 2.0845 a 15.9155 15.9155 0 0 1 0 31.831 a 15.9155 15.9155 0 0 1 0 -31.831'
```

components/common/circular_chart/circular_chart.tsx
```diff
@@ -0,0 +1,55 @@
...
+                />
+                <path
+                    className='circle'
+                    strokeDasharray={`${value}, 100`}
+                    d='M18 2.0845
+                a 15.9155 15.9155 0 0 1 0 31.831
+                a 15.9155 15.9155 0 0 1 0 -31.831'
```
michelengelen, line 41:
```suggestion
                    d='M18 2.0845 a 15.9155 15.9155 0 0 1 0 31.831 a 15.9155 15.9155 0 0 1 0 -31.831'
```

i18n/en.json
```diff
@@ -1815,6 +1815,11 @@
   "admin.reload.reloadDescription.featureName": "Reload Configuration from Disk",
   "admin.reload.reloadDescription.recycleDatabaseConnections": "Environment > Database > Recycle Database Connections",
   "admin.reload.reloadFail": "Reloading unsuccessful: {error}",
+  "admin.reporting.workspace_optimization.chip_problems": "Problems: {count}",
+  "admin.reporting.workspace_optimization.chip_suggestions": "Suggestions: {count}",
+  "admin.reporting.workspace_optimization.chip_warnings": "Warnings: {count}",
+  "admin.reporting.workspace_optimization.overall_workspace_score": "Overall Workspace Score",
+  "admin.reporting.workspace_optimization.overall_workspace_score_description": "Stay on top of how optimized your workspace is! This score is a total of the categories of key indicators of health and growth below. Ensure your workspace is in its most optimized state so you and your user base get the most out of Mattermost.",
```
michelengelen, line 1822: not related: we should port the other translations (currently under `workspaceOptimization.*`)

michelengelen (changes requested):
A few suggestions, but looks good overall! 👍🏼

components/admin_console/workspace-optimization/chips_list.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export type ChipsInfoType = { [key in ChipsInfoKeys]: number };
+
+type ChipsListProps = {
+    chipsData: ChipsInfoType;
+};
+
+const ChipsList = ({
+    chipsData,
+}: ChipsListProps): JSX.Element | null => {
```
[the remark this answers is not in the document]
pablovelezvidal: maybe the campaign was about not to remove the return type but the function default-de facto definition as React.FC which is commonly incorrectly used. This is the reference for even removing it from create-react-app https://github.com/facebook/create-react-app/pull/8177 .

components/admin_console/workspace-optimization/chips_list.tsx
```diff
@@ -0,0 +1,74 @@
...
+                    defaultMessage={'Problems: {count}'}
+                    values={{count}}
+                />
+            );
+            break;
+        }
+
+        return (
+            <Chip
+                key={chipKey}
```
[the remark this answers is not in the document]
pablovelezvidal: AFAIK, the key needs to be unique only in the context it is run, I mean, the element must have a unique and different key for its siblings. What you suggest makes sense, but in this case, since we loop over the unique keys of an object, adding the namespace won't make any difference

components/admin_console/workspace-optimization/dashboard.scss
```diff
@@ -15,6 +15,134 @@
...
+        font-size: 12px;
+        font-style: normal;
+        font-weight: bold;
+        line-height: 16px;
+        text-align: center;
+
+        &.error {
+            background-color: rgba(var(--sys-error-text-color-rgb), 0.3);
+
+            @extend .error;
```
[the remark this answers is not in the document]
pablovelezvidal: of course, good catch. Thanks!

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -291,9 +294,14 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
         },
     };
 
-    type ChipsInfoType = Omit<ItemStatus, 'none' | 'ok'>;
+    const overallScoreChipsData: ChipsInfoType = {
```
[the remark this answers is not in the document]
pablovelezvidal: ok

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -325,41 +333,11 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
 
             // chips will only be displayed for info aka Success, warning and error aka Problems
             if (item.status && item.status !== 'none' && item.status !== 'ok') {
-                chipsInfo[item.status] += 1;
+                accordionDataChips[item.status] += 1;
+                overallScoreChipsData[item.status] += 1;
```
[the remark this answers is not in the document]
pablovelezvidal: ok

components/admin_console/workspace-optimization/dashboard.tsx
pablovelezvidal: ok

components/admin_console/workspace-optimization/dashboard.tsx
pablovelezvidal: nope, we actually only need the wrapper. Good catch, thanks

components/common/circular_chart/circular_chart.scss
pablovelezvidal: the idea is that this component is common, so for general purpose these will be the available colors, then in the console, we style it using the `--sys-`ones as it is defined in `dashboard.scss lines 73 to 85`

components/common/circular_chart/circular_chart.tsx
pablovelezvidal: I think I rather to keep this isolated this way. Keep the other chip type in the context of the sys console.

i18n/en.json
pablovelezvidal: ok

components/admin_console/workspace-optimization/chips_list.tsx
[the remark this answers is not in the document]
michelengelen: yeah, you are right ... it was about that! 👍🏼

components/admin_console/workspace-optimization/chips_list.tsx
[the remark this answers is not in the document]
michelengelen: you are right ... and it is acceptable like it is atm ... I am just thinking about the future, when features might be added

components/admin_console/workspace-optimization/chips_list.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import {FormattedMessage} from 'react-intl';
+
+import Chip from 'components/common/chip/chip';
+
+import './dashboard.scss';
+
+type ChipsInfoKeys = 'info' | 'warning' | 'error';
+
+export type ChipsInfoType = { [key in ChipsInfoKeys]: number };
```
michelengelen, line 13: when we only use it like this we can also remove the extra type definition:

```suggestion
export type ChipsInfoType = { [key in 'info' | 'warning' | 'error']: number };
```

components/common/circular_chart/circular_chart.tsx
```diff
@@ -0,0 +1,51 @@
...
+    type: 'info' | 'warning' | 'error' | 'success';
+};
+
+const CircularChart = ({
+    value,
+    isPercentage,
+    width,
+    height,
+    type,
+}: CircularChartProps): JSX.Element | null => {
```
michelengelen, line 21: we will not return `null` here so it can be removed from the return type

components/common/circular_chart/circular_chart.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return (
+        <div className='CircularChart'>
+            <svg
+                viewBox='0 0 36 36'
+                className={`circular-chart ${type}`}
+                width={width ? width.toString() : '36'}
+                height={height ? height.toString() : '36'}
```
michelengelen, line 28: this is abolutely nitpicking, so dont feel obligated to even react to it! :D :P

`width` or `height` equaling `0` would be falsey, even if it is intended, so it would be better to check for the actual value of it: `width >= 0`

michelengelen (changes requested):
looking great ... just 1 small suggestion and nits! 👍🏼

components/common/circular_chart/circular_chart.tsx
pablovelezvidal: ok

components/common/circular_chart/circular_chart.tsx
pablovelezvidal: lol, the extra-validation doesn't hurt ;)

components/admin_console/workspace-optimization/chips_list.tsx
pablovelezvidal: ok

michelengelen (approved):

components/admin_console/workspace-optimization/dashboard.scss
```diff
@@ -15,6 +15,134 @@
...
+                        stroke: var(--sys-sidebar-text-active-border);
+                    }
+
+                    &.error .circle {
+                        stroke: var(--sys-error-text);
+                    }
+
+                    .circle-bg {
+                        fill: none;
+                        stroke: var(--center-channel-bg);
```
BenCookie95, line 91: Should this be a sys colour?

components/admin_console/workspace-optimization/dashboard.scss
```diff
@@ -15,6 +15,134 @@
...
+                }
+
+                @keyframes progress {
+                    0% {
+                        stroke-dasharray: 0 100;
+                    }
+                }
+
+                .percentage {
+                    fill: var(--center-channel-color);
```
BenCookie95, line 109: Should this be a sys colour?

BenCookie95 (approved):
Nice work, just a few small things

components/admin_console/workspace-optimization/dashboard.scss
pablovelezvidal: yes, good catch. Thanks

components/admin_console/workspace-optimization/dashboard.scss
pablovelezvidal: yes, good catch. Thanks

PR #9715, opened 2022-01-27
e2e: fix 'handle new post' spec read/unread assertions

Phrynobatrachus:
#### Summary
As far as I can tell, the assertions in these tests didn't match the
expected behavior. The first test asserted that the test user will see
a channel they were just added to as read, when they actually see
both the unread highlight and mention badge from the join system message.

There was also a problem with an implicit dependency between the first and second
tests. The second test again expected that the test channel starts as read from
the test user's point of view, but it reuses the same channel from the first test,
which ends on unread.

Certainly possible that I'm missing something obvious here, please correct me if I'm wrong.

#### Release Note
```
NONE
```

### DIFF
--- a/e2e/cypress/integration/websocket/handle_new_post_spec.js
+++ b/e2e/cypress/integration/websocket/handle_new_post_spec.js
@@ -33,15 +33,18 @@ describe('Handle new post', () => {
     });
 
     it('should mark channel as unread when a message is sent in another channel', () => {
-        // * Verify that the channel starts read
-        cy.visit(`/${team1.name}/channels/${channel1.name}`);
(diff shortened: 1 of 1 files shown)


josephbaylon (approved):
LGTM. Tested and passed locally. Thanks @Phrynobatrachus !!

furqanmlk:
Thanks @Phrynobatrachus 
Can you find these tests in Zypher ?

Phrynobatrachus:
Found one related, made 2 new ones for the others.

furqanmlk (approved):
LGTM

image

PR #9730, opened 2022-02-01
Growth spike/add remaining UI elements

pablovelezvidal:
#### Summary
Add the remaining UI components to the health dashboard, as described in the board card:
Peding UI elements:

The ok/successfull green SVG for the success scenarios.
The overal score header Problems big red alert Circle
Implement the buttons for CTA and learn more.
Add ability to accordion component to dont be able to expand if it does not contain elements and not even show the chevron.

#### Ticket Link
https://community.mattermost.com/boards/workspace/7dqusmf8stn3dens5ckhyatbdo/boattb15z9fr4m85etwrdekkduh/v37ny5c5eafnk3dypexwzagip7a/cg98h6gnnsprx5b8nf8f4cdycyh

#### Screenshots
Captura de pantalla 2022-02-01 a las 2 03 43

#### Release Note
```release-note
NONE
```


components/common/accordion/message_item.tsx
```diff
@@ -10,7 +10,7 @@ type Props = {
     severity: 'info' | 'warning' | 'error';
 }
 
-const MessageItem: React.FC<Props> = ({title, description}: Props): JSX.Element | null => {
+const MessageItem = ({title, description}: Props): JSX.Element | null => {
```
pablovelezvidal, line 13: just clean up for the effort of removing the unnecessary React.FC type from functional/hook components

components/trial_benefits_modal/trial_benefits_modal.tsx
```diff
@@ -34,14 +34,6 @@ type Props = {
     trialJustStarted?: boolean;
 }
 
-const ConsolePages = {
```
pablovelezvidal, line 37: moved this to constants, since in the trial benefits modal we have something similar to the learn more links and those point to system console pages, I reused/copied some of the logic there and then realized that it makes sense to keep those system console urls in a central place.

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -354,7 +422,12 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
                 defaultMessage='Workspace Optimization'
             />
             <div className='admin-console__wrapper'>
-                <OverallScore chips={<ChipsList chipsData={overallScoreChips}/>}/>
+                <OverallScore
+                    chips={<ChipsList chipsData={overallScoreChips}/>}
+
+                    // temp to see either the alert image or the circular chart
+                    chartValue={Math.floor(Math.random() * 100) + 1}
```
pablovelezvidal, line 429: random number, if below 50, shows the overal score image red icon, if over, shows the circular chart, so @michaelgamble can see both in action. Either way, I guess we will calculate this value here in the dashboard controller and pass it as a value to the OverallScore component.

components/admin_console/workspace-optimization/cta_buttons.tsx
```diff
@@ -0,0 +1,76 @@
...
+import './dashboard.scss';
+
+type CtaButtonsProps = {
+    learnMoreLink: string;
+    learnMoreText: string;
+    actionLink: string;
+    actionText: React.ReactNode;
+    telemetryAction?: string;
+    actionButtonCallback?: () => void;
+};
```
michelengelen, line 20: I am ok with how it is built ... just one suggestion: We might not always have both buttons, so we probably should make the text and link properties optional and just return `null` in a guard for the buttons itself. WDYT?

components/admin_console/workspace-optimization/cta_buttons.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    const actionButton = (
+        <button
+            className='actionButton annnouncementBar__purchaseNow'
+            onClick={handleActionButtonClick}
+        >
+            {actionText}
+        </button>
+    );
+    return (
```
michelengelen, line 68:
```suggestion

    return (
```

components/admin_console/workspace-optimization/dashboard.scss
```diff
@@ -208,6 +220,21 @@
...
+                    margin: 15px 0;
+
+                    button {
+                        max-height: 32px;
+                        padding: 5px 20px !important;
+                        margin-top: 0;
+                        margin-right: 10px;
+                        border-radius: 4px !important;
+                        font-weight: 600;
+                    }
```
michelengelen, line 236: don't we have common button styles somewhere in the application that we could reuse here?

And are those `!important` really necessary? I would rather increase specificity and don't use `!important`

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -64,6 +70,15 @@ const AccordionItem = styled.div<{iconColor: string}>`
     }
 `;
 
+const successIcon = (
+    <div className='success'>
+        <SuccessIconSvg
```
michelengelen, line 75: is it possible to pass the `className` directly to the svg? that way we could avoid the extra div

components/common/accordion/accordion_item.tsx
```diff
@@ -75,9 +75,9 @@ const AccordionItem = ({
                         {data.extraContent}
                     </div>
                 )}
-                <div className='accordion-item-header__chevron'>
+                {data.items.length > 0 && <div className='accordion-item-header__chevron'>
```
michelengelen, line 78: there is a early return above for `items.length === 0`, so this check is unnecessary.

michelengelen: or is this in preparation for the "success" accordion item when there is nothing to do?

components/admin_console/workspace-optimization/overall-score.tsx
```diff
@@ -4,19 +4,22 @@ import React from 'react';
 import {FormattedMessage} from 'react-intl';
 
 import CircularChart from 'components/common/circular_chart/circular_chart';
+import WarningSvg from 'components/common/svg_images_components/warning_svg';
```
michelengelen, line 7: SH00T ... I have forgotten to tell you that in compass-icons I built SVG components for the icons as well in version 0.1.19 ... so if we upgrade the version you can just use those and remove the svgs from the common folder .... i am sorry i haven't told you earlier! If you have questions regarding this feel free to ping me!

components/trial_benefits_modal/trial_benefits_modal.tsx
michelengelen: good catch

sass/base/_css_variables.scss
```diff
@@ -6,8 +6,8 @@
     --button-color: #fff;
     --center-channel-bg: #fff;
     --center-channel-color: #3d3c40;
-    --dnd-indicator: #f74343;
-    --error-text: #fd5960;
+    --dnd-indicator: #d24b4e;
+    --error-text: #d24b4e;
```
michelengelen, line 10: do these values align with the design system?

michelengelen (changes requested):
awesome work ... I only have some small suggestions

components/admin_console/workspace-optimization/cta_buttons.tsx
```diff
@@ -0,0 +1,76 @@
...
+}: CtaButtonsProps): JSX.Element => {
+    const history = useHistory();
+
+    const redirectToConsolePage = (route: string) => {
+        history.push(route);
+
+        if (telemetryAction) {
+            trackEvent(
+                TELEMETRY_CATEGORIES.WORKSPACE_OPTIMIZATION_DASHBOARD,
+                'health_dashboard_action_' + telemetryAction,
```
mkraft, line 38:
```suggestion
                'workspace_dashboard_action_' + telemetryAction,
```

mkraft (changes requested):

components/admin_console/workspace-optimization/cta_buttons.tsx
michelengelen: good catch

components/admin_console/workspace-optimization/dashboard.tsx
pablovelezvidal: well, this is definitely temporal, we were hardcoding a number value before, so having a random number doesn't make any difference. I added a big TODO to remove it, but we need to define how are we going to calculate the value we pass to the overall score chart, including what is the threshold to show the red alert icon.

components/admin_console/workspace-optimization/cta_buttons.tsx
pablovelezvidal: ok, if there is that scenario where we don't have both buttons, then I guess we should make all the params that create buttons optionals. I went with this initial approach because I didn't see any example/scenario where we didn't use both buttons.

components/admin_console/workspace-optimization/dashboard.scss
pablovelezvidal: > don't we have common button styles somewhere in the application that we could reuse here?
yes, the buttons already have the majority of the css styles set as an inline class from a button-class that is used across the system console that looks like the one in the design. These styles are just to make them fit with the design.

> And are those !important really necessary? I would rather increase specificity and don't use !important
Yes, they are important, and about being more specific, this class is applied as a class module, so the next specificity level would be to put the style inline with the button tag, which I don't consider necessary. I know important is not good but in this case it is a necessary evil 👿

components/admin_console/workspace-optimization/dashboard.tsx
pablovelezvidal: yes, it is possible, but the div is the one that handles the styles for the background and the radius that makes it look as a circle, and for handling the state changes where the accordion item list lenght is 0 and we switch between normal icon circle or the light green one for the success icon

components/common/accordion/accordion_item.tsx
pablovelezvidal: or is this in preparation for the "success" accordion item when there is nothing to do?

> Yes! exactly, that way the item just looks like a card.

components/admin_console/workspace-optimization/overall-score.tsx
pablovelezvidal: no worries, I already created a card in the actions board to replace all the SVGs for the compass icons in a different pull request so we can move forward with this one and avoid another big merge conflict with the base branch.

https://community.mattermost.com/boards/workspace/7dqusmf8stn3dens5ckhyatbdo/boattb15z9fr4m85etwrdekkduh/v37ny5c5eafnk3dypexwzagip7a/c74odxpk9ppr18y6wt8prnt3boh

sass/base/_css_variables.scss
pablovelezvidal: yes! those were already added before.

pablovelezvidal:
@michelengelen @michaelgamble @mkraft gentle ping on this one :)

michelengelen (approved):

mkraft (approved):

mkraft:
@furqanmlk I've added you early to this PR, as requested, however this one will probably be merged into the main feature branch without QA review approval.

PR #9732, opened 2022-02-01
MM-41271 : Don't render sidebar channels when collapsed

M-ZubairAhmed:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41271

#### Related Pull Requests
None

#### Release Note

```release-note
Removes collapsed sidebar menu from the dom on sidebar collapse and vise versa on sidebar expand
```


components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -82,18 +82,44 @@ type Props = {
...
+    }
+
     isCollapsed = (props: Props) => {
         return props.isCategoryDragged || (props.isCategoryCollapsed && !this.props.isUnread && !props.isCurrentChannel);
     }
 
+    componentDidMount() {
+        const channelElement = this.getRef();
+        if (channelElement) {
+            channelElement.addEventListener('transitionend', this.handleTransitionEnd);
```
hmhealey, line 107: Is there some reason that we can't add this listener by passing an `onTransitionEnd` prop to the component?

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -215,6 +248,10 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
             );
         }
 
+        if (!this.state.show) {
```
hmhealey, line 251: Could you move this up closer to the top of the render? I know it won't make much of a difference to skip the whole setup of `wrappedComponent`, but it's nice to return early when rendering null doesn't depend on any of the other work done by this function.

hmhealey (changes requested):

esethna:
@M-ZubairAhmed @hmhealey Will this have a negative perceived performance impact when a user expands a channel category?

hmhealey:
No, it shouldn't. We expect this to be a decent performance improvement for anyone with a lot of channels in their sidebar who keeps any categories collapsed though.

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -173,11 +180,12 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
                     index={channelIndex}
                 >
                     {(provided, snapshot) => {
+                        const {onTransitionEnd: draggableTransitionEnd, ...draggableProps} = provided.draggableProps;
```
M-ZubairAhmed, line 183: draggableProps also injects onTransitionEnd, so we need to extract it from the props and pass on in our function

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -173,11 +180,12 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
...
                         return (
                             <li
                                 draggable='false'
                                 ref={this.setRef(provided.innerRef)}
-                                className={classNames('SidebarChannel', {
+                                className={classNames('SidebarChannel animating', {
```
M-ZubairAhmed, line 188: the word animating is misnomer in this context, if you see https://github.com/mattermost/mattermost-webapp/blob/9c8d80bfa9e03282325894b25c8438b1148c9760/sass/layout/_sidebar-left.scss#L1099-L1102
it just adds the transition property. better name should have been `animate` `transition` but i didnt wanted to change that in this pr

components/sidebar/sidebar_channel/sidebar_channel.tsx
jespino, line 183: This is a bit obscure way of setting the `draggableTransitionEnd` variable and at the same time extracting the draggableProps. IMO that are too many responsabilities for a single line of code.

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -173,11 +189,13 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
                     index={channelIndex}
                 >
                     {(provided, snapshot) => {
+                        const {handleTransitionEnd, restOfDraggableProps} = this.wrapTransitionEndHandler(provided.draggableProps);
+
```
M-ZubairAhmed, line 193: draggableProps also injects onTransitionEnd, so we need to extract it from the props via factory function

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -82,32 +82,23 @@ type Props = {
...
 export default class SidebarChannel extends React.PureComponent<Props, State> {
     static defaultProps = {
         isDraggable: true,
     }
 
-    isCollapsed = (props: Props) => {
-        return props.isCategoryDragged || (props.isCategoryCollapsed && !this.props.isUnread && !props.isCurrentChannel);
-    }
-
-    componentDidUpdate(prevProps: Props) {
```
M-ZubairAhmed, line 97: as explained in https://github.com/mattermost/mattermost-webapp/pull/9732/files#r798465348 animating class has to be present always

components/sidebar/sidebar_channel/sidebar_channel.tsx
M-ZubairAhmed: ✅ I removed this event listener and added the handler directly to the element

components/sidebar/sidebar_channel/sidebar_channel.tsx
M-ZubairAhmed: There were few UI gliches when we werent rendering the draggable components so i moved this check above https://github.com/mattermost/mattermost-webapp/pull/9732/files#diff-4c9f89f66f73c204771dc4e66a58fb0c67b346703ea3753e8c3337a99e9f9540R175

components/sidebar/sidebar_channel/sidebar_channel.tsx
M-ZubairAhmed: ✅  Updated as per request/discussion

components/sidebar/sidebar_channel/sidebar_channel.tsx
hmhealey: Okay, that should be fine. I assume the heaviest part of those extra channels in the sidebar is the actual `channelComponent`, so that should still be a performance improvement. I will have to update my queries I've been asking people to run since they'll still count those empty draggable items

components/sidebar/sidebar_channel/sidebar_channel.tsx
```diff
@@ -121,6 +112,37 @@ export default class SidebarChannel extends React.PureComponent<Props, State> {
...
+        }
+
+        // On fully collapsed
+        if (event.propertyName === 'height' && this.isCollapsed(this.props)) {
+            this.setState({show: false});
+        }
+
+        // On fully expanded
+        if (event.propertyName === 'height' && !this.isCollapsed(this.props)) {
+            this.setState({show: true});
```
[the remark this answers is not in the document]
M-ZubairAhmed: Yes i fixed it. I changed my approach to using the css 3 animation. This was done because
- browser can optimize for [animation steps](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Animations/Using_CSS_animations)
- animation start and end event listeners are typed correctly in react types, where as transition start doesn't exist in react dom types.

sass/layout/_sidebar-left.scss
```diff
@@ -1094,32 +1098,31 @@
...
         }
-    }
 
-    .SidebarChannel.animating {
-        transition-duration: 0.15s, 0.15s, 0.15s, 0.15s, 0.15s; /* collapse animation speed */
-        transition-property: height, padding, color, opacity, visibility;
-        transition-timing-function: ease, ease, ease, ease-in, step-start;
+        &.expanded {
+            animation-direction: normal;
+            animation-duration: 0.4s;
```
M-ZubairAhmed, line 1104: If you notice transition duration and animation duration are different: 0.3 and 0.4 respectively. Take a look at the fps comparison with both being dissimilar and being same.

- With dissimilar duration

image

- With same duration
image

The one with the same duration shows few frames were dropped at the starting of the menu collapsing. This reduced in frame rate resulted in choppy transition. On the other hand the one with different duration shows no frame drop resulting in near 60 fps. Height collapsing is a layout shift operation which requires more effort than opacity since browser has to paint change in height in every step. Change in duration gives browser enough room to perform the transition.

sass/layout/_sidebar-left.scss
```diff
@@ -1001,7 +1001,11 @@
         color: rgba(var(--sidebar-text-rgb), 0.6);
         list-style-type: none;
         opacity: 1;
+        transition-duration: 0.3s;
```
hmhealey, line 1004: This is twice as slow as the old transition. We should try this with the same 0.15s duration to see if it's still smooth at that rate, but if that doesn't work, we should confirm this with the UX team

M-ZubairAhmed: i have moved it back to 0.15s

hmhealey (approved):

devinbinnie (approved):

M-ZubairAhmed:
triggering e2e again after fixing related e2e tests

jgilliam17:
@M-ZubairAhmed 
Tested manually - no issues. Verified sidebar collapse and expand is working as expected. Also, new messages in collapsed categories are received as expected, as is the sorting to Unreads from collapsed categories.
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/148517-bfef81d-onprem-ent-webapp-pr-9732-mattermost/mm-ee-testbfef81d/mochawesome.html) has a failure that seems related to this PR MM-T2003 Can you please take a look? Thanks

M-ZubairAhmed:
@jgilliam17 it seems like that is the old report, i have fixed all the related e2es for this pr, here is the latest report https://automation-dashboard.vercel.app/cycles/1455f4cf-3731-4819-af4d-4e986cfca0a4

jgilliam17:
@M-ZubairAhmed Thanks for the latest report. Yes, everything looks good, no PR related failures. Good to merge.

jgilliam17 (approved):

PR #9735, opened 2022-02-02
MM-40148: decouples thread data from counts

koox00:
#### Summary

Currently fetch threads returns counts for threads as well.
This raises some performance concerns.

As of now we support `onlyThreads` and `onlyCounts` on the getThreadsForUser
api.
This commit utilizes these two query parameters to decouple counts from
actual thread data.
Introduces `getThreadCounts` action which only returns counts and
`getThreads` now will only return threads.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40148

#### Related Pull Requests

- Has server changes (https://github.com/mattermost/mattermost-server/pull/19349)

#### Release Note

```release-note
NONE
```


components/threading/global_threads/global_threads.tsx
```diff
@@ -108,15 +108,18 @@ const GlobalThreads = () => {
         return selectedThreadId && list.length === 1 && list.includes(selectedThreadId);
     };
 
+    const shouldLoadThreads = (counts && counts.total !== 0) && (isEmpty(threadIds) || isOnlySelectedThreadInList(threadIds));
+    const shouldLoadUnreadThreads = (counts && counts.total_unread_threads !== 0) && (isEmpty(unreadThreadIds) || isOnlySelectedThreadInList(unreadThreadIds));
```
larkox, line 112: Nit: Probably not much gain, but this is calculated on every render. If you move it to the effect body, it is calculated only when the effect is triggered. Depending on the cost of `isOnlySelectedThreadInList` may make more or less sense.

packages/mattermost-redux/src/actions/threads.ts
```diff
@@ -31,18 +31,36 @@ import {forceLogoutIfNecessary} from './helpers';
...
             forceLogoutIfNecessary(error, dispatch, getState);
             dispatch(logError(error));
             return {error};
         }
 
+        return {data};
+    };
+}
+
+export function getThreads(userId: string, teamId: string, {before = '', after = '', perPage = ThreadConstants.THREADS_CHUNK_SIZE, unread = false} = {}) {
```
larkox, line 50: I'm gonna steal this way of defining objects in the signature of the function :laughing:

packages/mattermost-redux/src/reducers/entities/threads/counts.ts
```diff
@@ -139,7 +139,7 @@ export function countsIncludingDirectReducer(state: ThreadsState['counts'] = {},
     case ChannelTypes.RECEIVED_CHANNEL_DELETED:
     case ChannelTypes.LEAVE_CHANNEL:
         return handleLeaveChannel(state, action, extra);
-    case ThreadTypes.RECEIVED_THREADS:
+    case ThreadTypes.RECEIVED_THREAD_COUNTS:
```
larkox, line 142: I may be missing something (or maybe it was handled before) but shouldn't be also some changes in the websockets? When we receive a new post, or things like that?

larkox (approved):
Nothing blocking, just one small nit and a question.

components/threading/global_threads/global_threads.tsx
koox00: I would think that `isOnlySelectedThreadInList` is not that costly.

It could be changed though to 
```js
         return selectedThreadId && list.length === 1 && list[0] === selectedThreadId; 
```

packages/mattermost-redux/src/actions/threads.ts
koox00: destructuring an object in function args is a neat trick :)

packages/mattermost-redux/src/reducers/entities/threads/counts.ts
koox00: ThreadUpdated ws event, fires a thread read action, which then should update counts,
but it seems that although it's updating `total_unread_{threads|mentions}` counts, it's not updating `total`!
Thanks @larkox, will fix in this PR!

PS: the drawback of that would be that infinite loading could assume that there are no threads anymore to fetch, that's the only place `total` is used.

koox00: > Thanks @larkox, will fix in this PR!

OK, it seems that ATM we have no means to derive a thread updated is a new thread or an existing one,
so we don't know whether to update the `total` count or not.

hmhealey (approved):
Looks good to me assuming that outstanding conversation about the websocket stuff isn't a blocker

ashishbhate (approved):

packages/mattermost-redux/src/reducers/entities/threads/counts.ts
koox00: @larkox, @hmhealey we'll need either to add a new event (ThreadCreated) or just add a prop like `new_thread: true` on `ThreadUpdated`, since this is already an issue, I want to add a new ticket for that and work on it ASAP.

Merge the other one first, since it will solve an issue in another PR as well https://github.com/mattermost/mattermost-server/pull/18492#issuecomment-1031258760

ashishbhate:
I've opened a new PR: https://github.com/mattermost/mattermost-webapp/pull/10064 with a some minor changes to bring it up to date with master. I've also pushed it to the mattermost org to make testing it easier with the server PR.

PR #9749, opened 2022-02-03
MM-41511 Fix repeated calls to makeGetCategory

hmhealey:
The main change here is to make it so that we're not calling `makeGetCategory` in the `mapStateToProps` since that creates a new instance of the selector each time which prevents it from doing any memoization. Normally, I would've moved just it out into the outer `makeMapStateToProps` to memoize it on a per-component basis, but I figured I'd pull it out into its own selector at the same time so that we only have a single instance of that selector instead of one-per-component.

This is also an effort to improve some performance issues we're seeing on community. The profiles we've collected on community contain a large number of calls which appear to be to the `clearChannelSelection` action passed to `SidebarChannelLink`, but I think there's something wrong with the source mapping there since the line number points to the `mapStateToProps` of the `SidebarChannelLink` instead. I'm hoping that this reduces the number of calls showing up there

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41511

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/sidebar/sidebar_channel/sidebar_channel_link/index.ts
+++ b/components/sidebar/sidebar_channel/sidebar_channel_link/index.ts
@@ -5,21 +5,21 @@ import {connect} from 'react-redux';
 import {bindActionCreators, Dispatch} from 'redux';
 
 import {getCurrentUserId, getMyChannelMemberships} from 'mattermost-redux/selectors/entities/common';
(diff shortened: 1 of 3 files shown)


jwilander (approved):

AshishDhama (approved):
LGTM!

pablovelezvidal (approved):
LGTM 👍

PR #9750, opened 2022-02-03
Fix failing unit tests in TutorialTip

hmhealey:
Looks like there was a conflict between two recent PRs in this file which let each build fine separately but caused a conflict when they were merged. Hopefully the resulting behaviour is correct since this'll fix the tests to check for that behaviour.

#### Release Note
```release-note
NONE
```


AshishDhama (approved):

AGMETEOR (approved):

AGMETEOR:
Thanks @hmhealey . Was just about to push a PR. @neallred there’s already a PR here :)

hmhealey:
Thanks for the fast reviews. Seems like everyone's working on the same piece of code today 😄

AshishDhama (approved):
I think this is also fine

PR #9645, opened 2022-01-12
MM-40894 - not able to download license in desktop app

pablovelezvidal:
#### Summary
This PR replaces several occurrences of the anchor property target from _new, which is an incorrect value to the property  to _blank that is the correct way of calling it. Also, replaces the _new from _self for the download invoice link that was not working on the desktop app.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40894

#### Screenshots

https://user-images.githubusercontent.com/10082627/149197201-fef845fb-557b-4304-bead-44d0b24d1636.mp4

#### Release Note
```release-note
NONE
```

PR #9751, opened 2022-02-03
Mm 40894 not able to download license from download history

pablovelezvidal:
#### Summary
This PR removes the value _new from anchor target property. _new is not a valid value, it just works to open a new tab window because it tries to open an existing tab named _new, since it fails to find it, it results in opening a new empty tab. Also, replacing it to _self so it works fine for downloading invoices from the desktop app which is the place where originally the problem was found.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40894

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/9645

#### Screenshots

https://user-images.githubusercontent.com/10082627/152431906-80358f2c-9f7e-4fd5-927d-f4a274fa035e.mp4

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/admin_console/billing/billing_history.test.tsx
+++ b/components/admin_console/billing/billing_history.test.tsx
@@ -1,0 +1,139 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+import {Provider} from 'react-redux';
+import configureStore from 'redux-mock-store';
+
+import thunk from 'redux-thunk';
+
+import {shallow} from 'enzyme';
+
+import {mountWithIntl} from 'tests/helpers/intl-test-helper';
+
+import BillingHistory from './billing_history';
+
+describe('components/admin_console/billing/billing_history', () => {
+    // required state to mount using the provider
+    const state = {
+        entities: {
+            general: {
+                license: {
+                    IsLicensed: 'true',
+                },
+            },
+            cloud: {
+                invoices: {
+                    in_1KNb3DI67GP2qpb4ueaJYBt8: {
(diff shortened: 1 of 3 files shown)


components/admin_console/billing/billing_history.tsx
```diff
@@ -151,75 +147,77 @@ const BillingHistory: React.FC<Props> = () => {
...
+                                    style='currency'
+                                    currency='USD'
+                                />
+                            </td>
+                            <td>
+                                {getPaymentStatus(invoice.status)}
+                            </td>
+                            <td className='BillingHistory__table-invoice'>
+                                <a
+                                    target='_self'
```
pablovelezvidal, line 211: This is the real change of this PR, setting back the target value to _self so it downloads the invoices in the desktop app

components/admin_console/billing/billing_history.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+                <BillingHistory/>
+            </Provider>,
+        );
+
+        const invoiceTableRows = wrapper.find('table.BillingHistory__table tr.BillingHistory__table-row');
+
+        expect(invoiceTableRows.length).toBe(2);
+    });
+
+    test('Billing history section download button has the target property set as _self so it works well in desktop app', () => {
```
pablovelezvidal, line 132: Added unit tests to cover this scenario from a future regression by again changing the target property value

stevemudie (approved):
It works!

Looks good @pablovelezvidal! I was able to download the invoice from both the Subscription and the Billing Information sections of the System Console using the Desktop app without any issues.

devinbinnie (approved):

neallred (approved):
lgtm!

PR #9752, opened 2022-02-03
[MM-41408] Remove autotour feature

JulienTant:
#### Summary
The result of the A/B testing shows that we have a better conversion rate without the auto tour. This PR drops the Autotour feature flag and implementation.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41408

#### Related Pull Requests
- Has server changes https://github.com/mattermost/mattermost-server/pull/19495

#### Screenshots
N/A

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/crt_tour/crt_list_tutorial_tip/crt_list_tutorial_tip.tsx
+++ b/components/crt_tour/crt_list_tutorial_tip/crt_list_tutorial_tip.tsx
@@ -14,15 +14,11 @@ import TutorialTip from 'components/tutorial/tutorial_tip';
 import {ShortcutKey, ShortcutKeyVariant} from 'components/shortcut_key';
 import {useMeasurePunchouts} from 'components/tutorial/tutorial_tip/hooks';
 
-type Props = {
-    autoTour: boolean;
-};
-
 export const UpShortcut = () => <ShortcutKey variant={ShortcutKeyVariant.TutorialTip}>{'UP'}</ShortcutKey>;
 
 export const DownShortcut = () => <ShortcutKey variant={ShortcutKeyVariant.TutorialTip}>{'DOWN'}</ShortcutKey>;
 
-const CRTListTutorialTip = ({autoTour}: Props) => {
+const CRTListTutorialTip = () => {
     const isMobileView = useSelector((state: GlobalState) => getIsMobileView(state));
     const dispatch = useDispatch();
     const onPrevNavigateTo = () => dispatch(openLhs());
@@ -67,7 +63,6 @@ const CRTListTutorialTip = ({autoTour}: Props) => {
             stopPropagation={true}
             screen={screen}
             overlayClass='tip-overlay--threads-list'
-            autoTour={autoTour}
             onPrevNavigateTo={onPrevNavigateTo}
             punchOut={useMeasurePunchouts(punchOutIds, [])}
         />
--- a/components/tutorial/tutorial_tip/tutorial_tip.tsx
+++ b/components/tutorial/tutorial_tip/tutorial_tip.tsx
@@ -24,10 +24,6 @@ const TutorialSteps = {
     [Preferences.CRT_TUTORIAL_STEP]: Constants.CrtTutorialSteps,
     [Preferences.CRT_THREAD_PANE_STEP]: Constants.CrtThreadPaneSteps,
 };
-const TutorialAutoTourStatus = {
-    [Preferences.TUTORIAL_STEP]: Preferences.TUTORIAL_STEP_AUTO_TOUR_STATUS,
-    [Preferences.CRT_TUTORIAL_STEP]: Preferences.CRT_TUTORIAL_AUTO_TOUR_STATUS,
-};
 
 type Preference = {
     user_id: string;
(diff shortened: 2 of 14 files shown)


pablovelezvidal (approved):
LGTM, thanks @JulienTant 👍

JulienTant:
Hey @AshishDhama - I've been told you're working on tooltips so you might be interested in reviewing this. We're dropping support for Auto tour (when the tooltip shows up without the user clicking on a green dot) as a result of our AB testing showing more engagement from the users who click by themselves.

components/tutorial/tutorial_tip/tutorial_tip.tsx
```diff
@@ -255,20 +212,14 @@ export default class TutorialTip extends React.PureComponent<Props, State> {
         if (this.props.extraFunc) {
             this.props.extraFunc();
         }
-
-        if (onNextNavigateTo && nextStep === true && autoTour) {
```
neallred, line 259: I think this is still needed by the CRT tour:
```
components/crt_tour/crt_list_tutorial_tip/crt_list_tutorial_tip.tsx
components/crt_tour/crt_welcome_tutorial_tip/crt_welcome_tutorial_tip.tsx
```

components/crt_tour/crt_list_tutorial_tip/crt_list_tutorial_tip.tsx
```diff
@@ -14,15 +14,11 @@ import TutorialTip from 'components/tutorial/tutorial_tip';
 import {ShortcutKey, ShortcutKeyVariant} from 'components/shortcut_key';
 import {useMeasurePunchouts} from 'components/tutorial/tutorial_tip/hooks';
 
-type Props = {
-    autoTour: boolean;
```
neallred, line 18: For context, the reason growth team wants to remove autotour is that it did not do as well in a conversion A/B test.

Something I learned today is that growth spike work on separate tour sets still desires to have autotour. Does that sound right, @AshishDhama ? If that's so, I think that's something else we'll want to come to agreement on in the Mattermost use case thread on tour tips.

neallred:
Aside from whether we still need auto tour for the other tour sets and aside from still needing the onNext... onPrev... functions regardless of the auto tour decision, this set of changes looks good to me!

components/tutorial/tutorial_tip/tutorial_tip.tsx
AshishDhama: @neallred Yeah right

AshishDhama:
@esethna @matthewbirtch I think we still need ``CRT tour`` to be auto, We have this check if the user dismisses the tour tip we don't show the next tour tip automatically instead silently move forward but if the user hits next or previous we show the tour automatically in any case. Also, should we keep the same behavior for the new onboarding tour in that case the first tip will always be auto, and then based on user action if they dismiss it then we move forward silently (don't show tour automatically) but keep showing the pulsating dot and on next or prev show automatically.
@neallred The new onboarding channel tour flow I am working on will replace the current one so whatever we learned may apply to this as well but not CRT. Tell me If I am wrong but in AB testing I think we had only two options either go through the tour one by one (on clicking pulsating dot manually) or keep showing tour tips automatically until they reach the last one and finish the tour so the only option for the user to get rid of tour is either to opt-out or finish. instead of soft dismiss (dismiss would have worked till the next refresh) like have put for CRT Tour now. We also didn’t have the previous button and close button as well which I added as part of the CRT tour.
cc// @marianunez @dontoisme

neallred:
>  in AB testing I think we had only two options either go through the tour one by one (on clicking pulsating dot manually) or keep showing tour tips automatically until they reach the last one and finish the tour so the only option for the user to get rid of tour is either to opt-out or finish. instead of soft dismiss (dismiss would have worked till the next refresh) like have put for CRT Tour now

That is correct.

> We also didn’t have the previous button and close button as well which I added as part of the CRT tour.

Also correct.

My two cents is that dismissal behavior (by clicking on the shadow overlay) should be consistent across all tours; I have no opinion on what exactly the behavior should be.

marianunez:
> or keep showing tour tips automatically until they reach the last one and finish the tour so the only option for the user to get rid of tour is either to opt-out or finish.

I believe this is not the case, the only thing that is automatic in this A/B test is to start the first tip - so instead of having the pulsating dot waiting for the user to click on it, it **automatically** shows the popup of the first tip but everything else they have to click through it. The result of the A/B test was that this change of having the user click to start the tour or showing the first popup automatically made no difference in the completing the tour.

neallred:
@marianunez I think they did each pop up for the user automatically when the user clicked next, or when they refreshed the page. I think that's what this removed `autoShow` method did: https://github.com/mattermost/mattermost-webapp/pull/9752/files#diff-096d7f8dbc4f08b6237a49f4df18c4b316067bcd4f12b331cd9db480eaf3386fL154-L175

marianunez:
For the purposes of this PR, the goal is to retire the A/B test so we need to remove the Feature Flag and if @AshishDhama and the user-case onboarding group still want to keep the behavior as it aligns better with the new tour - no issues with that. The goal is just to complete the A/B @JulienTant

AshishDhama:
> The result of the A/B test was that this change of having the user click to start the tour or showing the first popup automatically made no difference in completing the tour.

@marianunez Thanks, knowing the results of AB test helps.

matthewbirtch:
> > The result of the A/B test was that this change of having the user click to start the tour or showing the first popup automatically made no difference in completing the tour.
> 
> @marianunez Thanks, knowing the results of AB test helps.

@AshishDhama just to confirm, we still want the first tour point to display automatically when the user has chosen 'Take a tour of channels' from the new task list, so yes, we do still need this - and we do still want to keep it for CRT tour

JulienTant:
Closing this PR until the growth spike is over

PR #9754, opened 2022-02-04
Fix and update channel switch performance telemetry tracking

jwilander:
#### Summary
I noticed two things in the telemetry data we were receiving:

1. The `channel_switch` events were no longer being reported.
2. The old `channel_switch` and existing `team_switch` data was reporting seemed low at around 200ms.

It was easy to spot that 1) was broken because of a mismatched name when marking the `channel_switch` event.

For 2) we were measuring when ChannelView updated which didn't mean that posts had been loaded and rendered. I moved that measuring to PostList and ensured that the posts had been loaded/mounted before marking the switch as complete. I also introduced the idea of `fresh` switches into the telemetry which indicates whether the posts for that channel had been loaded previously or not. This still isn't a perfect measurement but it will be more accurate to what the user perceives as the performance and gives us a bit more insight than we had before.

I also started console logging the performance events in developer mode and removed one dead line of code.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/post_view/post_list/post_list.jsx
+++ b/components/post_view/post_list/post_list.jsx
@@ -8,11 +8,34 @@ import LoadingScreen from 'components/loading_screen';
 import {PostRequestTypes} from 'utils/constants';
 
 import {getOldestPostId, getLatestPostId} from 'utils/post_utils';
+import {clearMarks, mark, measure, trackEvent} from 'actions/telemetry_actions.jsx';
 
 import VirtPostList from 'components/post_view/post_list_virtualized/post_list_virtualized';
 
 const MAX_NUMBER_OF_AUTO_RETRIES = 3;
 export const MAX_EXTRA_PAGES_LOADED = 10;
+
+// Measures the time between channel or team switch started and the post list component rendering posts.
+// Set "fresh" to true when the posts have not been loaded before.
+function markAndMeasureChannelSwitchEnd(fresh = false) {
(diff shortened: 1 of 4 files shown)


components/post_view/post_list/post_list.jsx
```diff
@@ -8,12 +8,35 @@ import LoadingScreen from 'components/loading_screen';
...
+import {clearMarks, mark, measure, trackEvent} from 'actions/telemetry_actions.jsx';
 
 import VirtPostList from 'components/post_view/post_list_virtualized/post_list_virtualized';
 
 const MAX_NUMBER_OF_AUTO_RETRIES = 3;
 export const MAX_EXTRA_PAGES_LOADED = 10;
 
+// Measures the time between channel or team switch started and the post list component rendering posts.
+// Set "fresh" to true when the posts have not been loaded before.
+function markAndMeasureChannelSwitchEnd(fresh = false) {
```
hmhealey, line 20: Note that fresh may not always be exactly what we expect here since posts could have been preloaded in the background if we're switching to an unread channel

hmhealey (approved):

M-ZubairAhmed (approved):
This looks good, excited for the telemetry

components/post_view/post_list/post_list.jsx
jwilander: Got it. Good to consider when analyzing the data.

PR #9758, opened 2022-02-04
Enable performance telemetry tracking for production deployments

jwilander:
#### Summary
Per discussion here https://community.mattermost.com/core/pl/seageg6dhidkmmx9rz9urgk9mw.

#### Release Note
```release-note
Enable performance telemetry tracking for production deployments not running in developer mode. Disabled when telemetry is toggled off.
```


actions/telemetry_actions.jsx
```diff
@@ -14,6 +17,15 @@ const SUPPORTS_MEASURE_METHODS = isSupported([
     performance.clearMeasures,
 ]);
 
+export function isTelemetryEnabled(store = reduxStore) {
```
jwilander, line 20: Not a fan of passing the store in but couldn't find a cleaner method that would allow for unit testing.

jwilander: Good call, that's much cleaner. I'll update the PR.

jwilander:
@hmhealey PTAL

actions/telemetry_actions.jsx
```diff
@@ -14,6 +17,15 @@ const SUPPORTS_MEASURE_METHODS = isSupported([
     performance.clearMeasures,
 ]);
 
+export function isTelemetryEnabled(state = store.getState()) {
```
hmhealey, line 20: Maybe remove the default value from here? If we're only calling `shouldTrackPerformance` without an argument, I think it'd be best to have that be the only place its done to keep people using the other selectors correctly.

hmhealey (approved):
One small suggestion, but otherwise looks good

M-ZubairAhmed (approved):

cwarnermm:
@jwilander - Are [telemetry documentation](https://docs.mattermost.com/manage/telemetry.html) updates needed as part of this PR?

jwilander:
@cwarnermm nope, this falls into the same categories called out in the documentation so all good 👍

PR #9760, opened 2022-02-06
removed leading whitespace causing header description bug

scottaudet:
#### Summary

  Removed header description bug where leading whitespace caused markdown code block

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/19213

#### Related Pull Requests

  N/A

#### Screenshots

|  before  |  after  |
|----|----|
| Screen Shot 2022-02-06 at 14 48 16|Screen Shot 2022-02-06 at 14 47 19|

#### Release Note

```
NONE
```
```release-note

```

### DIFF
--- a/components/edit_channel_header_modal/edit_channel_header_modal.tsx
+++ b/components/edit_channel_header_modal/edit_channel_header_modal.tsx
@@ -100,21 +100,24 @@ export default class EditChannelHeaderModal extends React.PureComponent<Props, S
     }
 
     public handleSave = async (): Promise<void> => {
-        const {header} = this.state;
-        if (header === this.props.channel.header) {
-            this.hideModal();
-            return;
-        }
+        const header = this.cleanHeaderText();
 
-        this.setState({saving: true});
-
-        const {channel, actions} = this.props;
-        const {error} = await actions.patchChannel(channel.id!, {header});
-        if (error) {
-            this.setState({serverError: error, saving: false});
-        } else {
-            this.hideModal();
+        if (header != this.props.channel.header) {
+            this.setState({saving: true});
+            const {channel, actions} = this.props;
+            const {error} = await actions.patchChannel(channel.id!, {header});
+            if (error)
+                this.setState({serverError: error, saving: false});
+                console.log(error)
         }
+        console.log(this.state)
+        this.hideModal()
+    }
+
+    private cleanHeaderText = () => {
+        let {header} = this.state
+        header = header?.trim()
+        return header
     }
 
     private hideModal = (): void => {


components/edit_channel_header_modal/edit_channel_header_modal.tsx
```diff
@@ -100,21 +100,24 @@ export default class EditChannelHeaderModal extends React.PureComponent<Props, S
...
-        const {error} = await actions.patchChannel(channel.id!, {header});
-        if (error) {
-            this.setState({serverError: error, saving: false});
-        } else {
-            this.hideModal();
-        }
+    private cleanHeaderText = () => {
+        let {header} = this.state
+        header = header?.trim()
+        return header
```
M-ZubairAhmed, line 120: Do we need a separate function ? it looks simple enough to me to be written in a single line

M-ZubairAhmed, line 120: Do we need a separate function ? it looks simple enough to me to be written in a single line

M-ZubairAhmed, line 120: Do we need a separate function ? it looks simple enough to me to be written in a single line

M-ZubairAhmed (changes requested):

scottaudet:
Thanks @hmhealey. The header is given to the UI in a markdown component so the spaces trigger the code block functionality of a .md file. Realistically, a user shouldn't be able to make an all space header and any leading/trailing spaces don't change the comprehension of the name. I'm glad that this fix can help! I fixed the linting issues so hopefully this CI passes.

components/edit_channel_header_modal/edit_channel_header_modal.tsx
M-ZubairAhmed: @scottaudet i am curious what sync issues have you come across with that. May be we can look out for that behaviour when writing code in future.
But wasnt able to face issue with sync without the function
```js
        const header = this.state.header?.trim() ?? '';
```

scottaudet: I think I forget to have web pack running when I was testing that change so it wasn't updating in local host 😣 Looks good for me - checking in

M-ZubairAhmed (approved):

hmhealey (approved):
Thanks for clearing that up. This looks good to me now

jgilliam17:
Thanks @scottaudet 
When adding only one line code block to header description, outline is added, but it does not show the whole code block on hover
@hmhealey would this be out of scope here? Something similar is present on daily, wasn't sure if it would be fixed here or needs to be filed separately 
Screen Shot 2022-02-23 at 1 43 12 PM

hmhealey:
Yeah, that's out of scope. That's an existing issue around us simply allowing a lot of Markdown stuff in the header that probabaly doesn't make sense, but we've never bothered fixing it since it's a lot of work.

jgilliam17:
Thanks @scottaudet 
Tested, looks good to merge 
- Verified creating channel header description with leading white spaces renders as expected, with whitespaces removed.
E2E failed to trigger last time, re-running again before approving.

jgilliam17:
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/155528-a56f73e-onprem-ent-webapp-pr-9760-mattermost/mm-ee-testa56f73e/mochawesome.html) looks good, no PR related failures.

jgilliam17 (approved):

PR #9766, opened 2022-02-07
Cypress/e2e/fixed cypress test cases

furqanmlk:
#### Summary
Fixed mark as unread test case
Fixed mark mention as unread test case
Fixed tutorial guided tour test case 

#### Screenshots

image

#### Release Note
```release-note
NONE
```


saturninoabril:
I wonder what's the main difference of similar tests at e2e/cypress/integration/onboarding/tutorial_guided_tour_spec.js. If none, I'd suggest to just combine. Otherwise, I suggest to either modify MM-T4148 to add new step or create a separate test test, since duplicated test case ID (MM-T4148) will cause error when saving report to Zephyr. See https://mattermost.atlassian.net/browse/MM-41726 for ticket to fix the tests. cc @AGMETEOR

Phrynobatrachus (approved):

saturninoabril (approved):
Thanks @furqanmlk! Great to have new test cases in Zephyr!

Non-blocking: I forgot to comment last time but I'd recommend to move the other describe block into its own spec file or move all tests into one describe block (but not nested). Reason being is that whenever the first one failed and exited, the other one will never get a chance to be executed.

furqanmlk:
> Thanks @furqanmlk! Great to have new test cases in Zephyr!
> 
> 
> 
> Non-blocking: I forgot to comment last time but I'd recommend to move the other describe block into its own spec file or move all tests into one describe block (but not nested). Reason being is that whenever the first one failed and exited, the other one will never get a chance to be executed.

Ah thanks for the reminder. 
I completely forgot about each spec should one describe block. 
Let me fix it.

AshishDhama:
@furqanmlk we have updated the whole tour flow this I think is no longer true, could you please update your branch.

AshishDhama:
@furqanmlk have updated onboarding tour e2e here https://github.com/mattermost/mattermost-webapp/pull/9904, as onboarding tour is same for admin and non admin so have removed admin case. I have used the test Id created by you

PR #9768, opened 2022-02-07
MM-41176 Don't send Rudder requests when blocked

hmhealey:
Currently, we repeatedly try to send telemetry to Rudder even if the ad blocker has blocked the initial setup. This leads to a ton of error spam in the dev tools when trying to debug things on community.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41176
Closes https://github.com/mattermost/mattermost-server/issues/19380

#### Related Pull Requests
https://github.com/mattermost/focalboard/pull/2285

#### Release Note
```release-note
Stop attempting to send telemetry events when blocked by an ad blocker
```


components/root/root.jsx
```diff
@@ -214,6 +214,10 @@ export default class Root extends React.PureComponent {
                 },
                 anonymousId: '00000000000000000000000000',
             });
+
+            rudderAnalytics.ready(() => {
+                Client4.setTelemetryHandler(new RudderTelemetryHandler());
```
crspeller, line 219: Do you also have to delete the invocation on 186?

crspeller (changes requested):

components/root/root.jsx
hmhealey: Good catch. That's what I get for doing a `git add -p` to skip some debugging code

crspeller (approved):
Yay!

hmhealey:
@jespino Do you have a chance to look at this in the next day or two? If not, I can reassign

jespino:
@hmhealey sure, I'll take a look today, sorry for the delay

PR #9769, opened 2022-02-07
MM-41637 Remove API telemetry on most GET requests

hmhealey:
We make a ton of requests during page load which is sure to partly explain why first load takes so long. In particular, we send a massive number of requests to the telemetry (which get blocked by an ad blocker) which is likely to block and slow down the initial load of the web page since the browser can only run a certain number of requests in parallel.

image

A lot of that telemetry could prove useful, but I think these particular ones aren't useful at the moment because of just how many are sent. If you look at the video clip below, you'll see tons of events sent with the ones I've removed here (the type is shown in the bottom right under the `properties` field in the JSON):
https://user-images.githubusercontent.com/3277310/152891392-1ac5bd45-bcf0-4e6e-b154-bb898678eb8a.mov

I also looked into potentially batching these, but I can't find anything in the Rudder API for batching requests.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41637

#### Release Note
```release-note
Removed extra telemetry events tracked during page load
```


jasonblais:
I don't have major concerns, but I'm not sure if that will impact our DAU & MAU tracking.

@sri-byte Are you open to taking a look at how our [Cloud DAU](https://mattermost.looker.com/explore/data_warehouse_l/user_events_telemetry?qid=pOI8WHcB2HSp35wpaZiDgo&origin_space=221&toggle=vis) is measured? We use the `user count` field from Events telemetry, but I'm not sure what the logic is to generate the `user count` field. Would like to know if removing these API calls will impact the DAU measurement.

sri-byte:
@jasonblais From what I see, removing these wouldn't affect `user count`, as we count the unique user id's. 
@hmhealey Is it possible I can get the link to the list shown in the video?

jasonblais:
@sri-byte Thanks! Just to confirm, if these API calls aren't tracked, would the count of unique user ID's be lower on a given day, or still the same?

sri-byte:
when we create the user_events_telemetry table, we are dropping these particular events, I'm not sure why this filter is in place, but heres the code ```
COALESCE("ANALYTICS".EVENTS.SEGMENT_WEBAPP_EVENTS.type, "ANALYTICS".EVENTS.SEGMENT_WEBAPP_EVENTS.event)      IN ('api_channel_get',
                                            'api_channel_get_by_name_and_teamname',
                                            'api_channels_join_direct',
                                            'api_posts_get_after',
                                            'api_posts_get_before',
                                            'api_profiles_get',
                                            'api_profiles_get_by_ids',
                                            'api_profiles_get_by_usernames',
                                            'api_profiles_get_in_channel',
                                            'api_profiles_get_in_group_channels',
                                            'api_profiles_get_in_team',
                                            'api_profiles_get_not_in_channel',
                                            'api_profiles_get_not_in_team',
                                            'api_profiles_get_without_team',
                                            'channel_switch',
                                            'page_load',
                                            'lhs_dm_gm_count',
                                            'ui_channel_selected',
                                            'ui_channel_selected_v2',
                                            'ui_direct_channel_x_button_clicked',
                                            'application_backgrounded',
                                            'application_installed',
                                            'application_opened',
                                            'application_updated') 

```

**Comment by sri-byte** [comment]
I'm going through the list Harrison is removing in this PR, will get back shortly.

**Comment by jasonblais** [comment]
@sri-byte great, thank you for confirming. Removing these API telemetry shouldn't be a concern then. I'll approve the PR.

**Review by jasonblais — APPROVED** [review]
Won't impact daily active usage data, per Sri's comment below:

> ...there are other important events we track (I can see quite a few this.trackEvent('api', 'api_users_create', this.trackEvent('api', 'api_users_login')) in the client4.ts file), outside the ones Harrison is removing and we should be able to accurately do our DAU measurement if these events get triggered.

**Comment by sri-byte** [comment]
Its safe to say we can remove these, 
```
api_get_known_users
api_profiles_get
api_profiles_get_by_ids
api_profiles_get_by_usernames
api_profiles_get_in_team
api_profiles_get_not_in_team
api_profiles_get_without_team
api_profiles_get_in_group_channels
api_profiles_get_not_in_channel
api_profiles_get_in_channel
api_channel_get_by_name_and_teamName
api_channel_get
api_posts_get_before
api_posts_get_after
```
Is it possible we can keep these @hmhealey 
```
api_groups_get_not_associated_to_team
api_groups_get_not_associated_to_channel
api_groups_get_associated_to_team
api_groups_get_associated_to_channel
api_teams_get_team_by_name
```

hmhealey:
I'll take a look to see how often those are fired. The ones you've listed as safe to remove seem like the more impactful ones, but I'll double check to see how much of a difference the others make

M-ZubairAhmed (approved):

hmhealey:
@sri-byte Yep, I can re-add those ones. None of them are involved in that request spam I'm seeing on first load of community.

catalintomai:
@hmhealey - could we port/cherry-pick this change to earlier MM versions? We've discovered that most if not all of these events you've removed are responsible for a significant portion of our daily Rudderstack cost (about 98%). Removing them only from the next MM version might be too slow and cost a lot in the long-run. Thanks. cc: @jasonblais , @sri-byte

hmhealey:
Yeah, we could probably do that assuming the release team is cool with it. Which versions would we be wanting to backport it to? Just the ESR or more than that?

catalintomai:
> Yeah, we could probably do that assuming the release team is cool with it. Which versions would we be wanting to backport it to? Just the ESR or more than that?
> 
> Also, that's funny that Mattermod managed to find that "slash command" in your comment 😛

@hmhealey , thanks for getting back, I'll talk to Amy about it, it would be great if this could be pushed to earlier versions (no guarantee that customers would update, but everything helps). I think porting this just to ESR would be enough. 
And, yeah, I was surprised as well about the cherry pick command, seems a somewhat crude approach = )

PR #9788, opened 2022-02-10
[MM-41398] Fix Selecting a Status from Recent Status Uses Original Expiration Time

mylonsuren:
#### Summary

Fixes issue where selecting a status from Recent Status used original expiration time. Also fixes consistency between selecting the status that was used previously vs. other recent statuses/suggestions.

#### Ticket Link

Fixes [MM-41398](https://mattermost.atlassian.net/browse/MM-41398)

#### Screenshots

|  Before  |  After  |
|----|----|
|image|image|

#### Release Note

```release-note
NONE
```


mkraft (approved):

BenCookie95 (approved):
Is there a test we could add for this?

mylonsuren:
> Is there a test we could add for this?

I don't believe this fix is unit-testable, but I've added two e2e tests to cover this case.

@furqanmlk Can you please take a look at the e2e tests I've added and let me know if there's anything I should fix.

furqanmlk (approved):
Working as expected
**Before**

image

**After**

image

PR #9795, opened 2022-02-11
[MM-33679] Fix Accessibility for Custom Status and Modals

mylonsuren:
#### Summary

Adds voiceover accessibility support for custom status and modal headers.

#### Ticket Link

Fixes [MM-33679](https://mattermost.atlassian.net/browse/MM-33679)

#### Screenshots

| before | after |
|----|----|
| imageimage | Screen Shot 2022-02-10 at 6 55 36 PMScreen Shot 2022-02-10 at 6 58 24 PM|

#### Release Note

```release-note
NONE
```


mkraft (approved):

components/generic_modal.tsx
```diff
@@ -130,7 +131,8 @@ export default class GenericModal extends React.PureComponent<Props, State> {
                 enforceFocus={this.props.enforceFocus}
                 restoreFocus={true}
                 role='dialog'
-                aria-labelledby='genericModalLabel'
+                aria-label={this.props.ariaLabel}
+                aria-labelledby={this.props.ariaLabel ? undefined : 'genericModalLabel'}
```
BenCookie95, line 135: Could do `this.props.ariaLabel || 'genericModalLabel'`, 0/5

BenCookie95 (approved):
LGTM! Nice work

components/generic_modal.tsx
mylonsuren: I would've preferred do to something like that as well, but it guarantees that `aria-labelledby` is always defined which would take priority over `aria-label`. The issue stems from `aria-labelledby` not being able to read the value from the `FormattedMessage` component that is used for the header, so I just give priority to `aria-label` for components that use it to resolve the issue.

furqanmlk (approved):
Working as expected. Test case has been created [MM-T4654](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/MM-T4654)

image

PR #9805, opened 2022-02-13
MM-40306 : Investigate removing XRegExp

M-ZubairAhmed:
#### Summary
- [X] Replace `XRegExp.cache` in `components/suggestion/at_mention_provider/at_mention_provider.jsx`
- [X] Replace `XRegExp.cache` in `utils/text_formatting.tsx`
- [x] Replace `XRegExp.replace` in `utils/text_formatting.tsx`
- [x] Add updated `mattermost-marked` with removed `XRegExp` library to package.json

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40306

#### Related Pull Requests
https://github.com/mattermost/marked/pull/21

#### Screenshots
N/A

#### Release Note

```release-note
Removed and replaced "XRegExp" dependency with native RegEx.
```


utils/text_formatting.tsx
```diff
@@ -343,25 +337,8 @@ export function sanitizeHtml(text: string): string {
...
-  emailValidRestrictedCharacters +
-  '.]+\\")@';
-const reEmail = XRegExp.cache(
-    '(^|[^\\pL\\d])(' +
-    emailStartPattern +
-    '[\\pL\\d.\\-]+[.]\\pL{2,4}(?=$|[^\\p{L}]))',
-    'g',
-);
+// eslint-disable-next-line no-useless-escape
+const emailRegex = /(^|[^\p{L}\d])((?:[\p{L}\p{Nd}!#$%&'*+-/=?^_`\{|\}~](?:[\p{L}\p{Nd}!#$%&'*+-/=?^_`\{|\}~]|.(?!.|@))*|"[\p{L}\p{Nd}!#$%&'*+-/=?^_`\{|\}~\s(),:;<>@[].\]+")@[\p{L}\d.-]+[.]\p{L}{2,4}(?=$|[^\p{L}]))/gu;
```
M-ZubairAhmed, line 341: @hmhealey this regex is failing for one case where `email.@domain.com` could you please check it out

utils/text_formatting.tsx
```diff
@@ -343,25 +337,8 @@ export function sanitizeHtml(text: string): string {
...
-  emailValidCharacters +
-  ']|\\.(?!\\.|@))*|\\"[' +
-  emailValidRestrictedCharacters +
-  '.]+\\")@';
-const reEmail = XRegExp.cache(
-    '(^|[^\\pL\\d])(' +
-    emailStartPattern +
-    '[\\pL\\d.\\-]+[.]\\pL{2,4}(?=$|[^\\p{L}]))',
-    'g',
-);
```
M-ZubairAhmed, line 364: I moved to direct regex instead of string, because of so many escaping and with `u` flag it needs even more escaping. So just a regular regex worked

utils/text_formatting.tsx
```diff
@@ -343,25 +337,8 @@ export function sanitizeHtml(text: string): string {
...
+// eslint-disable-next-line no-useless-escape
```
M-ZubairAhmed, line 340: This eslint error as it complains for unnecessary escaping of `{` and `}` but with `u` flag it is required.

utils/text_formatting.tsx
michelengelen: Wow, that regex looks extremely complex!

M-ZubairAhmed: yes thats because of so many escapes need in unicode flag https://github.com/mattermost/mattermost-webapp/pull/9805#discussion_r805305369

michelengelen: @M-ZubairAhmed this one is not in any way less complex, but it checks (fails) for the dot problem you mentioned: https://regex101.com/r/sI6yF5/1

I have used that before and it works like a charm.

M-ZubairAhmed: @michelengelen the one you mentioned fails for `email.@domain.com` i wonder if we can modify the regex to catch that also. I am a noob in regex

components/suggestion/at_mention_provider/at_mention_provider.jsx
```diff
@@ -15,10 +13,11 @@ import Provider from '../provider.jsx';
 import AtMentionSuggestion from './at_mention_suggestion.jsx';
 
 const profilesInChannelOptions = {active: true};
+const regexForAtMention = /(?:^|\W)@([\p{L}\d_. -]*)$/ui;
```
larkox, line 16: I don't know enough about regex or xregexp to be sure if this is a intended change, but after the `\d`, I see two main differences with what was being passed to xregexp:
- We are missing a `\` (or `\\`)
- The order is different

utils/text_formatting.tsx
larkox: Before changing to a simpler regex, it would be nice to watch over the history of changes to know why this become like this. Maybe there are some corner cases that we wanted to address with this, and we would need to make sure any simplification still cover those cases.

utils/text_formatting.tsx
```diff
@@ -636,6 +612,8 @@ function highlightCurrentMentions(
     return output;
 }
 
+const hashtagRegex = /(^|\W)(#\p{L}[\p{L}\d_.-]*[\p{L}\d])/gu;
```
larkox, line 615: Same missing `\\` and order change as before.

larkox (approved):
Some minor comments, but LGTM.

components/suggestion/at_mention_provider/at_mention_provider.jsx
M-ZubairAhmed: - `\\` is need for escaping when we have regex as string
- yes the order doesn't matter for this character class `[\p{L}\d_. -]` because it will try to match a single character in that set. Reason for putting `-` in the end was `\d-` tries to create a range from `\d` to `_` and it becomes invalid regex. This wasnt before as now with addition of `u` flag regex becomes a bit strict (read pain)

utils/text_formatting.tsx
M-ZubairAhmed: same as https://github.com/mattermost/mattermost-webapp/pull/9805#discussion_r816675976

M-ZubairAhmed:
@larkox yes i will look at the blame and check. Also i am talking with @michelengelen his suggestion is to replace the email regex with

utils/text_formatting.tsx
michelengelen: @larkox absolutely ... It is just that the regex that has been used is not very performant and the first capture group does not do anything at all

M-ZubairAhmed: Good news i was able to fix the existing regex to pass the tests

utils/text_formatting.tsx
M-ZubairAhmed: https://regex101.com/r/Adx2SD/1

M-ZubairAhmed:
📂 E2E test report 📂 

https://automation-dashboard.vercel.app/cycles/d06a8aad-b9a3-449e-8380-57829b359bc6

hmhealey (approved):
Looks good to me. I was a bit worried about the bigger change to the email regex, but this code is heavily unit tested, so it should be fine

utils/text_formatting.tsx
hmhealey: I'm pretty sure that first group was supposed to be a word boundary, but I couldn't use `\b` because a URL could start with something that wasn't considered to be a word character. It might be that Unicode letters aren't recognized as word characters so `\b` doesn't work for them.

If emails consistently started with a non-word character, then we could use `\B` like you'll see in the emoji and at mention parsing, but since it can start with either word or non-word characters, it means we have to do some uglier stuff like in this regex

jgilliam17 (approved):
Thanks @M-ZubairAhmed 
Tested, looks good to merge.
- Verified mentions, emails and hashtags are rendering correctly. 
- [E2E report ](https://mattermost-cypress-report.s3.amazonaws.com/157824-a91b932-onprem-ent-webapp-pr-9805-mattermost/mm-ee-testa91b932/mochawesome.html)looks good, no PR related failures.

PR #9808, opened 2022-02-14
MM-41805: Improve selectors performance of "getDisplayedChannels" called when typing in quick switcher

M-ZubairAhmed:
#### Summary
- Improved selector performance of `getDisplayedChannels` from being called 315 times to 7 for a given search key in Quick switcher
- Changed from importing entire `utils` module to multiple imports in `channel_header`
- Removed explicitly given types in few selector files

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41805

#### Related Pull Requests
None

#### Screenshots
Shared in [Developers: Webapp channel on community](https://community-daily.mattermost.com/core/pl/xmaaeyjk1jypurna1kani9nage)

#### Release Note

```release-note
NONE
```


packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -189,9 +189,9 @@ export const getCurrentChannel: (state: GlobalState) => Channel = createSelector
     'getCurrentChannel',
     getAllChannels,
     getCurrentChannelId,
-    (state: GlobalState): UsersState => state.entities.users,
+    (state: GlobalState) => state.entities.users,
```
M-ZubairAhmed, line 192: Explicitly overriding types isnt required since the related store is typed

components/channel_header/index.js
```diff
@@ -47,12 +47,14 @@ import {isFileAttachmentsEnabled} from 'utils/file_utils';
 
 import ChannelHeader from './channel_header';
 
+const EMPTY_CHANNEL = {};
+
```
michelengelen, line 51: that change is surely not needed as long as the variable is only used once

components/channel_header/index.js
```diff
@@ -47,12 +47,14 @@ import {isFileAttachmentsEnabled} from 'utils/file_utils';
...
 function makeMapStateToProps() {
     const doGetProfilesInChannel = makeGetProfilesInChannel();
     const getCustomStatus = makeGetCustomStatus();
 
     return function mapStateToProps(state) {
-        const channel = getCurrentChannel(state) || {};
+        const channel = getCurrentChannel(state) || EMPTY_CHANNEL;
```
michelengelen, line 57: regarding my comment above this is a unneeded change

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -564,7 +566,7 @@ export default class SidebarChannelList extends React.PureComponent<Props, State
                         renderTrackVertical={renderTrackVertical}
                         renderView={renderView}
                         onScroll={this.onScroll}
-                        style={{position: 'absolute'}}
+                        style={scrollbarStyles}
```
michelengelen, line 569:
```suggestion
                        style={({position: 'absolute'} as CSSProperties)}
```

selectors/views/channel_sidebar.ts
```diff
@@ -27,9 +27,14 @@ import {memoizeResult} from 'mattermost-redux/utils/helpers';
...
-export function isUnreadFilterEnabled(state: GlobalState): boolean {
-    return state.views.channelSidebar.unreadFilterEnabled && !shouldShowUnreadsCategory(state);
-}
+export const isUnreadFilterEnabled = createSelector(
+    'isUnreadFilterEnabled',
+    (state: GlobalState) => state.views.channelSidebar.unreadFilterEnabled,
+    shouldShowUnreadsCategory,
+    (isUnreadFilterEnabled, showUnreadsCategory) => {
+        return isUnreadFilterEnabled && !showUnreadsCategory;
+    },
```
michelengelen, line 36:
```suggestion
    (isUnreadFilterEnabled, showUnreadsCategory) => isUnreadFilterEnabled && !showUnreadsCategory,
```

selectors/views/channel_sidebar.ts
```diff
@@ -211,28 +216,32 @@ function maxDefined(a: number, b?: number) {
     return typeof b === 'undefined' ? a : Math.max(a, b);
 }
 
+function concatChannels(channelsA: Channel[], channelsB: Channel[]) {
+    return [...channelsA, ...channelsB];
+}
+
+const memoizedConcatChannels = memoizeOne(concatChannels);
```
michelengelen, line 223: 👍🏼 🚀

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
M-ZubairAhmed: So the reason we changed it because on every render a new object will get created. The component will do a shallow compare with Object.is for that prop. Now because a new object was created on render its reference is changed and previous reference is lost. React will trigger a render again because of that. Globally declaring the object means object is only created once at the start of the execution and it will remain sane through out.

components/channel_header/index.js
M-ZubairAhmed: see https://github.com/mattermost/mattermost-webapp/pull/9808#discussion_r807094640
so until we get the request back from server about the channel or until store is populated we dont have the reference of the object `EMPTY_CHANNEL` to change

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
```diff
@@ -59,6 +59,8 @@ export function renderThumbVertical(props: any) {
         />);
 }
 
+const scrollbarStyles: CSSProperties = {position: 'absolute'};
+
```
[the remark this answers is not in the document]
M-ZubairAhmed: related https://github.com/mattermost/mattermost-webapp/pull/9808#discussion_r807094640

packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -265,13 +264,17 @@ export const isCurrentChannelMuted: (state: GlobalState) => boolean = createSele
 export const isCurrentChannelArchived: (state: GlobalState) => boolean = createSelector(
     'isCurrentChannelArchived',
     getCurrentChannel,
-    (channel: Channel): boolean => channel.delete_at !== 0,
+    (channel) => {
+        return channel.delete_at !== 0;
+    },
```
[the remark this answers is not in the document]
M-ZubairAhmed: ✅  Updated as per request/discussion

packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -265,13 +264,17 @@ export const isCurrentChannelMuted: (state: GlobalState) => boolean = createSele
...
 );
 
 export const isCurrentChannelDefault: (state: GlobalState) => boolean = createSelector(
     'isCurrentChannelDefault',
     getCurrentChannel,
-    (channel: Channel): boolean => isDefault(channel),
+    (channel) => {
+        return isDefault(channel);
+    },
```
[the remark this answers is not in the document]
M-ZubairAhmed: ✅  Updated as per request/discussion

packages/mattermost-redux/src/selectors/entities/channels.ts
```diff
@@ -317,7 +320,7 @@ export function makeGetChannelUnreadCount(): (state: GlobalState, channelId: str
         (state: GlobalState, channelId: string) => getChannelMessageCount(state, channelId),
         (state: GlobalState, channelId: string) => getMyChannelMembership(state, channelId),
         isCollapsedThreadsEnabled,
-        (messageCount: ChannelMessageCount, member: ChannelMembership, crtEnabled: boolean) => {
+        (messageCount: ChannelMessageCount, member: ChannelMembership, crtEnabled) => {
```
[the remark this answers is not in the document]
M-ZubairAhmed: ✅  Updated as per request/discussion

packages/mattermost-redux/src/selectors/entities/preferences.ts
```diff
@@ -195,12 +195,16 @@ export function isCollapsedThreadsAllowed(state: GlobalState): boolean {
...
-}
+export const isCollapsedThreadsEnabled = createSelector(
+    'isCollapsedThreadsEnabled',
+    isCollapsedThreadsAllowed,
+    getCollapsedThreadsPreference,
+    getConfig,
+    (collapsedThreadAllowed, collapsedThreadPreference, config) => {
+        return collapsedThreadAllowed &&
+            (collapsedThreadPreference === Preferences.COLLAPSED_REPLY_THREADS_ON || (config.CollapsedThreads as string) === 'always_on');
+    },
```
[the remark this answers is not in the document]
M-ZubairAhmed: ✅  Updated as per request/discussion

selectors/views/channel_sidebar.ts
M-ZubairAhmed: ✅  Updated as per request/discussion

selectors/views/custom_status.ts
```diff
@@ -50,10 +50,13 @@ export const getRecentCustomStatuses = createSelector(
...
-export function isCustomStatusEnabled(state: GlobalState) {
-    const config = getConfig(state);
-    return config && config.EnableCustomUserStatuses === 'true';
-}
+export const isCustomStatusEnabled = createSelector(
+    'isCustomStatusEnabled',
+    (state: GlobalState) => getConfig(state),
+    (config) => {
+        return config && config.EnableCustomUserStatuses === 'true';
+    },
```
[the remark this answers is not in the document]
M-ZubairAhmed: ✅  Updated as per request/discussion

components/channel_header/index.js
M-ZubairAhmed: https://github.com/mattermost/mattermost-webapp/pull/9808#discussion_r807647580

components/sidebar/sidebar_channel_list/sidebar_channel_list.tsx
michelengelen: yes, you are right .. I did not think about that! Nice find BTW!

michelengelen (approved):
LGTM

packages/mattermost-redux/src/selectors/entities/preferences.ts
```diff
@@ -195,12 +195,14 @@ export function isCollapsedThreadsAllowed(state: GlobalState): boolean {
     );
 }
 
-export function isCollapsedThreadsEnabled(state: GlobalState): boolean {
-    const isAllowed = isCollapsedThreadsAllowed(state);
-    const userPreference = getCollapsedThreadsPreference(state);
-
-    return isAllowed && (userPreference === Preferences.COLLAPSED_REPLY_THREADS_ON || getConfig(state).CollapsedThreads as string === 'always_on');
-}
+export const isCollapsedThreadsEnabled = createSelector(
```
hmhealey, line 198: Using `createSelector` here shouldn't be necessary. Since the results of the input selectors are just primitive values and a reference to the config, it probably takes about the same time to see if their results have been to check if their results have changes as it does to just return the result of that calculation. `createSelector` is only really needed if either the computation of the final result function is expensive or if it's returning an object/array

selectors/views/channel_sidebar.ts
```diff
@@ -27,9 +27,12 @@ import {memoizeResult} from 'mattermost-redux/utils/helpers';
 
 import {DraggingState, GlobalState} from 'types/store';
 
-export function isUnreadFilterEnabled(state: GlobalState): boolean {
-    return state.views.channelSidebar.unreadFilterEnabled && !shouldShowUnreadsCategory(state);
-}
+export const isUnreadFilterEnabled = createSelector(
```
hmhealey, line 30: Same here since the input selectors just return booleans

selectors/views/custom_status.ts
```diff
@@ -50,10 +50,11 @@ export const getRecentCustomStatuses = createSelector(
     },
 );
 
-export function isCustomStatusEnabled(state: GlobalState) {
-    const config = getConfig(state);
-    return config && config.EnableCustomUserStatuses === 'true';
-}
+export const isCustomStatusEnabled = createSelector(
```
hmhealey, line 53: And finally here

selectors/views/channel_sidebar.ts
```diff
@@ -211,28 +214,32 @@ function maxDefined(a: number, b?: number) {
     return typeof b === 'undefined' ? a : Math.max(a, b);
 }
 
+function concatChannels(channelsA: Channel[], channelsB: Channel[]) {
+    return [...channelsA, ...channelsB];
+}
+
+const memoizedConcatChannels = memoizeOne(concatChannels);
```
hmhealey, line 221: What's the difference between `memoizeOne` and the old `memoizeResult` function? Comparing the source for that, they look mostly the same except `memoizeOne` is spreading the arguments array which is supposedly worse according to the comments that I copied from `reselect` at some point. I don't know if those comments are still relevant since those comments are rather old, but I don't know if either would perform better or worse anyway.

Also, declaring this out here isn't great because it could allow others to accidentally use it and overwrite the memoized array. The reason it was written kind of weirdly before was to hide `memoizedConcat` from other functions in this file so that it couldn't be accidentally misused

selectors/views/channel_sidebar.ts
```diff
@@ -211,28 +214,32 @@ function maxDefined(a: number, b?: number) {
...
-export const getDisplayedChannels = (() => {
-    const memoizedConcat = memoizeResult((unreadChannels: Channel[], channelsInCategoryOrder: Channel[]) => {
-        return [...unreadChannels, ...channelsInCategoryOrder];
-    }) as (a: Channel[], b: Channel[]) => Channel[];
-
-    return (state: GlobalState) => {
-        // If the unread filter is enabled, only unread channels are shown
-        if (isUnreadFilterEnabled(state)) {
-            return getUnreadChannels(state);
+export const getDisplayedChannels = createSelector(
```
hmhealey, line 225: Wrapping this in `createSelector` will actually make it behave somewhat worse since any changes to `getUnreadChannels` or `getChannelsInCategoryOrder` will cause the selector to recompute even if their results aren't used. The previous way this was written, it waited until seeing if the unread filter or category was enabled before calling the selectors used for it.

I'm not sure if these changes will end up making this any better at the moment. Perhaps if we could add some tests for cases where this would perform better, I'd feel more confident with changing this, but I don't know if these changes will have any noticeable difference in performance

hmhealey (changes requested):

selectors/views/channel_sidebar.ts
hmhealey: As discussed, we're going to leave this as-is. Any differences between these two versions are likely not noticeable either way, and this way of writing the selector is more straightforward to understand since it's a pattern we follow everywhere else.

packages/mattermost-redux/src/selectors/entities/preferences.ts
M-ZubairAhmed: yes makes sense since it would not be much performance gain. Reverted.

selectors/views/custom_status.ts
M-ZubairAhmed: reverted

selectors/views/channel_sidebar.ts
M-ZubairAhmed: replaced it with `memoizeResult`

hmhealey (approved):

selectors/views/channel_sidebar.ts
M-ZubairAhmed: reverted

M-ZubairAhmed:
E2E test report https://automation-dashboard.vercel.app/cycles/d8123e78-7a5a-4f96-a6e4-3ed512323d67

jgilliam17:
Тhanks @M-ZubairAhmed 
Tested, looks good to merge.
- Verified Find Channels modal is working as before.
- [E2E report](https://automation-dashboard.vercel.app/cycles/9e7d6fa2-46de-420f-be60-e0151fd453fa) looks good, no PR related failures.

PR #9822, opened 2022-02-15
finalizing guest accounts metric

michelengelen:
#### Summary
- added fetching of guest accounts
- added test for whitelisted URLs (if existing)
- added threshold suggestion (5% of `totalUsers`)

refactored data mutation
- liveUrl status based on react state
- serverVersion status based on react state

#### Ticket Link
n/a

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/19437

#### Screenshots
n/a

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/admin_console/workspace-optimization/dashboard.tsx
+++ b/components/admin_console/workspace-optimization/dashboard.tsx
@@ -50,6 +50,8 @@ const successIcon = (
 const WorkspaceOptimizationDashboard = (props: Props) => {
     const [loading, setLoading] = useState(true);
     const [versionData, setVersionData] = useState<{type: string; version: string; status: ItemStatus}>({type: '', version: '', status: 'none'});
+    const [guestAccountStatus, setGuestAccountStatus] = useState<ItemStatus>('none');
+    const [liveUrlStatus, setLiveUrlStatus] = useState<ItemStatus>('none');
     const {formatMessage} = useIntl();
     const {getAccessData, getConfigurationData, getUpdatesData, getPerformanceData, getDataPrivacyData, getEaseOfManagementData} = useMetricsData();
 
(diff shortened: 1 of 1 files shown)


components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -59,25 +61,16 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
...
-                data.access.items[index].status = status === 'OK' ? 'ok' : 'error';
-            };
-            const onError = () => {
-                data.access.items[0].status = 'error';
-            };
-            return testSiteURL(onSuccess, onError, location.origin);
-        }
-
-        return Promise.resolve();
+        const onSuccess = ({status}: any) => setLiveUrlStatus(status === 'OK' ? 'ok' : 'error');
```
pablovelezvidal, line 71: nit: we should put all these string values either in a enum or simple constant values, as a best practice.

pablovelezvidal (approved):
LGTM! 🚀

mkraft (approved):

components/admin_console/workspace-optimization/dashboard.tsx
```diff
@@ -120,17 +113,34 @@ const WorkspaceOptimizationDashboard = (props: Props) => {
             }
 
             setVersionData({type, version: result.tag_name, status});
-            return;
+        }
+    };
+
+    const fetchGuestAccounts = async () => {
+        if (TeamSettings?.EnableOpenServer && GuestAccountsSettings?.Enable) {
+            let usersArray = await fetch('/api/v4/users/invalid_emails').then((result) => result.json());
```
BenCookie95, line 121: Should we pass some query parameters for page & per_page? Just in case it turns out there are thousands of records returned?

michelengelen: good question ... is this a very common use case? And how do we determine if we need to fetch all?

BenCookie95: I don't ever think we should fetch all, just fetch the first 50 here

michelengelen: Yeah, I know what you mean, but consider this: For a server that has 20K+ users having 5 guests is pretty reasonable and would not be very actionable. On the other hand a server with 50 users and 5 guest accounts this is a different story

BenCookie95: I don't mind keeping the threshold at 5% if we can at least limit the number of users returned from that endpoint

michelengelen: well ... maybe we can adjust the API a bit to accept a parameter to only return the number of users with email domains that differ from the server domain? On the other hand that would mean we would not be able to check for whitelisted domains

BenCookie95: So a total count of invalid users? That's a good idea but with our current timelines I don't think we have time for any more adjustments.

@mkraft what's your take on this?

mkraft: I think if it's not clearly defined and performant by this point this widget can wait for a future iteration. It was a nice-to-have anyhow and we don't want to delay the rest of our great work because of it.

michelengelen: ok, so for now we can just merge it as is and revisit later, right?

Any objections @BenCookie95 ?

mkraft: My intents are:
* avoid delaying getting this to QA
* no big performance slowdown if it makes it in

michelengelen: question is: is the API slowing down significantly if the numbers are high? Can we test that somehow, maybe on community-daily?

michelengelen: OK, I will remove that metric for now and we can re-visit when the MVP is out

michelengelen:
@BenCookie95 and @pablovelezvidal could you please re-review and merge when everything is fine? THX! 🙇🏼

BenCookie95 (approved):

PR #9824, opened 2022-02-15
Add First Contentful Paint telemetry

M-ZubairAhmed:
#### Summary
This add console log telemetry for measuring FCP. The element in consideration is post textbox, which measures the time it takes to paint on the screen. As soon as its done the performance observer is disconnected to free up performance buffer. It currently works for Chrome.

#### Related Pull Requests
N/A

#### Screenshots
image

#### Release Note

```release-note

```


M-ZubairAhmed:
@jwilander @hmhealey This will be here if we ever need it in future. Closing it out to reduce the noise.

PR #9833, opened 2022-02-17
MM-41955 Reduce useless changes to user state

hmhealey:
As discussed, we want to reduce accidental state changes caused by reducers returning new state objects which are identical to the the previous values. This does that for all of the most used parts of users state (`profiles`, `statuses`, `profilesInChannel`). I didn't bother doing it with a lot of the other ones like `profilesInTeam` or `profilesNotInGroup` since they're only really used in the System Console and modals for managing users in channels/teams/groups, so they won't impact general performance as much. Ideally, I'd actually like to get those ones out of redux entirely, but that's future work.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41955

#### Release Note
```release-note
Improved performance of code for storing users in web app
```


components/root/root.jsx
```diff
@@ -181,10 +180,6 @@ export default class Root extends React.PureComponent {
     }
 
     onConfigLoaded = () => {
-        if (isDevMode()) {
-            enableDevModeFeatures();
```
hmhealey, line 185: This added a warning when trying to access a non-existent `.length` property of a Set or Map which lodash does in `isEqual` to try to identify the type of object. Since we have TypeScript now that will warn us if we do that by mistake, that warning isn't needed any more anyway.

packages/mattermost-redux/src/reducers/entities/users.ts
```diff
@@ -458,25 +454,31 @@ function profilesNotInGroup(state: RelationOneToMany<Group, UserProfile> = {}, a
     }
 }
 
+function saveToState<T>(state: Record<string, T>, key: string, value: T): Record<string, T> {
```
M-ZubairAhmed, line 457: can we think of a better name, because its not actually saving but merging 0/5 🤔

M-ZubairAhmed (approved):

packages/mattermost-redux/src/reducers/entities/users.ts
hmhealey: Maybe `addToState`? I just put this helper here, but it seems like it'd be really helpful throughout all the redux state. I'm pretty sure we could use these helpers in Joram's PR too, but I don't want to change too much else right now.

packages/mattermost-redux/src/reducers/entities/users.ts
```diff
@@ -171,50 +183,50 @@ function profiles(state: IDMappedObjects<UserProfile> = {}, action: GenericActio
...
-    case UserTypes.RECEIVED_PROFILES_LIST:
-        return Object.assign({}, state, profileListToMap(action.data));
-    case UserTypes.RECEIVED_PROFILES:
-        return Object.assign({}, state, action.data);
+    case UserTypes.RECEIVED_PROFILES_LIST: {
+        const users: UserProfile[] = action.data;
 
-    case UserTypes.LOGIN: { // Used by the mobile app
-        const {user} = action.data;
+        return users.reduce((nextState, user) => {
```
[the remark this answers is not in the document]
hmhealey: It's kind of a weird one since we don't really know if it's `state` or `nextState` at this point since the previous iterations could've returned either the existing state or a new one. This is the best naming convention I've come up with so far when using reduce though since just calling it `state` would cause it to shadow the outer one.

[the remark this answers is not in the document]
jwilander: I'm good with it 👍

jgilliam17:
[E2E report ](https://mattermost-cypress-report.s3.amazonaws.com/152432-d9e07f2-onprem-ent-webapp-pr-9833-mattermost/mm-ee-testd9e07f2/mochawesome.html)shows many failures (35+) related to autocomplete 
Mention autocomplete is showing a loading spinner e.g. for no results and also when match is found
Screen Shot 2022-02-24 at 10 29 06 PM

hmhealey:
Those should be fixed now. Turns out the autocomplete was throwing an error which broke most of those tests

jgilliam17 (approved):
Thanks @hmhealey 
Tested, looks good to merge.
- [E2E report](https://mattermost-cypress-report.s3.amazonaws.com/157495-15f1781-onprem-ent-webapp-pr-9833-mattermost/mm-ee-test15f1781/mochawesome.html) looks good, no PR related failures.

amyblais:
@hmhealey Needs a manual cherry-pick.

PR #9847, opened 2022-02-18
MM-41741: fixes leaving an unread channel

koox00:
#### Summary

Currently when you leave a currently viewing channel, and that channel
is unread, and you have the setting "Group unread channels separately"
to ON the app crashes.

Since we remove the channel both from myMembers and channels,
but we keep an unread channel when while we are viewing it in the
unreads group, we need to guard against myMembers to ensure that the
channels in the unread list are always joined channels.

This commit does exactly that.

PS: the regression happened when we started removing channels from redux store upon leaving them

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41741

#### Release Note

```release-note
* Fixes app crash upon leaving a currently viewing unread channel with channel unread grouping set to ON
```


michelengelen (approved):
LGTM

amyblais:
@hmhealey @jgilliam17 This is a candidate to fix before Cloud March 2nd release.

e2e/cypress/integration/channel_sidebar/group_unreads_separately_spec.js
```diff
@@ -105,6 +105,25 @@ describe('Channel sidebar - group unreads separately', () => {
         // * Verify channel is not in the UNREADS category
         cy.get('.SidebarChannelGroup:contains(CHANNELS)').should('be.visible').get('.SidebarChannel:not(.unread):contains(Off-Topic)').should('be.visible');
     });
+
+    it('Leaving an unread channel when unread category is ON', () => {
```
jgilliam17, line 109:
```suggestion
    it('MM-T4655 Leaving an unread channel when unread category is ON', () => {
```

jgilliam17 (commented):
Thanks @koox00 
Tested, looks good no crash, seeing one issue 
When rejoining the same channel, channel is added to Unreads on the sidebar. I expected it in Channels category.
- Added test case to Zephyr and suggested ID below.

Willyfrog (changes requested):
Looks good, but I'm missing having a unittest for the case at least (nice that you added the e2e test btw!)

koox00:
> When rejoining the same channel, channel is added to Unreads on the sidebar. I expected it in Channels category.

Can we add another task for that, since that's true for current master as well?
I'd prefer to keep this one just to fix the crash.

hmhealey (commented):
Could we also clear the state returned by [the `lastUnreadChannel` reducer ](https://github.com/mattermost/mattermost-webapp/blob/master/reducers/views/channel.ts#L112) as well when leaving the channel? We might end up with other code that depends on it, causing us to run into this same issue later.

That might also fix the thing Jelena was reporting above where the channel returns to the Unreads category after rejoining it.

Willyfrog (approved):

hmhealey (approved):

jgilliam17:
Thanks @koox00 
Tested, looks good to merge.
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/155988-3be03a8-onprem-ent-webapp-pr-9847-mattermost/mm-ee-test3be03a8/mochawesome.html) also looks good, no PR related failures. 
Filed re-joining the same channel issue separately.
https://mattermost.atlassian.net/browse/MM-42140

jgilliam17 (approved):

PR #9848, opened 2022-02-18
Do not render ChannelHeaderMobile when not in mobile view to prevent heavy selector computations

jwilander:
#### Summary
The NotifyCounts component which is mounted as a child component of the tree under ChannelHeaderMobile uses a very heavy selector `getUnreadStatusInCurrentTeam` that recomputes often. Since the mobile channel header is only needed when we're in mobile view, this PR changes it to only mount when we're looking at mobile view.

@jgilliam17 for testing this one can you make sure that the mobile channel header still appears when loading the app in a mobile browser or when shrinking the app in a desktop browser to be small?

#### Release Note

```release-note
NONE
```


M-ZubairAhmed (approved):

components/channel_layout/center_channel/index.ts
```diff
@@ -45,6 +45,7 @@ const mapStateToProps = (state: GlobalState, ownProps: Props) => {
         lhsOpen: getIsLhsOpen(state),
         rhsOpen: getIsRhsOpen(state),
         rhsMenuOpen: getIsRhsMenuOpen(state),
+        isMobile: state.views.channel.mobileView,
```
M-ZubairAhmed, line 48: would we use `getIsMobileView` selector? 0/5 on it thought

hmhealey: +1 to that

jwilander: Yep, good call. I'll update it.

hmhealey (changes requested):
I'm not sure why, but this causes the header to not appear in the onboarding screen for some reason
Screen Shot 2022-02-22 at 11 17 29 AM

jwilander:
@M-ZubairAhmed think you could take this PR over too? I haven't had time to get back to it yet

M-ZubairAhmed:
🧪 e2e report 🧪 

https://automation-dashboard.vercel.app/cycles/01e6ff9f-2cbb-46b1-ab1f-95f45bfe09da

M-ZubairAhmed:
Reopened https://github.com/mattermost/mattermost-webapp/pull/10811

PR #9849, opened 2022-02-18
[MM-41904] Optimize group list permissions

BenCookie95:
#### Summary
Moved the function for creating `groupPermissionsMap` to a selector in order to reduce the number of times this gets computed.

Added some unit tests to cover the new functions we added. We have 4 permissions that get assigned to a user with the correct role:
- Permissions.EDIT_CUSTOM_GROUP
- Permissions.CREATE_CUSTOM_GROUP
- Permissions.MANAGE_CUSTOM_GROUP_MEMBERS
- Permissions.DELETE_CUSTOM_GROUP

We also have another permission called `custom_group_user` that gets assigned when you are a member of a group. As of right now that permission is not used but I guess it's useful to keep it in case we change how group permissions work.

My understanding of the roles/permissions is a little raw so correct me if the above statement is incorrect.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41904

#### Release Note
```release-note
NONE
```


jespino (approved):
LGTM.

packages/mattermost-redux/src/selectors/entities/roles.ts
```diff
@@ -50,6 +50,43 @@ export const getMyGroupRoles: (state: GlobalState) => Record<string, Set<string>
...
+    getMyGroupRoles,
+    getRoles,
+    getMySystemPermissions,
+    (state) => state.entities.groups.groups,
+    (myGroupRoles, roles, systemPermissions, allGroups) => {
+        const groups = Object.entries(allGroups).filter((entry) => (entry[1].allow_reference && entry[1].delete_at === 0)).map((entry) => entry[1]);
+
+        const permissions = new Set<string>();
+        groups.forEach((group) => {
+            if (myGroupRoles[group.id!]) {
```
mkraft, line 64: a variable (like `roleNames`) could be instantiated with the value of `myGroupRoles[group.id!]` to avoid re-accessing the object.

packages/mattermost-redux/src/selectors/entities/roles.ts
```diff
@@ -50,6 +50,43 @@ export const getMyGroupRoles: (state: GlobalState) => Record<string, Set<string>
     },
 );
 
+export const getGroupListPermissions: (state: GlobalState) => Record<string, GroupPermissions> = createSelector(
```
mkraft, line 53: It would be nice to add some documentation to this, for example:

```suggestion
/** Returns a map of permissions—keyed by group id—for groups in state that are not deleted, mentionable, and that the current user is a member of.*/
export const getGroupListPermissions: (state: GlobalState) => Record<string, GroupPermissions> = createSelector(
```

BenCookie95: Added a comment

mkraft (approved):

furqanmlk (approved):
LGTM
**Permission Unchecked**

image

**Permission Checked**
image

PR #5907, opened 2020-07-15
MM 16068 announcement bar tooltip show appropriately

fullstackguyhere:
#### Summary
This PR is mainly to fix the issue of Tooltips always displaying when a user hovers over an Announcement banner, even when the text isn't truncated.
It fixes the following issues:-

1. Only show the tooltip if the text in the banner truncates
2. Fix the z-index of the tooltip, it sits behind the banner

Along with these two issues, additionally fixed:-
3. Fix the additional issue of not being able to click links or other clickable content inside the tooltip. The reason being pointer-events proper for .tooltip class set to none. So I fixed this for announcement bar-related tooltips.

Apart from these issues, had to add constant OVERLAY_ANNOUCMENT_HIDE_DELAY to increase delay for hiding overlays since if the content inside the tooltip is clickable, the user needs more than 400 microseconds to click it. 

#### Ticket Link
Issue on Github: https://github.com/mattermost/mattermost-server/issues/14404

Issue on JIRA: https://mattermost.atlassian.net/browse/MM-16068

#### Related Pull Requests
Not applicable here

PR #9851, opened 2022-02-20
MM-16068: Improve announcement banner tooltip

sinansonmez:
#### Summary
Currently tooltip is always displayed on hover regardless of text is truncated or not

The announcement banner is improved on the following use-cases
- If text is not long enough, the text is not truncated and tooltip is not displayed on hover
- if text without URL is long enough, the text is truncated and tooltip is displayed on hover
- if text with URL is long enough, the text is truncated and tooltip is displayed on hover. Tooltip is hidden with 600 ms (implemented based on previous [feedback](https://github.com/mattermost/mattermost-webapp/pull/5907#issuecomment-680290989))

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/14404

#### Screenshots
**before**

https://user-images.githubusercontent.com/37421564/154868584-9b42b7c0-5151-49eb-81cb-5bfaa48b1095.mov

**after with long text**

https://user-images.githubusercontent.com/37421564/154868655-cfe6cd56-d44a-4aaf-923b-722cba024a3c.mov

**after with short text**

https://user-images.githubusercontent.com/37421564/154868678-3c59c16e-77c1-4d85-99ca-505d2967f4de.mov

#### Release Note
```release-note
Tooltip is only displayed when text is too long in the announcement banner
```


utils/url.tsx
```diff
@@ -185,3 +185,8 @@ export function isPermalinkURL(url: string): boolean {
 
     return isInternalURL(url, siteURL) && (regexp.test(url));
 }
+
+export function isStringContainingUrl(text: string): boolean {
```
sinansonmez, line 189: I have limited knowledge on Regex. I am open to any suggestion to improve this function

M-ZubairAhmed:
@sinansonmez can you please fix the conflicts

utils/url.tsx
hmhealey: As far as I can tell, that regex looks okay (I've written more than a few regexes that have caused the whole app to freeze on bad input), but I think we can probably simplify this. We only care that the string contains a URL and not what exactly that URL is, which would be a lot harder, so I think we just need the first part of yours that checks whether or not a string starts with a URL protocol. We should also add something to check if the string starts with `www.` since we also autolink links like that

Here's a couple examples of the bare minimum that we autolink currently:
Screen Shot 2022-02-22 at 1 02 33 PM

hmhealey (changes requested):
Thanks for the PR! I have a couple requests, but this looks good so far

utils/url.tsx
sinansonmez: Regex is fairly simplified, as suggested checking the following cases, url includes one of them
- `https://`
- `http://`
- `www.`

utils/constants.jsx
```diff
@@ -1610,6 +1610,7 @@ export const Constants = {
     },
     OVERLAY_TIME_DELAY_SMALL: 100,
     OVERLAY_TIME_DELAY: 400,
+    OVERLAY_ANNOUNCEMENT_HIDE_DELAY: 600,
```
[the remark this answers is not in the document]
sinansonmez: Done. I will keep in mind this rule for future PRs

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
```diff
@@ -112,11 +145,11 @@ export default class AnnouncementBar extends React.PureComponent<Props> {
                 <FormattedMarkdownMessage id={this.props.message}/>
             );
         }
-        const announcementTooltip = (
+        const announcementTooltip = this.state.showTooltip ? (
             <Tooltip id='announcement-bar__tooltip'>
                 {this.props.tooltipMsg ? this.props.tooltipMsg : message}
             </Tooltip>
-        );
+        ) : <></>;
```
devinbinnie, line 152: Nit: Can we have this return `null` so we don't have a `React.Fragment` in the snapshot? It's just a bit cleaner :)

utils/url.tsx
devinbinnie: This should be fine as long as we don't care about other protocols. From a security standpoint this is actually probably better to not generate links for other protocols :P

We could potentially want to use `mailto:` in the announcement banner, but 0/5 on whether we allow that or not.

devinbinnie (approved):
LGTM! Just have a couple non-blocking comments. Thanks @sinansonmez :)

components/announcement_bar/default_announcement_bar/announcement_bar.tsx
sinansonmez: Thanks for the comment. In the beginning I tried to pass null but in case of no tooltip is needed, `OverlayTrigger ` component was throwing error because I was passing null for `overlap` prop on [line 167](https://github.com/mattermost/mattermost-webapp/blob/e73b161f528db8f774eb428b28cf36350c85dbec/components/announcement_bar/default_announcement_bar/announcement_bar.tsx#L167)

devinbinnie: Ahh so it's a type issue. We can leave it for now.

esethna (approved):
Looks great based on the screenshots! Thanks @sinansonmez, excellent work!

furqanmlk (commented):
@sinansonmez 
I tested other banner with Announcement Banner.
Looks like `Expired banner` is getting pushed to next line

image

I tested on master, expired banner is displayed correctly

sinansonmez:
@furqanmlk thanks for catching it. I fixed the issue with my latest commit.

utils/url.tsx
hmhealey: Strictly speaking, we support any number of different URL protocols including custom ones both with and without the `//`, but I don't expect anyone to reasonably use them here, so let's ignore those for now to avoid complicating things 😄

hmhealey (approved):

hmhealey:
Looks good to me!

furqanmlk (approved):
Thanks @sinansonmez 
Working as expected.

furqanmlk:
@hmhealey or @devinbinnie can you please merge this PR ?

hmhealey:
Thanks again for the PR!

PR #9854, opened 2022-02-22
adding Persian as a beta language

ctlaltdieliet:
#### Summary
Adding Persian as an new official (beta) language to Mattermost

#### Related Pull Requests
- [Has mobile changes](https://github.com/mattermost/mattermost-mobile/pull/5995)

#### Release Note

```release-note
Adding Persian as an official supported language (Beta)
```


ctlaltdieliet:
Pinging @jasonblais and @cwarnermm

i18n/i18n.jsx
```diff
@@ -124,28 +125,34 @@ const languages = {
         order: 15,
         url: uk,
     },
+    ko: {
```
hmhealey, line 128: I'm not sure if this will fix the error around the missing `fa.json`, but this is actually getting overwritten since it's reusing the same key name as Korean is below
```suggestion
    fa: {
```

hmhealey (changes requested):

ctlaltdieliet:
Thank you  @hmhealey ! Peer reviews do matter!

hmhealey (approved):
Thanks! This looks good to me now, but I think we're going to need to block it on the translation updates so that that the json file exists

cwarnermm:
@hmhealey - Thanks! CC'ing @spirosoik for visibility

hmhealey:
@ctlaltdieliet There's a snapshot test failing since it's relying on one of the values that you had to change. Would you be able to fix that?

ctlaltdieliet:
> @ctlaltdieliet There's a snapshot test failing since it's relying on one of the values that you had to change. Would you be able to fix that?
@hmhealey Done :)

hmhealey:
Thanks!

jgilliam17:
Thanks @ctlaltdieliet 
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/157221-b264621-onprem-ent-webapp-pr-9854-mattermost/mm-ee-testb264621/mochawesome.html) looks good, no PR related failures
Wondering why Channels and Favorites are localized on the sidebar, but not on the menu
Screen Shot 2022-03-01 at 10 17 01 AM
For other strings still shown in English, I assume those are ok and just missing translation at the moment, correct?
Screen Shot 2022-03-01 at 10 32 04 AM

hmhealey:
I think both those are fine since they're not directly tied to this PR

>  Wondering why Channels and Favorites are localized on the sidebar, but not on the menu

That seems like a bug since it's also untranslated in other languages which are out of beta. That text probably can't be translated which needs to be fixed in the code.

>  For other strings still shown in English, I assume those are ok and just missing translation at the moment, correct?

That one's just a missing string. We'll just claim "beta" on that one. It'll be fixed by the translators eventually

jgilliam17:
Тhanks @hmhealey 
I added https://mattermost.atlassian.net/browse/MM-42401 for follow up. Looks good to merge.

jgilliam17 (approved):

PR #9857, opened 2022-02-22
MM-39537: fixes delay on bolding channel in LHS

koox00:
#### Summary

Currently we await for the API to mark a post as unread, this commit
changes the order so we optimistically mark a post as unread and then
wait for the actual request to finish.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39537

#### Release Note

```release-note
NONE
```


michelengelen (approved):
LGTM and in case anything goes wrong in the API call we still have the `catch`, so that change should be fine to go!

koox00:
> LGTM and in case anything goes wrong in the API call we still have the `catch`, so that change should be fine to go!

@michelengelen yep, BTW this seems like it was written before with optimistic update in mind, 
but for some dark reason, it didn't do that...

hmhealey (approved):

hmhealey:
Yeah, it was definitely written before we were trying to do things more optimistically. All of the APIs written using `bindClientFunc` are also not optimistic

PR #9869, opened 2022-02-23
Fix possible leak in SidebarChannel

hmhealey:
Thanks to Jesús for pointing this out. Everywhere else that we use the accessibility events uses a state change to attach any additional listeners, so they can't be affected by this same issue.

#### Release Note
```release-note
Fix potential memory leak in sidebar when using accessibility hotkeys
```


devinbinnie (approved):

jespino (approved):
LGTM

PR #9876, opened 2022-02-24
Fixing possible (low priority) memory leaks

jespino:
#### Summary
This memory leaks are specially low priority because they happens on unmount,
and this 2 components only get unmounted when you stop seeing the global header
(than means, when you switch to the admin console, or when you switch to the
backstage.

Probably would be useful to talk about this other way of leaking memory in the
Developers Meeting.

Thanks to @hmhealey for bringing up this kind of issues.

#### Ticket Link
NOT YET

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/global_header/left_controls/history_buttons/history_buttons.tsx
+++ b/components/global_header/left_controls/history_buttons/history_buttons.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useEffect, useState} from 'react';
+import React, {useEffect, useState, useCallback} from 'react';
 import styled from 'styled-components';
 import IconButton from '@mattermost/compass-components/components/icon-button';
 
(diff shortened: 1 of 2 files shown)


components/global_header/left_controls/history_buttons/history_buttons.tsx
```diff
@@ -62,7 +62,7 @@ const HistoryButtons = (): JSX.Element => {
         );
     };
 
-    const handleButtonMessage = (message: {origin: string; data: {type: string; message: {enableBack: boolean; enableForward: boolean}}}) => {
+    const handleButtonMessage = useCallback((message: {origin: string; data: {type: string; message: {enableBack: boolean; enableForward: boolean}}}) => {
```
devinbinnie, line 65: Just curious, how come we don't format the logic the same way as above? (in which we place the function within the `useEffect` call)

devinbinnie (approved):

hmhealey (approved):

components/global_header/left_controls/history_buttons/history_buttons.tsx
hmhealey: I think I prefer the first way of doing this since it keeps everything contained in the `useEffect`. They do seem to do the same thing though

jespino: Actually, both options are good options, I don't know exactly why I did this differently.

PR #9881, opened 2022-02-24
[MM-42075] Add "(Beta)" to User Groups

mylonsuren:
#### Summary

Adds "(Beta)" after parts of the app that list `User Groups`.

#### Ticket Link

Fixes [MM-42075](https://mattermost.atlassian.net/browse/MM-42075) 

#### Screenshots

| image | image |
|----|----|
| image | image|

#### Release Note

```release-note
NONE
```


mkraft (approved):

BenCookie95 (approved):
Do we need a quick design approval as well?

mylonsuren:
> Do we need a quick design approval as well?

Yeah was wondering as well, also noticed that there is an overflow issue with the `+` for `Create New User Group`, not sure if that's being addressed separately per this [thread](https://community-daily.mattermost.com/core/pl/13x8u81xn7gipfckg8b7rqxspr), but it should also account for the added "(Beta)" text that's now part of it.

cc: @michaelgamble

mkraft:
> Do we need a quick design approval as well?

If design can a) look at it in time, and b) we can implement some alternative in time then I'm OK with doing a different solution. As long as this feature isn't released without some beta identifier.

BenCookie95:
> > Do we need a quick design approval as well?
> 
> If design can a) look at it in time, and b) we can implement some alternative in time then I'm OK with doing a different solution. As long as this feature isn't released without some beta identifier.

Yeah I agree with Martin

BenCookie95:
> Create New User Group

Good point about the menu item for `Create New User Group`, we won't be able to see the beta part even if increase the menu width. the proposed width was 232px

michaelgamble (changes requested):
My recommendation is that we follow in line with how the "Collapsed reply thread" feature handled it, where we show the (Beta) tag at the point of enabling the feature only, and leave the rest of the feature to look how we ultimately intend it to look. 

What that means from my perspective is remove it from everywhere except from the system console setting.

@mkraft @laneycs Does that work for you?

michaelgamble (changes requested):
Found a visual bug in the focus state
image

Found a bug in the menu item getting cut off.. it should be at a z-level that is not being cut off by the modal
image

BenCookie95:
> Found a visual bug in the focus state image
> 
> Found a bug in the menu item getting cut off.. it should be at a z-level that is not being cut off by the modal image

We will create separate tickets for those

mylonsuren:
> > Found a visual bug in the focus state image
> > Found a bug in the menu item getting cut off.. it should be at a z-level that is not being cut off by the modal image
> 
> We will create separate tickets for those

1) First bug related to the focus state should be resolved.
2) Ticket created here: [MM-42189](https://mattermost.atlassian.net/browse/MM-42189) for the hidden menu bug.

michaelgamble (approved):
LGTM

cwarnermm:
Documented via https://github.com/mattermost/docs/pull/5406

PR #9883, opened 2022-02-24
[MM-42068] Adding source query param to team and channel group endpoints

BenCookie95:
#### Summary
Adding `group_source` to the query parameters in order to filter out custom groups

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42068
https://mattermost.atlassian.net/browse/MM-42070

#### Release Note

```release-note
Fixes a bug where custom groups were appearing in the UI when linking groups to channels/teams
```

### DIFF
--- a/packages/mattermost-redux/src/selectors/entities/groups.ts
+++ b/packages/mattermost-redux/src/selectors/entities/groups.ts
@@ -10,6 +10,8 @@ import {getTeam} from 'mattermost-redux/selectors/entities/teams';
 import {UserMentionKey} from 'mattermost-redux/selectors/entities/users';
 import {GlobalState} from 'mattermost-redux/types/store';
 
+import Constants from 'utils/constants';
+
 import {getCurrentUserLocale} from './i18n';
 
 const emptyList: any[] = [];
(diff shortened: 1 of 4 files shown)


mkraft (approved):

mylonsuren (commented):
Should a test case be added to verify that custom groups are no longer included in the response?

BenCookie95:
> Should a test case be added to verify that custom groups are no longer included in the response?

There's a test on server to verify that the endpoint returns groups based on that query param, so I think we're good

mylonsuren (approved):

BenCookie95:
Discovered an issue so need to add another part to this

BenCookie95:
@mkraft & @mylonsuren This needs a re-review, had to adjust the redux selectors

packages/mattermost-redux/src/selectors/entities/groups.ts
```diff
@@ -148,7 +150,7 @@ export const getGroupsNotAssociatedToTeam: (state: GlobalState, teamID: string)
     getAllGroups,
     (state: GlobalState, teamID: string) => getTeamGroupIDSet(state, teamID),
     (allGroups, teamGroupIDSet) => {
-        return Object.entries(allGroups).filter(([groupID]) => !teamGroupIDSet.has(groupID)).map((entry) => entry[1]);
+        return Object.entries(allGroups).filter(([groupID]) => !teamGroupIDSet.has(groupID)).filter((entry) => (entry[1].source === Constants.LDAP_SERVICE)).map((entry) => entry[1]);
```
mylonsuren, line 153: Could we do something similar to ln. 173 in `getGroupsNotAssociatedToChannel` and removed the extra filter here?

```suggestion
        return Object.entries(allGroups).filter(([groupID, group]) => !teamGroupIDSet.has(groupID) && group.source === Constants.LDAP_SERVICE).map((entry) => entry[1]);
```

mylonsuren (changes requested):

packages/mattermost-redux/src/selectors/entities/groups.ts
BenCookie95: Yeah good suggestion

mkraft (approved):

mylonsuren (approved):

furqanmlk (approved):
Working as expected

**Before**

image

**After**

image

PR #9885, opened 2022-02-24
Read from myTeams if there is no currentTeam

neallred:
#### Summary
For instances created from cloud portal,  read from `myTeams` as a back up as `currentTeam` does not yet have data. This was causing separate errors for both the paths that:
* tried to create channels
* tried to redirect user to the main channels view

Opting to defer QA until this is running from instances created from test portal, as this is where the issue was observed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42019

#### Screenshots
Was able to find this out on an instance created from test portal:

myTeams-but-no-current-team

#### Release Note
```release-note
NONE
```

### DIFF
--- a/components/preparing_workspace/preparing_workspace.tsx
+++ b/components/preparing_workspace/preparing_workspace.tsx
@@ -135,6 +135,9 @@ export default function PreparingWorkspace(props: Props) {
     const isSelfHosted = useSelector(getLicense).Cloud !== 'true';
     const currentTeam = useSelector(getCurrentTeam);
     const myTeams = useSelector(getMyTeams);
+    // In cloud instances created from portal,
+    // new admin user has a team in myTeams but not in currentTeam.
+    const inferredTeam = currentTeam || myTeams?.[0];
     const config = useSelector(getConfig);
     const configSiteUrl = config.SiteURL;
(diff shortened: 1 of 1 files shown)


marianunez (approved):

components/preparing_workspace/preparing_workspace.tsx
```diff
@@ -233,7 +233,9 @@ export default function PreparingWorkspace(props: Props) {
             return;
         }
 
-        let team = currentTeam;
+        // In cloud instances created from portal,
+        // new admin user has a team in myTeams but not in currentTeam.
+        let team = (currentTeam || myTeams?.[0]);
```
julmondragon, line 238: Are parenthesis required by lint?

julmondragon: As the same validation is done in multiple places of `PreparingWorkspace` could be consolidated in a constant.

julmondragon (approved):
Nice catch.
Just some nit:

components/preparing_workspace/preparing_workspace.tsx
neallred: I don't think they're required, but I left them since the linter didn't mind. I can change. Consolidating to a constant sounds good to me!

PR #9887, opened 2022-02-24
MM-39768 Rename ToggleModalButtonRedux to ToggleModalButton

hmhealey:
This is getting rid of the last instance of the original ToggleModalButton in favour of the redux-connected version, and then it renames the redux-connected version to replace the original one.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39876

#### Release Note
```release-note
NONE
```


hanzei:
@hmhealey I don't think I'm the right person to review webapp refactoring code. Would you please request a review from someone else?

michelengelen (approved):
LGTM ... question: is there a plan to port this over to typescript and FC as well?

hmhealey:
Lol, I was figuring you'd ask that. I actually started to do it, but I ran into issues with the types of `dialogType` and `dialogProps` in TypeScript. We'd need to do something like typing the component as `ToggleModalButton<T>` where `T` is the given dialogue's props similar to how the `openModal` action is typed, but that causes issues in certain use cases, so I didn't think it was worth the effort to do right now.

michelengelen:
> ... I ran into issues with the types of `dialogType` and `dialogProps` in TypeScript. ...

I know what you mean. I encountered the same problem once and circumvented it somehow I cannot remember.

pablovelezvidal (approved):
LGTM!

michelengelen:
> > > ... I ran into issues with the types of `dialogType` and `dialogProps` in TypeScript. ...
> > 
> > 
> > I know what you mean. I encountered the same problem once and circumvented it somehow I cannot remember.
> 
> Realistically, we could also just getting rid of the component eventually. Even if it was converted to a functional component, it's a lot of lines just to add an `openModal` call to a button that's in a component that's probably already redux-connected anyway

Yes, absolutely ... at the end of the day it only fires up a function on click anyways, so I would say the same ... just provide a simple/stupid button that has a click handler built in ... that should do the trick.

PR #9897, opened 2022-02-25
only first admin sees preparing-workspace screen

neallred:
#### Summary
Second admin should not see use case onboarding.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42103 , but only the `First admin vs second admin scenario:` scenario listed in the QA description.

#### Related Pull Requests
Not related in the code sense, but there is a separate server PR that addresses the `>= 10` posts part of the bug:
* https://github.com/mattermost/mattermost-server/pull/19642

#### Screenshots

https://user-images.githubusercontent.com/13738432/155773390-6c3f863e-dab5-4549-9689-a076534c113c.mp4

#### Release Note

```release-note
NONE
```

### DIFF
--- a/components/preparing_workspace/preparing_workspace.tsx
+++ b/components/preparing_workspace/preparing_workspace.tsx
@@ -7,6 +7,7 @@ import {RouterProps} from 'react-router-dom';
 import {FormattedMessage, useIntl} from 'react-intl';
 
 import {GeneralTypes} from 'mattermost-redux/action_types';
+import {General} from 'mattermost-redux/constants';
 import {savePreferences} from 'mattermost-redux/actions/preferences';
 import {createChannel} from 'mattermost-redux/actions/channels';
 import {getFirstAdminSetupComplete as getFirstAdminSetupCompleteAction} from 'mattermost-redux/actions/general';
@@ -81,6 +82,7 @@ const WAIT_FOR_REDIRECT_TIME = 2000 - START_TRANSITIONING_OUT;
 export type Actions = {
     createTeam: (team: Team) => ActionResult;
     checkIfTeamExists: (teamName: string) => ActionResult;
+    getProfiles: (page: number, perPage: number, options: Record<string, any>) => ActionResult;
 }
 
 type Props = RouterProps & {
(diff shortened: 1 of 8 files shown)


packages/mattermost-redux/src/selectors/entities/users.ts
```diff
@@ -692,25 +692,27 @@ export function searchProfilesInGroup(state: GlobalState, groupId: Group['id'],
...
+export function checkIsFirstAdmin(currentUser: UserProfile, users: IDMappedObjects<UserProfile>): boolean {
+    if (!currentUser) {
+        return false;
+    }
+    if (!currentUser.roles.includes('system_admin')) {
+        return false;
+    }
+    const userIds = Object.keys(users);
+    for (const userId of userIds) {
+        const user = users[userId];
```
JulienTant, line 704: Looks like this could be replaced by a `for (const user of users)` - do I miss something?

JulienTant (changes requested):

packages/mattermost-redux/src/selectors/entities/users.ts
neallred: We can almost do that. JS doesn't allow for...of object iteration , but we can iterate on the `Object.values` which I think is cleaner code and has the perks of being faster: https://jsbench.me/6il02vjfuw/1

JulienTant (approved):
LGTM, thanks for taking the time to make a benchmark :)

packages/mattermost-redux/src/selectors/entities/users.ts
JulienTant: image
Slight improvement on my machine, only +~525op/s haha

julmondragon (approved):
Looks good!

components/preparing_workspace/preparing_workspace.tsx
```diff
@@ -168,6 +170,7 @@ export default function PreparingWorkspace(props: Props) {
     const [submitError, setSubmitError] = useState<string | null>(null);
     useEffect(() => {
         showOnMountTimeout.current = setTimeout(() => setShowFirstPage(true), 40);
+        props.actions.getProfiles(0, General.PROFILE_CHUNK_SIZE, {roles: General.SYSTEM_ADMIN_ROLE});
```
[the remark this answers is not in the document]
neallred: Good questions! I also assume it will be very unlikely to have 101 system admin profiles in any system. The service does support sorting by creation. However, most uses of the selectors don't read the data directly from the server, but from redux, which doesn't retain the sort order since we store it in redux as an object.

jgilliam17:
Thanks @neallred 
Tested, looks good to merge 
- Verified only first system admin sees the wizard.  The onboarding wizard and checklist are not shown to other users. 
- I'm not running E2Es for now as discussed with @neallred, @AshishDhama is working on e2e test fixes that will be handled in a separate PR.

jgilliam17 (approved):

PR #9901, opened 2022-02-25
[MM-41995] Channel Info RHS

JulienTant:
#### Summary
This PR adds a new 'Channel info' button at the very far right of the channel header, opening the new channel info panel in the RHS.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41995

#### Related Pull Requests
N/a

##### For a channel
image

##### For a Group channel
image

##### For a DM
image

#### Release Note
```release-note
Added a right-hand side panel to see and interact with channel information.
```


components/channel_info_rhs/menu.tsx
```diff
@@ -0,0 +1,115 @@
...
+    icon: JSX.Element;
+    text: string;
+    onClick: () => void;
+}
+
+const menuItem = ({icon, text, className, onClick}: MenuItemProps) => {
+    return (
+        <div className={className}>
+            <div
+                className='MenuItem--main-action MenuItem--container'
```
neallred, line 27: Since we're using styled components for this component, I'd recommend breaking these subcomponents up into more styled components. The naming kind of looks like BEM conventions which is confusing because it isn't what's intended here.

components/channel_info_rhs/menu.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    .MenuItem--main-action {
+        &__text {
+            padding-left: 8px;
+            flex: 1;
+        }
+    }
+
+    &:hover {
+       background: rgba(63, 67, 80, 0.08);
```
neallred, line 62: Ditto the earlier comment about replacing with theming variables.

components/channel_info_rhs/menu.tsx
```diff
@@ -0,0 +1,115 @@
...
+const Icon = styled.div`
+    color: rgba(63, 67, 80, 0.56);
+`;
+
+interface MenuItemProps {
+    className?: string;
+
+    icon: JSX.Element;
+    text: string;
+    onClick: () => void;
```
neallred, line 20: Nit: Consider making `onClick` an optional component and omit passing the noop function down below.

components/channel_info_rhs/rhs.tsx
```diff
@@ -0,0 +1,184 @@
...
+import MoreDirectChannels from 'components/more_direct_channels';
+
+import Menu from './menu';
+import AboutArea from './about_area';
+import TopButtons from './top_buttons';
+import Header from './header';
+
+const Divider = styled.div`
+    width: 88%;
+    border: 1px solid rgba(var(--center-channel-color-rgb), 0.08);
```
neallred, line 27: Looks like the opacity should be `0.04`

components/channel_info_rhs/top_buttons.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    // Copy Button State
+    const copyIcon = copyLink.copiedRecently ? 'icon-check' : 'icon-link-variant';
+    const copyText = copyLink.copiedRecently ? formatMessage({id: 'channel_info_rhs.top_buttons.copied', defaultMessage: 'Copied'}) : formatMessage({id: 'channel_info_rhs.top_buttons.copy', defaultMessage: 'Copy Link'});
+
+    const toggleFavorite = () => {
+        props.toggleFavorite();
+    };
+
+    const toggleMute = () => {
```
neallred, line 95: Ditto earlier comments about preferring passing `props.<the function>` to the component over defining a new function when the new function doesn't do anything additional.

e2e/cypress/integration/channel/channel_info_rhs_spec.js
```diff
@@ -0,0 +1,228 @@
...
+            // # Click on the channel info button
+            cy.get('#channel-info-btn').click();
+
+            // * Verify that we can toggle the mute status
+            cy.uiGetRHS().findByText('Mute').should('be.visible').click();
+            cy.uiGetRHS().findByText('Muted').should('be.visible').click();
+            cy.uiGetRHS().findByText('Mute').should('be.visible');
+        });
+
+        it('should be able to app people', () => {
```
neallred, line 93: Nit: `to app people` -> `to add people`

e2e/cypress/integration/channel/channel_info_rhs_spec.js
```diff
@@ -0,0 +1,228 @@
...
+            // # Go to test channel
+            cy.visit(`/${testTeam.name}/channels/${testChannel.name}`);
+
+            // # Click on the channel info button
+            cy.get('#channel-info-btn').click();
+
+            // * Verify that we can toggle the favorite status
+            cy.uiGetRHS().findByText('Favorite').should('be.visible').click();
+            cy.uiGetRHS().findByText('Favorited').should('be.visible').click();
+            cy.uiGetRHS().findByText('Favorite').should('be.visible');
```
neallred, line 77: We could extract some of the more common actions and assertions into things like `verifiyMute` and `verifyFavorite`. e.g.

```tsx
function verifyFavorite() {
    cy.uiGetRHS().findByText('Favorite').should('be.visible').click();
    cy.uiGetRHS().findByText('Favorited').should('be.visible').click();
    cy.uiGetRHS().findByText('Favorite').should('be.visible');
}
``

File: e2e/cypress/integration/channel/channel_info_rhs_spec.js [hunk 8]
```diff
@@ -0,0 +1,228 @@
...
+// - [*] indicates an assertion (e.g. * Check the title)
+// - Use element ID when selecting an element. Create one if none.
+// ***************************************************************
+
+// Stage: @prod
+// Group: @channel @rhs @channel_info
+
+import {stubClipboard} from '../../utils';
+
+describe('Channel Info RHS', () => {
```
neallred, line 15: Nice tests!

neallred (commented):
Overall looks good to me! The biggest functionality thing is probably adding theme support.

components/channel_info_rhs/about_area.tsx
```diff
@@ -0,0 +1,300 @@
...
+        break;
+    }
+
+    const noDescription = isDirectChannel && dmUser?.user.is_bot;
+
+    return (
+        <Container>
+
+            {channelPurpose && (
+                <ChannelPurpose className={channelPurposeClass ? `ChannelPurpose__${channelPurposeClass}` : ''}>
```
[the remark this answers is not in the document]
JulienTant: nts: `overflow-wrap: anywhere;`

components/channel_info_rhs/header.tsx
```diff
@@ -0,0 +1,70 @@
...
+    );
+
+    return (
+        <div className='sidebar--right__header'>
+            <span className='sidebar--right__title'>
+                <FormattedMessage
+                    id='channel_info_rhs.header.title'
+                    defaultMessage='Info'
+                />
+                {channelName &&
```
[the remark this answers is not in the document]
JulienTant: I can't see where your mouse is so I'm not sure what you're pointing at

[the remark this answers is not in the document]
neallred: ahh, that's true. The "Off-Topic" text on the right. If you hover on it or click on it it changes color.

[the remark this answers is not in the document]
JulienTant: Oh yeah, that's a channel switcher, but the RHS now follows us so I guess its not needed anymore!

components/channel_info_rhs/menu.tsx
JulienTant: the onClick will be mandatory, i just use the noop while I develop to avoid having my linter screaming at me because my tags are missing the onClck option haha!

components/channel_info_rhs/index.ts
```diff
@@ -0,0 +1,89 @@
...
+        isFavorite,
+        isMuted,
+        isInvitingPeople,
+        gmUsers,
+        canManageMembers,
+        canManageProperties,
+    } as ChannelInfoRhsProps;
+
+    switch (channel.type) {
+    case Constants.DM_CHANNEL:
```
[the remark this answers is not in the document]
JulienTant: Yes I'll turn it into an `if`, I used to have much more code down there

components/channel_info_rhs/about_area.tsx
```diff
@@ -0,0 +1,300 @@
...
+    & .status-wrapper {
+        height: 50px;
+    }
+`;
+
+const Usernames = styled.p`
+    font-family: Metropolis, sans-serif;
+    font-size: 18px;
+    line-height: 24px;
+    color: #3F4350;
```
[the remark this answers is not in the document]
JulienTant: Well it's not exactly the same color. I had this issue with quite  a lot of them, I tried doing some matching but I don't know how flexible we sre regarding this. Do I use the figma color or find a close match in the `_css_variables.scss`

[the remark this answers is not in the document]
neallred: Ahh, I picked the wrong one. `--center-channel-text` should be exact. The difference between the RGB and the hex is, we usually use the hex when we don't need transparency, and the `-rgb` variant when we do. In general I'd lean towards picking a close match in `_css_variables.scss`.

components/channel_header/components/header_icon_wrapper.tsx
```diff
@@ -78,6 +78,13 @@ const HeaderIconWrapper: React.FC<Props> = (props: Props) => {
             messageID: t('channel_header.channelFiles'),
             message: 'Channel files',
         },
+        channelInfo: {
```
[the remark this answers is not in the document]
JulienTant: That's part of the behavior of all channel info icons:
https://github.com/mattermost/mattermost-webapp/blob/3d224e9b651bea6ff885eda2255782b045780424/components/channel_header/components/header_icon_wrapper.tsx#L121

matthewbirtch (changes requested):
Nice work on this @JulienTant. Most of my comments are focused on UI polish, but there are a few functional things as well. Let me know if any of the below is unclear.

1. The icon for the `i` button is not the right one. Can you ensure it's using the compass icon component using the `information-outline` icon rather than an SVG? It also looks like it isn't perfectly centered within the button
<table>
<tr>
<td><strong>designed</strong></td>
<td><strong>implemented</strong></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

2. When I choose 'View info' from the channel menu, it should also open the new RHS rather than the modal
image

3. It looks like we might need to update the version of the compass icon font. the mute and unmute icons were updated a couple weeks ago to fix the jumping around that happens when you toggle between muted and unmuted.

4. Can we reduce the delay for the `Copied` state to return back to normal? I think we can cut it in half.

5. The edit icon button for the channel header/purpose is not square - it's wider than it is tall. The button shape should be 24px x 24px. Also, the icon itself should be 14.4px font-size
image

6. For the same edit button, can we make sure it aligns with the margin (see design screenshot with guide line for reference)
image

7. When the channel purpose or header are empty can we change it so they looks like this? Also, the text label should be clickable.


9. Since we updated the language in the RHS to 'Add a channel header', can we do the same in the channel header itself when empty? Right now it still says 'description'

image

10. Update the URL text color to `color: rgba(var(--center-channel-color-rgb), .64);`

11. When the header and purpose are set, we need to reduce the space between them. It looks like there is margin set on the `p` tag within as well as the purpose container that is causing double margin there.

12. For the hover state on the buttons at the top, it looks like there is an animated transition on them. Can we make the hover state immediate without any transition? Or reduce the duration to 200ms or something really short.

13. For the menu items near the bottom, can we remove the padding and the rounded corner so that it goes right to the edge on hover? Sorry, I believe you implemented as designed, but it feels better with it going edge to edge like this:
image

14. In the DM info RHS, the status icon is misaligned and smaller than 
15. Also, it looks the DM is not displaying the name based on the team name preference. It's always showing the username.
16. When the title/position is displayed for the DM user, the content is not vertically centered with the avatar
image

JulienTant:
I pushed a fix for those items:
- [x] The icon for the i button is not the right one. Can you ensure it's using the compass icon component using the information-outline icon rather than an SVG? It also looks like it isn't perfectly centered within the button
**** [x] Can we reduce the delay for the Copied state to return back to normal? I think we can cut it in half.
**** [x] The edit icon button for the channel header/purpose is not square - it's wider than it is tall. The button shape should be 24px x 24px. Also, the icon itself should be 14.4px font-size
**** [x] Since we updated the language in the RHS to 'Add a channel header', can we do the same in the channel header itself when empty? Right now it still says 'description'
**** [x] Update the URL text color to color: rgba(var(--center-channel-color-rgb), .64);
**** [x] For the hover state on the buttons at the top, it looks like there is an animated transition on them. Can we make the hover state immediate without any transition? Or reduce the duration to 200ms or something really short.
**** [x] For the menu items near the bottom, can we remove the padding and the rounded corner so that it goes right to the edge on hover? Sorry, I believe you implemented as designed, but it feels better with it going edge to edge like this:

matthewbirtch:
> I removed the transition, except for the "Copy Link" button. Tell me if you want me to remove the animation for the Copy button too

No, I'm good with the transition on the Copy Link. Although the transition doesn't animate when it goes back to the default. Can we keep the transition for both the copied state and when it returns to the default state?

JulienTant:
**Done**
- [x] In the DM info RHS, the status icon is misaligned and smaller than
- [x] When I choose 'View info' from the channel menu, it should also open the new RHS rather than the modal
➡️ Wow I totally missed that!

- [x] Also, it looks the DM is not displaying the name based on the team name preference. It's always showing the username.
- [x] When the header and purpose are set, we need to reduce the space between them. It looks like there is margin set on the p tag within as well as the purpose container that is causing double margin there.
- [x] When the title/position is displayed for the DM user, the content is not vertically centered with the avatar


**What's left**
- [ ] It looks like we might need to update the version of the compass icon font. the mute and unmute icons were updated a couple weeks ago to fix the jumping around that happens when you toggle between muted and unmuted.
➡️ I'll check with the team to see how to do that

- [ ] For the same edit button, can we make sure it aligns with the margin (see design screenshot with guide line for reference)
➡️ Hum maybe the size issue fixed this because it looks aligned for me (sorry for the red it makes it easier to see 😛 )
image

- [ ] When the channel purpose or header are empty can we change it so they looks like this? Also, the text label should be clickable.
➡️ Just to make sure I understand, you want me to reuse exactly what's done in the channel header right? So 12px grayed text with the pencil appearing on above?

JulienTant:
Hey @matthewbirtch, thanks for all the detailed feedback! I'm done dealing with those so feel free to take a new look.

I added this:
- When opening the RHS for an archived channel, the 'archive' icon will show next to the channel name in the RHS header. It was the case before and some e2e tests existed to make sure that is the case.

matthewbirtch (changes requested):
Thanks @JulienTant. I discovered a few other issues after looking at a few other aspects of the interface.

1. On screens < 768, there is a `i` icon in the header. Tapping that opens up a modal that displays the channel header, but I think we can just make that open up the new info RHS instead
image

2. It looks like we're using the wrong variable for the text color. It should be `center-channe-color-rgb`. This is really noticeable when you turn on a dark theme. Everywhere you're using `center-channel-text-rgb` should be replaced with `center-channel-color-rgb`

image

matthewbirtch (changes requested):
Thanks @JulienTant. I discovered a few other issues after looking at a few other aspects of the interface.

1. On screens < 768, there is a `i` icon in the header. Tapping that opens up a modal that displays the channel header, but I think we can just make that open up the new info RHS instead
image

2. It looks like we're using the wrong variable for the text color. It should be `center-channe-color-rgb`. This is really noticeable when you turn on a dark theme. Everywhere you're using `center-channel-text-rgb` should be replaced with `center-channel-color-rgb`

image

3. For the **hover state** of `Add a channel purpose` or `Add a channel header`, can we also change the color to `color: rgba(var(--center-channel-color-rgb),0.64);`

matthewbirtch:
Sorry for the duplicate review....go with the 2nd one and ignore the 1st.

JulienTant:
> 3\. For the **hover state** of `Add a channel purpose` or `Add a channel header`, can we also change the color to `color: rgba(var(--center-channel-color-rgb),0.64);`

The non-hover color is already `rgba(var(--center-channel-color-rgb),0.64);`, so I changed the hover to `a:0.72` to copy the behavior of the channel header bar. Is that ok?

matthewbirtch:
> > 3. For the **hover state** of `Add a channel purpose` or `Add a channel header`, can we also change the color to `color: rgba(var(--center-channel-color-rgb),0.64);`
> 
> The non-hover color is already `rgba(var(--center-channel-color-rgb),0.64);`, so I changed the hover to `a:0.72` to copy the behavior of the channel header bar. Is that ok?

yep, that works

pablovelezvidal (approved):
LGTM! awesome work @JulienTant

matthewbirtch (approved):
Awesome work on this first PR for the Channel info @JulienTant!

matthewbirtch:
The "Info" title in the RHS header on mobile looks misaligned. It's sitting too high
image

components/channel_info_rhs/about_area_gm.tsx
```diff
@@ -0,0 +1,122 @@
...
+                                userId={user.id}
+                                username={user.username}
+                                channelId={channel.id}
+                                popoverPlacement='left'
+                            />
+                        </ProfilePictureContainer>
+                    ))}
+                </ProfilePictures>
+                <Usernames>
+                    {gmUsers.map((user, i, {length}) => (
```
neallred, line 88: Neat trick for getting the length of users!

neallred (approved):
LGTM! Nice work, @JulienTant !!

esethna (approved):
Nice work!!

JulienTant:
> Minor detail I missed: The "Info" title in the RHS header on mobile looks misaligned. It's sitting too high image

I fixed this! Thanks!

stevemudie (approved):
This is so cool! Nice work @JulienTant!

dontoisme:
@JulienTant I ran into small issue that the info panel is non-scrollable on small screens with an extended header at max length. I'm not sure this is blocking the release but more an FYI.
Screen Shot 2022-03-17 at 9 54 17 AM

e2e/cypress/integration/channel/channel_info_rhs_spec.js
```diff
@@ -0,0 +1,395 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
stevemudie, line 1: @JulienTant Since this is a new E2E it needs a test case number. You can use MM-T4685 here.

JulienTant:
> @JulienTant I ran into small issue that the info panel is non-scrollable on small screens with an extended header at max length. 

I came up with a solution by making the "more" button position dynamic by analyzing the children elements size.

JulienTant:
@stevemudie I'm re-adding the QA Review tag and dismissed your review as discussed in DM.

JulienTant:
Just found and fixed a bug, the gradient of the "more" button was forced to white instead of using a background color:

Before: 
image

After:
image

stevemudie:
@JulienTant I have all of the Zephyr suggestions in place so that's all set.

There is one other thing that was noticed though. In dark mode there's an odd reverse shadow that you might want to look into when there's a situation where there's more to read.

White shadow in dark theme

stevemudie:
> There is one other thing that was noticed though. In dark mode there's an odd reverse shadow that you might want to look into when there's a situation where there's more to read.

NvM I just noticed you fixed it already.

JulienTant:
> @JulienTant I have all of the Zephyr suggestions in place so that's all set.

Thanks! So much conflicts to resolve T_T

> There is one other thing that was noticed though. In dark mode there's an odd reverse shadow that you might want to look into when there's a situation where there's more to read.
> 
> White shadow in dark theme

I found this bug this yesterday and fixed it!

dontoisme (approved):
Looks good to me :D

PR #9906, opened 2022-02-28
MM-33823: fixes deleting entered text on suggestions

koox00:
#### Summary

Currently if you move the caret (left or right) whilst the suggestion box is open
and then hit enter, all text after the caret will get deleted.

This commit keeps the text as it was, without entering the suggestion.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-33823

#### Release Note

```release-note
NONE
```


mickmister (commented):
I'm not sure of this implications of this change, so I'll need to defer review to another dev that's more familiar with this code. Thanks for this fix @koox00 :tada:

jgilliam17:
Shoot, I re-ran the test I missed the earlier post

crspeller (approved):

jgilliam17:
Thanks @koox00 
[E2E report ](https://automation-dashboard.vercel.app/cycles/77865e73-236d-47b0-a01c-584a691c7c38)looks good, no PR related failures.
The message is no longer being deleted when inserting mention.
In one use case cursor jumps to the right and mention fails to be added when doing this:  add `@`, move cursor 1-2 spaces using right arrow and select from the autocomplete See video

https://user-images.githubusercontent.com/52937121/156647964-1a8ce41b-7454-4745-8eed-6580a769840f.mov

matthewbirtch:
> > In one use case cursor jumps to the right and mention fails to be added when doing this: add @, move cursor 1-2 spaces using right arrow and select from the autocomplete See video
> 
> I think that should be the expected one and, that all cases would be like this since you have moved the cursor, what do you think @matthewbirtch?
> 
> @jgilliam17 what's the other behavior that's happening, does it add the mention in some cases?

If you move the cursor to the left or right while the autocomplete is open I wonder if we should just close the autocomplete at that point?

However, in the case shown the video from @jgilliam17, the behavior seems wrong. I think that should be avoidable though if we close the autocomplete once users move the cursor to the left or right when the autocomplete is open

jgilliam17:
@koox00 Apologies, that comment wasn't clear. When adding `@`, followed by few spaces autocomplete remains open and nothing is added if a selection on autocomplete is made.

koox00:
> If you move the cursor to the left or right while the autocomplete is open I wonder if we should just close the autocomplete at that point?

@matthewbirtch what if I want to change a letter in the text?

> However, in the case shown the video from @jgilliam17, the behavior seems wrong

what behavior would be right?

matthewbirtch:
> > If you move the cursor to the left or right while the autocomplete is open I wonder if we should just close the autocomplete at that point?
> 
> @matthewbirtch what if I want to change a letter in the text?

Sorry, I meant if you move the cursor left or right away to another word or empty space

> > However, in the case shown the video from @jgilliam17, the behavior seems wrong
> 
> what behavior would be right?

For starters, if you choose an item from the autocomplete menu it should add that item to the message where the cursor was. In the video example, it didn't add the username and it just moved the cursor to a random location (seemingly)

jgilliam17:
@koox00  I'm removing myself for now. Please add me as a reviewer again when the PR is ready for QA.

lieut-data:

PR #9908, opened 2022-02-28
MM-41827 Count preloaded channels on first team load

hmhealey:
When we switch teams, we preload posts in unread channels so that it'll speed up how long it takes for the user to switch to those channels when the user switches to them. This adds a metric to see how many channels that we're loading on the first time the user views each of their teams.

I added a bit of extra information to track:
1. If this is the first team the user has preloaded channels for. This'll let us differentiate potentially when the user is preloading a lot of channels for their first viewed team because they have a lot of DMs/GMs.
2. Their total number of channels on that team (both channels in the team and DMs/GMs that are cross team). That'll let us determine if they have a lot of unread channels simply because they're in a lot of channels.

I also slightly changed the performance debug messages to be more generic so that they can be used for other types of messages. See the screenshot below for how that looks.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41827

#### Screenshots
Different formats of log messages I tried. The 3rd one is the final result
Screen Shot 2022-02-28 at 5 36 09 PM

#### Release Note
```release-note
Added telemetry for channel preloading
```


M-ZubairAhmed (approved):
This looks a good addition to me in telemetry

M-ZubairAhmed:
@hmhealey this can be QA differed i guess ?

hmhealey:
I was hoping to test this on the test server, but it looks like those metrics aren't sent for test servers, so time to test on production

cwarnermm:
@hmhealey - Is a product documentation updated needed for this PR? Specifically the [Telemetry](https://docs.mattermost.com/manage/telemetry.html) page?

PR #9909, opened 2022-02-28
MM-40662: Fix non-emoji text in markdown not rendered as emoji but added to recent emojis

sinansonmez:
#### Summary
This fixes an issue related to emoji such as ':smile:' is written within markdown. This is not rendered as emoji in the message but it is added to recently used emojis' list. With the changes, the emoji is not added to recently used emojis' list

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/19175

#### Screenshots
**before**

https://user-images.githubusercontent.com/37421564/156071033-9b4cd72e-4a9b-454c-a738-6100acd38061.mov

**after**

https://user-images.githubusercontent.com/37421564/156071062-8b1320b9-1e44-4b85-af47-80b94900cecb.mov

#### Release Note
```release-note
Fixes an issue related to when emoji shortcode is written in markdown, the emoji wasn't rendered as emoji in the message but it was added to recently used emojis.
```

### DIFF
--- a/utils/emoticons.tsx
+++ b/utils/emoticons.tsx
@@ -26,12 +26,13 @@ export const emoticonPatterns: { [key: string]: RegExp } = {
 export const EMOJI_PATTERN = /(:([a-zA-Z0-9_+-]+):)/g;
 
 export function matchEmoticons(text: string): RegExpMatchArray | null {
-    let emojis = text.match(EMOJI_PATTERN);
+    const markdownCleanedText = text.replace(/`(.*?)`/g, '');
+    let emojis = markdownCleanedText.match(EMOJI_PATTERN);
 
     for (const name of Object.keys(emoticonPatterns)) {
         const pattern = emoticonPatterns[name];
 
-        const matches = text.match(pattern);
+        const matches = markdownCleanedText.match(pattern);
         if (matches) {
             if (emojis) {
                 emojis = emojis.concat(matches);


utils/emoticons.tsx
```diff
@@ -26,12 +26,13 @@ export const emoticonPatterns: { [key: string]: RegExp } = {
 export const EMOJI_PATTERN = /(:([a-zA-Z0-9_+-]+):)/g;
 
 export function matchEmoticons(text: string): RegExpMatchArray | null {
-    let emojis = text.match(EMOJI_PATTERN);
+    const markdownCleanedText = text.replace(/`(.*?)`/g, '');
```
sinansonmez, line 29: I can move this part to utils. Since as it is only used here maybe it makes sense to keep it here. Please let me know what do you prefer.

M-ZubairAhmed:
@esethna please verify is the new behavior is what is expected

lieut-data (changes requested):
Thanks, @sinansonmez!

I'll defer to @hmhealey's expertise on all things Markdown, but I think we'll need something more robust than a regular expression pass. For example, I can reproduce this issue using code blocks too:

CleanShot 2022-03-01 at 09 53 34@2x

My suspicion is that we actually need to render the markdown first and only extract emoticons from those segments of the parsed text that are outside of these undesired cases.

esethna (approved):
I'll defer to dev on the technical implementation, but from a behavioral perspective this makes sense (ie don't add to recent emojis if it isn't actually rendered in the message)

hmhealey (changes requested):
Similar to @lieut-data's comment, I've also found that it's possible to include emojis in links as well by posting something like
```
[link](http://example.com/:grinning:)
```

Instead of just using a simple replace though, we could probably use a version of the Markdown renderer for this though since it knows what text actually appears in the post. If you take a look at the [containsAtChannel](https://github.com/mattermost/mattermost-webapp/blob/master/utils/post_utils.ts#L138) utility, it uses a modified `MentionableRenderer` to remove all text that's ignored like code blocks and the URL of links

sinansonmez:
Thanks for the guidance @hmhealey now I see your point about renderer (sorry for repetition).

I actually tested bunch of option including link, code span (single line and multiline), list, list item, br, em, paragraph etc. `MentionableRenderer` does the job so I don't need to create another renderer only for this. I updated the PR accordingly 

I kindly ask QA engineer to test all the possible cases throughly (as they ussually do 💯) such as table etc.

hmhealey (approved):
Thanks! This looks good to me now. I think we already have other tests that cover the MentionableRenderer's use in that other function, so I'd assume all those cases will work here as well

lieut-data (changes requested):
Thanks for this, @sinansonmez! Would you be willing to extend `utils/emoticons.test.tsx` with a test that covers this use case, helping us to make sure we don't regress this case in the future?

sinansonmez:
Thanks @lieut-data I added tests to cover this use case

lieut-data (approved):
Brilliant, thank you!!

sinansonmez:
thanks @lieut-data and @hmhealey will you also request QA review from someone so after QA it can be merged?

lieut-data:
Yep, I've asked @jgilliam17 to help with QA (or route as appropriate!)

jgilliam17:
Thanks @sinansonmez 
Tested, looks good to merge. 
- Verified emoji text, non-rendered, used in markdowns, code, links, tables etc. is not added to recently used emoji
[E2E report](https://mattermost-cypress-report.s3.amazonaws.com/167365-726c449-onprem-ent-webapp-pr-9909-mattermost/mm-ee-test726c449/mochawesome.html) looks good, no PR related failures.

jgilliam17 (approved):

PR #9917, opened 2022-03-01
MM-41892 Fix untranslatable text in search results

hmhealey:
Straightforward enough. Those IDs couldn't be identified by `make i18n-extract` without the `t` function because they aren't passed directly into `formatMessage` or one of the similar translation methods.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41892

#### Release Note
```release-note
Made text in search results translatable
```


neallred (approved):
:ship:

julmondragon (approved):
Thank you @hmhealey !

PR #9930, opened 2022-03-03
MM-42278 Initial Playwright setup for visual testing

saturninoabril:
#### Summary
This PR introduced Playwright initially for visual testing, then eventually for accessibility and frontend performance testing. Once it reached stability and maturity based on our use case, completed other important items like reporting, integration, pipeline, etc. and accepted for dev/community to write tests, it will be used to automate tests of new features while Cypress will go on maintenance mode. Files and folders are to be found at `/e2e/aesthetic/`.  Used "aesthetic" as it will test initially the "beauty" of the webapp but very much open to suggestion for a better name.

Files and folders of Cypress were moved to `/e2e/functional/`.

Visual testing is known to be problematic for elements that render dynamic information like date/time, name of teams/channels/users, etc. To deal with it, I introduced option on that makes it easier to hide specific element thru new props of custom theme (`hideInVisualTesting`). There are other ways to do that like specifically adding new user preference, but I thought it makes sense to extend the use of custom theme.

I've extended `Client4` and made some changes to its properties and relocated couple of helper functions to make it compatible as node client. `Client4` was more of a browser client, though.

For local development and in-house visual test run, screenshots were created using the official [Playwright docker container](https://playwright.dev/docs/docker) which uses Linux/Ubuntu (focal) distribution. Using such container makes the screenshot consistent even with developers using different local environment and will match with the same container we'll use in the pipeline.

To test screenshots in wide-range of different environments (OS, browser and mobile clients) and easy enough for QA to handle during release testing, I've added the ability to run both using [applitools](https://applitools.com/) and [percy](https://percy.io/). Nothing official yet but both provided so we'll get enough time to evaluate which tool is best for us.

To follow in separate PR:
- Support for Cloud server
- Support with Calls product
- Reporting - to community, dashboard and Zephyr
- Gitlab pipeline
- Docs on best practices, semi-opinionated patterns and styles

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42278

#### Related Pull Requests
- Has server changes ([link here](https://github.com/mattermost/mattermost-server/pull/19681))

#### Release Note
```release-note
NONE
```


hmhealey:
I haven't had a chance to look through all this, but with the number of changes in this PR, is it possible for us to break it down any further? Also, I've actually got similar changes to Client4 that I've been working on which will conflict with these, so we'll have to consolidate those.

As for dealing with those problematic elements, I wonder if we could instead have a fixed username and time somehow since I feel like hiding elements of the page could cause other problems by having the tests not reflect the actual UI. If we're doing that though, I'd rather do it by injecting CSS directly into the web page instead of adding something into the theme to do that. Looking at the docs for Playwright, it sounds like we might be able to use [something like this](https://playwright.dev/docs/api/class-page#page-add-style-tag) to add those rules without changing the theming code

saturninoabril:
@hmhealey Thanks for the general review! I haven't had a chance to work/look into this but will try within this week.

Sorry for the massive changes. I do understand the overhead in reviewing the changes, unfortunately, they are related to each other and I thought it's the minimum change I could provide. The changes typically were:
- moved `/e2e` to `/e2e/functional/`
- modify `Client4` to be compatible with node client
- basic Playwright setup at `/e2e/aesthetic/`
- few tests with screenshots
And submitting separately will not make sense since they are related to each other. I hope it's fine but let me know how you'd like to proceed.

Well noted on your (upcoming) changes in Client4.

We can use fixed values but at one point, we'll still going to deal with dynamic element. Besides, randomness will help us with test isolation as well. I didn't know there's a [way/feature](https://playwright.dev/docs/api/class-page#page-add-style-tag) to do that directly in Playwright. Thanks for sharing and I'll take a look.

saturninoabril:
@Phrynobatrachus @furqanmlk Adding you for quick check and feel free to post comments.

PR #9949, opened 2022-03-05
MM-41824 : Add telemetry to count the number of request on page load

M-ZubairAhmed:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41822

#### Related Pull Requests
N/A

#### Screenshots
https://user-images.githubusercontent.com/17708702/158206708-600fb416-d7d6-4692-a194-d30d3229994e.mov

#### Release Note

```release-note
Telemetry added for number of api requests on initial page load.
```


actions/telemetry_actions.jsx
```diff
@@ -92,22 +99,70 @@ export function measure(name1, name2) {
...
+        if (entry.initiatorType === 'xmlhttprequest' || entry.initiatorType === 'fetch') {
+            totalRequests += 1;
+            maxResourceSize = Math.max(maxResourceSize, entry.encodedBodySize);
+        }
+    });
+
+    clearResourceTimings();
+
+    if (totalRequests > 0) {
+        trackEvent('pageRequests', initiatedBy, {
```
M-ZubairAhmed, line 159: open to suggestion on category `pageReqests`

actions/telemetry_actions.jsx
```diff
@@ -92,22 +99,70 @@ export function measure(name1, name2) {
...
+        return;
+    }
+
+    let totalRequests = 0;
+    let maxResourceSize = 0; // in Bytes
+
+    performance.getEntriesByType('resource').forEach((entry) => {
+        if (entry.initiatorType === 'xmlhttprequest' || entry.initiatorType === 'fetch') {
+            totalRequests += 1;
+            maxResourceSize = Math.max(maxResourceSize, entry.encodedBodySize);
```
M-ZubairAhmed, line 152: does sending max resource size makes sense, i thought it would give an idea on how large resource is, ?

M-ZubairAhmed: Any other telemetry derived from resource timing api is can send ?

actions/telemetry_actions.jsx
```diff
@@ -92,22 +99,70 @@ export function measure(name1, name2) {
...
+    }
+
+    performance.clearResourceTimings();
+}
+
+/**
+ * Send an telemetry for number of requests to the server for an event.
+ * @param {String} initiatedBy The name of the event that initiated the tracking
+ */
+export function trackNumOfRequestEvent(initiatedBy) {
```
jwilander, line 141: Only concern with this is that it's not tied to the "initiatedBy". So if this function is called in two different places around the same time the metrics could get all messed up for. For example, if immediately after page load someone does a channel switch we might report all the requests as part of the channel switch if the page load timeout hasn't fired yet. Maybe when we "mark" the start of an event like `page_load` we only let `trackNumOfRequestEvent` run when it's called with `page_load`. Essentially a lock on the function. Not sure if it's necessary or overkill. What do you think?

actions/telemetry_actions.jsx
```diff
@@ -28,9 +31,13 @@ export function shouldTrackPerformance(state = store.getState()) {
...
     Client4.trackEvent(category, event, props);
-    if (isDevMode() && category === 'performance' && props) {
-        // eslint-disable-next-line no-console
-        console.log(`${event}: ${props.duration}ms, fresh: ${props.fresh}`);
+    if (isDevMode() && (category === 'performance' || category === 'pageRequests') && props) {
+        const propKeys = Object.keys(props);
+        if (propKeys.length > 0) {
+            const propsAsString = propKeys.map((key) => `${key}: ${props[key]}`).join(', ');
+            // eslint-disable-next-line no-console
+            console.log(`${event} -> ${propsAsString}`);
```
hmhealey, line 39: I also made a change like this in https://github.com/mattermost/mattermost-webapp/pull/9908, so that's definitely going to conflict 😛

actions/telemetry_actions.jsx
```diff
@@ -92,22 +99,70 @@ export function measure(name1, name2) {
...
     // Must be wrapped in setTimeout because loadEventEnd property is 0
     // until onload is complete, also time added because analytics
     // code isn't loaded until a subsequent window event has fired.
     const tenSeconds = 10000;
     setTimeout(() => {
-        const {loadEventEnd, navigationStart} = window.performance.timing;
-        const pageLoadTime = loadEventEnd - navigationStart;
-        trackEvent('performance', 'page_load', {duration: pageLoadTime});
+        trackPageLoadTimeEvent();
+        trackNumOfRequestEvent('page_load');
```
hmhealey, line 122: Perhaps instead of having two separate events for this, we could add these into the events for page_load. channel_switch, and team_switch? So instead of sending just `duration`, we also add `totalRequests` and `maxResourceSize`

actions/telemetry_actions.jsx
hmhealey: Instead of clearing the timings between calls for this, if we tied this to the page_load/channel_switch event, we could count only resource entries that occurred within the start and end of that event by using the `startTime` and `duration`. That should allow us to count requests for two events in parallel

actions/telemetry_actions.jsx
hmhealey: I wonder if `encodedBodySize` or `decodedBodySize` would be more helpful to have here? I assume `encodedBodySize` includes like gzip compression if that's enabled on the server. I'm trying to think about which we'll care more about since encoded means that it'll be more realistic, but it'll then be influenced by server settings.

Then again, all of this will be influenced by server settings like CRT.

actions/telemetry_actions.jsx
M-ZubairAhmed: ha like minds :P

actions/telemetry_actions.jsx
M-ZubairAhmed: changed it to `performance` itself

components/post_view/post_list/post_list.jsx
```diff
@@ -4,36 +4,47 @@
...
         'TeamLink#click',
         'PostList#component',
     ]);
 
-    if (dur1 !== -1) {
-        trackEvent('performance', 'channel_switch', {duration: Math.round(dur1), fresh});
+    // If channel was switched by clicking on sidebar channel link
+    if (channelSwitchDuration !== -1) {
+        trackEvent('performance', 'channel_switch', {duration: Math.round(channelSwitchDuration), fresh});
+        trackNumOfRequestEvent('channel_switch');
```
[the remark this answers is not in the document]
M-ZubairAhmed: yes it would make sense but how much of setTimeout we should put. How do we decide that 🤔

actions/telemetry_actions.jsx
M-ZubairAhmed: not handling in this pr

actions/telemetry_actions.jsx
```diff
@@ -104,7 +112,24 @@ export function trackLoadTime() {
...
         const {loadEventEnd, navigationStart} = window.performance.timing;
         const pageLoadTime = loadEventEnd - navigationStart;
-        trackEvent('performance', 'page_load', {duration: pageLoadTime});
+
+        let numOfRequest = 0;
+        let maxAPIResourceSize = 0; // in Bytes
+        let longestAPIResource = '';
+        let longestAPIResourceDuration = 0;
+        performance.getEntriesByType('resource').forEach((resourceTimingEntry) => {
+            if (resourceTimingEntry.initiatorType === 'xmlhttprequest' || resourceTimingEntry.initiatorType === 'fetch') {
```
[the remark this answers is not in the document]
M-ZubairAhmed: This gives almost exact number of request on page load +/- 2

components/root/root.jsx
```diff
@@ -280,6 +280,13 @@ export default class Root extends React.PureComponent {
                 prevProps.history.push('/terms_of_service');
             }
         }
+
+        performance.addEventListener('resourcetimingbufferfull', this.handleResourceTimingBufferFull);
+    }
+
+    handleResourceTimingBufferFull = () => {
+        // eslint-disable-next-line no-console
+        console.log('Resource timing buffer full');
```
jwilander, line 289: Is this still needed?

M-ZubairAhmed: oh yeah will remove it

actions/telemetry_actions.jsx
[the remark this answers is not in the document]
M-ZubairAhmed: Also that(xhr/fetch) is the target we are trying to reduce with Graphql

jwilander (approved):
LGTM

hmhealey (approved):

cwarnermm:
@M-ZubairAhmed - Is a product documentation updated needed for this PR? Specifically the [Telemetry](https://docs.mattermost.com/manage/telemetry.html) page?

hmhealey:
I commented [here](https://github.com/mattermost/mattermost-webapp/pull/9908#issuecomment-1087687732) on the same thing just to keep the conversation in one place

PR #9969, opened 2022-03-08
[MM-42322] add thread auto follow setting in system console

AshishDhama:
#### Summary
Add thread auto follow setting above collapsed reply threads setting in system console, ``ThreadAutoFollow`` needs to be true to enable Collapsed reply threads, exposing that to UI. 

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-42322

#### Related Pull Requests
``NONE``

#### Screenshots
Screenshot 2022-03-10 at 10 06 44 AM

#### Release Note
```release-note
UI: exposing threadAutoFollow config setting to UI
```


michelengelen (approved):
LGTM

components/admin_console/admin_definition.jsx
```diff
@@ -6052,6 +6052,16 @@ const AdminDefinition = {
...
+                        label_default: 'Automatically Follow Threads',
+                        help_text: t('admin.experimental.threadAutoFollow.desc'),
+                        help_text_default: 'This setting must be enabled to support [Collapsed Reply Threads](https://docs.mattermost.com/messaging/organizing-conversations.html) and may impact your database server performance. If you cannot easily scale up and tune your database, or if you are running the Mattermost application server and database server on the same machine, we recommended disabling ThreadAutoFollow until Collapsed Reply Threads is promoted to general availability in Q2 2022. Learn more about these [performance considerations here](https://support.mattermost.com/hc/en-us/articles/4413183568276). When enabled, threads a user starts, participates in, or is mentioned in are automatically followed. Entries are added to the ThreadMembership table to track followed threads for each user and the read or unread state of each followed thread. Enabling this configuration setting doesn’t retroactively follow threads for actions taken prior to the setting being enabled.',
```
jfrerich, line 6061:
```suggestion
                        help_text_default: 'This setting must be enabled to support [Collapsed Reply Threads](https://docs.mattermost.com/messaging/organizing-conversations.html) and may impact your database server performance. If you cannot easily scale up and tune your database, or if you are running the Mattermost application server and database server on the same machine, we recommended disabling `ThreadAutoFollow` until Collapsed Reply Threads is promoted to general availability in Q2 2022. Learn more about these [performance considerations here](https://support.mattermost.com/hc/en-us/articles/4413183568276). When enabled, threads a user starts, participates in, or is mentioned in are automatically followed. Entries are added to the `ThreadMembership` table to track followed threads for each user and the read or unread state of each followed thread. Enabling this configuration setting doesn’t retroactively follow threads for actions taken prior to the setting being enabled.',
```

jfrerich: I didn't test this, but I think you can `codeblock` the `ThreadAutoFollow` and `ThreadMembership` so they appear as shown in the Jira ticket.

jfrerich (changes requested):
There are two pieces of text that need to be code blocked in order to exactly match the text in the Jira ticket description.

jfrerich:
There is an extra `for` that snuck in on that last commit 😃 

image

AshishDhama:
@jfrerich thanks

jfrerich (approved):
Looks nice!

jfrerich:
I would update the image at the top of the PR too.  Thanks for the fixes!

esethna (changes requested):
@AshishDhama please add a line break between the two paragraphs

Also on the test server it seems the setting is in a disabled state since it says it's controlled via env variable? Is that just a problem on the test server? We want any admins to be able to easily toggle the setting from the console

image

esethna:
Also @AshishDhama can we remove the ` in Q2 2022` part of the sentence. I'm thinking it might be best not to "hard code" any dates into the text

AshishDhama:
> Also on the test server it seems the setting is in a disabled state since it says it's controlled via env variable? Is that just a problem on the test server? We want any admins to be able to easily toggle the setting from the console

@esethna Yeah this is a warning as we cannot override setting if it’s set via ENV variable, which we requested https://mattermost.atlassian.net/browse/MM-42108 (link to conversation with SRE team)
cc// @angeloskyratzakos Please correct me if I have wrong understanding.

esethna:
@AshishDhama just to understand, what's expected in production cloud deployments and production on-prem deployments? 

For production cloud deployments I would actually expect that this setting is not shown at all (since it only affects backend resources which we ourselves manage). Sorry if that wasn't clear in the ticket, my fault

For production on-prem, this setting should be togglable via the system console UI

AshishDhama:
> I'm not familiar with how the spinwick test servers are deployed, what matters most to me is understanding what will be expected in production. Just to understand, can you clarify what's expected in **production cloud deployments** and **production on-prem deployments**?
> 
> For production cloud deployments I would actually expect that this setting is not shown at all (since it only affects backend resources which we manage ourselves for all cloud instances). Sorry if that wasn't described in the ticket, my fault
> 
> For production on-prem, this setting should be togglable via the system console UI

@marianunez @neallred Could you guys please guide me here.
@esethna Does it make sense to hide Collapsed reply threads option as well. I mean user cannot enable it with ``ThreadsAutoFollow``.

neallred:
For on prem, this value will be controllable by admins. For both test and production cloud, it is overridden and admins won't be able to control it, so agree that  I think we can hide it for cloud environments. Looks like [the code that generates the UI](https://github.com/mattermost/mattermost-webapp/blob/master/components/admin_console/admin_console.tsx#L143-L148) supports hiding items by defining an `isHidden` function property that gets called with the license. Typically we check for whether something is cloud by seeing if `license.Cloud === 'true'`.

michelengelen:
good call @neallred ... it is located in `components/admin_console/admin_definition.jsx` ... thats probably the best way to handle that

esethna (approved):
Makes sense @neallred, thanks for the clarification. In that case I'll approve to unblocking merging.

AshishDhama:
@esethna Have tested on suggested server it's working.

cwarnermm:
Documented via https://github.com/mattermost/docs/pull/5520

